PR #2921, opened 2024-03-15
Fix storage emulator connection string for blobs

davidfowl:
- This broke after the CDK merge. We were missing test coverage.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2921)

### DIFF
--- a/tests/Aspire.Hosting.Tests/Azure/AzureBicepResourceTests.cs
+++ b/tests/Aspire.Hosting.Tests/Azure/AzureBicepResourceTests.cs
@@ -117,6 +117,24 @@ public class AzureBicepResourceTests
     }
 
     [Fact]
+    public async Task AddAzureCosmosDBEmulator()
+    {
+        var builder = DistributedApplication.CreateBuilder();
+
+        var cosmos = builder.AddAzureCosmosDB("cosmos").RunAsEmulator(e =>
+        {
+            e.WithEndpoint("emulator", e => e.AllocatedEndpoint = new(e, "localost", 10001));
+        });
+
+        Assert.True(cosmos.Resource.IsContainer());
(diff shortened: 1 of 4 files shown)


tests/Aspire.Hosting.Tests/Azure/AzureBicepResourceTests.cs
```diff
@@ -116,6 +116,24 @@ public async Task AssertManifestLayout()
...
+        {
+            e.WithEndpoint("emulator", e => e.AllocatedEndpoint = new(e, "localost", 10001));
+        });
+
+        Assert.True(cosmos.Resource.IsContainer());
+
+        var cs = AzureCosmosDBEmulatorConnectionString.Create(10001);
+
+        Assert.Equal(cs, cosmos.Resource.ConnectionStringExpression.ValueExpression);
+        Assert.Equal(cs, await ((IResourceWithConnectionString)cosmos.Resource).GetConnectionStringAsync());
```
eerhardt, line 134: It's odd that these 2 strings are the same. I would have thought the ValueExpression wouldn't have the port in it.

eerhardt (approved):
LGTM

tests/Aspire.Hosting.Tests/Azure/AzureBicepResourceTests.cs
davidfowl: Yes this is strange. It's because we don't build the right connection string for the manifest for these emulators. I'll look at it in a follow up PR (just for clarity)

mitchdenny (approved):

PR #2931, opened 2024-03-15
DcpDependencyCheck: throw instead of crashing process

### LINKED ISSUE #2928: Throw instead of calling `Environment.Exit` in `DcpDependencyCheck`
If the cancellation token passed to `DcpDependencyCheck.EnsureDcpDependenciesAsync(...)` is canceled before completion, this will currently crash the process. Instead, it should clean up and propagate the cancellation up the stack.

This is causing testing to crash when the test builder is not used as intended. We should not terminate the process this way, or we should provide a way to opt-out of crashing the process. There are 4 `Environment.Exit` calls in `DcpDependencyCheck`. They should all be converted into exceptions which propagate up the stack. This would improve the debugging experience.

Are the error codes passed to `Environment.Exit` important?

ReubenBond:
Fixes #2928
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2931)

### DIFF
--- a/src/Aspire.Hosting.Testing/DistributedApplicationEntryPointInvoker.cs
+++ b/src/Aspire.Hosting.Testing/DistributedApplicationEntryPointInvoker.cs
@@ -97,10 +97,6 @@ internal sealed class DistributedApplicationEntryPointInvoker
                     // build to throw
                     _appTcs.TrySetException(new InvalidOperationException($"The entry point exited without building a {nameof(DistributedApplication)}."));
                 }
-                catch (TargetInvocationException tie) when (tie.InnerException?.GetType().Name == "HostAbortedException")
-                {
-                    // The host was stopped by our own logic
-                }
                 catch (TargetInvocationException tie)
                 {
                     exception = tie.InnerException ?? tie;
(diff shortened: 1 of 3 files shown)


danegsta:
We return those particular error codes to allow IDEs to tell if Aspire failed due to a missing container runtime requirements (they currently drive warnings in VS, for example).

ReubenBond:
Do the IDEs do this currently? The issue with Environment.Exit is that it cannot be intercepted and a developer just sees a crash, maybe with an error code.
Currently, it crashes on cancellation, not just on actual dependency issues. Cancellation is normal/expected.

davidfowl:
This should throw an exception and then have it fail with the specific exit code when running from the IDE.

ReubenBond:
> This should throw an exception and then have it fail with the specific exit code when running from the IDE. Using the env var.

Is this something I can do in this PR? What do I search for? I'm not aware of an env var

danegsta:
We could have VS set an environment variable when running F5/CTRL+F5; alternatively, we could add the ability to apply an exit code/reason to the exception and use a last chance exception handler to exit with the appropriate code if found.

davidfowl:
Lets close on the approach tomorrow. I also need to change this code so that the container runtime makes it all the way out of the dependency check so it can be used later.

davidfowl:
Closing in favor of https://github.com/dotnet/aspire/pull/3026.

PR #2932, opened 2024-03-15
Throw when accessing DistributedApplicationExtensions before application has started

### LINKED ISSUE #2901: Make DistributedApplicationExtensions methods throw if DistributedApplication instance isn't started
The `DistributedApplicationTestingBuilder` type is available to facilitate writing integration tests against an AppHost project, however using it correctly can be a bit difficult as a series of methods must be called in the correct order to get a successful result. It would be helpful if methods like `GetHttpClient` and `GetEndpoint` on `DistributedApplicationExtensions` would throw if they're called on an instance of `DistributedApplication` that hasn't been started yet rather than the underlying call to `GetEndpointUriStringCore` failing with an empty endpoint name because it hasn't been allocated yet.

For example, the following test will fail with the message `Endpoint '' for resource 'mywebapp1' not found.` because it's missing the call to `await app.StartAsync()`:

```csharp
[Fact]
public async Task HttpClientGetTest()
{
    var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Program>();
    // Missing call to await app.StartAsync() here

    var httpClient = app.CreateHttpClient("mywebapp1");
    var result1 = await httpClient.GetFromJsonAsync<WeatherForecast[]>("/weatherforecast");
    Assert.NotNull(result1);
    Assert.True(result1.Length > 0);
}
```

ReubenBond:
Fixes #2901

Note that this relies on https://github.com/dotnet/aspire/issues/2928
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2932)


davidfowl (approved):

src/Aspire.Hosting.Testing/DistributedApplicationExtensions.cs
```diff
@@ -98,6 +100,15 @@ private static string GetEndpointUriStringCore(DistributedApplication app, strin
         return endpoint.Url;
     }
 
+    static void ThrowIfNotStarted(DistributedApplication app)
+    {
+        var lifetime = app.Services.GetRequiredService<IHostApplicationLifetime>();
+        if (!lifetime.ApplicationStarted.IsCancellationRequested)
+        {
+            throw new InvalidOperationException("The application has not started.");
```
DamianEdwards, line 108: These message should be in resources now so we can localize. Can we make this a little more descriptive, e.g. "The application must be started before resolving endpoints or connection strings."

ReubenBond: Pushed update. PTAL. I thought we stopped localizing exception messages.

DamianEdwards (approved):

PR #2944, opened 2024-03-16
Make HostNameResolver internal

davidfowl:
- We should be resolving the container host internally and exposing it on endpoint references
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2944)

### DIFF
--- a/src/Aspire.Hosting/Utils/HostNameResolver.cs
+++ b/src/Aspire.Hosting/Utils/HostNameResolver.cs
@@ -8,7 +8,7 @@ namespace Aspire.Hosting.Utils;
 /// <summary>
 /// Helpers to resolve host names when running in a containers.
 /// </summary>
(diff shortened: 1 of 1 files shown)


mitchdenny (approved):

src/Aspire.Hosting/Utils/HostNameResolver.cs
```diff
@@ -8,7 +8,7 @@ namespace Aspire.Hosting.Utils;
 /// <summary>
 /// Helpers to resolve host names when running in a containers.
 /// </summary>
-public class HostNameResolver
+internal class HostNameResolver
```
eerhardt, line 11: `internal static class HostNameResolver`

No reason to ever create one of these.

PR #2949, opened 2024-03-16
Fix for invalid parameter names.

### LINKED ISSUE #2783: Do not allow `-` for builder.AddBicepTemplate().WithParameter() function.
Something like this is currently valid for the AppHost code:

```csharp
var bicep1 = builder.AddBicepTemplate("test", "test.bicep")
                   .WithParameter("param-01", parameter)
```

However, `param-01` is an invalid name for a parameter and the program should return an error here.

mitchdenny:
Fixes #2783
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2949)

### DIFF
--- a/src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
+++ b/src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
@@ -7,7 +7,9 @@ namespace Aspire.Hosting.Azure.Utils;
 
 internal static partial class BicepParameterNameValidator
 {
-    [GeneratedRegex("^[a-z_]+[a-z0-9_]*$", RegexOptions.IgnoreCase)]
+    // See rules from Bicep's highlightjs implementation:
+    // https://github.com/Azure/bicep/blob/a992bdf2d4d7c5c7dec684b7d0de4db9cb260f8a/src/highlightjs/src/bicep.ts#L12
+    [GeneratedRegex("^[a-z_][a-z0-9_]*$", RegexOptions.IgnoreCase)]
     private static partial Regex GetBicepParameterExpression();
 
     internal static void ThrowIfInvalid(string bicepParameterName)
@@ -17,7 +19,7 @@ internal static partial class BicepParameterNameValidator
         if (!regex.IsMatch(bicepParameterName))
         {
             throw new ArgumentException(
-                "Bicep parameter names must only contain alpha, numeric, and _ characters with a leading alpha character.",
+                "Bicep parameter names must only contain alpha, numeric, and _ characters and must start with an alpha or _ characters.",
                 nameof(bicepParameterName)
                 );
         }
--- a/src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
+++ b/src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
@@ -1,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Azure.Utils;
+
+internal static partial class BicepParameterNameValidator
+{
+    [GeneratedRegex("^[a-z_]+[a-z0-9_]*$", RegexOptions.IgnoreCase)]
+    private static partial Regex GetBicepParameterExpression();
+
+    internal static void ThrowIfInvalid(string bicepParameterName)
+    {
+        var regex = GetBicepParameterExpression();
+
+        if (!regex.IsMatch(bicepParameterName))
+        {
+            throw new ArgumentException(
+                "Bicep parameter names must only contain alpha, numeric, and _ characters with a leading alpha character.",
+                nameof(bicepParameterName)
+                );
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


davidfowl:
This doesn't cover setting the values directly into the dictionary.

mitchdenny:
No, it doesn't. We would need to either sub-class `Dictionary<K, T>` for that or perhaps add the following methods to the class and not expose the underlying dictionary. We'd need `SetParameter(...), `GetParameter(key)` and `GetParameters()`.

src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
```diff
@@ -0,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Azure.Utils;
+
+internal static partial class BicepParameterNameValidator
+{
+    [GeneratedRegex("^[a-z_]+[a-z0-9_]*$", RegexOptions.IgnoreCase)]
```
davidfowl, line 10: Point to a document that explains this.

vhvb1989: This is related: https://github.com/Azure/bicep/issues/1494#issuecomment-776128864

Looks like bicep has some identifier's rules similar to C#.

However, I didn't find specific naming rules on bicep docs. It is only seen from src: https://github.com/Azure/bicep/blob/a992bdf2d4d7c5c7dec684b7d0de4db9cb260f8a/src/highlightjs/src/bicep.ts#L12

src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+    private static partial Regex GetBicepParameterExpression();
+
+    internal static void ThrowIfInvalid(string bicepParameterName)
+    {
+        var regex = GetBicepParameterExpression();
+
+        if (!regex.IsMatch(bicepParameterName))
+        {
+            throw new ArgumentException(
+                "Bicep parameter names must only contain alpha, numeric, and _ characters with a leading alpha character.",
```
vhvb1989, line 20: nit. It can start with `_`, like `_foo`

tests/Aspire.Hosting.Tests/Azure/AzureBicepProvisionerTests.cs
```diff
@@ -84,6 +84,38 @@ public async Task ResourceWithTheSameBicepTemplateAndParametersHaveTheSameCheckS
...
+    [InlineData("alpha")]
+    [InlineData("a1pha")]
+    [InlineData("_alpha")]
+    [InlineData("__alpha")]
+    [InlineData("alpha1_")]
+    [InlineData("Alpha1_A")]
+    public void WithParameterAllowsParameterNamesWhichAreValidBicepIdentifiers(string bicepParameterName)
+    {
+        var builder = DistributedApplication.CreateBuilder();
+        builder.AddAzureConstruct("construct", _ => { })
```
vhvb1989, line 115: I wish we could shift-left the use-case when a parameter is added but it does not exists.  For example:

```bicep
var bicep1 = builder.AddBicepTemplate("test", "test.bicep")
                   .WithParameter("paramFoo", parameter)
```

Can we make .netAspire to check if `paramFoo` exists on `test.bicep` ?
Can you just load the `test.bicep` as string to a variable and run `string.Contains(bicepString, "param paramFoo") and fail if that's false?

That would help users a lot to quickly detect typos; instead of waiting until azd is trying to compile the output bicep file.

davidfowl: Too hacky. We could builld an analyzer but we still need a runtime check.

src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
mitchdenny: I think the regex in the highlightjs implementation might be the most concise. I actually asked Open AI for a BNF for Bicep so I could figure it out, but it said that it couldn't source an authoritative document and instead inferred it from usage.

tests/Aspire.Hosting.Tests/Azure/AzureBicepProvisionerTests.cs
mitchdenny: There is also no guarantee that the file exists at compile time.

davidfowl: This still needs a comment though. You can say it's from the bicep bnf

src/Aspire.Hosting.Azure/Utils/BicepParameterNameValidator.cs
davidfowl: This still needs a comment though. You can say it's from the bicep bnf

danmoseley, line 10: Culture invariant also

davidfowl (approved):

PR #2950, opened 2024-03-16
Change preview attribute to experimental.

### LINKED ISSUE #2918: Use ExperimentalAttribute
https://github.com/dotnet/aspire/pull/2875 brought in "preview" APIs (specifically Azure CDK APIs) that we want to warn consumers may change in a later release. These APIs have been marked with `[RequiresPreviewFeatures]`.

We should instead be using `[ExperimentalAttribute]` on these APIs. This attribute fits this scenario more. You can read about the distinction between the two attributes at https://github.com/dotnet/runtime/issues/77869.

cc @mitchdenny @joperezr @terrajobst

mitchdenny:
Fixes #2918 

Also changes the overloads with callbacks exposing CDK types to not be optional. In all cases we have an overload that does not expose the callback so we don't need it to be optional.

### DIFF
--- a/src/Aspire.Hosting.Azure/Extensions/AzureAppConfigurationExtensions.cs
+++ b/src/Aspire.Hosting.Azure/Extensions/AzureAppConfigurationExtensions.cs
@@ -23,7 +23,7 @@ public static class AzureAppConfigurationExtensions
     public static IResourceBuilder<AzureAppConfigurationResource> AddAzureAppConfiguration(this IDistributedApplicationBuilder builder, string name)
     {
 #pragma warning disable ASPIRE0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
-        return builder.AddAzureAppConfiguration(name, (_, _, _) => { });
+        return builder.AddAzureAppConfiguration(name, null);
 #pragma warning restore ASPIRE0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
     }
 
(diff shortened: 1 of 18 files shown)


src/Aspire.Hosting.Azure/Extensions/AzureAppConfigurationExtensions.cs
```diff
@@ -47,12 +47,9 @@ public static IResourceBuilder<AzureAppConfigurationResource> AddAzureAppConfigu
...
 
-            if (configureResource != null)
-            {
-                var resource = (AzureAppConfigurationResource)construct.Resource;
-                var resourceBuilder = builder.CreateResourceBuilder(resource);
-                configureResource(resourceBuilder, construct, store);
-            }
+            var resource = (AzureAppConfigurationResource)construct.Resource;
+            var resourceBuilder = builder.CreateResourceBuilder(resource);
+            configureResource(resourceBuilder, construct, store);
```
eerhardt, line 52: We should be throwing an ArgumentNullException if someone passes in null, and not letting a NullReferenceException occur.

Personally, I would just keep the signature `Action<IResourceBuilder<AzureAppConfigurationResource>, ResourceModuleConstruct, AppConfigurationStore>? configureResource`. You can drop the default `= null` if you want.

Then above, you can just pass in `null` instead of a dummy `(_, _, _) => { }`

mitchdenny: Done. Although I find something deeply unsettling about passing null into a method these days ;)

davidfowl:
Can you do this too while you're here https://github.com/dotnet/aspire/issues/2952

davidfowl (approved):

src/Aspire.Hosting.Azure/Extensions/AzureSearchExtensions.cs
```diff
@@ -71,12 +71,9 @@ void ConfigureSearch(ResourceModuleConstruct construct)
...
 
-            if (configureResource is not null)
-            {
-                var resource = (AzureSearchResource)construct.Resource;
-                var resourceBuilder = builder.CreateResourceBuilder(resource);
-                configureResource(resourceBuilder, construct, search);
-            }
+            var resource = (AzureSearchResource)construct.Resource;
+            var resourceBuilder = builder.CreateResourceBuilder(resource);
+            configureResource?.Invoke(resourceBuilder, construct, search);
```
eerhardt, line 76: (super nit) It's kind of weird to create something (the `resourceBuilder`) and then not use it when `configureResource` is null. Obviously, it doesn't really matter. But it just looks odd reading the code.

eerhardt (approved):

PR #2953, opened 2024-03-16
Make malformed user secrets more obvious.

### LINKED ISSUE #2939: Azure Provisioner crashes silently if user secerts is invalid JSON.
Modify the user secrets such that it is invalid JSON. Launch an AppHost that uses an Azure Provisioner and you'll see that it doesn't start creating resource groups etc. But there is no visible sign in the dashboard that anything went wrong because we fire and forget provisioning.

mitchdenny:
Fixes #2939 

Now when failure occurs we'll get an error message for each impacted Azure resource as well as an error message right up front in the console log. The resources will also be in a failed state to make it obvious that folks need to dig in.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2953)

### DIFF
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
@@ -171,10 +171,11 @@ internal sealed class AzureProvisioner(
         {
             try
             {
-                var userSecrets = await provisioningContextLazy.Value.ConfigureAwait(false);
+                var provisioningContext = await provisioningContextLazy.Value.ConfigureAwait(false);
+
                 // Ensure directory exists before attempting to create secrets file
                 Directory.CreateDirectory(Path.GetDirectoryName(userSecretsPath)!);
-                await File.WriteAllTextAsync(userSecretsPath, userSecrets.ToString(), cancellationToken).ConfigureAwait(false);
+                await File.WriteAllTextAsync(userSecretsPath, provisioningContext.UserSecrets.ToString(), cancellationToken).ConfigureAwait(false);
 
                 logger.LogInformation("Azure resource connection strings saved to user secrets.");
             }
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
@@ -151,39 +151,8 @@ internal sealed class AzureProvisioner(
 
         var userSecretsPath = GetUserSecretsPath();
 
-        JsonObject? userSecrets = null;
-        try
-        {
-            userSecrets = userSecretsPath is not null && File.Exists(userSecretsPath)
-                ? JsonNode.Parse(await File.ReadAllTextAsync(userSecretsPath, cancellationToken).ConfigureAwait(false))!.AsObject()
-                : [];
-        }
-        catch (JsonException ex)
-        {
-            // If we fail to parse the JSON then we can't really continue because we won't be able to update
-            // the JSON with secrets. So rather than start provisioning on a bad JSON file we exit out. However
-            // to help the user to figure out what is going on we log an error on the main console logger as
-            // well as enumerate through each of the resources and throw up an error in those logs as well.
-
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -151,9 +151,36 @@ async Task AfterProvisionAsync(IAzureResource resource)
...
+            {
+                var resourceLogger = loggerService.GetLogger(resource);
+                resourceLogger.LogError(
+                    ex,
+                    "Failed to provision resource '{ResourceName}' because user secrets file '{UserSecretsPath}' was not valid JSON.",
+                    resource.Name,
+                    userSecretsPath
+                    );
+                resource.ProvisioningTaskCompletionSource?.SetException(ex);
+            }
```
davidfowl, line 180: Instead of this logic, can we move this to `GetProvisioningContextAsync`? That way there's a single piece of logic setting the TCS and logging to the per resource logger.

mitchdenny: I moved the user secret loading logic so it is loaded when we get the provisioning context. If that fails the sub-resources will fail, I still special case the message to make it easier for folks to diagnose, and I still catch the exception when we go to write the user secrets so that there is something on the console output (this will help diagnose failures like this in CI scenarios).

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -173,11 +169,19 @@ async Task AfterProvisionAsync(IAzureResource resource)
         // If we created any resources then save the user secrets
         if (userSecretsPath is not null)
         {
-            // Ensure directory exists before attempting to create secrets file
-            Directory.CreateDirectory(Path.GetDirectoryName(userSecretsPath)!);
-            await File.WriteAllTextAsync(userSecretsPath, userSecrets.ToString(), cancellationToken).ConfigureAwait(false);
+            try
+            {
+                var userSecrets = await provisioningContextLazy.Value.ConfigureAwait(false);
```
davidfowl, line 174: I am very confused by this. Shouldn't this be the provisioningContext?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -173,11 +169,19 @@ async Task AfterProvisionAsync(IAzureResource resource)
...
+                // Ensure directory exists before attempting to create secrets file
+                Directory.CreateDirectory(Path.GetDirectoryName(userSecretsPath)!);
+                await File.WriteAllTextAsync(userSecretsPath, userSecrets.ToString(), cancellationToken).ConfigureAwait(false);
```
davidfowl, line 177: This should be:

```suggestion
                await File.WriteAllTextAsync(userSecretsPath, context.UserSecrets.ToString(), cancellationToken).ConfigureAwait(false);
```

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
davidfowl, line 174: I am very confused by this. Shouldn't this be the provisioningContext?
```suggestion
                var provisioningContext = await provisioningContextLazy.Value.ConfigureAwait(false);
```

mitchdenny: Fixed. That was a bit of a brainfart. I didn't catch it because everything can ToString :)

davidfowl (approved):

davidfowl:
I just realized we should have made a UserSecrets lazy getter. The problem with this approach is that we need the entire provisioning context to save the user secrets. Even when we're fully cached, we'll end up going to azure hitting your azure sub https://github.com/dotnet/aspire/blob/16c0bb2c1e74e2dfd4dff4a18e0f32dfdd1a5bc1/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs#L174.

We should fix this.

PR #2974, opened 2024-03-18
Check playground manifests against baselines.

mitchdenny:
The purpose of this PR is to inject some steps into the build that will help us detect if the playground projects have not had their manifests and/or Bicep outputs regenerated. If someone adjusts the playground project we want to make sure we are seeing what the impacts are to the manifests across the spectrum of playground projects we have ... this will break the they get updated.


radical:
```diff
-      "path": "../../../../../Users/midenn/AppData/Local/Temp/tmp0cmpqc.tmp.bicep"
+      "path": "../../../../../../../../tmp/tmpMUnPth.tmp.bicep"
```

Are these paths correct?

mitchdenny:
> ```diff
> -      "path": "../../../../../Users/midenn/AppData/Local/Temp/tmp0cmpqc.tmp.bicep"
> +      "path": "../../../../../../../../tmp/tmpMUnPth.tmp.bicep"
> ```
> 
> Are these paths correct?

Nope. I need to figure out why that is happening.

PR #2988, opened 2024-03-18
Expose IDistributedApplicationBuilder on DistributedApplicationModel.

mitchdenny:
To discuss.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2988)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/DistributedApplicationModel.cs
+++ b/src/Aspire.Hosting/ApplicationModel/DistributedApplicationModel.cs
@@ -9,11 +9,17 @@ namespace Aspire.Hosting.ApplicationModel;
 /// Represents a distributed application.
 /// </summary>
 /// <param name="resources">The resource collection used to initiate the model.</param>
+/// <param name="applicationBuilder">The <see cref="IDistributedApplicationBuilder" /> that is attached to the model.</param>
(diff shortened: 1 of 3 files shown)


davidfowl:
What do we do with the service collection?

mitchdenny:
We already have instances where there is a captured builder exposed and accessed after the container is built. It don't see this as being any worse than that. An alternative might be to inject the IDistributedApplicationBuilder into the container - or provide an interface that offers a factory for returning the builder for a particular resource.

eerhardt:
I'm not sure I understand the goal here. Can you explain the reasoning for wanting this?

In general, what use is the `IDistributedApplicationBuilder` after it has been built? You can't add anything to it.

> What do we do with the service collection?

The service collection just returns the `_innerBuilder.Services` where `_innerBuilder` is a `HostApplicationBuilder`. `HostApplicationBuilder` will make the collection readonly once it is built:

https://github.com/dotnet/runtime/blob/ffe7e26ccc29f5ca1737dd7d84fa98b1b7c10d5f/src/libraries/Microsoft.Extensions.Hosting/src/HostApplicationBuilder.cs#L245-L246

PR #2994, opened 2024-03-19
Fix TimeProvider error by registering separate type as scoped

JamesNK:
The dashboard auth change and the dashboard time zone change conflict with each other and error. TimeProvider is registered as scoped and auth wants a singleton (i.e. the system time).

This PR fixes the problem by making the browser time provider a separate DI registration than the standard TimeProvider registration.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2994)

### DIFF
--- a/src/Aspire.Dashboard/Extensions/TimeProviderExtensions.cs
+++ b/src/Aspire.Dashboard/Extensions/TimeProviderExtensions.cs
@@ -1,11 +1,13 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Aspire.Dashboard.Model;
+
 namespace Aspire.Dashboard.Extensions;
 
 internal static class TimeProviderExtensions
 {
-    public static DateTime ToLocal(this TimeProvider timeProvider, DateTimeOffset utcDateTimeOffset)
+    public static DateTime ToLocal(this BrowserTimeProvider timeProvider, DateTimeOffset utcDateTimeOffset)
     {
         var dateTime = TimeZoneInfo.ConvertTimeFromUtc(utcDateTimeOffset.UtcDateTime, timeProvider.LocalTimeZone);
         dateTime = DateTime.SpecifyKind(dateTime, DateTimeKind.Local);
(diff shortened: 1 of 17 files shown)


mitchdenny (approved):

src/Aspire.Dashboard/Extensions/TimeProviderExtensions.cs
```diff
@@ -1,24 +1,26 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Aspire.Dashboard.Model;
+
 namespace Aspire.Dashboard.Extensions;
 
 internal static class TimeProviderExtensions
 {
-    public static DateTime ToLocal(this TimeProvider timeProvider, DateTimeOffset utcDateTimeOffset)
+    public static DateTime ToLocal(this BrowserTimeProvider timeProvider, DateTimeOffset utcDateTimeOffset)
```
drewnoakes, line 10: This narrowing isn't technically required as far as I can tell (`BrowserTimeProvider` is a `TimeProvider`). Perhaps revert the changes in this class, or rename the class to `BrowserTimeProviderExtensions`.

src/Aspire.Dashboard/Utils/FormatHelpers.cs
```diff
@@ -50,7 +51,7 @@ static string GetLongTimePatternWithMillisecondsCore(CultureInfo cultureInfo)
 
     private static string GetShortDateLongTimePatternWithMilliseconds(CultureInfo cultureInfo) => GetMillisecondFormatStrings(cultureInfo).ShortDateLongTimePattern;
 
-    public static string FormatTime(TimeProvider timeProvider, DateTime value, bool includeMilliseconds = false, CultureInfo? cultureInfo = null)
+    public static string FormatTime(BrowserTimeProvider timeProvider, DateTime value, bool includeMilliseconds = false, CultureInfo? cultureInfo = null)
```
drewnoakes, line 54: Ditto, changes in this file.

drewnoakes (approved):

src/Aspire.Dashboard/Extensions/TimeProviderExtensions.cs
JamesNK: I narrowed it because I don't want someone to accidentally use plain of `TimeProvider` by accident. (this helped me catch a few places I had missed)

If there is a scenario where these extension methods are used on a normal provider then we can revisit.

JamesNK: Might rename to BrowserTimeProviderExtensions in the future. I won't right now to avoid merge delay.

PR #2998, opened 2024-03-19
Make user secrets lazy load.

mitchdenny:
Turns user secret loading into a lazy operation that is awaited in each resource provisioning operation.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2998)

### DIFF
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
@@ -133,6 +133,14 @@ internal sealed class AzureProvisioner(
             cancellationToken), cancellationToken);
     }
 
+    private static async Task<JsonObject> GetUserSecretsAsync(string? userSecretsPath, CancellationToken cancellationToken)
+    {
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -173,9 +182,11 @@ async Task AfterProvisionAsync(IAzureResource resource)
             {
                 var provisioningContext = await provisioningContextLazy.Value.ConfigureAwait(false);
```
davidfowl, line 183: Remove this.

davidfowl (approved):

davidfowl (changes requested):
Hmm, I just realized this isn't right. I might have been wrong here 😭. We can't save this as it'll overwrite the file.

mitchdenny:
> Hmm, I just realized this isn't right. I might have been wrong here 😭. We can't save this as it'll overwrite the file.

I don't follow. It seems to work as expected.

davidfowl (approved):

PR #3013, opened 2024-03-19
Add method to configure Redis container persistence

### LINKED ISSUE #3009: Redis WithDataBindMount/WithDataVolume methods don't configure Redis to enable persistence
The Redis container image doesn't use persistent storage unless explicitly configured to do so (via the `--save` argument, e.g. `--save 60 1`, see [Redis container image README](https://hub.docker.com/_/redis/) and [Redis docs](https://redis.io/docs/management/persistence/#snapshotting)). The newly added `WithDataBindMount` and `WithDataVolume` extension methods for the Redis resource only add the container storage mounts but don't actually enable persistence for Redis.

We should update the `WithDataBindMount` and `WithDataVolume` methods to enable Redis persistence on the container. We might also consider adding a `WithPersistence(TimeSpan interval, long keysChangedCount)` method that calls `WithArgs($"--save {internal.TotalSeconds} {keysChangedCount}")` to capture the concept.

DamianEdwards:
Fixes #3009

Here's a recording of Redis persistence working in TestShop to preserve the basket across debug sessions:

https://github.com/dotnet/aspire/assets/249088/f9c1e3fc-9ba1-49c7-95dd-a247b17605be

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3013)

### DIFF
--- a/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
@@ -55,22 +55,34 @@ public static class RedisBuilderExtensions
     }
 
     /// <summary>
-    /// Adds a named volume for the data folder to a Redis container resource.
+    /// Adds a named volume for the data folder to a Redis container resource and enables Redis persistence.
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the resource name. </param>
     /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RedisResource> WithDataVolume(this IResourceBuilder<RedisResource> builder, string? name = null, bool isReadOnly = false)
-        => builder.WithVolume(name ?? $"{builder.Resource.Name}-data", "/data", isReadOnly);
+        => builder.WithVolume(name ?? $"{builder.Resource.Name}-data", "/data", isReadOnly)
+                  .WithPersistence(TimeSpan.FromSeconds(60));
 
     /// <summary>
-    /// Adds a bind mount for the data folder to a Redis container resource.
+    /// Adds a bind mount for the data folder to a Redis container resource and enables Redis persistence.
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="source">The source directory on the host to mount into the container.</param>
     /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RedisResource> WithDataBindMount(this IResourceBuilder<RedisResource> builder, string source, bool isReadOnly = false)
-        => builder.WithBindMount(source, "/data", isReadOnly);
+        => builder.WithBindMount(source, "/data", isReadOnly)
+                  .WithPersistence(TimeSpan.FromSeconds(60));
+
+    /// <summary>
+    /// Configures the persistence settings for a Redis container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="interval">The interval between snapshot exports.</param>
+    /// <param name="keysChangedThreshold">The number of key change operations required to trigger a snapshot at the interval.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<RedisResource> WithPersistence(this IResourceBuilder<RedisResource> builder, TimeSpan interval, long keysChangedThreshold = 1)
+        => builder.WithAnnotation(new RedisPersistenceCommandLineArgsCallbackAnnotation(interval, keysChangedThreshold), ResourceAnnotationMutationBehavior.Replace);
 }
--- a/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
@@ -57,32 +57,75 @@ public static class RedisBuilderExtensions
     /// <summary>
     /// Adds a named volume for the data folder to a Redis container resource and enables Redis persistence.
     /// </summary>
+    /// <remarks>
+    /// Use <see cref="WithPersistence(IResourceBuilder{RedisResource}, TimeSpan?, long)"/> to adjust Redis persistence configuration, e.g.:
+    /// <code>
+    /// var cache = builder.AddRedis("cache")
+    ///                    .WithDataVolume()
+    ///                    .WithPersistence(TimeSpan.FromSeconds(10), 5);
+    /// </code>
+    /// </remarks>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the resource name. </param>
-    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
+    /// <param name="isReadOnly">
+    /// A flag that indicates if this is a read-only volume. Setting this to <c>true</c> will disable Redis persistence.<br/>
+    /// Defaults to <c>false</c>.
+    /// </param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RedisResource> WithDataVolume(this IResourceBuilder<RedisResource> builder, string? name = null, bool isReadOnly = false)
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Redis/RedisPersistenceCommandLineArgsCallbackAnnotation.cs
```diff
@@ -0,0 +1,21 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Globalization;
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting.Redis;
+
+internal class RedisPersistenceCommandLineArgsCallbackAnnotation(TimeSpan interval, long keysChangedThreshold)
```
davidfowl, line 9: Why?

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -55,22 +55,34 @@ public static IResourceBuilder<RedisResource> WithRedisCommander(this IResourceB
...
+
+    /// <summary>
+    /// Configures the persistence settings for a Redis container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="interval">The interval between snapshot exports.</param>
+    /// <param name="keysChangedThreshold">The number of key change operations required to trigger a snapshot at the interval.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<RedisResource> WithPersistence(this IResourceBuilder<RedisResource> builder, TimeSpan interval, long keysChangedThreshold = 1)
+        => builder.WithAnnotation(new RedisPersistenceCommandLineArgsCallbackAnnotation(interval, keysChangedThreshold), ResourceAnnotationMutationBehavior.Replace);
```
davidfowl, line 87: builder.WithArgs?

src/Aspire.Hosting.Redis/RedisPersistenceCommandLineArgsCallbackAnnotation.cs
DamianEdwards: So that it can be easily overridden with another call. See the tests.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
DamianEdwards: WithArgs just keeps adding more args, whereas I want this to be adjustable via multiple calls and boil down to just one set of args to add.

davidfowl: This screams missing primitive. Does it compose with other calls to WithArgs?

DamianEdwards: Args don't really compose today, they just stack. In the case that a logical operation requires adding more than one arg (like in this case `["--save", "60",  "1"]` to get `--save 60 1`) you lose any structure. They're just a list of args to add, not even indexed, so implementing this kind of deduping in the callback of a single args annotation callback would be super ugly. Given we already have the concept of annotation mutation behavior, this seemed like the right approach.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -55,22 +55,34 @@ public static IResourceBuilder<RedisResource> WithRedisCommander(this IResourceB
...
+    /// Adds a named volume for the data folder to a Redis container resource and enables Redis persistence.
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the resource name. </param>
     /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RedisResource> WithDataVolume(this IResourceBuilder<RedisResource> builder, string? name = null, bool isReadOnly = false)
-        => builder.WithVolume(name ?? $"{builder.Resource.Name}-data", "/data", isReadOnly);
+        => builder.WithVolume(name ?? $"{builder.Resource.Name}-data", "/data", isReadOnly)
+                  .WithPersistence(TimeSpan.FromSeconds(60));
```
davidfowl, line 66: Why a default value?

DamianEdwards: I think it's reasonable to have a default here but to your point it should likely be defined by the WithPersistence method and just documented here. Folks can call WithPersistence with args to set custom values.

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -164,4 +164,105 @@ public async Task MultipleRedisInstanceProducesCorrectRedisHostsVariable()
...
+        Assert.Equal(1, persistenceAnnotation.KeysChangedThreshold);
+    }
+
+    [Fact]
+    public void WithPersistenceReplacesPreviousAnnotationInstances()
+    {
+        var appBuilder = DistributedApplication.CreateBuilder();
+        var redis = appBuilder.AddRedis("myRedis")
+                              .WithDataVolume()
+                              .WithPersistence(TimeSpan.FromSeconds(10), 2);
```
mitchdenny, line 250: Could this API design be confusing to developers? It would be easy for the developer to assume that either of these extension methods along should do what they want? A developer might see `WithPersistence` and thing that is what they want bit without either `WithDataVolme` or `WithDataBindMount` it isn't going to do anything.

I can see why the split but I do think it'll be confusing for folks.

mitchdenny:
No problems with the implementation. Just sharing some concerns about the the design and potential for confusion.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -55,22 +55,34 @@ public static IResourceBuilder<RedisResource> WithRedisCommander(this IResourceB
...
 
     /// <summary>
-    /// Adds a named volume for the data folder to a Redis container resource.
+    /// Adds a named volume for the data folder to a Redis container resource and enables Redis persistence.
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the resource name. </param>
     /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RedisResource> WithDataVolume(this IResourceBuilder<RedisResource> builder, string? name = null, bool isReadOnly = false)
```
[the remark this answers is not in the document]
DamianEdwards: Good question. I remember thinking about this and deciding to leave it there but now I can't remember why. I'll do some more testing today and verify the behavior.

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
DamianEdwards: `WithPersistence` will work without a bind mount or volume, it will just persist to the container file system and be lost when the container is stopped. I can add more details to the doc comments to try and make it clearer. The alternative would be to remove the `WithPersistence` method and promote its arguments to `WithDataVolume` and `WithDataBindMount`.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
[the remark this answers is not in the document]
DamianEdwards: One could use the same volume between two different Redis resources, with one having only read access. I updated the methods so that if `isReadOnly == true` then Redis persistence is not configured, as the snapshot attempt will fail obviously if the directory is read only. I tried it out locally and it works well, with data from the volume being loaded but new operations not being persisted.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
DamianEdwards: I moved the defaults to `WithPersistence` and updated the doc comments to make it clearer how to change the persistence configuration.

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
DamianEdwards: I updated the doc comments to make it clearer that calling `WithDataBindMount` or `WithDataVolume` is required to get the persistence to last across sessions.

davidfowl (approved):

PR #2851, opened 2024-03-13
Add support for `docker build --build-args`

IEvangelist:
Add support for `docker build --build-args` from the `PublishAsDockerFile` API.

- Fixes #2441.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2851)

PR #3015, opened 2024-03-19
Address peer feedback.

IEvangelist:
See https://github.com/dotnet/aspire/pull/2851#pullrequestreview-1938029704
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3015)

### DIFF
--- a/src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
+++ b/src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
@@ -281,38 +281,16 @@ public sealed class ManifestPublishingContext(DistributedApplicationExecutionCon
             {
                 var buildArg = args[i];
 
-                Writer.WritePropertyName(buildArg.Name);
-
-                var value = buildArg.Value;
-                if (value is null) // null means let docker build pull from env var.
+                var valueString = buildArg.Value switch
                 {
-                    Writer.WriteNullValue();
-                    continue;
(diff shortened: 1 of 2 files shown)


davidfowl:
I think before we merge this, we should verify what it means to have a non string argument. Doesn't this translate into a command line argument anyways? (strings)

eerhardt:
> I think before we merge this, we should verify what it means to have a non string argument. Doesn't this translate into a command line argument anyways? (strings)

I'd be fine with `"true"` in the manifest (i.e. only supporting JSON strings). 

I'm not sure how the current code works today with `"{{bool.TrueString}}"`.

IEvangelist:
> > I think before we merge this, we should verify what it means to have a non string argument. Doesn't this translate into a command line argument anyways? (strings)
> 
> I'd be fine with `"true"` in the manifest (i.e. only supporting JSON strings).
> 
> I'm not sure how the current code works today with `"{{bool.TrueString}}"`.

That was the test code, it evaluated as `"SOME_BOOL": "True"`. I'd assume it's probably pretty common for true/false values. Even STJ doesn't have native support for mapping the string representation of bool to a bool property type. This change adds `bool` support only, otherwise it's functionally the same as it was before.

eerhardt (approved):

Issue #2850, opened 2024-03-13
`EndToEnd` tests failing to start up `testproject` with `dotnet run`


radical:
[Build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=600755&view=logs&j=22fea640-1099-5f32-ec5d-316ad83f359a&t=d268830e-a3ea-51fa-cf04-927cd28b9001) and [failing log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-heads-main-41184ebf5ee84503b6/Aspire.EndToEnd.Tests/1/console.da51e905.log?helixlogtype=result)

```
...
[apphost] info: Aspire.Hosting.Dcp.api-server[0]
[apphost]       API server started	{"Address": "127.0.0.1", "Port": 46087}
[apphost] info: Aspire.Hosting.Dcp.dcpctrl.IdeExecutableRunner[0]
[apphost]       Executables cannot be started via IDE: missing required environment variable 'DEBUG_SESSION_PORT'
[apphost] info: Aspire.Hosting.Dcp.dcpctrl[0]
[apphost]       starting controller manager
...
[apphost] fail: Microsoft.Extensions.Hosting.Internal.Host[11]
[apphost]       Hosting failed to start
[apphost]       System.Threading.Tasks.TaskCanceledException: A task was canceled.
[apphost]          at k8s.Watcher`1.CreateWatchEventEnumerator(Func`1 streamReaderCreator, Action`1 onError, CancellationToken cancellationToken)+MoveNext()
[apphost]          at k8s.Watcher`1.CreateWatchEventEnumerator(Func`1 streamReaderCreator, Action`1 onError, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
[apphost]          at Aspire.Hosting.Dcp.KubernetesService.WatchAsync[T](String namespaceParameter, CancellationToken cancellationToken)+MoveNext() in /_/src/Aspire.Hosting/Dcp/KubernetesService.cs:line 170
[apphost]          at Aspire.Hosting.Dcp.KubernetesService.WatchAsync[T](String namespaceParameter, CancellationToken cancellationToken)+MoveNext() in /_/src/Aspire.Hosting/Dcp/KubernetesService.cs:line 170
[...]

PR #3018, opened 2024-03-19
[tests] Use helix's retry mechanism for flaky tests

radical:
In particular:
`Aspire.EndToEnd.Tests`: https://github.com/dotnet/aspire/issues/2850

`Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.Tests`, and
`Aspire.Microsoft.EntityFrameworkCore.SqlServer.Tests` get aborted with
timeout:

`Aborting test run: test run timeout of 600000 milliseconds exceeded.`

Related: https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.Helix/Sdk#test-retry

cc @eerhardt @joperezr @davidfowl 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3018)

### DIFF
--- a/eng/test-configuration.json
+++ b/eng/test-configuration.json
@@ -3,8 +3,8 @@
   "defaultOnFailure": "fail",
   "localRerunCount": 1,
   "retryOnRules": [
-    { "testAssembly": { "wildcard": "Aspire.EndToEnd.*" }, "failureMessage": { "regex": "App run failed" } },
-    { "testAssembly": { "wildcard": "Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*" } },
-    { "testAssembly": { "wildcard": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } }
+    { "testAssembly": { "regex": "Aspire.EndToEnd.*" }, "failureMessage": { "regex": "App run failed" } },
+    { "testAssembly": { "regex": "Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*" } },
+    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } }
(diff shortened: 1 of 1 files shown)


radical:
I'll check that this does work before we merge.

eng/test-configuration.json
```diff
@@ -0,0 +1,10 @@
+{
+  "version": 1,
+  "defaultOnFailure": "fail",
```
eerhardt, line 3: What other values are available here?

radical: https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.Helix/Sdk#defaultonfailure

`fail` here would retry only if there is a matching rule. My thinking here was that we shouldn't add retries for *any* failure, rather add them specifically for flaky cases. Which are these particular ones right now.

radical:
Retry works:

Screenshot 2024-03-19 at 18 52 43

The second time is also a fail because the simulated error always fails.

eerhardt (approved):
:shipit:

joperezr (approved):
LGTM. I know that the ef core tests are already tracked by a different issue, but for the end to end tests do we plan on eventually removing these retries? Or do we plan on keeping those?

radical:
Ideally the retries can be removed at some point in the future!

eerhardt:
> Ideally the retries can be removed at some point in the future!

Agreed. We need to stabilize the EndToEnd tests so they pass consistently. We are currently getting random failures when calling `dotnet run` on the AppHost project, but we don't have enough logging to figure out what the error is.

The idea here is to minimize the impact on devs so they see these random failures less often while we add more logging and stabilize these tests.

PR #3023, opened 2024-03-19
Make username and password defaults consistent

### LINKED ISSUE #2403: Review username and password defaults for all container and azure resources and make them consistent.
We should support:

1. Picking a username and password (using Parameter resources)
1. Defaulting a username and randomly generating a password
1. Picking a username and generating a password

Related https://github.com/dotnet/aspire/issues/2389

eerhardt:
* Support RabbitMQ username and password from Parameter
* Support MySql password from Parameter
* Support Oracle password from Parameter
* Support SQL Server password from Parameter

Fix #2403
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3023)

### DIFF
--- a/src/Aspire.Hosting.MySql/MySqlServerResource.cs
+++ b/src/Aspire.Hosting.MySql/MySqlServerResource.cs
@@ -18,11 +18,13 @@ public class MySqlServerResource : ContainerResource, IResourceWithConnectionStr
     public MySqlServerResource(string name, ParameterResource? password) : base(name)
     {
         PrimaryEndpoint = new(this, PrimaryEndpointName);
-        PasswordInput = new(this, "password");
-
         PasswordParameter = password;
 
-        Annotations.Add(InputAnnotation.CreateDefaultPasswordInput());
+        if (PasswordParameter is null)
+        {
+            Annotations.Add(InputAnnotation.CreateDefaultPasswordInput());
+            PasswordInput = new(this, "password");
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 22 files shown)


src/Aspire.Hosting.MySql/MySqlServerResource.cs
```diff
@@ -14,33 +14,45 @@ public class MySqlServerResource : ContainerResource, IResourceWithConnectionStr
...
     /// </summary>
     public EndpointReference PrimaryEndpoint { get; }
 
-    internal InputReference PasswordInput { get; }
+    private InputReference PasswordInput { get; }
+
+    /// <summary>
+    /// Gets the parameter that contains the MySQL server password.
+    /// </summary>
+    public ParameterResource? PasswordParameter { get; }
```
davidfowl, line 38: Do we want to expose this or the PasswordReference ?

mitchdenny:
@eerhardt can you regenerate the manifests in the playgrounds?

mitchdenny:
We've discussed this before I understand the reasoning, but I still don't like that we emit `inputs` fields for generated passwords that we aren't going to use. I think we should figure out an elegant way to avoid this because its just plain confusing :|

src/Aspire.Hosting.MySql/MySqlServerResource.cs
eerhardt: I don't want the `InputRefererence` to leak out of the object since the "input" can only be referenced from inside the resource.

Long-term I think we should consider merging Parameters and Inputs. I think the concept of "InputAnnotation" is confusing. Instead we should just have Parameters, and Parameters can have default/generated values.

davidfowl (approved):
LGTM

PR #3026, opened 2024-03-19
Support new IDE protocol in app model

### LINKED ISSUE #2928: Throw instead of calling `Environment.Exit` in `DcpDependencyCheck`

karolz-ms:
Because it was decided that we should support the old version of the protocol for one preview release, the code inside `ApplicationExecutor` must do different things depending on the version of DCP it is talking to. As a result, I refactored a bit the DCP version/dependency check. Adding @ReubenBond who I was told is doing some work in this area right now.

Fixes #2928

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3026)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -65,7 +65,8 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                                           IDashboardEndpointProvider dashboardEndpointProvider,
                                           DistributedApplicationExecutionContext executionContext,
                                           ResourceNotificationService notificationService,
-                                          ResourceLoggerService loggerService)
+                                          ResourceLoggerService loggerService,
+                                          IDcpDependencyCheckService dcpDependencyCheckService)
 {
     private const string DebugSessionPortVar = "DEBUG_SESSION_PORT";
 
@@ -106,7 +107,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
             }
             PrepareServices();
             PrepareContainers();
-            PrepareExecutables();
+            await PrepareExecutablesAsync(cancellationToken).ConfigureAwait(false);
 
             await PublishResourcesWithInitialStateAsync().ConfigureAwait(false);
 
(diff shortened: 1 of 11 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -928,9 +938,10 @@ private void PreparePlainExecutables()
         }
     }
 
-    private void PrepareProjectExecutables()
+    private async Task PrepareProjectExecutablesAsync(CancellationToken cancellationToken)
     {
         var modelProjectResources = _model.GetProjectResources();
+        var dcpInfo = await dcpDependencyCheckService.GetDcpInfoAsync(cancellationToken).ConfigureAwait(false);
```
davidfowl, line 944: Perfect! I need this information to get the container runtime 😄 for the host name in this file.

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -29,79 +31,96 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
         _dcpOptions = dcpOptions.Value;
     }
 
-    public async Task EnsureDcpDependenciesAsync(CancellationToken cancellationToken = default)
+    public async Task<DcpInfo?> GetDcpInfoAsync(CancellationToken cancellationToken = default)
```
davidfowl, line 34: Any reason to not make this cache the result?

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -29,79 +31,96 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
...
+                return dcpInfo;
+            }
+            catch (Exception ex) when (ex is not DistributedApplicationException)
             {
-                return; // Best effort
+                Console.Error.WriteLine(string.Format(
+                    CultureInfo.InvariantCulture,
+                    Resources.DcpDependencyCheckFailedMessage,
+                    ex.ToString()
+                ));
```
mitchdenny, line 109: Use logger instead.

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -29,79 +31,96 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
...
+                Environment.Exit((int)DcpVersionCheckFailures.DcpVersionFailed);
```
mitchdenny, line 110: Can we throw here instead of doing an exit.

davidfowl:
@karolz-ms I crapped on your PR an integrated @ReubenBond 's changes https://github.com/dotnet/aspire/pull/2931.

davidfowl:
@danegsta we are throwing here. We can tweak this in a follow up PR to set a specific exit code for unhandled exceptions (that was a good idea 😄).

davidfowl (approved):

Issue #2422, opened 2024-02-26
Make dcp and dashboard communication reliable and robust


davidfowl:
It seems like we have a set of issues that all have to do with launching and connecting to DCP and making sure that it's reliable. We should harder this code to make sure we a resilient to launching DCP (retrying if it fails to launch) and making sure it's healthy and if it goes unhealthy having a good way to recover.

 - https://github.com/dotnet/aspire/issues/2198 
- https://github.com/dotnet/aspire/issues/2019
- https://github.com/dotnet/aspire/issues/2187
- https://github.com/dotnet/aspire/issues/1976
- https://github.com/dotnet/aspire/issues/2017
- https://github.com/dotnet/aspire/issues/2539
- https://github.com/dotnet/aspire/issues/2542
- https://github.com/dotnet/aspire/issues/881

davidfowl:
cc @karolz-ms @danegsta

karolz-ms:
All what I have seen so far in terms of inability to start DCP/dashboard fell into one of 3 buckets:

1. Mismatched Aspire/DCP (workload) binaries. We should be on the look-out to minimize opportunities for such mismatch and I believe the workload set work that Jose and co are doing should be a big help. However, once you are in this situation, no amount of compensation/retrying will help. One thing that we can do now is to improve the error message like https://github.com/dotnet/aspire/issues/1976 suggested, e.g. by instructing developers how to ensure that their workload installation is healthy (`DcpHostService` would be the place I'd consider). Maybe a troubleshooting documentation page could also help here.

2. DCP shutting down too slowly. @danegsta has fixed all known problems in this area and these fixes have already been integrated into Aspire (DCP release 0.1.54).

3. "Arctic" run times out. By that I mean the very first run of a particular DCP version on a given machine. On my dev machine, which is no slouch, it routinly takes close to 10 seconds for the "advanced threat protection" AV software to scan new DCP binaries before they start running, every time I build. This makes me think that our retry timeout for establishing communication between app host and DCP is currently too small (5 seconds) and this PR https://github.com/dotnet/aspire/pull/2435/files#diff-a3e877b57b19124a6d9810e8072e6f1892acf96a26246fc4fcfa34570ed77028 is bumping it up to 20 seconds.

These are the things that IMO have/will make a difference--hope this helps and happy to hear more thoughts on the subject.

davidfowl:
[...]

PR #3029, opened 2024-03-20
Retry on watch longpoll disconnect.

mitchdenny:
Partially addresses #2422 

This PR introduces a Polly resilience pipeline to retry the watch resource long poll operation when it disconnects. We detect the disconnect by overving that the exception that is thrown has an inner exception of type `EndOfStreamException`. If we see this then we'll log a debug message for future supportability purposes, but otherwise exit.

We swallow `OperationCancelledExceptions` and exit cleanly. All other exceptions are logged.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3029)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -156,21 +158,44 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
...
+                        }
+                        finally
+                        {
+                            semaphore.Release();
+                        }
                     }
-                }
+                }, cancellationToken).ConfigureAwait(false);
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
```
mitchdenny, line 200: We preserve existing behavior where we swallow all exceptions and log an error except for cancellations which bubble out.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -156,21 +158,44 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
...
-                await foreach (var (eventType, resource) in kubernetesService.WatchAsync<T>(cancellationToken: cancellationToken))
+                ShouldHandle = new PredicateBuilder().HandleInner<EndOfStreamException>(),
+                BackoffType = DelayBackoffType.Constant,
+                MaxRetryAttempts = int.MaxValue,
+                UseJitter = true,
+                MaxDelay = TimeSpan.FromSeconds(30),
+                OnRetry = (retry) =>
                 {
-                    await semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    _logger.LogDebug(
```
mitchdenny, line 170: This is hard to test. You have to drastically alter KubernetesService to inject this error synthetically and DCP has quite a long timeout for long poll (like Kubernetes). I've asked @karolz-ms for an environment variable we can inject into DCP when launching it to bring down the long poll timeout for resource watching so we can simulate this more easily.

davidfowl:
cc @martincostello

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -156,21 +158,44 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
...
-                    try
-                    {
-                        await handler(eventType, resource).ConfigureAwait(false);
-                    }
-                    finally
+            var pipeline = new ResiliencePipelineBuilder().AddRetry(retryUntilCancelled).Build();
+
+            try
+            {
+                await pipeline.ExecuteAsync(async (pipelineCancellationToken) =>
```
martincostello, line 183: Is it much refactoring/cognitive overload to refactor this to pass the state through into Polly so you can avoid the need to allocation the closure here by making the lambda static, or is the context where this happens not perf-sensitive enough for that to be a concern?

mitchdenny: Its not that perf sensitive. We execute this once per resource type of project, executable, and container. And if we fail with an end of stream (after ~30 mins) its not a huge cost.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -185,25 +187,51 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
 
         async Task WatchKubernetesResource<T>(Func<WatchEventType, T, Task> handler) where T : CustomResource
         {
+            var retryUntilCancelled = new RetryStrategyOptions()
+            {
+                ShouldHandle = new PredicateBuilder().HandleInner<EndOfStreamException>(),
+                BackoffType = DelayBackoffType.Exponential,
+                MaxRetryAttempts = int.MaxValue,
+                UseJitter = true,
+                MaxDelay = TimeSpan.FromSeconds(30),
```
davidfowl, line 196: 30 seconds seems short no?

davidfowl (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -185,25 +187,51 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
...
+                            semaphore.Release();
+                        }
                     }
-                }
+                }, cancellationToken).ConfigureAwait(false);
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
-                _logger.LogError(ex, "Watch task over kubernetes {ResourceType} resources terminated", typeof(T).Name);
+                _logger.LogCritical(ex, "Watch task over kubernetes {ResourceType} resources terminated", typeof(T).Name);
```
davidfowl, line 234: Can we say something more about what action the user can take. Like mmake sure dcp is running etc./

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: Discussed offline. This is the only code that is hitting the dcp resource watch endpoint, hitting it every 30 seconds is probably fine. We hit it more frequently than that with an exponential backoff to 30 seconds. Realistically - if it doesn't respond by 30 seconds it is probably dead and a different (fatal) exception would have thrown and we would have aborted.

PR #3031, opened 2024-03-20
Use API key auth for OTLP in local development

JamesNK:
Previous PR added API key auth to OTLP endpoint. There is the opportunity to secure the local dev environment by using API key auth and having the host co-ordinate the dashboard and user apps.

This PR:
* Generates a random API key each time the app host is run.
* Passes environment variables to dashboard to enable API key auth, and the expected API key.
* Passes environment variables to ups to add API key header to OTLP requests.
* Can be disabled by passing `DOTNET_DISABLE_OTLP_API_KEY_AUTH=true` setting to apphost.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3031)

### DIFF
--- a/src/Aspire.Hosting/DistributedApplicationBuilder.cs
+++ b/src/Aspire.Hosting/DistributedApplicationBuilder.cs
@@ -74,20 +74,11 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
 
         AppHostDirectory = options.ProjectDirectory ?? _innerBuilder.Environment.ContentRootPath;
 
-        var configData = new Dictionary<string, string?>
+        _innerBuilder.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
         {
             // Make the app host directory available to the application via configuration
             ["AppHost:Directory"] = AppHostDirectory
-        };
-
-        if (!IsOtlpApiKeyAuthDisabled(_innerBuilder.Configuration))
-        {
-            // Set a random API key for the OTLP exporter.
-            // Passed to apps as a standard OTEL attribute to include in OTLP requests and the dashboard to validate.
-            configData["AppHost:OtlpApiKey"] = Guid.NewGuid().ToString();
-        }
-
-        _innerBuilder.Configuration.AddInMemoryCollection(configData);
+        });
 
         // Core things
         _innerBuilder.Services.AddSingleton(sp => new DistributedApplicationModel(Resources));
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -36,13 +36,16 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
...
+            var apiKey = configuration["AppHost:OtlpApiKey"];
+
             context.EnvironmentVariables["OTEL_EXPORTER_OTLP_ENDPOINT"] = resource is ContainerResource
                 ? HostNameResolver.ReplaceLocalhostWithContainerHost(url, configuration)
                 : url;
 
             // Set the service name and instance id to the resource name and UID. Values are injected by DCP.
             context.EnvironmentVariables["OTEL_RESOURCE_ATTRIBUTES"] = "service.instance.id={{- .Name -}}";
             context.EnvironmentVariables["OTEL_SERVICE_NAME"] = "{{- index .Annotations \"otel-service-name\" -}}";
+            context.EnvironmentVariables["OTEL_EXPORTER_OTLP_HEADERS"] = $"x-otlp-api-key={apiKey}";
```
davidfowl, line 48: Check if apiKey is null.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -72,10 +72,13 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
...
 
-        // Make the app host directory available to the application via configuration
         _innerBuilder.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
         {
-            ["AppHost:Directory"] = AppHostDirectory
+            // Make the app host directory available to the application via configuration
+            ["AppHost:Directory"] = AppHostDirectory,
+            // Set a random API key for the OTLP exporter.
+            // Passed to apps as a standard OTEL attribute to include in OTLP requests and the dashboard to validate.
+            ["AppHost:OtlpApiKey"] = Guid.NewGuid().ToString()
```
davidfowl, line 81: Make it possible to opt out. Either by checking if it's already set or checking another value.

JamesNK:
FYI @martinjt

I've looked at how your https://github.com/practical-otel/opentelemetry-aspire-collector/ works and I believe it will need some changes but should continue to be possible if API key auth is added to OTLP endpoints.

You'll need to:
* Read the API key from the `AppHost:OtlpApiKey` config setting.
* Pass the API key to collector container as an env var
* Update the config file to include an `x-otlp-api-key` header to the [exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/README.md).
* Consider removing the `OTEL_EXPORTER_OTLP_HEADERS` env var that is passed to apps. They don't need it if they're talking to the collector.

davidfowl (approved):

martinjt:
Cool, I can do that, I'm assuming thar will be Preview 5?

JamesNK:
Yes, preview 5.

PR #3033, opened 2024-03-20
Auth for dashboard web app and resource service client

drewnoakes:
Adds auth to both:

- The dashboard's gRPC resource service client, using client certificates
- The dashboard's web app, using OpenIdConnect

Several new configuration options exist. These are documented in the README changes in the PR.

Here's a recording of sign in with Google:

dashboard-openid-connect-login

Follow up items:

- [x] #3199
- [x] #3200
- [x] #3201
- [x] #3197
- [x] #3194
 
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3033)


src/Aspire.Dashboard/README.md
```diff
@@ -8,3 +10,26 @@ The dashboard is configured with the following keys:
...
+
+## Auth
+
+The dashboard also supports authentication for its various network endpoints.
+
+To opt-out of authentication, set the following environment variables to `1`:
+
+- `DOTNET_RESOURCE_SERVICE_DISABLE_AUTH`
+- `DOTNET_DASHBOARD_WEB_DISABLE_AUTH`
+- `DOTNET_DASHBOARD_OTLP_DISABLE_AUTH`
```
drewnoakes, line 22: These two configuration keys are not currently in use, though I expect they will be.

@JamesNK are you expecting OTEL auth to be enabled by default with explicit opt-out, as per the resource service and dashboard web app?

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -98,6 +103,28 @@ GrpcChannel CreateChannel()
...
+
+                if (string.IsNullOrWhiteSpace(certificatePath))
+                {
+                    throw new InvalidOperationException(
+                        $"Resource service auth hasn't been disabled via {ResourceServiceDisableAuthVariableName}, " +
+                        $"and no certificate was provided via {ResourceServiceClientCertificatePathVariableName} and " +
+                        $"{ResourceServiceClientCertificatePasswordVariableName}. Review the dashboard's configuration.");
+                }
+
+                httpHandler.SslOptions = new SslClientAuthenticationOptions
```
drewnoakes, line 120: We may wish to have a `RemoteCertificateValidationCallback` here that validates the server certificate's thumbprint (and corresponding config to pass that cert in).

drewnoakes:
We currently don't have a resource service implementation that works with certificates. To test this, I created certs with:

```
# create root CA details
openssl genrsa -aes256 -out rootCA.key 4096
openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 3650 -out rootCA.crt -subj "//CN=MyRootCA"

# server
openssl genrsa -aes256 -out server.key 4096
openssl req -new -key server.key -out server.csr -subj "//CN=localhost"
openssl x509 -req -in server.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out server.crt -days 365 -sha256
openssl pkcs12 -export -out server.pfx -inkey server.key -in server.crt

# client
openssl genrsa -aes256 -out client.key 4096
openssl req -new -key client.key -out client.csr -subj "//CN=localhost"
openssl x509 -req -in client.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out client.crt -days 365 -sha256
openssl pkcs12 -export -out client.pfx -inkey client.key -in client.crt
```

Then the following changes to the app host project:

```diff
diff --git a/src/Aspire.Hosting/Aspire.Hosting.csproj b/src/Aspire.Hosting/Aspire.Hosting.csproj
index 0707a57c..bc25bdd0 100644
--- a/src/Aspire.Hosting/Aspire.Hosting.csproj
+++ b/src/Aspire.Hosting/Aspire.Hosting.csproj
@@ -26,6 +26,7 @@
   <ItemGroup>
     <PackageReference Include="Grpc.AspNetCore" />
     <PackageReference Include="KubernetesClient" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Certificate" />
     <PackageReference Include="Microsoft.Extensions.Hosting" />
   </ItemGroup>
```

```diff
diff --git a/src/Aspire.Hosting/Dashboard/DashboardService.cs b/src/Aspire.Hosting/Dashboard/DashboardService.cs
index 16cec459a..4c2b1da2d 100644
--- a/src/Aspire.Hosting/Dashboard/DashboardService.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardService.cs
@@ -4,6 +4,7 @@
 using System.Text.RegularExpressions;
 using Aspire.V1;
 using Grpc.Core;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.Extensions.Hosting;
 
 namespace Aspire.Hosting.Dashboard;
@@ -15,6 +16,7 @@ namespace Aspire.Hosting.Dashboard;
 /// An instance of this type is created for every gRPC service call, so it may not hold onto any state
 /// required beyond a single request. Longer-scoped data is stored in <see cref="DashboardServiceData"/>.
 /// </remarks>
+[Authorize]
 internal sealed partial class DashboardService(DashboardServiceData serviceData, IHostEnvironment hostEnvironment, IHostApplicationLifetime hostApplicationLifetime)
     : V1.DashboardService.DashboardServiceBase
 {
```

```diff
diff --git a/src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs b/src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
index 20ee29d0..71d2df3a 100644
--- a/src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
@@ -3,13 +3,16 @@
 
 using System.Diagnostics;
 using System.Net;
+using System.Security.Cryptography.X509Certificates;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Dcp;
+using Microsoft.AspNetCore.Authentication.Certificate;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -72,6 +75,53 @@ internal sealed class DashboardServiceHost : IHostedService
             // Turn on HTTPS
             builder.WebHost.UseKestrelHttpsConfiguration();
 
+            #region TEMPORARY TEST CODE
+
+            // Auth
+            builder.Services
+                .AddAuthentication(CertificateAuthenticationDefaults.AuthenticationScheme)
+                .AddCertificate(options =>
+                {
+                    // Disallow self-signed certs.
+                    options.AllowedCertificateTypes = CertificateTypes.Chained;
+
+                    // Revocation checks require an online CA, which we don't have during testing.
+                    options.RevocationMode = X509RevocationMode.NoCheck;
+
+                    options.Events = new CertificateAuthenticationEvents()
+                    {
+                        OnAuthenticationFailed = context =>
+                        {
+                            _logger.LogError(context.Exception, "Failed authentication.");
+
+                            return Task.CompletedTask;
+                        },
+                        OnCertificateValidated = context =>
+                        {
+                            // TODO validate certificate is as expected
+
+                            _logger.LogInformation("Authentication complete.");
+
+                            return Task.CompletedTask;
+                        }
+                    };
+                });
+
+            builder.Services.AddAuthorization();
+
+            builder.Services.Configure<KestrelServerOptions>(options =>
+            {
+                options.ConfigureHttpsDefaults(options =>
+                {
+                    options.ServerCertificate = new X509Certificate2(@"d:\repos\tmp\certs\server.pfx", "server", X509KeyStorageFlags.DefaultKeySet);
+                    options.ClientCertificateMode = ClientCertificateMode.RequireCertificate;
+                });
+            });
+
+            #endregion
+
             // Configuration
             builder.Services.AddSingleton(configuration);
 
@@ -80,7 +130,7 @@ internal sealed class DashboardServiceHost : IHostedService
             builder.Services.AddSingleton(loggerOptions);
             builder.Services.Add(ServiceDescriptor.Singleton(typeof(ILogger<>), typeof(Logger<>)));
 
-            builder.Services.AddGrpc();
+            builder.Services.AddGrpc(options => options.EnableDetailedErrors = true);
             builder.Services.AddSingleton(applicationModel);
             builder.Services.AddSingleton(kubernetesService);
             builder.Services.AddSingleton<DashboardServiceData>();
@@ -91,6 +141,13 @@ internal sealed class DashboardServiceHost : IHostedService
 
             _app = builder.Build();
 
+            #region TEMPORARY TEST CODE
+
+            _app.UseAuthentication();
+            _app.UseAuthorization();
+
+            #endregion
+
             _app.MapGrpcService<DashboardService>();
         }
         catch (Exception ex)
```

The config must be updated to ensure `DOTNET_RESOURCE_SERVICE_ENDPOINT_URL` is set to `localhost`, as by default it finds a random port on `127.0.0.1` which then fails certificate name validation. The host name must match the CN of the cert.

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -33,6 +35,9 @@ namespace Aspire.Dashboard.Model;
 internal sealed class DashboardClient : IDashboardClient
 {
     private const string ResourceServiceUrlVariableName = "DOTNET_RESOURCE_SERVICE_ENDPOINT_URL";
+    private const string ResourceServiceDisableAuthVariableName = "DOTNET_RESOURCE_SERVICE_DISABLE_AUTH";
+    private const string ResourceServiceClientCertificatePathVariableName = "DOTNET_RESOURCE_SERVICE_CLIENT_CERTIFICATE_PATH";
```
JamesNK, line 39: File path won't be enough on its own. There will be people that want to use a cert from the store. I know of security standards where deploying certificates isn't allowed. They must come from the OS store.

JamesNK: What ASP.NET Core has for getting certs for endpoints: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-8.0#certificate-sources

* Path and Password to load .pfx files.
* Path, KeyPath and Password to load .pem/.crt and .key files.
* Subject and Store to load from the certificate store.

src/Aspire.Dashboard/README.md
JamesNK: There is an `DOTNET_DASHBOARD_OTLP_AUTH_MODE` with a setting of None. It defaults to None. An upcoming change for security is to make it not have a default value so people explicitly say none.

I think an enum like what OTLP is doing is better because it can serve double duty: a way to explicitly say what auth you want (including none) and a switch for multiple auth options (which could become a thing in the future if not a reality today).

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -685,6 +685,7 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
 
             context.EnvironmentVariables["ASPNETCORE_URLS"] = appHostApplicationUrl;
             context.EnvironmentVariables["DOTNET_RESOURCE_SERVICE_ENDPOINT_URL"] = grpcEndpointUrl;
+            context.EnvironmentVariables["DOTNET_RESOURCE_SERVICE_DISABLE_AUTH"] = "1"; // No auth in local dev experience
```
JamesNK, line 688: What's the prior art on 1/0 vs true/false?

I prefer true/false strings but my mind is open.

JamesNK:
> We currently don't have a resource service implementation that works with certificates.

Integration tests for this would be valuable. You can create a very simple ASP.NET Core app with a dummy OTLP endpoint and cert auth, then call it from the dashboard.

I added a number of integration tests for various OTLP scenarios. You can look to them as examples for how to do integration tests with ASP.NET Core apps.

Launching the dashboard to call the OTLP endpoint would be the most realistic, but I don't know a could way to then assert results. Could just test DashboardClient in isolation.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
drewnoakes: In the Aspire repo, `DOTNET_WATCH` uses `0` and `1`, so I followed that pattern for consistency. The `GetBool` extension method will handle `true` and `false` values too, however. I don't have a strong opinion either way.

drewnoakes:
> Integration tests for this would be valuable

This is on my radar. I won't have time to add them before the p5 snap.

src/Aspire.Dashboard/README.md
drewnoakes: We can use an enum here. I guess the values would be `None` or `Certificate` (`ClientCertificate`?). Other modes, such as `ApiKey`, aren't planned for now.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: I would use true/false. 1/0 would be supported by nature of using that helper.

JamesNK: `GetBool` supporting reading 0/1 is ok, but I think our docs should always show true/false, and values set by Aspire host should be true/false.

drewnoakes: I see more precedence for 0/1 than for true/false, so I'm inclined to leave this as is.

- `DOTNET_CLI_TELEMETRY_OPTOUT ` https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry#how-to-opt-out
- `DOTNET_RUNNING_IN_CONTAINER` https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#jit-memory-barriers
- `DOTNET_EnableDiagnostics` https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_enablediagnostics
- `DOTNET_MULTILEVEL_LOOKUP` https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_multilevel_lookup

drewnoakes: It's a bit of a mess really. Looking through https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables, you see some that take true/false, but most of them use 1/0.

drewnoakes: So do we want to break with the norm and use true/false in docs/templates/playgrounds/samples/etc?

src/Aspire.Dashboard/Model/DashboardClient.cs
drewnoakes: Pushed an update that supports key stores as well. The configuration support for Kestrel doesn't seem applicable to gRPC clients (or at least I couldn't work out how to do it).

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
drewnoakes: I'm going to convert these to values like `ResourceServiceClient:Insecure:DisableAuth` so will doc as `true`, seeing as the 0/1 seems more aligned with the env var presentation in .NET.

davidfowl: I agree

drewnoakes:
@kvenkatrajan @JamesNK @davidfowl this is ready for review.

I'll investigate the test failure tomorrow.

davidfowl:
It would be great to have an example using this with keycloak. Could be a playground project (and sample). Not required to get this in but we should do it (especially since it'll serve as a way of testing the functionality).

src/Aspire.Dashboard/Components/_Imports.razor
```diff
@@ -10,3 +11,6 @@
 @using Aspire.Dashboard.Components
 @using Aspire.Dashboard.Components.Controls
 @using Microsoft.Extensions.Localization
+
+@* Require authorization for all pages *@
+@attribute [Authorize]
```
davidfowl, line 16: Do we show the user information anywhere?

davidfowl:
@javiercn @halter73 @SteveSandersonMS can I get you to take a look at how we're doing auth in Blazor server? One thing I think we may want to do is show the user information somewhere.

@kvenkatrajan @leslierichardson95 do we want more UI around logging in? The current mode will just challenges for auth immediately without a user gesture.

javiercn:
@davidfowl I've taken a look and it looks good.

The user should be automatically available in Blazor during SSR render and in interactive Server.

davidfowl:
Should we be using the top level attribute or the authorize view on the router so we can show login screen options?

javiercn:
> Should we be using the top level attribute or the authorize view on the router so we can show login screen options?

Do you mean the AuthorizeView with the `NotAuthorized` fragment?

src/Aspire.Dashboard/Components/_Imports.razor
drewnoakes: Not yet. I would like to, along with a sign out button.

I didn't find a way to perform OIDC sign out from Blazor server though. Everything I saw doc'd revolved around using HttpContext in action handler methods, which doesn't work in Blazor server, afaict, throwing errors about the headers being read only because the response has started.

drewnoakes:
> It would be great to have an example using this with keycloak. Could be a playground project (and sample). Not required to get this in but we should do it (especially since it'll serve as a way of testing the functionality).

This would be possible for the web app's auth, but not for the resource service's.

Would the user need to provide their own OIDC authority, client ID and secret? I'm not familiar with keycloak, if that's related.

davidfowl:
> This would be possible for the web app's auth, but not for the resource service's.

Yep.

> Would the user need to provide their own OIDC authority, client ID and secret? I'm not familiar with keycloak, if that's related.

Yes. We have an example of using [keycloak here](https://github.com/dotnet-presentations/eshop-app-workshop/blob/caf3ec803e4ac5f1361fbcb1592b592258c26a85/src/eShop.AppHost/Program.cs#L14-L15). @DamianEdwards can help since he spent all of that time setting it up

src/Aspire.Dashboard/Components/_Imports.razor
davidfowl: AFAIK you can use the AuthenticationStateProvider

davidfowl: BTW why is this always on? We don't have auth when running locally right?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -103,6 +111,30 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
 
         builder.Services.AddLocalization();
 
+        if (authMode == DashboardWebAppAuthMode.OpenIdConnect)
```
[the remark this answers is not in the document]
DamianEdwards: Should we not put this in the `ConfigureAuthentication` method called on line 86?

src/Aspire.Dashboard/Components/_Imports.razor
DamianEdwards: Should we make this more explicit by creating an authorization policy and referring to it here?

drewnoakes: > why is this always on?

My observation was that `[Authorize]` only does anything when you actually have authorization middleware configured. The dashboard only does that when you opt-in to the `OpenIdConnect` auth mode.

DamianEdwards: Isn't it always turned on here? https://github.com/dotnet/aspire/pull/3033/files#diff-ccedb77216895077db571ce0a3c547948ca85798f4ce0523e4eb936216f27f15L196

DamianEdwards: > I can look into that. What benefit would that have?

It's a bit confusing now that there's so many authentication schemes being configured, and different schemes between the OTLP endpoint and the rest of the dashboard UI. Having the UI explicitly configured to require authn/z based on the scheme that was configured for dashboard auth will help make it clearer IMO.

JamesNK: Specify a policy name on the attribute. Makes it easier to understand if each part of the dashboard has a named policy for auth.

OTLP services is already doing this.

JamesNK: https://github.com/dotnet/aspire/blob/0783a609d369634c1a0aa5d399b78c79a52caafe/src/Aspire.Dashboard/Otlp/Grpc/OtlpLogsService.cs#L14

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -62,6 +63,13 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
         builder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
 #endif
 
+        var authMode = builder.Configuration.GetEnum<DashboardWebAppAuthMode>("DashboardWebApp:AuthMode");
+
+        if (authMode == null)
+        {
+            throw new InvalidOperationException("Unable to read DashboardWebApp:AuthMode. Supported values are Unsecured and OpenIdConnect.");
+        }
```
JamesNK, line 71: Move this configuration into LoadDashboardConfig alongside other custom config.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -408,6 +440,12 @@ private static DashboardStartupConfiguration LoadDashboardConfig(IConfiguration
             OtlpApiKey = otlpApiKey
         };
     }
+
+    private enum DashboardWebAppAuthMode
+    {
+        Unsecured,
```
JamesNK, line 446: Unsecured vs None? I went with None but these should be consistent. Can you change the OTLP auth enum to use Unsecured. And update readme.

drewnoakes: Yes, I'll make a pass to get things consistent. We should also review the use of `A:B:C` vs `DOTNET_*` style config throughout the dashboard.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -103,6 +111,30 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
 
         builder.Services.AddLocalization();
 
+        if (authMode == DashboardWebAppAuthMode.OpenIdConnect)
+        {
+            // Configure OpenID Connect (OIDC)
+            var authentication = builder.Services.AddAuthentication(options =>
```
JamesNK, line 117: This should be in the ConfigureAuthentication method

src/Aspire.Dashboard/DashboardWebApplication.cs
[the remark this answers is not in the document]
JamesNK: Yes, it should be in ConfigureAuthentication.

JamesNK:
There are failing integration tests. Even with an Unsecured setting, requests still require auth.

I used `OtlpCompositeAuthenticationHandler` to meet the authorization requirement when auth was disabled. I don't know if it's the best way, but it worked for me.

How much testing have you done that the different endpoint's auth don't conflict with each other? I couldn't see code to select which auth is run, although I haven't that before so I'm not sure how to go about doing it.

src/Aspire.Dashboard/DashboardWebApplication.cs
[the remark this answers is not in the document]
drewnoakes: Merging them flags the question of what then becomes the default scheme. In `ConfigureAuthentication` it's `OtlpComposite`, in my new code it's `Cookies`.

Seems like we want quite different config for the web app and OTLP, but they're both in the same `WebApplicationBuilder`. I'll follow up on the comments below around the use of policies to see if that helps somehow.

JamesNK:
At the very least I think one integration test with browser auth enabled and the server returning a 401 is required to ensure this is working.

There are a lot of complex moving parts here and we don't develop with auth enabled. It is very easy to create regressions. The integration tests around OTLP auth have been merged only a week and have caught a number already.

src/Aspire.Dashboard/DashboardWebApplication.cs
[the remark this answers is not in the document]
drewnoakes: Worked out how to do this. We need a single call to `AddAuthentication`, after which we register various authentication schemes. Then we have different policies for OTLP and WebApp, and configure each to use the authentication schemes we want. In this way, we can avoid having a default authentication scheme at all.

drewnoakes:
> At the very least I think one integration test with browser auth enabled and the server returning a 401 is required to ensure this is working.

I took a look at this today, as I agree a simple test would be good. We want to validate that we receive a `302 Found` with a `Location` at the OpenID provider to which the user would be redirected.

The challenge with this is that it has to be a live OpenId provider, as the server internally reaches out to `https://<authority>/.well-known/openid-configuration`. I'm going to investigate spinning up a dummy OpenId provider during the test. I might not complete that in time for p5.

src/Aspire.Dashboard/Components/_Imports.razor
drewnoakes: A policy's been added for the web app's HTTP traffic. Now, when OpenIdConnect is selected, requests are challenged correctly. OTLP routes to its authentication and policy.

Today I learned what authentication schemes and authorization policies are, then produced the update here. The approach looks good (for all I know) but would appreciate expert reviews from y'all.

drewnoakes: @JamesNK I removed the default authentication scheme and made the OTLP policy reference that scheme explicitly.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -382,55 +447,64 @@ private static DashboardStartupConfiguration LoadDashboardConfig(IConfiguration
...
+            throw new InvalidOperationException($"At least one URL for Aspire dashboard browser endpoint with {DashboardOtlpUrlDefaultValue} is required.");
         }
 
+        var webAppAuthMode = configuration.GetEnum<WebAppAuthMode>(WebAppAuthModeKey);
+
         var otlpUris = configuration.GetUris(DashboardOtlpUrlVariableName, new(DashboardOtlpUrlDefaultValue));
         if (otlpUris.Length != 1)
         {
-            throw new InvalidOperationException($"At least one URL for Aspire dashboard browser endpoint with {DashboardOtlpUrlDefaultValue} is required.");
+            throw new InvalidOperationException($"One URL for Aspire dashboard OTLP endpoint with {DashboardUrlDefaultValue} is required, not {otlpUris.Length}.");
```
drewnoakes, line 458: These two exception messages were reversed somehow and would have been misleading.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -382,55 +447,64 @@ private static DashboardStartupConfiguration LoadDashboardConfig(IConfiguration
...
         }
 
-        var allowAnonymous = configuration.GetBool(DashboardInsecureAllowAnonymousVariableName) ?? false;
-        var otlpAuthMode = configuration.GetEnum<OtlpAuthMode>(DashboardOtlpAuthModeVariableName);
+        var otlpAuthMode = configuration.GetEnum<OtlpAuthMode>(OtlpAuthModeKey);
```
drewnoakes, line 461: This overload of `GetEnum` throws if the value is blank or unknown, so we no longer need to test for `null` below. Ditto elsewhere in this PR.

drewnoakes:
I've pushed an update that unifies the OTLP and web app authentication/authorization code where possible, and have validated both work as expected.

Also fixed conflicts with main.

@JamesNK, as discussed offline, some OTLP configuration values are renamed in this PR:

- `DOTNET_DASHBOARD_OTLP_AUTH_MODE` → `Otlp:AuthMode`
- `DOTNET_DASHBOARD_OTLP_API_KEY` → `Otlp:ApiKey`
- `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS` is no longer needed → set `Otlp:AuthMode` to `Unsecured`
- For `Otlp:AuthMode`, the option `None` → `Unsecured`

src/Aspire.Dashboard/Components/_Imports.razor
drewnoakes: Have captured the ideas here in a TODO list in the PR description.

src/Aspire.Dashboard/Authentication/WebAppAuthenticationHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Security.Claims;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Authentication;
+
+public sealed class WebAppAuthenticationHandler(
```
JamesNK, line 13: I've been thinking about what to call the non-OTLP endpoint. I like Frontend. It's very clear what the difference is between frontend and OTLP.

Also, for consistency with the Otlp auth that does this, it should have composite in the name.

So, `FrontendCompositeAuthenticationHandler`.

src/Aspire.Dashboard/Components/_Imports.razor
```diff
@@ -10,3 +13,6 @@
 @using Aspire.Dashboard.Components
 @using Aspire.Dashboard.Components.Controls
 @using Microsoft.Extensions.Localization
+
+@* Require authorization for all pages of the web app *@
+@attribute [Authorize(Policy = WebAppAuthorizationDefaults.PolicyName)]
```
JamesNK, line 18: FrontendAuthorizationDefaults.PolicyName

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -12,19 +12,25 @@
...
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace Aspire.Dashboard;
 
 public class DashboardWebApplication : IAsyncDisposable
 {
-    internal const string DashboardInsecureAllowAnonymousVariableName = "DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS";
-    internal const string DashboardOtlpAuthModeVariableName = "DOTNET_DASHBOARD_OTLP_AUTH_MODE";
-    internal const string DashboardOtlpApiKeyVariableName = "DOTNET_DASHBOARD_OTLP_API_KEY";
+    internal const string WebAppAuthModeKey = "DashboardWebApp:AuthMode";
```
JamesNK, line 29: I want to put all the dashboard config in a `Dashboard` container.
```suggestion
    internal const string FrontendAuthModeKey = "Dashboard:Frontend:AuthMode";
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -12,19 +12,25 @@
...
+
+    internal const string OtlpAuthModeKey = "Otlp:AuthMode";
+    internal const string OtlpApiKeyKey = "Otlp:ApiKey";
```
JamesNK, line 32:
```suggestion
    internal const string OtlpAuthModeKey = "Dashboard:Otlp:AuthMode";
    internal const string OtlpApiKeyKey = "Dashboard:Otlp:ApiKey";
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -358,9 +362,70 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
                 };
             });
 
+        if (dashboardStartupConfig.WebAppAuthMode == WebAppAuthMode.OpenIdConnect)
```
JamesNK, line 365: Is there a downside to always registering these?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -358,9 +362,70 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
...
+                        .RequireAuthenticatedUser()
+                        .Build());
+            }
+            else
+            {
+                options.AddPolicy(
+                    name: WebAppAuthorizationDefaults.PolicyName,
+                    policy: new AuthorizationPolicyBuilder()
+                        .RequireAssertion(_ => true)
+                        .Build());
```
JamesNK, line 427: What about doing what OTLP does and require a claim? The handler registers the required claim if auth mode is Unsecured. Avoids the need for different policies.

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -98,6 +100,28 @@ GrpcChannel CreateChannel()
                 KeepAlivePingPolicy = HttpKeepAlivePingPolicy.WithActiveRequests
             };
 
+            var authMode = configuration.GetEnum<ResourceClientAuthMode>("ResourceServiceClient:AuthMode");
```
JamesNK, line 103:
```suggestion
            var authMode = configuration.GetEnum<ResourceClientAuthMode>("Dashboard:ResourceServiceClient:AuthMode");
```

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -98,6 +100,28 @@ GrpcChannel CreateChannel()
...
+
+            if (authMode == ResourceClientAuthMode.Certificate)
+            {
+                // Auth hasn't been suppressed, so configure it.
+                var sourceType = configuration.GetEnum<DashboardClientCertificateSource>("ResourceServiceClient:ClientCertificate:Source");
```
JamesNK, line 108:
```suggestion
                var sourceType = configuration.GetEnum<DashboardClientCertificateSource>("Dashboard:ResourceServiceClient:ClientCertificate:Source");
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -12,19 +12,25 @@
...
 using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.FluentUI.AspNetCore.Components;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace Aspire.Dashboard;
 
 public class DashboardWebApplication : IAsyncDisposable
 {
-    internal const string DashboardInsecureAllowAnonymousVariableName = "DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS";
```
JamesNK, line 25: I still want this config key to set frontend and OTLP auth modes to Unsecured. The most common usage of the dashboard image will be someone playing around with it on their local machine. They won't want to use auth.

This is a command line/env var friendly way (no double underscore escape) to set both values to unsecured.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -382,55 +447,64 @@ private static DashboardStartupConfiguration LoadDashboardConfig(IConfiguration
         var browserUris = configuration.GetUris(DashboardUrlVariableName, new(DashboardUrlDefaultValue));
         if (browserUris.Length == 0)
         {
-            throw new InvalidOperationException($"One URL for Aspire dashboard OTLP endpoint with {DashboardUrlDefaultValue} is required.");
+            throw new InvalidOperationException($"At least one URL for Aspire dashboard browser endpoint with {DashboardOtlpUrlDefaultValue} is required.");
```
JamesNK, line 450: The changes here are fine, but they'll be replaced with my PR to use options.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -382,55 +447,64 @@ private static DashboardStartupConfiguration LoadDashboardConfig(IConfiguration
...
             BrowserUris = browserUris,
+            WebAppAuthMode = webAppAuthMode,
             OtlpUri = otlpUris[0],
-            OtlpAuthMode = otlpAuthMode.Value,
+            OtlpAuthMode = otlpAuthMode,
             OtlpApiKey = otlpApiKey
         };
     }
+
+    private enum WebAppAuthMode
```
JamesNK, line 494:
```suggestion
    private enum FrontendAuthMode
```

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -124,6 +148,44 @@ GrpcChannel CreateChannel()
                     LoggerFactory = _loggerFactory,
                     ThrowOperationCanceledOnCancellation = true
                 });
+
+            X509CertificateCollection GetFileCertificate()
+            {
+                var filePath = configuration["ResourceServiceClient:ClientCertificate:FilePath"];
+                var password = configuration["ResourceServiceClient:ClientCertificate:Password"];
```
JamesNK, line 155:
```suggestion
                var filePath = configuration["Dashboard:ResourceServiceClient:ClientCertificate:FilePath"];
                var password = configuration["Dashboard:ResourceServiceClient:ClientCertificate:Password"];
```

src/Aspire.Dashboard/Model/DashboardClient.cs
JamesNK: What do you think about changing `ResourceServiceClient` to `ResourcesClient`. It's a bit shorter.

src/Aspire.Dashboard/README.md
```diff
@@ -17,13 +17,38 @@ For example, the default certificate used by HTTPS endpoints can be configured u
...
+    - `ResourceServiceClient:ClientCertificate:Password` (optional, string)
+  - `KeyStore` to load the cert from a key store, configured with:
+    - `ResourceServiceClient:ClientCertificate:Subject` (required, string)
+    - `ResourceServiceClient:ClientCertificate:KeyStore:Name` (optional, [`StoreName`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.storename), defaults to `My`)
+    - `ResourceServiceClient:ClientCertificate:KeyStore:Location` (optional, [`StoreLocation`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.storelocation), defaults to `CurrentUser`)
+- `ResourceServiceClient:Ssl` (optional, [`SslClientAuthenticationOptions`](https://learn.microsoft.com/dotnet/api/system.net.security.sslclientauthenticationoptions))
+
+To opt-out of authentication, set `ResourceServiceClient:AuthMode` to `Unsecured`. This completely disables all security for the resource service client. This setting is used during local development, but is not recommended if you attempt to host the dashboard in other settings.
+
+## Dashboard web app auth
```
JamesNK, line 41:
```suggestion
## Dashboard frontend authentication
```

JamesNK: This section should be next to `OTLP endpoint authentication`. It's odd to have resources client inbetween the two endpoint auth sections.

src/Aspire.Dashboard/README.md
```diff
@@ -17,13 +17,38 @@ For example, the default certificate used by HTTPS endpoints can be configured u
...
+
+The dashboard's web application supports OpenID Connect. Set `DashboardWebApp:AuthMode` to `OpenIdConnect`, then add the following configuration:
+
+- `Authentication:Schemes:OpenIdConnect:Authority` &mdash; URL to the identity provider (IdP)
```
JamesNK, line 45: This is fine for now but I'd like to see if we have move this auth configuration to `Dashboard:Frontend:OpenIdConnect`. I think we can do that by binding another config section.

Can happen in a future PR.

src/Aspire.Dashboard/README.md
```diff
@@ -17,13 +17,38 @@ For example, the default certificate used by HTTPS endpoints can be configured u
...
+    - `ResourceServiceClient:ClientCertificate:FilePath` (required, string)
+    - `ResourceServiceClient:ClientCertificate:Password` (optional, string)
+  - `KeyStore` to load the cert from a key store, configured with:
+    - `ResourceServiceClient:ClientCertificate:Subject` (required, string)
+    - `ResourceServiceClient:ClientCertificate:KeyStore:Name` (optional, [`StoreName`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.storename), defaults to `My`)
+    - `ResourceServiceClient:ClientCertificate:KeyStore:Location` (optional, [`StoreLocation`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.storelocation), defaults to `CurrentUser`)
+- `ResourceServiceClient:Ssl` (optional, [`SslClientAuthenticationOptions`](https://learn.microsoft.com/dotnet/api/system.net.security.sslclientauthenticationoptions))
+
+To opt-out of authentication, set `ResourceServiceClient:AuthMode` to `Unsecured`. This completely disables all security for the resource service client. This setting is used during local development, but is not recommended if you attempt to host the dashboard in other settings.
```
JamesNK, line 39: Is `ResourceServiceClient:AuthMode` required? I'm guessing it is required if `DOTNET_RESOURCE_SERVICE_ENDPOINT_URL` is specified.

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
```diff
@@ -29,8 +28,11 @@ public static class IntegrationTestHelpers
         {
             [DashboardWebApplication.DashboardUrlVariableName] = "http://127.0.0.1:0",
             [DashboardWebApplication.DashboardOtlpUrlVariableName] = "http://127.0.0.1:0",
-            [DashboardWebApplication.DashboardOtlpAuthModeVariableName] = nameof(OtlpAuthMode.None)
+            [DashboardWebApplication.OtlpAuthModeKey] = "Unsecured",
+            ["ResourceServiceClient:AuthMode"] = "Unsecured",
```
JamesNK, line 32: Add a constant

tests/Aspire.Dashboard.Tests/Integration/StartupTests.cs
```diff
@@ -143,26 +135,25 @@ public async Task Configuration_NoOtlpAuthMode_Error()
...
     public async Task Configuration_AllowAnonymous_NoError()
     {
         // Arrange
-        await using var app = IntegrationTestHelpers.CreateDashboardWebApplication(_testOutputHelper,
+        await using var app = IntegrationTestHelpers.CreateDashboardWebApplication(testOutputHelper,
             additionalConfiguration: data =>
             {
-                data.Remove(DashboardWebApplication.DashboardOtlpAuthModeVariableName);
-                data[DashboardWebApplication.DashboardInsecureAllowAnonymousVariableName] = bool.TrueString;
+                data[DashboardWebApplication.OtlpAuthModeKey] = "Unsecured";
```
JamesNK, line 156: Change back.
Also remove frontend auth mode default.

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
```diff
@@ -29,8 +28,11 @@ public static class IntegrationTestHelpers
         {
             [DashboardWebApplication.DashboardUrlVariableName] = "http://127.0.0.1:0",
             [DashboardWebApplication.DashboardOtlpUrlVariableName] = "http://127.0.0.1:0",
-            [DashboardWebApplication.DashboardOtlpAuthModeVariableName] = nameof(OtlpAuthMode.None)
+            [DashboardWebApplication.OtlpAuthModeKey] = "Unsecured",
```
JamesNK, line 31: Use name of instead of strings, e.g. `nameof(OtlpAuthMode.Unsecured)`

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
JamesNK: This shouldn't be a default because no resource endpoint is specified.

halter73:
> I'm going to investigate spinning up a dummy OpenId provider during the test.

You can take a look at a dummy OpenId provider I just added to test a trimming issue in Microsoft.AspNetCore.Components.WebAssembly.Authentication. I only did enough for oidc-client-js to let me complete the flow. It's possible that the server-side OpenIdConnectHandler might need more endpoints for stuff like JWKS, but it should still serve as a decent starting point.

https://github.com/dotnet/aspnetcore/blob/3f99d45b0b7d8f0427a3d98acc63098694613362/src/Components/test/testassets/Components.TestServer/RemoteAuthenticationStartup.cs#L37-L94

> Should we be using the top level attribute or the authorize view on the router so we can show login screen options?

As you know from https://github.com/dotnet/aspire/pull/3011, endpoint routing doesn't integrate super cleanly with older Blazor concepts like `<NotFound>`. Same goes for top-level authorize attributes and `<AuthorizeView>/<NotAuthorized>`, because the authorization middleware will have the authentication handler issue a challenge before Blazor gets a chance to interact with the request.

I recommend using the top-level authorize attribute for when the entire page needs authorization. You just need to setup the login redirect using the authentication handler like you would do for any typical server rendered MVC/Razor Pages/Minimal APIs/etc... app. `<AuthorizeView>` is still useful for non-page components that need to show different content based on whether the user is logged in or not. For example, we use this to show a [login/logout link/form in the Blazor project template](https://github.com/dotnet/aspnetcore/blob/3f99d45b0b7d8f0427a3d98acc63098694613362/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Layout/NavMenu.razor#L44-L73).

Since [you're using global InteractiveServerRenderMode](https://github.com/drewnoakes/aspire/blob/b7a165fe872396ae54d0dfe6613a19c75db469cf/src/Aspire.Dashboard/Components/App.razor#L28), normal server-side auth approaches should work fine. If you want to try auto render mode (which I don't necessarily recommend), you can look at the custom [server](https://github.com/dotnet/aspnetcore/blob/3f99d45b0b7d8f0427a3d98acc63098694613362/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Account/PersistingRevalidatingAuthenticationStateProvider.cs#L39) and [client](https://github.com/dotnet/aspnetcore/blob/3f99d45b0b7d8f0427a3d98acc63098694613362/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp.Client/PersistentAuthenticationStateProvider.cs#L24) `AuthenticationStateProvider` implementations in the Blazor project that use `PersistentComponentState` to flow a customized subset of claims from the server to the client. This is also demonstrated in the latest [Secure an ASP.NET Core Blazor Web App with OIDC docs](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?view=aspnetcore-8.0&pivots=without-bff-pattern) which go further to demonstrate using the BFF pattern so the client APP can authenticate with a backend using the JwtBearerHandler via a YARP proxy.

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: I mentioned on another PR that having a single kill switch for all security measures means that users won't get the opportunity to knowingly opt-out of future security features. We want each to be opted out individually to prevent that.

Very few users will need to ever set these config keys. Those who do will be guided by documentation and helpful error messages. The safety outweighs the convenience factor here.

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: Loads a bunch of code into the process that most users will never need. If there's an upside I'm missing, let me know. We're unlikely to ever make the auth settings changeable throughout the process lifetime.

src/Aspire.Dashboard/Model/DashboardClient.cs
drewnoakes: The "Resource Service" is a concept that exists in the system. This is a client for that concept. I don't think shortening it is worth the ambiguity.

src/Aspire.Dashboard/README.md
drewnoakes: It's currently required in all cases.

src/Aspire.Dashboard/README.md
drewnoakes: `Authentication:Schemes:<schemename>` seems to be the de facto way of doing this, no? If we move to our own config container, I'd be concerned we miss out on other built-in configuration controls.

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
drewnoakes: Enum is not public.

tests/Aspire.Dashboard.Tests/Integration/StartupTests.cs
drewnoakes: Discussed earlier.

src/Aspire.Dashboard/Authentication/WebAppAuthenticationHandler.cs
drewnoakes: I pushed an update that renames "web app" to "frontend" throughout.

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: I'll contact experts offline.

src/Aspire.Dashboard/README.md
JamesNK: Schemes include functionality to bind their options to a convention-based location in `IConfiguration`. It's 1 line of code to add support for binding to an extra location.

I don't know yet whether we want to do this. If there are sacrifices, then it might be left at the default location.

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: What would requiring a claim give us? In the OLTP case it seems that any request passing through the `OtlpCompositeAuthenticationHandler` is given that claim during authentication, which is then validated in authorization. Given the authentication scheme is only used by that policy, using a claim seems redundant. I'd vote for removing it from the OLTP code and replacing `OtlpCompositeAuthenticationHandler` with `AddPolicyScheme` instead (similar to 1db68e4984d50fda5f5e80f385eef6bbeb076f61), though maybe I'm missing something.

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: Is it idiomatic that an app makes its own namespace in its own config file? I don't mind either way. Perhaps you can do that in #3119 after this merges.

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: Actually I'm not sure that `OtlpCompositeAuthenticationHandler` can be given the same treatment as it uses multiple authentication schemes.

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: It doesn't need to be in its own config file. You can mix settings in config files. This happens in appsettings.json.

Yes, it's normal to have configuration for a thing sit under its own section. For example, all of Kestrel's settings are under `Kestrel`, all of logging's settings are under `Logging`, etc.

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: Yes, there are multiple schemes of OTLP.

I like the consistency of auth mode for both endpoints being checked in auth handlers, but what is here is ok if it works. Can always change it in the future if we want.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -358,9 +362,76 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
...
+                // Frontend is unsecured so our policy doesn't need any special handling.
+                options.AddPolicy(
+                    name: FrontendAuthorizationDefaults.PolicyName,
+                    policy: new AuthorizationPolicyBuilder()
+                        .RequireAssertion(_ => true)
+                        .Build());
+            }
+            else
+            {
+                throw new NotImplementedException($"Unexpected {nameof(FrontendAuthMode)} enum member.");
```
davidfowl, line 433: nit: NotSupportedException

drewnoakes: This code should be unreachable. There should be no path that gets here.

```suggestion
                throw new NotSupportedException($"Unexpected {nameof(FrontendAuthMode)} enum member.");
```

JamesNK (approved):

davidfowl (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: Actually `OtlpCompositeAuthenticationHandler` _can_ be converted to use `AddPolicyScheme` along with `ForwardDefaulSelector`, with a callback to dispatch accordingly. That would clean the code up a bit.

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -124,6 +148,44 @@ GrpcChannel CreateChannel()
...
+                var subject = configuration["ResourceServiceClient:ClientCertificate:Subject"];
+
+                if (subject is null or [])
+                {
+                    throw new InvalidOperationException("ResourceServiceClient:ClientCertificate:Source is \"KeyStore\", but no Certificate:FilePath is configured.");
+                }
+
+                using var store = new X509Store(storeName: StoreName.My, storeLocation: StoreLocation.CurrentUser);
+
+                configuration.Bind("ResourceServiceClient:ClientCertificate:KeyStore");
```
JamesNK, line 176: I'm updating this code to use strongly typed options, and I noticed this line. I don't think it does anything.

I'm guessing you were intending to make store configurable like Kestrel, but didn't finish.

drewnoakes: My understanding of that line is that it allows config values to apply to the store object. 

The idea is documented in https://github.com/dotnet/aspire/tree/main/src/Aspire.Dashboard#resources

JamesNK: This line is trying to bind IConfiguration values to a string: [Bind(IConfiguration, Object)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationbinder.bind?view=dotnet-plat-ext-8.0#microsoft-extensions-configuration-configurationbinder-bind(microsoft-extensions-configuration-iconfiguration-system-object))

drewnoakes: Thanks, yes that's a bug.

drewnoakes: Fixed in #3139.

PR #3035, opened 2024-03-20
Change console log line number color

### LINKED ISSUE #2842: Console logs: Different color for line numbers and text
AZDO has a slightly different color for line numbers:
image

GitHub actions has a slightly different color for line numbers:
image

Aspire currently **doesn't**:
image

IMO the line number text should be slightly different to show that's not part of the content and won't be selected with other text.

tlmii:
Resolves #2842 

After chatting with @leslierichardson95, here's where I landed:

Before:
image
After:
image

That's about as light as we can go for the line numbers while maintaining contrast. I also changed the timestamp to be brighter, like in Docker Desktop (though not as bright), rather than dimmer. 

I _think_ I preferred the timestamp being dimmed to keep it separate from the main content... but I do like having the line numbers dimmed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3035)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
+++ b/src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
@@ -1,15 +1,40 @@
 .log-overflow {
+    --console-theme-black: #0C0C0C;
+    --console-theme-blue: #0037DA;
+    --console-theme-cyan: #3A96DD;
+    --console-theme-green: #13A10E;
+    --console-theme-magenta: #CD13E8;
+    --console-theme-bg-red: #F80F24;
+    --console-theme-fg-red: #C50F1F;
+    --console-theme-white: #CCCCCC;
+    --console-theme-yellow: #C19C00;
+
+    --console-theme-bright-black: #767676;
+    --console-theme-bright-blue: #3B78FF;
+    --console-theme-bright-cyan: #61D6D6;
+    --console-theme-bright-green: #16C60C;
+    --console-theme-bright-magenta: #DA01FA;
+    --console-theme-bright-red: #E74856;
+    --console-theme-bright-white: #F2F2F2;
+    --console-theme-bright-yellow: #F9F1A5;
+
+    --line-number-color: #848484;
+    --timestamp-color: var(--line-number-color);
+    --console-background-color: var(--console-theme-black);
+    --console-background-color-hover: #1D1D1D;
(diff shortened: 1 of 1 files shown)


JamesNK:
> I think I preferred the timestamp being dimmed to keep it separate from the main content... but I do like having the line numbers dimmed.

Timestamp should be dimmed. It's not as important as other content and darker color helps focus on other content.

IMO timestamp and line number should both be dimmed but different amounts.

tlmii:
> IMO timestamp and line number should both be dimmed but different amounts.

There's not a lot of space for two noticeably different dimmed amounts before we fail contrast checks. But I'll give it a shot. Looking at the before/after more I really do think the dimmed timestamp focuses the important content a lot better.

JamesNK:
Could also consider increasing the spacing between line numbers and content. VS code for example:

image

They don't need to be so close together.

drewnoakes:
I like the line after screenshot a lot.

While we're at it, why not trim the date off the front of the value if it's today's date (in local time)? We do this in the resource grid, and it makes the UI a bit cleaner.

drewnoakes:
> Could also consider increasing the spacing between line numbers and content.

The font size could be a little larger too, IMO.

ilmax:
> I like the line after screenshot a lot.
> 
> While we're at it, why not trim the date off the front of the value if it's today's date (in local time)? We do this in the resource grid, and it makes the UI a bit cleaner.

I'm on the other side of the spectrum and this seems like a regression to me, 99% of the time I don't care about time until I find what I'm looking for in the log messages

tlmii:
OK, here's another stab at it.

- Increased the spacing between the line number of the timestamp/content to match AzDO
- Added hover color to line background similar to AzDO/GH logs and tweaked the line number color slightly for valid contrast
- Changed the timestamp back to dimmed, same color as line number
- Reorganized all the "console theme" colors to into variables to make it easier to modify them (and, theoretically, to make it easier to have actual console themes in the future)

Original before:
image
First attempt above:
image
After:
image

I _think_ this works for me. Separation between line number and text helps to differentiate the actual log entry from the line number. Dimmed timestamp keeps the focus on the content. Background highlight on hover helps with the recognition of individual log entries.

kvenkatrajan (approved):

JamesNK (approved):

PR #3036, opened 2024-03-20
Always require OTLP auth mode config when starting dashboard

JamesNK:
Require OTLP auth mode when starting the dashboard so users must make an explicit decision to have no auth.

@davidfowl FYI, this was review feedback.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3036)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -22,6 +22,7 @@ namespace Aspire.Dashboard;
 
 public class DashboardWebApplication : IAsyncDisposable
 {
+    internal const string DashboardInsecureAllowAnonymousVariableName = "DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS";
     internal const string DashboardOtlpAuthModeVariableName = "DOTNET_DASHBOARD_OTLP_AUTH_MODE";
     internal const string DashboardOtlpApiKeyVariableName = "DOTNET_DASHBOARD_OTLP_API_KEY";
     internal const string DashboardOtlpUrlVariableName = "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL";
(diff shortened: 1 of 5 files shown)


JamesNK:
TODO: update standalone dashboard docs and sample

davidfowl:
Instead of throwing, can we make it show a page that shows the error and explains what to do?

JamesNK:
It's possible, but it would be a lot more work. We'd need to make the new UI, put checks throughout the app disable UI and functionality, e.g. ignore incoming OTLP data, don't allow visiting regular UI, hide links to regular UI.

JamesNK:
Since there is auth for OTLP **and** auth for the browser, it would be annoying to have to explicitly set both to `None`. Maybe there should be a config setting such as `DOTNET_DASHBOARD_ALLOW_ANONYMOUS` that defaults the modes to none.

What do you think @drewnoakes @davidfowl?

drewnoakes:
There are security considerations in having single variable to opt out for all security features. If we add new kinds of security in future, should users that opted out before be automatically opted out of the new security too? The user may have reviewed the decision during installation. They may wish to review newer features as well.

If we are going to have a single variable, its name should evoke considerable caution. Something stronger than `DOTNET_DASHBOARD_ALLOW_ANONYMOUS`. Perhaps `INSECURE_DOTNET_DASHBOARD_DISABLE_AUTH`.

drewnoakes (approved):

JamesNK:
> There are security considerations in having single variable to opt out for all security features. If we add new kinds of security in future, should users that opted out before be automatically opted out of the new security too? The user may have reviewed the decision during installation. They may wish to review newer features as well.
> 
> If we are going to have a single variable, its name should evoke considerable caution. Something stronger than `DOTNET_DASHBOARD_ALLOW_ANONYMOUS`. Perhaps `INSECURE_DOTNET_DASHBOARD_DISABLE_AUTH`.

It's not opting out of all security features, just dashboard endpoint authentication. For example, it wouldn't disable configured HTTP security between the dashboard and resource service.

What about `DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS`? That keeps DOTNET at the front, similar to other .NET environment variables. Kind of long though.

Issue #1603, opened 2024-01-08
Create threat model for app model


adityamandaleeka:
@mitchdenny 

Relates to #237

- [ ] Enable ability to globally override image registry
    - https://github.com/dotnet/aspire/pull/3037
- [ ] Can we produce an SBOM with a closure of our dependencies (container)
- [x] Increase entropy on password generation beyond 128 bits.
    - https://github.com/dotnet/aspire/pull/2717
    - https://github.com/dotnet/aspire/pull/2740
- [ ] Ensure that port conflicts on container and process launch results in failure from DCP that we must respond to (port squatting)
- [ ] Investigating securing comms between AppHost to DCP to Docker/Podman
- [ ] Investigating securing comms between Dashboard to AppHost.

GrabYourPitchforks:
> Increase entropy on password generation beyond 128 bits.

BTW, here's how I performed the entropy calculation. The parameters _a_, _b_, _c_, _d_ are the arguments to this function:

https://github.com/dotnet/aspire/blob/813fc19c3acf1a644f23157f608a40a2f4ce829f/src/Aspire.Hosting/Utils/PasswordGenerator.cs#L29

```math
\log_2( \frac{23^a \cdot 23^b \cdot 10^c \cdot 11^d \cdot (a + b + c + d)!}{a! \cdot b! \cdot c! \cdot d!} )
```

With _a := 6_, _b := 6_, _c := 2_, _d := 2_, this shakes out at ~91 bits of entropy.

PR #3037, opened 2024-03-20
Hook which rewrites the container registry in annotations.

mitchdenny:
This PR introduces a `WithContainerRegistry` method that works on the `IDistributedApplicationBuilder` and registers a hook which overwrites the container registry on all the `ContainerImageAnnotations` in the application model.

Related to #1603 

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3037)


tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -81,6 +81,32 @@ public void EnsureWorkerProjectDoesNotGetBindingsGenerated()
...
+        var builder = DistributedApplication.CreateBuilder(GetManifestArgs()).WithContainerRegistry("myprivateregistry.company.com");
+        var redis = builder.AddRedis("redis");
+        builder.Build().Run();
+
+        var redisManifest = await ManifestUtils.GetManifest(redis.Resource);
+        var expectedManifest = """
+            {
+              "type": "container.v0",
+              "connectionString": "{redis.bindings.tcp.host}:{redis.bindings.tcp.port}",
+              "image": "myprivateregistry.company.com/redis:7.2.4",
```
joperezr, line 96: NIT: I think that there was a recent change that made it such that you use a variable for the version instead of the version number so that it is easier to update the version in the future.

joperezr (approved):
LGTM, thanks @mitchdenny! 

FYI: @GrabYourPitchforks

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
joperezr: This is what I was thinking about: https://github.com/dotnet/aspire/blob/1ee0f5acec3e5a8aea4e626e7b8002e9cdff4518/src/Aspire.Hosting.Redis/RedisContainerImageTags.cs#L9

davidfowl:
I'm not sure we want to merge this as is, we need to discuss how this would be used practically.

mitchdenny:
I think that there are two schools of thought here about how this tooling should work:

1. Global override
2. Policy enforcement

## Global override
The global override option is what we have here. We call a single method, and before we spin up containers we override the container registry for each container we start. The underlying mechanics could change but the `WithContainerRegistry` call or something like it is what we'd end up with.

## Policy enforcement
Instead of overriding values we check to make sure that the container registry part of the container image annotation is set to a specific value, and if it isn't, blow up.

## Why global override is better?
I landed on global override. My reasoning is that as soon as you fall afoul of policy enforcement you are going to want to override the the image registry that is used. That means that for every resource that has a container image annotation you are going to need to use the `WithImageRegistry` method.

That seems pretty heavy handed but there is another more fatal problem with requiring people to do that. I would expect that in time we'll see higher level abstractions. For example `AddGitHubWebhookRouterReceiver` which might add multiple resources to the model. This wouldn't necessarily be something where the `WithImageRegistry` approach works. Rather than get all extension authors to consider this supply chain requirement, it is better to do this centrally.

## Problems with current implementation
I agree that we shouldn't merge this yet. It was initially just to facilitate conversation. I think that mutating the resources via a hook is the wrong approach. There are two places where we need to resolve the container registry. The first is in manifest writing, and the second is when we hand the resource over to DCP to spin up as a container. It should be easy enough to intercept these two locations and do the substitution with a global override.

I think that the global override probably needs to be something that we can express in code (like we have in this PR) but perhaps also something that could be set via an environment variable.

afscrome:
What about prefixing the container repository as well as changing the registry.  e.g. if your company blocks direct access to `docker.io` and proxies you through an ECR pull through cache, you may want to rewrite `docker.io/*` --> `ACCOUNT.dkr.ecr.REGION.amazonaws.com/docker.io/*` - changing both the registry, as well as a prefix to the image repository.

joperezr:
@mitchdenny @davidfowl what is the state of this PR? This is one of the follow-up items that we are tracking and need to get done by P6, so ideally we'd want to get this in now.

mitchdenny:
@joperezr I just made a change to make the API something that I am happy with. I changed from using an extension method on the builder (which would be the only one of its kind) to using an option which I think gives it the appropriate level of visibility.

It is still implemented as a hook. After thinking about alternative approaches, I think hooks here are a better option.

PR #3038, opened 2024-03-20
Set container host based on resolved container runtime

### LINKED ISSUE #2639: Resolve the correct container runtime and chose appropriate host name
When I tried to run `AddPostgres` with `WithPgAdmin` as well as `AddRedis` with `WithRedisCommander`, nothing worked on Podman native.

I struggled to find the root cause and finally found the problem when I ran:

```sh
> podman ps
> podman exec -it <pg admin container id> sh
> cat servers.json
{"Servers":{"1":{"Name":"postgreSQL","Group":"Aspire instances","Host":"host.docker.internal","Port":5432,"Username":"postgres","SSLMode":"prefer","MaintenanceDB":"postgres","PasswordExecCommand":"echo \u0027<password>\u0027"}}} 
```

Look at the `Host`, it still pointed to `host.docker.internal` although I run on `Podman`. And the same thing happened to `Redis` with `RedisCommander` And I think this is a bug.

OS: Windows 11
.NET 8.0.200
.NET Aspire version: `8.0.0-preview.4.24127.7+2cbacd4bf15147d0ce67e915f6932bffb3f7e720`
Podman version: `4.9.0`

/cc @davidfowl

davidfowl:
- Use info from dcp to determine the container host name. This is host.docker.internal for host.containers.internal for podman. It can be overridden by configuration as well.
- Introduced HostUrl a way to tell the system to rewrite localhost urls (or urls on the host) to use the container host name (which is usually unknown to the resource author).
- Default AllocatedEndpoint's container host name to null.
- Updated tests to verify multiple host names.

Fixes #2639
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3038)


mitchdenny (approved):
LGTM

src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs
```diff
@@ -43,7 +43,7 @@ public AllocatedEndpoint(EndpointAnnotation endpoint, string address, int port,
     /// <summary>
     /// The address of the container host.
```
drewnoakes, line 44: Doc what null means?

src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs
```diff
@@ -43,7 +43,7 @@ public AllocatedEndpoint(EndpointAnnotation endpoint, string address, int port,
...
     /// </summary>
-    public string ContainerHostAddress { get; private set; }
+    public string? ContainerHostAddress { get; private set; }
```
drewnoakes, line 46: Doc what null means?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -89,9 +90,16 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
     private readonly ConcurrentDictionary<(string, string), List<string>> _resourceAssociatedServicesMap = [];
     private readonly ConcurrentDictionary<string, CancellationTokenSource> _logStreams = new();
     private readonly ConcurrentDictionary<IResource, bool> _hiddenResources = new();
+    private DcpInfo? _dcpInfo;
+
+    private string DefaultContainerHostName => configuration["AppHost:ContainerHostname"] ?? _dcpInfo?.Containers?.ContainerHostName ?? "host.docker.internal";
 
     public async Task RunApplicationAsync(CancellationToken cancellationToken = default)
     {
+        _dcpInfo = await dcpDependencyCheckService.GetDcpInfoAsync(cancellationToken).ConfigureAwait(false);
```
karolz-ms, line 99: Put that code after DcpModelCreationStart() please so that the cost to get DCP info is included.

src/Aspire.Hosting/Dcp/IDcpDependencyCheckService.cs
```diff
@@ -35,4 +35,11 @@ internal sealed class DcpContainersInfo
 
     [JsonPropertyName("error")]
     public string? Error { get; set; }
+
+    internal string ContainerHostName =>
+        Runtime switch
+        {
+            "podman" => "host.containers.internal",
```
karolz-ms, line 42: @danegsta what do you think about making container host part of the DCP container runtime info?

karolz-ms (approved):
Just one suggestion, otherwise LGTM!

src/Aspire.Hosting/Dcp/IDcpDependencyCheckService.cs
danegsta: We should be returning the identified runtime as part of `dcp info` already.

danegsta: I like that idea

davidfowl: @danegsta that issue is assigned to you already 😄

danegsta: DCP change to expose the address should be in PR in a few minutes

PR #3045, opened 2024-03-20
Support configuring dashboard with one endpoint for browser and OTLP

### LINKED ISSUE #2756: Aspire Dashboard HTTPS and SINGLE PORT hosting of both OTLP and Blazor app
I have successfully modified the project to host both OTLP and the Blazor app on a single port.  This allowed hosting as an Azure App Service with HTTPS used for both the received telemetry and the blazor dashboard app.

@DamianEdwards Is there a reason we cannot make this the default - only requiring a single port for the dashboard project? Or at least could we provide an option to start it up in this way to ensure we can host the dashboard with a single HTTPS port that handles both the OTLP inbound telemetry plus the blazor app.

Note: To deploy such a solution, you need to first follow the steps here to enable GRPC on the Azure App Service for Windows.  This is a preview feature:
https://github.com/Azure/app-service-linux-docs/tree/master/HowTo/gRPC/Windows

Note that I removed the "throw" inside the OtlpInterceptor.cs.

I started Kestrel up like this:
```cs
        builder.WebHost.ConfigureKestrel(kestrelOptions =>
        {
            kestrelOptions.ListenLocalhost(8080, options =>
            {
                if (builder.Environment.IsDevelopment()) options.UseHttps();
                options.Protocols = HttpProtocols.Http2;
                options.UseOtlpConnection();
            });
        });
```
I only started it using HTTPS in DEV mode, because I used the Azure Web App TLS termination in production.

I am not sure if this was needed, but I also added this to the appSettings.json:
```json
  "Kestrel": {
    "EndpointDefaults": {
      "Protocols": "Http2"
    }
  }
```

In summary, when deployed to an Azure App Service, this allows everything to run under TLS and only needs a single port.

Note: I did attempt to get this running on the Linux App Service, but hit a few issues and did not get time to resolve.

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/2756

If the dashboard was started with the same port for OTLP and browser endpoints, two endpoints were started and the dashboard fails because of a port conflict.

This PR changes the dashboard:
* Check if configured endpoints are the same, and if they are then only one endpoint is started. This allows the dashboard to be hosted in a single port environment such as Windows Azure App Service - https://github.com/Azure/app-service-linux-docs/blob/master/HowTo/gRPC/Windows/README.md
* Checks single endpoint for HTTPS. If it isn't secured with HTTPS, then a warning is logged that browser access is only possible via a proxy that terminates TSL.

TODO: Update dashboard readme.

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -209,24 +209,34 @@ public class DashboardWebApplication : IAsyncDisposable
     // possible from the caller. e.g., using environment variables to configure each endpoint's TLS certificate.
     private void ConfigureKestrelEndpoints(WebApplicationBuilder builder, DashboardStartupConfiguration dashboardStartupConfig)
     {
-        // Translate high-level config settings such as DOTNET_DASHBOARD_OTLP_ENDPOINT_URL and ASPNETCORE_URLS
-        // to Kestrel's schema for loading endpoints from configuration.
-        var browserEndpointNames = new List<string>(capacity: dashboardStartupConfig.BrowserUris.Length);
+        var singleEndpoint = dashboardStartupConfig.BrowserUris.Length == 1 && dashboardStartupConfig.BrowserUris[0] == dashboardStartupConfig.OtlpUri;
+
         var initialValues = new Dictionary<string, string?>();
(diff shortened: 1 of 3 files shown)


davidfowl:
Doe the auth code need a fixup?

JamesNK:
I don't think so. You could configure something that isn't useful by having a single endpoint that requires client certificate auth - which isn't possible from a browser.

davidfowl (approved):

Issue #2819, opened 2024-03-12
Remove dependencies on unstable OpenTelemetry packages in Aspire Components


eerhardt:
As our plan is to stabilize .NET Aspire for `v8.0.0` we need to remove unstable dependencies.

We have the following dependencies on unstable OpenTelemetry instrumentation packages in our Aspire Components:

* OpenTelemetry.Instrumentation.EventCounters
* OpenTelemetry.Instrumentation.SqlClient
* OpenTelemetry.Instrumentation.StackExchangeRedis

None of these packages has a plan to ship a stable/GA version by the time we ship `v8.0.0` of .NET Aspire. We need to formulate a plan for how to handle these dependencies.

### OpenTelemetry.Instrumentation.EventCounters

Used by all 5 EntityFramework components and Aspire.Microsoft.Data.SqlClient to turn the existing EventCounters into OTel metrics / Meters.

I think the following 2 options are possible here:

1. Remove these Metrics for now, and ask the lower level libraries (like efcore, sqlclient, etc) to produce metrics natively using the `Meter` APIs.
2. "Vendor"/copy the `OpenTelemetry.Instrumentation.EventCounters` into our Aspire Components for now until either: 
    a. The EventCounters library ships stable.
    b. The underlying libraries get native metrics using the `Meter` APIs.

### OpenTelemetry.Instrumentation.SqlClient

This is used by `Aspire.Microsoft.Data.SqlClient` (and after #2814 `Aspire.Microsoft.EntityFrameworkCore.SqlServer`) to produce traces / Activities for the SqlClient ADO.NET provider.

Tracing is more critical than Metrics, so I don't think it is a valid option to remove this instrumentation from the component. I think the only viable option here is to "vendor"/copy the instrumentation library code into the Aspire components.

### OpenTelemetry.Instrumentation.StackExchangeRedis

This is used by the 3 Redis components to produce traces / Activities when communicating with a Redis server.

For the same reason above as SqlClient, I don't think it is a valid option to remove this instrumentation from the component. I think the only viable option here is to "vendor"/copy the instrumentation library code into the Aspire components.

cc @joperezr @DamianEdwards @AndriySvyryd @roji @mgravell

samsp-msft:
Another possible option for EventCounters is to use the DiagnosticSource hooks that is being used for tracing to collect statistics and fabricate metrics based on those. The metrics can then be designed to match the OTel semantic conventions with dimensions etc.
[...]

PR #3059, opened 2024-03-21
Remove dependency from OpenTelemetry.Instrumentation.SqlClient

sebastienros:
Used two a separate forlder for `/Shared` since I expect the other OTel libraries we need to include will need this too.

See https://github.com/dotnet/aspire/issues/2819
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3059)


src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/GlobalSuppressions.cs
```diff
@@ -0,0 +1,11 @@
+// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style", "IDE0028:Simplify collection initialization", Justification = "Vendor code", Scope = "namespaceanddescendants", Target = "~N:OpenTelemetry.Instrumentation.SqlClient")]
```
eerhardt, line 8: Can we just use .editorconfig for this?

THIRD-PARTY-NOTICES.TXT
```diff
@@ -57,3 +57,20 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
+
+License notice for OpenTelemetry .NET (https://github.com/open-telemetry/opentelemetry-dotnet)
```
eerhardt, line 61: note that the next one (StackExchange.Redis) comes from a different repo: https://github.com/open-telemetry/opentelemetry-dotnet-contrib. We may need to update this when we pull that code in.

src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs
```diff
@@ -0,0 +1,303 @@
...
+
+namespace OpenTelemetry.Instrumentation.SqlClient;
+
+/// <summary>
+/// Options for <see cref="SqlClientInstrumentation"/>.
+/// </summary>
+/// <remarks>
+/// For help and examples see: <a href="https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.SqlClient/README.md#advanced-configuration" />.
+/// </remarks>
+public class SqlClientTraceInstrumentationOptions
```
eerhardt, line 19: These need to be `internal`. We don't want to ship these as public from our library.

eerhardt:
@reyang - Can you review the attribution to ensure we aren't violating the license here?

THIRD-PARTY-NOTICES.TXT
reyang:
```suggestion
License notice for OpenTelemetry
```

THIRD-PARTY-NOTICES.TXT
```diff
@@ -57,3 +57,20 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
+----------------------------------------------------------------------------------------------
+
+Copyright 2023 OpenTelemetry Authors
```
reyang, line 64: This is what OpenTelemetry uses https://github.com/open-telemetry/opentelemetry-dotnet/blob/af57ae87871aac7e201a03255ac8c81e5747bc07/src/OpenTelemetry.Api/Baggage.cs#L1-L2

```suggestion
Copyright The OpenTelemetry Authors
```

src/Vendoring/README.md
```diff
@@ -0,0 +1,32 @@
+# Vendoring code sync instructions
```
reyang, line 1: @rajkumar-rangaraj FYI

reyang:
> @reyang - Can you review the attribution to ensure we aren't violating the license here?

I've added couple suggestions, it looks good to me once these are addressed 👍

joperezr (commented):
So we decided to go with building their code ourselves as opposed to re-disting their libraries inside our package?

This makes me nervous as we won't be running their tests in this code, nor would we be applying code fixes, and we don't really know how long are we going to be doing this.

Did we explore the option of just harvesting their library and including it in our package instead?

cc @eerhardt

src/Vendoring/README.md
```diff
@@ -0,0 +1,32 @@
...
+### Instructions
+
+- Copy files from `src/OpenTelemetry.Instrumentation.SqlClient`:
+    - `OpenTelemetry.Instrumentation.SqlClient.csproj`
+    - `**\*.cs`
+
+### Customizations
+
+- Added `#nullable disable` in files that require it.
+- Added `GlobalSuppressions.cs` to fix incompatible coding style.
```
joperezr, line 32: We should also add here that all public types need to be made internal. As @eerhardt pointed out we need to make sure we are not exposing any of this as public API.

src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+using OpenTelemetry.Instrumentation.SqlClient;
+using OpenTelemetry.Instrumentation.SqlClient.Implementation;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.Trace;
+
+/// <summary>
+/// Extension methods to simplify registering of dependency instrumentation.
+/// </summary>
+public static class TracerProviderBuilderExtensions
```
joperezr, line 19: Needs to be internal

src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs
reyang: One suggestion for future releases, consider leveraging the public API analyzer in CI.

* https://github.com/open-telemetry/opentelemetry-dotnet/pull/806
* https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/.publicApi/Stable/PublicAPI.Shipped.txt

src/Vendoring/README.md
reyang: https://github.com/dotnet/aspire/pull/3059#discussion_r1534387563

src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/GlobalSuppressions.cs
sebastienros: These are specific to the files from OTel, but I will check if we can set rules for specific paths.

THIRD-PARTY-NOTICES.TXT
sebastienros: Is it worth filing an issue in Azure SDK? I took it from there: https://github.com/Azure/azure-sdk-for-net/blob/main/NOTICE.txt#L232-L235

reyang: > Is it worth filing an issue in Azure SDK? I took it from there: https://github.com/Azure/azure-sdk-for-net/blob/main/NOTICE.txt#L232-L235

@rajkumar-rangaraj could you follow up and handle this?

rajkumar-rangaraj: I will get this handled in Azure SDK. 
When we earlier vendored in Azure SDK, OpenTelemetry had that license format.

It's worth to update the Readme to check on the license formats too.

src/Vendoring/OpenTelemetry.Instrumentation.SqlClient/GlobalSuppressions.cs
sebastienros: Done

src/Vendoring/README.md
sebastienros: Done

eerhardt:
> So we decided to go with building their code ourselves as opposed to re-disting their libraries inside our package?
> 
> This makes me nervous as we won't be running their tests in this code, nor would we be applying code fixes, and we don't really know how long are we going to be doing this.
> 
> Did we explore the option of just harvesting their library and including it in our package instead?
> 
> cc @eerhardt

Redisting libraries in a nuget package isn't without drawbacks either. If the app brings in the OpenTelemetry.Instrumentation.SqlClient package through another dependency, there will be conflicts since 2 nuget packages brought in the same assembly. We would also need to sign it, etc, in order to redist it.

> and we don't really know how long are we going to be doing this.

My understanding from @reyang is that the SqlClient library is targeted to ship stable later this calendar year. But I assume those plans can change. The other one we will need to give the same treatment is the StackExchange.Redis OTel library, which doesn't have a plan on going stable AFAIK.

src/Components/Aspire.Microsoft.Data.SqlClient/Aspire.Microsoft.Data.SqlClient.csproj
```diff
@@ -15,15 +15,19 @@
...
+    <Compile Include="$(VendoringDir)OpenTelemetry.Shared\**\*.cs" LinkBase="OpenTelemetry.Instrumentation.SqlClient.Shared" />
+  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="AspNetCore.HealthChecks.SqlServer" />
-    <PackageReference Include="Microsoft.Data.SqlClient" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" />
```
eerhardt, line 28: What's this?

src/Components/Aspire.Microsoft.Data.SqlClient/Aspire.Microsoft.Data.SqlClient.csproj
```diff
@@ -15,15 +15,19 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="$(VendoringDir)OpenTelemetry.Instrumentation.SqlClient\**\*.cs" LinkBase="OpenTelemetry.Instrumentation.SqlClient" />
+    <Compile Include="$(VendoringDir)OpenTelemetry.Shared\**\*.cs" LinkBase="OpenTelemetry.Instrumentation.SqlClient.Shared" />
+  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="AspNetCore.HealthChecks.SqlServer" />
-    <PackageReference Include="Microsoft.Data.SqlClient" />
```
eerhardt, line 20: We want this, so our dependency on Microsoft.Data.SqlClient gets our version from the Directory.Packages.props.

sebastienros: Was not even on purpose, I don't know how I removed it. Maybe mistaken for the telemetry one.

src/Components/Aspire.Microsoft.Data.SqlClient/Aspire.Microsoft.Data.SqlClient.csproj
sebastienros: I think when I imported some source this was required for a specific type. Now not anymore so I assume that I removed a file that was the culprit. Removed.

eerhardt (approved):
LGTM. Thanks!

reyang (approved):

PR #3070, opened 2024-03-21
Use the right noun and remove Owner

davidfowl:
- Owner.Owner looks weird. EndpointReferenceExpression has an Endpoint which is an EndpointReference. EndpointReference have an associated resource.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3070)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
@@ -15,7 +15,7 @@ public sealed class EndpointReference : IManifestExpressionProvider, IValueProvi
     private bool? _isAllocated;
 
     /// <summary>
-    /// Gets the owner of the endpoint reference.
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
```diff
@@ -17,7 +17,7 @@ public sealed class EndpointReference : IManifestExpressionProvider, IValueProvi
     /// <summary>
     /// Gets the owner of the endpoint reference.
```
JamesNK, line 18:
```suggestion
    /// Gets the resource owner of the endpoint reference.
```

JamesNK (approved):

mitchdenny (approved):

PR #3075, opened 2024-03-21
Refactor console logs to render on server and use virtualization

### LINKED ISSUE #2507: Dashboard console logs - High CPU usage with large amount of logs
1. Start TestShop
2. Configure Aspire dashboard to log everything with a debug level
3. View Aspire dashboard console logs. There are 1000s of lines.
4. Scrolling up and down isn't smooth. CPU usage for the tab is constant 20% (max?)

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/2507

* Move sorting and rendering to the server
* Enable virtualization

TODO:
* ~Line numbers need work. Especially if the total logs are truncated. For example, if logs are truncated from 20,000 to 10,000, the first visible line should have a line number of 10,001.~
* ~Some old JS needs to be removed~
* Haven't tested whether line wrapping is a problem with virtualization yet. If it is then we'll disable line wrapping, like what AzDo does.
* Selecting text to copy doesn't work well with virtualization. AzDO also has this limitation. Perhaps add a "Raw logs" button to view raw text and people can copy from that. Most console log UIs I've seen offer this kind of feature.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3075)


JamesNK:
@tlmii I haven't looked at console logs closely before. Why does the current source code sort log entries and build HTML in the client?

tlmii:
> @tlmii I haven't looked at console logs closely before. Why does the current source code sort log entries and build HTML in the client?

Short answer: This was one of the first things we implemented... it started off all in Blazor (sans virtualization - I didn't know about the Virtualize component (not to mention a lot of other Blazor stuff at the time)), someone commented that that'd be too much for Blazor to push to the client as the dataset got larger and we should only send changes to the client manually. So we did. And it hasn't been a priority to fix since then.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
```diff
@@ -46,31 +58,76 @@ await foreach (var batch in batches.WithCancellation(cancellationToken))
...
         }
     }
 
-    internal async Task ClearLogsAsync(CancellationToken cancellationToken = default)
+    private static void InsertSorted(List<LogEntry> logEntries, LogEntry logEntry)
     {
-        await _cancellationSeries.ClearAsync();
+        if (logEntry.ParentId != null)
+        {
+            // If we have a parent id, then we know we're on a non-timestamped line that is part
```
tlmii, line 74: Was hoping the need to do this interleaving would go away at some point. I guess that some point is not yet...

src/Aspire.Dashboard/Components/Controls/LogViewer.razor
```diff
@@ -6,7 +6,27 @@
...
 @implements IAsyncDisposable
 
-<div class="log-overflow">
+<div class="log-overflow continuous-scroll-overflow">
     <div class="log-container" id="logContainer">
+        <Virtualize Items="_logEntries" ItemSize="20" OverscanCount="20">
+            <div class="line-row-container">
+                <div class="line-row">
+                    <span class="line-area" role="log">
+                        <span class="line-number">@(_logEntries.IndexOf(context) + 1)</span>
```
tlmii, line 15: This is going to end up scanning the whole log entry any time you're scrolled to the bottom, isn't it? This is probably part of what you were referencing in your description. Wonder if we could track the index on the item itself. Would require changing ones that are already in the list as we go past them to insert another, but that's not _too_ hard.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor
```diff
@@ -6,7 +6,27 @@
 @inject IJSRuntime JS
 @implements IAsyncDisposable
 
-<div class="log-overflow">
+<div class="log-overflow continuous-scroll-overflow">
     <div class="log-container" id="logContainer">
+        <Virtualize Items="_logEntries" ItemSize="20" OverscanCount="20">
```
tlmii, line 11: Since the entries being added are fairly lightweight, we could probably go with an even higher OverscanCount for this to help with what I expect will be commonly occurring rapid scrolling. Not sure how to determine what would be a good number though. One full extra page on a reasonable size monitor?

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
```diff
@@ -79,10 +79,8 @@
     align-self: flex-start;
     flex-shrink: 0;
     color: var(--line-number-color);
-}
-
-::deep .line-number::before {
-    content: counter(line-number);
```
tlmii, line 85: Can remove the `counter-reset` and `counter-increment` lines above as well

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
JamesNK: I just replicated what JS was doing. I'm not a fan of it. Reordering data from the server just before it is rendered is probably the wrong place. If reordering is required, it should happen before storing the data inside the host.

Can you talk more about why it is required? What was the original issue?

src/Aspire.Dashboard/Components/Controls/LogViewer.razor
JamesNK: Yes, this is a problem for lots of reasons. Still TODO.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
tlmii: Yeah, sorry, I probably could have been more clear on that comment. I had hoped it was going to go away with streaming logs, but @karolz-ms indicated there were still two separate streams so there might be an issue.

The reordering was required because the logs arrived out of order from the two different streams (stdout and stderr).  This was especially true with logs being written to the file system - the first stderr line could have been seconds or minutes after the first stdout line, but since the two files were read at the same time, we'd get them in unpredictable order. Though whatever the source is (docker logs acted slightly differently) they'd still need to be interleaved at some point.

Looking through the current code on the dashboard side, I don't see anywhere obvious where we see the two different streams anymore. Maybe @karolz-ms can speak to how that works in the streaming world now?

JamesNK: There are still two streams, they now come from k8s service:

https://github.com/dotnet/aspire/blob/0783a609d369634c1a0aa5d399b78c79a52caafe/src/Aspire.Hosting/Dcp/ResourceLogSource.cs#L28-L29

I think this would still be a problem.

JamesNK:
@davidfowl I modified the resource logger API to have a struct record to represent line information. It now has a line number on it.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
karolz-ms: DCP could combine the streams (and apply timestamps to Executable logs), but then we need to figure out how to convey what entries come from stdout vs stderr. E.g. change the format to JSON lines or something similar, but then there is a question of how to deal with non-printable characters etc.

JamesNK: If DCP combined the streams, would they be in the correct order? How does DCP get its data?

JamesNK: > E.g. change the format to JSON lines or something similar, but then there is a question of how to deal with non-printable characters etc.

I guess a JSON serializer would escape them to /u342345 format.

tlmii:
When the page is scrolled to the bottom, there's a bit of a flicker every time the content is updated (the gif doesn't catch all the flickers):

ConsoleLogsFlicker

I switched back to main branch and that doesn't seem to be happening there. Just a function of the way the Virtualize works? I don't think its a blocker since this PR makes the page usable with lots of content which is much more important. But it is kind of annoying when watching the logs come in waiting for something.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
```diff
@@ -34,21 +34,19 @@
     background: var(--console-background-color);
     color: var(--console-font-color);
     font-family: 'Cascadia Mono', Consolas, monospace;
-    font-size: 12px;
+    font-size: 13px;
```
tlmii, line 37: I'm fine with changing it to 13px, just noting that 12px is what AzDO and GH Actions both use so that's why it was picked.

src/Aspire.Dashboard/Model/ResourceLogLine.cs
```diff
@@ -0,0 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Dashboard.Model;
+
+public readonly record struct ResourceLogLine(int LineNumber, string Content, bool IsErrorMessage)
```
tlmii, line 6: Should this have the same doc comments as the LogLine in Hosting?

src/Aspire.Dashboard/ConsoleLogs/TimestampParser.cs
```diff
@@ -12,7 +13,7 @@ public static partial class TimestampParser
 {
     private static readonly Regex s_rfc3339RegEx = GenerateRfc3339RegEx();
 
-    public static bool TryColorizeTimestamp(BrowserTimeProvider timeProvider, string text, bool convertTimestampsFromUtc, out TimestampParserResult result)
+    public static bool TryColorizeTimestamp(string text, [NotNullWhen(true)] out TimestampParserResult? result)
```
tlmii, line 16: Should this be renamed now since its just finding it, not colorizing it anymore?

src/Aspire.Dashboard/Model/ResourceLogLine.cs
JamesNK: LogLine only has public comments because it's a public API.

This one is just for us. The names are self explanatory.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
JamesNK: It felt small to me compared to other text. And there was feedback in the line number color issue that text was small.

I tried 14px and it was too big. 13px felt good. IMO stick with it, the the app out, and see if we want to change its size again.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
```diff
@@ -46,33 +54,103 @@ await foreach (var batch in batches.WithCancellation(cancellationToken))
                 continue;
             }
 
-            List<LogEntry> entries = new(batch.Count);
-
-            foreach (var (content, isErrorOutput) in batch)
+            foreach (var (lineNumber, content, isErrorOutput) in batch)
             {
-                entries.Add(logParser.CreateLogEntry(content, isErrorOutput));
+                if (_baseLineNumber is null)
```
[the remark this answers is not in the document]
JamesNK: Yes, it's required when there are more than 10,000 lines. That is how much hosting buffers.

I'll add comments.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
karolz-ms: For Containers the logs arrive in the order the container orchestrator is providing them. But since we are using the CLI to talk to the orchestrator, it boils down to reading `stdout` and `stderr`. 

The same applies to Executables. When DCP launches Executables directly, it is also responsible for reading `stdout` and `stderr` streams. When using the IDE to launch Executables (i.e. dealing with VS projects), it is relying on the IDE to provide the logs. 

Bottom line this is probably sufficient for UI to have logs emitted ROUGHLY at the same time to surface close in the combined stream. But there will definitively be some jitter involved, and if the expectation is that developers will be able to infer the order in which things happened based on the order in which they appeared in the (combined) log stream, especially if we are talking about microseconds delay between writes to `stdout` vs `stderr`, that is probably a stretch.

JamesNK:
> When the page is scrolled to the bottom, there's a bit of a flicker every time the content is updated (the gif doesn't catch all the flickers):
> 
> ConsoleLogsFlicker [ ConsoleLogsFlicker ](https://private-user-images.githubusercontent.com/9613109/315889006-7b818e8b-850b-4067-956e-e9c63322624e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTEzNzQxMDQsIm5iZiI6MTcxMTM3MzgwNCwicGF0aCI6Ii85NjEzMTA5LzMxNTg4OTAwNi03YjgxOGU4Yi04NTBiLTQwNjctOTU2ZS1lOWM2MzMyMjYyNGUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDMyNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAzMjVUMTMzNjQ0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MGQwMDYwZmIwYjY4YzIxMzRmZjgxMGY3OTg0OWJiZWRhYWUzZWU1YjhkMjc5YWYyM2RhNDBhOTQ0ZjQ3MTI1ZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.ulf05MNiAhkCX9C9QY8UgwoTaoIqQm4nzVXaRT606Yo) [ ](https://private-user-images.githubusercontent.com/9613109/315889006-7b818e8b-850b-4067-956e-e9c63322624e.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTEzNzQxMDQsIm5iZiI6MTcxMTM3MzgwNCwicGF0aCI6Ii85NjEzMTA5LzMxNTg4OTAwNi03YjgxOGU4Yi04NTBiLTQwNjctOTU2ZS1lOWM2MzMyMjYyNGUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDMyNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAzMjVUMTMzNjQ0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MGQwMDYwZmIwYjY4YzIxMzRmZjgxMGY3OTg0OWJiZWRhYWUzZWU1YjhkMjc5YWYyM2RhNDBhOTQ0ZjQ3MTI1ZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.ulf05MNiAhkCX9C9QY8UgwoTaoIqQm4nzVXaRT606Yo)
> 
> I switched back to main branch and that doesn't seem to be happening there. Just a function of the way the Virtualize works? I don't think its a blocker since this PR makes the page usable with lots of content which is much more important. But it is kind of annoying when watching the logs come in waiting for something.

I think it's a problem with `Virtualize`. When new content is added to Virtualize, it adds content to the end, but doesn't move the scroll position with the content. That means new data updates the console and increases the scroll bar size, then our JS sets the scroll position to the end, which then triggers `Virtualize` to load the end data. There are a bunch of steps instead of new data being loaded once and displayed.

I created an issue about it: https://github.com/dotnet/aspnetcore/issues/54742

I don't think there is a way to improve it (short of copying the source code of Virtualize and changing how it works) right now, but hopefully it can be improved in the next .NET.

cc @SteveSandersonMS @mkArtakMSFT

JamesNK:
Review please 🙏

tlmii (approved):
One thing I'm seeing fairly consistently here is that if I launch TestShop, let the Resources page load, go to Console Logs, and choose the dashboard entry from the list, the page doesn't scroll with the logs. It seems like it just stays at wherever it is when the page loads (several pages of logs down from the top, but I guess not all the way at the bottom since it doesn't auto-scroll). Sometimes if I leave the page and come back, it'll act as expected.

That particular log source is the one that is already more than a page of data by the time you get to it and still continues adding logs, so its hard to repro it for other resources.

I think the rest of this looks good other than this and the flicker (which I know is not from our code) so I'll go ahead and approve it. Workaround is easy enough - just use keyboard or mouse to scroll to the bottom.

JamesNK:
I've noticed it not scrolling with new content as the page loads as well. I'll see if I can figure it out.

JamesNK:
> I've noticed it not scrolling with new content as the page loads as well. I'll see if I can figure it out.

I've fixed it.

PR #3080, opened 2024-03-21
Use application name in implicit volume names

### LINKED ISSUE #3058: Implicit volume names should include the application name to make conflicts between apps less likely
The new `WithDataVolume` methods added on various container-based resources will implicitly name the volume after the resource name if a volume name is not supplied. The implicit names should include the AppHost application name as a prefix to reduce the chance of conflicts between resources of different apps.

DamianEdwards:
Implicit volume names are now in the form `$"{applicationName}-{resourceName}-{suffix}"`, unless the application name contains invalid chars for a volume name, in which case it's `$"volume-{resourceName}-{suffix}"`.

Fixes #3058
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3080)

### DIFF
--- a/eng/testing/xunit.runner.json
+++ b/eng/testing/xunit.runner.json
@@ -1,5 +1,4 @@
 {
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
-  "longRunningTestSeconds": 120,
-  "parallelizeAssembly": false
+  "longRunningTestSeconds": 120
 }
--- a/eng/testing/xunit.runner.json
+++ b/eng/testing/xunit.runner.json
@@ -1,4 +1,5 @@
 {
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
-  "longRunningTestSeconds": 120
+  "longRunningTestSeconds": 120,
+  "parallelizeAssembly": false
 }
--- a/src/Shared/VolumeNameGenerator.cs
+++ b/src/Shared/VolumeNameGenerator.cs
@@ -1,0 +1,52 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting.Utils;
+
+internal static class VolumeNameGenerator
+{
+    /// <summary>
+    /// Creates a volume name with the form <c>$"{applicationName}-{resourceName}-{suffix}</c>, e.g. <c>"myapplication-postgres-data"</c>.
+    /// </summary>
+    /// <remarks>
+    /// If the application name contains chars that are invalid for a volume name, the prefix <c>"volume-"</c> will be used instead.
+    /// </remarks>
+    public static string CreateVolumeName<T>(IResourceBuilder<T> builder, string suffix) where T : IResource
+    {
+        if (!HasOnlyValidChars(suffix))
+        {
+            throw new ArgumentException($"The suffix '{suffix}' contains invalid characters. Only [a-zA-Z0-9_.-] are allowed.", nameof(suffix));
+        }
+
+        // Create volume name like "myapplication-postgres-data"
+        var applicationName = builder.ApplicationBuilder.Environment.ApplicationName;
+        var resourceName = builder.Resource.Name;
+        return $"{(HasOnlyValidChars(applicationName) ? applicationName : "volume")}-{resourceName}-{suffix}";
+    }
+
+    private static bool HasOnlyValidChars(string name)
+    {
+        // According to the error message from docker CLI, volume names must be of form "[a-zA-Z0-9][a-zA-Z0-9_.-]"
+        var nameSpan = name.AsSpan();
+
+        for (var i = 0; i < nameSpan.Length; i++)
+        {
+            var c = nameSpan[i];
+
+            if (i == 0 && !(Char.IsAsciiLetter(c) || Char.IsNumber(c)))
+            {
+                // First char must be a letter or number
+                return false;
+            }
+            else if (!(Char.IsAsciiLetter(c) || Char.IsNumber(c) || c == '_' || c == '.' || c == '-'))
+            {
+                // Subsequent chars must be a letter, number, underscore, period, or hyphen
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/tests/Aspire.Hosting.Tests.SharedShim/Aspire.Hosting.Tests.SharedShim.csproj
+++ b/tests/Aspire.Hosting.Tests.SharedShim/Aspire.Hosting.Tests.SharedShim.csproj
@@ -1,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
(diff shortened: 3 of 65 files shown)


src/Shared/VolumeNameGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+
+internal static class VolumeNameGenerator
+{
+    /// <summary>
+    /// Creates a volume name with the form <c>$"{applicationName}-{resourceName}-{suffix}</c>, e.g. <c>"myapplication-postgres-data"</c>.
+    /// </summary>
+    /// <remarks>
+    /// If the application name contains chars that are invalid for a volume name, the prefix <c>"volume-"</c> will be used instead.
+    /// </remarks>
+    public static string CreateVolumeName<T>(IResourceBuilder<T> builder, string suffix) where T : IResource
```
radical, line 16: super duper nit: should this be `VolumeNameGenerator.Create<T>`?

DamianEdwards: Perhaps? We're not super consistent with naming these kinds of things from what I can see, e.g. `PathNormalizer.NormalizePathForCurrentPlatform` vs. `PasswordGenerator.Generate`.

eng/testing/xunit.runner.json
```diff
@@ -1,4 +1,5 @@
 {
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
-  "longRunningTestSeconds": 120
+  "longRunningTestSeconds": 120,
+  "parallelizeAssembly": false
```
joperezr, line 4: Was this to fix the AppHost issue?

joperezr: If so, is it not possible to only remove parallelization of tests within that assembly instead?

tests/testproject/TestProject.WorkerA/Program.cs
```diff
@@ -6,5 +6,5 @@
 var builder = Host.CreateApplicationBuilder(args);
 builder.Services.AddHostedService<Worker>();
 
-var host = builder.Build();
+using var host = builder.Build();
```
joperezr, line 9: I assume there must be an analyzer somewhere that flags cases where you are not disposing IDisposable types. We should consider enabling that to prevent these test issues to happen again in the future. I'll log an issue to track this.

src/Shared/VolumeNameGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        if (!HasOnlyValidChars(suffix))
+        {
+            throw new ArgumentException($"The suffix '{suffix}' contains invalid characters. Only [a-zA-Z0-9_.-] are allowed.", nameof(suffix));
+        }
+
+        // Create volume name like "myapplication-postgres-data"
+        var applicationName = builder.ApplicationBuilder.Environment.ApplicationName;
+        var resourceName = builder.Resource.Name;
+        return $"{(HasOnlyValidChars(applicationName) ? applicationName : "volume")}-{resourceName}-{suffix}";
```
radical, line 26: If the `applicationName` has invalid chars then should it be sanitized (like replace invalid ones with `_`) and used instead of defaulting to `volume`?
Does a user need to be able to reliably predict this name? Just wondering because here we silently switch to `"volume"` if `applicationName` has invalid chars.

joperezr (approved):
Not super familiar with this code, but changes LGTM.

tests/Aspire.Hosting.Tests.SharedShim/Aspire.Hosting.Tests.SharedShim.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
radical, line 4:
```suggestion
    <TargetFramework>$(NetCurrent)</TargetFramework>
```

eng/testing/xunit.runner.json
DamianEdwards: This was to fix the test failures due to hitting the io handle limit on Linux.

tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj
```diff
@@ -30,7 +30,7 @@
     <ProjectReference Include="..\..\src\Aspire.Hosting.Nats\Aspire.Hosting.Nats.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.PostgreSQL\Aspire.Hosting.PostgreSQL.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\testproject\TestProject.AppHost\TestProject.AppHost.csproj" IsAspireProjectResource="false" />
-
+    <ProjectReference Include="..\Aspire.Hosting.Tests.SharedShim\Aspire.Hosting.Tests.SharedShim.csproj" IsAspireProjectResource="false" Aliases="AspireHostingShared" />
```
radical, line 33: TIL `Aliases=..`, 👍

mitchdenny:
What happens if you have the repo checked out into two different directories. Because the volumes are persistent they'll conflict with one another. perhaps the volume name needs to be based on a hash of the apphost directory?

src/Shared/VolumeNameGenerator.cs
DamianEdwards: I started down the path of sanitization but ended up opting for this instead for simplicity sake. I think it'll be rare that an application name contains invalid chars so the extra complexity of char substitution (including dealing with what char to substitute if it's the first character that's invalid) didn't seem worth it.

DamianEdwards:
> What happens if you have the repo checked out into two different directories. Because the volumes are persistent they'll conflict with one another. perhaps the volume name needs to be based on a hash of the apphost directory?

That seems like a rare case that the user can handle themselves by specifying a custom volume name that handles the situation. I don't really want to make the default names more complex to handle what seems like a rare situation that the user can workaround themselves. Conflicts from resource names being the same between different apps seemed much more likely and worth solving.

eng/testing/xunit.runner.json
radical: We should set this only for `Aspire.Hosting`, so the other assemblies can still run in parallel. I'm assuming that we start off all the test assemblies in parallel and xunit decides how many to actually run concurrently. And this would be the case where we have a single step of `run non-helix steps`.

DamianEdwards: Thing is, this [apparently defaults to `false`](https://xunit.net/docs/configuration-files#parallelizeAssembly), so I'm not actually sure if this is what fixed it. I didn't realize the file was shared when I made the change (it's linked into the `Aspire.Hosting.Tests` project).

DamianEdwards: Happy to change it so that `Aspire.Hosting.Tests` has its own copy of this file and just have the change there.

radical: > defaults to false

nvm then. we can drop the change then, but if you want to keep it then separate copy would be great, IMHO.

radical (approved):

Issue #3077, opened 2024-03-21
Add Icons & Readmes to Aspire.Hosting.Azure.* packages


mitchdenny:
We don't currently have any.

davidfowl:
Fixed by https://github.com/dotnet/aspire/pull/3082

Issue #2955, opened 2024-03-16
Add README/icons to Aspire.Hosting.* and component packages


davidfowl:
We have them for Aspire components but not for the hosting packages.  Need icons + README.md for:

Hosting packages - README + Icon needed
- [ ] Aspire.Hosting.AWS
- [ ] Aspire.Hosting.Dapr
- [ ] Aspire.Hosting.Kafka
- [ ] Aspire.Hosting.MongoDB
- [ ] Aspire.Hosting.MySQL
- [ ] Aspire.Hosting.Nats
- [ ] Aspire.Hosting.NodeJs
- [ ] Aspire.Hosting.Oracle
- [ ] Aspire.Hosting.Orleans
- [ ] Aspire.Hosting.RabbitMQ

Hosting packages - REAMDE only needed
- [ ] Aspire.Hosting.Azure
- [ ] Aspire.Hosting.Azure.OperationalInsights
- [ ] Aspire.Hosting.PosgreSQL
- [x] Aspire.Hosting.Redis
- [ ] Aspire.Hosting.Seq
- [ ] Aspire.Hosting.SqlServer

Hosting packages - Icon only needed
- [ ] Aspire.Hosting.Azure.KeyVault

Components - Icon only needed
- [ ] Aspire.Confluent.Kafka
- [ ] Aspire.MongoDB.Driver
- [ ] Aspire.Oracle.EntityFrameworkCore
- [ ] Aspire.RabbitMQ.Client

PR #3082, opened 2024-03-21
Icons for Aspire.Hosting.Azure packages

timheuer:
Part 1: Icons
Using Azure Icons from Brand Central
#3077
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3082)


timheuer:
Took a stab at working this for the refactoring @mitchdenny

src/Aspire.Hosting.Azure.CognitiveServices/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+var myService = builder.AddProject<Projects.MyService>()
+                       .WithReference(openai);
+```
+
+The `AddAzureOpenAI` method will read connection information from the AppHost's configuration (for example, from "user secrets") under the `ConnectionStrings:openai` config key. The `WithReference` method passes that connection information into a connection string named `openai` in the `MyService` project. In the _Program.cs_ file of `MyService`, the connection can be consumed using:
+
+```csharp
+builder.AddAzureOpenAIClient("openai");
+```
```
eerhardt, line 35: One thing we do in the Components READMEs is to reference which `Aspire.Hosting.X` packages you need to reference to do the hosting side of the changes.

Maybe we could do the same thing here the other way?

src/Components/Aspire.Azure.AI.OpenAI/README.md
```diff
@@ -121,16 +121,16 @@ dotnet add package Aspire.Hosting.Azure
 Then, in the _Program.cs_ file of `AppHost`, add an Azure AI OpenAI service and consume the connection using the following methods:
```
eerhardt, line 121: @mitchdenny - it looks like all the Azure Components READMEs are now broken since they just say

````
In your AppHost project, install the Aspire Azure Hosting library with [NuGet](https://www.nuget.org):
```dotnetcli
dotnet add package Aspire.Hosting.Azure
```
````

These need to be updated now that they are split into different packages.

src/Aspire.Hosting.Azure.CognitiveServices/README.md
timheuer: Yeah good point -- i basically extracted the 'app host extension' sections from the component ones here to be consistent, but now that is lacking what about a simpler thing here and link to the components -- that way the better instructions are there with more context?

src/Components/Aspire.Azure.AI.OpenAI/README.md
timheuer: I can modify those in this PR as well.

src/Aspire.Hosting.Azure.CognitiveServices/README.md
timheuer: What do you think of this @eerhardt:
image

Rather than repeating the install instructions, etc. --  just link to that package

eerhardt: 👍 I like it.

davidfowl (approved):

davidfowl:
@timheuer are you going to take a stab at the other packages 😄 (I just realized that we need to do the same for all of the OSS stuff, redis etc)

mitchdenny:
Wonderful! Thanks!

timheuer:
> @timheuer are you going to take a stab at the other packages 😄 (I just realized that we need to do the same for all of the OSS stuff, redis etc)

Let me log an issue with remaining just as a punchlist so we don't miss any...

timheuer:
Let's track remaining as your prior https://github.com/dotnet/aspire/issues/2955 (I enumerated all the missing package assets)

PR #3090, opened 2024-03-22
Remove InputAnnotation

eerhardt:
Removes `"inputs"` on all resources. And instead only ParameterResources have "inputs" - the "value" input.

This means that password and user name parameters need to be "synthesized" when they aren't provided. To make this work, we create a new ParameterResource, but don't add it to the model. At manifest generation time, we add some infrastructure that ensures all referenced Resources are written to the manifest, even if they aren't explicitly in the DistributedApplicationModel list.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3090)


src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -22,20 +21,25 @@ public static class ParameterResourceBuilderExtensions
...
     {
-        return builder.AddParameter(name, () =>
-        {
-            var configurationKey = $"Parameters:{name}";
-            return builder.Configuration[configurationKey] ?? throw new DistributedApplicationException($"Parameter resource could not be used because configuration key '{configurationKey}' is missing.");
-        }, secret: secret);
+        return builder.AddParameter(name, parameterDefault => GetParameterValue(builder, name, parameterDefault), secret: secret);
+    }
+
+    private static string GetParameterValue(IDistributedApplicationBuilder builder, string name, ParameterDefault? parameterDefault)
```
davidfowl, line 27: nit: This should take IConfiguration

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -114,7 +104,62 @@ public static IResourceBuilder<T> PublishAsConnectionString<T>(this IResourceBui
...
+    /// </summary>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">Name of parameter resource</param>
+    /// <param name="secret">Flag indicating whether the parameter should be regarded as secret.</param>
+    /// <param name="parameterDefault">The <see cref="GenerateParameterDefault"/> that describes how the parameter's value should be generated.</param>
+    /// <returns>The created <see cref="ParameterResource"/>.</returns>
+    public static ParameterResource CreateGeneratedParameter(
+        IDistributedApplicationBuilder builder, string name, bool secret, GenerateParameterDefault parameterDefault)
+    {
+        var parameterResource = new ParameterResource(name, parameterDefault => GetParameterValue(builder, name, parameterDefault), secret);
```
davidfowl, line 159: nit: 
```suggestion
        var parameterResource = new ParameterResource(name, parameterDefault => GetParameterValue(builder.Configuration, name, parameterDefault), secret);
```

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -375,4 +501,44 @@ private void WriteContainerMounts(ContainerResource container)
...
+
+    private async Task WriteReferencedResources(DistributedApplicationModel model)
+    {
+        foreach (var resource in _referencedResources.Values)
+        {
+            if (!model.Resources.Contains(resource))
+            {
+                await WriteResourceAsync(resource).ConfigureAwait(false);
+            }
+        }
```
davidfowl, line 542: nit: It's more efficient to remove resources that are already in the model, then write the what's left over.

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -375,4 +501,44 @@ private void WriteContainerMounts(ContainerResource container)
...
+        {
+            // add the resource to the ReferencedResources for now. After the whole model is written,
+            // these will be written to the manifest
+            _referencedResources.TryAdd(resource.Name, resource);
+        }
+        else if (value is IValueWithReferences objectWithReferences)
+        {
+            foreach (var dependency in objectWithReferences.References)
+            {
+                if (!_currentDependencyStack.Contains(dependency))
```
davidfowl, line 524: perf: Use a hashset to tracking visited objects?

davidfowl (approved):
Small nits, can be addressed in follow up

playground/DatabaseMigration/DatabaseMigration.AppHost/aspire-manifest.json
```diff
@@ -65,6 +51,24 @@
...
+        "value": {
+          "type": "string",
+          "secret": true,
+          "default": {
+            "generate": {
+              "minLength": 22,
+              "minLower": 1,
+              "minUpper": 1,
+              "minNumeric": 1
+            }
```
eerhardt, line 68: @vhvb1989 - what if we made the `azd` rule that if a parameter has a `default`, we don't prompt. And a new optional property like `default.generate.manualPrompt: true` can put on the parameter to make `azd` prompt the user? That way by default, generated parameters just get generated. And only if someone opts in to prompting will it prompt.

Thoughts?

vhvb1989: having `allowManualPrompt` in `default.generate` sounds weird to me, because it would be next to the other rules, leading to questions like, what is allowManualPrompt=true + minLength=10.

I would add `allowManualPrompt` as a field for inputs.value

vhvb1989: I added support for azd for this with a field `NoManual` in the parameter metadata, here: https://github.com/Azure/azure-dev/pull/3580

So, a bicep parameter can be:
```bicep
@metadata({azd: {
  type: 'autoOrManualGeneration'
  config: {length:22}
  noManual:true
  }
})
@secure()
param pg1_password string
```

The noManual makes azd to skip the asking to customer and directly generating.

I can map any setting you add to the manifest to that azd feature

Issue #3089, opened 2024-03-22
Treating Azure OpenAI Deployments as resources


aaronpowell:
With the AOAI integration we can add a service resource into the AppHost like so:

```csharp
var openai = builder.AddAzureOpenAI("openai")
                    .AddDeployment(new("gpt-35-turbo", "gpt-35-turbo", "0613"));

builder.AddProject<Projects.OpenAIEndToEnd_WebStory>("webstory")
       .WithReference(openai);
```

But within our client project we're going to need to know the name of the deployment to invoke it in the SDK:

```csharp
    private ChatCompletionsOptions chatCompletionOptions = new ChatCompletionsOptions()
    {
        DeploymentName = "gpt-35-turbo",
        Messages =
        {
            new ChatRequestSystemMessage("Pick a random topic and write a sentence of a fictional story about it.")
        }
    };
```

Since the name of the deployment is any arbitrary string there's the possibility of it being miskeyed in the client application. My workaround approach to this is to have code like so:

```csharp
.AddDeployment(new(builder.Configuration["Azure:AI:ChatDeploymentName"] ?? "gpt-4", "gpt-4", "1106"))
```

Where I use the config or fallback to a known default, but I then need to ensure that the config value is passed to the client using `WithEnvironment`.

It'd be convenient if there was a way that the deployment names were flowed into the client application, either allowing them to be created as parameters with `AddParameter` (and then passing that into the `AzureOpenAIDeployment` constructor`) or when the `AzureOpenAIResource` is passed to the projects as a resource the deployment info is also passed (probably as a different environment variable) - although the discovery of this could be challenging and it'd fall down if you're using a connection string for local dev like this:

```csharp
var ai = builder.ExecutionContext.IsPublishMode ?
    builder.AddAzureOpenAI(ServiceNames.AzureOpenAI)
        .AddDeployment(new(builder.Configuration["Azure:AI:ChatDeploymentName"] ?? "gpt-4", "gpt-4", "1106"))
        .AddDeployment(new(builder.Configuration["Azure:AI:EmbeddingDeploymentName"] ?? "text-embedding-ada-002", "text-embedding-ada-002", "2")) :
    builder.AddConnectionString(ServiceNames.AzureOpenAI);
```

Since in this scenario the deployment info isn't available until provisioning happens.

mitchdenny:
Looping in @tg-msft.

This is one of the areas where having "deep linked connection strings" are really helpful. When you define a database resource via Aspire like the following:

[...]

PR #3092, opened 2024-03-22
Fix bug with deployments not being poulated.

mitchdenny:
Fix a bug discovered when looking at this issue: #3089

I wasn't adding the deployment to the list of deployments that were passed into the callback. Will need to check other resources for this bug as well but wanted this one fixed ASAP.

Also modified the playground code to show an example of how you can leverage the fact that the Open AI resource is a Bicep resource to pipe out the model name so it only has to be specified once.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3092)

### DIFF
--- a/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
+++ b/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
@@ -5,11 +5,16 @@ var builder = DistributedApplication.CreateBuilder(args);
 
 builder.AddAzureProvisioning();
 
-var openai = builder.AddAzureOpenAI("openai")
-                    .AddDeployment(new("gpt-35-turbo", "gpt-35-turbo", "0613"));
+#pragma warning disable ASPIRE0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+var openai = builder.AddAzureOpenAI("openai", (_, _, _, deployments) => {
+    var deployment = deployments.Single();
+    deployment.AddOutput("modelName", x => x.Name);
+}).AddDeployment(new("gpt-35-turbo", "gpt-35-turbo", "0613"));
+#pragma warning restore ASPIRE0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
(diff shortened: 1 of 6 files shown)


playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
```diff
@@ -5,11 +5,16 @@
 
 builder.AddAzureProvisioning();
 
-var openai = builder.AddAzureOpenAI("openai")
-                    .AddDeployment(new("gpt-35-turbo", "gpt-35-turbo", "0613"));
+#pragma warning disable ASPIRE0001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+var openai = builder.AddAzureOpenAI("openai", (_, _, _, deployments) => {
+    var deployment = deployments.Single();
+    deployment.AddOutput("modelName", x => x.Name);
```
davidfowl, line 11: Isn't this an input?

mitchdenny: We define the model name via AddDeployment ... the callback exposes it as a CDK resource, which we can then just pipe out. There are a couple of different ways to spin this using the tools we give people today but I think this is probably the easiest (unless we had deep linking connection strings - then none of this would be needed).

davidfowl: I'm just asking if this isn't part of the Deployment info we pass in?

mitchdenny: It is, so arguably it could just be defined as a standard C# variable. We still need the bug fixed though ;)

davidfowl: Make it a variable 😄

mitchdenny: Done.

davidfowl (approved):

PR #3093, opened 2024-03-22
Clean up distributed app builders to fix test failures

JamesNK:
Problem:

1. Test creates `DistributedApplicationBuilder`.
2. `DistributedApplicationBuilder` creates `HostApplicationBuilder`.
3. `HostApplicationBuilder` creates configuration, including a `FileSystemWatcher`.
4. Test never builds builder and never disposes of host. The `FileSystemWatcher` hangs around until finalizer runs.

There are many tests like this. Too many watchers exceeds a Linux limit and errors start.

Fix proposed here:

* ~Have an internal method on `DistributedApplicationBuilder` to clean up `HostApplicationBuilder`'s config. Tests will need to be updated to use it.~
* ~Have a test helper type `BuilderContainer` that makes it easy to create and "dispose" builders.~
* Have `TestDistributedApplicationBuilder` that wraps an inner `DistributedApplicationBuilder`. It implements dispose, and disposing the test builder will create a host and then dispose it.

~Only tests in `AWSCloudFormationResourceTests` are updated. All tests that create a builder but never create a host that is disposed will need to be changed.~
I updated the tests I could find that didn't build and dispose a host.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3093)


davidfowl (approved):

mitchdenny (approved):

tests/Aspire.Hosting.Tests/Utils/TestDistrubtedApplicationBuilder.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Aspire.Hosting.Utils;
+
+/// <summary>
+/// DistributedApplication.CreateBuilder() creates a builder that includes configuration to read from appsettings.json.
+/// The builder has a FileSystemWatcher, which can't be cleaned up unless a DistributedApplication is built and disposed.
+/// This class wraps the builder and provides a way to automatically dispose it to prevent test failures from excessive
+/// FileSystemWatcher instances from many tests.
+/// </summary>
+public sealed class TestDistrubtedApplicationBuilder : IDisposable, IDistributedApplicationBuilder
```
radical, line 16:
```suggestion
public sealed class TestDistributedApplicationBuilder : IDisposable, IDistributedApplicationBuilder
```

radical (approved):

PR #3104, opened 2024-03-22
Remove dependency from OpenTelemetry.Instrumentation.StackExchangeRedis

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3104)


src/Components/Aspire.StackExchange.Redis.DistributedCaching/Aspire.StackExchange.Redis.DistributedCaching.csproj
```diff
@@ -6,12 +6,19 @@
     <PackageTags>$(ComponentCachePackageTags) distributedcaching distributedcache redis</PackageTags>
     <Description>A Redis® implementation for IDistributedCache that integrates with Aspire, including health checks, logging, and telemetry.</Description>
     <PackageIconFullPath>$(SharedDir)redis-cube-red_white-rgb.png</PackageIconFullPath>
+    <IsAotCompatible>false</IsAotCompatible>
```
sebastienros, line 9: I think this would not be necessary if the `Shared` code was in a custom assembly that is itself marked aot-not-compatible, right? Is it worth it?

sebastienros:
Tests are broken by the change at `await notifier.ActivityReceived.WaitAsync(TimeSpan.FromSeconds(10));`

src/Components/Aspire.StackExchange.Redis/Aspire.StackExchange.Redis.csproj
```diff
@@ -7,19 +7,27 @@
     <Description>A generic Redis® client that integrates with Aspire, including health checks, logging, and telemetry.</Description>
     <PackageIconFullPath>$(SharedDir)redis-cube-red_white-rgb.png</PackageIconFullPath>
     <NoWarn>$(NoWarn);SYSLIB1100;SYSLIB1101</NoWarn>
+    <IsAotCompatible>false</IsAotCompatible>
```
eerhardt, line 10:
```suggestion
```

The library was AOT compatible before this change. We shouldn't be making it incompatible. We need to remove whatever code is causing this. I believe you may have copied too much.

src/Components/Aspire.StackExchange.Redis.DistributedCaching/Aspire.StackExchange.Redis.DistributedCaching.csproj
eerhardt:
```suggestion
```

This should be removed. These libraries are AOT compatible.

src/Components/Aspire.StackExchange.Redis/Aspire.StackExchange.Redis.csproj
eerhardt: I've pushed a change to revert this. I also opened https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1625 to port the fix to the underlying OTel code.

sebastienros: I updated the code sync instructions to apply the PR code until this is released.

src/Components/Aspire.StackExchange.Redis/Aspire.StackExchange.Redis.csproj
```diff
@@ -14,12 +14,19 @@
     <Compile Include="..\Common\ConfigurationSchemaAttributes.cs" Link="ConfigurationSchemaAttributes.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="$(VendoringDir)OpenTelemetry.Instrumentation.StackExchangeRedis\**\*.cs" LinkBase="OpenTelemetry.Instrumentation.StackExchangeRedis" />
+    <!-- Tests need access to StackExchangeRedisInstrumentationOptions which is made internal in /Vendoring -->
```
eerhardt, line 19: This is an interesting/unfortuante side-effect of this change. If users need to customize/configure the SERedis instrumentation options, previously they could. Now with this change, since we are making these options internal in our own package, customers can no longer customize it. Instead, they would have to disable `Tracing` from our component, and add it themselves, with their own configuration.

I'm not sure this changes our plans, but it is something we should be aware of.

src/Vendoring/README.md
```diff
@@ -10,31 +10,39 @@ git checkout tags/Instrumentation.SqlClient-1.7.0-beta.1
...
 
-## OpenTelemetry.Instrumentation.SqlClient
+## OpenTelemetry.Instrumentation.StackExchangeRedis
 
 ```console
-git clone https://github.com/open-telemetry/opentelemetry-dotnet.git
+git clone https://github.com/open-telemetry/opentelemetry-dotnet-contrib.git
 git fetch --tags
-git checkout tags/Instrumentation.SqlClient-1.7.0-beta.1
+git checkout tags/Instrumentation.StackExchangeRedis-1.0.0-rc9.13
```
eerhardt, line 31: What's your thoughts on putting these versions in? As a way to document which version we are currently on? How do we plan on keeping it up to date?

eerhardt (approved):

src/Vendoring/README.md
sebastienros: Just to show which versions are used right now. When there is a new release this has to be updated. Otherwise there is no way to know which is currently used and if work needs to be done. Also the way the otel repos is done each lib is release individually which doesn't help.

src/Components/Aspire.StackExchange.Redis/Aspire.StackExchange.Redis.csproj
sebastienros: Implementation detail, they can't change it until it's stable.

sebastienros: Or we need to create a custom options class (or change this one's namespace).

eerhardt: No, I don't think we need a custom options class for this. Following the above approach will work for people who hit this. The "real fix" here is to get SE Redis to use Activities intrinsically and not even have this Instrumentation library.

PR #3111, opened 2024-03-22
Templates: only add health checks in development

### LINKED ISSUE #3109: Update templates to only add default health checks endpoints in development
The templates currently add two health check endpoints in apps that reference the ServiceDefaults project and call the `MapDefaultEndpoints` method. These endpoints aren't secured and perform work based on all health checks registered in the project without any timeout or caching configured which isn't a good practice. Nothing in the default end-to-end experiences today actually relies on these endpoints but they're still useful to manually validate that health checks added by Aspire components are running.

We should update the templates to only add the health check endpoints when the environment is development and link to docs on how to properly configure health checks endpoints including restricting access via routing and integrating with request timeouts and output caching, e.g.:

```csharp
// Adding health checks endpoints to applications in non-development environments has security implications.
// See https://aka.ms/dotnet/aspire/healthchecks for details before enabling these endpoints in non-development environments.
if (app.Environment.IsDevelopment())
{
    // All health checks must pass for app to be considered ready to accept traffic after starting
    app.MapHealthChecks("/health");

    // Only health checks tagged with the "live" tag must pass for app to be considered alive
    app.MapHealthChecks("/alive", new HealthCheckOptions
    {
        Predicate = r => r.Tags.Contains("live")
    });
}
```

DamianEdwards:
Fixes #3109
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3111)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
@@ -101,9 +101,6 @@ public static class Extensions
 
     public static WebApplication MapDefaultEndpoints(this WebApplication app)
     {
-        // Uncomment the following line to enable the Prometheus endpoint (requires the OpenTelemetry.Exporter.Prometheus.AspNetCore package)
-        // app.MapPrometheusScrapingEndpoint();
-
         // Adding health checks endpoints to applications in non-development environments has security implications.
         // See https://aka.ms/dotnet/aspire/healthchecks for details before enabling these endpoints in non-development environments.
         if (app.Environment.IsDevelopment())
@@ -116,6 +113,9 @@ public static class Extensions
             {
                 Predicate = r => r.Tags.Contains("live")
             });
+
+            // Uncomment the following line to enable the Prometheus endpoint (requires the OpenTelemetry.Exporter.Prometheus.AspNetCore package)
+            // app.MapPrometheusScrapingEndpoint();
         }
 
         return app;
--- a/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
@@ -104,14 +104,19 @@ public static class Extensions
         // Uncomment the following line to enable the Prometheus endpoint (requires the OpenTelemetry.Exporter.Prometheus.AspNetCore package)
         // app.MapPrometheusScrapingEndpoint();
 
-        // All health checks must pass for app to be considered ready to accept traffic after starting
-        app.MapHealthChecks("/health");
-
-        // Only health checks tagged with the "live" tag must pass for app to be considered alive
-        app.MapHealthChecks("/alive", new HealthCheckOptions
+        // Adding health checks endpoints to applications in non-development environments has security implications.
+        // See https://aka.ms/dotnet/aspire/healthchecks for details before enabling these endpoints in non-development environments.
+        if (app.Environment.IsDevelopment())
         {
-            Predicate = r => r.Tags.Contains("live")
-        });
+            // All health checks must pass for app to be considered ready to accept traffic after starting
+            app.MapHealthChecks("/health");
+
+            // Only health checks tagged with the "live" tag must pass for app to be considered alive
+            app.MapHealthChecks("/alive", new HealthCheckOptions
+            {
+                Predicate = r => r.Tags.Contains("live")
+            });
+        }
 
         return app;
     }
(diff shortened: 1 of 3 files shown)


danmoseley:
do we need follow up PR's for eshop and eshoponazure?

DamianEdwards:
> do we need follow up PR's for eshop and eshoponazure?

Probably not a bad idea, but eShop should be doing its own threat modelling and security reviews too.

src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
```diff
@@ -104,14 +104,19 @@ public static WebApplication MapDefaultEndpoints(this WebApplication app)
...
         // app.MapPrometheusScrapingEndpoint();
 
-        // All health checks must pass for app to be considered ready to accept traffic after starting
-        app.MapHealthChecks("/health");
-
-        // Only health checks tagged with the "live" tag must pass for app to be considered alive
-        app.MapHealthChecks("/alive", new HealthCheckOptions
+        // Adding health checks endpoints to applications in non-development environments has security implications.
+        // See https://aka.ms/dotnet/aspire/healthchecks for details before enabling these endpoints in non-development environments.
+        if (app.Environment.IsDevelopment())
```
joperezr, line 109: While this works, it feels a bit like instead of making this method a no-op we'd ideally condition the callsites in the Program.cs files for each microservice instead. I can understand why it's more convinient to do it here, but just adding the obvious comment in case we want to consider it.

joperezr: Also, should the condition wrap the commented out code for Prometheus too?

joperezr (approved):
Left a couple of comments/questions, but this looks good overall. Thanks for taking care of this @DamianEdwards!

src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
DamianEdwards: > we'd ideally condition the callsites in the Program.cs files for each microservice instead

The problem with that is that the call to `MapDefaultEndpoints` is added by the VS tooling for enlisting in Aspire orchestration so I don't think it really scales to do it there.

> Also, should the condition wrap the commented out code for Prometheus too?

What's the thinking there? That the comment applies to both? I don't think putting the Prometheus endpoint in dev only, commented out, is that useful, but I could see that the comment about security implications of these kinds of endpoints could be useful to expand to cover Prometheus too.

joperezr: > The problem with that is that the call to MapDefaultEndpoints is added by the VS tooling for enlisting in Aspire orchestration so I don't think it really scales to do it there.

Yeah, that's what I thought. We could of course ask the VS Tooling folks to also emit the condition but that would be likely hard to do at this stage. I'm fine with this approach, I just think that in an ideal world I feel like the other one would be a bit more accurate.

> What's the thinking there? That the comment applies to both? I don't think putting the Prometheus endpoint in dev only, commented out, is that useful, but I could see that the comment about security implications of these kinds of endpoints could be useful to expand to cover Prometheus too.

Yeah, mainly thinking about security implications. Your comment above the if is specific about health checks, but technically speaking it applies to any endpoints you expose with your app, and given these are all enabled for all of your microservices by default I thought it might make sense to have the user explicitly take action to expose these in prod too.

src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
```diff
@@ -101,17 +101,22 @@ public static IHostApplicationBuilder AddDefaultHealthChecks(this IHostApplicati
 
     public static WebApplication MapDefaultEndpoints(this WebApplication app)
     {
-        // Uncomment the following line to enable the Prometheus endpoint (requires the OpenTelemetry.Exporter.Prometheus.AspNetCore package)
-        // app.MapPrometheusScrapingEndpoint();
+        // Adding health checks endpoints to applications in non-development environments has security implications.
+        // See https://aka.ms/dotnet/aspire/healthchecks for details before enabling these endpoints in non-development environments.
+        if (app.Environment.IsDevelopment())
```
eerhardt, line 106: Was there anything relying on these endpoints? For example, does ACA/k8s query them at all? If not, how does ACA know if the app is health or not?

DamianEdwards: Nothing in our E2E relies on these today.

PR #3114, opened 2024-03-22
Service Discovery API refactoring

ReubenBond:
Our requirements on Service Discovery have evolved since we first created it, eg we added more semantics to the input string. There were also some vestigial types, members, and patterns to clean up. [API review](https://github.com/dotnet/aspnetcore/issues/53715#issuecomment-1986258344) also revealed some things which could be simplified or changed to align with prevailing patterns.

This PR addresses these points by refactoring the API:

* Move types in the *.Abstractions assembly out of the .Abstractions namespace, and clean up namespaces in general
* Hide types & members which do not need to be exposed. Delete unused types (eg, unused `IServiceEndPointSelector` impls).
* Seal most types
* Simplify the resolver API by removing the `ResolutionStatus` type
* Rename `IHttpClientBuilder` extension from `UseServiceDiscovery` to `AddServiceDiscovery` to align with existing patterns. The existing method is kept and marked as `[Obsolete]` for now.
* Add a `ServiceEndPointQuery` type to communicate the parsed input from the core resolver to the providers
* Resolve `IServiceEndPointSelector` from DI instead of using a factory (and register it as transient).
* Add missing overloads to hosting methods. Specifically, overloads which accept configuration delegates.
* General code style cleanup (primary constructors, etc)
* Rename `ServiceEndPointCollectionSource` to `ServiceEndPointBuilder`, make internal, and extract public interface: `IServiceEndPointBuilder`
* Rename `ServiceEndPointCollection` to `ServiceEndPointSource` to match existing ASP.NET pattern

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3114)


src/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndPointQuery.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// <summary>
+    /// Gets the string which the query was constructed from.
+    /// </summary>
+    public string OriginalString { get; }
+
+    /// <summary>
+    /// Gets the collection of included URI schemes.
+    /// </summary>
+    public string[] IncludeSchemes { get; }
```
davidfowl, line 116: Why array an not IReadOnlyList<string>

src/Microsoft.Extensions.ServiceDiscovery.Abstractions/ServiceEndPointQuery.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Gets the string which the query was constructed from.
+    /// </summary>
+    public string OriginalString { get; }
+
+    /// <summary>
+    /// Gets the collection of included URI schemes.
```
adityamandaleeka, line 114: Consider saying something about the ordering here?

src/Microsoft.Extensions.ServiceDiscovery.Dns/DnsServiceEndPointResolverBase.cs
```diff
@@ -95,61 +94,29 @@ public async ValueTask<ResolutionStatus> ResolveAsync(ServiceEndPointCollectionS
...
-                    // soon and to retry for some period until they do.
-                    _lastStatus = ResolutionStatus.FromPending(exception);
-                }
-                else
-                {
-                    _lastStatus = ResolutionStatus.FromException(exception);
-                }
-            }
-            else if (endPoints is not { Count: > 0 })
+            if (endPoints is not { Count: > 0 })
```
adityamandaleeka, line 109: Is there a reason why you put the `not` version of this in the `if` instead of flipping it? Not a big deal but I think the other way would be slightly more readable.

src/Microsoft.Extensions.ServiceDiscovery.Dns/DnsServiceEndPointResolverBase.cs
```diff
@@ -169,7 +136,7 @@ private void SetResult(List<ServiceEndPoint>? endPoints, Exception? exception, T
 
         TimeSpan GetRefreshPeriod()
         {
-            if (_lastStatus.StatusCode is ResolutionStatusCode.Success)
+            if (_hasEndpoints)
             {
                 return MinRetryPeriod;
             }
```
adityamandaleeka, line 142: For the bit below (can't actually comment there in this PR), should we protect against overflow?

src/Microsoft.Extensions.ServiceDiscovery.Yarp/ServiceDiscoveryReverseProxyServiceCollectionExtensions.cs
```diff
@@ -11,7 +10,7 @@ namespace Microsoft.Extensions.DependencyInjection;
 /// <summary>
 /// Extensions for <see cref="IServiceCollection"/> used to register the ReverseProxy's components.
 /// </summary>
-public static class ReverseProxyServiceCollectionExtensions
+public static class ServiceDiscoveryReverseProxyServiceCollectionExtensions
```
adityamandaleeka, line 13: That's a name...

ReubenBond: Thankfully, it's just an extension class, so no one needs to actually type this.

src/Microsoft.Extensions.ServiceDiscovery.Dns/DnsServiceEndPointResolverBase.cs
ReubenBond: I'll pin to max if the value overflows.

ReubenBond:
image
We should merge this for this fact alone

adityamandaleeka (approved):

PR #3116, opened 2024-03-23
Allow packages to not go stable

### LINKED ISSUE #2541: Add the ability to keep packages in preview
When we ship for 8.0 GA, there are going to be some Components that shouldn't be shipped as "stable". For example, https://www.nuget.org/packages/Azure.AI.OpenAI/ does not have a stable version yet. If it doesn't have a stable version before we ship for 8.0 GA, we shouldn't (can't?) ship our `Aspire.Azure.AI.OpenAI` component as a stable version.

We should have some build infrastructure that allows for specific packages to be built with a preview tag, even when the rest of the packages are "stable".

cc @joperezr @DamianEdwards @sebastienros @radical

joperezr:
Fixes #2541

cc: @eerhardt 

With these changes, now individual projects can decide to stay marked as preview even when the rest of the repo is producing stable packages. Once we branch for GA, We will update `StabilizePackageVersion` and set it to `true` meaning that the repo will start producing stable packages. That said, individual projects will be able to stay in prerelease by adding the following property in their .csproj:

```xml
    <PropertyGroup>
        <!-- This project will remain producing prerelease versions, even when the rest of the repo is producing stable packages -->
        <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
    </PropertyGroup>
``` 

This model is the same as the one used by dotnet/runtime repo.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3116)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -7,6 +7,9 @@
     <VersionPrefix>8.0.0</VersionPrefix>
     <PreReleaseVersionLabel>preview.5</PreReleaseVersionLabel>
     <UseVSTestRunner>true</UseVSTestRunner>
+    <!-- Enable to remove prerelease label. -->
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
   </PropertyGroup>
   <PropertyGroup>
     <!-- Package versions defined directly in <reporoot>/Directory.Packages.props -->


joperezr:
Oh, I wasn't aware of the property's existence 😋. 

I'm assuming this would just work if that is set with the changes I have here, but I'll double-check.

eerhardt:
> I'm assuming this would just work if that is set with the changes I have here

Agreed. I guess I was mostly commenting on the PR description above in the top post. For the libraries we don't want to go stable, we should just need to set `<SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>` in their .csproj.

eerhardt (approved):
LGTM

radical:
> For the libraries we don't want to go stable, we should just need to set `<SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>` in their .csproj.

The PR description should be updated to reference this property.

radical (approved):

joperezr:
Sorry this took a while, I have just validated that the property indeed works with these changes and packages would remain unstable.

joperezr:
> > For the libraries we don't want to go stable, we should just need to set `<SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>` in their .csproj.
> 
> The PR description should be updated to reference this property.

Updated.

joperezr:
You are correct, fixing that now.

PR #3118, opened 2024-03-23
Fix SqlClient component activity name

sebastienros:
The conformance tests should have failed but they did not. And did not for other components with the same problem (Redis). We need to investigate.

Note that traces work fine without this change because we are calling `AddSqlClientInstrumentation()` which uses the same wrong activity name. But this could be an issue if other libraries are using the name coming from the nuget package.

/cc @rajkumar-rangaraj @reyang since the azure sdk has the same problem.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3118)

### DIFF
--- a/src/Vendoring/.editorconfig
+++ b/src/Vendoring/.editorconfig
@@ -15,3 +15,5 @@ dotnet_diagnostic.CA1852.severity = silent
 # IDE0060: Remove unused parameters
 dotnet_diagnostic.IDE0060.severity = silent
 
+# IDE0005: Using directive is unnecessary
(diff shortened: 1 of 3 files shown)


rajkumar-rangaraj:
Good catch, we will fix this in Azure SDK. 
This issue may not impact the Azure SDK as we back off from internal SQL instrumentation if the customer brings a reference to OpenTelemetry.Instrumentation.SqlClient package.

eerhardt (approved):

PR #3119, opened 2024-03-23
Organize dashboard config to use strongly typed options, support primary/secondary API keys and rotation

### LINKED ISSUE #2996: Make dashboard configuration align with standard .NET configuration
Most configuration in .NET uses Microsoft.Extensions.Configuration, which supports great features such as sections.

Aspire dashboard has a rapidly growing number of configuration options:
* Endpoints
* Auth
* OTLP limits
* Logging (built-in config)
* Kestrel (built-in config)

Right the config options we've added to Aspire don't follow the .NET naming standard for config or sections. For example, setting OTLP auth mode has a key called `DOTNET_DASHBOARD_OTLP_AUTH_MODE` instead of something like `AspireDashboard__Otlp__AuthMode` (double underscores indicates a section delimiter in env vars).

There are pros and cons to each. `DOTNET_DASHBOARD_OTLP_AUTH_MODE` is optimized for command line and env var usage, but it doesn't match built-in config values, and isn't nested into sections in JSON files.

We should consider consistently using the .NET standard for naming and organizing dashboard configuration.

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/2996

* Change dashboard config to be loaded via [options](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-8.0)
  * Use options validation
  * Use options post config to apply shortcut config over the top of structured config, e.g. `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL` is set to `Dashboard:Otlp:EndpointUrl`.
* Change OLTP API key auth to have primary and secondary keys
* Change OTLP API key auth to use `IOptionsMonitor<T>` to get latest keys after config changes
* Add `DOTNET_DASHBOARD_CONFIG_FILE_PATH` which configs a location to load JSON from a file. This can be used in cloud hosting scenarios to point at a mounted volume. Changes to the file can be applied to dashboard without restarting the app (for API key rotation without process restart)
* Tests

TODO:
* ~Resource service endpoint URL doesn't use new config yet~
* ~Update dashboard readme~
* ~Remove old code~
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3119)

### DIFF
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Security.Cryptography.X509Certificates;
 
 namespace Aspire.Dashboard.Configuration;
 
@@ -47,6 +48,8 @@ public sealed class ResourceServiceClientCertificateOptions
     public string? FilePath { get; set; }
     public string? Password { get; set; }
     public string? Subject { get; set; }
+    public string? StoreName { get; set; }
+    public StoreLocation? Location { get; set; }
 }
 
 public sealed class OtlpOptions
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -48,7 +48,7 @@ public sealed class ResourceServiceClientCertificateOptions
     public string? FilePath { get; set; }
     public string? Password { get; set; }
     public string? Subject { get; set; }
-    public string? StoreName { get; set; }
+    public string? Store { get; set; }
     public StoreLocation? Location { get; set; }
 }
 
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -13,7 +13,7 @@ public sealed class DashboardOptions
     public OtlpOptions Otlp { get; set; } = new OtlpOptions();
     public FrontendOptions Frontend { get; set; } = new FrontendOptions();
     public ResourceServiceClientOptions ResourceServiceClient { get; set; } = new ResourceServiceClientOptions();
-    public TelemetryLimits TelemetryLimits { get; set; } = new TelemetryLimits();
+    public TelemetryLimitOptions TelemetryLimits { get; set; } = new TelemetryLimitOptions();
 }
 
 public sealed class ResourceServiceClientOptions
@@ -130,7 +130,7 @@ public sealed class FrontendOptions
     }
 }
 
-public sealed class TelemetryLimits
+public sealed class TelemetryLimitOptions
 {
     public int MaxLogCount { get; set; } = 10_000;
     public int MaxTraceCount { get; set; } = 10_000;
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -1,0 +1,138 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Aspire.Dashboard.Configuration;
+
+public sealed class DashboardOptions
+{
+    public string? ApplicationName { get; set; }
+    public OtlpOptions Otlp { get; set; } = new OtlpOptions();
+    public FrontendOptions Frontend { get; set; } = new FrontendOptions();
+    public ResourceServiceClientOptions ResourceServiceClient { get; set; } = new ResourceServiceClientOptions();
+    public TelemetryLimits TelemetryLimits { get; set; } = new TelemetryLimits();
+}
+
+public sealed class ResourceServiceClientOptions
+{
+    private Uri? _parsedUrl;
+
+    public string? Url { get; set; }
+    public ResourceClientAuthMode? AuthMode { get; set; }
+    public ResourceServiceClientCertificateOptions ClientCertificates { get; set; } = new ResourceServiceClientCertificateOptions();
+
+    public Uri? GetUri() => _parsedUrl;
+
+    internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
+    {
+        if (!string.IsNullOrEmpty(Url))
+        {
+            if (!Uri.TryCreate(Url, UriKind.Absolute, out _parsedUrl))
+            {
+                errorMessage = $"Failed to parse resource service client endpoint URL '{Url}'.";
+                return false;
+            }
+        }
+
+        errorMessage = null;
+        return true;
+    }
+}
+
+public sealed class ResourceServiceClientCertificateOptions
+{
+    public DashboardClientCertificateSource? Source { get; set; }
+    public string? FilePath { get; set; }
+    public string? Password { get; set; }
+    public string? Subject { get; set; }
+}
+
+public sealed class OtlpOptions
+{
+    private Uri? _parsedEndpointUrl;
+
+    public string? PrimaryApiKey { get; set; }
+    public string? SecondaryApiKey { get; set; }
+    public OtlpAuthMode? AuthMode { get; set; }
+    public string? EndpointUrl { get; set; }
+
+    public Uri GetEndpointUri()
+    {
+        Debug.Assert(_parsedEndpointUrl is not null, "Should have been parsed during validation.");
+        return _parsedEndpointUrl;
+    }
+
+    internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
+    {
+        if (string.IsNullOrEmpty(EndpointUrl))
+        {
+            errorMessage = "OTLP endpoint URL is not configured. Specify a Dashboard:Otlp:EndpointUrl value.";
+            return false;
+        }
+        else
+        {
+            if (!Uri.TryCreate(EndpointUrl, UriKind.Absolute, out _parsedEndpointUrl))
+            {
+                errorMessage = $"Failed to parse OTLP endpoint URL '{EndpointUrl}'.";
+                return false;
+            }
+        }
+
+        errorMessage = null;
+        return true;
+    }
+}
+
+public sealed class FrontendOptions
+{
+    private List<Uri>? _parsedEndpointUrls;
+
+    public string? EndpointUrls { get; set; }
+    public FrontendAuthMode? AuthMode { get; set; }
+
+    public IReadOnlyList<Uri> GetEndpointUris()
+    {
+        Debug.Assert(_parsedEndpointUrls is not null, "Should have been parsed during validation.");
+        return _parsedEndpointUrls;
+    }
+
+    internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
+    {
+        if (string.IsNullOrEmpty(EndpointUrls))
+        {
+            errorMessage = "One or more frontend endpoint URLs are not configured. Specify a Dashboard:Frontend:EndpointUrls value.";
+            return false;
+        }
+        else
+        {
+            var parts = EndpointUrls.Split(';', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            var uris = new List<Uri>(parts.Length);
+            foreach (var part in parts)
+            {
+                if (!Uri.TryCreate(part, UriKind.Absolute, out var uri))
+                {
+                    errorMessage = $"Failed to parse frontend endpoint URLs '{EndpointUrls}'.";
+                    return false;
+                }
+
+                uris.Add(uri);
+            }
+            _parsedEndpointUrls = uris;
+        }
+
+        errorMessage = null;
+        return true;
+    }
+}
+
+public sealed class TelemetryLimits
+{
+    public int MaxLogCount { get; set; } = 10_000;
+    public int MaxTraceCount { get; set; } = 10_000;
+    public int MaxMetricsCount { get; set; } = 50_000; // Allows for 1 metric point per second for over 12 hours.
+    public int MaxAttributeCount { get; set; } = 128;
+    public int MaxAttributeLength { get; set; } = int.MaxValue;
+    public int MaxSpanEventCount { get; set; } = int.MaxValue;
+}
--- a/src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
@@ -1,0 +1,86 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Configuration;
+
+public sealed class ValidateDashboardOptions : IValidateOptions<DashboardOptions>
+{
+    public ValidateOptionsResult Validate(string? name, DashboardOptions options)
+    {
+        var errorMessages = new List<string>();
+
+        if (!options.Frontend.TryParseOptions(out var frontendParseErrorMessage))
+        {
+            errorMessages.Add(frontendParseErrorMessage);
+        }
+
+        if (!options.Otlp.TryParseOptions(out var otlpParseErrorMessage))
+        {
+            errorMessages.Add(otlpParseErrorMessage);
+        }
+
+        if (!options.ResourceServiceClient.TryParseOptions(out var resourceServiceClientParseErrorMessage))
+        {
+            errorMessages.Add(resourceServiceClientParseErrorMessage);
+        }
+
+        switch (options.Otlp.AuthMode)
+        {
+            case OtlpAuthMode.Unsecured:
+                break;
+            case OtlpAuthMode.ApiKey:
+                if (string.IsNullOrEmpty(options.Otlp.PrimaryApiKey))
+                {
+                    errorMessages.Add("PrimaryApiKey is required when OTLP authentication mode is API key. Specify a Dashboard:Otlp:PrimaryApiKey value.");
+                }
+                break;
+            case OtlpAuthMode.ClientCertificate:
+                break;
+            case null:
+                errorMessages.Add($"OTLP endpoint authentication is not configured. Either specify DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS with a value of true, or specify Dashboard:Otlp:AuthMode. Possible values: {string.Join(", ", typeof(OtlpAuthMode).GetEnumNames())}");
+                break;
+            default:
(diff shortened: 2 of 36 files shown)


src/Shared/DashboardConfigNames.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Aspire.Hosting;
+
+internal static class DashboardConfigNames
+{
+    internal const string DashboardConfigFilePathVariableName = "DOTNET_DASHBOARD_CONFIG_FILE_PATH";
+    internal const string DashboardInsecureAllowAnonymousVariableName = "DOTNET_DASHBOARD_INSECURE_ALLOW_ANONYMOUS";
+    internal const string DashboardOtlpAuthModeVariableName = "Dashboard:Otlp:AuthMode";
+    internal const string DashboardOtlpPrimaryApiKeyVariableName = "Dashboard:Otlp:PrimaryApiKey";
+    internal const string DashboardOtlpSecondaryApiKeyVariableName = "Dashboard:Otlp:SecondaryApiKey";
```
davidfowl, line 12: These are environment variables for the config system right? This won't work on anything but windows.

```suggestion
    internal const string DashboardOtlpAuthModeVariableName = "Dashboard__Otlp__AuthMode";
    internal const string DashboardOtlpPrimaryApiKeyVariableName = "Dashboard__Otlp__PrimaryApiKey";
    internal const string DashboardOtlpSecondaryApiKeyVariableName = "Dashboard__Otlp__SecondaryApiKey";
```

JamesNK: These keys are used to both set env vars and read from config.

Will `config["Dashboard__Otlp__AuthMode"]` work or must it be `config["Dashboard:Otlp:AuthMode"]`?

If `config["Dashboard__Otlp__AuthMode"]` doesn't work then there will just need to be a couple of extra constants for the env var names.

davidfowl: You need separate ones for env variables vs config then.

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Configuration;
+
+public sealed class ValidateDashboardOptions : IValidateOptions<DashboardOptions>
+{
+    public ValidateOptionsResult Validate(string? name, DashboardOptions options)
```
davidfowl, line 10: Check for valid uris in here?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -63,20 +67,30 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
         builder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
 #endif
 
-        var dashboardConfig = LoadDashboardConfig(builder.Configuration);
+        if (builder.Configuration[DashboardConfigNames.DashboardConfigFilePathName.ConfigKey] is { Length: > 0 } configFilePath)
+        {
+            builder.Configuration.AddJsonFile(configFilePath, optional: true, reloadOnChange: true);
```
davidfowl, line 72: Should it be optional if you specified it?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -63,20 +67,30 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
...
+        }
+
+        var dashboardConfigSection = builder.Configuration.GetSection("Dashboard");
+        builder.Services.Configure<DashboardOptions>(dashboardConfigSection);
+        builder.Services.AddSingleton<IPostConfigureOptions<DashboardOptions>, PostConfigureDashboardOptions>();
+        builder.Services.AddSingleton<IValidateOptions<DashboardOptions>, ValidateDashboardOptions>();
```
davidfowl, line 78: This is fine, but there are methods to make this easier. 
```
builder.Services.AddOptions<T>().PostConfigure(..).ValidateOptions(...)
```

JamesNK: I looked at them but they're based around registering funcs. I wanted to register types.

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
JamesNK: I made parsing part of setting the properties. It will get validated when the value is set (error if it fails).

davidfowl: Just beware of this https://github.com/dotnet/runtime/issues/44381

JamesNK: Hmmm, ok! I'll change it.

JamesNK: Done. If I had more time, I'd look at what other places do here, but I think this is functional.

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -34,8 +37,6 @@ namespace Aspire.Dashboard.Model;
 /// </remarks>
 internal sealed class DashboardClient : IDashboardClient
 {
-    private const string ResourceServiceUrlVariableName = "DOTNET_RESOURCE_SERVICE_ENDPOINT_URL";
```
JamesNK, line 37: Setting the resource service won't be very common. I replaced it with `Dashboard:ResourceServiceClient:Url`.

davidfowl: Why won't it be common?

JamesNK: Let me rephrase. It won't be common for people to start the dashboard with their own resource service. Aspire tooling will (DCP) and some cloud hosters might setup a resource service, but people just playing around with the dashboard won't.

I removed it to keep envvar/cmd arg optimized variables like this focused on the most commonly used settings (endpoint addresses and endpoint security). Do you want it added back?

davidfowl: Yea I think we should add it back. It's the other primary function of the dashboard. We'll have other implementations soon.

JamesNK:
Review please 🙏

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+public sealed class ResourceServiceClientCertificateOptions
+{
+    public DashboardClientCertificateSource? Source { get; set; }
+    public string? FilePath { get; set; }
+    public string? Password { get; set; }
+    public string? Subject { get; set; }
+    public string? StoreName { get; set; }
+    public StoreLocation? Location { get; set; }
```
JamesNK, line 52: @drewnoakes I changed StoreName and Location just properties on the cert options.

Couple of reasons:
* This matches what Kestrel endpoint config does.
* Having just store name and location in a nested key store section seems odd when FilePath/Password isn't nested under a file section.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        errorMessage = null;
+        return true;
+    }
+}
+
+public sealed class TelemetryLimits
+{
+    public int MaxLogCount { get; set; } = 10_000;
+    public int MaxTraceCount { get; set; } = 10_000;
```
davidfowl, line 136: Are these per resource?

davidfowl:
@eerhardt we should use the config binder source gen and config schema generator, right?

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
JamesNK: Yes

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -110,6 +123,10 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
 
         _app = builder.Build();
 
+        _dashboardOptionsMonitor = _app.Services.GetRequiredService<IOptionsMonitor<DashboardOptions>>();
+        // Get options to run validation during startup.
+        _ = _dashboardOptionsMonitor.CurrentValue;
```
davidfowl, line 128: There's a method for this https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.optionsbuilderextensions.validateonstart?view=dotnet-plat-ext-8.0

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -210,39 +227,59 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
...
+    private static DashboardOptions GetDashboardOptions(WebApplicationBuilder builder, IConfigurationSection dashboardConfigSection)
+    {
+        var dashboardOptions = new DashboardOptions();
+        dashboardConfigSection.Bind(dashboardOptions);
+        new PostConfigureDashboardOptions(builder.Configuration).PostConfigure(name: string.Empty, dashboardOptions);
+        var result = new ValidateDashboardOptions().Validate(name: string.Empty, dashboardOptions);
+        if (result.Failed)
+        {
+            throw new OptionsValidationException(optionsName: string.Empty, typeof(DashboardOptions), result.Failures);
+        }
```
davidfowl, line 243: Does this look reasonable to the user who launches this standalone?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -322,23 +359,17 @@ static Func<EndpointInfo> CreateEndPointAccessor(ListenOptions options, bool isH
...
-                o.OtlpAuthMode = dashboardStartupConfig.OtlpAuthMode;
-            })
-            .AddScheme<OtlpApiKeyAuthenticationHandlerOptions, OtlpApiKeyAuthenticationHandler>(OtlpApiKeyAuthenticationDefaults.AuthenticationScheme, o =>
-            {
-                o.OtlpApiKey = dashboardStartupConfig.OtlpApiKey;
-            })
+            .AddScheme<OtlpCompositeAuthenticationHandlerOptions, OtlpCompositeAuthenticationHandler>(OtlpCompositeAuthenticationDefaults.AuthenticationScheme, o => { })
+            .AddScheme<OtlpApiKeyAuthenticationHandlerOptions, OtlpApiKeyAuthenticationHandler>(OtlpApiKeyAuthenticationDefaults.AuthenticationScheme, o => { })
             .AddScheme<OtlpConnectionAuthenticationHandlerOptions, OtlpConnectionAuthenticationHandler>(OtlpConnectionAuthenticationDefaults.AuthenticationScheme, o => { })
             .AddCertificate(options =>
```
davidfowl, line 369: Not important for this PR but it would look nicer if these were hidden in a helper extension method. Sorta like AddCertifcate is 😄

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -151,31 +150,19 @@ GrpcChannel CreateChannel()
...
 
             X509CertificateCollection GetKeyStoreCertificate()
             {
-                var subject = configuration["ResourceServiceClient:ClientCertificate:Subject"];
-
-                if (subject is null or [])
-                {
-                    throw new InvalidOperationException("ResourceServiceClient:ClientCertificate:Source is \"KeyStore\", but no Certificate:FilePath is configured.");
-                }
+                var subject = _dashboardOptions.ResourceServiceClient.ClientCertificates.Subject!;
```
davidfowl, line 161: Subject is never null?

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
```diff
@@ -14,16 +16,6 @@
...
-    public int TraceCountLimit { get; init; }
-    public int MetricsCountLimit { get; init; }
-    public int AttributeCountLimit { get; init; }
-    public int AttributeLengthLimit { get; init; }
-    public int SpanEventCountLimit { get; init; }
-}
-
 public sealed class TelemetryRepository
 {
     internal const string LogCountLimitKey = "DOTNET_DASHBOARD_OTEL_LOG_COUNT_LIMIT";
```
davidfowl, line 21: Shouldn't these be in a PostConfigure too? Where are these used?

src/Aspire.Dashboard/README.md
```diff
@@ -1,65 +1,94 @@
...
-### Frontend
+- `ASPNETCORE_URLS` specifies one or more HTTP endpoints through which the dashboard frontend is served. The frontend endpoint is used to view the dashboard in a browser. Defaults to http://localhost:18888.
+- `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL` specifies the OTLP endpoint. OTLP endpoint hosts an OTLP service and recevies telemetry. Defaults to http://localhost:18889.
+- `DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS` specifies the dashboard doesn't use authentication and accepts anonymous access. This setting is a shortcut to configuring `Dashboard:Frontend:AuthMode` and `Dashboard:Otlp:AuthMode` to `Unsecured`.
+- `DOTNET_DASHBOARD_CONFIG_FILE_PATH` specifies the path for an optional JSON configuration file.
 
-The dashboard's web application frontend supports OpenID Connect (OIDC). Set `Frontend:AuthMode` to `OpenIdConnect`, then add the following configuration:
+### Frontend authentication
+
+The dashboard's frontend supports OpenID Connect (OIDC). Set `Dashboard:Frontend:AuthMode` to `OpenIdConnect`, then add the following configuration:
```
davidfowl, line 41: I would put this under an OIDC/OAuth2 subheading (we're going to add more auth types)

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: It's a little funky. The exception message is all the failure messages joined together with semi-colons. It's informative but ugly. Probably also a big stack trace.

When P5 docker image is published I'll see what it looks like and what can be done to improve it.

JamesNK: Issue: https://github.com/dotnet/aspire/issues/3168

src/Aspire.Dashboard/Model/DashboardClient.cs
```diff
@@ -119,7 +118,7 @@ GrpcChannel CreateChannel()
                     ClientCertificates = certificates
                 };
 
-                configuration.Bind("ResourceServiceClient:Ssl", httpHandler.SslOptions);
+                configuration.Bind("Dashboard:ResourceServiceClient:Ssl", httpHandler.SslOptions);
```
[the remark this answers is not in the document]
JamesNK: @drewnoakes ?

I think it's used as a simple way to set TLS config in tests without having to explicitly add more config.

src/Aspire.Dashboard/Model/DashboardClient.cs
JamesNK: Validation will fail if it's null. Won't get to this point. I'll add a comment.

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
JamesNK: These should be removed.

src/Aspire.Dashboard/Model/DashboardClient.cs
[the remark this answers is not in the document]
davidfowl: I think we should own all of the options types and copy settings over explicitly. Less fragile. Can be follow up.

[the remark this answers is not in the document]
JamesNK: What do you think about continuing to do this, but not advertising it in docs? In other words, it's an internal implementaiton detail. A private API.

Then we add new options if people ask for them.

src/Aspire.Dashboard/README.md
JamesNK: I like the idea. Let's update the readme like that once more auth options are added.

src/Aspire.Dashboard/Model/DashboardClient.cs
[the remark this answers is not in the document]
davidfowl: Can we preserve the schema?

davidfowl (approved):
Left feedback, but this LGTM

src/Aspire.Dashboard/Model/DashboardClient.cs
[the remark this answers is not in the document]
JamesNK: I don't understand. What do you mean?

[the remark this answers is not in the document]
davidfowl: You said don't document it, I assume that to mean, we don't document it because we might change it later so it would be a breaking change if we do.

I'm asking if we know what the existing schema is, and can we preserve it to avoid breaking people.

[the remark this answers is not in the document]
JamesNK: It's the public getters and setters on https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslclientauthenticationoptions?view=net-8.0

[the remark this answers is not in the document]
davidfowl: I'm worried about this type adding something then suddenly config binding breaks. Or we switch to the source generator then it breaks.

drewnoakes (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -67,20 +69,33 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
         builder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
 #endif
 
-        var dashboardConfig = LoadDashboardConfig(builder.Configuration);
+        // Allow for a user specified JSON config file on disk. Throw an error if the specified file doesn't exist.
```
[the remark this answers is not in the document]
JamesNK: It does. There is a unit test of it throwing.

src/Shared/DashboardConfigNames.cs
```diff
@@ -0,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting;
+
+internal static class DashboardConfigNames
```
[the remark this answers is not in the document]
JamesNK: It's internal because it is shared code with the host, and the host is referenced. Internal is correct.

PR #3124, opened 2024-03-24
Allowing for versioning in deployments.

mitchdenny:
This is a draft exploring two changes:

1. Change CDK callback to be a context object (currently called a definition).
2. Provide a generic overload for AddAzureKeyVault(...) which allows the type of the definition to be supplied.
3. Depending on the definition the shape of the Bicep produced varies.
4. There is a default definition that the non-generic overloads use.

Trying to kill two birds with one stone here. We want to simplify the callbacks on the CDK resource types, but we also want a way to allow for changing what it means to deploy a KeyVault if best practices change over time.

This PR won't go in as is (we'll back out some of the default/versioning stuff) but just want to get a feel for what we would ship in the future would feel like (we will probably introduce the context/definition object in GA).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3124)

### DIFF
--- a/playground/cdk/CdkSample.AppHost/Program.cs
+++ b/playground/cdk/CdkSample.AppHost/Program.cs
@@ -4,6 +4,7 @@
 #pragma warning disable ASPIRE0001 // Because we use the CDK callbacks.
 
 using Aspire.Hosting.Azure;
+using Aspire.Hosting.Azure.KeyVault;
 using Azure.Provisioning.KeyVaults;
 using Azure.ResourceManager.ApplicationInsights.Models;
 using Azure.ResourceManager.OperationalInsights.Models;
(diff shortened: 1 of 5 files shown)


playground/cdk/CdkSample.AppHost/Program.cs
```diff
@@ -26,9 +27,15 @@
...
 var signaturesecret = builder.AddParameter("signaturesecret");
-var keyvault = builder.AddAzureKeyVault("mykv", (_, construct, keyVault) =>
+var keyvault = builder.AddAzureKeyVault("mykv", (def) =>
 {
-    var secret = new KeyVaultSecret(construct, name: "mysecret");
+    var secret = new KeyVaultSecret(def.Construct, def.KeyVault, name: "mysecret");
+    secret.AssignProperty(x => x.Properties.Value, signaturesecret);
+});
+
+var secureKeyVault = builder.AddAzureKeyVault<MoreSecureAzureKeyVaultAzureKeyVaultDefinition>("securekv", (def) =>
```
mitchdenny, line 36: This invocation of AddAzureKeyVault has a different definition type. The definition is responsible for controlling the shape of the resources as they get defined. It is a generic overload so that any custom properties defined can be got at (there are one in this simple exammple).

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
```diff
@@ -35,22 +34,30 @@ public static IResourceBuilder<AzureKeyVaultResource> AddAzureKeyVault(this IDis
...
-            keyVault.Properties.Tags["aspire-resource-name"] = construct.Resource.Name;
-
-            var keyVaultAdministratorRoleAssignment = keyVault.AssignRole(RoleDefinition.KeyVaultAdministrator);
-            keyVaultAdministratorRoleAssignment.AssignProperty(x => x.PrincipalId, construct.PrincipalIdParameter);
-            keyVaultAdministratorRoleAssignment.AssignProperty(x => x.PrincipalType, construct.PrincipalTypeParameter);
-
             var resource = (AzureKeyVaultResource)construct.Resource;
             var resourceBuilder = builder.CreateResourceBuilder(resource);
-            configureResource?.Invoke(resourceBuilder, construct, keyVault);
+            var definition = (TDefinition)Activator.CreateInstance(typeof(TDefinition), construct, resourceBuilder)!;
```
davidfowl, line 58: I think we should explore making this a static interface method instead of a constructor.

mitchdenny: Yeah ... was just trying to figure out what I should call it :)

Issue #795, opened 2023-11-13
Add Qdrant Aspire Component


eerhardt:
We should make an Aspire Component for working with a Qdrant vector database.

See https://www.nuget.org/packages/Qdrant.Client/.

We should make sure that the Aspire Qdrant Component integrates well with [Semantic Kernel](https://github.com/microsoft/semantic-kernel). See also https://www.nuget.org/packages/Microsoft.SemanticKernel.Connectors.Memory.Qdrant/

glennc:
We should have a resource for Qdrant as well, should we have a separate issue for that or merge them into one?

davidfowl:
They typically happen at the same time

timheuer:
I actually started on a qdrant resource one last night...I can contribute maybe a starting point...

davidfowl:
You're a dev now @timheuer

glennc:
I have created a couple of throwaway ones for this as well @timheuer happy to look over yours and get it together :)

timheuer:
> I have created a couple of throwaway ones for this as well @timheuer happy to look over yours and get it together :)

PR: https://github.com/dotnet/aspire/pull/3131

eerhardt:
Completed with https://github.com/dotnet/aspire/pull/3131.

Issue #1765, opened 2024-01-22
Add support for Connection Properties


sebastienros:
As discussed [here](https://github.com/dotnet/aspire/pull/1475#discussion_r1445146076) we should not create custom connection string formats for the sake of providing connection information if the Azure Portal and Azure SDK don't support it.

Instead we agreed to add a new "section" in configuration to define custom properties that can contain separate values like keys and other non-secret information.

eerhardt:
@davidfowl @mitchdenny - do we think this is going to happen before the 8.0 GA of Aspire?

davidfowl:
AFAIK no.

maddymontaquila:
We went all in on connection strings so  not planning this

davidfowl:
RIP

PR #3131, opened 2024-03-24
Initial work for Qdrant resource and component

timheuer:
Initial proposal for #795

This PR adds a Resource and initial Component (plus playground and tests)

Resource exposes two API access endpoints (REST + gRPC). The QdrantClient for .NET uses gRPC by default but Semantic Kernel does not use that client library so the other exposed endpoint is helpful.

REST endpoint by default exposes a Web UI dashboard -- this is excluded in Publish (per documentation in source, confirmed with Qdrant contributors).

Endpoints are secred non-optionally with an API Key (using ParameterResource -- will generate random if not provided).

Component used QdrantClient .NET version (recommended from Qdrant). This component uses gRPC endpoint by default. Settings will read from standard env vars for ConnectionString + API key or from config for the component itself if provided.

The component doesn't seem to be rich in trace/metrics so those are not implemented...ideas?

Little runtime demo with playground+dashboard apps:

https://github.com/dotnet/aspire/assets/4821/eec32fbc-7096-4485-bdbb-27441f76a094

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3131)


playground/Qdrant/Qdrant.ServiceDefaults/Extensions.cs
```diff
@@ -0,0 +1,126 @@
+using Microsoft.AspNetCore.Builder;
```
davidfowl, line 1: The playground samples share a service defaults.

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// </summary>
+    /// <param name="builder">An <see cref="IResourceBuilder{T}"/> for <see cref="ProjectResource"/></param>
+    /// <param name="qdrantResource">The Qdrant server resource</param>
+    /// <returns></returns>
+    public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuilder<ProjectResource> builder, IResourceBuilder<QdrantServerResource> qdrantResource)
+    {
+        builder.WithEnvironment(context =>
+        {
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = qdrantResource.Resource.ConnectionStringExpression;
+            context.EnvironmentVariables[$"Parameters__{qdrantResource.Resource.ApiKeyParameter.Name}"] = qdrantResource.Resource.ApiKeyParameter.Value;
```
davidfowl, line 95: Why `Parameters__`? This should be written into a component specific config section. Also you should directly set the parameter, not the parameter value.

e.g.:

```
```suggestion
            context.EnvironmentVariables[$"Qdrant__ApiKey"] = qdrantResource.Resource.ApiKeyParameter;
```

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// Add a reference to a Qdrant settings for a project.
+    /// </summary>
+    /// <param name="builder">An <see cref="IResourceBuilder{T}"/> for <see cref="ProjectResource"/></param>
+    /// <param name="qdrantResource">The Qdrant server resource</param>
+    /// <returns></returns>
+    public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuilder<ProjectResource> builder, IResourceBuilder<QdrantServerResource> qdrantResource)
+    {
+        builder.WithEnvironment(context =>
+        {
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = qdrantResource.Resource.ConnectionStringExpression;
```
davidfowl, line 94:
```suggestion
            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = new ConnectionStringReference(qdrantResource.Resource, optional: false);
```

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+            context.EnvironmentVariables[$"Parameters__{qdrantResource.Resource.ApiKeyParameter.Name}"] = qdrantResource.Resource.ApiKeyParameter.Value;
+
+            var restEndpointUrl = qdrantResource.Resource.GetEndpoint("rest")?.Url;
+            if (restEndpointUrl != null)
+            {
+                context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = restEndpointUrl;
+            }
```
davidfowl, line 101: We're missing an API here, ideally this would be:

```suggestion
            var restEndpointUrl = qdrantResource.Resource.GetEndpoint("rest");
            if (restEndpointUrl.Exists)
            {
                context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = restEndpointUrl;
            }
```

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        if (builder.Configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            settings.ConnectionString = connectionString;
+        }
+
+        if (builder.Configuration[$"Parameters:{connectionName}-ApiKey"] is string apiKey)
+        {
+            settings.ApiKey = apiKey;
+        }
```
davidfowl, line 69: This component needs a schema, it shouldn't use parameters, but instaed it should use something like `Qdrant:{connectionName}:ApiKey`

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
timheuer: Certainly `Qdrant__XXXX` is better.  For component-specific sectio though, since this is named resource value, were you thinking this would actually be `Aspire__Qdrant__[name]__ApiKey`?

timheuer: Actually, looking at other components, is this correct? Looking at MySQL for example, there would be one Aspire:MySqlConnectorSettings:ConnectionString value -- if I have 2 MySQL resources, this would be shared between them as the component config is not per-named resource...correct? /cc @eerhardt

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
timheuer: @davidfowl it has a schema, see my comment on the other PR feedback as well -- agreed a better named one is better, but the component config schema's don't account for per-named resources (and their typed settings class)

playground/Qdrant/Qdrant.ServiceDefaults/Extensions.cs
timheuer: Fixed in next update

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
timheuer: Fixed in next update

roji:
Really nice to see this!

> The QdrantClient for .NET uses gRPC by default but Semantic Kernel does not use that client library so the other exposed endpoint is helpful.

FWIW reimplementing the SK connector over the SDK is something that's on my TODO list... that would make the SK connector more efficient thanks to gRPC etc.

> The component doesn't seem to be rich in trace/metrics so those are not implemented...ideas?

Can you point out what trace/metrics in e.g. the Python client look like? This may be something we can improve.

timheuer:
> FWIW reimplementing the SK connector over the SDK is something that's on my TODO list... that would make the SK connector more efficient thanks to gRPC etc.

This is great to hear @roji! For now the Resource will expose both endpoints so the SK SDK as-is can get the `ConnectionString` REST endpoint to use in the connector.

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
eerhardt: 2 thoughts here:

1. For other components that look like this (need a URL and an API Key), we've defined a connection string format ourselves. See 
https://github.com/dotnet/aspire/blob/0d3d68e0d59c1e45d36484761f1516bcfc6e1bbd/src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs#L50-L72

https://github.com/dotnet/aspire/blob/0d3d68e0d59c1e45d36484761f1516bcfc6e1bbd/src/Components/Aspire.Azure.Search.Documents/AzureSearchSettings.cs#L58-L80

I'm not sure we should invent yet another pattern for this component.

2. IMO, the correct fix/design here is https://github.com/dotnet/aspire/issues/1765.

src/Components/Aspire.Qdrant/Aspire.Qdrant.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <PackageIconFullPath>$(SharedDir)QdrantLogo_256x.png</PackageIconFullPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\Common\HealthChecksExtensions.cs" Link="HealthChecksExtensions.cs" />
+    <Compile Include="..\Common\ConfigurationSchemaAttributes.cs" Link="ConfigurationSchemaAttributes.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Qdrant.Client" />
```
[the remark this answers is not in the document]
timheuer: Fixed, plumed through to config

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
timheuer: What is the suggested change here? In this case the ApiKey is not a part of the connection string, it's passed as an arg to the client lib.

eerhardt: > In this case the ApiKey is not a part of the connection string, it's passed as an arg to the client lib.

That is no different than the 2 scenarios I linked to above. Here's where the ApiKeys get passed as an arg to the client lib for them:

https://github.com/dotnet/aspire/blob/126b1855ba08d7670747d59d2a8048d36d1c1c99/src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIExtensions.cs#L118-L119

https://github.com/dotnet/aspire/blob/126b1855ba08d7670747d59d2a8048d36d1c1c99/src/Components/Aspire.Azure.Search.Documents/AspireAzureSearchExtensions.cs#L114

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Gets the primary endpoint for the Qdrant database.
+    /// </summary>
+    public EndpointReference PrimaryEndpoint => _primaryEndpoint ??= new(this, PrimaryEndpointName);
+
+    /// <summary>
+    /// Gets the connection string expression for the Qdrant database.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression =>
+       ReferenceExpression.Create(
+            $"http://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
```
eerhardt, line 41: > http://

This seems wrong. I don't think we should be hard-coding the scheme in the Expression.

davidfowl: that's right, the scheme is part of the endpoint reference:

PrimaryEndpoint.Property(EndpointProperty.Scheme)

timheuer: @eerhardt How do we handle resources that don't have a TLS certificate by default? That's the situation here. https://qdrant.tech/documentation/guides/security/#tls

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
timheuer: Should IConnectionStrings move more generically from Azure.Component?

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
davidfowl: We can't do https here but it should use the scheme defined by the endpoint.

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
eerhardt: > Should IConnectionStrings move more generically from Azure.Component?

No, the only reason it exists is because the base `AzureComponent` needs to call it on the Settings object. In your case, the `AspireQdrantExtensions` class can just call an internal method on the `QdrantSettings` class.

timheuer: Fixed in latest update

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
timheuer: changed to scheme already in latest PR feedback.  As an aside, SSL could be enabled using dotnet dev-certs but felt hacky (but got it working).  Anyhow, PR update to use scheme

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Gets the primary endpoint for the Qdrant database.
+    /// </summary>
+    public EndpointReference PrimaryEndpoint => _primaryEndpoint ??= new(this, PrimaryEndpointName);
+
+    /// <summary>
+    /// Gets the connection string expression for the Qdrant database.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression =>
+       ReferenceExpression.Create(
+            $"Endpoint={PrimaryEndpoint.Scheme}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)};Key={ApiKeyParameter}");
```
davidfowl, line 41:
```suggestion
            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Scheme)}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)};Key={ApiKeyParameter}");
```

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <param name="qdrantResource">The Qdrant server resource</param>
+    /// <returns></returns>
+    public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuilder<ProjectResource> builder, IResourceBuilder<QdrantServerResource> qdrantResource)
+    {
+        builder.WithEnvironment(context =>
+        {
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = new ConnectionStringReference(qdrantResource.Resource, optional: false);
+
+            var restEndpointUrl = $"Endpoint={qdrantResource.Resource.GetEndpoint("rest").Url};Key={qdrantResource.Resource.ApiKeyParameter.Value}";
+            if (restEndpointUrl is not null)
```
mitchdenny, line 90: This will never be null?

davidfowl: yeah we need a new API for this.

timheuer: I realize I'm checking the wrong thing for null, but is it really that `.Exists()` is just going to be a null check anyway?

```
var restEndpoint = qdrantResource.Resource.GetEndpoint("rest");
if (restEndpoint is not null) //Exists?
{
    context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = $"Endpoint={restEndpoint.Url};Key={qdrantResource.Resource.ApiKeyParameter.Value}";
}
```

or more a `TryGetEndpoint`?

davidfowl: it would make sure there's a backing endpoint annotation. Since this change is in main, maybe we just expose it in this PR.

timheuer: It's possible for an `EndpointReference` on a `Resource` to not have annotations?!

davidfowl: The EndpointReference is a promise so it never returns null. You get the reference and asks questions about it 😄 .

timheuer: Wait, so is it already there? `IsAllocated` seems to validate the annotation exists...

davidfowl: Right that's if the endpoint is allocated, I guess you can use that for this scenario. Exists is different. Exists says do I have the EndpointAnnotation. IsAllocated tells you if the endpoint is allocated by DCP. In publish mode, none of the endpoints are allocated, but they all exist.

davidfowl: Wait is this endpoint ever not defined?

timheuer: No it always is defined.  The resource is created always with 2 endpoints, primary and this named one.
```
return builder.AddResource(qdrant)
    .WithImage(QdrantContainerImageTags.Image, QdrantContainerImageTags.Tag)
    .WithHttpEndpoint(hostPort: port, containerPort: QdrantPortGrpc, name: QdrantServerResource.PrimaryEndpointName)
    .WithHttpEndpoint(hostPort: port, containerPort: QdrantPortHttp, name: "rest")

// other parts snipped
```

Maybe that makes you feel better about the simpler check?

But if not, is this what you had in mind for the Exists? (virtually same as Allocated, just not checking the Allocated check?

```
public bool Exists => GetAnnotation() is not null;

private EndpointAnnotation GetAnnotation()
{
    var endpoint = _endpointAnnotation ??= Resource.Annotations.OfType<EndpointAnnotation>().SingleOrDefault(a => StringComparers.EndpointAnnotationName.Equals(a.Name, EndpointName));
    ArgumentNullException.ThrowIfNull(endpoint, $"The endpoint `{EndpointName}` does not exist for the resource `{Resource.Name}`.");
    return endpoint;
}
```

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -84,13 +84,12 @@ public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuil
...
             context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = new ConnectionStringReference(qdrantResource.Resource, optional: false);
 
-            var restEndpointUrl = $"Endpoint={qdrantResource.Resource.GetEndpoint("rest").Url};Key={qdrantResource.Resource.ApiKeyParameter.Value}";
-            if (restEndpointUrl is not null)
-            {
-                context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = restEndpointUrl;
-            }
+            // REST endpoint
+            var restEndpoint = qdrantResource.Resource.GetEndpoint("rest");
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = $"Endpoint={restEndpoint?.Url};Key={qdrantResource.Resource.ApiKeyParameter}";
```
davidfowl, line 92:
```suggestion
            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = $"Endpoint={restEndpoint};Key={qdrantResource.Resource.ApiKeyParameter}";
```

davidfowl: My bad! 
```suggestion
            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = ReferenceExpression.Create($"Endpoint={restEndpoint?.Url};Key={qdrantResource.Resource.ApiKeyParameter}");
```

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+    public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuilder<ProjectResource> builder, IResourceBuilder<QdrantServerResource> qdrantResource)
+    {
+        builder.WithEnvironment(context =>
+        {
+            // primary endpoint (gRPC)
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = new ConnectionStringReference(qdrantResource.Resource, optional: false);
+
+            // REST endpoint
+            var restEndpoint = qdrantResource.Resource.GetEndpoint("rest");
+            context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = ReferenceExpression.Create($"Endpoint={restEndpoint};Key={qdrantResource.Resource.ApiKeyParameter}");
```
davidfowl, line 92: One way to make this a bit cleaner is to make it a property on the resource and just use it here.

eerhardt, line 92: Who reads this `_rest` environment variable?

tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs
```diff
@@ -0,0 +1,220 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
eerhardt, line 1: We should also have component tests. See the other components for examples.

It doesn't look like there is a Qdrant test container (yet). https://github.com/testcontainers/testcontainers-dotnet/issues/992. But we can make the tests before that, and just have them check if a connection can be made and skip if it can't. Then you can run the tests with a local Qdrant container running.

src/Components/Aspire.Qdrant/README.md
```diff
@@ -0,0 +1,107 @@
+# Aspire.Qdrant.Client library
```
eerhardt, line 1: Can you also add entries for this component in:

* https://github.com/dotnet/aspire/blob/main/src/Components/Aspire_Components_Progress.md
* https://github.com/dotnet/aspire/blob/main/src/Components/Telemetry.md

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+        else
+        {
+            builder.Services.AddKeyedSingleton(serviceKey, (sp, key) => ConfigureQdrant());
+        }
+
+        QdrantClient ConfigureQdrant()
+        {
+            if (settings.Endpoint is not null)
+            {
+                return new QdrantClient(settings.Endpoint, apiKey: settings.Key);
```
eerhardt, line 86: Is there any sort of logging, tracing, metrics, health checks that can be enabled?

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+    public static IResourceBuilder<QdrantServerResource> WithDataBindMount(this IResourceBuilder<QdrantServerResource> builder, string source, bool isReadOnly = false)
+        => builder.WithBindMount(source, "/qdrant/storage", isReadOnly);
+
+    /// <summary>
+    /// Add a reference to a Qdrant settings for a project.
+    /// </summary>
+    /// <param name="builder">An <see cref="IResourceBuilder{T}"/> for <see cref="ProjectResource"/></param>
+    /// <param name="qdrantResource">The Qdrant server resource</param>
+    /// <returns></returns>
+    public static IResourceBuilder<ProjectResource> WithReference(this IResourceBuilder<ProjectResource> builder, IResourceBuilder<QdrantServerResource> qdrantResource)
```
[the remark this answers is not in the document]
davidfowl: Good catch!

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        if (builder.Configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            settings.ParseConnectionString(connectionString);
+        }
+
+        if (builder.Configuration[$"{DefaultConfigSectionName}:{connectionName}:Key"] is string apiKey)
+        {
+            settings.Key = apiKey;
+        }
```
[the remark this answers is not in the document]
timheuer: What if I want to override the key value.

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
timheuer: As noted in the original PR, Qdrant exposes two endpoints for access: gRPC (which the client component uses by default), and a REST endpoint (btw, I'm using their terms).  The second is currently what you'd use to compose with SK (although Shay is hoping to change that but not there yet) *and also* is where the Qdrant management web UI is exposed (note: this is disabled in publish mode).

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
timheuer: On the client, not yet. I was looking to actually wire up the metrics for the server _via_ the client, but not doing that ;-).

Shay is hopefully going to get some time to work on adding Tracing but it's just not in yet. Tracking: https://github.com/qdrant/qdrant-dotnet/issues/49

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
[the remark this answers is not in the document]
timheuer: Yep, will change -- FYI I noticed this is in AWS Hosting package a well.

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
eerhardt: It looks like Logging is supported.

https://github.com/qdrant/qdrant-dotnet/blob/b3b80fddbb82bc36c9de9ea7100da93b3624c742/src/Qdrant.Client/QdrantClient.cs#L59-L63

tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs
roji: Yeah, FYI I was in touch recently with @russcam about reviving that PR, hopefully he'll have a bit of time soon.

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
timheuer: Like this?

Resource:
    public ReferenceExpression RestConnectionStringExpression =>
        ReferenceExpression.Create(
            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Scheme)}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}_rest;Key={ApiKeyParameter}");

WithReference overload:
// primary endpoint (gRPC)
context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}"] = qdrantResource.Resource.ConnectionStringExpression;

// REST endpoint
context.EnvironmentVariables[$"ConnectionStrings__{qdrantResource.Resource.Name}_rest"] = qdrantResource.Resource.RestConnectionStringExpression;

src/Components/Aspire.Qdrant/AspireQdrantExtensions.cs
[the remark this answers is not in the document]
timheuer: ahh, i see what you mean, you aren't saying remove it from a config, just remove this logic here.  If so, ack.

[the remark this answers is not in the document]
eerhardt: Yep. just those 5 lines of code can be deleted. The rest looks right to me.

timheuer:
@eerhardt I think I addressed most of your feedback (thanks for the pointers on logging).

tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs
russcam: I've updated the PR and requested the maintainers to reopen it (I don't have ability to reopen it- guessing there's some repo setting preventing this).

mitchdenny:
On the hosting side this is looking pretty good. I'm wondering whether when we deploy to ACA we should favor the use of the dev services (it supports quadrant). If that is the case then perhaps there should be an Aspire.Hosting.Azure.Qdrant package?

davidfowl:
We don’t favor any of the aca services anymore so I would day no. Not as part of this PR and we’d do something like redis first (which requires container app cdk support).

Might be worth spiking PublishAsAcaRedis or something similar

src/Shared/QdrantLogo_256x.png
glennc, line 1: Are there any permission gathering we need to do in order to include the logo? Are we inline with their usage guides?

timheuer: Yes already did -- put the disclaimer of use on the README

eerhardt (commented):
This is looking really great. Thanks for the contribution @timheuer!

I pushed some minor changes and fix ups to the PR.

If you are ready to take this out of "Draft", I think it can be approved. The one remaining comment I have can be addressed in a follow up PR if we think we should do it.

playground/Qdrant/Qdrant.ApiService/Program.cs
```diff
@@ -0,0 +1,95 @@
...
+                ["title"] = "Toy Story"
+            }
+        },
+        new PointStruct
+        {
+            Id = 4,
+            Vectors = new [] {0.106827796f, 0.21676421f},
+            Payload =
+            {
+                ["title"] = "Pulp Function"
```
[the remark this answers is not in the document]
timheuer: From a sample #shrug

eerhardt (approved):
Thanks again for the contribution!

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
```diff
@@ -0,0 +1,50 @@
...
+    public ReferenceExpression ConnectionStringExpression =>
+       ReferenceExpression.Create(
+            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Scheme)}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)};Key={ApiKeyParameter}");
+
+    /// <summary>
+    /// Gets the connection string expression for the Qdrant REST endpoint.
+    /// </summary>
+    public ReferenceExpression RestConnectionStringExpression =>
+        ReferenceExpression.Create(
+            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Scheme)}://{PrimaryEndpoint.Property(EndpointProperty.Host)}:6333;Key={ApiKeyParameter}");
```
davidfowl, line 49: Why is this port hard coded?

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+        string name,
+        IResourceBuilder<ParameterResource>? apiKey = null,
+        int? port = null)
+    {
+        var apiKeyParameter = apiKey?.Resource ??
+            ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-Key", special: false);
+        var qdrant = new QdrantServerResource(name, apiKeyParameter);
+        return builder.AddResource(qdrant)
+            .WithImage(QdrantContainerImageTags.Image, QdrantContainerImageTags.Tag)
+            .WithHttpEndpoint(port: port, targetPort: QdrantPortGrpc, name: QdrantServerResource.PrimaryEndpointName)
```
davidfowl, line 42: Using the same port for grpc and http is not correct. Specifying a port will clash.

src/Aspire.Hosting.Qdrant/QdrantServerResource.cs
timheuer: Converting this feedback to a bug since this PR merged FYI @eerhardt

eerhardt: Resolved this with https://github.com/dotnet/aspire/pull/3422

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
eerhardt: Resolved this with https://github.com/dotnet/aspire/pull/3422

PR #3132, opened 2024-03-25
Make reading Kubeconfig from DCP resilient

mitchdenny:
This PR adds some retry logic to `EnsureKubernetes`. As part of the process the method has become async and I'm making use of a semaphore to stop the stampede to read the file it'll only read once whilst the other threads slowly spin if they get caught up in the semaphore.


src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +284,54 @@ public void Dispose()
...
+            // being written. We rely on DCP to write the configuration file but it may happen in parallel to
+            // this code executing is its possible the file does not exist, or is still being written by
+            // the time we get to it.
+            //
+            // This retry will retry reading the file 5 times (by default, but configurable) with a pause
+            // of 3 seconds between each attempt. This means it could take up to 15 seconds to fail. We emit
+            // debug level logs for each retry attempt should we need to help a customer debug this.
+            var configurationReadRetry = new RetryStrategyOptions()
+            {
+                ShouldHandle = new PredicateBuilder().Handle<IOException>(e => e.Message.StartsWith("The process cannot access the file")),
```
davidfowl, line 311: Looking at the exception text feels too specific. I think handling the IO exception is just fine.

mitchdenny: Ended up changing the code a fair bit to improve debugability and ended up catching a k8s client specific exception.

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+        if (_kubernetes != null)
+        {
+            logger.LogDebug("Kubernetes ensured at first chance shortcut.");
+            return;
+        }
+
+        while (!await _kubeconfigReadSemaphore.WaitAsync(100, cancellationToken).ConfigureAwait(false))
+        {
+            logger.LogDebug("Waiting for semaphore to read kubeconfig.");
+        }
```
davidfowl, line 302: Why the 100ms timeout?

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+            //
+            // This retry will retry reading the file 5 times (by default, but configurable) with a pause
+            // of 3 seconds between each attempt. This means it could take up to 15 seconds to fail. We emit
+            // debug level logs for each retry attempt should we need to help a customer debug this.
+            var configurationReadRetry = new RetryStrategyOptions()
+            {
+                ShouldHandle = new PredicateBuilder().Handle<KubeConfigException>(),
+                BackoffType = DelayBackoffType.Constant,
+                MaxRetryAttempts = dcpOptions.Value.KubernetesConfigReadRetryCount,
+                MaxDelay = TimeSpan.FromSeconds(dcpOptions.Value.KubernetesConfigReadRetryIntervalSeconds),
```
davidfowl, line 333: Compute this once

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+            //
+            // It looks like it is possible for us to attempt to read the file before it is written/finished
+            // being written. We rely on DCP to write the configuration file but it may happen in parallel to
+            // this code executing is its possible the file does not exist, or is still being written by
+            // the time we get to it.
+            //
+            // This retry will retry reading the file 5 times (by default, but configurable) with a pause
+            // of 3 seconds between each attempt. This means it could take up to 15 seconds to fail. We emit
+            // debug level logs for each retry attempt should we need to help a customer debug this.
+            var configurationReadRetry = new RetryStrategyOptions()
```
davidfowl, line 328: Can't we create this object once?

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+                    logger.LogDebug(
+                        retry.Outcome.Exception,
+                        "Reading Kubernetes configuration file from '{DcpKubeconfigPath}' failed. Retry pending. (iteration {Iteration}).",
+                        locations.DcpKubeconfigPath,
+                        retry.AttemptNumber
+                        );
+                    return ValueTask.CompletedTask;
+                }
+            };
+            var pipeline = new ResiliencePipelineBuilder().AddRetry(configurationReadRetry).Build();
```
davidfowl, line 345: Same here, we only need to do this once right?

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+            // this code executing is its possible the file does not exist, or is still being written by
+            // the time we get to it.
+            //
+            // This retry will retry reading the file 5 times (by default, but configurable) with a pause
+            // of 3 seconds between each attempt. This means it could take up to 15 seconds to fail. We emit
+            // debug level logs for each retry attempt should we need to help a customer debug this.
+            var configurationReadRetry = new RetryStrategyOptions()
+            {
+                ShouldHandle = new PredicateBuilder().Handle<KubeConfigException>(),
+                BackoffType = DelayBackoffType.Constant,
```
davidfowl, line 331: Not exponential?

karolz-ms:
I think a better fix would be to address this on the DCP side. Namely, produce the `kubeconfig` file with the following sequence:

1. Create a temporary file
2. Write the content and close the file
3. Rename the file so that it matches the name the readers expect.

This should be a pretty easy fix on DCP side

karolz-ms:
Had a brief chat with @davidfowl , working on a DCP fix for `kubeconfig` access conflict.

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Just so we don't have a bunch of threads spinning in a tight loop.

I wanted to log that some spinning was occurring to make diagnostics easier hence why I didn't just specify -1

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
 
-            var config = KubernetesClientConfiguration.BuildConfigFromConfigFile(kubeconfigPath: locations.DcpKubeconfigPath, useRelativePaths: false);
-            _kubernetes = new DcpKubernetesClient(config);
+        try
+        {
+            // This retry was created in relation to this comment in GitHub:
+            //
+            //     https://github.com/dotnet/aspire/issues/2422#issuecomment-2016701083
+            //
+            // It looks like it is possible for us to attempt to read the file before it is written/finished
```
karolz-ms, line 320: In Aspire P5 this won't be true anymore.

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: I think non exponential is desirable here. Remember this is the retry loop which only one thread should be in at a time, and its a 3 second interval. i would expect only one retry to be necessary, but we allow up to 10 before we blow up.

mitchdenny: .. these are also configurable via environment variables as an escape hatch.

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Moved into a helper method which caches the entire pipeline.

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Moved into a helper method which caches the pipeline.

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+            // being written. We rely on DCP to write the configuration file but it may happen in parallel to
+            // this code executing is its possible the file does not exist, or is still being written by
+            // the time we get to it.
+            //
+            // This retry will retry reading the file 5 times (by default, but configurable) with a pause
```
karolz-ms, line 325: Given that the whole startup sequence is blocked on being able to communicate with DCP, waiting 3 seconds between attempts is excessive. After the first attempt it is very likely that the relevant portions of the directory contents will be in OS cache and the check for file existence will be very cheap. I would probe repeatedly on a constant 100-200 ms interval.

src/Aspire.Hosting/Dcp/KubernetesService.cs
karolz-ms: Consider an ETW event for this.

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +285,78 @@ public void Dispose()
...
+            // debug level logs for each retry attempt should we need to help a customer debug this.
+            var configurationReadRetry = new RetryStrategyOptions()
+            {
+                ShouldHandle = new PredicateBuilder().Handle<KubeConfigException>(),
+                BackoffType = DelayBackoffType.Constant,
+                MaxRetryAttempts = dcpOptions.Value.KubernetesConfigReadRetryCount,
+                MaxDelay = TimeSpan.FromSeconds(dcpOptions.Value.KubernetesConfigReadRetryIntervalSeconds),
+                OnRetry = (retry) =>
+                {
+                    logger.LogDebug(
```
karolz-ms, line 336: Consider keeping the stats and just log the outcome. E.g. took x milliseconds, probed y times, before succeeding/failing. The logs for each individual attempt aren't that interesting.

karolz-ms (dismissed):
Overall I like the direction, my main concern is super low frequency of probing for the file in this attempt.

karolz-ms:
> We have to do this both on the DCP side and here. We have to add the retry logic here because we are looking for it before DCP has put the file down on disk. What you are proposing for DCP will improve things as well.

Makes sense. I gave you some suggestions, HTH!

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Still logging but less noisy by dropping the exception. On my machine I seem to go through the loop 6-8 times (fairly beefy desktop machine). As a result of the reduced interval I increased the number of attempts we'll make before giving up.

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Probing every 100ms now

src/Aspire.Hosting/Dcp/KubernetesService.cs
mitchdenny: Tidied this comment up.

mitchdenny:
@karolz-ms could you please reset your vote?

src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -280,16 +286,80 @@ public void Dispose()
...
 
-        lock (Model.Dcp.Schema)
+        await _kubeconfigReadSemaphore.WaitAsync(-1, cancellationToken).ConfigureAwait(false);
+
+        if (_kubernetes != null)
         {
-            if (_kubernetes != null) { return; }
+            _kubeconfigReadSemaphore.Release();
+            return;
+        }
```
davidfowl, line 334: I would put this in the try.

davidfowl (approved):
Small nit

PR #3139, opened 2024-03-25
Fix key store config binding

drewnoakes:
This was missed in #3033.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3139)

### DIFF
--- a/src/Aspire.Dashboard/Model/DashboardClient.cs
+++ b/src/Aspire.Dashboard/Model/DashboardClient.cs
@@ -171,9 +171,11 @@ internal sealed class DashboardClient : IDashboardClient
                     throw new InvalidOperationException("ResourceServiceClient:ClientCertificate:Source is \"KeyStore\", but no Certificate:FilePath is configured.");
                 }
 
-                using var store = new X509Store(storeName: StoreName.My, storeLocation: StoreLocation.CurrentUser);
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

JamesNK (dismissed):
As far as I can tell this still won't do something useful. The properties on [X509Store](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509store?view=net-8.0) are read only. I've never looked at the internal code of configuration binding but I'm guessing it sets properties with reflection. If there are no setters, then it can't do anything.

Name and location need to be specified in the [constructor](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509store.-ctor?view=net-8.0#system-security-cryptography-x509certificates-x509store-ctor(system-string-system-security-cryptography-x509certificates-storelocation)).

drewnoakes:
You're right. I'll push an update.

davidfowl:
I spoke to @drewnoakes about a follow up to avoid using this type, but if its broken now then we should do that work here.

drewnoakes:
Added a type to bind config to, then populate the `KeyStore` from that type. PTAL.

PR #3152, opened 2024-03-25
Ensure RabbitMQ WithDataVolume/WithDataBindMount methods actually enable persistent storage

### LINKED ISSUE #2247: AddRabbitMQ method with volume mounts does not persist queues across restarts by default
RabbitMQ uses its HOSTNAME to set up paths for queues on its internal storage. As the DCP generates a random hostname across restarts, this changes with every invocation of the container. This results in unexpected behavior from the developer's point of view - queues are not persisted across restarts even with persistent storage configured for the container.

This is an example of a statement that will NOT persist queues:

```
 var rabbitMq = builder
     .AddRabbitMQContainer("rabbitmq-host", 5672, password: rabbitMqPassword)
     .WithVolumeMount("rabbitmq-host-vol", "/var/lib/rabbitmq", VolumeMountType.Named);
```

If a NODENAME environment variable is set, RabbitMQ uses that to override the HOSTNAME on the container for internal configuration purposes. So directly specifiying the NODENAME variable With i.e. `WithEnvironment("NODENAME", "rabbit@localhost")` resolves the issue, as in the following statement:

```
 var rabbitMq = builder
     .AddRabbitMQContainer("rabbitmq-host", 5672, password: rabbitMqPassword)
     .WithEnvironment("NODENAME", "rabbit@localhost")
     .WithVolumeMount("rabbitmq-host-vol", "/var/lib/rabbitmq", VolumeMountType.Named);
```

Suggestion : set NODENAME by default to a fixed value for RabbitMQ hosts (it's only used internally, so won't affect other RabbitMQ containers running in the same docker engine) - but still allow the user to override by setting it specifically like above. Alternatively, as a minimum, update the RabbitMQ container documentation for Aspire to mention this specifically.

DamianEdwards:
This updates the `WithDataVolume` and `WithDataBindMount` methods for RabbitMQ to set the `RABBITMQ_NODENAME` environment variable to a stable value so that RabbitMQ's internal storage is laid out with a stable storage directory. I also updated TestShop so that it uses persistent storage for its RabbitMQ messaging. I'm able to run TestShop with the `orderprocessor` resource commented out and make some "orders" by added items to the basket and then clicking the basket summary icon, then stop the session, uncomment `orderprocessor`, and start a new session, and see that `orderprocessor` processes the order messages from the previous session.

Fixes #2247
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3152)

### DIFF
--- a/playground/TestShop/AppHost/Program.cs
+++ b/playground/TestShop/AppHost/Program.cs
@@ -13,7 +13,7 @@ var catalogService = builder.AddProject<Projects.CatalogService>("catalogservice
                             .WithReplicas(2);
 
 var rabbitMqPassword = builder.AddParameter("rabbitmq-password", secret: true);
-var messaging = builder.AddRabbitMQ("messaging", password: rabbitMqPw)
+var messaging = builder.AddRabbitMQ("messaging", password: rabbitMqPassword)
                        .WithDataVolume()
                        .PublishAsContainer();
 
--- a/playground/TestShop/AppHost/Program.cs
+++ b/playground/TestShop/AppHost/Program.cs
@@ -12,7 +12,7 @@ var catalogService = builder.AddProject<Projects.CatalogService>("catalogservice
                             .WithReference(catalogDb)
                             .WithReplicas(2);
 
-var rabbitMqPw = builder.AddParameter("rabbitmqpw", secret: true);
+var rabbitMqPassword = builder.AddParameter("rabbitmq-password", secret: true);
 var messaging = builder.AddRabbitMQ("messaging", password: rabbitMqPw)
                        .WithDataVolume()
                        .PublishAsContainer();
--- a/playground/TestShop/AppHost/Program.cs
+++ b/playground/TestShop/AppHost/Program.cs
@@ -12,7 +12,10 @@ var catalogService = builder.AddProject<Projects.CatalogService>("catalogservice
                             .WithReference(catalogDb)
                             .WithReplicas(2);
 
-var messaging = builder.AddRabbitMQ("messaging").PublishAsContainer();
+var rabbitMqPw = builder.AddParameter("rabbitmqpw", secret: true);
+var messaging = builder.AddRabbitMQ("messaging", password: rabbitMqPw)
+                       .WithDataVolume()
+                       .PublishAsContainer();
 
 var basketService = builder.AddProject("basketservice", @"..\BasketService\BasketService.csproj")
                            .WithReference(basketCache)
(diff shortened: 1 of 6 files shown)


src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -58,5 +60,21 @@ public static IResourceBuilder<RabbitMQServerResource> WithDataVolume(this IReso
     /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<RabbitMQServerResource> WithDataBindMount(this IResourceBuilder<RabbitMQServerResource> builder, string source, bool isReadOnly = false)
-        => builder.WithBindMount(source, "/var/lib/rabbitmq", isReadOnly);
+        => builder.WithBindMount(source, "/var/lib/rabbitmq", isReadOnly)
+                  .RunWithStableNodeName();
+
+    private static IResourceBuilder<RabbitMQServerResource> RunWithStableNodeName(this IResourceBuilder<RabbitMQServerResource> builder)
+    {
+        if (builder.ApplicationBuilder.ExecutionContext.IsRunMode)
```
davidfowl, line 68: What happens when you publish?

DamianEdwards: The assumption is when you publish the host name of the container is stable, rather than what happens in dev where we randomize the port every time.

eerhardt:
Is it possible to add some tests?

playground/TestShop/AppHost/Program.cs
```diff
@@ -12,7 +12,10 @@
                             .WithReference(catalogDb)
                             .WithReplicas(2);
 
-var messaging = builder.AddRabbitMQ("messaging").PublishAsContainer();
+var rabbitMqPw = builder.AddParameter("rabbitmqpw", secret: true);
```
eerhardt, line 15: (nit) `rabbitmqpw` is really hard to read/parse at first.

```suggestion
var rabbitMqPassword = builder.AddParameter("rabbitmq-password", secret: true);
```

playground/TestShop/AppHost/Program.cs
```diff
@@ -12,7 +12,10 @@
...
+var messaging = builder.AddRabbitMQ("messaging", password: rabbitMqPw)
+                       .WithDataVolume()
+                       .PublishAsContainer();
```
eerhardt, line 18: Why is `.PublishAsContainer();` necessary?

DamianEdwards: I didn't add that, it was there already.

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
DamianEdwards: More details about how RabbitMQ auto-assigns node names in their docs: https://www.rabbitmq.com/docs/clustering#node-names and https://www.rabbitmq.com/docs/configure

From the latter link:

> The node name should be unique per Erlang-node-and-machine combination. To run multiple nodes, see the [clustering guide](https://www.rabbitmq.com/docs/clustering).

> Default:
>   Unix: rabbit@$HOSTNAME
>   Windows: rabbit@%COMPUTERNAME%

So ultimately, what happens after publish depends on what `$HOSTNAME` results in on the container where RabbitMQ is running.

davidfowl (approved):

PR #3159, opened 2024-03-25
Detect conflicting settings when using EnrichEF

### LINKED ISSUE #2524: Enrich EF APIs overwrite an existing ExecutionStrategy by default
When someone calls `EnrichSqlServerDbContext` after they have configured their DbContext with an execution strategy, for example:

```C#
builder.Services.AddDbContext<MyDb1Context>(options =>
{
    options.UseNpgsql(npgsql =>
    {
        npgsql.ExecutionStrategy(c => new MyExecutionStrategy(c));
    });
});
builder.EnrichNpgsqlDbContext<MyDb1Context>();
```

We are overwriting the ExecutionStrategy by default (since settings.Retry is true by default)

https://github.com/dotnet/aspire/blob/c5f1b95332f4e462b2e608dfc006afca0f629661/src/Components/Aspire.Microsoft.EntityFrameworkCore.SqlServer/AspireSqlServerEFCoreSqlClientExtensions.cs#L111-L119

We should consider if this is the best behavior. Other options I can think of would be to detect if the ExecutionStrategy is already set, and log a warning.

cc @DamianEdwards

sebastienros:
Fixes https://github.com/dotnet/aspire/issues/2524
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3159)


src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
```diff
@@ -117,12 +121,52 @@ void ConfigureDbContext(DbContextOptionsBuilder dbContextOptionsBuilder)
...
+                        if (executionStrategy != null)
+                        {
+                            if (executionStrategy.GetType() == typeof(NpgsqlRetryingExecutionStrategy))
+                            {
+                                // Keep custom Retry strategy
+                            }
+                            else if (executionStrategy.GetType() != typeof(NpgsqlExecutionStrategy))
+                            {
+                                throw new InvalidOperationException($"{nameof(NpgsqlEntityFrameworkCorePostgreSQLSettings)}.Retry can't be set when a custom Execution Strategy is configured.");
+                            }
```
davidfowl, line 144: Can't use is?

sebastienros:
Just for Npgsql for now to validate the fix. I think the way the execution strategy is checked is too brittle (passing a `null!` DbContext).

@AndriySvyryd @roji

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
```diff
@@ -117,12 +121,52 @@ void ConfigureDbContext(DbContextOptionsBuilder dbContextOptionsBuilder)
...
+            if (settings.Retry || settings.CommandTimeout.HasValue)
             {
-                return;
+                builder.PatchServiceDescriptor<TContext>(optionsBuilder => optionsBuilder.UseNpgsql(options =>
+                {
+                    var extension = optionsBuilder.Options.FindExtension<NpgsqlOptionsExtension>();
+
+                    if (settings.Retry)
+                    {
+                        var executionStrategy = extension?.ExecutionStrategyFactory?.Invoke(new ExecutionStrategyDependencies(null!, optionsBuilder.Options, null!));
```
roji, line 133: @sebastienros just to confirm, you're referring to this right?

/cc @AndriySvyryd

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
eerhardt: You don't want to use `is` because that would return `true` for that type and any derived type. Here we are saying "if the executionStrategy is concretely an NpgsqlExecutionStrategy, we will call `EnableRetryOnFailure()`. For anything else, throw"

davidfowl: Got it. A comment would be great 😄

roji: You're saying you specifically want to exclude subclasses of e.g. NpgsqlRetryingExecutionStrategy? I think it makes sense to assume that such subclasses would also be retrying (otherwise they wouldn't/shouldn't derive from NpgsqlRetryingExecutionStrategy), but ok...

eerhardt: Good point @roji. I think Line 137 should be ` if (executionStrategy is NpgsqlRetryingExecutionStrategy)`.

Line 141 should still be `else if (executionStrategy.GetType() != typeof(NpgsqlExecutionStrategy))`.

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
sebastienros: Yes, `new ExecutionStrategyDependencies(null!` I don't know any other way to get the actual `ExtensionStrategy`, and our tests are actually passing a concrete DbContext because we own the type and know how to construct it. There this code makes me sad, the ctor in other EF providers could definitely guard against it, right?

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
sebastienros: Done

tests/Aspire.Microsoft.EntityFrameworkCore.Cosmos.Tests/EnrichCosmosDbTests.cs
```diff
@@ -116,4 +116,22 @@ public void EnrichSupportCustomOptionsLifetime()
         var context = host.Services.GetRequiredService<ITestDbContext>() as TestDbContext;
         Assert.NotNull(context);
     }
+
+    [Fact]
+    public void EnrichWithConflictingRequestTimeoutThrows()
+    {
+        var builder = Host.CreateEmptyApplicationBuilder(null);
+
+        Assert.Throws<InvalidOperationException>(() =>
```
eerhardt, line 125: Can we just assert around the line we expect to throw?

tests/Aspire.Microsoft.EntityFrameworkCore.SqlServer.Tests/EnrichSqlServerTests.cs
```diff
@@ -84,6 +84,27 @@ public void EnrichCanConfigureDbContextOptions()
...
+            });
+        });
+
+        Assert.Throws<InvalidOperationException>(() =>
+        {
+            builder.EnrichSqlServerDbContext<TestDbContext>(settings => settings.CommandTimeout = 456);
+            using var host = builder.Build();
+            var context = host.Services.GetRequiredService<TestDbContext>();
+        });
+    }
```
eerhardt, line 106:
```suggestion
        builder.EnrichSqlServerDbContext<TestDbContext>(settings => settings.CommandTimeout = 456);
        using var host = builder.Build();

        Assert.Throws<InvalidOperationException>(() =>
        {
            var context = host.Services.GetRequiredService<TestDbContext>();
        });
    }
```

eerhardt (approved):
LGTM. As long as @roji and @AndriySvyryd approve.

tests/Aspire.Microsoft.EntityFrameworkCore.SqlServer.Tests/EnrichSqlServerTests.cs
sebastienros: All providers updated

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
AndriySvyryd: Yes, there's no other way for 8. But, I assume all of this will be removed for the next version, right? You'd just need to make sure that the user calls Enrich before AddDbContext, so all Aspire configuration can be silently overridden.

For SqlServer you can get this behavior in 8 by calling `UseAzureSqlDefaults` instead of `EnableRetryOnFailure`

AndriySvyryd (approved):

src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
eerhardt: > You'd just need to make sure that the user calls Enrich before AddDbContext

I don't think that's the API experience we are looking for. IMO, the current experience makes more sense:

1. AddDbContext
2. EnrichDbContext

If you flip it the other way around, it doesn't really make sense. How can you "Enrich" something before you add it?

AndriySvyryd: Ok, perhaps the question is whether Enrich just provides a recommended default configuration or whether it should be authorative over anything specified in `AddDbContext`.

Personally, I don't see why it should override or throw for any conflicting configuration. A suppressible warning seems more appropriate.

PR #3160, opened 2024-03-26
Add support for specifying custom container run arguments

### LINKED ISSUE #2737: Expose ability to run containers with access to GPUs
One of the prerequisites for running AI container workloads on NVIDIA GPUs is the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html).

Once that's installed, when you run the container, you need to pass the `--gpus` flag. 

Right now there doesn't appear to be a way to configure a container resource in the AppHost project such that it can be started with GPU support, e.g. on Docker for Windows, the container must be run like `docker run -d --gpus=all ...`.

danegsta:
PR adds a new annotation for specifying custom arguments to the container run command issued by the orchestrator. This is intended to support advanced scenarios like enabling GPU support in Docker containers, where adding annotations for individual flags may not make sense.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3160)


davidfowl:
What goes into the manifest?

danegsta:
Only concern I have with adding the run arguments to the manifest is that they can potentially be runtime specific (the `--gpus` flag, for example, isn't supported in podman).

DamianEdwards:
Yeah I think these can't be in the manifest as they're opaque and very much tied to the container host. We can build higher-level container runtime feature constructs (e.g. GPU support) and those could go in the manifest and could leverage this PR's changes when running locally, if we want to.

DamianEdwards:
I'd really like to get this in as it enables GPU support for AI scenarios with Aspire which we have some folks experimenting with right now. @davidfowl @danegsta how do you feel about my comment above RE not putting this in the manifest?

danegsta:
I'm already a bit nervous about adding them to the manifest without any additional runtime context, so I'd be fine with backing out my changes to write to the manifest until we've had a chance to figure out exactly how we want to add this data.

davidfowl:
If they don't go into the manifest does that mean this is a runtime only scenario?

DamianEdwards:
I think so yes. We could choose to add first-class modeling for GPU support later as a portable concept, but this is low level for when you want to pass args directly to the container runtime.

davidfowl:
When you run on kuberentes aren't those args potentially applicable? How is this different from docker build args?

DamianEdwards:
Honestly I don't know. Are they? But even just enabling GPU uses different runtime args between docker and podman.

danegsta:
It looks like Podman 5.0 added support for the same `--gpu` argument as Docker. Doesn't really change the calculous on adding generic container run arguments to the manifest, but maybe this should become a more targeted feature for enabling GPU support rather than exposing the low level ability to add extra run arguments?

danegsta:
@DamianEdwards I backed out the logic to write the manifest.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -175,6 +175,46 @@ public static IResourceBuilder<T> WithEnvironment<T>(this IResourceBuilder<T> bu
...
     }
 
+    /// <summary>
+    /// Adds a callback to be executed with a list of arguments to add to the container run command when a container resource is started.
+    /// </summary>
+    /// <typeparam name="T">The resource type.</typeparam>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="args">The arguments to be passed to the container run command when the container resource is started.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<T> WithContainerRunArgs<T>(this IResourceBuilder<T> builder, params string[] args) where T : IResourceWithArgs
```
davidfowl, line 185: @danegsta I think we want these methods constrained on container resource and moved to container resource extensions.

davidfowl (approved):

davidfowl:
@danegsta backport?

prom3theu5:
This is cool
Would it be possible to emit them to the manifest please as currently they aren't so downstream tools have no way to know that run args are expected to be passed.

DamianEdwards:
We decided not to emit them as they're container-host specific and thus don't translate well, e.g. args for `docker run` are different than those for `podman run`. But we can always change that decision. Please log an issue and we can discuss there.

PR #3177, opened 2024-03-26
Stop the apphost from running if the application url is http and ASPIRE_ALLOW_UNSECURED_TRANSPORT is not set

mitchdenny:
I'm making use of the options validation APIs to blow up if the URL is http and the allow insecure environment is not set. The environment variable that binds to the option is:

```dotenv
ASPIRE_ALLOW_UNSECURED_TRANSPORT = true
```

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3177)


src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
```diff
@@ -54,7 +54,8 @@ internal sealed class DashboardServiceHost : IHostedService
         ILoggerFactory loggerFactory,
         IConfigureOptions<LoggerFilterOptions> loggerOptions,
         ResourceNotificationService resourceNotificationService,
-        ResourceLoggerService resourceLoggerService)
+        ResourceLoggerService resourceLoggerService
```
davidfowl, line 57: undo

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    public DashboardTokenProvider(IOptions<DashboardAuthenticationOptions> dashboardAuthenticationOptions)
+    {
+        BrowserToken = dashboardAuthenticationOptions.Value.BrowserToken ?? GenerateToken();
+        OltpToken = GenerateToken();
+        ResourceServerToken = GenerateToken();
+    }
+
+    private static string GenerateToken()
+    {
+        var rawToken = PasswordGenerator.Generate(24, true, true, true, true, 6, 6, 6, 6);
```
eerhardt, line 28: Why `6, 6, 6, 6`? This should really be:

```suggestion
        var rawToken = PasswordGenerator.Generate(24);
```

You shouldn't care how many lower, upper, numbers, special symbols are produced. Using `minXXX` lowers the entropy.

eerhardt:
Are there any tests that can be written for this new functionality?

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+        var rawTokenBytes = Encoding.UTF8.GetBytes(rawToken);
+        var encodedToken = Convert.ToHexString(rawTokenBytes).ToLower();
+        return encodedToken;
+    }
+
+    public string BrowserToken { get; init; }
+    public string OltpToken { get; init; }
+    public string ResourceServerToken { get; init; }
```
eerhardt, line 36:
```suggestion
    public string BrowserToken { get; }
    public string OltpToken { get; }
    public string ResourceServerToken { get; }
```

These should be readonly properties.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -708,6 +711,19 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            {
+                // Configure for insecure running.
+                context.EnvironmentVariables["DISABLEAUTHPLACEHOLDER"] = true;
+            }
+            else
+            {
+                // Configure for secure running.
+                context.EnvironmentVariables["BROWSERTOKENPLACEHOLDER"] = dashboardTokenProvider.BrowserToken;
+                context.EnvironmentVariables["OTLPTOKENPLACEHOLDER"] = dashboardTokenProvider.OltpToken;
+                context.EnvironmentVariables["RESOURCESERVERTOKENPLACEHOLDER"] = dashboardTokenProvider.ResourceServerToken;
```
davidfowl, line 724: needs more underscores

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -89,9 +91,12 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
         _innerBuilder.Services.AddSingleton<ResourceLoggerService>();
 
         // Dashboard
+        _innerBuilder.Services.AddOptions<DashboardAuthenticationOptions>().Bind(Configuration.GetSection("Aspire:Dashboard:Authentication"));
```
eerhardt, line 94: Can we enable the config binder source generator? That way we don't have to do this reflection at startup.

```xml
<EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
```

in the .csproj.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -708,6 +711,19 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
 
             var grpcEndpointUrl = await _dashboardEndpointProvider.GetResourceServiceUriAsync(context.CancellationToken).ConfigureAwait(false);
 
+            if (dashboardAuthenticationOptions.Value.AllowUnsecureTransport.GetValueOrDefault(false))
+            {
+                // Configure for insecure running.
+                context.EnvironmentVariables["DISABLEAUTHPLACEHOLDER"] = true;
```
eerhardt, line 717: Who respects these env vars?

Do they also need to go in the `StartDashboardAsDcpExecutableAsync` method below?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: They are placeholders, and will be changed to the real values before merge. I just need James' change in first.

eerhardt: Which PR? https://github.com/dotnet/aspire/pull/3119 is already merged.

src/Aspire.Hosting/Dashboard/DashboardAuthenticationOptionsValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Aspire.Hosting.Dashboard;
+
+internal class DashboardAuthenticationOptionsValidator(IConfiguration configuration) : IValidateOptions<DashboardAuthenticationOptions>
+{
+    public ValidateOptionsResult Validate(string? name, DashboardAuthenticationOptions options)
+    {
+        if (configuration[KnownEnvironmentVariables.AspNetCoreUrls] is not { } applicationUrls)
+        {
+            throw new DistributedApplicationException($"AppHost does not have applicationUrl in launch profile, or {KnownEnvironmentVariables.AspNetCoreUrls} environment variable set.");
```
drewnoakes, line 15: The comma makes it unclear what the "not" applies to. Would this be clearer?

> AppHost does not have applicationUrl in launch profile and the ASPNETCORE_URLS environment variable is not set.

src/Aspire.Hosting/Dashboard/DashboardAuthenticationOptionsValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        var firstApplicationUrl = applicationUrls.Split(";").First();
+
+        if (!Uri.TryCreate(firstApplicationUrl, UriKind.Absolute, out var parsedFirstApplicationUrl))
+        {
+            throw new DistributedApplicationException($"AppHost applicationUrl in launch profile, or {KnownEnvironmentVariables.AspNetCoreUrls} environment variable could not be parsed as a URI.");
```
drewnoakes, line 22: Consider:

> Unable to parse launch URI from either applicationUrl in the launch profile, or the ASPNETCORE_URLS environment variable.

or just add another comma:

> AppHost applicationUrl in launch profile, or ASPNETCORE_URLS environment variable, could not be parsed as a URI.

Consider also showing the URI in the message, unless someone feels that's leaking private info.

src/Aspire.Hosting/Dashboard/DashboardAuthenticationOptionsValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        if (parsedFirstApplicationUrl.Scheme == "http" && !options.AllowUnsecureTransport.GetValueOrDefault(false))
+        {
+            throw new DistributedApplicationException($"AppHost cannot use a non-TLS URL for applicationUrl in launch profile, or {KnownEnvironmentVariables.AspNetCoreUrls} unless DashboardAuthenticationOptions.AllowUnsecureTransport is set to true.");
+        }
+
+        _ = configuration;
```
drewnoakes, line 30: Remove?

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    public DashboardTokenProvider(IOptions<DashboardAuthenticationOptions> dashboardAuthenticationOptions)
+    {
+        BrowserToken = dashboardAuthenticationOptions.Value.BrowserToken ?? GenerateToken();
+        OltpToken = GenerateToken();
+        ResourceServerToken = GenerateToken();
+    }
+
+    private static string GenerateToken()
+    {
+        var rawToken = PasswordGenerator.Generate(24, true, true, true, true, 0, 0, 0, 0);
```
drewnoakes, line 28: Named arguments for the literals here would make this easier to read without an IDE.

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+        var rawTokenBytes = Encoding.UTF8.GetBytes(rawToken);
+        var encodedToken = Convert.ToHexString(rawTokenBytes).ToLower();
```
drewnoakes, line 30: .NET 9 has `Convert.ToHexStringLower`. Maybe add a TODO here or add conditional code for `#if NET9_OR_GREATER`, etc.

https://learn.microsoft.com/en-us/dotnet/api/system.convert.tohexstringlower?view=net-9.0

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: Yeah ... it wasn't last night :)

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: Ended up not using Bind at all. We are using custom environment variable names so I am just doing this via `PostConfigure<TOptions>(...)`.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: This code has been removed in the PR for now. It will go in after I backport this change to P5.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: This code has been removed for now.

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
mitchdenny: I'll go one better and default the values so no one needs to think about it :)

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Text;
+using Aspire.Hosting.Utils;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Hosting.Dashboard;
+
+internal interface IDashboardTokenProvider
+{
+    public string BrowserToken { get; }
+    public string OltpToken { get; }
```
JamesNK, line 13:
```suggestion
    public string DashboardOtlpToken { get; }
```

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+    public string ResourceServerToken { get; }
+}
+
+internal class DashboardTokenProvider : IDashboardTokenProvider
+{
+    public DashboardTokenProvider(IOptions<TransportOptions> transportOptions)
+    {
+        BrowserToken = transportOptions.Value.BrowserToken ?? GenerateToken();
```
JamesNK, line 21: Add debug level logging here about whether the token from the options are used or one is generated.

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System.Text;
+using Aspire.Hosting.Utils;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Hosting.Dashboard;
+
+internal interface IDashboardTokenProvider
+{
+    public string BrowserToken { get; }
```
JamesNK, line 12:
```suggestion
    public string DashboardFrontendToken { get; }
```

src/Aspire.Hosting/Utils/PasswordGenerator.cs
```diff
@@ -55,8 +55,8 @@ internal static class PasswordGenerator
     /// log base 2 [67^x * 23^a * 23^b * 10^c * 11^d * (a + b + c + d)! / (a! * b! * c! * d!)]
     /// </remarks>
     public static string Generate(int minLength,
-        bool lower, bool upper, bool numeric, bool special,
-        int minLower, int minUpper, int minNumeric, int minSpecial)
+        bool lower = true, bool upper = true, bool numeric = true, bool special = true,
+        int minLower = 0, int minUpper = 0, int minNumeric = 0, int minSpecial = 0)
```
JamesNK, line 59: IMO don't add defaults. Just use named arguments in the few places this is called. Keeps it clear about what is generated.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -119,6 +124,11 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
         LogBuilderConstructed(this);
     }
 
+    private void MapTransportOptionsFromCustomKeys(TransportOptions options)
+    {
+        options.AllowUnsecureTransport = Configuration.GetBool(KnownEnvironmentVariables.AllowUnsecuredTransport, false);
```
JamesNK, line 129: Only set this if the configuration is present. Allows someone to configure `AllowUnsecureTransport` directly without it always being overriden.

e.g.
https://github.com/dotnet/aspire/blob/23104ee048dcf99bc1bdcb5ff8ad01ee85921d3b/src/Aspire.Dashboard/Configuration/PostConfigureDashboardOptions.cs#L33-L34

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+        var firstApplicationUrl = applicationUrls.Split(";").First();
+
+        if (!Uri.TryCreate(firstApplicationUrl, UriKind.Absolute, out var parsedFirstApplicationUrl))
+        {
+            return ValidateOptionsResult.Fail($"The 'applicationUrl' setting of the launch profile has value '{firstApplicationUrl}' which could not be parsed as a URI.");
+        }
+
+        if (parsedFirstApplicationUrl.Scheme == "http" && !options.AllowUnsecureTransport.GetValueOrDefault(false))
+        {
+            return ValidateOptionsResult.Fail($"Use of a non-TLS URL in the 'applicationUrl' setting of the launch profile unless the 'ASPIRE_ALLOW_UNSECURED_TRANSPORT' environment variable is set to 'true'. See https://aka.ms/dotnet/aspire/allowunsecuredtransport for more details.");
```
JamesNK, line 27: Get ASPIRE_ALLOW_UNSECURED_TRANSPORT from const

JamesNK:
Use the API key from the provider. Can remove this code:

https://github.com/dotnet/aspire/blob/23104ee048dcf99bc1bdcb5ff8ad01ee85921d3b/src/Aspire.Hosting/DistributedApplicationBuilder.cs#L170-L175

And update the place when that config key was used.

mitchdenny:
> Use the API key from the provider. Can remove this code:
> 
> https://github.com/dotnet/aspire/blob/23104ee048dcf99bc1bdcb5ff8ad01ee85921d3b/src/Aspire.Hosting/DistributedApplicationBuilder.cs#L170-L175
> 
> And update the place when that config key was used.
> 
> Edit: Actually, I think it still needs to be set in the config so it can flow into AddOtlpEnvironment. I think you might run into order issues here. Value is needed before the container is built.

We can address this in a subsequent PR, this PR is (primarily) about failing if the ASPIRE_ALLOW_UNSECURED_TRANSPORT variable is not set when using a http endpoint.

src/Aspire.Hosting/Utils/PasswordGenerator.cs
drewnoakes: If nothing's actually using these arguments, and the type's not public, I'd consider removing them. Can always add them later.

mitchdenny: Reverted default args and added named args to callsite.

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return ValidateOptionsResult.Fail($"AppHost does not have applicationUrl in launch profile, or {KnownEnvironmentVariables.AspNetCoreUrls} environment variable set.");
+        }
+
+        var firstApplicationUrl = applicationUrls.Split(";").First();
+
+        if (!Uri.TryCreate(firstApplicationUrl, UriKind.Absolute, out var parsedFirstApplicationUrl))
+        {
+            return ValidateOptionsResult.Fail($"The 'applicationUrl' setting of the launch profile has value '{firstApplicationUrl}' which could not be parsed as a URI.");
+        }
```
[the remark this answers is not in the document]
mitchdenny: @davidfowl what was the consensus on this one. Did we want to fail if ANY of the URLs where http?

src/Shared/KnownEnvironmentVariables.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting;
+
+internal static class KnownEnvironmentVariables
```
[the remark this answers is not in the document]
mitchdenny: I ended up going with KnownConfigNames since we use that convention in a few places.

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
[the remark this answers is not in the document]
davidfowl: We’re said the first one determines the scheme

tests/cosmos.module.bicep
```diff
@@ -0,0 +1,42 @@
+targetScope = 'resourceGroup'
```
JamesNK, line 1: What's this file for?

src/Shared/KnownConfigNames.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting;
+
+internal static class KnownConfigNames
```
JamesNK, line 6: KnownHostConfigNames

JamesNK (approved):

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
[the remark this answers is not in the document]
JamesNK: What happens if there is https first and http second and I visit the dashboard with the http address?

src/Shared/KnownConfigNames.cs
mitchdenny: Some of these values are going to likely be shared between components (e.g. dashboard and apphost). They represent the protocol which is why I wasn't getting specific to which process nominally owns them.

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+internal class TransportOptionsValidator(IConfiguration configuration, DistributedApplicationExecutionContext executionContext) : IValidateOptions<TransportOptions>
+{
+    public ValidateOptionsResult Validate(string? name, TransportOptions options)
+    {
+        if (executionContext.IsPublishMode)
+        {
+            return ValidateOptionsResult.Success;
+        }
+
+        if (configuration[KnownConfigNames.AspNetCoreUrls] is not { } applicationUrls)
```
davidfowl, line 18: This needs to check null or empty

src/Aspire.Hosting/Dashboard/TransportOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Dashboard;
+
+internal class TransportOptions
+{
+    public bool? AllowUnsecureTransport { get; set; }
```
davidfowl, line 8: Do we need to differentiate null vs false anywhere?

davidfowl (changes requested):

src/Aspire.Hosting/Dashboard/TransportOptions.cs
mitchdenny: Nope. I've changed this to just a plain old bool. We resolve the truthiness in PostConfigure so we know what this value is by the time people are reading this.

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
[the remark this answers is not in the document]
davidfowl: The dashboard should [redirect to https](https://github.com/dotnet/aspire/blob/e24eb81ab71faf4d0e18021a543a755513c313b2/src/Aspire.Dashboard/DashboardWebApplication.cs#L197)

tests/testproject/TestProject.AppHost/TestProgram.cs
```diff
@@ -35,6 +36,11 @@ private TestProgram(string[] args, Assembly assembly, bool includeIntegrationSer
 
         AppBuilder = DistributedApplication.CreateBuilder(new DistributedApplicationOptions { Args = args, DisableDashboard = disableDashboard, AssemblyName = assembly.FullName });
 
+        if (allowUnsecuredTransport)
+        {
+            AppBuilder.Configuration[KnownConfigNames.AllowUnsecuredTransport] = "true";
```
davidfowl, line 41: Lets expose the option on `DistributedApplicationOptions`

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -15,7 +15,7 @@ public ValidateOptionsResult Validate(string? name, TransportOptions options)
             return ValidateOptionsResult.Success;
         }
 
-        if (configuration[KnownConfigNames.AspNetCoreUrls] is not { } applicationUrls)
+        if (configuration[KnownConfigNames.AspNetCoreUrls] is not { } applicationUrls || string.IsNullOrEmpty(applicationUrls))
```
davidfowl, line 18:
```suggestion
        if (string.IsNullOrEmpty(configuration[KnownConfigNames.AspNetCoreUrls]))
```

davidfowl:
None of this should fail if you disable the dashboard right?

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Aspire.Hosting.Dashboard;
+
+internal class TransportOptionsValidator(IConfiguration configuration, DistributedApplicationExecutionContext executionContext) : IValidateOptions<TransportOptions>
+{
+    public ValidateOptionsResult Validate(string? name, TransportOptions options)
+    {
+        if (executionContext.IsPublishMode)
+        {
+            return ValidateOptionsResult.Success;
+        }
```
davidfowl, line 16: I think we also want to noop if the dashboard isn't running.

src/Aspire.Hosting/Dashboard/DashboardTokenProvider.cs
```diff
@@ -0,0 +1,64 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
davidfowl, line 1: I would remove this from this PR and focus solely on the insecure mode. We can do this in a P6 branch.

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
mitchdenny: Even if we aren't running the dashboard, the resource server is still running.

davidfowl: It's not 😄

src/Aspire.Hosting/Dashboard/TransportOptionsValidator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        var firstApplicationUrl = applicationUrls.Split(";").First();
+
+        if (!Uri.TryCreate(firstApplicationUrl, UriKind.Absolute, out var parsedFirstApplicationUrl))
+        {
+            return ValidateOptionsResult.Fail($"The 'applicationUrl' setting of the launch profile has value '{firstApplicationUrl}' which could not be parsed as a URI.");
+        }
+
+        if (parsedFirstApplicationUrl.Scheme == "http" && !effectiveAllowUnsecureTransport)
```
davidfowl, line 33: We need more checks here:

DOTNET_DASHBOARD_OTLP_ENDPOINT_URL

DOTNET_RESOURCE_SERVICE_ENDPOINT_URL

These endpoints are being set today by our template. These should also be checked for http. They only support a single address so that simplifies things.

playground/AWS/AWS.AppHost/Properties/launchSettings.json
```diff
@@ -9,7 +9,8 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
         "DOTNET_ENVIRONMENT": "Development",
-        "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "http://localhost:16216"
+        "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "http://localhost:16216",
+        "ASPIRE_ALLOW_UNSECURED_TRANSPORT": "true"
```
eerhardt, line 13: If we are putting this in all our playground apps, do we also need it in the templates?

src/Aspire.Hosting/Dashboard/TransportOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Dashboard;
+
+internal class TransportOptions
+{
+    public bool AllowUnsecureTransport { get; set; }
+    public string? BrowserToken { get; set; }
```
eerhardt, line 9: This property doesn't appear to be used.

```suggestion
```

playground/AWS/AWS.AppHost/Properties/launchSettings.json
eerhardt: Or, if this isn't supposed to be in the templates, but instead we are supposed to use HTTPS, shouldn't all the playground apps be updated to use HTTPS instead?

DamianEdwards: Not going in templates and the goal should be to update all the playground apps (or maybe all but one) to use HTTPS.

mitchdenny: I'll update the PR once I am back online this morning.

davidfowl: I agree, one should use https, all of the rest should be http with this flag.

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -14,8 +14,8 @@ public class TestingBuilderTests
     public async Task HasEndPoints()
     {
         var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.TestingAppHost1_AppHost>();
+        appHost.Configuration[KnownConfigNames.AllowUnsecuredTransport] = "true";
```
davidfowl, line 17: `DistributedApplicationTestingBuilder` turns off the dashboard, so we don't need this.

playground/AWS/AWS.AppHost/Properties/launchSettings.json
DamianEdwards: Not sure who you're agreeing with, but I said the exact opposite :smile:

mitchdenny: #confused

davidfowl: LOL

mitchdenny: OK I've got a http and https launch profile in all our launch settings file with settings that mimic (for the most part) what we get from the templates. There are some exceptions in some playgrounds for specific reasons.

tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
```diff
@@ -48,3 +48,12 @@ private sealed record WeatherForecast(DateOnly Date, int TemperatureC, string? S
         public int TemperatureF => 32 + (int)(TemperatureC / 0.5556);
     }
 }
+
+public class DistributedApplicationFixtureWithAllowUnsecureTransport<TEntryPoint> : DistributedApplicationFixture<TEntryPoint> where TEntryPoint: class
+{
+    protected override void OnBuilding(DistributedApplicationBuilder applicationBuilder)
+    {
+        applicationBuilder.Configuration[KnownConfigNames.AllowUnsecuredTransport] = "true";
```
davidfowl, line 56: This isn't needed right?

davidfowl (approved):
small nits but this is good.

tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
mitchdenny: Correct. I'll remove this before merging.

PR #3180, opened 2024-03-26
Remove obsolete CopySrcInsteadOfClone source-build property

MichaelSimons:
Copy is now the default/only behavior.  This property was removed with https://github.com/dotnet/source-build/issues/4117
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3180)


MichaelSimons:
hmmm, I forgot aspire is on a 8.0 version of arcade.  This can't be removed until 9.0 arcade is picked up.

Issue #2367, opened 2024-02-23
Dapr --app-port argument not consistently passed to the dapr executable


darraghjones:
I see the following error in the dapr logs because the app-port argument is not passed to the dapr executable.

time="2024-02-23T00:06:55.4110487Z" level=warning msg="App channel is not initialized. Did you configure an app-port?" app_id=web instance=DESKTOP-SE6C36R scope=dapr.runtime.channels type=log ver=1.12.5

You can see this happening below running the AspireWithDapr sample 

<img width="1830" alt="image" src="https://github.com/dotnet/aspire/assets/431192/da5743e0-a767-4adc-9f4e-2c9feb5f7b73">

The app-port argument is being passed to the API service dapr instance, but I couldn't see any reason why it would behave differently.

darraghjones:
Ah, it seems this has to do with the web running on https. If i change this to http, the app-port argument is passed along to dapr.

davidfowl:
Not sure what we can do about this

davidfowl:
Allowed specifying the endpoint to use for the sidecar resource. This determines the app port. 

Fixed in https://github.com/dotnet/aspire/pull/3184

PR #3184, opened 2024-03-26
#2367 dapr change the app port depending on the app protocol

paule96:
if you dapr connected apps uses anything else then `http` they will currently not correctly connecting to the sidecare. (more details to that issue in #2367 )

To fix this now, a use can set the app protocol in the sidecare options and this will then take the correct endpoint information.

Open questions:

- [x] Can someone test the playground project for dapr?
  - for me on Linux it seems to be broken in

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3184)


paule96:
okay, I know now why the playground project is not working. There seems to be a bug unrelated to this PR that it is required to have dapr `1.13.0` as a reference for aspire >= preview 4.
The problem is there is only dapr `1.12.0` available on the feeds configured in the NuGet.config.

paule96:
Okay after I integrated locally the nuget.org feed I was able to test my change. For me it looks like it is working.

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -146,13 +146,13 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+                        
                         context.EnvironmentVariables.TryAdd("DAPR_GRPC_ENDPOINT", grpc);
                         context.EnvironmentVariables.TryAdd("DAPR_HTTP_ENDPOINT", http);
                     }));
 
-            daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "grpc", port: sidecarOptions?.DaprGrpcPort));
-            daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "http", port: sidecarOptions?.DaprHttpPort));
+            daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "grpc", port: sidecarOptions?.DaprGrpcPort, uriScheme: "http"));
+            daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "http", port: sidecarOptions?.DaprHttpPort, uriScheme: "http"));
             daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "metrics", port: sidecarOptions?.MetricsPort));
```
paule96, line 156: @philliphoff do you know what the uri schema for metrics and profiling should be?

philliphoff: Both endpoints are HTTP-based, with the metrics endpoint being a basic HTTP endpoint and, I believe, the profiling endpoint being gRPC.

paule96: found that in the docs for the profiling option:

| Name | Environment Variable |  Default | Description |
| ------ | ------------------------  | ------- | ------------- |
| --enable-profiling |   | false | Enable “pprof” profiling via an HTTP endpoint |


So it's HTTP too. Changed it accordingly.

paule96:
@davidfowl and @karolz-ms,

because you both were involved in #3305 can I ask you to review this PR too?

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -146,17 +146,16 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
 
                         var http = daprCli.GetEndpoint("http");
                         var grpc = daprCli.GetEndpoint("grpc");
-
+
```
davidfowl, line 149:
```suggestion
```

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -146,17 +146,16 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
 
                         var http = daprCli.GetEndpoint("http");
                         var grpc = daprCli.GetEndpoint("grpc");
-
+                        
                         context.EnvironmentVariables.TryAdd("DAPR_GRPC_ENDPOINT", grpc);
                         context.EnvironmentVariables.TryAdd("DAPR_HTTP_ENDPOINT", http);
                     }));
-
```
davidfowl, line 153: Why did you remove the space?

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -174,7 +173,8 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
                         EndpointReference? httpEndPoint = null;
                         if (resource is IResourceWithEndpoints resourceWithEndpoints)
                         {
-                            httpEndPoint = resourceWithEndpoints.GetEndpoint("http");
+                            var endpointName = sidecarOptions?.AppProtocol ?? "http";
```
davidfowl, line 176: This coupling between the dapr protocol and the endpoint name of the app is interesting. Maybe we should have 2 separate options and a default fallback where we assume the app protocol and endpoint name are the same.

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -146,17 +146,16 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
                         context.EnvironmentVariables.TryAdd("DAPR_HTTP_ENDPOINT", http);
                     }));
-
             daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, uriScheme: "http", name: "grpc", port: sidecarOptions?.DaprGrpcPort));
             daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, uriScheme: "http", name: "http", port: sidecarOptions?.DaprHttpPort));
             daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, uriScheme: "http", name: "metrics", port: sidecarOptions?.MetricsPort));
             if (sidecarOptions?.EnableProfiling == true)
             {
-                daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "profile", port: sidecarOptions?.ProfilePort));
+                daprCli.Annotations.Add(new EndpointAnnotation(ProtocolType.Tcp, name: "profile", port: sidecarOptions?.ProfilePort, uriScheme: "http"));
```
davidfowl, line 158: Good catch!

playground/dapr/AppHost/Program.cs
```diff
@@ -1,15 +1,21 @@
+using Aspire.Hosting.Dapr;
```
davidfowl, line 1: We don't want to change the sample, can you write a test instead? We now have dapr tests.

paule96: cool was not sure. When I started with that PR there were no tests for dapr. :)

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: @davidfowl you mean something like that:

```csharp
builder.AddProject<Projects.DaprServiceA>("servicea", "https")
       .WithDaprSidecar(new DaprSidecarOptions{
              AppEndpoint = "MyEndpoint",
              AppProtocol = "HTTP"
       })
```

And there are the following cases:

- `AppEndpoint` and `AppProtocol ` is `null`
  - use `http` for both
- `AppEndpoint` is **not** `null` and `AppProtocol` is `null`
  - use as the `AppProtocol` the protocol that is defined by the `AppEndpoint` URL. 
  - if the extraction failes, fall back to `http`
- `AppEndpoint` is `null` and `AppProtocol` is **not** `null`
  - Get the app endpoint with the `AppProtocol` name and set it as `AppEndpoint`
  - if the extraction failes, fall back to `http` endpoint
    - if that fails exception??
- `AppEndpoint` is not `null` and `AppProtocol` is not `null`
  - use both as they are defined

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: yes we both fixed the same thing. :)

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: fixed

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: resolved

paule96:
@davidfowl please review again :) I changed the logic and also added tests

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -170,16 +171,11 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
-
-                        EndpointReference? httpEndPoint = null;
-                        if (resource is IResourceWithEndpoints resourceWithEndpoints)
-                        {
-                            var endpointName = sidecarOptions?.AppProtocol ?? "http";
-                            httpEndPoint = resourceWithEndpoints.GetEndpoint(endpointName.ToLower());
-
-                            if (httpEndPoint.IsAllocated && sidecarOptions?.AppPort is null)
+                        var endPoint = GetEndpointReference(sidecarOptions, resource);
+                        if(endPoint is not null){
```
davidfowl, line 175: nit: Fix the formatting.

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -256,6 +255,33 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
         appModel.Resources.AddRange(sideCars);
     }
 
+    static (EndpointReference appEndpoint, string protocol)? GetEndpointReference(DaprSidecarOptions? sidecarOptions, IResource resource)
+    {
+        if (resource is IResourceWithEndpoints resourceWithEndpoints)
```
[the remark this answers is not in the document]
paule96: done

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: done

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -255,6 +257,47 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
         appModel.Resources.AddRange(sideCars);
     }
 
+    /// <summary>
+    /// Get's the applications endpoint and the protocl for it.
```
davidfowl, line 261:
```suggestion
    /// Get's the applications endpoint and the protocol for it.
```

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -255,6 +257,47 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+    /// - <see cref="DaprSidecarOptions.AppEndpoint"/>
+    /// The logic is like:
+    /// - if both are null try to use 'http' for both
+    /// - if <see cref="DaprSidecarOptions.AppProtocol"/> is not null try to get an endpoint with the name of the protocol
+    /// - if <see cref="DaprSidecarOptions.AppEndpoint"/> is not null try to use the schema as the protocol
+    /// - if bot are not null just use them.
+    /// </summary>
+    /// <param name="sidecarOptions">The current options that are the base of the logic.</param>
+    /// <param name="resource">The application resource the sidecar is connected to.</param>
+    /// <returns>The appendpoint that should be used and the protocol.</returns>
```
davidfowl, line 273:
```suggestion
    // This method resolves the application's endpoint and the protocol that the dapr side car will use.
    // It depends on DaprSidecarOptions.AppProtocol and DaprSidecarOptions.AppEndpoint.
    // - If both are null default to 'http' for both.
    // - If AppProtocol is not null try to get an endpoint with the name of the protocol.
    // - if AppEndpoint is not null try to use the scheme as the protocol.
    // - if both are not null just use both options.
```

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -255,6 +257,39 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+    // It depends on DaprSidecarOptions.AppProtocol and DaprSidecarOptions.AppEndpoint.
+    // - If both are null default to 'http' for both.
+    // - If AppProtocol is not null try to get an endpoint with the name of the protocol.
+    // - if AppEndpoint is not null try to use the scheme as the protocol.
+    // - if both are not null just use both options.
+    static (EndpointReference appEndpoint, string protocol)? GetEndpointReference(DaprSidecarOptions? sidecarOptions, IResource resource)
+    {
+        if (resource is IResourceWithEndpoints resourceWithEndpoints)
+        {
+            if (sidecarOptions == null || (sidecarOptions.AppProtocol == null && sidecarOptions.AppEndpoint == null))
```
davidfowl, line 270: Can you rewrite this as a switch expression?

paule96: @davidfowl Do you have time for a bit of guidance here.
I tried it but failed and I don't know what I do wrong:

```csharp
return (sidecarOptions, resourceWithEndpoints) switch{
                (DaprSidecarOptions? p0, IResourceWithEndpoints p1) when p0 == null || (p0.AppProtocol == null && p0.AppEndpoint == null) => (p1.GetEndpoint("http"), "http"),
                _ => throw new ArgumentException($"The combination of {nameof(DaprSidecarOptions.AppEndpoint)} and {nameof(DaprSidecarOptions.AppProtocol)}")
            }
```

That's my current code. And it will result in:

"CS0119", 'DaprSidecarOptions' is a type, which is not valid in the given context"

paule96: @davidfowl nevermind... I just used `var` and it works...

davidfowl: I think it would look something like this:

```C#
if (resource is not IResourceWithEndpoints resourceWithEndpoints)
{
    return null;
}

return (sidecarOptions?.AppProtocol, sidecarOptions?.AppEndpoint) switch
{
    // If both are not null, use AppEndpoint and AppProtocol
    (string appProtocol, string appEndpoint) => (resourceWithEndpoints.GetEndpoint(appEndpoint), appProtocol),

    // If both are null, default to http
    (null, null) => (resourceWithEndpoints.GetEndpoint("http"), "http"),

    // If AppProtocol is null, use AppEndpoint
    (null, string appEndpoint) => (resourceWithEndpoints.GetEndpoint(appEndpoint), appEndpoint),

    // If both are not null, use AppEndpoint and AppProtocol
    (string appProtocol, null) => (resourceWithEndpoints.GetEndpoint(appProtocol), appProtocol);
}
```

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -267,25 +267,13 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
-            {
-                var appEndpoint = resourceWithEndpoints.GetEndpoint(sidecarOptions.AppProtocol);
-                return (appEndpoint, sidecarOptions.AppProtocol);
-            }
-            else if (sidecarOptions.AppProtocol != null && sidecarOptions.AppEndpoint != null)
-            {
-                var appEndpoint = resourceWithEndpoints.GetEndpoint(sidecarOptions.AppEndpoint);
-                return (appEndpoint, sidecarOptions.AppProtocol);
-            }
+            return (sidecarOptions, resourceWithEndpoints) switch{
```
davidfowl, line 270: Why not get rid of teh when expressions and just check the 2 values? That makes it much cleaner.

paule96: how does that work? It's my first time with the new switch expression :O

RussKie: It's the same as the old `switch` with less "boilerplate" if you like. You may find [these docs](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/switch-expression) helpful.

In a nutshel, this method switches on a number of different cases (quite elaborate) where args are typles and for each case returns tuples.

Does this makes sense?

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -255,6 +257,27 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+    // It depends on DaprSidecarOptions.AppProtocol and DaprSidecarOptions.AppEndpoint.
+    // - If both are null default to 'http' for both.
+    // - If AppProtocol is not null try to get an endpoint with the name of the protocol.
+    // - if AppEndpoint is not null try to use the scheme as the protocol.
+    // - if both are not null just use both options.
+    static (EndpointReference appEndpoint, string protocol)? GetEndpointReference(DaprSidecarOptions? sidecarOptions, IResource resource)
+    {
+        if (resource is IResourceWithEndpoints resourceWithEndpoints)
+        {
+            return (sidecarOptions, resourceWithEndpoints) switch{
```
RussKie, line 270: [styling] Here and few other places the curly has to be on the new line

tests/Aspire.Hosting.Tests/Dapr/DaprTests.cs
```diff
@@ -77,10 +77,92 @@ public async Task WithDaprSideCarAddsAnnotationAndSidecarResource()
...
+                e.UriScheme = "http";
+                e.AllocatedEndpoint = new(e, "localhost", 8000);
+            })
+            .WithEndpoint("https", e =>
+            {
+                e.Port = 8001;
+                e.UriScheme = "https";
+                e.AllocatedEndpoint = new(e, "localhost", 8001);
+            });
+        if(schema == null && endPoint == null && port == null){
```
RussKie, line 122: [styling] spacing
```suggestion
        if (schema == null && endPoint == null && port == null) {
```

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -255,6 +257,27 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+                (var p0, var p1) when p0 == null || (p0.AppProtocol == null && p0.AppEndpoint == null) => (p1.GetEndpoint("http"), "http"),
```
RussKie, line 271: ❗  Equality operator can be overloaded which can load to hard to track bugs.
```suggestion
                (var p0, var p1) when p0 is null || (p0.AppProtocol is null && p0.AppEndpoint is null) => (p1.GetEndpoint("http"), "http"),
```

ditto other places

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: Sadly no. I did read the documentation but have no clue how it should work with two a parameters

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: done

tests/Aspire.Hosting.Tests/Dapr/DaprTests.cs
paule96: done

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: done

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
paule96: can you provide an example, maybe with the code from this pr

davidfowl: My comment here described it https://github.com/dotnet/aspire/pull/3184#discussion_r1560554150

paule96: sorry didn't saw this comment. Fixed it now.

davidfowl:
This looks great

tests/Aspire.Hosting.Tests/Dapr/DaprTests.cs
```diff
@@ -77,10 +77,96 @@ public async Task WithDaprSideCarAddsAnnotationAndSidecarResource()
...
+        var ports = new Dictionary<string, int>
+        {
+            ["http"] = 3500,
+            ["grpc"] = 50001,
+            ["metrics"] = 9090
+        };
+
+        foreach (var e in endpoints)
+        {
+            e.AllocatedEndpoint = new(e, "localhost", ports[e.Name], targetPortExpression: $$$"""{{- portForServing "{{{e.Name}}}" -}}""");
```
karolz-ms, line 156: Is it necessary to set the `targetPortsExpression` here?

karolz-ms (commented):
Looks great overall, thanks so much @paule96 for taking care of this. I have just one question regarding the use of `targetPortExpression` in the test

davidfowl (approved):

karolz-ms (approved):

tests/Aspire.Hosting.Tests/Dapr/DaprTests.cs
paule96: I'm not sure if the dapr related code would work without it. I will take a look later.

paule96: @karolz-ms  yes it is important. If you don't do that you will get the following exception inside of `Aspire.Hosting.ApplicationModel.EndpointReference`

image

karolz-ms: OK thanks for confirmation

PR #3207, opened 2024-03-27
De-duplicate environment variables passed to the dashboard

JamesNK:
The dashboard can be started as a raw executable, or it can be specified as a project (done in playground apps) and then annotated with settings as required.

These two modes duplicated how config via env vars was passed to the dashboard. This PR removes duplication.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3207)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -679,43 +679,6 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


davidfowl:
I like this but I’d prefer if we made the dashboard an app model resource in both cases and share the code to configure it (this is what I was looking at last night)

JamesNK:
IMO we can improve that in a later PR. Having env vars calculated in a separate method is a good improvement regardless of the future.

I want to remove the duplication ASAP. We will change it a lot with new auth modes and other config.

drewnoakes:
This would be helpful for what I'm working on too, so I'm supportive of this. @davidfowl would it conflict with changes you want to land soon?

davidfowl (approved):

Issue #2590, opened 2024-03-02
Improvements to the azure deployment for local dev experience


davidfowl:
- [x] - Status goes from Running to Starting - This happens because we default to Running if nothing is set, then the azure provisioning logic changes it to Starting.
- [x] - Child resources are always running - Any azure resource with children are always running. If the parent resources is failed, so should the child resource fail.
- [x] - Show deployment id - Today we show individual resources that were provisioned, it would be better to show the deployment.
- [x] - ~Show incremental progress while deploying - Today we have a single deployment operation and we should show status.~
- [x] - ~Garbage collect unreferenced resources~
- [x] #3251
- [x] - [Improve the error messages when the azure subscription information is not properly configured (remove stack traces from errors).] #3251
- [x] - Show better error message for when the Azure CLI is not installed. #3251
- [x] - Show better status messages on provisioning operations (https://github.com/dotnet/aspire/pull/3210)
- [x] - https://github.com/dotnet/aspire/issues/1329

davidfowl:
Filed follow up issues

Issue #2415, opened 2024-02-26
Resource server protocol changes


davidfowl:
As part of the work to support custom resources in the dashboard (https://github.com/dotnet/aspire/pull/2390) there are several improvements that we should consider making to the resource server API:
 
- [ ] #5065
- [x] Make the list of endpoints/urls more general purpose and support named urls. For e.g. we want to be able to navigate to the portal for azure resources.
- [x] Feedback here about the service and endpoints split (https://github.com/dotnet/aspire/pull/2413#issue-2153052580) cc @JamesNK 
- [ ] Model replicas explicitly. Today we have a flat list of resources and replicas are found based on naming. We can make this explicit.
- [x] Allow setting both the state and the state icon or icon type (https://github.com/dotnet/aspire/pull/3210)
- [x] #4849

kvenkatrajan:
@davidfowl , I am not sure I understand this request : 

> Make the list of endpoints/urls more general purpose and support named urls. For e.g. we want to be able to navigate to the portal for azure resources.

Can you give example of variations of urls for local/azure resources you would expect? 

The other 2 are clear - Thanks!

JamesNK:
What about adding references to resources? i.e. what resources is a resource referencing. We can show that in the resource detail pane.

```cs
builder.AddProject<Projects.MyFrontend>("frontend")
       .WithReference(basketService)
       .WithReference(catalogService.GetEndpoint("http"));
```

In the example above, the `frontend` resource would have two references. A reference would have:

* Resource name
* Endpoint name
* Value - URI string?

You could try to reverse engineer this by looking at environment variables, but it would be a pain in the butt. Much easier to model it explicitly.

drewnoakes:
Another potential addition is window size for logs. The server can limit the amount of history it sends easily enough, and I believe ACA does this today. The dashboard client currently appends logs without rolling old messages out of the window, which will eat memory indefinitely. We should synchronise the window size between client and server, so the server doesn't send data the client won't display, and so the client can roll old messages out. Perhaps the client specified the window size in the request.

davidfowl:
I'm going to punt this to the backlog for now.

davidfowl:
If anyone has cycles to work on this, I would like the status change for p5 😄

eerhardt:
@drewnoakes - this issue is marked as "Medium" priority in 8.2. Since it affects external resources from following security best practices, should this be moved higher and commited for 8.2?

drewnoakes:
[...]

PR #3210, opened 2024-03-27
Support custom styles for resource style

### LINKED ISSUE #2816: Dashboard "Removed" state still shows green status
We should probably change this to a warning or error status 
image

davidfowl:
- Added support for sending custom style to the dashboard via the resource server protocol.
- Use this for azure resource to show more descriptive state updates.

Contributes to https://github.com/dotnet/aspire/issues/2590
Contributes to https://github.com/dotnet/aspire/issues/2415
Fixes https://github.com/dotnet/aspire/issues/2816


src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -155,6 +155,11 @@ message Resource {
 
     // The list of urls that this resource exposes
     repeated Url urls = 13;
+
+    // The style of the state. This is used to determine the state icon.
+    // Supported styles are success", "warning", "info". Any other style
```
drewnoakes, line 160:
```suggestion
    // Supported styles are "success", "warning", "info". Any other style
```

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -155,6 +155,11 @@ message Resource {
 
     // The list of urls that this resource exposes
     repeated Url urls = 13;
+
+    // The style of the state. This is used to determine the state icon.
```
drewnoakes, line 159: Did you consider encoding a FluentUI icon name (e.g. `ErrorCircle`), style (e.g. `Filled`) and colour (e.g. `Blue`) here instead? That would give a lot more control over the UI to the server.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -152,6 +154,13 @@ await foreach (var kv in keyVaults.GetAllAsync(cancellationToken: cancellationTo
 
             if (keyVault is null)
             {
+                await notificationService.PublishUpdateAsync(resource, state => state with
+                {
+                    State = "Provisioning Keyvault",
+                    StateStyle = "info"
+
+                }).ConfigureAwait(false);
```
drewnoakes, line 162: To make consistent with the code below:

```suggestion
                })
                .ConfigureAwait(false);
```

drewnoakes (approved):

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
davidfowl: I would prefer that actually. So just parse the enum?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -104,11 +98,21 @@ async Task AfterProvisionAsync(IAzureResource resource)
...
+                    StateStyle = "success"
+                })
+                .ConfigureAwait(false);
             }
             catch (Exception)
             {
-                await SetStateAsync(resource, "FailedToStart").ConfigureAwait(false);
+                await UpdateStateAsync(resource, s => s with
+                {
+                    State = "Failed to Provision",
```
timheuer, line 112: later question: these likely need to be localized?

davidfowl: Yep.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
drewnoakes: Looks like they're not enums: https://github.com/microsoft/fluentui-blazor/tree/main/src/Assets/FluentUI.Icons/Icons

This method might help, though it doesn't look cheap so results should be cached:

https://github.com/microsoft/fluentui-blazor/blob/main/src/Assets/FluentUI.Icons/Icons.cs#L20

You pass it an instance of [`IconInfo`](https://github.com/microsoft/fluentui-blazor/blob/dev/src/Core/Components/Icons/IconInfo.cs) which takes name, size and variant (filled/regular).

davidfowl: I decided to stick with fixed names.

PR #3213, opened 2024-03-27
Add token auth to dashboard frontend

### LINKED ISSUE #3195: Embiggen API key
`Guid.NewGuid().ToString()` isn't secure enough. What API should be used to generate the right level of randomness?

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/3195

Initial code for token auth functionality in dashboard

* Add token page with input + button
* Clicking on the button invokes a JavaScript method that then calls an API which validates token and does the sign in. This is required because cookie authentication doesn't like signing in when Blazor is in interactive mode (i.e. SignalR)
* Configure auth and redirects to send unauthorized browsers to the token page, and then the token page redirects back to the app once the token is validated.

To do:

* ~Actually validate the token~
* ~An endpoint that support getting the token from a query string (IDE mode)~
* ~Protect navigating to dashboard pages after the token page has loaded.~
* ~Disable page shortcuts on the token page.~
* Provide accurate screenshot of token in console for tooltip info (alternatively, if we won't have that in time for this PR, disable the tooltip or remove the screenshot)
* ~Read incoming token from IDE tooling in host~
* ~Configure cookie auth length~
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3213)


JamesNK:
@tlmii I think should make changes to the token page directly to this PR for efficiency. What do you think?

davidfowl:
Screenshot/video of the flow?

tlmii:
> @tlmii I think you should make changes to the token page directly to this PR for efficiency. What do you think?

That seems reasonable. I did my first mockups last night. Chatting with some folks this morning, then I can push some changes

tlmii:
> Screenshot/video of the flow?

@davidfowl The full flow isn't implemented yet (still a few bullet points in the todo list) but here's a couple screenshots of what I just pushed for what we came up with for the page itself:

image
image

cc @kvenkatrajan @leslierichardson95

src/Aspire.Dashboard/Components/Layout/MainLayout.razor
```diff
@@ -31,11 +32,22 @@
             <FluentIcon Value="@(new Icons.Regular.Size24.Settings())" Color="Color.Neutral" />
         </FluentButton>
     </FluentHeader>
-    <NavMenu />
-    <FluentBodyContent Class="custom-body-content">
-        <FluentToastProvider />
-        @Body
-    </FluentBodyContent>
+
+    <AuthorizeView>
```
tlmii, line 36: We (at least initially) opted for a full-page login screen, so no header or anything either, so I just switched it to an empty layout template instead and didn't really make use of this. If we keep that other method of doing it, I guess this can be removed?

src/Aspire.Dashboard/Components/Pages/Token.razor
```diff
@@ -0,0 +1,40 @@
+﻿@page "/token"
+@layout EmptyLayout
+@attribute [AllowAnonymous]
+
+<PageTitle>Token!</PageTitle>
```
tlmii, line 5: I didn't touch the page title here yet since I had no great ideas on what it should say. @kvenkatrajan @leslierichardson95 ?

src/Aspire.Dashboard/Components/Layout/MainLayout.razor
leslierichardson95: Looks good so far!  I'm struggling with a good title currently but maybe something like ".NET Aspire Dashboard Login"?  It also might be good to have a header in the login prompting the user to input a token because it looks a little barren otherwise.

JamesNK:
> > Screenshot/video of the flow?

Way too purple IMO. It's a highlight color, not an everywhere color.

tlmii:
> > > Screenshot/video of the flow?
> 
> Way too purple IMO. It's a highlight color, not an everywhere color.

image
image

I forgot to change the color of the version info, will do that later.

src/Aspire.Dashboard/Components/Pages/Token.razor
tlmii: I changed this to ".NET Aspire Token Validation" at @leslierichardson95 's suggestion, but we're still open to feedback as to what this should say.

JamesNK:
This is the k8s dashboard sign in:

image

I think above the textbox we should have a title: "{ApplicationName} dashboard" e.g. "TestShop dashboard".

Then some text like:

> Please enter the dashboard token to log in. A token was generated when the dashboard started. See `Log in to the .NET Aspire dashboard` for more information.".

src/Aspire.Dashboard/Components/Pages/Token.razor.cs
```diff
@@ -51,6 +55,8 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
         if (firstRender)
         {
             _jsModule = await JS.InvokeAsync<IJSObjectReference>("import", "/Components/Pages/Token.razor.js");
+
+            _tokenTextField?.FocusAsync();
```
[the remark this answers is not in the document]
JamesNK: It doesn't return task?!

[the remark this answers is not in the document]
JamesNK: @vnbaaij Was this intentional?

src/Aspire.Dashboard/Components/Pages/Token.razor
```diff
@@ -0,0 +1,40 @@
+﻿@page "/token"
+@layout EmptyLayout
+@attribute [AllowAnonymous]
+
+<PageTitle>.NET Aspire Token Validation</PageTitle>
```
JamesNK, line 5: All other pages are "{ApplicationName} resources", "{ApplicationName} structured logs", etc.

IMO make this "{ApplicationName} log in"

drewnoakes:
The new UI is looking amazing. Great work.

> I think above the textbox we should have a title: "{ApplicationName} dashboard" e.g. "TestShop dashboard".

100%. Identifying it is useful and personalization is nice.

> Please enter the dashboard token to log in.

I like this change.

> Change "More info" to "See Log in to the .NET Aspire dashboard for more information."

I am in favour of this change in documentation, but not in UI. We want as little text as possible. Making the longest text on the screen one of the least important is distracting for a dialog box, where focus and flow is important.

JamesNK:
I guess it's pretty long.

This Windows login screen:

image

How about changing the link to be exactly why someone would click it: `How do I get the dashboard token?`

tlmii:
It's not evident from the screenshots above, but my intention with the (?) was that the answer to "How do I get the dashboard token?" would be right there without having to go to the documentation (the screenshot here is just a placeholder):

image

The "More Info" was only supposed to be for if you really wanted to dig in on why you're seeing it or how to change it to something else (or disable it).

src/Aspire.Dashboard/Components/Pages/Token.razor.cs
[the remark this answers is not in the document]
tlmii: Yeah that tripped me up at first, since ElementReference.FocusAsync _does_ return a task.

JamesNK:
> The "More Info" was only supposed to be for if you really wanted to dig in on why you're seeing it or how to change it to something else (or disable it).

Ok.

More feedback: Chabge `Submit` button text to `Log in`.

tlmii:
Another stab at it. Added header, instructional text, changed page title, tweaked a few layout things, localized the text, etc. Latest screenshots:

image
image

There's one part of the layout I'm really not fond of and I want to fix. But it'll look the same afterwards, so wanted to get this iteration out first.

JamesNK:
Feedback:

- I think the title/text should be left aligned with the text box
- I asked for the text, but now I see it, it feels a little unnecessary. Maybe removed it and have the placeholder updated to "Enter token to log in..."? That prompts the user and makes things a bit cleaner. What do you think?

tlmii:
> * Is the validation message style done? It's workable but feels rough.
> 
> Edit: I read https://www.nngroup.com/articles/errors-forms-design-guidelines/. How about placing the validation error immediately below the text box? When it appears the `More Info` link moves down to make space. Bonus points: A red exclamation point icon.

I was trying to find a way to make the validation message's arrival not change the size of the "dialog" and (probably more importantly) not move the Submit button. That's why I had it below everything. But I stopped focusing on that to address the other feedback. I'll take a look

JamesNK:
Yeah, I agree on not changing the dialog size or moving the button.

Can just more info move? Or maybe naturally have a little more vertical space between the textbox and more info/button. Then have the error message appear in the space between more info and the textbox. It would be a little cramped, but maybe that's ok for something that is just for showing errors.

tlmii:
Yeah, I'm working on a couple ideas. Probably won't have another round until the morning though.

src/Aspire.Dashboard/Components/Pages/Token.razor.cs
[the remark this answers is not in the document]
vnbaaij: It does not return a Task at the moment. It is on the list to change that, but as it will be a breaking change we decided to wait for a next major.

tlmii:
Thoughts?

image

(haven't pushed this yet, still tweaking, but wanted to send one last screenshot tonight)

Getting the triangle on the validation message is not the easy thing I expected - the visibility of validation-message isn't handled automatically or anything so I can't just drop it in there. Plus the two separate validation messages. I'll give it more thought after some rest.

tlmii:
Probably would have been helpful to note that the (?) icon is an actual link in that screenshot (replacing the more info button) rather than just the tool tip anchor. Not sure about it. But it makes things cleaner.

JamesNK:
I prefer the button below the text box. That is common in the examples I've looked at.

JamesNK:
> Getting the triangle on the validation message is not the easy thing I expected - the visibility of validation-message isn't handled automatically or anything so I can't just drop it in there. Plus the two separate validation messages. I'll give it more thought after some rest.

What about a pure CSS solution: https://stackoverflow.com/questions/19255296/is-there-a-way-to-use-svg-as-content-in-a-pseudo-element-before-or-after

vnbaaij:
May I suggest to use the `FluentValidationMessage` for the error state.

It automatically provides the exclamation mark error icon and uses the correct font size and spacing in accordance with the Fluent Design. Example from the demo site (https://www.fluentui-blazor.net/Forms)
image

JamesNK:
Recording of current end-to-end below. It demos:

* Login UI
* Printing login URL with token in app host
* Logging in with token in UI
* Logging in with URL in app host

browser-token

tlmii:
> May I suggest to use the `FluentValidationMessage` for the error state.
> 

Good call, I didn't think to check for that since the existing component was already there when I started to implement. I'll switch to that. Thanks!

JamesNK:
@tlmii Minor UI change while you're in the area: the input should be password rather than text.

tlmii:
Here's where we are now after more conversations with people and addressing the above as best I can. I also removed the second validation-message section and used the ValidationMessageStore for the post-submit errors, so the look and feel was consistent w/ the validation errors:

Default View:
image
image

Required validation failure:
image

Invalid token failure:
image

Where do I find my token flyout:
image

Long app names will truncate the header with ellipsis:
image

Long validation messages (if its a single long message, not if there are multiple) will wrap and increase the size of the login dialog. 
image
Note that I don't like that the (!) icon is centered to the entire block of text (rather than the first line). There doesn't seem to be a really straightforward way to change this. Setting `align-items: baseline` _sort of_ works, but it actually aligns it too high on the first line and looks wrong, especially when there's only a single line. I'm not going to spend any more time on that at the moment, as right now none of our validation messages come close to wrapping.

JamesNK:
@tlmii Don't worry about long validation messages. There are only two and they're both short.

JamesNK:
This is ready for review. Prioritize testing and looking at this please.

@mitchdenny @DamianEdwards @drewnoakes

src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+using Aspire.Dashboard.Configuration;
+using Aspire.Dashboard.Utils;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Options;
+using System.Web;
+
+namespace Aspire.Dashboard.Model;
+
+internal sealed class ValidateTokenMiddleware
```
DamianEdwards, line 14: Curious as to why this is a middleware and not just an authentication handler?

DamianEdwards: Can you log an issue to follow up later? I think we should be targeting using the framework primitives to implement this and if we wanted it as a new viable standalone authentication option for ASP.NET Core apps (as I know @davidfowl has mentioned) that would be a requirement.

src/Shared/TokenGenerator.cs
```diff
@@ -0,0 +1,38 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Security.Cryptography;
+
+namespace Aspire.Hosting;
+
+internal static class TokenGenerator
+{
+    public static string GetToken()
```
drewnoakes, line 10: `GetToken` makes it seem like you might get the same value if you called the method multiple times. Consider `CreateToken` or similar.

src/Shared/TokenGenerator.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        // Generate a 128-bit entropy token 
+        var tokenBytes = GenerateEntropyToken(16); // 16 bytes = 128 bits
```
drewnoakes, line 13: Even though the code is defined below, I find named literal arguments easier to read.

```suggestion
        var tokenBytes = GenerateEntropyToken(size: 16); // 16 bytes = 128 bits 
```

src/Shared/KnownConfigNames.cs
```diff
@@ -5,8 +5,10 @@ namespace Aspire.Hosting;
 
 internal static class KnownConfigNames
```
drewnoakes, line 6: Unrelated to this PR, but this class name seems misleading. These are env var names, not config keys.

src/Shared/StringUtils.cs
```diff
@@ -7,7 +7,7 @@ namespace Aspire.Hosting.Utils;
 
 internal static class StringUtils
 {
-    public static bool TryGetUriFromDelimitedString(string input, string delimiter, [NotNullWhen(true)] out Uri? uri)
+    public static bool TryGetUriFromDelimitedString(string? input, string delimiter, [NotNullWhen(true)] out Uri? uri)
```
drewnoakes, line 10: Can also annotate `input`:

```suggestion
    public static bool TryGetUriFromDelimitedString([NotNullWhen(true)] string? input, string delimiter, [NotNullWhen(true)] out Uri? uri)
```

tests/Aspire.Hosting.Tests/ProjectResourceTests.cs
```diff
@@ -282,10 +284,18 @@ public async Task VerifyManifest(bool disableForwardedHeaders)
         Assert.Equal(expectedManifest, manifest.ToString());
     }
 
-    private static IDistributedApplicationBuilder CreateBuilder(DistributedApplicationOperation operation = DistributedApplicationOperation.Publish)
+    private static IDistributedApplicationBuilder CreateBuilder(string[]? args = null, DistributedApplicationOperation operation = DistributedApplicationOperation.Publish)
     {
-        var args = operation == DistributedApplicationOperation.Publish ? new[] { "--publisher", "manifest" } : Array.Empty<string>();
-        var appBuilder = DistributedApplication.CreateBuilder(args);
+        var resovledArgs = new List<string>();
```
drewnoakes, line 289: nit: typo

```suggestion
        var resolvedArgs = new List<string>();
```

drewnoakes (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -780,10 +780,19 @@ private async Task StartDashboardAsDcpExecutableAsync(CancellationToken cancella
             KeyValuePair.Create(DashboardConfigNames.ResourceServiceUrlName.EnvVarName, resourceServiceUrl),
             KeyValuePair.Create(DashboardConfigNames.DashboardOtlpUrlName.EnvVarName, otlpEndpointUrl),
             KeyValuePair.Create(DashboardConfigNames.ResourceServiceAuthModeName.EnvVarName, "Unsecured"),
-            KeyValuePair.Create(DashboardConfigNames.DashboardFrontendAuthModeName.EnvVarName, "Unsecured"),
         };
 
-        if (configuration["AppHost:OtlpApiKey"] is { } otlpApiKey)
+        if (configuration["AppHost:BrowserToken"] is { Length: > 0 } browserToken)
```
mitchdenny, line 785: Should we use IOptions<T> for this.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -80,12 +78,40 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
...
 
+        // Set configuration
+        ConfigurePublishingOptions(options);
         _innerBuilder.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
         {
             // Make the app host directory available to the application via configuration
             ["AppHost:Directory"] = AppHostDirectory
         });
 
+        if (!options.DisableDashboard && !IsDashboardUnsecured(_innerBuilder.Configuration))
```
mitchdenny, line 89: Maybe extract this into its own method as well?

src/Shared/KnownConfigNames.cs
mitchdenny: Debatable :) They are perfectly valid config keys. All environment variable names are valid config key names.

src/Aspire.Hosting/DistributedApplicationLifecycle.cs
```diff
@@ -17,6 +21,11 @@ public Task StartAsync(CancellationToken cancellationToken)
 
     public Task StartedAsync(CancellationToken cancellationToken)
     {
+        if (distributedApplicationOptions.DashboardEnabled && configuration["AppHost:BrowserToken"] is { Length: > 0 } browserToken)
```
mitchdenny, line 24: I thought that VS would still continue to look for the existing `Now listening on:` message. I'm assuming based on this that they've agreed to look for the extra message? This wouldn't work for `dotnet watch` for example?

src/Aspire.Dashboard/Configuration/PostConfigureDashboardOptions.cs
```diff
@@ -35,5 +35,9 @@ public void PostConfigure(string? name, DashboardOptions options)
             options.Frontend.AuthMode = FrontendAuthMode.Unsecured;
             options.Otlp.AuthMode = OtlpAuthMode.Unsecured;
         }
+        if (options.Frontend.AuthMode == FrontendAuthMode.BrowserToken && string.IsNullOrEmpty(options.Frontend.BrowserToken))
+        {
+            options.Frontend.BrowserToken = TokenGenerator.GetToken();
```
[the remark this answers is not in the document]
JamesNK: This is planning for the future (frontend defaults to browser token auth). Will remove from this PR.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -105,9 +105,13 @@ internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
 public sealed class FrontendOptions
 {
     private List<Uri>? _parsedEndpointUrls;
+    private byte[]? _browserTokenBytes;
 
     public string? EndpointUrls { get; set; }
     public FrontendAuthMode? AuthMode { get; set; }
+    public string? BrowserToken { get; set; }
+
+    public byte[]? GetBrowserTokenBytes() => _browserTokenBytes;
```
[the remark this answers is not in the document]
JamesNK: The options should only be set by a framework and unit tests. I'll add a comment.

src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                }
+                else
+                {
+                    // Failure. Remove the token from the query string.
+                    var qs = HttpUtility.ParseQueryString(context.Request.QueryString.ToString());
+                    qs.Remove("t");
+
+                    var newQuerystring = qs.ToString();
+                    context.Response.Redirect($"{context.Request.Path}?{newQuerystring}");
```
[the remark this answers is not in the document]
JamesNK: I think this is fine. If what is matched changes then this will just keep working.

src/Aspire.Dashboard/Utils/CompareHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System.Buffers;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Aspire.Dashboard.Utils;
+
+internal static class CompareHelpers
+{
+    // This method is used to compare two keys in a way that avoids timing attacks.
```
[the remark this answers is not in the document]
JamesNK: It doesn't need to be random, just constant. This should be fine.

To be honest I don't know if a timing attack is possible against such a small value when the overhead of a HTTP request is so large, but just being thorough.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
JamesNK: This is fine. They're separate concerns.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
JamesNK: I agree. I didn't want to rewrite the host in this PR.

src/Aspire.Dashboard/Authentication/OtlpApiKey/OtlpApiKeyAuthenticationHandler.cs
```diff
@@ -34,9 +32,9 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
                 return Task.FromResult(AuthenticateResult.Fail($"Multiple '{ApiKeyHeaderName}' headers in request."));
             }
 
-            if (!CompareApiKey(options.GetPrimaryApiKeyBytes(), apiKey.ToString()))
+            if (!CompareHelpers.CompareKey(options.GetPrimaryApiKeyBytes(), apiKey.ToString()))
             {
-                if (options.GetSecondaryApiKeyBytes() is not { } secondaryBytes || !CompareApiKey(secondaryBytes, apiKey.ToString()))
+                if (options.GetSecondaryApiKeyBytes() is not { } secondaryBytes || !CompareHelpers.CompareKey(secondaryBytes, apiKey.ToString()))
```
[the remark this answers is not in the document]
JamesNK: Why should it?

src/Shared/KnownConfigNames.cs
JamesNK: I agree with Drew. And this file should be specific to hosting. Dashboard [has its own file](https://github.com/dotnet/aspire/blob/f0494ca5691ce279180fb284df32e582bf335834/src/Shared/DashboardConfigNames.cs). Host has it. If some are duplicated then that's fine. The two apps just happen to have some config settings with the same name.

src/Aspire.Hosting/DistributedApplicationLifecycle.cs
JamesNK: VS isn't looking for this. It is hard coding appending login to the URL. And it already knows the token.

JamesNK:
@tlmii Clicking on the help link on the login page takes you to a new page that shows "Unauthorized". Need to disable that.

src/Shared/KnownConfigNames.cs
mitchdenny: I was talking about them being env var names ;)

mitchdenny (approved):

src/Aspire.Dashboard/Components/Pages/Login.razor
```diff
@@ -0,0 +1,49 @@
...
+@inject IStringLocalizer<Dashboard.Resources.Login> Loc
+
+<PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.Login.PageTitle)" Loc="@Loc" /></PageTitle>
+
+<div class="token-backdrop">
+    <EditForm EditContext="@EditContext" OnValidSubmit="@SubmitAsync" FormName="token_validation">
+        <DataAnnotationsValidator />
+        <div class="token-form-container">
+            <div class="token-logo">
+                <AspireLogo Height="128" Width="128" />
```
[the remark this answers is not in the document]
tlmii: Safari has issues with displaying our logo svg as a downloaded file. When we moved to embedding it as an icon like this for the header, it fixed the issue. So I stuck with it. It's actually the same size (svg content) for the header, which since that's on all requests might be a bigger concern. I'll file an issue to see if there's any way to optimize the size but I know the designers are incredibly busy at the moment so I don't expect it'll be handled by them soon.

src/Aspire.Hosting/DistributedApplicationLifecycle.cs
DamianEdwards: The question about `dotnet watch` is relevant though as that definitely watches console output to know when to launch the browser.

tlmii:
> @tlmii Clicking on the help link on the login page takes you to a new page that shows "Unauthorized". Need to disable that.

Fixed. That was not the behavior I expected for a link to `#` on a non-root page.

src/Aspire.Hosting/DistributedApplicationLifecycle.cs
JamesNK: It will continue to look for `Now listening on:`.

src/Aspire.Dashboard/Utils/CompareHelpers.cs
```diff
@@ -0,0 +1,55 @@
...
+            stackalloc byte[StackAllocThreshold] :
+            (requestPooled = RentClearedArray(requestSpanLength))).Slice(0, requestSpanLength);
+
+        try
+        {
+            // Always succeeds because the byte span is always as big or bigger than required.
+            Encoding.UTF8.GetBytes(requestKey, requestBytesSpan);
+
+            // Trim request bytes to the same length as expected bytes. Need to be the same size for fixed time comparison.
+            var equals = CryptographicOperations.FixedTimeEquals(expectedKeyBytes, requestBytesSpan.Slice(0, expectedKeyBytes.Length));
```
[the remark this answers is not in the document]
JamesNK: stackalloc always sets data to zero unless the skip initialization attribute is added to the method. Or am I wrong?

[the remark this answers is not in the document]
JamesNK: Either way, after changes, all bytes in the sliced span (either from stackalloc or rented array) are overwritten by GetBytes call.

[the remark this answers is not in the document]
GrabYourPitchforks: > stackalloc always sets data to zero unless the skip initialization attribute is added to the method/assembly. Or am I wrong?

You are correct. However, the dotnet org is lamentably very aggressive about applying this attribute, so we can't assume it won't be added to this project in vNext.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -448,6 +491,14 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
                             .RequireAuthenticatedUser()
                             .Build());
                     break;
+                case FrontendAuthMode.BrowserToken:
+                    options.AddPolicy(
+                        name: FrontendAuthorizationDefaults.PolicyName,
+                        policy: new AuthorizationPolicyBuilder(
+                            FrontendAuthenticationDefaults.AuthenticationScheme)
+                            .RequireAuthenticatedUser()
```
GrabYourPitchforks, line 499: Is an auth cookie the only way for this condition to be fulfilled? It's an ironclad guarantee that no other auth mechanism (NTLM, client cert, whatever) can fulfill this condition?

src/Aspire.Dashboard/wwwroot/img/TokenExample.png
drewnoakes, line 1: This file can be compressed from 9kB to 7kB using https://tinypng.com.

tlmii: We can optimize after we get the final screenshot. That was just a placeholder. See #3348.

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: Hmm, I'm not 100% sure. I think the way to be sure would be to have a specific claim added by cookies auth when mode is BrowserToken and then requiring it here.

What do you think?

JamesNK: I've updated the authz policy to require a specific claim that is only set when cookie auth is signed in via a browser token.

Issue #3190, opened 2024-03-26
[WebToolsE2E] With no Docker or Podman installed, create Aspire Starter App with redis cache option and run, the error message calls out Docker not found but doesn't mention Podman


balachir:
**INSTALL STEPS**
1. Clean machine: Win11 x64 23h2 ENU
2. Install Dev17.10.0 Preview 3.0 [main-34726.61] (Includes Aspire 8.0.0-preview.4.24156.9)
3. Install Aspire 8.0 P5 version 8.0.0-preview.5.24175.10 + Apply feeds
4. No Docker or Podman installed

**REPRO STEPS**
1. Create Aspire Starter App using command-line with redis cache option
    > dotnet new aspire-starter -o AspireStarterApp03withRedis --use-redis-cache

2. Navigate to AppHost folder and run the app
    > dotnet run

**ACTUAL**
You get the following error, which is expected. But the error is specific about Docker not being installed.

**Question**: Should the error be more general to indicate that either Docker or Podman needs to be installed.

![image](https://github.com/dotnet/aspire/assets/8246794/9c308f1a-f944-4ebf-8cef-52e66ec8997c)

**Notes**
With VS New Project dialog (NPD) too, the checkbox in NPD is specific about recommending Docker. See **AzDO 2004980**

![image](https://github.com/dotnet/aspire/assets/8246794/f9d0a2ca-8bd7-4462-8907-d043f195ebf5)

balachir:
@DamianEdwards @davidfowl for scenarios like these, are we fine with recommending Docker or should the messaging be more general and recommend either Docker or Podman?

cc: @karolz-ms

DamianEdwards:
It should be generalized to something like "a supported container runtime"

timheuer:
> It should be generalized to something like "a supported container runtime"

David Pine just updated the doc point here about supported container runtime as well -- but I think the tool tip in VS doesn't allow hyperlink directly but we should put a readable aka.ms link (instead of FWLink)

v-cazhang:
In AzDO bug 2004980, Vijay mentioned that:

This is in the templates:
https://github.com/dotnet/aspire/blob/ac2b068eb10096852a5e48d01615d9effa3331f2/src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json#L84-L85

DamianEdwards:
Yep I'm fixing this

DamianEdwards:
Fixed

[...]

PR #3224, opened 2024-03-27
Update container runtime not found exception message

DamianEdwards:
#3190
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3224)

### DIFF
--- a/src/Aspire.Hosting/Properties/Resources.Designer.cs
+++ b/src/Aspire.Hosting/Properties/Resources.Designer.cs
@@ -79,7 +79,8 @@ namespace Aspire.Hosting.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Container runtime &apos;{0}&apos; could not be found. The error from the container runtime check was: {1}..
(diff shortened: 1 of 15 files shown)


timheuer (approved):
lgtm

balachir:
@DamianEdwards for the scenario described in https://github.com/dotnet/aspire/issues/3190, with neither docker or podman installed, with your change, the exception message will still output `Container runtime 'docker' could not be found`. But now, there will be the additional aka.ms link with more info. Correct?

How does the '{0}' in the error message resolve to 'docker'? Is that because DCP tries to default to docker?

balachir (approved):

DamianEdwards:
@balachir From my understanding, DCP attempts to auto-detect which container runtime is installed and use that, but in the case where the error coming back doesn't have a container runtime name, `Aspire.Hosting` assumes Docker as the name and then adds the link now.

PR #3229, opened 2024-03-27
Retry on YamlException when loading Kubernetes client configuration

ReubenBond:
I encountered this while running an integration test:
image

My assumption is that this is caused by a race between writing and ready the Kubernetes client configuration file. By handling the resulting `YamlException`, we allow the process to be retried.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3229)

### DIFF
--- a/src/Aspire.Hosting/Dcp/KubernetesService.cs
+++ b/src/Aspire.Hosting/Dcp/KubernetesService.cs
@@ -297,6 +297,8 @@ internal sealed class KubernetesService(ILogger<KubernetesService> logger, IOpti
         {
             var configurationReadRetry = new RetryStrategyOptions()
             {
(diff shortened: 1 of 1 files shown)


davidfowl (approved):
Add a comment

ReubenBond:
Done: 
```csharp
// Handle exceptions caused by races between writing and reading the configuration file.
// If the file is loaded while it is still being written, this can result in a YamlException being thrown.
ShouldHandle = new PredicateBuilder().Handle<KubeConfigException>().Handle<YamlException>(),
```

davidfowl:
@joperezr why does the bot not like @ReubenBond ?

PR #3230, opened 2024-03-27
Add WithManagementPlugin() method for RabbitMQ management plugin

### LINKED ISSUE #2067: Add method to enable the RabbitMQ management plugin
Comment says that tag is `3-management` but we are just using `3`: https://github.com/dotnet/aspire/blob/73b9e0bede781852e6a980284b22b0003ec1591e/src/Aspire.Hosting/RabbitMQ/RabbitMQBuilderExtensions.cs#L16

Should we be using `3-management`? (note: Visual Studio connected services uses `rabbitmq:3-management-alpine` for local dev)

DamianEdwards:
This change adds a `WithManagementPlugin()` extension method for the RabbitMQ resource to update the container image with the management plugin enabled and expose its HTTP endpoint. The method checks that the image name and tag are known and supported before updating the tag to enable the management plugin, otherwise it throws.

It also updates the doc comments to better match the format used for the Redis resource.

Fixes #2067

image

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3230)

### DIFF
--- a/src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
+++ b/src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
@@ -45,7 +45,7 @@ public static class RabbitMQBuilderExtensions
         if (builder.ExecutionContext.IsRunMode)
         {
             // Configure for management plugin
-            rabbitmq.WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.TagMangement)
+            rabbitmq.WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.TagManagement)
                     .WithHttpEndpoint(containerPort: 15672, name: RabbitMQServerResource.ManagementEndpointName);
         }
         else
--- a/src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
+++ b/src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
@@ -16,7 +16,7 @@ public static class RabbitMQBuilderExtensions
     /// Adds a RabbitMQ container to the application model.
     /// </summary>
     /// <remarks>
-    /// The default image is "rabbitmq". The default tag is "3-management" when running and "3" when publishing.
+    /// The default image and tag are "rabbitmq" and "3".
     /// </remarks>
     /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
     /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
@@ -35,23 +35,13 @@ public static class RabbitMQBuilderExtensions
 
         var rabbitMq = new RabbitMQServerResource(name, userName?.Resource, passwordParameter);
         var rabbitmq = builder.AddResource(rabbitMq)
+                              .WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.Tag)
                               .WithEndpoint(hostPort: port, containerPort: 5672, name: RabbitMQServerResource.PrimaryEndpointName)
                               .WithEnvironment(context =>
                               {
                                   context.EnvironmentVariables["RABBITMQ_DEFAULT_USER"] = rabbitMq.UserNameReference;
                                   context.EnvironmentVariables["RABBITMQ_DEFAULT_PASS"] = rabbitMq.PasswordParameter;
                               });
-
-        if (builder.ExecutionContext.IsRunMode)
-        {
-            // Configure for management plugin
-            rabbitmq.WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.TagManagement)
-                    .WithHttpEndpoint(containerPort: 15672, name: RabbitMQServerResource.ManagementEndpointName);
-        }
-        else
-        {
-            rabbitmq.WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.Tag);
-        }
 
         return rabbitmq;
     }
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting.RabbitMQ/RabbitMQServerResource.cs
```diff
@@ -9,6 +9,7 @@ namespace Aspire.Hosting.ApplicationModel;
 public class RabbitMQServerResource : ContainerResource, IResourceWithConnectionString, IResourceWithEnvironment
 {
     internal const string PrimaryEndpointName = "tcp";
+    internal const string MangementEndpointName = "management";
```
eerhardt, line 12:
```suggestion
    internal const string ManagementEndpointName = "management";
```

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -30,14 +34,26 @@ public static class RabbitMQBuilderExtensions
...
+                              {
+                                  context.EnvironmentVariables["RABBITMQ_DEFAULT_USER"] = rabbitMq.UserNameReference;
+                                  context.EnvironmentVariables["RABBITMQ_DEFAULT_PASS"] = rabbitMq.PasswordParameter;
+                              });
+
+        if (builder.ExecutionContext.IsRunMode)
+        {
+            // Configure for management plugin
+            rabbitmq.WithImage(RabbitMQContainerImageTags.Image, RabbitMQContainerImageTags.TagMangement)
+                    .WithHttpEndpoint(containerPort: 15672, name: RabbitMQServerResource.MangementEndpointName);
```
eerhardt, line 49:
```suggestion
                    .WithHttpEndpoint(containerPort: 15672, name: RabbitMQServerResource.ManagementEndpointName);
```

tests/Aspire.Hosting.Tests/RabbitMQ/AddRabbitMQTests.cs
```diff
@@ -24,18 +24,27 @@ public void AddRabbitMQContainerWithDefaultsAddsAnnotationMetadata()
...
 
-        var endpoint = Assert.Single(containerResource.Annotations.OfType<EndpointAnnotation>());
-        Assert.Equal(5672, endpoint.ContainerPort);
-        Assert.False(endpoint.IsExternal);
-        Assert.Equal("tcp", endpoint.Name);
-        Assert.Null(endpoint.Port);
-        Assert.Equal(ProtocolType.Tcp, endpoint.Protocol);
-        Assert.Equal("tcp", endpoint.Transport);
-        Assert.Equal("tcp", endpoint.UriScheme);
+        var primaryEndpoint = Assert.Single(containerResource.Annotations.OfType<EndpointAnnotation>().Where(e => e.Name == "tcp"));
```
eerhardt, line 27: Should we make this test be a Theory and run it for both "RunMode" and "PublishMode"?

eerhardt (approved):

src/Aspire.Hosting.RabbitMQ/RabbitMQContainerImageTags.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.RabbitMQ;
+
+internal static class RabbitMQContainerImageTags
+{
+    public const string Image = "rabbitmq";
+    public const string Tag = "3";
+    public const string TagMangement = $"{Tag}-management";
```
attilah, line 10: `TagMangement` typo here as well.

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -30,14 +34,26 @@ public static class RabbitMQBuilderExtensions
...
-                      });
+        var rabbitmq = builder.AddResource(rabbitMq)
+                              .WithEndpoint(hostPort: port, containerPort: 5672, name: RabbitMQServerResource.PrimaryEndpointName)
+                              .WithEnvironment(context =>
+                              {
+                                  context.EnvironmentVariables["RABBITMQ_DEFAULT_USER"] = rabbitMq.UserNameReference;
+                                  context.EnvironmentVariables["RABBITMQ_DEFAULT_PASS"] = rabbitMq.PasswordParameter;
+                              });
+
+        if (builder.ExecutionContext.IsRunMode)
```
timheuer, line 45: It does feel like this is a diff approach than other local management experiences and implicit instead of explicit (even in run mode). e.g. why would we just add pgadmin when in runmode then? Even though it isn't a separate management container, I feel like we should model it as such.

src/Aspire.Hosting.RabbitMQ/RabbitMQServerResource.cs
DamianEdwards: doh

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
DamianEdwards: Because it's in the same container it makes composition of the various container extension methods much harder to reason about. When it's a separate container instance we don't have to worry so much about whether the tag or whole image has been changed on the resource being extended. If we wanted to model this as a separate method we'd need to think about how we handle situations where the main resource's image or tag aren't the default.

DamianEdwards: Spoke to @davidfowl about this some more and I'm going to make the change to put this in a separate method and handle the known tag formats. If you call it and the container image registry, name or tag is of an unsupported value it'll throw.

PR #3233, opened 2024-03-27
Rename config value to use unsecured instead of insecure

JamesNK:
We're using unsecured everywhere else. And this setting defaults the frontend and OTLP endpoint auth modes to unsecured. Seems like it should be consistent.

If merged, I'll backport to P5.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3233)


drewnoakes (approved):
LGTM

PR #3235, opened 2024-03-28
Dashboard should request log streams only when the user gestures to display them

### LINKED ISSUE #2789: Dashboard should request log streams only when the user gestures to display them (for given resource)
With https://github.com/dotnet/aspire/pull/2731/ the dashboard requests log streams for all application resources at the startup or shortly after. This is not efficient, especially for Containers, where each log stream is associated with a separate container orchestrator (CLI) running in the background.

eerhardt:
Change the AppHost to only subscribe to DCP's logs when there is a subscriber for the logs.

Fix #2789
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3235)


JamesNK:
What happens to historical logs? For example, will requesting the console logs for an app show all longs since the app started? (if less than the limit)

eerhardt:
> What happens to historical logs? For example, will requesting the console logs for an app show all longs since the app started? (if less than the limit)

The "backlog" gets cleared after the last subscriber to a resource leaves. Later, when a "new" subscriber comes along, they are all replayed from DCP because we call `StartLogStream(executable);` again.

So even if no one was subscribed for a while, the whole logs come back once someone does.

davidfowl:
One small race is that starting logs before the container is ready, it won’t show up. Super tight race and I don’t know if it’ll be worth complicating the code but I thought I’d let you know

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -348,6 +331,21 @@ void Log(LogLine log)
 
                 state.OnNewLog += Log;
 
+                // ensure the backlog snapshot is taken after subscribing to OnNewLog
+                // to ensure the backlog snapshot contains the correct logs. The backlog
+                // gets cleared when there are no subscribers, so we ensure we are subscribing first.
+                LogLine[] backlogSnapshot;
+                lock (state._backlog)
```
davidfowl, line 338: You can’t lock in this method. There’s a compiler bug allowing this currently https://github.com/dotnet/roslyn/issues/72443

eerhardt: I think that bug is different. In that bug it is `yield return`ing from inside the `lock`. Here we aren't doing anything async in the lock, we are just making the snapshot.

I can move this out into a synchronous method (which should obviously be fine being called from an async iterator.

eerhardt:
> One small race is that starting logs before the container is ready, it won’t show up.

Can you describe this a bit more? When what is starting the logs before the container is ready?

eerhardt:
> One small race is that starting logs before the container is ready, it won’t show up. Super tight race and I don’t know if it’ll be worth complicating the code but I thought I’d let you know

This didn't turn out to be too bad. I added another loop that joined the "has subscribers" and "logs available" information, and only starts the logs once both are true.

tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
```diff
@@ -48,7 +48,7 @@ async Task<List<ResourceEvent>> GetValuesAsync(CancellationToken cancellationTok
         {
             var values = new List<ResourceEvent>();
 
-            await foreach (var item in notificationService.WatchAsync().WithCancellation(cancellationToken))
+            await foreach (var item in notificationService.WatchAsync(cancellationToken))
```
davidfowl, line 51: Is this right?

davidfowl: Looked at the IL

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -195,6 +199,70 @@ private void WatchResourceChanges(CancellationToken cancellationToken)
...
+            await foreach (var entry in _logInformationChannel.Reader.ReadAllAsync(cancellationToken))
+            {
+                var logsAvailable = false;
+                var hasSubscribers = false;
+                if (resourceLogState.TryGetValue(entry.ResourceName, out (bool, bool) stateEntry))
+                {
+                    (logsAvailable, hasSubscribers) = stateEntry;
+                }
+
+                logsAvailable = entry.LogsAvailable ?? logsAvailable;
```
davidfowl, line 228: Once this is true, it'll always to be true, maybe add a comment.

davidfowl (approved):

PR #3243, opened 2024-03-28
Improve writing dashboard startup config failure messages

### LINKED ISSUE #3168: Make dashboard print a good error message when configuration is invalid
See https://github.com/dotnet/aspire/pull/3119#discussion_r1538599584

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/3168

* Write configuration errors to logging instead of throwing exception
* Add missing validation of frontend auth mode

Before:

image

After:

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3243)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Net;
 using System.Reflection;
 using System.Security.Claims;
@@ -33,7 +35,9 @@ public sealed class DashboardWebApplication : IAsyncDisposable
     internal const string DashboardUrlDefaultValue = "http://localhost:18888";
 
     private readonly WebApplication _app;
+    private readonly ILogger<DashboardWebApplication> _logger;
     private readonly IOptionsMonitor<DashboardOptions> _dashboardOptionsMonitor;
+    private readonly IReadOnlyList<string> _validationFailures;
     private Func<EndpointInfo>? _frontendEndPointAccessor;
     private Func<EndpointInfo>? _otlpServiceEndPointAccessor;
 
@@ -48,6 +52,8 @@ public sealed class DashboardWebApplication : IAsyncDisposable
     }
 
     public IOptionsMonitor<DashboardOptions> DashboardOptionsMonitor => _dashboardOptionsMonitor;
+
+    public IReadOnlyList<string> ValidationFailures => _validationFailures;
 
     /// <summary>
     /// Create a new instance of the <see cref="DashboardWebApplication"/> class.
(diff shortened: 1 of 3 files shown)


JamesNK:
@mitchdenny What does the host do on configuration or startup failure? We should think about consistency.

davidfowl:
> @mitchdenny What does the host do on configuration or startup failure? We should think about consistency.

Can you do it in both places?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -224,22 +229,53 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
...
+    private static void HandleConfigurationError(ILogger<DashboardWebApplication> logger, string[] failureMessages)
+    {
+        WriteVersion(logger);
+        logger.LogError("Failed to start the dashboard due to {Count} configuration error(s).", failureMessages.Length);
+        foreach (var message in failureMessages)
+        {
+            logger.LogError("{ErrorMessage}", message);
+        }
+
+        Environment.Exit(-1);
```
davidfowl, line 241: I don't think we want to do this from the host though. It's a library, not an executable.

JamesNK: I don't understand. Would you prefer to set the exit code in `Program.cs` instead of `DashboardWebApplication.cs`?

JamesNK: Or are you saying that if we have similar code for printing out config errors in the host then it shouldn't Environment.Exit?

davidfowl: > Or are you saying that if we add similar code for printing out config errors in the host then it shouldn't Environment.Exit?

This, but we need to unwind.

Since the dashboard owns main, this should happen in main.

JamesNK: I agree. Having it where it currently is caused test failures anyway.

mitchdenny:
This is the way we do it in the AppHost:

https://github.com/dotnet/aspire/blob/9c4533e939faafbc26202694808e37039e22d835/src/Aspire.Hosting/DistributedApplicationBuilder.cs#L98-L99

JamesNK:
I want to register a type. I don't want to use a callback.

JamesNK:
Review please 🙏

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -17,14 +17,23 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
...
-        if (!options.Otlp.TryParseOptions(out var otlpParseErrorMessage))
+        switch (options.Frontend.AuthMode)
         {
-            errorMessages.Add(otlpParseErrorMessage);
+            case FrontendAuthMode.Unsecured:
+                break;
+            case FrontendAuthMode.OpenIdConnect:
+                break;
+            case null:
+                errorMessages.Add($"Frontend endpoint authentication is not configured. Either specify {DashboardConfigNames.DashboardUnsecuredAllowAnonymousName.ConfigKey}=true, or specify {DashboardConfigNames.DashboardFrontendAuthModeName.ConfigKey}. Possible values: {string.Join(", ", typeof(FrontendAuthMode).GetEnumNames())}");
```
drewnoakes, line 27: Error messages should be localised.

src/Aspire.Dashboard/Program.cs
```diff
@@ -3,5 +3,11 @@
...
-// TODO potentially inline DashboardWebApplication in this file
-new DashboardWebApplication().Run();
+var app = new DashboardWebApplication();
+if (app.ValidationFailures.Count > 0)
+{
+    return -1;
+}
+
+app.Run();
+return 0;
```
drewnoakes, line 13: Just make the `Run` method return `int`?

drewnoakes (approved):

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
JamesNK: If we want to do that then create a new issue. It's beyond the scope of this PR.

PR #3251, opened 2024-03-28
Automatically AddAzureProvisioning when Azure resource type is used.

mitchdenny:
Partially addresses https://github.com/dotnet/aspire/issues/2590

This PR will automatically call `AddAzureProvisioning`.

This PR also improves the error message if there is a provisioning issue due to missing configuration data.

This is what you will now see if a resource fails to provision because it is missing subscription information:

image

Resources that depend on the resource via enviroment variables have also had some improvements to their error logging:

image

I haven't removed the exceptions completely on the dependent resources because there are some cases where this could obscure a real issue. So if you turn on debug level logging for DCP you can get the full exception shown (example):

image

I do not put this debug information in the dashboard (its in the console). This is because the ResourceLoggerService infrastructure doesn't really honor any config settings for controlling verbosity levels. We should probably decide what we want to do here around controlling verbosity of resource loggers.

### DIFF
--- a/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
+++ b/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
@@ -59,7 +59,7 @@ public static class AzurePostgresExtensions
             postgres.AssignProperty(x => x.Sku.Tier, "'Burstable'");
             postgres.AssignProperty(x => x.Version, "'16'");
             postgres.AssignProperty(x => x.HighAvailability.Mode, "'Disabled'");
-            postgres.AssignProperty(x => x.StorageSizeInGB, "32");
+            postgres.AssignProperty(x => x.Storage.StorageSizeInGB, "32");
             postgres.AssignProperty(x => x.Backup.BackupRetentionDays, "7");
             postgres.AssignProperty(x => x.Backup.GeoRedundantBackup, "'Disabled'");
             postgres.AssignProperty(x => x.AvailabilityZone, "'1'");
--- a/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
+++ b/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
@@ -47,6 +47,8 @@ public static class AzurePostgresExtensions
         Action<IResourceBuilder<AzurePostgresResource>, ResourceModuleConstruct, PostgreSqlFlexibleServer>? configureResource,
         bool useProvisioner = false)
     {
+        builder.ApplicationBuilder.AddAzureProvisioning();
+
         var configureConstruct = (ResourceModuleConstruct construct) =>
         {
             var administratorLogin = new Parameter("administratorLogin");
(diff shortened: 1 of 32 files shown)


src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -57,7 +59,7 @@ public static class AzurePostgresExtensions
             postgres.AssignProperty(x => x.Sku.Tier, "'Burstable'");
             postgres.AssignProperty(x => x.Version, "'16'");
             postgres.AssignProperty(x => x.HighAvailability.Mode, "'Disabled'");
-            postgres.AssignProperty(x => x.Storage.StorageSizeInGB, "32");
```
davidfowl, line 60: Intentional?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -123,8 +123,11 @@ public override async Task GetOrCreateResourceAsync(AzureBicepResource resource,
 
         PopulateWellKnownParameters(resource, context);
 
-        var azPath = FindFullPathFromPath("az") ??
-            throw new InvalidOperationException("Azure CLI not found in PATH");
+        if (FindFullPathFromPath("az") is not { } azPath)
+        {
+            resourceLogger.LogCritical("Using Azure resources during local development requires the installation of the Azure CLI. See https://aka.ms/dotnet/aspire/azcli for instructions.");
```
davidfowl, line 128: If we don't throw, what happens to the provisioning status? Does it end up successful?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1144,17 +1145,27 @@ private async Task CreateExecutableAsync(AppResource er, ILogger resourceLogger,
...
-                if (value is not null)
+                    if (value is not null)
+                    {
+                        spec.Args.Add(value);
+                    }
+                }
+                catch (Exception ex)
                 {
-                    spec.Args.Add(value);
+                    resourceLogger.LogCritical("Failed to apply arguments. A dependency may have failed to start.");
```
davidfowl, line 1165: Why not throw a more specific exception and let this bubble out? That way you can handle it in a single place.

davidfowl:
We need AddBicep** and AddAzureConstruct to also call AddAzureProvisioning

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
mitchdenny: No ... side effect of trying 0.1.0-beta.1 of Azure.Provisioning. Will change that back.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: Its less interesting for args, but for environment variables I'm trying to log each environment variable that had trouble being injected. I figured a little bit of duplication here led to a better experience. But I could probably capture that in the exception.

davidfowl: So you want to keep trying as each one fails instead of failing on the first one?

davidfowl: What if we log from here but then throw at the end if any failed to be applied?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
mitchdenny: Fixed.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: Fixed.

davidfowl (approved):

PR #3260, opened 2024-03-28
Write ports to the manifest

davidfowl:
In order for tools to better describe multiple endpoints int the manifest, we must include port information. At runtime most of the port information is dynamically generated and thus not described (for anything but containers), that doesn't work well when trying to deploy to various environments. We allocate ports in situations where there are none to match the runtime behavior.

The exposed port (**port** in the manifest) is the port used to talk to this binding.
The target port (**containerPort** in the manifest) is the port the resource binds to.

Rules:
~1. For http and https schemes, if no port has been specified, we use 80 and 443 for http and https schemes.~
1. In the manifest, bindings *MUST* always have a target port. If there is none, one is dynamically allocated.
1. The exposed port is allocated if there's no container port is allocated.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3260)


playground/orleans/OrleansAppHost/aspire-manifest.json
```diff
@@ -29,31 +29,35 @@
         "Orleans__GrainStorage__Default__ProviderType": "AzureBlobStorage",
         "Orleans__GrainStorage__Default__ServiceKey": "grainstate",
         "ConnectionStrings__grainstate": "{grainstate.connectionString}",
-        "Orleans__ClusterId": "84bf51f8b9c14c58ac7119216fde5c20",
+        "Orleans__ClusterId": "5179116e0d41450e9bcb05f7936f1a66",
```
mitchdenny, line 32: Why did the cluster ID change?

davidfowl: Must be code in orleans

davidfowl: Yea it changes every run

davidfowl:
Replaced by https://github.com/dotnet/aspire/pull/3274

PR #3266, opened 2024-03-28
Show https endpoints more consistently

### LINKED ISSUE #3106: [WebToolsE2E] Endpoints show with + symbols and we need to do additional clicks to open the Endpoints displayed in the Dashboard
**REGRESSION INFO**: Worked-in: Aspire 8.0 P4

**INSTALL STEPS**
2. Install VS 17.10 Preview 3 build main-34721.267 (includes .NET SDK 8.0.200 and .NET workload Aspire 8.0.0-preview.4.24156.9)
3. Install Aspire 8.0.0-preview.5.24172.3, apply feeds

**REPRO STEPS**
1. New Aspire Starter Application > check 'use redis' > AspireStarterApp2
2. Ctrl + F5

**ACTUAL**
With both https and http endpoints available now, the Endpoints column does not display the full URL links. Instead it shows '+1' and '+2' that you need to click and then view the remaining endpoints.

With webfrontend in the screenshot below, you can see that the http endpoint is the one that's displayed but users are more likely to want to click the https endpoint, which is harder to get to.

image

Even with only http profile, the Endpoints column does not display the full URL even though space is available on the screen.

image

**EXPECTED**
This is the experience we had in Preview 4, which I feel was better

**http profile only**
image

**http and https profiles**
image

tlmii:
A couple of quick changes that hopefully mostly fix #3106 (for now at least).

- Give the Endpoints column a little bit more room and take a little bit of room away from the Name column (now that we show less in that column)
- If there's only one entry in the Endpoints column, display it directly and do not use the FluentOverflow component. It'll be displayed with normal ellipsis truncation.
- Sort the endpoints so that https comes first.

This does _not_ address the part of #3106 where the first endpoint (of more than one) would fit in the space, but overflow doesn't show it (because it doesn't fit next to the +n button). That would take a change in the underlying overflow component.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3266)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor
@@ -53,7 +53,7 @@
                         ViewKey="ResourcesList">
         <Summary>
             @{
-                var gridTemplateColumns = HasResourcesWithCommands ? "1fr 2fr 1.25fr 1.5fr 2.5fr 2fr 1fr 1fr 1fr" : "1fr 2fr 1.25fr 1.5fr 2.5fr 2fr 1fr 1fr";
+                var gridTemplateColumns = HasResourcesWithCommands ? "1fr 1.5fr 1.25fr 1.5fr 2.5fr 2.5fr 1fr 1fr 1fr" : "1fr 1.5fr 1.25fr 1.5fr 2.5fr 2.5fr 1fr 1fr";
             }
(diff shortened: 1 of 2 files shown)


tlmii:
An example of this in action, with resources that have no endpoints, non-url endpoints, single endpoints and multiple endpoints:

 
EndpointsTruncation

tlmii:
And a quick screenshot of the overall resources page for TestShop so you can see relative column sizes (I have not resized any columns at this point, just shunk my window size):

image

adamint (approved):

kvenkatrajan (approved):

kvenkatrajan:
@tlmii - Thanks for getting this fix in so quickly!

PR #3269, opened 2024-03-29
Re-enables IDE protocol v1 support.

karolz-ms:
Corresponding change in DCP will make it tolerate any combination of IDE tooling and app host libraries. That is, DCP will be able to use both old and new-style annotations and will probe the IDE for protocol supported, translating the requests as necessary.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3269)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -989,19 +989,30 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
             {
                 exeSpec.ExecutionType = ExecutionType.IDE;
 
+                if (_dcpInfo?.Version?.CompareTo(DcpVersion.MinimumVersionIdeProtocolV1) >= 0)
+                {
+                    projectLaunchConfiguration.DisableLaunchProfile = project.TryGetLastAnnotation<ExcludeLaunchProfileAnnotation>(out _);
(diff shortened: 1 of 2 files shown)


karolz-ms:
@mitchdenny meant to ask you for review as well but for some reason GH is not letting me... 😕

mitchdenny (approved):

davidfowl:
@karolz-ms did you try this with public VS? If not I can before merging.

karolz-ms:
@davidfowl I have not yet. Would very much appreciate a check from you

davidfowl:
We need this change in preview 6 right?

davidfowl (dismissed):
Still exploding 

image

davidfowl:
Is the new DCP in already? Does it handle old pre v1 protocol requests?

karolz-ms:
This change requires DCP version 0.1.62, which has been stuck in the pipeline since two days ago due to some AzDO issues... 😞

karolz-ms:
I verified that with the new DCP and VS 2022 17.10.0 Preview 2.0 (NOT IntPreview) the TestShop app still works with this change. So we just need to wait for the DCP insertion and then we can merge.

karolz-ms:
Verified again this morning with DCP 0.1.62 that was integrated into Aspire main over the weekend, works with publicly available VS 2022 Preview

PR #3271, opened 2024-03-29
Upgrade to new version of Azure.Provisioning

JoshLove-msft:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3271)


src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -52,12 +52,11 @@ public static class AzurePostgresExtensions
             var administratorLogin = new Parameter("administratorLogin");
             var administratorLoginPassword = new Parameter("administratorLoginPassword", isSecure: true);
 
-            var postgres = new PostgreSqlFlexibleServer(construct, administratorLogin, administratorLoginPassword, name: construct.Resource.Name);
+            var postgres = new PostgreSqlFlexibleServer(construct, administratorLogin, administratorLoginPassword, name: construct.Resource.Name, storageSizeInGB: 32);
             postgres.AssignProperty(x => x.Sku.Name, "'Standard_B1ms'");
             postgres.AssignProperty(x => x.Sku.Tier, "'Burstable'");
             postgres.AssignProperty(x => x.Version, "'16'");
             postgres.AssignProperty(x => x.HighAvailability.Mode, "'Disabled'");
-            postgres.AssignProperty(x => x.Storage.StorageSizeInGB, "32");
```
mitchdenny, line 60: Does this mean this value isn't settable via AssignProperty?

mitchdenny (approved):

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
JoshLove-msft: Actually, I was mistaken - overriding the property should work now because the underlying nested instances won't be null.

davidfowl:
P6?

eerhardt:
I think this is needed for preview5. The versions of the Azure.Provisioning we are using today isn't on nuget.org.

cc @joperezr

PR #3274, opened 2024-03-29
Endpoint changes

### LINKED ISSUE #2099: Add ability to express exposed services in app host
See https://github.com/Azure/azure-dev/issues/3284

cc @mitchdenny

davidfowl:
- In order for tools to better describe multiple endpoints int the manifest, we must include port information. At runtime most of the port information is dynamically generated and thus not described (for anything but containers), that doesn't work well when trying to deploy to various environments. We allocate ports in situations where there are none to match the runtime behavior. 
- Change also renames ContainerPort to TargetPort everywhere.
- Added `WithExternalHttpEndpoints` to mark all http and https endpoints as external.

The exposed port (port in the manifest) is the port used to talk to this binding.
The target port (targetPort in the manifest) is the port the resource binds to.

**From Copilot:**

These rules determine the values of the targetPort and exposedPort variables based on the properties of an endpoint object:
1.	If the targetPort property is already specified, use its value as the targetPort.
2.	If the endpoint belongs to a ProjectResource and the targetPort is not specified, set the targetPort to null. This is because ProjectResource objects get their default port from the deployment tool.
3.	If none of the above conditions are met, allocate a dynamic port for the targetPort.

Next, the exposedPort is determined based on the following conditions:
1.	If the exposedPort and targetPort are the same, set the exposedPort to null. This means that the exposed port does not need to be explicitly mentioned because it is the same as the target port.
2.	If the port property is specified, use its value as the exposedPort.
3.	If the targetPort is specified and the port is not, set the exposedPort to null. In this case, the exposedPort will default to the targetPort.
4.	If the endpoint has a URI scheme of "http" or "https" and both the port and targetPort are null, set the exposedPort to null. This allows the tool to infer the default HTTP and HTTPS ports.
5.	For other URI schemes, allocate a a dynamic port for the exposedPort.

Fixes https://github.com/dotnet/aspire/issues/2099


src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -277,9 +278,49 @@ public void WriteBindings(IResource resource, bool emitContainerPort = false)
                 Writer.WriteString("protocol", endpoint.Protocol.ToString().ToLowerInvariant());
                 Writer.WriteString("transport", endpoint.Transport);
 
-                if (emitContainerPort && endpoint.ContainerPort is { } containerPort)
+                int? targetPort = (resource, endpoint.UriScheme, endpoint.TargetPort) switch
```
mitchdenny, line 281: Nice!

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -264,11 +264,12 @@ public void WriteConnectionString(IResource resource)
     /// collection.
     /// </summary>
     /// <param name="resource">The <see cref="IResource"/> that contains <see cref="EndpointAnnotation"/> annotations.</param>
-    /// <param name="emitContainerPort">Flag to determine whether container port is emitted.</param>
-    public void WriteBindings(IResource resource, bool emitContainerPort = false)
+    public void WriteBindings(IResource resource)
     {
         if (resource.TryGetEndpoints(out var endpoints))
         {
+            var allocator = new PortAllocator();
```
mitchdenny, line 271: I'm wondering whether this should be more global in scope. If the target compute platform is something like Docker compose you could end up with port conflicts for the external ports right?

davidfowl: This is a *GREAT* question.  I think this does have to be global. We probably shouldn't assume anything.

davidfowl: Done

mitchdenny (approved):

PR #3277, opened 2024-03-29
Put all database connection strings in keyvault too

davidfowl:
- Change how we intercept postgres with IPostgresResource
- Add db connection strings into keyvault directly and reference them from there.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3277)


mitchdenny:
This is halfway to what we were talking about yesterday with introducing an interface for a resource and PublishAs/As returning the AzurePostgresResource.

davidfowl:
Thoughts on the general approach?

cc @ellismg @vhvb1989 This is the pattern we discussed where the entire connection string is stored in kv so we don't need to ever get it client side.

PR #3219, opened 2024-03-27
Update docker message in templates

DamianEdwards:
Relates to #3190
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3219)

PR #3286, opened 2024-03-29
Add additional project template string

tlmii:
This file gets created every time I build main right now. I _think_ it was missed in #3219 but I'm not entirely sure how that all works. 

@DamianEdwards @joperezr FYI, in case you know what's going on here.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3286)


DamianEdwards:
Yep, I forgot to build the template package. We should have a test so that CI fails if any files are add/changed when templates are built.

DamianEdwards (approved):

tlmii:
@joperezr Guessing I can backport this to Preview 6 with no issue. If we have to do another P5 during QB mode this is low-risk to sneak in but I'm guessing probably not worth it on its own?

PR #3293, opened 2024-03-30
Event Hubs: Update EventProcessorClient component to pull BlobClient from DI

### LINKED ISSUE #3254: EventProcessorClient should use DI for blob client (Event Hubs)
As per https://github.com/dotnet/aspire/pull/2870 comments by @eerhardt 

> We should just be able to call a different overload on the AzureClientFactoryBuilder, which gives us the IServiceProvider. See https://github.com/Azure/azure-sdk-for-net/blob/87d426597b261b88f8a6c6c21fab6031f857fb70/sdk/extensions/Microsoft.Extensions.Azure/src/AzureClientFactoryBuilder.cs#L164-L172. (Note it would take some refactoring of the base AzureComponent<TSettings, TClient, TClientOptions> class, as right now the abstract AddClient method takes a TBuilder, when it could just be passed an AzureClientFactoryBuilder.
> 
> I think this needs some configuration-by-convention love in the sense that we should probably only permit a keyed blob client so it can be configured in settings.
> 
> I see it a different way. It should allow for a keyed blob client, but shouldn't require it to be keyed. If no "blob service key" was configured, it should just get the BlobServiceClient from DI. If a "blob service key" was specified, then use it to get the BlobServiceClient from DI.
> 
> That way the simple case "just works". You just need to:
> 
> var builder = WebApplication.CreateBuilder(args);
> 
> builder.AddAzureBlobClient("blobConnectionName");
> builder.AddAzureEventProcessorClient("eventhubns");
> and the EventProcessorClient grabs the BlobClient. There's no need to force someone to use keys when there is only one.

oising:
- Update AzureComponent to pass concrete AzureClientFactoryBuilder and remove TBuilder generic
- Update associated components (RegisterClientFactory is replaced with concrete AddClient call)
- Update launchSettings.json in EH sample AppHost
- Update EventProcessorClient component to pull BlobServiceClient from DI

The BlobServiceClient is first looked for as a keyed service using `settings.BlobConnectionName` -- the `serviceKey` is always the same as the `connectionName` for keyed services. If this is not found, a default service is looked for. If this is not found, an `InvalidOperationException` is thrown with a helpful message. This is a breaking change from Preview 5.

Closes https://github.com/dotnet/aspire/issues/3254
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3293)


oising:
fyi @eerhardt

src/Components/Aspire.Azure.Data.Tables/AspireTablesExtensions.cs
```diff
@@ -114,7 +116,7 @@ private sealed class TableServiceComponent : AzureComponent<AzureDataTablesSetti
                 return !string.IsNullOrEmpty(connectionString) ? new TableServiceClient(connectionString, options) :
                     cred is not null ? new TableServiceClient(settings.ServiceUri, cred, options) :
                     new TableServiceClient(settings.ServiceUri, options);
-            }, requiresCredential: false);
```
sebastienros, line 117: What is the impact of changing this? I see that it's always `true` in `AddClient` methods

eerhardt: We shouldn't be changing this. We need to keep the current behavior.

eerhardt: I fixed most changes back to the way they were. I missed the EventHubs ones, but I need to step away. Can you take care of that? I can take care of it when I get back too.

src/Components/Aspire.Azure.Messaging.EventHubs/EventProcessorClientComponent.cs
```diff
@@ -3,58 +3,62 @@
...
+                    ? new EventProcessorClient(containerClient,
+                        settings.ConsumerGroup ?? EventHubConsumerClient.DefaultConsumerGroupName,
+                        settings.ConnectionString)
+                    : new EventProcessorClient(containerClient,
                         settings.ConsumerGroup ?? EventHubConsumerClient.DefaultConsumerGroupName, settings.Namespace,
                         settings.EventHubName, cred, options);
 
                 return processor;
-
-            }, requiresCredential: false);
```
sebastienros, line 53: There is no available method that provides options, cred, provider and requiresCredential. The ultimate private call has these but there is no public combination that allows to access it. Ideally on AddClient since this is the one we wanted. I assume the expectation was that if you provide a factory with `TokenCredential` you'd want `requiresCredential: true` but here we don't always.

eerhardt: I was able to successfully use this using a connection string. So I'm not sure we need to specify `requiresCredential: false`. Since this is a new component, we shouldn't be breaking anyone. But let's keep the other components doing what they are doing now. This one can use AddClient.

@tg-msft - do you know what this `requiresCredential` bool does/means exactly? Why can't I specify `requiresCredential: false` when I use the overloads with IServiceProvider?

src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
```diff
@@ -120,17 +120,20 @@ public sealed class AzureMessagingEventHubsConsumerSettings : AzureMessagingEven
...
+    /// Gets or sets the IServiceProvider service key used to obtain an Azure BlobServiceClient.
     /// </summary>
-    /// <remarks>Applies only to <see cref="EventProcessorClient"/></remarks>
-    public string? BlobClientConnectionName { get; set; }
+    /// <remarks>
+    /// A BlobServiceClient is required when using the Event Processor. If a BlobClientServiceKey is not configured,
+    /// an un-keyed BlobServiceClient will be retrieved from the IServiceProvider. If a BlobServiceClient is not available in
+    /// the IServiceProvider, an exception is thrown.
+    /// </remarks>
+    public string? BlobClientServiceKey{ get; set; }
```
davidfowl, line 130: Formatting

eerhardt (approved):
This looks great. I pushed a few minor changes.

Thanks for the great contributions @oising!

src/Components/Aspire.Azure.Messaging.EventHubs/EventProcessorClientComponent.cs
oising: Yeah, this is the problem I had -- there was no public overload available that allowed specifying it when you want the service provider. I couldn't immediately see what the point of this flag even was, so I figured it was vestigial as everything seemed to work regardless.

tg-msft: [requiresCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.core.extensions.iazureclientfactorybuilderwithcredential.registerclientfactory?view=azure-dotnet#parameters) is a niche feature for services that allow anonymous auth (i.e., reads on a Storage account used for hosting static websites).  I believe it only controls whether or not we throw an exception when resolving a client if there was no credential provided.  https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientRegistration.cs#L46

PR #3294, opened 2024-03-30
Improve service address allocation

### LINKED ISSUE #3265: Endpoint allocation can get stuck some times
This code in `ApplicationExecutor.CreateServicesAsync` should have a timeout to either fail or at least log a message that it is still waiting on address allocation for some services, and list those services.

https://github.com/dotnet/aspire/blob/93a1efbc4d8798903828eb5da7ffb50b873dc513/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L826-L845

This was hit in running `Aspire.EndToEnd.Tests` where 1/20 runs fails where it gets stuck waiting for a service to get allocated.

Screenshot 2024-03-28 at 15 25 23
In the above run it is stuck waiting for `redis`, till the test infrastructure times out after 5 minutes.

[Full log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-2800-merge-6d339efb36ed4a7986/Aspire.EndToEnd.Tests-5/1/console.af9c471f.log?helixlogtype=result). And corresponding dcp logs are [here](https://helix.dot.net/api/jobs/6d339efb-36ed-4a79-862f-93fb9c517d6c/workitems/Aspire.EndToEnd.Tests-5?api-version=2019-06-17) in `logs/dcp-diag-logs`, and the corresponding [build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=622703&view=results).

Note that this is not specific to redis.

cc @karolz-ms @eerhardt

karolz-ms:
Should fix https://github.com/dotnet/aspire/issues/3265
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3294)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -1045,7 +1045,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                 }
                 else
                 {
-#pragma warning disable CS0612 // These annotations are obsolete; remove in Aspire Preview 6
+#pragma warning disable CS0612 // These annotations are obsolete; remove after Aspire GA
                     annotationHolder.Annotate(Executable.CSharpProjectPathAnnotation, projectMetadata.ProjectPath);
 
                     // ExcludeLaunchProfileAnnotation takes precedence over LaunchProfileAnnotation.
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -836,7 +836,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                 BackoffType = DelayBackoffType.Constant,
                 MaxDelay = TimeSpan.FromSeconds(1),
                 UseJitter = true,
-                MaxRetryAttempts = 2,
+                MaxRetryAttempts = 1,
                 ShouldHandle = new PredicateBuilder().Handle<Exception>(),
                 OnRetry = (retry) =>
                 {
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -813,7 +813,9 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
         {
             AspireEventSource.Instance.DcpServicesCreationStart();
 
-            var needAddressAllocated = _appResources.OfType<ServiceAppResource>().Where(sr => !sr.Service.HasCompleteAddress).ToList();
+            var needAddressAllocated = _appResources.OfType<ServiceAppResource>()
+                .Where(sr => !sr.Service.HasCompleteAddress && sr.Service.Spec.AddressAllocationMode != AddressAllocationModes.Proxyless)
+                .ToList();
 
             await CreateResourcesAsync<Service>(cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/Dcp/KubernetesService.cs
```diff
@@ -54,6 +57,36 @@ internal sealed class KubernetesService(ILogger<KubernetesService> logger, IOpti
 
     public TimeSpan MaxRetryDuration { get; set; } = TimeSpan.FromSeconds(20);
 
+    public Task<T> GetAsync<T>(string name, string? namespaceParameter = null, CancellationToken cancellationToken = default)
+        where T : CustomResource
+    {
+        var resourceType = GetResourceFor<T>();
+
+        return ExecuteWithRetry(
```
davidfowl, line 65: @martincostello @mitchdenny We should use polly right?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -823,26 +825,53 @@ private async Task CreateServicesAsync(CancellationToken cancellationToken = def
...
-            {
-                if (evt == WatchEventType.Bookmark) { continue; } // Bookmarks do not contain any data.
+            const int watchAttempts = 2;
 
-                var srvResource = needAddressAllocated.Where(sr => sr.Service.Metadata.Name == updated.Metadata.Name).FirstOrDefault();
-                if (srvResource == null) { continue; } // This service most likely already has full address information, so it is not on needAddressAllocated list.
+            for (int attempt = 0; attempt < watchAttempts; attempt++)
+            {
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                cts.CancelAfter(TimeSpan.FromSeconds(10));
```
davidfowl, line 833: Make it configurable?

src/Aspire.Hosting/Dcp/KubernetesService.cs
martincostello: Looking at the code it looks like there's Polly retries at one level, and then this is retrying manually around that. I would have thought that if you need to retry the retries, that yeah you could use Polly here two with a differently configured pipeline/strategy.

mitchdenny: I'd rather see us use Polly than hand rolling retry logic. As for retries within retries I'd try to avoid that and agree with what Martin says (but he is the Polly expert :P)

karolz-ms: The code inside `KubernetesService` (and its retry logic) predates the use of Polly inside Aspire hosting.

If we want to standardize on Polly, we can do this, but given that this change will probably need to be backported to P5, I suggest we open a separate issue for that and address in P6/GA timeframe.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
sebastienros: Does it make sense to have "2" retries exactly but be able to configure the timeout for each attempt? Should it instead either:
- have a configurable timeout and retry as many times (with a non-configurable delay)
- or configure a number retries (with a non-configurable delay)

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -823,26 +825,53 @@ private async Task CreateServicesAsync(CancellationToken cancellationToken = def
...
                 {
-                    srvResource.Service.ApplyAddressInfoFrom(updated);
-                    needAddressAllocated.Remove(srvResource);
+                    // We do not specify the initial list version, so the watcher will give us all updates to Service objects.
+                    IAsyncEnumerable<(WatchEventType, Service)> serviceChangeEnumerator = kubernetesService.WatchAsync<Service>(cancellationToken: cts.Token);
+                    await foreach (var (evt, updated) in serviceChangeEnumerator)
+                    {
+                        if (evt == WatchEventType.Bookmark) { continue; } // Bookmarks do not contain any data.
+
+                        var srvResource = needAddressAllocated.Where(sr => sr.Service.Metadata.Name == updated.Metadata.Name).FirstOrDefault();
```
sebastienros, line 843:
```suggestion
                        var srvResource = needAddressAllocated.FirstOrDefault(sr => sr.Service.Metadata.Name == updated.Metadata.Name);
```

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: This change is a workaround for a bug that is not super well understood TBH. It is also part of the application startup sequence, so the performance is a big concern. The current behavior, if we hit the issue (missed resource change notification) is to wait indefinitely for a notification that never comes. In other words, a hang. That is not good, and that is also why it does not make sense to have unlimited number of retries. At this time, with what I know, two retries of the watch, followed by a direct query attempt, will strike the best balance between performance, reliability, and giving the developer feedback when something goes wrong. If/when we get more data (we do have automated test that hit this issue), we can re-assess. 

I do agree that making the timeout configurable is a good idea, let me do that.

karolz-ms: Done

karolz-ms:
@sebastienros @martincostello Thanks for the review, I think I addressed all the comments, LMK

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -823,26 +825,53 @@ private async Task CreateServicesAsync(CancellationToken cancellationToken = def
...
+            {
+                var dcpSvc = await kubernetesService.GetAsync<Service>(sar.Service.Metadata.Name, cancellationToken: cancellationToken).ConfigureAwait(false);
+                if (dcpSvc.HasCompleteAddress)
+                {
+                    sar.Service.ApplyAddressInfoFrom(dcpSvc);
+                }
+                else
                 {
-                    return; // We are done
+                    distributedApplicationLogger.LogWarning($"Unable to allocate a network port for service '{sar.Service.Metadata.Name}'; service may be unreachable and its clients may not work properly.");
```
sebastienros, line 871: Don't use interpolated strings

```suggestion
                    distributedApplicationLogger.LogWarning("Unable to allocate a network port for service '{ServiceName}'; service may be unreachable and its clients may not work properly.", sar.Service.Metadata.Name);
```

karolz-ms:
Don't merge yet, I have some offline feedback from @mitchdenny to address here

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -823,26 +825,53 @@ private async Task CreateServicesAsync(CancellationToken cancellationToken = def
...
+                            needAddressAllocated.Remove(srvResource);
+                        }
+
+                        if (needAddressAllocated.Count == 0)
+                        {
+                            return; // We are done
+                        }
+                    }
                 }
+                catch { }
```
mitchdenny, line 858: Can we be more specific with the catch here? Do we know what specific exceptions are likely to occur that we can retry and everything else is unexpected and therefore fatal? We should be looking to at least log these exceptions via `ILogger.LogDebug`.

karolz-ms: done

tests/Aspire.Hosting.Tests/Dcp/MockKubernetesService.cs
```diff
@@ -11,7 +11,20 @@ internal sealed class MockKubernetesService : IKubernetesService
...
+
+    public Task<T> GetAsync<T>(string name, string? namespaceParameter = null, CancellationToken _ = default) where T : CustomResource
+    {
+        var res = CreatedResources.OfType<T>().FirstOrDefault(r =>
+            r.Metadata.Name == name &&
+            string.Equals(r.Metadata.NamespaceProperty ?? string.Empty, namespaceParameter ?? string.Empty)
+        );
+        if (res == null)
+        {
+            throw new ArgumentException($"Resource '{name}' not found");
```
martincostello, line 24: Use the overload the specifies the param name too?

martincostello (approved):

tests/Aspire.Hosting.Tests/Dcp/MockKubernetesService.cs
karolz-ms: Actually, what is missing here is the reference to namespace, if provided. So I do not think I will use the overload with the param name for creating the Exception (there might be more than one parameter involved), but I need to include the namespace--thanks!

karolz-ms: Done. Just to clarify: the `somenamespace/someobject` is a Kubernetes convention for presenting object names; and if the object is not namespaced, then that becomes `/someobject`

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -999,7 +1045,7 @@ private void PrepareProjectExecutables()
                 }
                 else
                 {
-#pragma warning disable CS0612 // These annotations are obsolete; remove in Aspire Preview 6
+#pragma warning disable CS0612 // These annotations are obsolete; remove after Aspire GA
```
karolz-ms, line 1048: Not really related to the fix, just reusing this PR for 1-line comment update

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -823,26 +826,69 @@ private async Task CreateServicesAsync(CancellationToken cancellationToken = def
...
+            var withTimeout = new TimeoutStrategyOptions()
             {
-                if (evt == WatchEventType.Bookmark) { continue; } // Bookmarks do not contain any data.
+                Timeout = _options.Value.ServiceStartupWatchTimeout
+            };
+
+            var tryTwice = new RetryStrategyOptions()
+            {
+                BackoffType = DelayBackoffType.Constant,
+                MaxDelay = TimeSpan.FromSeconds(1),
```
mitchdenny, line 837: Is this enough time. Based on the settings here it will trice once, then wait a second and try again, then again. So if it isn't responsive after what 3 seconds its going to give up?

This means that the timeout value specified in the options won't be that effective since these hard coded settings means that a value have 2-3 seconds won't be effective.

mitchdenny (approved):

PR #3305, opened 2024-03-31
Added ability to resolve target port

davidfowl:
- Change how we process and expose target ports. This change makes it possible to avoid using any internals of DCP to access and resolve target ports from any resource.
- More launch profile processing to the project resource defaults.
- Remove logic in the manifest writing that wrote a special env variable. This should just work now  like everything else.
- Exposed TargetPort and Exists on EndpointReference.
- Use this new target property with dapr.
- Added a ReferenceExpressionBuilder which makes it possible to dynamically build a ReferenceExpression.
- Endpoints and env are evaluated earlier now so change remove tests that add default launch profile on real projects without allocating those endpoints.

cc @DamianEdwards this should fix dapr and the tests.

Goal is to have this go into p6 (API fit and finish).

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs
+++ b/src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs
@@ -18,7 +18,8 @@ public class AllocatedEndpoint
     /// <param name="address">The IP address of the endpoint.</param>
     /// <param name="containerHostAddress">The address of the container host.</param>
     /// <param name="port">The port number of the endpoint.</param>
-    public AllocatedEndpoint(EndpointAnnotation endpoint, string address, int port, string? containerHostAddress = null)
+    /// <param name="dcpServiceName">The name of the DCP service.</param>
+    public AllocatedEndpoint(EndpointAnnotation endpoint, string address, int port, string? containerHostAddress = null, string? dcpServiceName = null)
     {
         ArgumentNullException.ThrowIfNull(endpoint);
         ArgumentOutOfRangeException.ThrowIfLessThan(port, 1, nameof(port));
@@ -28,6 +29,7 @@ public class AllocatedEndpoint
         Address = address;
         ContainerHostAddress = containerHostAddress;
         Port = port;
+        DcpServiceName = dcpServiceName;
     }
 
     /// <summary>
@@ -64,6 +66,11 @@ public class AllocatedEndpoint
     /// URI in string representation.
     /// </summary>
     public string UriString => $"{UriScheme}://{EndPointString}";
+
+    /// <summary>
+    /// The associated service name created in DCP for this endpoint.
+    /// </summary>
+    public string? DcpServiceName { get; }
 
     /// <summary>
     /// Returns a string representation of the allocated endpoint URI.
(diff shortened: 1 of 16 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1034,18 +1035,14 @@ private void PrepareProjectExecutables()
...
                 exeSpec.Args.Add("--no-launch-profile");
 
-                var launchProfileName = project.SelectLaunchProfileName();
-                if (!string.IsNullOrEmpty(launchProfileName))
+                var launchProfile = project.GetEffectiveLaunchProfile();
+                if (launchProfile is not null && !string.IsNullOrWhiteSpace(launchProfile.CommandLineArgs))
                 {
-                    var launchProfile = project.GetEffectiveLaunchProfile();
-                    if (launchProfile is not null && !string.IsNullOrWhiteSpace(launchProfile.CommandLineArgs))
+                    var cmdArgs = launchProfile.CommandLineArgs.Split((string?)null, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
```
davidfowl, line 1041: This should move to the project defaults as well https://github.com/dotnet/aspire/issues/3306

karolz-ms (approved):
This is pretty neat; I really like how all the DCP details were encapsulated in this change.

Approving assuming that the launch profile command line argument handling will also go to ProjectResourceBuilderExtensions.

DamianEdwards:
you going to backport to p6 @davidfowl?

src/Aspire.Hosting/ApplicationModel/AllocatedEndpoint.cs
```diff
@@ -65,6 +67,11 @@ public AllocatedEndpoint(EndpointAnnotation endpoint, string address, int port,
     /// </summary>
     public string UriString => $"{UriScheme}://{EndPointString}";
 
+    /// <summary>
+    /// The associated service name created in DCP for this endpoint.
+    /// </summary>
+    public string? DcpServiceName { get; }
```
eerhardt, line 73: I'm not sure this is a great public name in our application model.

Users don't really know (and shouldn't know IMO) about DCP at all. Is there a better name we can use here?

DamianEdwards: `OrchestratorServiceName`?

eerhardt: Can we make it internal for now? We can always make it public later.

davidfowl: That makes it impossible to write a test. This is difficult because we need something that can appear in string interpolated expression. It doesn’t need to be the dcp expression but it has to be something that we can detect and replace with the dcp expression, at runtime .

There’s another alternative that comes to mind:

Make it internal:

Don’t evaluate this object inside of the IValueProvider implementation, but have that object flow all the way out to evaluation phase (GetValue) in ApplicationExecutor.

I think that means we’d need to manually process ReferenceExpressions in GetValue, handling target port specially.

PR #3330, opened 2024-04-02
Make resource group name mandatory config value (don't auto generate).

### LINKED ISSUE #3259: Generate and persist random resource group names, don't use the machine name.
Right now, we're using the machine name to get a stable resource group name if one wasn't specified. We should instead generate a random name and store it in user secrets.

mitchdenny:
Fixes #3259 

We are deciding not to auto generate the name at all. The reason is that if you auto generate the name then you end up creating a bunch of resources in a resource group that the developer doesn't know about. If customers are provisioning in a busy subscription then they could easily loose track of these resources.

By forcing developers to provide a resource group name themselves they use "their naming system" so that they have an increased chance on knowing whether they can delete those dev time resources or not.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3330)

### DIFF
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
@@ -15,7 +15,6 @@ using Azure.ResourceManager.Resources;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Configuration.UserSecrets;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
@@ -26,7 +25,6 @@ internal sealed class AzureProvisioner(
     IOptions<AzureProvisionerOptions> options,
     DistributedApplicationExecutionContext executionContext,
     IConfiguration configuration,
-    IHostEnvironment environment,
     ILogger<AzureProvisioner> logger,
     IServiceProvider serviceProvider,
     ResourceNotificationService notificationService,
@@ -362,14 +360,10 @@ internal sealed class AzureProvisioner(
             throw new MissingConfigurationException("An azure location/region is required. Set the Azure:Location configuration value.");
         }
 
-        var unique = $"{Environment.MachineName.ToLowerInvariant()}-{environment.ApplicationName.ToLowerInvariant()}";
-
-        // Name of the resource group to create based on the machine name and application name
-        var (resourceGroupName, createIfAbsent) = _options.ResourceGroup switch
+        if (string.IsNullOrEmpty(_options.ResourceGroup))
         {
-            null or { Length: 0 } => ($"rg-aspire-{unique}", true),
-            string rg => (rg, _options.AllowResourceGroupCreation ?? false)
-        };
+            throw new MissingConfigurationException("A resource group name is required. Set the Azure:ResourceGroup configuration value.");
+        }
 
         var resourceGroups = subscriptionResource.GetResourceGroups();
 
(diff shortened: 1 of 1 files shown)


eerhardt:
+1 to Damian's comment about this being a friction point.

I rather like that my machine name (which is based on my name) shows up in the Resource Group name. That way in the Azure Subscription (which is shared by hundreds of devs), I can find my resource groups easily by filtering based on my name.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -378,25 +372,25 @@ await foreach (var tenant in armClient.GetTenants().GetAllAsync(cancellationToke
...
-            var response = await resourceGroups.GetAsync(resourceGroupName, cancellationToken).ConfigureAwait(false);
+            var response = await resourceGroups.GetAsync(_options.ResourceGroup, cancellationToken).ConfigureAwait(false);
             resourceGroup = response.Value;
 
             logger.LogInformation("Using existing resource group {rgName}.", resourceGroup.Data.Name);
         }
         catch (Exception)
         {
-            if (!createIfAbsent)
+            if (!_options.AllowResourceGroupCreation ?? false)
```
eerhardt, line 382: Does this mean that I need to also set `AllowResourceGroupCreation` in my config in order to get it created? This wasn't necessary before.

mitchdenny:
> I'm personally not a fan of required configuration values like this as they're an obvious friction point in the path to first deployment. Have we thought about an API-based approach to setting the resource group name as well? I believe we removed `AddAzureProvisioning()` but if it were still there then that'd be a natural place to allow setting the resource group name.

I don't see how this adds any more friction beyond having to specify the subscription ID?

mitchdenny:
> +1 to Damian's comment about this being a friction point.
> 
> I rather like that my machine name (which is based on my name) shows up in the Resource Group name. That way in the Azure Subscription (which is shared by hundreds of devs), I can find my resource groups easily by filtering based on my name.

The concern raised in the linked issue was that this is leaking potentially confidential information into the names of the resource groups that are being created. If that isn't a concern then we probably don't need to do any of this.

In practice people are usually going to put their name/username in the RG name for dev purposes so they can find it later ... but does that mean we should default to doing that? :)

eerhardt:
> I don't see how this adds any more friction beyond having to specify the subscription ID?

Because it's "one more thing".

You F5 the first time. You get an error:

"Please set the SubscriptionID".

So you do that, and F5 again. Now you get a different error:

"Please set a Location".

So you do that, and F5 again.

Today at this point it works. With this change you get into another error:

"Please set a Resource Group Name".


Maybe we should be prompting for these things if they don't exist in the config, and writing them back to user secrets once people enter them in. I think I would be more amenable to a solution like that. Similar to what `azd up` does - asks me for all 3 one after the other. I don't need to try again 3 times.

mitchdenny:
> Maybe we should be prompting for these things if they don't exist in the config, and writing them back to user secrets once people enter them in. I think I would be more amenable to a solution like that. Similar to what `azd up` does - asks me for all 3 one after the other. I don't need to try again 3 times.

As it happens we are working on a solution for this :) I don't think the timing will work for GA, but VS has a JSONRPC API we can call to get it to prompt for these values all at once. I'm yet to experiment with it but the idea was to use this mechanism to prompt not only for subscription/location/rg information, but also for parameter values if your model requires it.

eerhardt:
> but VS has a JSONRPC API

What about "not VS""

sebastienros:
My attempt at solving it by implement the design from the issue. It store the generated RG in secret-store like the deployments are. https://github.com/dotnet/aspire/pull/3375

PR #3333, opened 2024-04-02
Add missing QueuePolicy in AWS Playground AppHost

djonser:
The sample CloudFormation template configures a subscription for the SQS Queue to receive messages from the SNS Topic, but no QueuePolicy is configured which results in: `User: arn:aws:iam::111111111111:user/DUB.SQS is not authorized to perform: sqs:sendmessage on resource: arn:aws:sqs:eu-west-1:222222222222:AspireSampleDevResources-ChatMessagesQueue-abcdef123456 because no resource-based policy allows the sqs:sendmessage action`.

This pull request adds a QueuePolicy allowing SNS to deliver messages to the ChatMessagesQueue.


djonser:
@dotnet-policy-service agree

RussKie (approved):

Issue #2730, opened 2024-03-08
Redeployment shouldn't remove manually assigned identities


WhitWaldo:
I've got an existing Key Vault and managed user-assigned identity with the appropriate roles for accessing it. After deploying the container app via `azd up`, I assigned the UAI to it and everything worked as expected.

An hour later after I'd made a few more changes and re-deployed via `azd up`, I realized the app wasn't working. I dug into the logs and realized that as part of the redeployment, even though the other resources remained unchanged, the container app no longer listed my already-existing UAI (in addition to the one deployed by Aspire). 

I sort of expected that re-deployments of container apps would kick off new revision deployments instead of deleting and deploying an entirely new container app, meaning that any changes I'd make to the app would persist, so please consider this a feature request - if I add my own identities to a deployed resource, updates to the deployment should start from whatever is already there and be additive for whatever it's trying to do rather than replace with a clean resource.

Thanks!

davidfowl:
This is because you now have 2 sources of truth. Azd has no idea about the changes you manually made to the environment and using azd up (which does azd provision) will not keep those changes. There's currently no support for [customizing the compute](https://github.com/Azure/azure-dev/issues/3292) outside of generating the bicep using [azd infra synth](https://learn.microsoft.com/en-us/dotnet/aspire/deployment/azure/aca-deployment-azd-in-depth?tabs=windows#generate-bicep-from-net-aspire-app-model), this is still being worked on.

Changes to the environment outside of the app model or IaC will be overwritten when re-provisioning. azd is optimized for being able to repeatably create environments with infrastructure.

Running azd deploy will only deploy your compute and leave the infrastructure alone.

WhitWaldo:
Typically, if you download an ARM template and re-apply it and nothing has changed, nothing happens. If you download the template and make a change to it, it only applies whatever has changed (except some of the older resources - looking at you vnets). And of course, if you deploy a brand new template, you get brand new resources. 

But I guess I would expect some sort of guardrail up that prevents developers from inadvertently deploying a destructive template. The deployments appear to be idempotent - I might suggest adding a validation step: if there already exists a resource group matching the target deployment and it's filled with the same resource names as have been generated in the newly-prepared local template, throw on the brakes and have the developer confirm that they want to proceed with the operation (and accept a -F to force it through without the check).

[...]

PR #2429, opened 2024-02-26
Azure Provisioning Integration

mitchdenny:
Contributes to #2423 

This PR introduces a new extension method called `AddAzureConstruct(...)` which allows developers to make use of the Azure CDK (`Azure.Provisioning`) to define Azure resources. This initial version provides basic integration for accepting parameters from the Aspire app model and using outputs.

The usage is as follows:

```csharp
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.

using Azure.Provisioning.Storage;
using Azure.ResourceManager.Storage.Models;

var builder = DistributedApplication.CreateBuilder(args);
builder.AddAzureProvisioning();

var sku = builder.AddParameter("storagesku");

var construct1 = builder.AddAzureConstruct("construct1", (construct) =>
{
    var account = construct.AddStorageAccount(
        name: "bob",
        kind: StorageKind.BlobStorage,
        sku: StorageSkuName.StandardLrs
        );
    account.AssignParameter(a => a.Sku.Name, construct.AddParameter(sku));

    account.AddOutput(data => data.PrimaryEndpoints.TableUri, "tableUri", isSecure: true);
});

builder.AddProject<Projects.CdkSample_ApiService>("api")
       .WithEnvironment("TABLE_URI", construct1.GetOutput("tableUri"));

// This project is only added in playground projects to support development/debugging
// of the dashboard. It is not required in end developer code. Comment out this code
// to test end developer dashboard launch experience. Refer to Directory.Build.props
// for the path to the dashboard binary (defaults to the Aspire.Dashboard bin output
// in the artifacts dir).
builder.AddProject<Projects.Aspire_Dashboard>(KnownResourceNames.AspireDashboard);

builder.Build().Run();
```

This would result in the following manifest:

```json
{
  "resources": {
    "storagesku": {
      "type": "parameter.v0",
      "value": "{storagesku.inputs.value}",
      "inputs": {
        "value": {
          "type": "string"
        }
      }
    },
    "construct1": {
      "type": "azure.bicep.v0",
      "path": "construct1.module.bicep",
      "params": {
        "storagesku": "{storagesku.value}"
      }
    },
    "api": {
      "type": "project.v0",
      "path": "../CdkSample.ApiService/CdkSample.ApiService.csproj",
      "env": {
        "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES": "true",
[...]

Issue #1864, opened 2024-01-26
ManagedIdentity and Workload Identity support 


dersia:
I really like what Aspire is doing and how it simplifies the development and deployment cycle. as I was looking through the issue for preview 3 I came across some issue that have considerations around connectionstrings.

in general I don't think that it is a good approach to have "full" connectionstrings in files like the manifest, though I understand that there might be circumstances where this might be preferable, I think we should have a better way of handling secrets espacially during deployment.

I am a big fan of Managed- and Workflow Identities and think it makes sense to use them when ever possible. there are some services that might not support ManagedIdentities and for this I would prefer to have them in a Keyvault and them have the connection to that KV be secured using a ManagedIdentity.

I think it would be great for Aspire to set in the manifest to use managed Identities where possible or where it is not possible a additional KV resource is added to the manifest and that in the manifest it ptells the tooling to put the credentials/connectionstrings into this KV and then use the secret from KV when added as a reference in another service.

the tooling (azd up, aspir8) could than create  the managed/workflow Identities, add the right role permissions and configure ACA or Aks to use that identity to connect to the services. this would make much safer production deployments and increase overall security.

I hope it is understandable what I am suggesting, sometimes things sound well formulated in my head, but they don't come out the way I'd like 😅
so please let me know if something is unclear.

davidfowl:
We have plans on how to model secrets that are going to be fleshed out in the next release. Once we're done with the refactoring, they'll never show up in the manifest verbatim (unless you do it manually as part of WithEnvironment).

dersia:
It's great that this is happening as of today already, but when I understand that correctly this is not part of the `manifest.json`, right? So probably this won't happen for Aspir8?

If it would be part of the manifest, we could just implement a "lite" version of this:

`AddManagedIdentityRole(string roleId)`. That would make it easier to add permissions the ManagedIdentity that is created by azd up to existing resources. Most of the time, when you start a new project you often want to connect to existing Azure Resources. We have a lot shared Cognitive Search Resources for example, or when we connect to Azure OpenAI.

davidfowl:
> It's great that this is happening as of today already, but when I understand that correctly this is not part of the manifest.json, right? So probably this won't happen for Aspir8?

Aspir8 doesn't support azure at all. It's a pure kubernetes solution that has no dependency on azure resources.

> If it would be part of the manifest, we could just implement a "lite" version of this:

[...]

PR #3339, opened 2024-04-02
Custom Manifest Output & Custom Identities for Projects

rudiv:
I'm not sure if this is helpful to the wider project, but the ability to support different Managed Identities for a deployed Project is super useful.

Some semi-related issues: https://github.com/dotnet/aspire/issues/2730 https://github.com/dotnet/aspire/issues/1864 

### What is this

For example, consider 4 apps (A, B, C, D). We may want to configure that A and B can access C, but only B can access D. Within ACA by default it's a free for all so we have to rely on our own auth implementation.

The best way to do this is with specific managed identities for each application and use MSAL for request authorisation, yet right now there is no System Assigned Identity and only the User Assigned Identity that's used for the Container Registry.

This change introduces 2 core concepts:
- `Aspire.Hosting`: Custom Manifest Output (via Annotation)
- `Aspire.Hosting.Azure`: User Assigned Identity support against the project resource

The former allows anyone to define a custom output in to the manifest that may be useful for other tools in the future, but is basically groundwork to support this Azure specific extension.

The latter can be used either with hardcoded paths or through a Bicep/AzureConstruct (https://github.com/dotnet/aspire/pull/2429)* Resource.

### Usage

```csharp
var projectResource = builder.AddProject<Projects.ServiceA>("servicea")
    .WithUserAssignedIdentity("IDEN1", "00000000-0000-0000-0000-000000000000", "/subscriptions/<subscription_id>/resourcegroups/my-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/my-user")
    .WithUserAssignedIdentity("IDEN2", bicepResource.GetOutput("clientId"), bicepResource.GetOutput("resourceId"));
```

This outputs the following to the manifest:

```json
"userAssignedIdentities": [
 {
   "clientId": "00000000-0000-0000-0000-000000000000",
   "resourceId": "/subscriptions/\u003Csubscription_id\u003E/resourcegroups/my-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/my-user",
   "env": "IDEN1"
 }
 {
   "clientId": "{identities.outputs.clientId}",
   "resourceId": "{identities.outputs.resourceId}",
   "env": "IDEN2"
 }
]
```

Which can then be consumed by `azd` (See https://github.com/Azure/azure-dev/pull/3634) that add in the appropriate Client IDs for the app to use, in this case `IDEN1_CLIENT_ID` and `IDEN2_CLIENT_ID`.

### Other approaches considered
Originally, I'd planned to just output the Resource ID from here that would be consumed by `azd` to wire up the `userAssignedIdentity` on the container as that's the only thing really required that can't be done outside of Bicep. The `WithUserAssignedIdentity` would then wire up both the output of this Resource ID and the Environment Variable separately.

Whilst this reduces the amount of work that `azd` would have to do, it feels a bit less "general purpose".

What I mean by that is two fold, in our scenario we'd like both the Client ID for use in external apps (done as part of this) and the Principal ID (can be easily added to this - right now we just look up using the resource ID) for assignment of this identity to an app registration to use in the Graph API.

Basically, having _more_ information in the manifest feels preferable to less, even if it means slightly more work for `azd` to perform.

------

As I said at the start, I'm not sure how this really fits in to Aspire as a whole as it's definitively deployment related, but it's helped me to wire things up much easier in a "proper" way as the alternative is modifying the generated Bicep and .yaml files for each container, whereas with this we only need to provide the Bicep for the Identity creation (and in the future, hopefully use the new Construct methods).

This as usual with my stuff relates to https://github.com/Azure/azure-dev/discussions/3184, and whilst I'm aware that changes are incoming I didn't see anything about managed identity per-app support in any other discussions.

`*` There probably needs `Aspire.Hosting.Azure.ManagedServiceIdentities` or similar which I'd be happy to contribute as part of this PR if you confirm this is a right path to take. It's obviously a relatively simple resource:

```csharp
var bicepResource = builder.AddAzureConstruct("identity", construct =>
{
    var uai = new UserAssignedIdentity(construct, name: "testIdentity");
    uai.AddOutput("resourceId", sa => sa.Id);
    uai.AddOutput("clientId", sa => sa.ClientId);
});
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3339)

### DIFF
--- a/src/Aspire.Hosting.Azure/ApplicationModel/UserAssignedIdentityAnnotation.cs
+++ b/src/Aspire.Hosting.Azure/ApplicationModel/UserAssignedIdentityAnnotation.cs
@@ -1,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting.Azure.ApplicationModel;
+
+/// <summary>
+/// Represents a user assigned identity that should be assigned to a project.
+/// </summary>
+public class UserAssignedIdentityAnnotation : CustomManifestOutputAnnotation
+{
+    /// <summary>
+    /// Initializes a new instance of <see cref="UserAssignedIdentityAnnotation"/>.
+    /// </summary>
+    /// <param name="identityId">The identity ID to be assigned to the </param>
+    public UserAssignedIdentityAnnotation(string identityId) : base("userAssignedIdentities", identityId)
+    {
+    }
+}
--- a/src/Aspire.Hosting.Azure/ApplicationModel/UserAssignedIdentityAnnotation.cs
+++ b/src/Aspire.Hosting.Azure/ApplicationModel/UserAssignedIdentityAnnotation.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Text.Json.Serialization;
 using Aspire.Hosting.ApplicationModel;
 
 namespace Aspire.Hosting.Azure.ApplicationModel;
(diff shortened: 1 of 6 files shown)


davidfowl:
This work is going to be post GA, but I've been working on this long-term branch that takes over bicep generation from azd https://github.com/dotnet/aspire/tree/davidfowl/gen-compute-infra. Instead of building up a bicep string, this will move to use the azure provisioning libraries for container apps (these don't exist yet). The way we extend or describe container app is something like https://github.com/Azure/azure-dev/issues/3292#issuecomment-1992226370. A callback that lets you describe the container app companion resource for the project/container resource.

Given that I think this scenario would look like:

```C#

var identity = builder.AddAzureUserAssignedIdentity("identity");

var storage = builder.AddAzureStorage("storage").AddBlob("blob");

// Dynamically create a role assignment between the identity and the storage account
builder.AddAzureRoleAssignment(storage, identity, RoleAssignment.StorageContributor);

var apiService = builder.AddProject<Projects.AspireAppWithHttps_ApiService>("apiservice")
                                      // AZURE_CLIENT_ID=identity.clientid
                                      .WithUserAssignedManagedIdentity(identity) 
                                      // Associate the identity with the container app
                                      .PublishAsAzureContainerApp(c => c.AddUserAssignedIdentity(identity)); 

builder.Build().Run();
```

I made up a ton of these APIs, I'm sure there will be higher and lower level in the fullness of time to represent what you want to. 

PS: Right now we're super busy cleaning up APIs and doing lots of fit and finish to get to GA. Soon after we will be looking at this in more detail.

Thanks for sticking in there 😄

ellismg:
@davidfowl - Happy to take the PR on the `azd` to support this if the manifest if you wanted to do something here before GA and you move all this logic into Aspire proper post GA.

rudiv:
> I'm not a fan of putting this in the manifest. It's too tied to azure in its current form. I understand the scenario we are trying to unblock but I rather not go down this route.

Understand this to an extent, and it's specifically why I added this in as a concept of a custom manifest output so that it's more general purpose than just tying everything to Azure / azd.

That is to say that custom manifest output is general purpose and could be used for much more than managed identity in the future. In my opinion it also opens up the manifest to other tools that are more specific to different environments (or even requirements, consider a scenario where you'd want to provide metadata to another app that isn't even deployment related, but you still want to be configured through your Aspire app from a developer perspective).

The Azure specific stuff is in the .Azure project (where it should be for that exact purpose?) and of course the changes I've made in `azd` are Azure specific too because, well it's an Azure Developer CLI 😃

I think what I'm getting at is that, whilst this specific implementation of the custom manifest output would be Azure specific, I don't feel that the concept actually is. I do understand that keeping the manifest as simple and generic as possible appears to be the goal here, but as old developers we all know that only goes so far.

This is quite useful to us now (using a custom build of azd) for deploying what we require without stepping down to Bicep too much or manually adjusting the YAML. Even if most Aspire users would just bung their services up using connection strings rather than identities, or stick with the default one that has Registry access etc etc, this is a very low-cost way to give even more flexibility.

PS. On your other reply @davidfowl I get where this is heading, it's just really hard to picture without the associated libs.

davidfowl:
I understand and appreciate the thinking and the PR but it's not the direction we want to go in.

rudiv:
OK, since the needle on support for this won't move until after GA. I've created a [package](https://github.com/rudiv/Aspire.Provisioning.RealWorld) that adds support for this for anyone else that might want more control.

Only MI / KV support at the moment, but will add more as we need it. Trying to move away from the plethora of custom bicep and towards the world I think we all envisage of Aspire being able to do everything in C#.

davidfowl:
Very cool package!

PR #3354, opened 2024-04-03
Add browser security headers to dashboard

JamesNK:
Content security policy is a guard against XSS. See https://learn.microsoft.com/en-us/aspnet/core/blazor/security/content-security-policy
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3354)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -247,6 +247,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
 
         _app.UseAuthorization();
 
+        _app.UseMiddleware<BrowserSecurityHeadersMiddleware>();
         _app.UseAntiforgery();
 
         _app.MapRazorComponents<App>().AddInteractiveServerRenderMode();
--- a/src/Aspire.Dashboard/Model/BrowserSecurityHeadersMiddleware.cs
+++ b/src/Aspire.Dashboard/Model/BrowserSecurityHeadersMiddleware.cs
@@ -1,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Dashboard.Authentication.OtlpConnection;
+
+namespace Aspire.Dashboard.Model;
+
+/// <summary>
+/// Middleware adds headers related to browser security that aren't built into ASP.NET Core:
+/// - Content-Security-Policy
+/// - Referrer-Policy
+/// - X-Content-Type-Options
+/// </summary>
+internal sealed class BrowserSecurityHeadersMiddleware
+{
+    private readonly RequestDelegate _next;
+    private readonly string _cspContent;
+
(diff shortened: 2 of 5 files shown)


DamianEdwards:
Can we add a test to verify CSP is set?

kvenkatrajan:
Do we need to specify X-Frame-Options header as well for click-jacking mitigation? : https://learn.microsoft.com/en-us/aspnet/core/blazor/security/server/interactive-server-side-rendering?view=aspnetcore-8.0#click-jacking

JamesNK:
> Can we add a test to verify CSP is set?

Like what? The browser uses the value and we don't have browser integration tests. And I'm not even sure how to add XSS content now that the dashboard is locked down to test with.

If you want, you can manually test by F5 an Aspire with the environment not set to development. You'll see the browser block BrowserLink and hot reload calls to VS.

JamesNK:
> Do we need to specify X-Frame-Options header as well for click-jacking mitigation? : [learn.microsoft.com/en-us/aspnet/core/blazor/security/server/interactive-server-side-rendering?view=aspnetcore-8.0#click-jacking](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/server/interactive-server-side-rendering?view=aspnetcore-8.0#click-jacking)

I can add it if there is no ask from anyone to put the site in an iframe.

DamianEdwards:
> Like what? The browser uses the value and we don't have browser integration tests. And I'm not even sure how to add XSS content to test with now that the dashboard is locked down.

I just mean check the required headers are sent in an integration test.

JamesNK:
All feedback applied.

JamesNK:
@SteveSandersonMS @javiercn It would be great if you could take a quick look at the header values and gives a thumbs up or down on the values chosen.

kvenkatrajan (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -247,6 +247,7 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
 
         _app.UseAuthorization();
 
+        _app.UseMiddleware<BrowserSecurityHeadersMiddleware>();
         _app.UseAntiforgery();
 
         _app.MapRazorComponents<App>().AddInteractiveServerRenderMode();
```
javiercn, line 253: You might want to check if you are targeting .NET 9.0. As compression is enabled by default and we setup an additional CSP policy value (ancestors-src: 'self'). 

I don't think it'll matter, as both policies will be applied and the most strict always win. Just letting you know in case you want to use a single header (you'll have to set one option in AddInteractiveServerRenderMode to null, and add it to your CSP policy header).

As per the other headers they look ok to me. It might be worth tightening up the CSP policy a bit more by avoiding unsafe-inline and using a nonce instead. Even though its only on CSS which is arguably way less problematic that on scripts.

Might also be worth hashing all the scripts and using hashes as opposed to `self`. That will protect you better against stored XSS if there is an encoding error.

PR #3370, opened 2024-04-03
Update table every second, reuse existing metric values if possible

### LINKED ISSUE #3363: When we navigate using TAB key after the table keyboard focus moves to top of the page instead 0f moving to next interactive control.:A11y_.NetAspireDashboard_Processruntimedotnetassembliescount_Table view_Keyboard.
"[Check out Accessibility Insights](https://nam06.safelinks.protection.outlook.com/?url=https://accessibilityinsights.io/&data=02%7c01%7cv-stfe%40microsoft.com%7cb67b2c4b646d4f9561a208d6f4b5c39b%7c72f988bf86f141af91ab2d7cd011db47%7c1%7c0%7c636965458850501301&sdata=mxhokIKNMb22llsjXHLgU3XZibj1Qfx37rpY4PU2sfE%3D&reserved=0)! - Identify accessibility bugs before check-in and make bug fixing faster and easier."

**Please do not close this bug. This bug should only be closed by Trusted Tester after verification.**

### Tags:
#A11yTCS; #A11yMAS; #A11ySev2; #Win11; #ChromiumEdge; #FTP; #DesktopWeb; #.NET Core; #SH_.Net Aspire Dashboard_Web_Mar2024; #Keyboard; #WCAG2.4.3;

### Environment Details:
Application Name: .Net Aspire Dashboard
URL: https://localhost:19888/
Chrome Version 122.0.2365.80 (Official build) (64-bit)
Window Version: Win 11
OS Build: 22621.3085

### Repro Steps:
1. Hit the URL: https://localhost:15887 and login with valid credentials.
2. TAB to "Metrics" tab item and hit ENTER key.
3. TAB to any existing instruction present under "Open telemetry instruction runtime" grid and hit ENTER key.
4. For example "Process runtime dotnet assemblies count" instruction selected.
5. "Process runtime dotnet assemblies count" control will open. TAB to "Table view" button and hit ENTER key.
6. Navigate using TAB key and observe the issue.

### Actual Results:
When we navigate using TAB key after the table keyboard focus moves to top of the page instead pf moving to next interactive control.

### Expected Results:
When we navigate using TAB key after the table keyboard focus should move to next interactive control.

### User Impact:
Keyboard only users will face difficulty here and will not be get information about the other interactive control present after the table.

### Attachment:
[3363_A11y_.NetAspireDashboard_Processruntimedotnetassembliescount_Table view.webm](https://github.com/dotnet/aspire/assets/90901519/d4ef0ff6-6cfa-4724-8724-fd8e12c27bca)

adamint:
We can associate rows with the underlying data's time range so that focus is not lost when the data is updated, if keyboard focus is active inside the table.

Fixes #3363

Animation

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
@@ -21,6 +21,7 @@ public partial class MetricTable : ChartBase
     private OtlpInstrument? _instrument;
     private bool _showCount;
     private bool _onlyShowValueChanges = true;
+    private DateTimeOffset? _lastUpdate;
 
     private readonly CancellationTokenSource _waitTaskCancellationTokenSource = new();
 
@@ -31,6 +32,14 @@ public partial class MetricTable : ChartBase
 
     protected override async Task OnChartUpdated(List<ChartTrace> traces, List<DateTimeOffset> xValues, bool tickUpdate, DateTimeOffset inProgressDataTime)
     {
+        // Only update table every second
+        if (inProgressDataTime - _lastUpdate < TimeSpan.FromSeconds(1))
+        {
+            return;
+        }
+
+        _lastUpdate = inProgressDataTime;
+
         if (!Equals(_instrument?.Name, InstrumentViewModel.Instrument?.Name) || _showCount != InstrumentViewModel.ShowCount)
         {
             _metrics.Clear();
@@ -42,8 +51,6 @@ public partial class MetricTable : ChartBase
         _showCount = InstrumentViewModel.ShowCount;
 
         _metrics = UpdateMetrics(out var xValuesToAnnounce, traces, xValues);
-
-        await InvokeAsync(StateHasChanged);
 
         if (xValuesToAnnounce.Count == 0)
         {
@@ -81,6 +88,11 @@ public partial class MetricTable : ChartBase
         for (var i = 0; i < xValues.Count; i++)
         {
             var xValue = xValues[i];
+            if (_metrics.TryGetValue(xValue, out var oldMetric))
+            {
+                newMetrics.Add(xValue, oldMetric);
+                continue;
+            }
 
             KeyValuePair<DateTimeOffset, MetricViewBase>? previousMetric = newMetrics.LastOrDefault(dt => dt.Key < xValue);
 
--- a/src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
@@ -32,7 +32,7 @@ public partial class MetricTable : ChartBase
 
     protected override async Task OnChartUpdated(List<ChartTrace> traces, List<DateTimeOffset> xValues, bool tickUpdate, DateTimeOffset inProgressDataTime)
     {
-        // Only update table every second
+        // Only update table every second to batch updates. New data coming every 200ms may be disorienting for screen-reader users.
         if (inProgressDataTime - _lastUpdate < TimeSpan.FromSeconds(1))
         {
             return;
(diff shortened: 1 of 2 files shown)


src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
```diff
@@ -43,8 +52,6 @@ protected override async Task OnChartUpdated(List<ChartTrace> traces, List<DateT
 
         _metrics = UpdateMetrics(out var xValuesToAnnounce, traces, xValues);
 
-        await InvokeAsync(StateHasChanged);
```
adamint, line 46: this was unnecessary

src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
```diff
@@ -31,6 +32,14 @@ public partial class MetricTable : ChartBase
 
     protected override async Task OnChartUpdated(List<ChartTrace> traces, List<DateTimeOffset> xValues, bool tickUpdate, DateTimeOffset inProgressDataTime)
     {
+        // Only update table every second
+        if (inProgressDataTime - _lastUpdate < TimeSpan.FromSeconds(1))
```
danmoseley, line 36: I don't know the code. If another update never comes in, does it mean that the data before the update we skipped here will display indefinitely? In other words, should this update be deferred rather than ignored?

adamint: > I think this logic is just evaluating how often to refresh the values (like a refresh interval)

Yep! No data is lost; the instrument updates its data every [200ms](https://github.com/dotnet/aspire/blob/88cece37d28245548e21735fe8a739a2e0ed7239/src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs#L91) after which this method is called.

src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor.cs
```diff
@@ -81,6 +88,11 @@ protected override async Task OnChartUpdated(List<ChartTrace> traces, List<DateT
         for (var i = 0; i < xValues.Count; i++)
         {
             var xValue = xValues[i];
+            if (_metrics.TryGetValue(xValue, out var oldMetric))
+            {
+                newMetrics.Add(xValue, oldMetric);
+                continue;
+            }
```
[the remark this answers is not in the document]
adamint: Currently, the behavior of `FluentDataGrid` is such that if you have focus on a row i (backed by some object items[i]) and then items[i] changes, focus is un-set. The current behavior of the metric table update function is to create new objects for each point on each update, which will result in focus being un-set within 200ms of focusing a row (since that's the time between instrument updates). 

If we preserve the old `MetricViewBase` object (since the data for the given timestamp would not have changed), focus will not be lost since the object will not have changed.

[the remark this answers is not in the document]
kvenkatrajan: And are dimensions a factor of metric? Can we have an update on multiple dimensions for the metrics at the same time? In that case we would want the focus to be respected - or am I overthinking this :)

tlmii (approved):
I'm not very familiar with the metrics code, but this change seems reasonable for the purpose.

My only question, for future proofing, would be: Are we sure that re-using the same object will always work? That just _feels_ a little "implementation detail". I know there's [ItemKey](https://www.fluentui-blazor.net/DataGrid#fluentdatagrid%3Ctgriditem%3E-class_properties) available on the data grid. Does that achieve a similar result, and if so, might that be more reliable?

JamesNK (changes requested):
I'd like some comments about why code is needed.

adamint:
> My only question, for future proofing, would be: Are we sure that re-using the same object will always work? That just feels a little "implementation detail". I know there's [ItemKey](https://www.fluentui-blazor.net/DataGrid#fluentdatagrid%3Ctgriditem%3E-class_properties) available on the data grid. Does that achieve a similar result, and if so, might that be more reliable?

Yes, that achieves the same result! I like that approach a lot more... did not realize ItemKey existed. I've updated the code to use that isntead

JamesNK:
It fails with a runtime error...

image

JamesNK (approved):

kvenkatrajan:
@adamint please backport for 8.0 - might be late for preview 6 - confirm.

PR #3372, opened 2024-04-03
Do not set TargetPort to Port when IsProxying is also set

ReubenBond:
When creating endpoint annotations, if a TargetPort is equal to the Port for a non-Container resource (and non-null) and IsProxying is true, that is an error: DCP cannot bind a proxy to a port which the target service is also bound to.

This PR does two things:
1. We do not set `TargetPort` to `Port` automatically if `IsProxying` is true.
2. If the resource is not a container and `TargetPort == Port && IsProxying`, then we throw during startup

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3372)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -44,9 +44,23 @@ public sealed class EndpointAnnotation : IResourceAnnotation
         _transport = transport;
         Name = name;
         Port = port;
-        TargetPort = targetPort ?? port;
+
+        TargetPort = targetPort;
+
+        // If the target port was not explicitly set and the service is not being proxied,
+        // we can set the target port to the port.
+        if (TargetPort is null && !isProxied)
+        {
+            TargetPort = port;
+        }
+
         IsExternal = isExternal ?? false;
         IsProxied = isProxied;
+
+        if (TargetPort is int && Port is int && TargetPort == Port && IsProxied)
+        {
+            throw new ArgumentException($"When {nameof(isProxied)} is true, {nameof(targetPort)} must not be equal to {nameof(port)}.");
+        }
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -56,11 +56,6 @@ public sealed class EndpointAnnotation : IResourceAnnotation
 
         IsExternal = isExternal ?? false;
         IsProxied = isProxied;
-
-        if (TargetPort is int && Port is int && TargetPort == Port && IsProxied)
-        {
-            throw new ArgumentException($"When {nameof(isProxied)} is true, {nameof(targetPort)} must not be equal to {nameof(port)}.");
-        }
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -1692,7 +1692,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
             {
                 if (sp.EndpointAnnotation.TargetPort is null)
                 {
-                    throw new InvalidOperationException($"The endpoint for container resource {modelResourceName} must specify the ContainerPort");
+                    throw new InvalidOperationException($"The endpoint for container resource '{modelResourceName}' must specify the ContainerPort");
                 }
 
                 sp.DcpServiceProducerAnnotation.Port = sp.EndpointAnnotation.TargetPort;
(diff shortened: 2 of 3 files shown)


src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -44,9 +44,23 @@ public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, strin
...
+        {
+            TargetPort = port;
+        }
+
         IsExternal = isExternal ?? false;
         IsProxied = isProxied;
+
+        if (TargetPort is int && Port is int && TargetPort == Port && IsProxied)
+        {
+            throw new ArgumentException($"When {nameof(isProxied)} is true, {nameof(targetPort)} must not be equal to {nameof(port)}.");
```
karolz-ms, line 62: This unfortunately depends on what this annotation is applied to. It will work for a singleton container; in fact, that is how we ran in Aspire P1 timeframe when there were no proxyless services. So this check will probably have to be moved to ApplicationExecutor.

karolz-ms (changes requested):

karolz-ms (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1692,7 +1692,7 @@ private void AddServicesProducedInfo(IResource modelResource, IAnnotationHolder
             {
                 if (sp.EndpointAnnotation.TargetPort is null)
                 {
-                    throw new InvalidOperationException($"The endpoint for container resource {modelResourceName} must specify the ContainerPort");
+                    throw new InvalidOperationException($"The endpoint for container resource '{modelResourceName}' must specify the ContainerPort");
```
eerhardt, line 1695:
```suggestion
                    throw new InvalidOperationException($"The endpoint for container resource '{modelResourceName}' must specify the TargetPort");
```

?

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -278,14 +278,17 @@ public void WriteBindings(IResource resource)
...
-                int? targetPort = (resource, endpoint.UriScheme, endpoint.TargetPort) switch
+                int? targetPort = (resource, endpoint.UriScheme, endpoint.TargetPort, endpoint.Port) switch
                 {
                     // The port was specified so use it
-                    (_, _, int port) => port,
+                    (_, _, int target, _) => target,
 
-                    // Project resources get their default port from the deployment tool
+                    // Container resources get their default listening port from the exposed port.
+                    (ContainerResource, _, _, int port) => port,
```
davidfowl, line 287: Hmm not sure this is right. We care about the target port itself being null in this case.

eerhardt: Doesn't it have to be `null` in this case? If it wasn't 284 would have taken it.

ReubenBond: Yes, it would already have matched, but it's good to be explicit anyway.

eerhardt (approved):
LGTM

davidfowl:
Can you re gen the manifests? I want to see if there are any changes

davidfowl:
cc @vhvb1989

eerhardt:
> > Can you re gen the manifests?
> 
> Is there a way to do that across all projects? We have some tests which check manifest output and they did not change

We should really add an official way to do it to the repo, but here's a hack powershell script that I used a few weeks ago:

```
$playground_apps = @(
    "C:\git\aspire2\playground\AWS\AWS.AppHost\",
    "C:\git\aspire2\playground\AzureSearchEndToEnd\AzureSearch.AppHost\",
    "C:\git\aspire2\playground\AzureStorageEndToEnd\AzureStorageEndToEnd.AppHost\",
    "C:\git\aspire2\playground\bicep\BicepSample.AppHost\",
    "C:\git\aspire2\playground\cdk\CdkSample.AppHost\",
    "C:\git\aspire2\playground\CosmosEndToEnd\CosmosEndToEnd.AppHost\",
    "C:\git\aspire2\playground\CustomResources\CustomResources.AppHost",
    "C:\git\aspire2\playground\dapr\AppHost\",
    "C:\git\aspire2\playground\DatabaseMigration\DatabaseMigration.AppHost",
    "C:\git\aspire2\playground\mongo\Mongo.AppHost\",
    "C:\git\aspire2\playground\mysql\MySqlDb.AppHost\",
    "C:\git\aspire2\playground\nats\Nats.AppHost\",
    "C:\git\aspire2\playground\OpenAIEndToEnd\OpenAIEndToEnd.AppHost\",
    "C:\git\aspire2\playground\orleans\OrleansAppHost",
    "C:\git\aspire2\playground\ParameterEndToEnd\ParameterEndToEnd.AppHost\",
    "C:\git\aspire2\playground\PostgresEndToEnd\PostgresEndToEnd.AppHost\",
    "C:\git\aspire2\playground\ProxylessEndToEnd\ProxylessEndToEnd.AppHost\",
    "C:\git\aspire2\playground\seq\Seq.AppHost\",
    "C:\git\aspire2\playground\signalr\SignalRAppHost\",
    "C:\git\aspire2\playground\SqlServerEndToEnd\SqlServerEndToEnd.AppHost\",
    "C:\git\aspire2\playground\TestShop\AppHost\")

foreach ($playground_app in $playground_apps)
{
Push-Location $playground_app
dotnet run -- --publisher manifest --output-path aspire-manifest.json
Pop-Location
}
```

ReubenBond:
I ran the script and compared main vs prior to this commit and got the same results.

PR #2049, opened 2024-02-02
Fix component governance alerts

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2049)

PR #3373, opened 2024-04-03
Lift up the dependency of Newtonsoft.Json in Cosmos components

joperezr:
FYI @Pilchie @danmoseley 

cc: @eerhardt @sebastienros 

https://github.com/dotnet/aspire/pull/2049 was trying to remove this CG alert but because this was a transitive dependency, it didn't actually remove the issue. This PR is adding an explicit reference to both of the Aspire Components which has the transitive dependency triggering CG alerts in order to address this.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3373)

### DIFF
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/Aspire.Microsoft.Azure.Cosmos.csproj
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/Aspire.Microsoft.Azure.Cosmos.csproj
@@ -21,6 +21,9 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
+
+    <!-- Pinned version for Component Governance - Remove when root dependencies are updated tracked by https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4302 -->
+    <PackageReference Include="Newtonsoft.Json" />
   </ItemGroup>
 
 </Project>
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/Aspire.Microsoft.Azure.Cosmos.csproj
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/Aspire.Microsoft.Azure.Cosmos.csproj
@@ -22,7 +22,9 @@
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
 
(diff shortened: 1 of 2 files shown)


danmoseley:
What's the risk of this causing a functional break? (Is the product already loading the newer version in our context, so this is purely about suppressing the dependency warning thus very low risk?)

joperezr:
Not necessarily, depends on the app on whether or not it is being lifted. I wouldn't necessarily say this is very low risk, which is why I think it is fine to keep it in main for now and give it some bake time before we consider if we want to backport.

danmoseley:
Right, I understand now. Yup, agreed.

sebastienros:
Weird that my PR was only declaring it in Directory.Packages.props, I am definitely aware it would be of no effect, it's just a constant. Thanks for fixing it.

danmoseley (approved):
Is there any additional test we can/should do here? If not, seems fine to merge into main.

src/Components/Aspire.Microsoft.Azure.Cosmos/Aspire.Microsoft.Azure.Cosmos.csproj
```diff
@@ -21,6 +21,9 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
+
+    <!-- Pinned version for Component Governance - Remove when root dependencies are updated tracked by https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4302 -->
```
eerhardt, line 25: 1. From reading the linked issues, it sounds like this will never be fixed. Does a comment to remove this add value?
2. Maybe we should link to the root issue instead of a dupe-of-a-dupe-of-a-dupe.

eerhardt (approved):
I just have some suggestions on the comments. The actual code changes look fine.

eerhardt:
> I wouldn't necessarily say this is very low risk, which is why I think it is fine to keep it in main for now and give it some bake time before we consider if we want to backport.

I don't think waiting is necessary. I would bring this change into preview6, definitely before we GA.

If the CosmosDB code doesn't work with the latest Newtonsoft.Json version, we have bigger issues.

The reason the CosmosDB package doesn't update their reference is because they consider it a breaking change on their customers who could be using some API that broke between 10 and 13.

joperezr:
I'd ex

> Is there any additional test we can/should do here? If not, seems fine to merge into main.

Cosmos components have good test coverage already which would exercise the new dependency version. As Eric points out, they haven't bumped the version mainly for compat reasons.

Pilchie:
Agree with @eerhardt here - happy to see this merged.

PR #3374, opened 2024-04-03
Fix validation error messages

drewnoakes:
Looks like a copy/paste bug. It could mislead a user who is trying to configure the dashboard correctly.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3374)

### DIFF
--- a/src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
@@ -88,10 +88,16 @@ public sealed class ValidateDashboardOptions : IValidateOptions<DashboardOptions
                                 errorMessages.Add("Dashboard:ResourceServiceClient:ClientCertificate:Source is \"KeyStore\", but no Dashboard:ResourceServiceClient:ClientCertificate:Subject is configured.");
                             }
                             break;
+                        case null:
+                            errorMessages.Add($"The resource service client is configured to use certificates, but no certificate source is specified. Specify Dashboard:ResourceServiceClient:ClientCertificate:Source. Possible values: {string.Join(", ", typeof(DashboardClientCertificateSource).GetEnumNames())}");
+                            break;
                         default:
                             errorMessages.Add($"Unexpected resource service client certificate source: {options.ResourceServiceClient.ClientCertificates.Source}");
                             break;
                     }
+                    break;
+                case null:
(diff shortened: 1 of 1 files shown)


src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -90,12 +89,12 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
                             }
                             break;
                         default:
```
JamesNK, line 91: There should be a case for null. It should say the value is required when a resource service is configured. See the switch statement for options.Otlp.AuthMode as an example.

JamesNK (commented):
Add unit tests that the right error message is present when a resource service URL is configured without an auth mode.

drewnoakes:
Please file a separate issue (with a link to this) for tests since we have multiple areas that need to be covered post preview 6.

tlmii (approved):
This change itself seems good. It does seem like James is right there should be a null case also. Not sure if you want to put that in this PR or do it separately (or someone file an issue). Approving with the assumption that one of those happens, because this does fix the stated issue.

drewnoakes:
Pushed an update to handle the two null config values.

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -88,11 +88,17 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
...
+                        case null:
+                            errorMessages.Add($"The resource service client is configured to use certificates, but no certificate source is specified. Specify Dashboard:ResourceServiceClient:ClientCertificate:Source. Possible values: {string.Join(", ", typeof(DashboardClientCertificateSource).GetEnumNames())}");
+                            break;
                         default:
                             errorMessages.Add($"Unexpected resource service client certificate source: {options.ResourceServiceClient.ClientCertificates.Source}");
                             break;
                     }
                     break;
+                case null:
+                    errorMessages.Add($"Resource service client authentication is not configured. Either specify {DashboardConfigNames.DashboardUnsecuredAllowAnonymousName.ConfigKey}=true, or specify {DashboardConfigNames.ResourceServiceAuthModeName.ConfigKey}. Possible values: {string.Join(", ", typeof(ResourceClientAuthMode).GetEnumNames())}");
```
JamesNK, line 100:
```suggestion
                    errorMessages.Add($"Resource service client authentication is not configured. Specify {DashboardConfigNames.ResourceServiceAuthModeName.ConfigKey}. Possible values: {string.Join(", ", typeof(ResourceClientAuthMode).GetEnumNames())}");
```
Allow anonymous doesn't impact dashboard to resource service communication.

JamesNK (approved):

JamesNK:
Please create an issue for adding tests.

PR #3375, opened 2024-04-03
Generate and store random RG name when provisioning

### LINKED ISSUE #3259: Generate and persist random resource group names, don't use the machine name.

sebastienros:
Fixes https://github.com/dotnet/aspire/issues/3259
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3375)


sebastienros:
Only issue with that is that it's hard to clean the subscription since my name/machine is not part of the resource group anymore. A solution could be to have an optional prefix (no friction) in the options to add to the unique name.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -362,14 +363,24 @@ await foreach (var tenant in armClient.GetTenants().GetAllAsync(cancellationToke
...
 
-        // Name of the resource group to create based on the machine name and application name
-        var (resourceGroupName, createIfAbsent) = _options.ResourceGroup switch
+        if (string.IsNullOrEmpty(_options.ResourceGroup))
         {
-            null or { Length: 0 } => ($"rg-aspire-{unique}", true),
-            string rg => (rg, _options.AllowResourceGroupCreation ?? false)
-        };
+            // Create a unique resource group name and save it in user secrets
+            resourceGroupName = $"rg-aspire-{RandomNumberGenerator.GetHexString(8, true)}-{environment.ApplicationName.ToLowerInvariant()}";
```
eerhardt, line 374:
```suggestion
            resourceGroupName = $"rg-aspire-{environment.ApplicationName.ToLowerInvariant()}-{RandomNumberGenerator.GetHexString(8, true)}";
```

Maybe put the constant thing at the front, so they are sorted together.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -362,14 +363,24 @@ await foreach (var tenant in armClient.GetTenants().GetAllAsync(cancellationToke
...
 
-        // Name of the resource group to create based on the machine name and application name
-        var (resourceGroupName, createIfAbsent) = _options.ResourceGroup switch
+        if (string.IsNullOrEmpty(_options.ResourceGroup))
         {
-            null or { Length: 0 } => ($"rg-aspire-{unique}", true),
-            string rg => (rg, _options.AllowResourceGroupCreation ?? false)
-        };
+            // Create a unique resource group name and save it in user secrets
+            resourceGroupName = $"rg-aspire-{environment.ApplicationName.ToLowerInvariant()}-{RandomNumberGenerator.GetHexString(8, true)}";
```
danmoseley, line 374: aside, I wish they had called it SecureRandomNumberGenerator. Each time I see this I have to stop and check it's the crypto one and not some local thing we have in the project.

sebastienros:
Ping

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
davidfowl, line 374: Lowercase the hex string

timheuer: is even ToLowerInvariant enough? Will RG rule naming allow diacritics, etc? - `azd` likely went through this same exercise already?

sebastienros: > Lowercase the hex string

This is what the second argument is for already.

sebastienros: While I added diacritics support, I tried and anything seems to work for resource groups:

image

I tried emojis, they passed validation but failed on creation.

timheuer: Just as a note, `azd` doesn't support this as environment name (which gets used in RG creation). Not sure if we want to be consistent but pointing out they use [a-zA-Z0-9] and underscore/dash validation only.

sebastienros: In the meantime I added try RG name support.

> this as environment name

You mean "resource group" name, and also that we should actually remove diacritics and then also filter on `[a-zA-Z0-9_\-]` to be sure it works with azd?

davidfowl: Is there an overall length constraint as well?

timheuer: > Is there an overall length constraint as well?

Yes. 90 on the RG name, but deployments are limited to 64 and if this gets used with deployment names as well in azureprovisioning, that's something to consider.

Here's what azd does: https://github.com/Azure/azure-dev/blob/6abd371120c98393a8c3ab0c0dba2a0a92285648/cli/azd/pkg/environment/environment.go#L114

sebastienros: Now it's normalizing with `azAZ90-_`, it's limiting to 90 chars, and it's also using an optional custom RG prefix because at least for us there is no way to record the ownership of RGs. Tried it locally (i.e. on Azure).

src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
```diff
@@ -14,6 +14,7 @@
     <Compile Include="..\Aspire.Hosting\Dcp\Process\ProcessSpec.cs" Link="Provisioning\Utils\ProcessSpec.cs" />
     <Compile Include="..\Aspire.Hosting\Dcp\Process\ProcessUtil.cs" Link="Provisioning\Utils\ProcessUtil.cs" />
     <Compile Include="..\Shared\Cosmos\CosmosConstants.cs" Link="Shared\CosmosConstants.cs" />
+    <Compile Include="$(SharedDir)ResourceGroupNameHelpers.cs" Link="shared\ResourceGroupNameHelpers.cs" />
```
eerhardt, line 17: Does this need to be in a `Shared` directory? It is only used by 1 project.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -362,14 +364,44 @@ await foreach (var tenant in armClient.GetTenants().GetAllAsync(cancellationToke
...
+            if (!string.IsNullOrWhiteSpace(_options.ResourceGroupPrefix))
+            {
+                prefix = _options.ResourceGroupPrefix;
+            }
+
+            var suffix = RandomNumberGenerator.GetHexString(8, lowercase: true);
+
+            var maxApplicationNameSize = ResourceGroupNameHelpers.MaxResourceGroupNameLength - prefix.Length - suffix.Length - 2; // extra '-'s
+
+            var normalizedApplicationName = ResourceGroupNameHelpers.NormalizeResourceGroupNameForAzd(environment.ApplicationName.ToLowerInvariant());
```
eerhardt, line 387: Why do we ToLower the ApplicationName? That makes it harder to read IMO because I don't get PascalCasing anymore.

src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
sebastienros: Moved to Aspire.Hosting.Azure

src/Aspire.Hosting.Azure/Utils/ResourceGroupNameHelpers.cs
```diff
@@ -0,0 +1,86 @@
...
+                continue;
+            }
+
+            stringBuilder.Append(c);
+        }
+
+        // Can't end with '.'
+        return stringBuilder.ToString().TrimEnd('.');
+    }
+
```
eerhardt, line 40:
```suggestion
```

Dead code

src/Aspire.Hosting.Azure/Utils/ResourceGroupNameHelpers.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <summary>
+    /// Try to convert or excludes any characters which are not valid resource group name components for AZD.
```
eerhardt, line 42: > for AZD

This is kind of weird. Why are we special casing "for AZD"?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
sebastienros: Was like this before the PR. If you get support or strongly think so then I'll change it.
NB: In the portal they are case-insensitive if that helps deciding we can keep the original casing.

src/Aspire.Hosting.Azure/Utils/ResourceGroupNameHelpers.cs
sebastienros: @timheuer 's comment was that if we ask AZD to create an RG that it won't accept then this could be an issue.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
eerhardt: @davidfowl - is there a reason for this? Seems like it has negative value IMO.

src/Aspire.Hosting.Azure/Utils/ResourceGroupNameHelpers.cs
eerhardt: I think it is fine to make this a comment, but naming the method "ForAzd" seems super weird.

sebastienros: Removed the mention.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -362,14 +364,44 @@ await foreach (var tenant in armClient.GetTenants().GetAllAsync(cancellationToke
...
-            null or { Length: 0 } => ($"rg-aspire-{unique}", true),
-            string rg => (rg, _options.AllowResourceGroupCreation ?? false)
-        };
+            // Generate an resource group name since none was provided
+
+            var prefix = "rg-aspire";
+
+            if (!string.IsNullOrWhiteSpace(_options.ResourceGroupPrefix))
+            {
+                prefix = _options.ResourceGroupPrefix;
```
eerhardt, line 380: Why aren't we normalizing the `ResourceGroupPrefix`?

sebastienros: For the same reason we are not doing it for the `ResourceGroup` option? I think no since this is a technical value provided by the user. For `ApplicationName` the user should not have to worry about naming limitations.

Otherwise let's do it for both. And maybe log a warning that the name was changed if provided?

eerhardt: Ok, if we think that since this option is named "ResourceGroupPrefix", users will be responsible for using the correct characters.

eerhardt (approved):
I still think we shouldn't be "ToLower"ing the Application name, but since that is the current behavior, I think these changes are fine.

timheuer (approved):

PR #3383, opened 2024-04-04
Uses new Azure.Provisioning packages which have been split by resource

mitchdenny:
This PR (draft until we get a `0.1.0-beta.x build`) incorporates a change that the Azure SDK team has made around the `Azure.Provisioning` package. They are breaking it up into multiple packages. Here is what is in this PR:

# Migration to Azure.Provisioning.* packages

The Azure SDK team has split up the `Azure.Provisioning` packages into mutliple packages based on resource. This is something that we had always planned on happening as it better aligns with the way their engineering system works. We held off doing it for P5 due to the tight timeframe but it is now ready for P6.

# Fix for AppInsights/LogAnalytics issue

Through some manual testing I found an edge case that I hadn't tested earlier in relation to integration of log analytics and app insights. If someone does not provide a log analytics workspace explicitly then we were not creating one (we used to with our Bicep based approach). This was not an issue for deployment with AZD because it would inject the log analytics workspace ID. However if you don't provide a workspace AND you use the Azure Provisioner the deployment would fail. This PR includes fixes for that issue as well as covering test cases so that we don't accidentally regress this behavior.

# Firewall rules

Reacted to some changes in the firewall configuration for resources via the CDK. Unfortunately we won't be able to add an IP based firewall for KeyVault, Storage, and Cosmos resources. For KeyVault and Storage - they don't treat ACA as a trusted service (probably got something to do with the highly ephemeral IP ranges for this service). For Cosmos you can enable the firewall, but upon inspection of the side effects of doing this manually in the Portal it appears that a set of hard coded IP values are added. This is probably not something that we want to do in Aspire so we'll be better off waiting until we figure out how to make Aspire work well with virutal networks.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3383)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -33,22 +33,22 @@
     <PackageVersion Include="Microsoft.Azure.SignalR" Version="1.25.1" />
     <PackageVersion Include="Microsoft.Extensions.Azure" Version="1.7.2" />
     <!-- Azure Management SDK for .NET dependencies -->
-    <PackageVersion Include="Azure.Provisioning" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.AppConfiguration" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.ApplicationInsights" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.CognitiveServices" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.CosmosDB" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.EventHubs" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.KeyVault" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.OperationalInsights" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.PostgreSql" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.Redis" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.Resources" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.Search" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.ServiceBus" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.SignalR" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.Sql" Version="0.1.0-alpha.20240403.2" />
-    <PackageVersion Include="Azure.Provisioning.Storage" Version="0.1.0-alpha.20240403.2" />
+    <PackageVersion Include="Azure.Provisioning" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.AppConfiguration" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.ApplicationInsights" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.CognitiveServices" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.CosmosDB" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.EventHubs" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.KeyVault" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.OperationalInsights" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.PostgreSql" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Redis" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Resources" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Search" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.ServiceBus" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.SignalR" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Sql" Version="0.2.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Storage" Version="0.2.0-beta.1" />
     <!-- ASP.NET Core dependencies -->
     <PackageVersion Include="Microsoft.AspNetCore.Authentication.Certificate" Version="$(MicrosoftAspNetCoreAuthenticationCertificatePackageVersion)" />
     <PackageVersion Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="$(MicrosoftAspNetCoreAuthenticationOpenIdConnectPackageVersion)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -33,7 +33,22 @@
     <PackageVersion Include="Microsoft.Azure.SignalR" Version="1.25.1" />
     <PackageVersion Include="Microsoft.Extensions.Azure" Version="1.7.2" />
     <!-- Azure Management SDK for .NET dependencies -->
-    <PackageVersion Include="Azure.Provisioning" Version="0.1.0-beta.2" />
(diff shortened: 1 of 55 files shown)


mitchdenny:
@joperezr @danmoseley ... heads up for an incoming change for P6/GA.

NuGet.config
```diff
@@ -19,11 +19,18 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="orleans-nightly" value="https://orleans.pkgs.visualstudio.com/orleans-public/_packaging/orleans-nightly/nuget/v3/index.json" />
     <add key="dotnet9-transport" value="https://dnceng.pkgs.visualstudio.com/public/_packaging/dotnet9-transport/nuget/v3/index.json" />
+    <add key="azure-sdk-for-net-dev" value="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" />
```
eerhardt, line 22: When will these be available on nuget.org? Doing this isn't keeping us in a "shippable state" and we will need to remove this feed and update the versions before we release preview6.

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
```diff
@@ -49,6 +49,11 @@ public static IResourceBuilder<AzureStorageResource> AddAzureStorage(this IDistr
                 sku: StorageSkuName.StandardGrs
                 );
 
+            // Unfortunately Azure Storage does not list ACA as one of the resource types in which
```
eerhardt, line 52: > Unfortunately 

Is there some sort of issue tracking this? Do we need to raise this problem somewhere?

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
```diff
@@ -23,26 +23,25 @@ resource eventHubsNamespace_skb4aVCrD 'Microsoft.EventHub/namespaces@2022-10-01-
     name: sku
   }
   properties: {
-    minimumTlsVersion: '1.2'
```
eerhardt, line 26: Where is this change coming from? Is 1.2 the default and we don't need to explicitly specify it in the bicep?

Also minor nit - why is it generating an empty `properties` object. Couldn't that just be omitted?

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
```diff
@@ -10,8 +10,8 @@ param principalId string
 param principalType string
 
 
-resource storageAccount_X1L2R0Ykm 'Microsoft.Storage/storageAccounts@2022-09-01' = {
-  name: toLower(take(concat('ehstorage', uniqueString(resourceGroup().id)), 24))
+resource storageAccount_59pNEh4K4 'Microsoft.Storage/storageAccounts@2022-09-01' = {
```
eerhardt, line 13: Do these names change _every time_ the bicep is generated? If so, how do we expect people to do something like infra synth and easily diff when there are so many diffs.

playground/ProxylessEndToEnd/ProxylessEndToEnd.AppHost/aspire-manifest.json
```diff
@@ -9,6 +9,7 @@
           "scheme": "tcp",
           "protocol": "tcp",
           "transport": "tcp",
+          "port": 9999,
```
eerhardt, line 12: I assume/know this isn't caused by your change, but are these ports supposed to go into the manifest?

src/Aspire.Hosting.Azure.ApplicationInsights/AzureApplicationInsightsExtensions.cs
```diff
@@ -76,15 +77,26 @@ public static IResourceBuilder<AzureApplicationInsightsResource> AddAzureApplica
...
             {
+                // If someone provides a workspace via the extension method we should use it.
                 appInsights.AssignProperty(p => p.WorkspaceResourceId, logAnalyticsWorkspace.Resource.WorkspaceId, AzureBicepResource.KnownParameters.LogAnalyticsWorkspaceId);
             }
-            else
+            else if (builder.ExecutionContext.IsRunMode)
             {
+                // ... otherwise if we are in run mode, the provisioner expects us to create one ourselves.
+                var autoInjectedLogAnalyticsWorkspace = new OperationalInsightsWorkspace(construct, name: $"law-{construct.Resource.Name}");
+                appInsights.Properties.Tags["aspire-resource-name"] = $"law-{construct.Resource.Name}";
```
eerhardt, line 87: Is this correct? Is it supposed to be:

```suggestion
                autoInjectedLogAnalyticsWorkspace.Properties.Tags["aspire-resource-name"] = $"law-{construct.Resource.Name}";
```

Also, can we remove the duplicate `$"law-{construct.Resource.Name}"` and just use `autoInjectedLogAnalyticsWorkspace.Name` to reuse the value we set on line 86?

NuGet.config
mitchdenny: This will be removed before the PR merges. Working through this with the Azure SDK team. This was just to get it working and validated so we can greenlight that.

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
mitchdenny: 1.2 would be the default.

The empty properties field might be necessary to satisfy the RP.

Letting @JoshLove-msft know about the change here.

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
mitchdenny: We changed the formula we use for resource name generation deliberately to address some linter warnings. This bicep was first generated using the old method and had not been updated until I regenerated everything.

These should be stable moving forward. It's a good call out though as we cannot change these moving forward.

playground/ProxylessEndToEnd/ProxylessEndToEnd.AppHost/aspire-manifest.json
mitchdenny: Yes. I just happen to be the one that regenerates all the manifests ;)

src/Aspire.Hosting.Azure.ApplicationInsights/AzureApplicationInsightsExtensions.cs
mitchdenny: That name field is the bicep construct name. And the other one would be an expression. 3 different values. But I can tweak this a bit so I only generate the string once.

Directory.Packages.props
```diff
@@ -33,7 +33,22 @@
...
+    <PackageVersion Include="Azure.Provisioning.KeyVault" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.OperationalInsights" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.PostgreSql" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Redis" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Resources" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Search" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.ServiceBus" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.SignalR" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Sql" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning.Storage" Version="0.1.0-beta.1" />
```
davidfowl, line 51: Why not use a variable here?

eng/refreshManifests.ps1
```diff
@@ -0,0 +1,2 @@
+..\build.cmd
```
davidfowl, line 1: Why does it build first?

Directory.Packages.props
mitchdenny: Because they will eventually start releasing at a different cadence. Right now these package are all in the sdk/provisioning folder but they will end up splitting them out into sdk/[service] folders and will tend to release on a cadence tied to the underlying management plane packages.

eng/refreshManifests.ps1
mitchdenny: To make sure that everything is up to date because the loop uses --no-build (to speed things up).

Directory.Packages.props
davidfowl: Can't we change it when that happens?

eng/refreshManifests.ps1
davidfowl: This shouldn't run build.cmd. Let users run that.

davidfowl:
As part of these firewall rule changes, these services still all work when using your local dev machine right? Those rules only apply when the compute is in azure?

mitchdenny:
Correct. I've been running the compute locally and using the services.

mitchdenny:
@davidfowl we need to get this into main.

davidfowl (approved):

davidfowl:
Lets merge it!

PR #3391, opened 2024-04-04
Fix dashboard token being written on publish

### LINKED ISSUE #3387: When publishing manifest, dashboard URL with token is output
Started happening after adding auth to the dashboard.

image

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/3387
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3391)


src/Aspire.Hosting/DistributedApplicationLifecycle.cs
```diff
@@ -21,13 +21,13 @@ public Task StartAsync(CancellationToken cancellationToken)
...
     public Task StartedAsync(CancellationToken cancellationToken)
     {
-        if (distributedApplicationOptions.DashboardEnabled && configuration["AppHost:BrowserToken"] is { Length: > 0 } browserToken)
-        {
-            LoggingHelpers.WriteDashboardUrl(logger, configuration["ASPNETCORE_URLS"], browserToken);
-        }
-
         if (executionContext.IsRunMode)
         {
+            if (distributedApplicationOptions.DashboardEnabled && configuration["AppHost:BrowserToken"] is { Length: > 0 } browserToken)
```
mitchdenny, line 26: Not a bad idea to omit this output if the browser token is missing 👍

But you might also wish to omit the output if IsPublishMode == true (look at the `DistributedApplicationExecutionContext` for this. If someone modifies their launchprofile to pass in the `--publisher manifest` argument then they'll produce a manifest but this output will still be output, because VS would have injected the browser token (based on my understanding of how the browser token injection mechanism is going to work).

mitchdenny: Approving now since that is more of an edge case (we certainly experience it in our repo but we are probably outliers).

mitchdenny (approved):

src/Aspire.Hosting/DistributedApplicationLifecycle.cs
JamesNK: It's nested inside an `IsRunMode` check. If there is a publisher value then won't that be false?

JamesNK: It isn't included when run from VS with `"commandLineArgs": "--publisher manifest --output-path test.json"` in launch profile:

image

PR #3401, opened 2024-04-04
Make status column icons colorblind friendly

drewnoakes:
Ensures these icons are differentiated not only by colour, but also by icon.

Also fixes this issue, pointed out offline by @timheuer:

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3401)

### DIFF
--- a/src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
+++ b/src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
@@ -51,12 +51,12 @@ else if (!string.IsNullOrEmpty(Resource.StateStyle))
                         Class="severity-icon" />
             break;
         case "success":
-            <FluentIcon Icon="Icons.Filled.Size16.Circle"
+            <FluentIcon Icon="Icons.Filled.Size16.CheckmarkCircle"
                         Color="Color.Success"
                         Class="severity-icon" />
             break;
         case "info":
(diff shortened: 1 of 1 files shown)


mitchdenny (approved):

tlmii (approved):

timheuer (approved):

adamint (approved):

danmoseley:
do any of the 3 remaining uses of ""Icons.Filled.Size16.Circle"" in that file need to be visually distinguishable?

drewnoakes:
I think the featureless circle makes sense for "unknown" state, which is two of the three. The other is "starting" state. Perhaps that could be an outlined circle.

drewnoakes:
Did we ever consider using a spinner for "starting" status items?

https://blazorfluentui.azurewebsites.net/spinnerPage

danmoseley:
Looks appropriate to me, my 2c

danmoseley:
> The other is "starting" state. Perhaps that could be an outlined circle.

@drewnoakes do we still need to distinguish this, to be accessible?

drewnoakes:
> @drewnoakes do we still need to distinguish this, to be accessible?

Yes, https://github.com/dotnet/aspire/pull/3689

I decided against using a spinner, mostly because it didn't seem possible to make the Blazor Fluent UI one that small.

PR #3400, opened 2024-04-04
Resource service supports API key authentication

drewnoakes:
Adds support for an API key to authenticate the resource service.

Unless unsecured, the app host will generate an API key and pass it to the dashboard via an environment variable. The dashboard then includes this key in a header for all gRPC calls. The app host's resource service validates that the expected key is received and rejects requests where the key is omitted.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3400)

PR #3402, opened 2024-04-04
Localize "Unknown" resource state label

drewnoakes:
This string appears whenever a resource status is unknown. It should be localized.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3402)


drewnoakes:
CI failure seems unrelated to these changes. Also occurring in https://github.com/dotnet/aspire/pull/3400.

tlmii (approved):

adamint (approved):

JamesNK:
Should this be localized? Other states aren't localized.

I think this change should be to put a note against the literal string that it shouldn't be localized because resource states aren't localized.

PR #3422, opened 2024-04-05
Addressing port/endpoint issues on Qdrant

### LINKED ISSUE #3421: Port clash and missing EndpointReference in Qdrant resource
RE: https://github.com/dotnet/aspire/pull/3131#pullrequestreview-1981525558

- Port clash if port specified, need separate port options
- Don't use primary endpoint reference for different endpoint

timheuer:
Fix #3421
- Separate port params
- update playground
- Modify test to verify port in manifest
- Separated EndpointReferences
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3422)

### DIFF
--- a/playground/Qdrant/Qdrant.AppHost/Program.cs
+++ b/playground/Qdrant/Qdrant.AppHost/Program.cs
@@ -3,7 +3,7 @@
 
 var builder = DistributedApplication.CreateBuilder(args);
 
-var qdrant = builder.AddQdrant("qdrant")
+var qdrant = builder.AddQdrant("qdrant", httpPort:5503)
     .WithDataVolume("qdrant_data");
 
 builder.AddProject<Projects.Qdrant_ApiService>("apiservice")
--- a/playground/Qdrant/Qdrant.AppHost/Program.cs
+++ b/playground/Qdrant/Qdrant.AppHost/Program.cs
@@ -3,7 +3,7 @@
 
 var builder = DistributedApplication.CreateBuilder(args);
 
-var qdrant = builder.AddQdrant("qdrant", httpPort:5503)
+var qdrant = builder.AddQdrant("qdrant")
     .WithDataVolume("qdrant_data");
 
 builder.AddProject<Projects.Qdrant_ApiService>("apiservice")
--- a/src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
@@ -13,7 +13,7 @@ namespace Aspire.Hosting;
 public static class QdrantBuilderExtensions
 {
     private const int QdrantPortGrpc = 6334;
-    private const int QdrantPortHttp = 6333;
+    private const int QdrantPortRest = 6333;
     private const string ApiKeyEnvVarName = "QDRANT__SERVICE__API_KEY";
     private const string EnableStaticContentEnvVarName = "QDRANT__SERVICE__ENABLE_STATIC_CONTENT";
 
@@ -28,13 +28,13 @@ public static class QdrantBuilderExtensions
     /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency</param>
     /// <param name="apiKey">The parameter used to provide the API Key for the Qdrant resource. If <see langword="null"/> a random key will be generated as {name}-Key.</param>
     /// <param name="grpcPort">The host port of gRPC endpoint of Qdrant database.</param>
-    /// <param name="httpPort">The host port of REST endpoint of Qdrant database.</param>
+    /// <param name="restPort">The host port of REST endpoint of Qdrant database.</param>
     /// <returns>A reference to the <see cref="IResourceBuilder{QdrantServerResource}"/>.</returns>
     public static IResourceBuilder<QdrantServerResource> AddQdrant(this IDistributedApplicationBuilder builder,
         string name,
         IResourceBuilder<ParameterResource>? apiKey = null,
         int? grpcPort = null,
-        int? httpPort = null)
+        int? restPort = null)
     {
         var apiKeyParameter = apiKey?.Resource ??
             ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-Key", special: false);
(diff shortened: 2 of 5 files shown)


src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -27,20 +27,22 @@ public static class QdrantBuilderExtensions
...
     /// <param name="apiKey">The parameter used to provide the API Key for the Qdrant resource. If <see langword="null"/> a random key will be generated as {name}-Key.</param>
-    /// <param name="port">The host port of Qdrant database.</param>
+    /// <param name="gRpcPort">The host port of gRPC endpoint of Qdrant database.</param>
+    /// <param name="httpPort">The host port of REST endpoint of Qdrant database.</param>
     /// <returns>A reference to the <see cref="IResourceBuilder{QdrantServerResource}"/>.</returns>
     public static IResourceBuilder<QdrantServerResource> AddQdrant(this IDistributedApplicationBuilder builder,
         string name,
         IResourceBuilder<ParameterResource>? apiKey = null,
-        int? port = null)
+        int? gRpcPort = null,
```
davidfowl, line 36:
```suggestion
        int? grpcPort = null,
```

playground/Qdrant/Qdrant.AppHost/Program.cs
```diff
@@ -3,7 +3,7 @@
 
 var builder = DistributedApplication.CreateBuilder(args);
 
-var qdrant = builder.AddQdrant("qdrant")
+var qdrant = builder.AddQdrant("qdrant", httpPort:5503)
```
davidfowl, line 6: Remove the port?

davidfowl, line 6:
```suggestion
var qdrant = builder.AddQdrant("qdrant")
```

remove

timheuer: Was in there to have a validated manifest test...you don't see the need?

eerhardt (approved):
LGTM

tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs
```diff
@@ -174,7 +174,7 @@ public async Task QdrantClientAppWithReferenceContainsConnectionStrings()
     public async Task VerifyManifest()
     {
         var appBuilder = DistributedApplication.CreateBuilder(new DistributedApplicationOptions() { Args = new string[] { "--publisher", "manifest" } } );
-        var qdrant = appBuilder.AddQdrant("qdrant");
+        var qdrant = appBuilder.AddQdrant("qdrant", httpPort:5503);
```
eerhardt, line 177: Instead of modifying this test, we should add a new one. I can push a new test to this branch.

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
```diff
@@ -27,20 +27,22 @@ public static class QdrantBuilderExtensions
...
     {
         var apiKeyParameter = apiKey?.Resource ??
             ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-Key", special: false);
         var qdrant = new QdrantServerResource(name, apiKeyParameter);
         return builder.AddResource(qdrant)
             .WithImage(QdrantContainerImageTags.Image, QdrantContainerImageTags.Tag)
-            .WithHttpEndpoint(port: port, targetPort: QdrantPortGrpc, name: QdrantServerResource.PrimaryEndpointName)
-            .WithHttpEndpoint(port: port, targetPort: QdrantPortHttp, name: QdrantServerResource.RestEndpointName)
+            .WithHttpEndpoint(port: grpcPort, targetPort: QdrantPortGrpc, name: QdrantServerResource.PrimaryEndpointName)
+            .WithHttpEndpoint(port: httpPort, targetPort: QdrantPortHttp, name: QdrantServerResource.RestEndpointName)
```
eerhardt, line 45: Is it confusing that the `httpPort` goes to the `rest` endpoint, and not the endpoint that is named `http`?

Maybe we should either rename the endpoints, or rename these port parameters?

timheuer: I thought about this as developing. I was trying to match the way the Qdrant docs refer to the exposed API endpoints (as gRPC and REST).  But the more I think about it, it's mixing weird names.  

As you re-raise the question here @eerhardt , I think maybe representing the endpoints what they are (grpc and http) eliminating the 'high level' speak is maybe better...naming them clearly for what the endpoint communicates.

Thoughts? Broader quick rename -- change default endpoint name from 'http' to 'grpc' and change 'rest' to 'http' everywhere?

eerhardt (approved):

src/Aspire.Hosting.Qdrant/QdrantBuilderExtensions.cs
eerhardt: I changed the default endpoint name to 'grpc' but left the other one 'rest`. I can rename it to 'http' everywhere if you'd like.

The part that I don't like about using 'http' is that they are both HTTP. Where at least "rest" gives a little more indication of the difference. But I can rename it to 'http' if you'd like.

timheuer: or the way you just did ;-)

eerhardt: Looking for a little bit of previous art, Dapr says:

https://github.com/dotnet/aspire/blob/e2b1b46044bb74e9d7ccce54d00ada58d411b947/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs#L154-L156

Not saying we'd have to match it, but it at least shows we use this naming pattern elsewhere.

eerhardt: I was thinking Qdrant docs being the winner. 😆 

I can honestly go either way. If I had to pick I would pick the way it is now in the PR mostly because it matches how Qdrant refers to them. But I can live with renaming "rest" => "http" if you'd like.

timheuer: Ugh, sorry not having this convo in original PR.  Let's rename to http -- the docs are mixed. But the **config** docs consistently use 'http_port' as what is being exposed, let's rename rest->http

PR #3433, opened 2024-04-05
Distinguish Orleans hosts & clients, parameterize CluserId/ServiceId, improve playground app

### LINKED ISSUE #3399: OrleansService generates a new GUID each time.
@ReubenBond @benjaminpetit  I noticed that every time we run manifest generation on the Orleans playground sample, the cluster ID is recreated. This is probably not what we want for deployment scenarios. Instead we should probably be auto injecting a parameter into the app model like @eerhardt did for Postgres admin usernames and passwords.

ReubenBond:
Fixes #3399
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3433)


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -400,10 +400,11 @@ private static void ApplyEndpoints<T>(this IResourceBuilder<T> builder, IResourc
...
     /// <param name="name">An optional name of the endpoint. Defaults to the scheme name if not specified.</param>
     /// <param name="env">An optional name of the environment variable to inject.</param>
+    /// <param name="isExternal">Indicates that this endpoint should be exposed externally at publish time.</param>
     /// <param name="isProxied">Specifies if the endpoint will be proxied by DCP. Defaults to true.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     /// <exception cref="DistributedApplicationException">Throws an exception if an endpoint with the same name already exists on the specified resource.</exception>
-    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, string? name = null, string? env = null, bool isProxied = true) where T : IResource
+    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null) where T : IResource
```
ReubenBond, line 407: @davidfowl cool with this? I know we default this value to false internally, but I don't want to rely on default behavior for these bindings: they should not be made external.

src/Aspire.Hosting.Orleans/OrleansService.cs
```diff
@@ -1,34 +1,59 @@
...
+public class OrleansService
 {
+    /// <summary>Initializes a new <see cref="OrleansService"/> instance.</summary>
+    /// <param name="builder">The distributed application builder.</param>
+    /// <param name="name">The service name.</param>
+    public OrleansService(IDistributedApplicationBuilder builder, string name)
+    {
+        Name = name;
+        Builder = builder;
+        ServiceId = ParameterResourceBuilderExtensions.CreateGeneratedParameter(builder, $"{name}-service-id", secret: false, new GenerateParameterDefault
```
davidfowl, line 20: @eerhardt this API is kinda crazy 😄

src/Aspire.Hosting/ResourceBuilderExtensions.cs
davidfowl: Yes this is fine.

ReubenBond:
@mitchdenny @davidfowl @eerhardt PTAL

src/Aspire.Hosting.Orleans/OrleansService.cs
eerhardt: Suggestions on making it better?

src/Aspire.Hosting.Orleans/OrleansService.cs
```diff
@@ -1,34 +1,59 @@
...
+            Numeric = true,
+            MinLength = 25
+        });
+        ClusterId = ParameterResourceBuilderExtensions.CreateGeneratedParameter(builder, $"{name}-cluster-id", secret: false, new GenerateParameterDefault
+        {
+            Lower = true,
+            Upper = false,
+            Special = false,
+            Numeric = true,
+            MinLength = 25
```
eerhardt, line 34:
```suggestion
            Upper = false,
            Special = false,
            MinLength = 25
        });
        ClusterId = ParameterResourceBuilderExtensions.CreateGeneratedParameter(builder, $"{name}-cluster-id", secret: false, new GenerateParameterDefault
        {
            Upper = false,
            Special = false,
            MinLength = 25
```

You don't need to specify the `true`s.

You can also create a helper method for this to reduce the code duplication:

```
ServiceId = CreateGeneratedParameter(builder,  $"{name}-service-id");
ClusterId = CreateGeneratedParameter(builder,  $"{name}-cluster-id");

static ParameterResource CreateGeneratedParameter(IDistributedApplicationBuilder builder, string name) => 
ParameterResourceBuilderExtensions.CreateGeneratedParameter(builder, name, secret: false, new GenerateParameterDefault
    {
        Upper = false,
        Special = false,
        MinLength = 25
    });

File: src/Aspire.Hosting.Orleans/OrleansService.cs [hunk 4]
```diff
@@ -1,34 +1,59 @@
...
     /// Gets the distributed application builder.
     /// </summary>
-    public IDistributedApplicationBuilder Builder { get; } = builder;
+    public IDistributedApplicationBuilder Builder { get; }
 
     /// <summary>
     /// Gets or sets the service identifier.
     /// </summary>
-    public string? ServiceId { get; set; }
+    internal object ServiceId { get; set; }
```
eerhardt, line 51: Why `object`?

```suggestion
    internal ParameterResource ServiceId { get; set; }
```

ReubenBond: It's either a parameter or a hard-coded string value.. I made it internal so at least this detail doesn't leak.

See `WithServiceId` and `WithClusterId`

eerhardt: > It's either a parameter or a hard-coded string value.

Might be good sticking that comment in here.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
eerhardt, line 407: Just remember we can't make changes like this after we ship stable.

It might make sense to get API compat turned on after 8.0 ships.

src/Aspire.Hosting.Orleans/OrleansService.cs
```diff
@@ -1,34 +1,55 @@
...
+
 namespace Aspire.Hosting.Orleans;
 
 /// <summary>
 /// Describes an Orleans service.
 /// </summary>
-/// <param name="builder">The distributed application builder.</param>
-/// <param name="name">The service name.</param>
-public class OrleansService(IDistributedApplicationBuilder builder, string name)
+public class OrleansService
```
eerhardt, line 11:
```suggestion
public sealed class OrleansService
```

Do we expect people to derive from this?

src/Aspire.Hosting.Orleans/OrleansServiceClientExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+public static class OrleansServiceClientExtensions
+{
+    /// <summary>
+    /// Adds an Orleans client to the resource.
+    /// </summary>
+    /// <param name="builder">The builder on which add the Orleans service builder.</param>
+    /// <param name="orleansServiceClient">The Orleans service client, containing clustering, etc.</param>
+    /// <returns>The resource builder.</returns>
+    /// <exception cref="InvalidOperationException">Clustering has not been configured.</exception>
+    public static IResourceBuilder<T> WithReference<T>(
```
eerhardt, line 21: When would I use `WithReference(OrleansServiceClient)` over using `WithReference(OrleansService)`?

Is it confusing to have both? What's the difference between:

```C#
builder.AddProject<Projects.OrleansServer>("silo")
       .WithReference(orleans);
```

and 
```C#
builder.AddProject<Projects.OrleansClient>("frontend")
       .WithReference(orleans.AsClient());
```

?

ReubenBond: You use the service one for projects with an Orleans server and the client one with the Orleans clients. Their configuration follows the same format, but servers have more (storage, etc)

src/Aspire.Hosting.Orleans/OrleansServiceExtensions.cs
```diff
@@ -348,25 +380,40 @@ public static class OrleansServiceExtensions
...
+            foreach (var (name, provider) in res.GrainDirectory)
+            {
+                provider.ConfigureResource(builder, $"GrainDirectory__{name}");
+            }
+
+            // Set silo-to-silo and client-to-gateway ports
+            // NOTE: These endpoints are specified as proxied even though we never expect any connections via the proxy, and that would be invalid anyway.
+            // this is a workaround for current Aspire/DCP behavior.
+            builder.WithEndpoint(scheme: "tcp", name: "orleans-silo", env: "Orleans__Endpoints__SiloPort", isProxied: true, isExternal: false);
+            builder.WithEndpoint(scheme: "tcp", name: "orleans-gateway", env: "Orleans__Endpoints__GatewayPort", isProxied: true, isExternal: false);
```
eerhardt, line 416: I'm a bit confused by this comment.

1. Isn't "isProxied" defaulted to `true`?
2. What exactly is this working around?

eerhardt (approved):
I just had one last question. The rest LGTM

src/Aspire.Hosting.Orleans/OrleansServiceExtensions.cs
ReubenBond: IsProxied is defaulted to true. In this case, we don't actually want a proxy, but we need to make sure IsProxied is set to true, because otherwise DCP will barf. It's a limitation. The proxy ends up being unused by the app (they connect directly to each other), so it's benign but worth mentioning in a comment and being explicit about the current behavior.

PR #3438, opened 2024-04-05
Do not use ExecutionType.IDE when not debugging

ReubenBond:
When running tests from Visual Studio IntPreview (a build from last week), I observe that we are telling DCP to execute projects & executables via the IDE even when we are not debugging the tests. This fails because the IDE does not have a debug session when not debugging. This fix works around the issue by only enabling IDE execution when the debugger is attached.

The specific error I see is:
```
  | [2024-04-02T22:04:08] Aspire.Hosting.Dcp.dcpctrl.ExecutableReconciler Error: failed to start Executable	{"Executable": {"name":"servicea_3e8589f4-ej16bj0"}, "Reconciliation": 97, "error": "run session could not be started: failed to connect to IDE run session notification endpoint: websocket: bad handshake"}
```

EDIT: I cannot seem to reproduce this on the current build, so perhaps it's been fixed in VS?

cc @BillHiebert @karolz-ms 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3438)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -1115,7 +1115,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
             var projectLaunchConfiguration = new ProjectLaunchConfiguration();
             projectLaunchConfiguration.ProjectPath = projectMetadata.ProjectPath;
 
-            if (!string.IsNullOrEmpty(configuration[DebugSessionPortVar]))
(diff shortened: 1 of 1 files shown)


karolz-ms:
This should work and we do NOT want this change. The execution type should only be driven by the presence of IDE debug session var. Having debugger detached is equivalent to running in Ctrl+F5 mode, still via IDE, but without debugging.

BillHiebert:
This should just work w/o doing anything. We need to investiage

ReubenBond:
Agreed that we don't want this change. I cannot reproduce this anymore. It was reproducing consistently on Tuesday, but I've updated VS since then. Maybe there was an issue which was already fixed somewhere.

Issue #808, opened 2023-11-14
Add a package description to Aspire.ProjectTemplates


davidfowl:
https://www.nuget.org/packages/Aspire.ProjectTemplates/8.0.0-preview.1.23557.2

PR #3443, opened 2024-04-05
Add User Profile Menu for OpenID Connect login

### LINKED ISSUE #3194: Display logout link for authenticated user
With #3033 the dashboard can be configured to support oidc. When authenticated a user may wish to log out of their current authenticated state. Provide a link for logout in the dashboard. 

CC: @drewnoakes

### LINKED ISSUE #3197: Add user info on dashboard views once logged in
With oidc - user can be authenticated. Add user info as per issue comment: https://github.com/dotnet/aspire/pull/3033#issuecomment-2012367564

CC:  @drewnoakes

tlmii:
Resolves #3194 
Resolves #3197 

- Introduces the `FluentProfileMenu` to the upper-right of the site header when the user is authenticated with OpenID Connect.
- Displays the user's full name and user id (often, but not always, an email) as well as their initials.
- Adds two config settings - `Dashboard:Frontend:OpenIdConnect:NameClaimType` and `Dashboard:Frontend:OpenIdConnect:UsernameClaimType` to allow for customization of which claims determine the values displayed in the profile menu
- Provides a sign out button on the profile menu allowing the user to sign out - this clears the cookie and redirects them to the IdP where they can also sign out if they wish.

A couple screenshots:

image
image

TODO:
- [ ] Figure out what the header of the profile menu should be. Currently says OpenID Connect since that's the auth mode we're in. Could maybe find a way to tie it to the name of the IdP?  
- [ ] Test with another IdP (so far: Entra ID and Keycloak)
- [ ] Figure out if there's anything we can do (another config setting? more docs?) to make the OpenIdConnect code we use interoperate with Keycloak's logout better (right now it gets an error because `post_logout_redirect_uri` is passed but `id_token_hint` is not, which they consider an error even though the spec doesn't appear to).

Note: Testing this PR will require you to use your own Idp since this will only appear with OpenID Connect auth enabled. If you want to test it with TestShop (or other playground apps) you'll also need to disable the automatic setting of the authmode. Easiest way I could find to do that was to comment out this line:

https://github.com/dotnet/aspire/blob/29e2b5747d7ed696b015be59518370847fced705/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs#L126

But maybe there's a better way.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3443)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
@@ -10,7 +10,7 @@
                                    FooterLink="">
                     <HeaderTemplate>
                         <FluentStack VerticalAlignment="VerticalAlignment.Center">
-                            <FluentLabel>@Loc[nameof(Resources.Login.SignedInAs)]</FluentLabel>
+                            <FluentLabel>@Loc[nameof(Resources.Login.LoggedInAs)]</FluentLabel>
                             <FluentSpacer />
                             <form @formname="sign-out-form" class="sign-out-form" action="authentication/logout" method="post">
                                 <AntiforgeryToken />
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
@@ -1,0 +1,33 @@
+﻿<AuthorizeView>
+    <Authorized>
+        @if (_showUserProfileMenu)
+        {
+            <div class="profile-menu-container">
+                <FluentProfileMenu Initials="@_initials"
+                                   EMail="@_username"
+                                   FullName="@_name"
+                                   ButtonSize="@ButtonSize"
+                                   FooterLink="">
+                    <HeaderTemplate>
+                        <FluentStack VerticalAlignment="VerticalAlignment.Center">
+                            <FluentLabel>OpenID Connect</FluentLabel>
+                            <FluentSpacer />
+                            <form @formname="sign-out-form" class="sign-out-form" action="authentication/logout" method="post">
+                                <AntiforgeryToken />
+                                <FluentButton Appearance="Appearance.Stealth" Type="ButtonType.Submit">Sign out</FluentButton>
+                            </form>
+                        </FluentStack>
+                    </HeaderTemplate>
+                    <ChildContent>
+                        <FluentStack class="inner-persona-container">
+                            <FluentPersona Initials="@_initials" ImageSize="@ImageSize" Class="inner-persona">
+                                <div class="full-name">@_name</div>
+                                <div class="user-id">@_username</div>
+                            </FluentPersona>
+                        </FluentStack>
+                    </ChildContent>
+                </FluentProfileMenu>
+            </div>
+        }
+    </Authorized>
+</AuthorizeView>
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
@@ -10,7 +10,7 @@
                                    FooterLink="">
                     <HeaderTemplate>
                         <FluentStack VerticalAlignment="VerticalAlignment.Center">
-                            <FluentLabel>OpenID Connect</FluentLabel>
+                            <FluentLabel>@Loc[nameof(Resources.Login.SignedInAs)]</FluentLabel>
                             <FluentSpacer />
                             <form @formname="sign-out-form" class="sign-out-form" action="authentication/logout" method="post">
                                 <AntiforgeryToken />
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor.cs
@@ -1,0 +1,72 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Security.Claims;
+using Aspire.Dashboard.Configuration;
+using Aspire.Dashboard.Extensions;
+using Aspire.Dashboard.Resources;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Authorization;
+using Microsoft.Extensions.Localization;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Components.Controls;
+
+public partial class UserProfile : ComponentBase
+{
+    [Inject]
+    public required IOptionsMonitor<DashboardOptions> DashboardOptions { get; set; }
+
+    [CascadingParameter]
+    public required Task<AuthenticationState> AuthenticationState { get; set; }
+
+    [Inject]
+    public required IStringLocalizer<Login> Loc { get; set; }
+
+    [Inject]
+    public required ILogger<UserProfile> Logger { get; set; }
+
+    [Parameter]
+    public string ButtonSize { get; set; } = "24px";
+
+    [Parameter]
+    public string ImageSize { get; set; } = "52px";
+
+    private bool _showUserProfileMenu;
+    private string? _name;
+    private string? _username;
+    private string? _initials;
+
+    protected override async Task OnParametersSetAsync()
+    {
+        if (DashboardOptions.CurrentValue.Frontend.AuthMode == FrontendAuthMode.OpenIdConnect)
+        {
+            var authState = await AuthenticationState;
+
+            var claimsIdentity = authState.User.Identity as ClaimsIdentity;
+
+            if (claimsIdentity?.IsAuthenticated == true)
+            {
+                _showUserProfileMenu = true;
+                _name = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.OpenIdConnect.GetNameClaimTypes());
+                if (string.IsNullOrWhiteSpace(_name))
+                {
+                    // Make sure there's always a name, even if that name is a placeholder
+                    _name = Loc[nameof(Login.AuthorizedUser)];
+                }
+
+                _username = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.OpenIdConnect.GetUsernameClaimTypes());
+                _initials = _name.GetInitials();
+            }
+            else
+            {
+                // If we don't have an authenticated user, don't show the user profile menu. This shouldn't happen.
+                _showUserProfileMenu = false;
+                _name = null;
+                _username = null;
+                _initials = null;
+                Logger.LogError("Dashboard:Frontend:AuthMode is configured for OpenIDConnect, but there is no authenticated user.");
+            }
+        }
+    }
+}
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor.cs
@@ -48,14 +48,14 @@ public partial class UserProfile : ComponentBase
             if (claimsIdentity?.IsAuthenticated == true)
             {
                 _showUserProfileMenu = true;
-                _name = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.OpenIdConnect.GetNameClaimTypes());
+                _name = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.Frontend.OpenIdConnect.GetNameClaimTypes());
                 if (string.IsNullOrWhiteSpace(_name))
                 {
                     // Make sure there's always a name, even if that name is a placeholder
                     _name = Loc[nameof(Login.AuthorizedUser)];
                 }
 
-                _username = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.OpenIdConnect.GetUsernameClaimTypes());
+                _username = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.Frontend.OpenIdConnect.GetUsernameClaimTypes());
                 _initials = _name.GetInitials();
             }
             else
--- a/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
+++ b/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
@@ -50,6 +50,7 @@
 
 ::deep.layout > header {
     grid-area: head;
+    /*overflow-x: clip;*/
 }
 
 ::deep.layout > .nav-menu-container {
@@ -74,9 +75,9 @@
     margin-left: auto;
 }
 
-::deep.layout > header fluent-button[appearance=stealth]:not(:hover)::part(control),
-::deep.layout > header fluent-anchor[appearance=stealth]:not(:hover)::part(control),
-::deep.layout > header fluent-anchor[appearance=stealth].logo::part(control),
+::deep.layout > header > .header-gutters > fluent-button[appearance=stealth]:not(:hover)::part(control),
+::deep.layout > header > .header-gutters > fluent-anchor[appearance=stealth]:not(:hover)::part(control),
+::deep.layout > header > .header-gutters > fluent-anchor[appearance=stealth].logo::part(control),
 ::deep.layout > .aspire-icon fluent-anchor[appearance=stealth].logo::part(control) {
     background-color: var(--neutral-layer-4);
 }
@@ -86,7 +87,7 @@
     margin-bottom: 0;
 }
 
-::deep.layout > header fluent-anchor {
+::deep.layout > header > .header-gutters > fluent-anchor {
     font-size: var(--type-ramp-plus-2-font-size);
 }
 
--- a/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
+++ b/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
@@ -50,7 +50,6 @@
 
 ::deep.layout > header {
     grid-area: head;
-    /*overflow-x: clip;*/
 }
 
 ::deep.layout > .nav-menu-container {
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -206,7 +206,7 @@ public sealed class OpenIdConnectOptions
         }
         else
         {
-            _usernameClaimTypes = UsernameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.RemoveEmptyEntries);
+            _usernameClaimTypes = UsernameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
         }
 
         errorMessages = messages;
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -196,7 +196,7 @@ public sealed class OpenIdConnectOptions
         }
         else
         {
-            _nameClaimTypes = NameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.RemoveEmptyEntries);
+            _nameClaimTypes = NameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
         }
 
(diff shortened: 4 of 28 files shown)


src/Aspire.Dashboard/Extensions/StringExtensions.cs
```diff
@@ -7,24 +7,6 @@ namespace Aspire.Dashboard.Extensions;
 
 internal static class StringExtensions
 {
-    /// <summary>
-    /// Shortens a string by replacing the middle with an ellipsis.
-    /// </summary>
-    /// <param name="text">The string to shorten</param>
-    /// <param name="maxLength">The max length of the result</param>
-    public static string TrimMiddle(this string text, int maxLength)
```
tlmii, line 15: This hasn't been used for a while, so I deleted it and the unit test cases.

src/Aspire.Dashboard/Components/Controls/UserProfile.razor
```diff
@@ -0,0 +1,33 @@
...
+        {
+            <div class="profile-menu-container">
+                <FluentProfileMenu Initials="@_initials"
+                                   EMail="@_username"
+                                   FullName="@_name"
+                                   ButtonSize="@ButtonSize"
+                                   FooterLink="">
+                    <HeaderTemplate>
+                        <FluentStack VerticalAlignment="VerticalAlignment.Center">
+                            <FluentLabel>OpenID Connect</FluentLabel>
```
tlmii, line 13: This is what I referenced in one of the TODOs in the description. We can:

1. Leave it as-is
2. Set it to something else constant
3. Set it to something else constant and localize it
4. Figure out some way to get an appropriate value from an IdP
5. Remove it altogether

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
```diff
@@ -49,6 +49,7 @@
 
 ::deep.layout > header {
     grid-area: head;
+    /*overflow-x: clip;*/
```
tlmii, line 52: Will remove this comment in next update

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
```diff
@@ -68,9 +69,9 @@
     margin-left: auto;
 }
 
-::deep.layout > header fluent-button[appearance=stealth]:not(:hover)::part(control),
-::deep.layout > header fluent-anchor[appearance=stealth]:not(:hover)::part(control),
-::deep.layout > header fluent-anchor[appearance=stealth].logo::part(control),
+::deep.layout > header > .header-gutters > fluent-button[appearance=stealth]:not(:hover)::part(control),
```
tlmii, line 72: Made a bunch of these header styles more specific to the elements we were trying to impact so they wouldn't change the style of things within the FluentProfileMenu

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -15,6 +15,7 @@ public sealed class DashboardOptions
     public FrontendOptions Frontend { get; set; } = new FrontendOptions();
     public ResourceServiceClientOptions ResourceServiceClient { get; set; } = new ResourceServiceClientOptions();
     public TelemetryLimitOptions TelemetryLimits { get; set; } = new TelemetryLimitOptions();
+    public OpenIdConnectOptions OpenIdConnect { get; set; } = new OpenIdConnectOptions();
```
tlmii, line 18: @JamesNK @drewnoakes I'm open to suggestions on how these options should be configured. I didn't see (though definitely could have missed) a way to handle it within the OpenIDConnect infrastructure. I know there's further work on configuring policies, etc.

src/Aspire.Dashboard/Components/Controls/UserProfile.razor.cs
```diff
@@ -0,0 +1,72 @@
...
+            var claimsIdentity = authState.User.Identity as ClaimsIdentity;
+
+            if (claimsIdentity?.IsAuthenticated == true)
+            {
+                _showUserProfileMenu = true;
+                _name = claimsIdentity.FindFirst(DashboardOptions.CurrentValue.OpenIdConnect.GetNameClaimTypes());
+                if (string.IsNullOrWhiteSpace(_name))
+                {
+                    // Make sure there's always a name, even if that name is a placeholder
+                    _name = Loc[nameof(Login.AuthorizedUser)];
```
tlmii, line 55: Open to suggestions on what, if anything, we should display in this case. I've seen this type of placeholder before so I went with it for now.

kvenkatrajan:
> igure out what the header of the profile menu should be. Currently says OpenID Connect since that's the auth mode we're in.
Can we update to "Signed in as:" Or "Profile" for now? @leslierichardson95 thoughts?

kvenkatrajan:
Looks wonderful by the way - great job 👍

leslierichardson95:
> >f igure out what the header of the profile menu should be. Currently says OpenID Connect since that's the auth mode we're in.
> 
> 
> 
> Can we update to "Signed in as:" Or "Profile" for now? @leslierichardson95 thoughts?

@kvenkatrajan I like "Signed in as:".

src/Aspire.Dashboard/Extensions/ClaimsIdentityExtensions.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Security.Claims;
+
+namespace Aspire.Dashboard.Extensions;
+
+public static class ClaimsIdentityExtensions
```
davidfowl, line 8: Not public

tlmii:
> > > f igure out what the header of the profile menu should be. Currently says OpenID Connect since that's the auth mode we're in.
> > 
> > 
> > Can we update to "Signed in as:" Or "Profile" for now? @leslierichardson95 thoughts?
> 
> @kvenkatrajan I like "Signed in as:".

Updated:

image

Sign Out being on the same line as "Signed in as:" feels a little off to me now. But it still makes more sense than what was there before I think.

src/Aspire.Dashboard/Extensions/ClaimsIdentityExtensions.cs
tlmii: Whoops. Fixed.

JamesNK:
FYI the browser token auth uses `Logged in` and `Login` in its UI. Consider changing text here to be consistent.

mitchdenny:
What are we doing around authorization?

kvenkatrajan:
> What are we doing around authorization?

We have #808 tracking this.

tlmii:
> FYI the browser token auth uses `Logged in` and `Login` in its UI. Consider changing text here to be consistent.

Switched to "Logged in as:"

If we can't find a way to pull the name of the IdP from somewhere, maybe we can make it an option. I'd like that better than this (and it'd be consistent with Teams, Azure Portal, etc). But I don't want to hold this PR up for that.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
tlmii: Moved it.

mitchdenny:
808 seemed related to project templates?

kvenkatrajan:
Ah its 808 on the other board: https://github.com/dotnet/aspire-private-planning/issues/808

src/Aspire.Dashboard/Components/Controls/UserProfile.razor
drewnoakes: Another option is just to remove the text. I think it's pretty clear what the popup means in context:

image

src/Aspire.Dashboard/README.md
```diff
@@ -53,10 +53,12 @@ Set `Dashboard:Frontend:AuthMode` to `BrowserToken`. Browser token authenticatio
...
-- `Authentication:Schemes:OpenIdConnect:ClientId` &mdash; Identity of the relying party (RP)
-- `Authentication:Schemes:OpenIdConnect:ClientSecret`&mdash; A secret that only the real RP would know
+- `Authentication:Schemes:OpenIdConnect:Authority` URL to the identity provider (IdP)
+- `Authentication:Schemes:OpenIdConnect:ClientId` Identity of the relying party (RP)
+- `Authentication:Schemes:OpenIdConnect:ClientSecret` A secret that only the real RP would know
 - Other properties of [`OpenIdConnectOptions`](https://learn.microsoft.com/dotnet/api/microsoft.aspnetcore.builder.openidconnectoptions) specified in configuration container `Authentication:Schemes:OpenIdConnect:*`
+- `Dashboard:Frontend:OpenIdConnect:NameClaimType` specifies the claim type(s) that should be used to display the authenticated user's full name. Can be a single claim type or a comma-delimited list of claim types. Defaults to `name`.
+- `Dashboard:Frontend:OpenIdConnect:UsernameClaimType` specifies the claim type(s) that should be used to display the authenticated user's username. Can be a single claim type or a comma-delimited list of claim types. Defaults to `preferred_username`.
```
drewnoakes, line 61: nit: the PR description doesn't mention `Frontend` in these config keys. Perhaps update the PR description in case someone uses that as a reference.

drewnoakes (approved):
Looks great.

Approving, though as I mentioned above I think the popup could be simplified:

image

src/Aspire.Dashboard/README.md
tlmii: Whoops, forgot to update that when I changed the code. Thanks for calling that out.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -163,3 +164,52 @@ public sealed class TelemetryLimitOptions
     public int MaxAttributeLength { get; set; } = int.MaxValue;
     public int MaxSpanEventCount { get; set; } = int.MaxValue;
 }
+
+public sealed class OpenIdConnectOptions
```
drewnoakes, line 168: For consistency with other config classes in this file:

```suggestion
// Don't set values after validating/parsing options.
public sealed class OpenIdConnectOptions
```

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -100,6 +100,11 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
             }
         }
 
+        if (!options.Frontend.OpenIdConnect.TryParseOptions(out var messages))
+        {
+            errorMessages.AddRange(messages);
+        }
```
drewnoakes, line 106: What about moving this to the `case FrontendAuthMode.OpenIdConnect:` on line 24 above?

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -163,3 +164,52 @@ public sealed class TelemetryLimitOptions
...
+        }
+
+        if (string.IsNullOrWhiteSpace(UsernameClaimType))
+        {
+            messages ??= [];
+            messages.Add("OpenID Connect claim type for username not configured. Specify a Dashboard:OpenIdConnect:UsernameClaimType value.");
+        }
+        else
+        {
+            _usernameClaimTypes = UsernameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.RemoveEmptyEntries);
```
drewnoakes, line 209:
```suggestion
            _usernameClaimTypes = UsernameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -271,6 +271,10 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
             });
 #endif
         }
+        else if (dashboardOptions.Frontend.AuthMode == FrontendAuthMode.OpenIdConnect)
+        {
+            _app.MapPost("/authentication/logout", () => TypedResults.SignOut(authenticationSchemes: [CookieAuthenticationDefaults.AuthenticationScheme, "OpenIdConnect"]));
```
drewnoakes, line 276:
```suggestion
            _app.MapPost("/authentication/logout", () => TypedResults.SignOut(authenticationSchemes: [CookieAuthenticationDefaults.AuthenticationScheme, OpenIdConnectDefaults.AuthenticationScheme]));
```

src/Aspire.Dashboard/Extensions/ClaimsIdentityExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+internal static class ClaimsIdentityExtensions
+{
+    /// <summary>
+    /// Searches the claims in the <see cref="ClaimsIdentity.Claims"/> for each of the claim types in <paramref name="claimTypes" />
+    /// in the order presented and returns the first one that it finds.
+    /// </summary>
+    public static string? FindFirst(this ClaimsIdentity identity, string[] claimTypes)
+    {
+        if (identity is null)
```
drewnoakes, line 16: `identity` is not annotated as accepting `null`.

src/Aspire.Dashboard/Extensions/StringExtensions.cs
```diff
@@ -49,4 +31,29 @@ static bool IsValidHtmlIdCharacter(char c)
...
+    /// white space only, <paramref name="defaultValue"/> is returned.
+    /// </summary>
+    public static string? GetInitials(this string name, string? defaultValue = default)
+    {
+        if (string.IsNullOrWhiteSpace(name))
+        {
+            return defaultValue;
+        }
+
+        var initials = name!.Split(" ", StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)
```
drewnoakes, line 47: I wouldn't expect the suppression to be needed here:

```suggestion
        var initials = name.Split(" ", StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)
```

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -163,3 +164,52 @@ public sealed class TelemetryLimitOptions
...
+    {
+        List<string>? messages = null;
+        if (string.IsNullOrWhiteSpace(NameClaimType))
+        {
+            messages ??= [];
+            messages.Add("OpenID Connect claim type for name not configured. Specify a Dashboard:OpenIdConnect:NameClaimType value.");
+        }
+        else
+        {
+            _nameClaimTypes = NameClaimType.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
tlmii: Well, I don't know how I did that after having to look up the docs for the entries available here that aren't in framework that I'm used to. Good catch.

src/Aspire.Dashboard/Extensions/StringExtensions.cs
```diff
@@ -49,4 +31,29 @@ static bool IsValidHtmlIdCharacter(char c)
             return char.IsLetterOrDigit(c) || c == '_' || c == '-';
         }
     }
+
+    /// <summary>
+    /// Returns the two initial letters of the first and last words in the specified <paramref name="name"/>.
+    /// If only one word is present, a single initial is returned. If <paramref name="name"/> is null, empty or
+    /// white space only, <paramref name="defaultValue"/> is returned.
+    /// </summary>
+    public static string? GetInitials(this string name, string? defaultValue = default)
```
drewnoakes, line 40: I think this can be annotated with:

```suggestion
    [NotNullIfNotNull(nameof(defaultValue))]
    public static string? GetInitials(this string name, string? defaultValue = default)
```

tlmii: Had to specify `[return: NotNullIfNotNull(nameof(defaultValue))]`, but you're right that's good.

src/Aspire.Dashboard/Extensions/ClaimsIdentityExtensions.cs
tlmii: An earlier revision accepted null and I forgot to clean that up when I changed it. Thanks.

src/Aspire.Dashboard/Extensions/StringExtensions.cs
```diff
@@ -49,4 +32,30 @@ static bool IsValidHtmlIdCharacter(char c)
...
         }
     }
+
+    /// <summary>
+    /// Returns the two initial letters of the first and last words in the specified <paramref name="name"/>.
+    /// If only one word is present, a single initial is returned. If <paramref name="name"/> is null, empty or
+    /// white space only, <paramref name="defaultValue"/> is returned.
+    /// </summary>
+    [return: NotNullIfNotNull(nameof(defaultValue))]
+    public static string? GetInitials(this string name, string? defaultValue = default)
```
drewnoakes, line 42: This can be written to only allocate the returned string. This will hardly ever be called, so it's not a big deal, but all the same :)

```c#
[NotNullIfNotNull(nameof(defaultValue))]
public static string? GetInitials(string name, string? defaultValue = default)
{
    ReadOnlySpan<char> s = name.AsSpan().Trim();

    if (s.Length == 0)
        return defaultValue;

    int lastSpaceIndex = s.LastIndexOf(' ');

    if (lastSpaceIndex == -1)
        return s[0].ToString().ToUpperInvariant();

    // The name contained two or more words. Return the initials from the first and last.
    return $"{char.ToUpperInvariant(s[0])}{char.ToUpperInvariant(s[lastSpaceIndex + 1])}";
}
```

src/Aspire.Dashboard/Extensions/StringExtensions.cs
tlmii: I didn't expect it either, but for some reason it was complaining. It also didn't show the `NotNullWhen` annotation on IsNullOrWhiteSpace when I F12'd to it. So I must have gotten into a bad state somehow. Removed it now and it's not complaining and I see the annotation so I guess something changed in the interim.

tests/Aspire.Dashboard.Tests/StringExtensionsTests.cs
```diff
@@ -9,21 +9,27 @@ namespace Aspire.Dashboard.Tests;
...
-    [InlineData("/home/username/src/repos/AspireStarterProject/MyApp/MyApp.csproj", 5, "/h…oj")]
-    [InlineData("/home/username/src/repos/AspireStarterProject/MyApp/MyApp.csproj", 38, "/home/username/src…/MyApp/MyApp.csproj")]
-    [InlineData("/home/username/src/repos/AspireStarterProject/MyApp/MyApp.csproj", 39, "/home/username/src/…/MyApp/MyApp.csproj")]
-    [InlineData("c:\\", 5, "c:\\")]
-    [InlineData("c:\\src\\repos\\AspireStarterProject\\MyApp\\MyApp.csproj", 5, "c:…oj")]
-    [InlineData("c:\\src\\repos\\AspireStarterProject\\MyApp\\MyApp.csproj", 32, "c:\\src\\repos\\As…App\\MyApp.csproj")]
-    [InlineData("c:\\src\\repos\\AspireStarterProject\\MyApp\\MyApp.csproj", 33, "c:\\src\\repos\\Asp…App\\MyApp.csproj")]
-    [InlineData("c:\\src\\repos\\AspireStarterProject\\MyApp\\MyApp.csproj", 48, "c:\\src\\repos\\AspireStar…oject\\MyApp\\MyApp.csproj")]
-    public void TrimMiddle(string path, int targetLength, string expectedResult)
+    [InlineData("", "DefaultValue", "DefaultValue")]
```
drewnoakes, line 12: Can add some more test cases here with spaces in odd places:

```suggestion
    [InlineData("", "DefaultValue", "DefaultValue")]
    [InlineData("   ", "DefaultValue", "DefaultValue")]
    [InlineData("\t", "DefaultValue", "DefaultValue")]
```

src/Aspire.Dashboard/Extensions/StringExtensions.cs
tlmii: Yeah I didn't spend much time trying to make that better, just went with what worked to focus on the UI. But yours works, so I'll take it (slightly modified for analyzer warnings)

Issue #1151, opened 2023-11-30
Automatically persist generated password for all containers with user names and passwords


DamianEdwards:
By default today, we auto-generate a password for the SQL Server container when `AddSqlServerContainer` is called in an AppHost project. A random password on each app start isn't a great experience though as anything created in one session cannot be used in the next if a persistent volume is configured.

We should consider how we might instead automatically manage preservation of the generated password and reuse in subsequent runs, e.g. in user secrets. One challenge of user secrets is it requires initializing the project with an ID (via an MSBuild property in the project file) and modifying the project file while the app is running is probably not a great idea.

mitchdenny:
@DamianEdwards I've got this PR where I am looking to add secret support to Aspire:

https://github.com/dotnet/aspire/pull/1487

This PR is the first phase which adds the underlying resource types, but in the comments I've added a note on how this could evolve to provide some increased management around managing container password secrets (as an example). The difference here is that we don't store a generated password, instead we get a password from configuration (locally) which could be in user secrets.

I think that this is a reasonable pay for play compromise? Even if we did want to go further and do generation and storage we'd probably layer it on top of this generic secrets mechanism rather than building it bespoke for each resource?

mitchdenny:
Punting to preview 4 because we won't have time to react to some of the discussions from our app model offsite and it will dictate what we do here.

mitchdenny:
We discussed this at one of the AppHost syncs. We still want to do this an it's an important building block for container persistence.

davidfowl:
Related to https://github.com/dotnet/aspire/issues/2438

davidfowl:
Putting this in 8.1. It's important for the volume experience.

davidfowl:
Sweet!

PR #3446, opened 2024-04-05
Update TestShop to store RabbitMQ password in user secrets

DamianEdwards:
This updates the TestShop app to automatically store the generated password for RabbitMQ in user secrets so that it runs after cloning without any further action from the developer.

This is a pattern we'll likely want to productize and have the relevant resources support (perhaps by default) as tracked in #1151
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3446)

### DIFF
--- a/playground/TestShop/AppHost/ParameterExtensions.cs
+++ b/playground/TestShop/AppHost/ParameterExtensions.cs
@@ -1,0 +1,117 @@
+using System.Diagnostics;
+using System.Reflection;
+using Aspire.Hosting.Publishing;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.UserSecrets;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Hosting;
+
+internal static class ParameterExtensions
+{
+    /// <summary>
+    /// Creates a password parameter that in the development environment is generated once and stored in the user secrets store.
+    /// </summary>
+    /// <remarks>
+    /// The password is only stable when in the development environment and the application is running. In all other cases, the password is generated each time.
+    /// </remarks>
+    public static IResourceBuilder<ParameterResource> CreateStablePassword(this IDistributedApplicationBuilder builder, string name,
+        bool lower = true, bool upper = true, bool numeric = true, bool special = true,
+        int minLower = 0, int minUpper = 0, int minNumeric = 0, int minSpecial = 0)
+    {
+        ParameterDefault generatedPassword = new GenerateParameterDefault
+        {
+            MinLength = 22, // enough to give 128 bits of entropy when using the default 67 possible characters. See remarks in PasswordGenerator.Generate
+            Lower = lower,
+            Upper = upper,
+            Numeric = numeric,
+            Special = special,
+            MinLower = minLower,
+            MinUpper = minUpper,
+            MinNumeric = minNumeric,
+            MinSpecial = minSpecial
+        };
+
+        if (builder.Environment.IsDevelopment() && builder.ExecutionContext.IsRunMode)
+        {
+            // In development mode, generate a new password each time the application starts
+            generatedPassword = new UserSecretsParameterDefault(builder.Environment.ApplicationName, name, generatedPassword);
+        }
+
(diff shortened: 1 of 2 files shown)


playground/TestShop/AppHost/ParameterExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            var value = parameterDefault.GetDefaultValue();
+            var configurationKey = $"Parameters:{parameterName}";
+            TrySetUserSecret(applicationName, configurationKey, value);
+            return value;
+        }
+
+        public override void WriteToManifest(ManifestPublishingContext context) => parameterDefault.WriteToManifest(context);
+
+        private static bool TrySetUserSecret(string applicationName, string name, string value)
```
mitchdenny, line 69: This is quite different to the way we handle user secrets in the Azure Provisioner (for example) where we load the `secrets.json` file directly and then save it. There are some quirks here in the way that `dotnet user-secrets` injects values into the secrets file. `Parameters:secret` name will be a flatted value at the root of the document.

mitchdenny: That said I'm supportive of doing something like this because it reduces friction when you clone the repo and hit F5.

DamianEdwards: Yeah I wouldn't suggest we productize this approach and was unaware we already had code in the repo to deal with saving user secrets. I ported this from what I'm doing in the samples repo. `dotnet user-secrets` is the only "official" code implementation we have that saves to the secret store today so this seemed like the easiest approach for now in the playground app until we implement this idea for real.

playground/TestShop/AppHost/ParameterExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            Special = special,
+            MinLower = minLower,
+            MinUpper = minUpper,
+            MinNumeric = minNumeric,
+            MinSpecial = minSpecial
+        };
+
+        if (builder.Environment.IsDevelopment() && builder.ExecutionContext.IsRunMode)
+        {
+            // In development mode, generate a new password each time the application starts
```
eerhardt, line 37: I don't think this comment makes sense. We want to get it from user secrets if it is there in development mode.

playground/TestShop/AppHost/ParameterExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Creates a password parameter that in the development environment is generated once and stored in the user secrets store.
+    /// </summary>
+    /// <remarks>
+    /// The password is only stable when in the development environment and the application is running. In all other cases, the password is generated each time.
+    /// </remarks>
+    public static IResourceBuilder<ParameterResource> CreateStablePassword(this IDistributedApplicationBuilder builder, string name,
+        bool lower = true, bool upper = true, bool numeric = true, bool special = true,
+        int minLower = 0, int minUpper = 0, int minNumeric = 0, int minSpecial = 0)
+    {
+        ParameterDefault generatedPassword = new GenerateParameterDefault
```
eerhardt, line 22:
```suggestion
        ParameterDefault parameterDefault = new GenerateParameterDefault
```

Since sometimes this isn't a "generatedPassword" since it could come from user secrets.

playground/TestShop/AppHost/ParameterExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                    catch (Exception) { }
+                }
+            }
+
+            return false;
+        }
+    }
+
+    sealed class ResourceBuilder
```
eerhardt, line 98: Do we have a problem with our Hosting code that you need to make these classes?

playground/TestShop/AppHost/ParameterExtensions.cs
eerhardt, line 69:
```suggestion
        private static void SetUserSecret(string applicationName, string name, string value)
```

Nothing uses this bool.

eerhardt (approved):

PR #3449, opened 2024-04-06
Fix WithReference extension return type

timheuer:
Not caught in PR, but just discovered in use, the `WithReference` changed the shape of the type from `ProjectResource` to `IResourceWithEnvironment` -- not making it compatible with other components.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3449)


eerhardt (approved):
Thanks! I'll cherry pick this into the release/8.0 PR

davidfowl:
This should be a generic method with a constraint IResourceWithEnviroment. See https://github.com/dotnet/aspire/blob/c4ef238b04ccacf5278f884a5e6e047ac954e5cb/src/Aspire.Hosting/ResourceBuilderExtensions.cs#L257

PR #3467, opened 2024-04-07
Make sure bind mount paths are relative to the app host directory

### LINKED ISSUE #3323: WithBindMount methods failing when running in integration tests because paths aren't resolved relative to AppHost project
AppHost projects using the `WithBindMount` method to mount a host directory in a container can fail when running in the context of an integration test because the source path passed to `WithBindMount` isn't being resolved relative to the AppHost project.

Workaround is to explicitly resolve bind mount paths to full paths in the AppHost, e.g.:

```csharp

var grafana = builder.AddContainer("grafana", "grafana/grafana")
                     .WithBindMount(GetFullPath("../grafana/config"), "/etc/grafana", isReadOnly: true)
                     .WithBindMount(GetFullPath("../grafana/dashboards"), "/var/lib/grafana/dashboards", isReadOnly: true)
                     .WithEndpoint(targetPort: 3000, name: "grafana-http", scheme: "http");

builder.AddProject<Projects.MetricsApp>("app")
       .WithEnvironment("GRAFANA_URL", grafana.GetEndpoint("grafana-http"));

builder.AddContainer("prometheus", "prom/prometheus")
       .WithBindMount(GetFullPath("../prometheus"), "/etc/prometheus", isReadOnly: true)
       .WithEndpoint(/* This port is fixed as it's referenced from the Grafana config */ port: 9090, targetPort: 9090);

using var app = builder.Build();

await app.RunAsync();

static string GetFullPath(string relativePath) =>
    Path.GetFullPath(Path.Combine(Projects.MetricsApp_AppHost.ProjectPath, relativePath));
```

@davidfowl mentioned he had a fix for this at one point but that it didn't quite get in. We should fix this for preview.6

davidfowl:
- Made it so that bind mounts annotations are resolved relative to the app host directory by default.
- Bind mounts without a rooted path will throw an error.
- Removed logic from ApplicationExecutor that checks the annotation data since this is checked in the constructor.
- Added tests (moved tests to Containers folder)

Fixes #3323
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3467)


tests/Aspire.Hosting.Tests/Nats/AddNatsTests.cs
```diff
@@ -44,8 +44,10 @@ public void AddNatsContainerWithDefaultsAddsAnnotationMetadata()
     [Fact]
     public void AddNatsContainerAddsAnnotationMetadata()
     {
+        var path = OperatingSystem.IsWindows() ? @"C:\tmp\dev-data" : "/tmp/dev-data";
```
mitchdenny, line 47: Probably shouldn't hard code these paths.

mitchdenny:
Looks pretty good. Just one comment about hard coding paths.

mitchdenny (approved):

tests/Aspire.Hosting.Tests/Nats/AddNatsTests.cs
davidfowl: Why not?

src/Aspire.Hosting/DistributedApplicationOptions.cs
```diff
@@ -39,7 +41,11 @@ public DistributedApplicationOptions()
 
     internal Assembly? Assembly => _assembly.Value;
 
-    internal string? ProjectDirectory => _projectDirectory.Value;
+    internal string? ProjectDirectory
+    {
+        get => _projectDirectory ?? _projectDirectoryLazy.Value;
```
eerhardt, line 46:
```suggestion
        get => _projectDirectory ??= _projectDirectoryLazy.Value;
```

Might as well set this after you evaluate it, right?

Another thought is to remove the field altogether. I don't see it being used in tests. Am I missing something?

davidfowl: Yea the property is being used in tests.

eerhardt: It's confusing to mark the field with the comment

davidfowl: What do you mean? This is only a field so we can set it.

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -136,140 +136,6 @@ public void ExcludeLaunchProfileOmitsBindings()
...
-        var resources = publisher.ManifestDocument.RootElement.GetProperty("resources");
-
-        var grafana = resources.GetProperty("grafana");
-        var bindings = grafana.GetProperty("bindings");
-        var httpBinding = bindings.GetProperty("http");
-        Assert.Equal(3000, httpBinding.GetProperty("targetPort").GetInt32());
-    }
-
-    [Fact]
-    public void EnsureContainerWithArgsEmitsContainerArgs()
```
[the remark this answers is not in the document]
davidfowl: My goal is to move all tests (incrementally) to resource specific folders (projects). This manifest generation test needs to die 😄 (or only be used for testing cross cutting things like writing the entire manifest).

src/Aspire.Hosting/DistributedApplicationOptions.cs
eerhardt: Typically when someone puts a comment like:

```C#
// for testing
internal Foo() {}
```

It means Foo is getting called/used by tests.

A whole other option is to just use a single field, and when it is `null` during `get` generate the value.

eerhardt (approved):

PR #3471, opened 2024-04-08
Enable HSTS middleware when all endpoints are HTTPS

JamesNK:
Adds middleware to add [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) response header to dashboard.

@drewnoakes I noticed there was already a commented out call to `UseHsts` in app startup. You added it when splitting the dashboard out. Do you remember why it is there and commented out?
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3471)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -217,13 +217,13 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         else
         {
             _app.UseExceptionHandler("/Error");
-            //_app.UseHsts();
         }
 
         _app.UseStatusCodePagesWithReExecute("/error/{0}");
 
         if (isAllHttps)
         {
+            _app.UseHsts();
             _app.UseHttpsRedirection();
         }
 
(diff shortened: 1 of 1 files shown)


JamesNK:
@SteveSandersonMS @javiercn @davidfowl Is it a good idea to enable HSTS? The dashboard is designed to run on localhost, and since HSTS applies to the entire localhost domain, it seems like it would be disruptive to a user's local development experience.

See https://weblog.west-wind.com/posts/2022/Oct/24/Fix-automatic-rerouting-of-http-to-https-on-localhost-in-Web-Browsers

JamesNK:
Might not be a problem because `UseHsts` skips localhost by default: https://github.com/dotnet/aspnetcore/blob/6196f76672ed4a4415f7a12e8ae17b8212ebf462/src/Middleware/HttpsPolicy/src/HstsOptions.cs#L40

kvenkatrajan (approved):

PR #3472, opened 2024-04-08
Don't wait until after we've started the entire app to print the token

davidfowl:
- Print it right after we print the dashboard url
- Refactored the dashboard resource to use DashboardOptions instead of DcpOptions

PS: We'll need to backport this to p6.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3472)

### DIFF
--- a/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
@@ -14,7 +14,7 @@ using Microsoft.Extensions.Options;
 namespace Aspire.Hosting.Dashboard;
 
 internal sealed class DashboardLifecycleHook(IConfiguration configuration,
-                                             IOptions<DcpOptions> dcpOptions,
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -147,6 +147,11 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
             {
                 distributedApplicationLogger.LogInformation("Now listening on: {DashboardUrl}", firstDashboardUrl.ToString().TrimEnd('/'));
             }
+
+            if (!string.IsNullOrEmpty(browserToken))
+            {
+                LoggingHelpers.WriteDashboardUrl(distributedApplicationLogger, dashboardUrls, browserToken);
```
davidfowl, line 153: This is the fix.

JamesNK (approved):

PR #3476, opened 2024-04-08
Add max-width to endpoint popup

JamesNK:
Very long endpoint would make the popup excessively large and possibly push it off the screen. This PR adds a max width and trims the endpoints with ellipsis.

Before:

image

After:

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3476)

### DIFF
--- a/src/Aspire.Dashboard/Components/ResourcesGridColumns/EndpointsColumnDisplay.razor
+++ b/src/Aspire.Dashboard/Components/ResourcesGridColumns/EndpointsColumnDisplay.razor
@@ -72,11 +72,11 @@ else
                     Endpoints
                 </Header>
                 <Body>
-                    <div style="max-height:400px; overflow-x:hidden; overflow-y: auto; padding:5px 10px;">
+                    <div class="endpoint-popup">
                         @foreach (var item in items)
                         {
                             var d = (DisplayedEndpoint)item.Data!;
-                            <div style="margin-bottom: 4px;">
+                            <div class="endpoint-link">
                                 @if (d.Url != null)
                                 {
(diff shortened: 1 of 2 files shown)


src/Aspire.Dashboard/Components/ResourcesGridColumns/EndpointsColumnDisplay.razor
```diff
@@ -72,11 +72,11 @@ else
...
-                    <div style="max-height:400px; overflow-x:hidden; overflow-y: auto; padding:5px 10px;">
+                    <div class="endpoint-popup">
                         @foreach (var item in items)
                         {
                             var d = (DisplayedEndpoint)item.Data!;
-                            <div style="margin-bottom: 4px;">
+                            <div class="endpoint-link">
                                 @if (d.Url != null)
                                 {
                                     <a href="@d.Url" target="_blank">@d.Text</a>
```
tlmii, line 82: If it can be truncated now, should we have a tooltip with the full value of `d.Text`?

JamesNK:
Review please 🙏

tlmii:
It feels like the title attribute discussion we were having offline wasn't fully resolved. But it also seems like this makes things better for everyone in more cases while trading one accessibility issue (urls could render larger than the screen size making it hard to see the full value) for another (keyboard users may, in some cases, not be able to see the whole url). I'll approve under that understanding but we should resolve the other discussion separately.

tlmii (approved):

davidfowl (approved):

PR #3492, opened 2024-04-08
Fixed issues around injecting urls

### LINKED ISSUE #2241: Which endpoints should be propagated to ASPNETCORE_URLS env var?
In #2239, I changed the logic to exclude endpoints which either:
* Don't have a "http" or "https" URI scheme
* Have an EnvironmentVariable set

Is this the logic we want to stick with or is there a more robust check we could make?

### LINKED ISSUE #3477: Adding endpoints to a project without a launch profile results in errors in orchestrator
```C#

builder.AddProject<Projects.Stress_ApiService>("stress-apiservice", launchProfileName: null)
       .WithHttpEndpoint(port: 10005, name: "http");

builder.Build().Run();
```

The output

```
fail: Aspire.Hosting.Dcp.dcpctrl.ExecutableReconciler[0]
      could not create Endpoint object  {"Executable": {"name":"stress-apiservice-4slf0l8"}, "Reconciliation": 11, "ServiceName": "stress-apiservice", "Workload": "/stress-apiservice-4slf0l8", "error": "service-producer annotation is invalid: missing information about the port to expose the service"}
fail: Aspire.Hosting.Dcp.dcpctrl.ExecutableReconciler[0]
      could not create Endpoint object  {"Executable": {"name":"stress-apiservice-4slf0l8"}, "Reconciliation": 12, "ServiceName": "stress-apiservice", "Workload": "/stress-apiservice-4slf0l8", "error": "service-producer annotation is invalid: missing information about the port to expose the service"}
```

Seems like there's some strange interaction between DCP when there's no launch profile and adding endpoints.

davidfowl:
ASP.NET Core Urls we only being injected when launch profiles. 

- Fixed issues with ASPNETCORE_URLS not working with launch profiles
- Skip ASPNETCORE_URLS for endpoints that are mapped to an environment
- Added tests

Fixes #2241
Fixes #3477
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3492)

### DIFF
--- a/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
@@ -98,6 +98,7 @@ public static class ProjectResourceBuilderExtensions
 
         builder.WithOtlpExporter();
         builder.ConfigureConsoleLogs();
+        builder.SetAspNetCoreUrls();
 
         var projectResource = builder.Resource;
 
@@ -157,41 +158,6 @@ public static class ProjectResourceBuilderExtensions
                     var value = Environment.ExpandEnvironmentVariables(envVar.Value);
                     context.EnvironmentVariables.TryAdd(envVar.Key, value);
                 }
-
-                if (context.EnvironmentVariables.ContainsKey("ASPNETCORE_URLS"))
-                {
-                    // If the user has already set ASPNETCORE_URLS, we don't want to override it.
-                    return;
-                }
-
-                var aspnetCoreUrls = new ReferenceExpressionBuilder();
-
-                var processedHttpsPort = false;
-                var first = true;
-
-                // Turn http and https endpoints into a single ASPNETCORE_URLS environment variable.
-                foreach (var e in builder.Resource.GetEndpoints().Where(e => e.EndpointAnnotation.UriScheme is "http" or "https"))
-                {
-                    if (!first)
-                    {
-                        aspnetCoreUrls.AppendLiteral(";");
-                    }
-
-                    if (!processedHttpsPort && e.EndpointAnnotation.UriScheme == "https")
-                    {
-                        // Add the environment variable for the HTTPS port if we have an HTTPS service. This will make sure the
-                        // HTTPS redirection middleware avoids redirecting to the internal port.
-                        context.EnvironmentVariables["ASPNETCORE_HTTPS_PORT"] = e.Property(EndpointProperty.Port);
-
-                        processedHttpsPort = true;
-                    }
-
-                    aspnetCoreUrls.Append($"{e.Property(EndpointProperty.Scheme)}://localhost:{e.Property(EndpointProperty.TargetPort)}");
-                    first = false;
-                }
-
-                // Combine into a single expression
-                context.EnvironmentVariables["ASPNETCORE_URLS"] = aspnetCoreUrls.Build();
             });
 
             // TODO: Process command line arguments here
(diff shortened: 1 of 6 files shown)


karolz-ms (approved):

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -287,4 +253,53 @@ private static bool IsWebProject(ProjectResource projectResource)
...
+                return;
+            }
+
+            var aspnetCoreUrls = new ReferenceExpressionBuilder();
+
+            var processedHttpsPort = false;
+            var first = true;
+
+            static bool IsValidAspNetCoreUrl(EndpointAnnotation e) =>
+                e.UriScheme is "http" or "https" && e.TargetPortEnvironmentVariable is null;
```
eerhardt, line 278: Why does having a `TargetPortEnvironmentVariable` mean it isn't a valid ASPNETCORE_URLS?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -421,6 +421,8 @@ private static void ApplyEndpoints<T>(this IResourceBuilder<T> builder, IResourc
         // Set the environment variable on the resource
         if (env is not null && builder.Resource is IResourceWithEndpoints resourceWithEndpoints and IResourceWithEnvironment)
         {
+            annotation.TargetPortEnvironmentVariable = env;
```
eerhardt, line 424: It's kind of weird that `WithEndpoint(..., env: "SOME_ENV")` means that `SOME_ENV` gets the "target port" value.

1. Is this assuming too much this deep?
2. Should `env` be renamed?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: It's super indirect. I didn't want to design a new API explicitly opting you into ASP.NET Core URLs, but that's another option.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
davidfowl: Are you concerned about the name of the variable or the behavior? I tried to remove this feature completely, but it makes executables listening on a port more verbose than it needs to be.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
eerhardt: One thing I think may help is to update the docs on WithEndpoint's "env" variable:

> /// <param name="env">An optional name of the environment variable to inject.</param>

It doesn't really describe what is happening here.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
eerhardt: > Are you concerned about the name of the variable or the behavior?

A little about the name of the variable. A little about the documentation on what it does. I think I understand the behavior.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -287,4 +253,53 @@ private static bool IsWebProject(ProjectResource projectResource)
...
+
+                    processedHttpsPort = true;
+                }
+
+                aspnetCoreUrls.Append($"{e.Property(EndpointProperty.Scheme)}://localhost:{e.Property(EndpointProperty.TargetPort)}");
+                first = false;
+            }
+
+            // Combine into a single expression
+            context.EnvironmentVariables["ASPNETCORE_URLS"] = aspnetCoreUrls.Build();
```
ReubenBond, line 302: If this is empty, is it better to set it or not set it?

ReubenBond (approved):

src/Aspire.Hosting/ResourceBuilderExtensions.cs
davidfowl: Updated

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: Fixed

eerhardt (approved):

JamesNK:
Is this the bug I hit yesterday?

davidfowl:
Yep!

JamesNK (approved):
I don't really know this code but it looks ok to me

davidfowl:
@ReubenBond can you back port your orleans change. This needs to go in afterwards.

PR #3502, opened 2024-04-08
update dependencies with minor/patch increments available

danmoseley:
With dependabot not working, I used the marvelous
https://github.com/dotnet-outdated/dotnet-outdated
then hand-edited a few lines as it can't handle updating where the original property was set.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3502)


danmoseley:
I don't know whether any of these increments add any value, btw.

danmoseley:
doh -- I didn't add nuget.org to the feeds before doing this, so it only saw packages that were internal or already mirrored. we can merge this as is, then I can do that again -- they will need me mirroring the packages

joperezr (approved):

Issue #3491, opened 2024-04-08
Api Review for Aspire Components -- Not super deep dive, overview.


joperezr:

eerhardt:
@sebastienros - is this work now complete?

PR #3503, opened 2024-04-08
API Review - Components

sebastienros:
#3491
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3503)

### DIFF
--- a/API/NoDocs/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
+++ b/API/NoDocs/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
@@ -1,0 +1,17 @@
+// Assembly 'Aspire.Azure.AI.OpenAI'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.AI.OpenAI;
+
+public sealed class AzureOpenAISettings : IConnectionStringSettings
+{
+    public Uri? Endpoint { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public string? Key { get; set; }
+    public bool Tracing { get; set; }
+    public AzureOpenAISettings();
+}
--- a/API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
+++ b/API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
@@ -1,0 +1,17 @@
+// Assembly 'Aspire.Azure.Messaging.EventHubs'
+
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Messaging.EventHubs;
+
+public abstract class AzureMessagingEventHubsBaseSettings : IConnectionStringSettings
+{
+    public string? ConnectionString { get; set; }
+    public string? Namespace { get; set; }
+    public string? EventHubName { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public bool Tracing { get; set; }
+    protected AzureMessagingEventHubsBaseSettings();
+}
--- a/API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsConsumerSettings.cs
+++ b/API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsConsumerSettings.cs
@@ -1,0 +1,8 @@
+// Assembly 'Aspire.Azure.Messaging.EventHubs'
+
+namespace Aspire.Azure.Messaging.EventHubs;
+
+public sealed class AzureMessagingEventHubsConsumerSettings : AzureMessagingEventHubsConsumerBaseSettings
+{
+    public AzureMessagingEventHubsConsumerSettings();
+}
--- a/API/NoDocs/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
+++ b/API/NoDocs/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
@@ -1,0 +1,18 @@
+// Assembly 'Aspire.Azure.Messaging.ServiceBus'
+
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Messaging.ServiceBus;
+
+public sealed class AzureMessagingServiceBusSettings : IConnectionStringSettings
+{
+    public string? ConnectionString { get; set; }
+    public string? Namespace { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public string? HealthCheckQueueName { get; set; }
+    public string? HealthCheckTopicName { get; set; }
+    public bool Tracing { get; set; }
+    public AzureMessagingServiceBusSettings();
+}
--- a/API/NoDocs/Aspire.Azure.Search.Documents/AzureSearchSettings.cs
+++ b/API/NoDocs/Aspire.Azure.Search.Documents/AzureSearchSettings.cs
@@ -1,0 +1,18 @@
+// Assembly 'Aspire.Azure.Search.Documents'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Search.Documents;
+
+public sealed class AzureSearchSettings : IConnectionStringSettings
+{
+    public Uri? Endpoint { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public string? Key { get; set; }
+    public bool HealthChecks { get; set; }
+    public bool Tracing { get; set; }
+    public AzureSearchSettings();
+}
--- a/API/NoDocs/Aspire.Azure.Security.KeyVault/AzureSecurityKeyVaultSettings.cs
+++ b/API/NoDocs/Aspire.Azure.Security.KeyVault/AzureSecurityKeyVaultSettings.cs
@@ -1,0 +1,17 @@
+// Assembly 'Aspire.Azure.Security.KeyVault'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Security.KeyVault;
+
+public sealed class AzureSecurityKeyVaultSettings : IConnectionStringSettings
+{
+    public Uri? VaultUri { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public bool HealthChecks { get; set; }
+    public bool Tracing { get; set; }
+    public AzureSecurityKeyVaultSettings();
+}
--- a/API/NoDocs/Aspire.Microsoft.Azure.Cosmos/AzureCosmosDBSettings.cs
+++ b/API/NoDocs/Aspire.Microsoft.Azure.Cosmos/AzureCosmosDBSettings.cs
@@ -1,0 +1,16 @@
+// Assembly 'Aspire.Microsoft.Azure.Cosmos'
+
+using System;
+using System.Runtime.CompilerServices;
+using Azure.Core;
+
+namespace Aspire.Microsoft.Azure.Cosmos;
+
+public sealed class AzureCosmosDBSettings
+{
+    public string? ConnectionString { get; set; }
+    public bool Tracing { get; set; }
+    public Uri? AccountEndpoint { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public AzureCosmosDBSettings();
+}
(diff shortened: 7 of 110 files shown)


danmoseley:
Just wondering, how does this new API folder work? Is there a build step that makes sure it stays up to date? I don't see one like this in aspnetcore, wondering how it works

danmoseley:
Or is this not to actually check in just for review

sebastienros:
@danmoseley just to comment on the API, this won't get merged. We did it the same way for dotnet/extensions.

API/NoDocs/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
```diff
@@ -0,0 +1,17 @@
+// Assembly 'Aspire.Azure.AI.OpenAI'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.AI.OpenAI;
+
+public sealed class AzureOpenAISettings : IConnectionStringSettings
```
tg-msft, line 10: Nit - Worth calling out that this pattern varies from the default of `<Namespace>Settings`.  I think this reads better than `AzureAIOpenAISettings` though and we should keep it.

API/NoDocs/Aspire.Azure.Search.Documents/AzureSearchSettings.cs
```diff
@@ -0,0 +1,18 @@
+// Assembly 'Aspire.Azure.Search.Documents'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Search.Documents;
+
+public sealed class AzureSearchSettings : IConnectionStringSettings
```
tg-msft, line 10: Nit - Worth calling out that this pattern varies from the default of `<Namespace>Settings`.  I think this reads better than `AzureSearchDocumentsSettings` though and we should keep it.

API/NoDocs/Aspire.Microsoft.Azure.Cosmos/AzureCosmosDBSettings.cs
```diff
@@ -0,0 +1,16 @@
+// Assembly 'Aspire.Microsoft.Azure.Cosmos'
+
+using System;
+using System.Runtime.CompilerServices;
+using Azure.Core;
+
+namespace Aspire.Microsoft.Azure.Cosmos;
+
+public sealed class AzureCosmosDBSettings
```
tg-msft, line 9: We should _consider_ including `Microsoft` in this name.  We're still using the Track 1 Cosmos client which starts with `Microsoft.*` and while we could still use `AzureCosmosSettings` for a future Track 2 client it might be a little confusing to have both of these.  Ditto on the extension method to add a client.

API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
```diff
@@ -0,0 +1,17 @@
+// Assembly 'Aspire.Azure.Messaging.EventHubs'
+
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Messaging.EventHubs;
+
+public abstract class AzureMessagingEventHubsBaseSettings : IConnectionStringSettings
```
tg-msft, line 9: Folks on my team are pretty allergic to the word `Base` and we're not using it in the EH client libraries today (see [Azure.Messaging.EventHubs](https://apiview.dev/Assemblies/Review/050a8e78e2e94b92b38ec4920307efda) and [Azure.Messaging.EventHubs.Processor](https://apiview.dev/Assemblies/Review/7c00b3c911484f67a433d30f8eb6fb06)).  Are we adding this base class because we think it'll help customers or because we're trying to make our implementation slightly easier?  I struggle to imagine people wanting to polymorph over an input only settings class.

API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsConsumerSettings.cs
```diff
@@ -0,0 +1,8 @@
+// Assembly 'Aspire.Azure.Messaging.EventHubs'
+
+namespace Aspire.Azure.Messaging.EventHubs;
+
+public sealed class AzureMessagingEventHubsConsumerSettings : AzureMessagingEventHubsConsumerBaseSettings
```
tg-msft, line 5: Did we talk with any of the EH folks about whether to use `EventHub` or `EventHubs`?  We use plural `EventHubs` in the package/namespace but it's singular in `EventHubConsumerClient`.

API/NoDocs/Aspire.Azure.Security.KeyVault/AzureSecurityKeyVaultSettings.cs
```diff
@@ -0,0 +1,17 @@
+// Assembly 'Aspire.Azure.Security.KeyVault'
+
+using System;
+using System.Runtime.CompilerServices;
+using Aspire.Azure.Common;
+using Azure.Core;
+
+namespace Aspire.Azure.Security.KeyVault;
+
+public sealed class AzureSecurityKeyVaultSettings : IConnectionStringSettings
```
tg-msft, line 10: Just to triple check one last time before GA - we're going to unify all the KV packages for Keys, Secrets, Certs, etc., into a single Aspire component?

API/NoDocs/Microsoft.Extensions.Hosting/AspireAzureEFCoreCosmosDBExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Aspire.Microsoft.EntityFrameworkCore.Cosmos;
+using Microsoft.EntityFrameworkCore;
+
+namespace Microsoft.Extensions.Hosting;
+
+public static class AspireAzureEFCoreCosmosDBExtensions
+{
+    public static void AddCosmosDbContext<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicProperties)] TContext>(this IHostApplicationBuilder builder, string connectionName, string databaseName, Action<EntityFrameworkCoreCosmosDBSettings>? configureSettings = null, Action<DbContextOptionsBuilder>? configureDbContextOptions = null) where TContext : DbContext;
+    public static void EnrichCosmosDbContext<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicProperties)] TContext>(this IHostApplicationBuilder builder, Action<EntityFrameworkCoreCosmosDBSettings>? configureSettings = null) where TContext : DbContext;
```
amcasey, line 13: We have somewhat analogous methods called ConfigureX, but they're also tied to the options pattern.  Probably not enough to cause confusion though?

amcasey: What does it do if you don't pass the action?  It reads from config.

amcasey: Seems like it's implicitly accepting the root `IConfiguration` as an argument.  If we made it explicit, it would be more obvious that it's reading from config and give the option of passing a different config.

amcasey: Alt: Augment.

API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public string? ConnectionString { get; set; }
+    public string? Namespace { get; set; }
+    public string? EventHubName { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public bool Tracing { get; set; }
+    protected AzureMessagingEventHubsBaseSettings();
```
amcasey, line 16:
```suggestion
    internal AzureMessagingEventHubsBaseSettings();
```

API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
amcasey: Inlining the properties seems fine.

API/NoDocs/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Aspire.Azure.Messaging.ServiceBus;
+
+public sealed class AzureMessagingServiceBusSettings : IConnectionStringSettings
+{
+    public string? ConnectionString { get; set; }
+    public string? Namespace { get; set; }
+    public TokenCredential? Credential { get; set; }
+    public string? HealthCheckQueueName { get; set; }
+    public string? HealthCheckTopicName { get; set; }
+    public bool Tracing { get; set; }
```
amcasey, line 16:
```suggestion
    public bool TracingEnabled { get; set; }
```

amcasey (commented):
Some notes from the API Review discussion.

API/NoDocs/Aspire.StackExchange.Redis/StackExchangeRedisSettings.cs
```diff
@@ -0,0 +1,13 @@
+// Assembly 'Aspire.StackExchange.Redis'
+
+using System.Runtime.CompilerServices;
+
+namespace Aspire.StackExchange.Redis;
+
+public sealed class StackExchangeRedisSettings
```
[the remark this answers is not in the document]
eerhardt: @sebastienros was 100% correct in the meeting. These `Settings` classes can't use the `IOptions` pattern because the `Settings` objects influence what gets put into the DI container. `Settings.Tracing = false` means "don't put the Tracing services into DI". Since these Settings objects need to be evaluated before the DI service provider is built, they can't be `IOptions`, which is why we choose the name `Settings`.

See:

https://github.com/dotnet/aspire/blob/e2ba459e1505c82219efef201d086481c1f59df2/src/Components/README.md?plain=1#L38-L40

API/NoDocs/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsBaseSettings.cs
eerhardt: A couple thoughts:

1. My current plan is to keep EventHubs in "pre-release" even when the rest of .NET Aspire ships stable `8.0.0`. So we can tweak this API later. This component came in pretty late in preview5 and follows a bit of a different pattern than the others. So I think we need a little more bake time.
2. We use the base class internally:

https://github.com/dotnet/aspire/blob/e2ba459e1505c82219efef201d086481c1f59df2/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs#L16-L31

inlining the properties will just mean some more code duplication internally, but that isn't a huge deal.

> Are we adding this base class because we think it'll help customers or because we're trying to make our implementation slightly easier?

It's mostly the "don't repeat yourself" and making our implementation slightly easier.

> allergic to the word Base

Agreed we can drop `Base`.

____

In the end, I think we could make it work to not have base classes here, but it feels less "Object Oriented" to say "let's not abstract commonalitites in these classes".  Note that consumers are never expected to create these objects, they get passed the object in the `configureSettings` callback.

API/NoDocs/Aspire.Microsoft.EntityFrameworkCore.SqlServer/MicrosoftEntityFrameworkCoreSqlServerSettings.cs
```diff
@@ -0,0 +1,15 @@
+// Assembly 'Aspire.Microsoft.EntityFrameworkCore.SqlServer'
+
+using System.Runtime.CompilerServices;
+
+namespace Aspire.Microsoft.EntityFrameworkCore.SqlServer;
+
+public sealed class MicrosoftEntityFrameworkCoreSqlServerSettings
+{
+    public string? ConnectionString { get; set; }
+    public bool Retry { get; set; }
```
sebastienros, line 10: DisableX default to `true`

sebastienros: [Obsolete] current properties

API/NoDocs/Aspire.Microsoft.Azure.Cosmos/AzureCosmosDBSettings.cs
sebastienros: Add Microsoft

PR #3512, opened 2024-04-09
Update versions to latest found on nuget.org

danmoseley:
We can decide which if any we want to port into 8.0

procedure used
1. build tree successfully using regular feeds to populate local cache with stuff currently in our feeds
2. replace nuget config to use only nuget.org feed `dotnet new nugetconfig --force`
3. `dotnet outdated -vl Major -u Aspire.sln` (installed from https://github.com/dotnet-outdated/dotnet-outdated?tab=readme-ov-file#installation)
4. fixup where properties were erased in the props file
5. revert nuget config
6. build again
7. where packages can't be found, kick off mirroring https://github.com/dotnet/arcade/blob/main/Documentation/MirroringPackages.md
(hand choose latest stable package if it's not the newest)
8. assuming mirroring is finished, build again and verify it works

But now I'm getting odd local build results so let's try here...
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3512)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -91,15 +91,9 @@
     <PackageVersion Include="Confluent.Kafka" Version="2.3.0" />
     <PackageVersion Include="Dapper" Version="2.1.35" />
     <PackageVersion Include="DnsClient" Version="1.7.0" />
-<<<<<<< HEAD
-    <PackageVersion Include="Google.Protobuf" Version="3.26.0" />
-    <PackageVersion Include="Grpc.AspNetCore" Version="2.60.0" />
-    <PackageVersion Include="Grpc.Net.ClientFactory" Version="2.60.0" />
-=======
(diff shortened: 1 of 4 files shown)


danmoseley:
Going forward we will need to script this (dependabot working may help) ideally including the mirroring due to the diverse collection of dependencies we have that actively update.. and this list will only get longer.

danmoseley:
actually it's building locally now I fixed my typo. ready for review.

danmoseley:
@normj what AWS versions should we reference for final release? this will be in May, we're aiming to have essentially all changes done this week but can take change later if it's important. 
Is this set OK? or previous set?

RussKie (approved):

eerhardt (approved):

danmoseley:
... and meanwhile dependabot staggers back to life and opens up PR's for some of these ...

PR #3514, opened 2024-04-09
Change starter template to make frontend external endpoints

### LINKED ISSUE #3495: Consider changing templates to use WithExternalHttpEndpoints
With the changes, out of the box a deployed Aspire starter app to Azure (or other place) is not marking the front-end accessible.

Consider adding `WithExternalHttpEndpoints` in the templates for the endpoints that need it (front-end, Aspire Starter)

timheuer:
Fixes #3495 to make the frontend web app in aspire-starter template to use the `WithExternalHttpEndpoints` 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3514)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
@@ -10,6 +10,7 @@ builder.AddProject<Projects.AspireStarterApplication__1_Web>("webfrontend")
 #if UseRedisCache
     .WithReference(cache)
 #endif
-    .WithReference(apiService);
+    .WithReference(apiService)
+    .WithExternalHttpEndpoints();
 
 builder.Build().Run();
--- a/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
@@ -7,10 +7,10 @@ var cache = builder.AddRedis("cache");
 var apiService = builder.AddProject<Projects.AspireStarterApplication__1_ApiService>("apiservice");
 
 builder.AddProject<Projects.AspireStarterApplication__1_Web>("webfrontend")
+    .WithExternalHttpEndpoints()
 #if UseRedisCache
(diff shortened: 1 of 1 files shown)


src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
```diff
@@ -10,6 +10,7 @@
 #if UseRedisCache
     .WithReference(cache)
 #endif
-    .WithReference(apiService);
+    .WithReference(apiService)
+    .WithExternalHttpEndpoints();
```
davidfowl, line 14: I would add it first not last.

timheuer: Curious why? is there some race?

davidfowl: Not a race but this exposes all http/https endpoints defined when called. That's the more correct behavior for our default experience. Other http endpoints won't get exposed externally, which would blow up in ACA

timheuer: Will make the change, but not understanding how it affects the WithReference calls in this case.

davidfowl: It doesn't. It's about future calls to `WithHttpEndpoint`. If they get made before `WithExternalHttpEndpoints` it'll likely explode on ACA.

davidfowl (approved):

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.AppHost/Program.cs
timheuer: Sounds like another good case for an analyzer ;-)

eerhardt (approved):

timheuer:
Assume we want to merge/backport this?

davidfowl:
yes

timheuer:
> yes

make it so all powerful one. 😜

Issue #3429, opened 2024-04-05
AppHost should write Dashboard logs to console when it fails


eerhardt:
When the Dashboard fails to launch, it is not easy to debug why. You get a "normal" console log output and things look like they are running. But when you try to hit the Dashboard it refuses the connection (because the process isn't running anymore).

We need to write the logs from the Dashboard to the console in this case. Users will have no idea how to figure out what happened. The only way I know today is to dig into DCP and ask it for the logs.

An easy way to get into this situation today is to install the `8.0-preview5` of the workload. And then use a recent `8.1` build of Aspire.Hosting in your AppHost. You can do this by just lifting up an Aspire.Hosting package to 8.1.

```xml
  <ItemGroup>
    <PackageReference Include="Aspire.Hosting.AppHost" Version="8.0.0-preview.5.24201.12" />
    <PackageReference Include="Aspire.Hosting.Qdrant" Version="8.1.0-preview.1.24204.10" />
  </ItemGroup>
```

Running the AppHost gives me:

```
info: Aspire.Hosting.DistributedApplication[0]
      Aspire version: 8.1.0-preview.1.24204.10+45df94ff325848fefe58a7668817c4c7b3981b4e
info: Aspire.Hosting.DistributedApplication[0]
      Distributed application starting.
info: Aspire.Hosting.DistributedApplication[0]
      Application host directory is: C:\Users\eerhardt\source\repos\AspireApp13\AspireApp13.AppHost
info: Aspire.Hosting.DistributedApplication[0]
      Now listening on: https://localhost:17150
info: Aspire.Hosting.DistributedApplication[0]
      Login to the dashboard at https://localhost:17150/login?t=2284270edba1ece85eb9b043fc27cb2f
info: Aspire.Hosting.DistributedApplication[0]
      Distributed application started. Press Ctrl+C to shut down.
```

but when I hit the Dashboard it fails:
![image](https://github.com/dotnet/aspire/assets/8291187/2ad37a3a-9b53-484f-8e6e-a8b3617ff6ea)

And then digging into the DCP logs of the Dashboard gives me:

```
Unhandled exception. System.InvalidOperationException: Failed to convert configuration value at 'Dashboard:Frontend:AuthMode' to type 'Aspire.Dashboard.Configuration.FrontendAuthMode'.
 ---> System.FormatException: BrowserToken is not a valid value for FrontendAuthMode.
 ---> System.ArgumentException: Requested value 'BrowserToken' was not found.
[...]

PR #3523, opened 2024-04-09
Flow dashboard logs through the app host's logger

### LINKED ISSUE #2911: Option to write dashboard console logs in the host console
When users run an Aspire app, they run the host. Output from the host is written to the console. If there is an error in the host, such as from the resource service, these errors are written to the console.

The dashboard is started with the host, but it runs out of process. The dashboard is launched by the DCP. Its logs aren't written to a console the user can see. If an error occurs in the dashboard, then there is no way to access the details. Errors reported to us won't have any logging information.

There should be a host config option to output dashboard logs to the console.

davidfowl:
- Subscribe to logs from the dashboard resources and print them to the console.

Fixes https://github.com/dotnet/aspire/issues/2911

This pull request primarily focuses on enhancing the `DashboardLifecycleHook` class in the `src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs` file. The changes include the addition of new dependencies, the implementation of the `IAsyncDisposable` interface, and the introduction of a method to watch and handle dashboard logs. Additionally, there are modifications to the `DistributedApplicationBuilder` class and updates to the unit tests to accommodate these changes.

Key changes include:

**Addition of new dependencies:**

* [`src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs`](diffhunk://#diff-dffe4637b4710f58a20faab4d5a5e45a64e6d6fc71443eb3136286afa68b3c88L20-R30): New dependencies were added to the `DashboardLifecycleHook` class, including `ResourceNotificationService`, `ResourceLoggerService`, and `ILoggerFactory`. This change allows the class to handle resource notifications and logs, and to create loggers.

**Implementation of the `IAsyncDisposable` interface:**

* [`src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs`](diffhunk://#diff-dffe4637b4710f58a20faab4d5a5e45a64e6d6fc71443eb3136286afa68b3c88R48-R71): The `DashboardLifecycleHook` class now implements the `IAsyncDisposable` interface, which provides a mechanism for releasing unmanaged resources asynchronously. A new `DisposeAsync` method was added to handle the disposal of resources.

**Introduction of a method to watch and handle dashboard logs:**

* [`src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs`](diffhunk://#diff-dffe4637b4710f58a20faab4d5a5e45a64e6d6fc71443eb3136286afa68b3c88R192-R298): A new method, `WatchDashboardLogsAsync`, was introduced to watch and handle dashboard logs. This method includes logic to parse and log JSON formatted log lines.

**Updates to the `DistributedApplicationBuilder` class:**

* [`src/Aspire.Hosting/DistributedApplicationBuilder.cs`](diffhunk://#diff-61cfd4a19122460bb61a3ea9ba0fd6f4f23a5e5de3e6c8781cd3164eab88dab1R79): An additional filter was added to the `DistributedApplicationBuilder` class to handle logging for the `Aspire.Hosting.Dashboard`.

**Updates to unit tests:**

* [`tests/Aspire.Hosting.Tests/Dashboard/DashboardResourceTests.cs`](diffhunk://#diff-5b1712292a80a3a815ccb6688a54d9bfd3d540d47028c2d2833b09fe6991c59bR4-R12): Unit tests were updated to accommodate the changes in the `DashboardLifecycleHook` class and the `DistributedApplicationBuilder` class. This includes the addition of a new test, `DashboardLifecycleHookWatchesWarningLogs`, to verify the new log watching functionality. [[1]](diffhunk://#diff-5b1712292a80a3a815ccb6688a54d9bfd3d540d47028c2d2833b09fe6991c59bR4-R12) [[2]](diffhunk://#diff-5b1712292a80a3a815ccb6688a54d9bfd3d540d47028c2d2833b09fe6991c59bR129-R133) [[3]](diffhunk://#diff-5b1712292a80a3a815ccb6688a54d9bfd3d540d47028c2d2833b09fe6991c59bL286-R403)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3523)

### DIFF
--- a/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
@@ -2,6 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics;
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using System.Text.Json.Serialization;
 using Aspire.Dashboard.Model;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Dcp;
@@ -17,8 +20,14 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
                                              IOptions<DashboardOptions> dashboardOptions,
                                              ILogger<DistributedApplication> distributedApplicationLogger,
                                              IDashboardEndpointProvider dashboardEndpointProvider,
-                                             DistributedApplicationExecutionContext executionContext) : IDistributedApplicationLifecycleHook
+                                             DistributedApplicationExecutionContext executionContext,
+                                             ResourceNotificationService resourceNotificationService,
+                                             ResourceLoggerService resourceLoggerService,
+                                             ILoggerFactory loggerFactory) : IDistributedApplicationLifecycleHook, IAsyncDisposable
 {
+    private readonly CancellationTokenSource _shutdownCts = new();
+    private Task? _dashboardLogsTask;
+
     public Task BeforeStartAsync(DistributedApplicationModel model, CancellationToken cancellationToken)
     {
         Debug.Assert(executionContext.IsRunMode, "Dashboard resource should only be added in run mode");
@@ -36,7 +45,30 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
             AddDashboardResource(model);
         }
 
+        _dashboardLogsTask = WatchDashboardLogsAsync();
+
         return Task.CompletedTask;
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        _shutdownCts.Cancel();
+
+        if (_dashboardLogsTask is not null)
+        {
+            try
+            {
+                await _dashboardLogsTask.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                // Expected when the application is shutting down.
+            }
+            catch (Exception ex)
+            {
+                distributedApplicationLogger.LogError(ex, "Unexpected error while watching dashboard logs.");
+            }
+        }
     }
 
     private void AddDashboardResource(DistributedApplicationModel model)
@@ -90,7 +122,8 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
             {
                 ExecutableResource => KnownResourceTypes.Executable,
                 ProjectResource => KnownResourceTypes.Project,
-                _ => KnownResourceTypes.Container
+                ContainerResource => KnownResourceTypes.Container,
+                _ => dashboardResource.GetType().Name
             },
             State = configuration.GetBool("DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES") is true ? null : KnownResourceStates.Hidden
         };
@@ -141,6 +174,10 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
                 context.EnvironmentVariables[DashboardConfigNames.DashboardOtlpAuthModeName.EnvVarName] = "Unsecured";
             }
 
+            // Change the dashboard formatter to use JSON so we can parse the logs and render them in the
+            // via the ILogger.
+            context.EnvironmentVariables["LOGGING__CONSOLE__FORMATTERNAME"] = "json";
+
             // We need to print out the url so that dotnet watch can launch the dashboard
             // technically this is too early, but it's late ne
             if (StringUtils.TryGetUriFromDelimitedString(dashboardUrls, ";", out var firstDashboardUrl))
@@ -154,4 +191,136 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
             }
         }));
     }
+
+    private async Task WatchDashboardLogsAsync()
+    {
+        async Task<string?> GetDashboardResourceIdentifier()
+        {
+            string? dashboardResourceId = null;
+
+            try
+            {
+                var timeout = Debugger.IsAttached ? Timeout.InfiniteTimeSpan : dashboardOptions.Value.DashboardStartupTimeout;
+
+                // The dashboard resource isn't immediately available so watch for it.
+                // Wait for dashboard to startup and be reported before giving up.
+                using var cts =  new CancellationTokenSource(timeout);
+                using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, _shutdownCts.Token);
+
+                await foreach (var notification in resourceNotificationService.WatchAsync(linkedCts.Token))
+                {
+                    // We want to capture the replica of the dashboard resource. This will allow us to get the logs for the dashboard.
+                    // When the resource id is not the same as the resource name, it's the replica.
+                    if (StringComparers.ResourceName.Equals(notification.Resource.Name, KnownResourceNames.AspireDashboard) &&
+                        notification.ResourceId != notification.Resource.Name)
+                    {
+                        dashboardResourceId = notification.ResourceId;
+                        break;
+                    }
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                if (distributedApplicationLogger.IsEnabled(LogLevel.Debug))
+                {
+                    distributedApplicationLogger.LogDebug("Timed out waiting for dashboard resource to start.");
+                }
+            }
+
+            return dashboardResourceId;
+        }
+
+        if (await GetDashboardResourceIdentifier().ConfigureAwait(false) is not string dashboardResourceId)
+        {
+            // The only way this can happen is if DCP is unhealthy, so log a warning.
+            distributedApplicationLogger.LogWarning("Failed to get the dashboard logs.");
+            return;
+        }
+
+        var loggerCache = new Dictionary<string, ILogger>(StringComparer.Ordinal);
+        var defaultDashboardLogger = loggerFactory.CreateLogger("Aspire.Hosting.Dashboard");
+
+        try
+        {
+            // We turned on the JSON formatter for the logger, so we can log the log lines as JSON.
+            await foreach (var batch in resourceLoggerService.WatchAsync(dashboardResourceId).WithCancellation(_shutdownCts.Token).ConfigureAwait(false))
+            {
+                foreach (var logLine in batch)
+                {
+                    DashboardLogMessage? logMessage = null;
+
+                    try
+                    {
+                        logMessage = JsonSerializer.Deserialize(logLine.Content, DashboardLogMessageContext.Default.DashboardLogMessage);
+                    }
+                    catch (JsonException)
+                    {
+                        if (defaultDashboardLogger.IsEnabled(LogLevel.Debug))
+                        {
+                            defaultDashboardLogger.LogDebug("Failed to parse dashboard log line as JSON: {LogLine}", logLine.Content);
+                        }
+
+                        // Failed to parse, it's not JSON, just log the content as is.
+                        var level = logLine.IsErrorMessage ? LogLevel.Error : LogLevel.Information;
+                        defaultDashboardLogger.Log(level, 0, logLine.Content, null, (s, _) => s);
+                    }
+
+                    if (logMessage is not null)
+                    {
+                        LogMessage(loggerFactory, loggerCache, logMessage);
+                    }
+                }
+            }
+        }
+        catch (OperationCanceledException)
+        {
+            // Expected when the application is shutting down.
+        }
+        catch (Exception ex)
+        {
+            defaultDashboardLogger.LogError(ex, "Error reading dashboard logs.");
+        }
+        finally
+        {
+            if (defaultDashboardLogger.IsEnabled(LogLevel.Debug))
+            {
+                defaultDashboardLogger.LogDebug("Stopped reading dashboard logs.");
+            }
+        }
+    }
+
+    private static void LogMessage(ILoggerFactory loggerFactory, Dictionary<string, ILogger> loggerCache, DashboardLogMessage logMessage)
+    {
+        if (!loggerCache.TryGetValue(logMessage.Category, out var logger))
+        {
+            // Looks strange to see Aspire.Hosting.Dashboard.Aspire.Dashboard.Category,
+            // so trim the prefix and append Aspire.Hosting.Why is this important?
+            // Well there are logs emitting from categories that don't start with Aspire.Dashboard so we want to prefix all logs so that they can be controlled by config.
+            var categoryTrimmed = logMessage.Category.StartsWith("Aspire.Dashboard.") ?
+                logMessage.Category["Aspire.Dashboard.".Length..] : logMessage.Category;
+
+            loggerCache[logMessage.Category] = logger = loggerFactory.CreateLogger($"Aspire.Hosting.Dashboard.{categoryTrimmed}");
+        }
+
+        // TODO: We should log the state as well.
+
+        logger.Log(logMessage.LogLevel, logMessage.EventId, logMessage.Message, null, (s, _) => s);
+    }
+}
+
+internal sealed class DashboardLogMessage
+{
+    public string Timestamp { get; set; } = string.Empty;
+    public int EventId { get; set; }
+    [JsonConverter(typeof(JsonStringEnumConverter<LogLevel>))]
+    public LogLevel LogLevel { get; set; }
+    public string Category { get; set; } = string.Empty;
+    public string Message { get; set; } = string.Empty;
+    public JsonObject? State { get; set; }
+}
+
+[JsonSerializable(typeof(DashboardLogMessage))]
+internal sealed partial class DashboardLogMessageContext : JsonSerializerContext
+{
+
 }
--- a/src/Aspire.Hosting/DistributedApplicationBuilder.cs
+++ b/src/Aspire.Hosting/DistributedApplicationBuilder.cs
@@ -76,10 +76,14 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
 
         _innerBuilder.Logging.AddFilter("Microsoft.Hosting.Lifetime", LogLevel.Warning);
         _innerBuilder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
+        _innerBuilder.Logging.AddFilter("Aspire.Hosting.Dashboard", LogLevel.None);
 
         // This is so that we can see certificate errors in the resource server in the console logs.
         // See: https://github.com/dotnet/aspire/issues/2914
         _innerBuilder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer", LogLevel.Warning);
+
+        // Add the logging configuration again to allow the user to override the defaults
+        _innerBuilder.Logging.AddConfiguration(_innerBuilder.Configuration.GetSection("Logging"));
 
         AppHostDirectory = options.ProjectDirectory ?? _innerBuilder.Environment.ContentRootPath;
 
--- a/tests/Aspire.Hosting.Tests/Dashboard/DashboardResourceTests.cs
+++ b/tests/Aspire.Hosting.Tests/Dashboard/DashboardResourceTests.cs
@@ -1,11 +1,15 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Concurrent;
+using System.Text.Json;
+using Aspire.Hosting.Dashboard;
 using Aspire.Hosting.Dcp;
 using Aspire.Hosting.Tests.Utils;
 using Aspire.Hosting.Utils;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Xunit;
 
 namespace Aspire.Hosting.Tests.Dashboard;
@@ -122,6 +126,11 @@ public class DashboardResourceTests
             {
                 Assert.Equal("DASHBOARD__OTLP__AUTHMODE", e.Key);
                 Assert.Equal("Unsecured", e.Value);
+            },
+            e =>
+            {
+                Assert.Equal("LOGGING__CONSOLE__FORMATTERNAME", e.Key);
+                Assert.Equal("json", e.Value);
             }
         );
     }
@@ -283,11 +292,121 @@ public class DashboardResourceTests
         using var builder = TestDistributedApplicationBuilder.Create(new DistributedApplicationOptions
         {
             DisableDashboard = true,
-            Args = ["--environment", "Development"] }
+            Args = ["--environment", "Development"]
+        }
         );
 
         // Container validation logic runs when the service provider is built.
         using var app = builder.Build();
+    }
+
+    [Theory]
+    [InlineData(LogLevel.Critical)]
+    [InlineData(LogLevel.Error)]
+    [InlineData(LogLevel.Warning)]
+    [InlineData(LogLevel.Information)]
+    [InlineData(LogLevel.Debug)]
+    [InlineData(LogLevel.Trace)]
+    public async Task DashboardLifecycleHookWatchesLogs(LogLevel logLevel)
+    {
+        using var builder = TestDistributedApplicationBuilder.Create();
+
+        var loggerProvider = new TestLoggerProvider();
+
+        builder.Services.AddLogging(b =>
+        {
+            b.AddProvider(loggerProvider);
+            b.AddFilter("Aspire.Hosting.Dashboard", logLevel);
+        });
+
+        var dashboardPath = Path.GetFullPath("dashboard");
+
+        builder.Services.Configure<DcpOptions>(o =>
+        {
+            o.DashboardPath = dashboardPath;
+        });
+
+        var app = builder.Build();
+
+        await app.ExecuteBeforeStartHooksAsync(default);
+
+        var model = app.Services.GetRequiredService<DistributedApplicationModel>();
+        var resourceNotifcationService = app.Services.GetRequiredService<ResourceNotificationService>();
+        var resourceLoggerService = app.Services.GetRequiredService<ResourceLoggerService>();
+
+        var dashboard = Assert.Single(model.Resources.OfType<ExecutableResource>());
+
+        Assert.NotNull(dashboard);
+        Assert.Equal("aspire-dashboard", dashboard.Name);
+
+        // Push a notification through to the dashboard resource.
+        await resourceNotifcationService.PublishUpdateAsync(dashboard, "aspire-dashboard-0", s => s with { State = "Running" });
+
+        // Push some logs through to the dashboard resource.
+        var logger = resourceLoggerService.GetLogger("aspire-dashboard-0");
+
+        // The logging watcher expects a JSON payload
+        var dashboardLogMessage = new DashboardLogMessage
+        {
+            Category = "Test",
+            LogLevel = logLevel,
+            Message = "Test dashboard message"
+        };
+
+        logger.Log(logLevel, 0, JsonSerializer.Serialize(dashboardLogMessage), null, (s, _) => s);
+
+        // Get the logger with the category we expect Aspire.Hosting.Dashboard.Test
+        var testLogger = loggerProvider.CreateLogger("Aspire.Hosting.Dashboard.Test") as TestLogger;
+
+        Assert.NotNull(testLogger);
+
+        // Get the first log message that was logged
+        var log = await testLogger.FirstLogTask.WaitAsync(TimeSpan.FromSeconds(30));
+
+        Assert.Equal("Test dashboard message", log.Message);
+        Assert.Equal(logLevel, log.LogLevel);
+
+    }
+
+    private sealed class TestLogger : ILogger
+    {
+        private readonly TaskCompletionSource<LogMessage> _tcs = new();
+
+        public Task<LogMessage> FirstLogTask => _tcs.Task;
+
+        public IDisposable? BeginScope<TState>(TState state) where TState : notnull => null;
+
+        public bool IsEnabled(LogLevel logLevel) => true;
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+        {
+            var message = new LogMessage
+            {
+                LogLevel = logLevel,
+                Message = formatter(state, exception)
+            };
+
+            _tcs.TrySetResult(message);
+        }
+            
+
+        public sealed class LogMessage
+        {
+            public LogLevel LogLevel { get; set; }
+            public string Message { get; set; } = string.Empty;
+        }
+    }
+
+    private sealed class TestLoggerProvider : ILoggerProvider
+    {
+        private readonly ConcurrentDictionary<string, TestLogger> _loggers = new();
+
+        public ILogger CreateLogger(string categoryName)
+        {
+            return _loggers.GetOrAdd(categoryName, _ => new TestLogger());
+        }
+
+        public void Dispose() { }
     }
 
     private sealed class MockDashboardEndpointProvider : IDashboardEndpointProvider
--- a/playground/mongo/Mongo.AppHost/appsettings.json
+++ b/playground/mongo/Mongo.AppHost/appsettings.json
@@ -3,7 +3,8 @@
     "LogLevel": {
       "Default": "Information",
       "Microsoft.AspNetCore": "Warning",
-      "Aspire.Hosting.Dcp": "Warning"
+      "Aspire.Hosting.Dcp": "Warning",
+      "Aspire.Hosting.Dashboard": "Warning"
     }
   }
 }
--- a/src/Aspire.Hosting/Dashboard/DashboardOptions.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardOptions.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Globalization;
 using Aspire.Hosting.Dcp;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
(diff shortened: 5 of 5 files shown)


src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                if (@event.Resource.Name == KnownResourceNames.AspireDashboard && @event.ResourceId != @event.Resource.Name)
+                {
+                    dashboardResourceId = @event.ResourceId;
+                    break;
+                }
+            }
+        }
+        catch (OperationCanceledException)
+        {
+            distributedApplicationLogger.LogError("Failed to get the dashboard logs.");
```
davidfowl, line 190: Not sure if I should log this. This could happen if dcp is unhealthy.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                                "Error" => LogLevel.Error,
+                                "Critical" => LogLevel.Critical,
+                                "Trace" => LogLevel.Trace,
+                                _ => LogLevel.Information
+                            };
+
+                            if (!loggerCache.TryGetValue(logMessage.Category, out var logger))
+                            {
+                                var categoryTrimmed = logMessage.Category.StartsWith("Aspire.Dashboard.") ?
+                                    logMessage.Category["Aspire.Dashboard.".Length..] : logMessage.Category;
```
davidfowl, line 230: Looks strange to see Aspire.Hosting.Dashboard.Aspire.Dashboard.Category, so I trim the prefix and append Aspire.Hosting. Why is this important? Well there are logs emitting from categories that don't start with Aspire.Dashboard so we want to prefix *all* logs so that they can be controlled by config.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -76,6 +76,7 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
 
         _innerBuilder.Logging.AddFilter("Microsoft.Hosting.Lifetime", LogLevel.Warning);
         _innerBuilder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
+        _innerBuilder.Logging.AddFilter("Aspire.Hosting.Dashboard", LogLevel.Error);
```
davidfowl, line 79: Not sure we want to do this. Maybe we should update the templates? The downside is that upgrading won't have it and there will be LOTS of noise.

eerhardt, line 79:
```suggestion
        _innerBuilder.Logging.AddFilter("Aspire.Hosting.Dashboard", LogLevel.Warning);
```
I think we should be showing warnings as well. If we don't show them, what value would they be? No one would ever see them.

Note that Kestrel being filtered to only Error is really a bug that we should fix.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
eerhardt: I think logging it is fine. If DCP is unhealthy, what other indication do we have?

Maybe just a warning though?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
     }
+
+    private async Task SubscribeDashboardLogsAsync()
+    {
+        string? dashboardResourceId = null;
+
+        try
+        {
+            using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(1));
+            await foreach (var @event in resourceNotificationService.WatchAsync(cts.Token))
```
eerhardt, line 177:
```suggestion
            await foreach (var notification in resourceNotificationService.WatchAsync(cts.Token))
```

So you don't have to use `@event` everywhere.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            defaultDashboardLogger.LogError(ex, "Error logging dashboard logs.");
+        }
+    }
+}
+
+internal class DashboardLogMessage
+{
+    public string Timestamp { get; set; } = string.Empty;
+    public int EventId { get; set; }
+    public string LogLevel { get; set; } = string.Empty;
```
eerhardt, line 258:
```suggestion
    public LogLevel LogLevel { get; set; }
```

Why not use the enum?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+    public string Category { get; set; } = string.Empty;
+    public string Message { get; set; } = string.Empty;
+    public JsonObject? State { get; set; }
+}
+
+[JsonSerializable(typeof(DashboardLogMessage))]
+[JsonSourceGenerationOptions(ReadCommentHandling = JsonCommentHandling.Skip)]
```
eerhardt, line 265: Do we really emit JSON comments in our logs?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
eerhardt: > Well there are logs emitting from categories that don't start with Aspire.Dashboard so we want to prefix all logs so that they can be controlled by config.

This would make a great comment in the code.

eerhardt:
Don't forgot to add tests.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Doesn't work as we get a string from the external process that needs to be converted to an enum.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: That's part of this change https://github.com/dotnet/aspire/pull/3523/files#diff-dffe4637b4710f58a20faab4d5a5e45a64e6d6fc71443eb3136286afa68b3c88R154

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
eerhardt: Why can't you use `JsonStringEnumConverter<LogLevel>`?

davidfowl: overhead? I can write an switch statement and avoid all of the overhead.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Oh comments, my bad. We don't. I can remove this (copy pasta).

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
eerhardt: > I can write a switch statement and avoid all of the overhead.

Following that same logic... you might as well use a Utf8JsonReader.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
davidfowl: Should we do the config thing or do we just want to hardcode it?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Look at all of that crazy code in the converter compared to this switch 😄

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
             }
         }));
     }
+
+    private async Task SubscribeDashboardLogsAsync()
+    {
+        string? dashboardResourceId = null;
```
JamesNK, line 172: Move resolving the dashboard resource ID to its own method

src/Aspire.Hosting/DistributedApplicationBuilder.cs
eerhardt: IMO hard-coding it is fine. Especially becuase 90%+ of the time people don't want to see the dashboard logs.

Is it possible for a config setting to override the hard-coded value? That would be ideal.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+        }
+
+        var defaultDashboardLogger = loggerFactory.CreateLogger("Aspire.Hosting.Dashboard");
+
+        var loggerCache = new Dictionary<string, ILogger>();
+
+        try
+        {
+            // We turned on the JSON formatter for the logger, so we can log the log lines as JSON.
+            await foreach (var batch in resourceLoggerService.WatchAsync(dashboardResourceId).ConfigureAwait(false))
```
JamesNK, line 206: This probably stops when things are shutting down, but I think it would be good to have a CTS that is passed here and canceled when the lifecycle hook is stopping.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -141,6 +149,10 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
                 context.EnvironmentVariables[DashboardConfigNames.DashboardOtlpAuthModeName.EnvVarName] = "Unsecured";
             }
 
+            // Change the dashboard formatter to use JSON so we can parse the logs and render them in the
+            // via the ILogger.
+            context.EnvironmentVariables["LOGGING__CONSOLE__FORMATTERNAME"] = "json";
```
JamesNK, line 154: I like it. Better than my idea of copying the output directly to the host console.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -36,6 +42,8 @@ public Task BeforeStartAsync(DistributedApplicationModel model, CancellationToke
             AddDashboardResource(model);
         }
 
+        _ = SubscribeDashboardLogsAsync();
```
JamesNK, line 45: Can we store the task and await it when shutting down? Log unexcepted errors from the task.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
eerhardt: To this code, it is just 1 attribute on a property vs a 10 line switch.

As for runtime overhead, I wonder what performs better - doing an alloc-free enum parsing (what the converter does) or allocating a string just to switch over it...?

src/Aspire.Hosting/DistributedApplicationBuilder.cs
davidfowl: I thought that was the case but I couldn't get it to override.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/DistributedApplicationBuilder.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +166,104 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
         }));
     }
+
+    private async Task SubscribeDashboardLogsAsync()
+    {
+        string? dashboardResourceId = null;
+
+        try
+        {
+            using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(1));
```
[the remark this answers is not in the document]
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

tests/Aspire.Hosting.Tests/Dashboard/DashboardResourceTests.cs
```diff
@@ -283,13 +292,117 @@ public void ContainerIsValidWithDashboardIsDisabled()
...
+
+        // Get the first log message that was logged
+        var log = await testLogger.FirstLogTask.WaitAsync(TimeSpan.FromSeconds(30));
+
+        Assert.Equal("Test dashboard message", log.Message);
+        Assert.Equal(logLevel, log.LogLevel);
+
+    }
+
+    // Write a TestLogger : ILogger that captures log messages and sets a tcs when a message is logged.
```
davidfowl, line 363: Copilot 😄

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -36,9 +45,30 @@ public Task BeforeStartAsync(DistributedApplicationModel model, CancellationToke
...
+
+        if (_dashboardLogsTask is not null)
+        {
+            try
+            {
+                await _dashboardLogsTask.ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                distributedApplicationLogger.LogError(ex, "Unexpected error while watching dashboard logs.");
```
mitchdenny, line 65: LogCritical?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -90,7 +120,8 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
             {
                 ExecutableResource => KnownResourceTypes.Executable,
                 ProjectResource => KnownResourceTypes.Project,
-                _ => KnownResourceTypes.Container
+                ContainerResource => KnownResourceTypes.Container,
```
mitchdenny, line 123: Is this so we can test it as a container?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                            }
+
+                            logger.Log(logMessage.LogLevel, logMessage.EventId, logMessage.Message, null, (s, _) => s);
+                        }
+                    }
+                    catch (JsonException)
+                    {
+                        // Failed to parse, it's not JSON, just log the content as is.
+                        var level = logLine.IsErrorMessage ? LogLevel.Error : LogLevel.Information;
+                        defaultDashboardLogger.Log(level, 0, logLine.Content, null, (s, _) => s);
```
mitchdenny, line 264: Should we log a warning here as well that the log output is malformed. Could be the sign of a stray Console.Write(...).

davidfowl: I don't think so. If they are stray console.writelines we shouldn't warn.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Should it be? @eerhardt left it at warning.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+    {
+        async Task<string?> GetDashboardResourceIdentifier()
+        {
+            string? dashboardResourceId = null;
+
+            try
+            {
+                // The dashboard resource isn't immediately available so watch for it.
+                // Wait 1 min for dashboard to startup and be reported before giving up.
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```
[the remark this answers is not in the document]
drewnoakes:
```suggestion
                // Wait for dashboard to startup and be reported before giving up.
                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
mitchdenny: I guess if we are in dispose we are going down anyway.

mitchdenny (approved):
LGTM. would be good to see the timeout put as an option we could change if necessary. Sometimes its helpful for debugging issues.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -199,7 +199,7 @@ private async Task WatchDashboardLogsAsync()
             try
             {
                 // The dashboard resource isn't immediately available so watch for it.
-                // Wait 1 min for dashboard to startup and be reported before giving up.
+                // Wait for dashboard to startup and be reported before giving up.
                 using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```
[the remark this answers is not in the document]
drewnoakes: Alternatively: increase for `DEBUG` builds, or detect if a debugger is attached (or both).

drewnoakes (approved):

davidfowl:
This needs to show no logs by default. I'll make that change so that it's config enabled.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -17,8 +20,14 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
                                              IOptions<DashboardOptions> dashboardOptions,
                                              ILogger<DistributedApplication> distributedApplicationLogger,
                                              IDashboardEndpointProvider dashboardEndpointProvider,
-                                             DistributedApplicationExecutionContext executionContext) : IDistributedApplicationLifecycleHook
+                                             DistributedApplicationExecutionContext executionContext,
+                                             ResourceNotificationService resourceNotificationService,
+                                             ResourceLoggerService resourceLoggerService,
+                                             ILoggerFactory loggerFactory) : IDistributedApplicationLifecycleHook, IAsyncDisposable
 {
+    private readonly CancellationTokenSource _shutdownTcs = new();
```
JamesNK, line 28:
```suggestion
    private readonly CancellationTokenSource _shutdownCts = new();
```
And everywhere else.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -36,9 +45,30 @@ public Task BeforeStartAsync(DistributedApplicationModel model, CancellationToke
...
+    {
+        _shutdownTcs.Cancel();
+
+        if (_dashboardLogsTask is not null)
+        {
+            try
+            {
+                await _dashboardLogsTask.ConfigureAwait(false);
+            }
+            catch (Exception ex)
```
JamesNK, line 63: Add catch for OperationCanceledException. Don't log for it.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            {
+                // The dashboard resource isn't immediately available so watch for it.
+                // Wait for dashboard to startup and be reported before giving up.
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+
+                await foreach (var notification in resourceNotificationService.WatchAsync(cts.Token))
+                {
+                    // We want to capture the replica of the dashboard resource. This will allow us to get the logs for the dashboard.
+                    // When the resource id is not the same as the resource name, it's the replica.
+                    if (notification.Resource.Name == KnownResourceNames.AspireDashboard && notification.ResourceId != notification.Resource.Name)
```
JamesNK, line 209:
```suggestion
                    if (string.Equals(notification.Resource.Name, KnownResourceNames.AspireDashboard, StringComparisons.ResourceName) && notification.ResourceId != notification.Resource.Name)
```
I'm not sure about resource ID. Is it case insensitive as well?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+        }
+
+        if (await GetDashboardResourceIdentifier().ConfigureAwait(false) is not string dashboardResourceId)
+        {
+            // The only way this can happen is if DCP is unhealthy, so log a warning.
+            distributedApplicationLogger.LogWarning("Failed to get the dashboard logs.");
+            return;
+        }
+
+        var loggerCache = new Dictionary<string, ILogger>();
```
JamesNK, line 230: Be explicit about logger name match?
```suggestion
        var loggerCache = new Dictionary<string, ILogger>(StringComparer.Ordinal);
```

JamesNK, line 230: Assuming zero trust between hosting and dashboard, should there be a limit of cached loggers?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                                // Looks strange to see Aspire.Hosting.Dashboard.Aspire.Dashboard.Category,
+                                // so trim the prefix and append Aspire.Hosting.Why is this important?
+                                // Well there are logs emitting from categories that don't start with Aspire.Dashboard so we want to prefix all logs so that they can be controlled by config.
+                                var categoryTrimmed = logMessage.Category.StartsWith("Aspire.Dashboard.") ?
+                                    logMessage.Category["Aspire.Dashboard.".Length..] : logMessage.Category;
+
+                                loggerCache[logMessage.Category] = logger = loggerFactory.CreateLogger($"Aspire.Hosting.Dashboard.{categoryTrimmed}");
+                            }
+
+                            logger.Log(logMessage.LogLevel, logMessage.EventId, logMessage.Message, null, (s, _) => s);
```
JamesNK, line 257: Does this lose structured data? Not really important because we're just logging to the console.

Maybe have a TODO to convert the state payload into the right type to pass into logging here.

JamesNK (changes requested):

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                        }
+                    }
+                    catch (JsonException)
```
ReubenBond, line 260: This catch could be around just the JSON bit to reduce indentation & improve readability

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            {
+                foreach (var logLine in batch)
+                {
+                    try
+                    {
+                        var logMessage = JsonSerializer.Deserialize(logLine.Content, DashboardLogMessageContext.Default.DashboardLogMessage);
+
+                        if (logMessage is not null)
+                        {
+                            if (!loggerCache.TryGetValue(logMessage.Category, out var logger))
```
ReubenBond, line 246: Good candidate for a local function or a separate method IMO

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            defaultDashboardLogger.LogError(ex, "Error reading dashboard logs.");
+        }
+        finally
+        {
+            defaultDashboardLogger.LogDebug("Stopped reading dashboard logs.");
+        }
+    }
+}
+
+internal class DashboardLogMessage
```
ReubenBond, line 284:
```suggestion
internal sealed class DashboardLogMessage
```

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
ReubenBond: If we turn the code to get-or-add the logger above into a local function, this will be nice and clear without any other change

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +189,111 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            return;
+        }
+
+        var loggerCache = new Dictionary<string, ILogger>();
+        var defaultDashboardLogger = loggerFactory.CreateLogger("Aspire.Hosting.Dashboard");
+
+        try
+        {
+            // We turned on the JSON formatter for the logger, so we can log the log lines as JSON.
+            await foreach (var batch in resourceLoggerService.WatchAsync(dashboardResourceId).WithCancellation(_shutdownTcs.Token).ConfigureAwait(false))
```
ReubenBond, line 236: Because we have no sequencing w.r.t shutdown order, the cancellation+await in `DisposeAsync` likely happens too late for us to guarantee that `resourceLoggerService` here wouldn't have already been disposed. I took a look at the DI implementation and it appears that we dispose in activation order, rather than the reverse, so dependencies will be disposed before dependents.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
JamesNK: Hmm, I was going to say that you could consider removing altogether since there is already a cache inside LoggerFactory: https://github.com/dotnet/runtime/blob/f0991722db63e0b4962439469efb4ea1b72923f3/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactory.cs#L139-L162

But it doesn't limit the number of loggers either...

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: We create the resource so no, it doesn't need to be insensitive.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: oops

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: There's an exception inside with this catch.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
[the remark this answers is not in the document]
davidfowl: I made it a setting and also removed the timeout when debugging.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
JamesNK: Not if there is an existing project with the name though. Other places check the name without case sensitivity: https://github.com/dotnet/aspire/blob/8e27fadd7db0a775bf23f38b9bc6a855671f101b/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs#L26

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Yea, not adding a limit here.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: True, I forgot we had this comparer. Fixed

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +193,135 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+        {
+            string? dashboardResourceId = null;
+
+            try
+            {
+                var timeout = Debugger.IsAttached ? Timeout.InfiniteTimeSpan : dashboardOptions.Value.DashboardStartupTimeout;
+
+                // The dashboard resource isn't immediately available so watch for it.
+                // Wait for dashboard to startup and be reported before giving up.
+                using var cts = new CancellationTokenSource(timeout);
```
JamesNK, line 209: link this with _shutdownCts

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -154,4 +193,135 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+            return;
+        }
+
+        var loggerCache = new Dictionary<string, ILogger>(StringComparer.Ordinal);
+        var defaultDashboardLogger = loggerFactory.CreateLogger("Aspire.Hosting.Dashboard");
+
+        try
+        {
+            // We turned on the JSON formatter for the logger, so we can log the log lines as JSON.
+            await foreach (var batch in resourceLoggerService.WatchAsync(dashboardResourceId).WithCancellation(_shutdownCts.Token).ConfigureAwait(false))
```
JamesNK, line 247: There is a call to dispose the CTS in DisposeAsync: `_shutdownCts.Dispose();`

It's possible that dispose happens before the token property on the CTS here is accessed. An ObjectDisposedException is thrown from `CancellationTokenSource.Token`.

Potential fixes:
* Pass the token from the CTS into the method as a parameter and then always use the parameter <- best imo
* Don't dispose _shutdownCts

davidfowl: Never dispose the cts, that's the rule 😄 . Done

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Done

JamesNK (approved):

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: >  I took a look at the DI implementation and it appears that we dispose in activation order, rather than the reverse, so dependencies will be disposed before dependents.

Disposable happens in reverse activation order https://github.com/dotnet/runtime/blob/d88c7ba88627b4b68ad523ba27cb354809eb7e67/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/ServiceProviderEngineScope.cs#L148

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -76,11 +76,15 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
 
         _innerBuilder.Logging.AddFilter("Microsoft.Hosting.Lifetime", LogLevel.Warning);
         _innerBuilder.Logging.AddFilter("Microsoft.AspNetCore.Server.Kestrel", LogLevel.Error);
+        _innerBuilder.Logging.AddFilter("Aspire.Hosting.Dashboard", LogLevel.None);
```
eerhardt, line 79: If the Dashboard crashes, the user doesn't get any indication of what happened, right? They just see a hang. 

https://github.com/dotnet/aspire/issues/3429

We should be at least logging errors by default.

davidfowl: We can’t. I tried the playground samples and lots of other errors showed up. 

I think it could be done if we filtered to the startup category by default, but it can’t be all errors.

Issue #3176, opened 2024-03-26
Data present in the table getting truncated after zooming page to 200% : A11y_.NetAspireDashboard_Resources_Resources_Zoom


PoojaNamde:
"[Check out Accessibility Insights](https://nam06.safelinks.protection.outlook.com/?url=https://accessibilityinsights.io/&data=02%7c01%7cv-stfe%40microsoft.com%7cb67b2c4b646d4f9561a208d6f4b5c39b%7c72f988bf86f141af91ab2d7cd011db47%7c1%7c0%7c636965458850501301&sdata=mxhokIKNMb22llsjXHLgU3XZibj1Qfx37rpY4PU2sfE%3D&reserved=0)! - Identify accessibility bugs before check-in and make bug fixing faster and easier."

**Please do not close this bug. This bug should only be closed by Trusted Tester after verification.**

### Tags:
#A11yTCS; #A11yMAS; #A11ySev2; #Win11; #ChromiumEdge; #FTP; #DesktopWeb; #.NET Core; #SH_.Net Aspire Dashboard_Web_Mar2024; #WCAG1.4.4; #Zoom;

### Prerequisites:
Set display resolution to 1280*768.

### Environment Details:
Application Name: .Net Aspire Dashboard
URL: https://localhost:19888/
Chrome Version 122.0.2365.80 (Official build) (64-bit)
Window Version: Win 11
OS Build: 22621.3085

### Repro Steps:

1. Hit the URL: http://localhost:15888/
2. TAB to "Resources" tab item and hit select the tab by pressing Enter key.
3. Zoom the page to 200% and observe the issue.

### Actual Results:
Data present in the table getting truncated after zooming page to 200%.

**Similar issue is also observed in throughout the application for the table screen.**

### Expected Results:
All data and controls on the page should be clearly visible when page is zoomed to 200%.

### User Impact:
Low vision users will not be able to see the content present in the page clearly and will miss the information. 

### Attachment:

![3176_A11y_ NetAspireDashboard_ResourceTab_AfterZoom](https://github.com/dotnet/aspire/assets/90901519/8d32b705-5a71-4356-8200-fe8f1b5632b2)
![3176_A11y_ NetAspireDashboard_ResourceTab_BeforeZoom](https://github.com/dotnet/aspire/assets/90901519/c49ed100-4336-4a41-978f-7175c815ec1b)
[...]

PR #3538, opened 2024-04-09
Add tooltips to Resource table

adamint:
Required for #3176
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3538)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor
@@ -1,11 +1,12 @@
 ﻿@page "/"
 @using Aspire.Dashboard.Components.ResourcesGridColumns
-@using Aspire.Dashboard.Model
 @using Aspire.Dashboard.Resources
 @using Aspire.Dashboard.Utils
 @using System.Globalization
+@using Humanizer
 @inject IStringLocalizer<Dashboard.Resources.Resources> Loc
 @inject IStringLocalizer<ControlsStrings> ControlsStringsLoc
+@inject IStringLocalizer<Columns> ColumnsLoc
 
 <PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.Resources.ResourcesPageTitle)" Loc="@Loc" /></PageTitle>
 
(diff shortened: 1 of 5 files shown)


src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -57,21 +58,25 @@
...
                         <SourceColumnDisplay Resource="context" FilterText="@_filter" />
                     </TemplateColumn>
-                    <TemplateColumn Title="@Loc[nameof(Dashboard.Resources.Resources.ResourcesEndpointsColumnHeader)]">
-                        <EndpointsColumnDisplay Resource="context" HasMultipleReplicas="HasMultipleReplicas(context)" />
+                    <TemplateColumn Title="@Loc[nameof(Dashboard.Resources.Resources.ResourcesEndpointsColumnHeader)]" Tooltip="true" TooltipText="@(ctx => GetEndpointsTooltip(ctx))">
+                        <EndpointsColumnDisplay
+                            Resource="context"
+                            HasMultipleReplicas="HasMultipleReplicas(context)"
+                            DisplayedEndpoints="@GetDisplayedEndpoints(context, out var additionalMessage)"
+                            AdditionalMessage="@additionalMessage" />
```
kvenkatrajan, line 79: Can you explain a bit more on why we are doing this?

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -57,21 +58,25 @@
             }
             <FluentDataGrid Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" ItemSize="46" Items="@FilteredResources" ResizableColumns="true" GridTemplateColumns="@gridTemplateColumns" RowClass="GetRowClass" Loading="_isLoading">
                 <ChildContent>
-                    <PropertyColumn Title="@Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeColumnHeader)]" Property="@(c => c.ResourceType)" Sortable="true" />
-                    <TemplateColumn Title="@ControlsStringsLoc[nameof(ControlsStrings.NameColumnHeader)]" Sortable="true" SortBy="@_nameSort">
+                    <PropertyColumn Title="@Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeColumnHeader)]" Property="@(c => c.ResourceType)" Sortable="true" Tooltip="true" TooltipText="@(c => c.ResourceType)"/>
```
kvenkatrajan, line 61: Does this need to be localized?

adamint: It's not currently being localized and comes from DCP, so I would say no. We don't know all the types that we might be receiving

kvenkatrajan: Ok will confirm if they are keywords - we may want to skip localization

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: We need to add tooltips to the columns; to avoid duplicating code inside EndpointsColumnDisplay. I modified EndpointsColumnDisplay to call `GetDisplayedEndpoints`, now in Resources.razor.cs

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: If they do need localization, it should happen in a follow-up

src/Aspire.Dashboard/Components/Pages/Resources.razor
kvenkatrajan: got it thanks

kvenkatrajan (approved):

src/Aspire.Dashboard/Components/Pages/Resources.razor
drewnoakes: IMO we should be localising resource types and statuses, either with display strings coming from DCP, or with lookups on the dashboard side when the set is fixed.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -294,6 +296,84 @@ private async Task ExecuteResourceCommandAsync(ResourceViewModel resource, Comma
...
+        {
+            return containerImage;
+        }
+
+        if (resource.Properties.TryGetValue(KnownProperties.Resource.Source, out var value) && value.HasStringValue)
+        {
+            return value.StringValue;
+        }
+
+        return null;
```
drewnoakes, line 324: Please de-duplicate this with the code in `SourceColumnDisplay`. It'll be easy for these to fall out of sync with each other.

Doing this should improve the code in both places. It can be done with a new static method that returns a few different values:

- display value (for column)
- full value (for copy/tooltip
- sub text (optional, for `ContentAfterValue`)

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -294,6 +296,84 @@ private async Task ExecuteResourceCommandAsync(ResourceViewModel resource, Comma
...
+
+        var maxShownEndpoints = 3;
+        var tooltipBuilder = new StringBuilder(string.Join(", ", displayedEndpoints.Take(maxShownEndpoints).Select(endpoint => endpoint.Text)));
+
+        if (displayedEndpoints.Count > maxShownEndpoints)
+        {
+            tooltipBuilder.Append(CultureInfo.CurrentCulture, $" + {displayedEndpoints.Count - maxShownEndpoints}");
+        }
+
+        return tooltipBuilder.ToString();
```
drewnoakes, line 349: Ditto, de-dupe with `EndpointsColumnDisplay`.

tlmii:
Agree with @drewnoakes on the need to de-dupe that logic - it seems likely to end up diverging. Though if this is going to be backported to release/8.0 (guessing it is, as part of the final accessibility items) I wonder if we could do that de-dupe early in post-GA just to get this in this week? Just throwing that out there as I know there's a lot of OOF time.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
adamint: The content of the tooltip purposely diverges from the display in EndpointsColumnDisplay, I'm not sure this needs to be 'de-duped'. If so, I can file a follow-up

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: It seems like that's something that should be coming from DCP;otherwise we will have to keep a parallel list of available resource types.

src/Aspire.Dashboard/Components/ResourcesGridColumns/SourceColumnDisplay.razor
```diff
@@ -3,48 +3,27 @@
 @using Aspire.Dashboard.Resources
 @inject IStringLocalizer<Columns> Loc
 
-@if (Resource.IsProject() && Resource.TryGetProjectPath(out var projectPath))
+@if (ContentAfterValue is not null)
```
drewnoakes, line 6: Is it possible to put this `@if` inside the `<GridValue>`, around the `<ContentAfterValue>`?

drewnoakes (approved):

JamesNK (approved):

PR #3541, opened 2024-04-09
Change ParameterResource's "Failed to Start" to "Configuration Missing"

### LINKED ISSUE #3301: Change ParameterResource's "Failed to Start" to "Configuration Missing"
We now have the ability to set both the state text and the icon. We should use a more prescriptive error message and perhaps a log message explaining how to make parameters work.

eerhardt:
Also log the exception message for the Parameter resource so users know what the problem is.

Fix #3301
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3541)

### DIFF
--- a/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
@@ -82,7 +82,7 @@ public static class ParameterResourceBuilderExtensions
     /// </summary>
     private sealed class WriteParameterLogsHook(ResourceLoggerService loggerService, string resourceName, string message) : IDistributedApplicationLifecycleHook
     {
-        public Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
+        public Task BeforeStartAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
         {
             loggerService.GetLogger(resourceName).LogError(message);
 
@@ -91,7 +91,7 @@ public static class ParameterResourceBuilderExtensions
 
         public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken) => Task.CompletedTask;
 
-        public Task BeforeStartAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken) => Task.CompletedTask;
+        public Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken) => Task.CompletedTask;
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
@@ -88,10 +88,6 @@ public static class ParameterResourceBuilderExtensions
 
             return Task.CompletedTask;
         }
-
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -58,13 +61,39 @@ private static string GetParameterValue(IConfiguration configuration, string nam
...
         return builder.AddResource(resource)
                       .WithInitialState(state);
     }
 
+    /// <summary>
+    /// Writes the message to the specified resource's logs.
+    /// </summary>
+    private sealed class WriteParameterLogsHook(ResourceLoggerService loggerService, string resourceName, string message) : IDistributedApplicationLifecycleHook
+    {
+        public Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
```
davidfowl, line 85: This is too late as it runs after all resources are started.

eerhardt: Too late for what?

We just want to log the exception message into the log for the resource.

davidfowl: But you want it to show up quickly, not after all resources have started. If you're provisioning azure resources for e.g. this won't show up until after. These events run after all resources are provisioned.

davidfowl: BeforeStartAsync works since this resource fails super early.

davidfowl:
Should we also change the source to Configuration?

eerhardt:
> Should we also change the source to Configuration?

I don't know. I kind of like `Parameters:{name}"` showing up in the dashboard... as a indication of what you need to set. Although the log will have this information too.

Screenshot 2024-04-10 at 7 47 18 AM

All the other "Source" strings say where they come from: name of the container, name of the project. It feels consistent to have the name of the parameter here.

davidfowl:
That's what I thought but it seems like nobody figured it out.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -58,13 +61,39 @@ private static string GetParameterValue(IConfiguration configuration, string nam
...
+        public Task BeforeStartAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
+        {
+            loggerService.GetLogger(resourceName).LogError(message);
+
+            return Task.CompletedTask;
+        }
+
+        public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken) => Task.CompletedTask;
+
+        public Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken) => Task.CompletedTask;
```
davidfowl, line 94: These aren't needed

davidfowl (approved):

eerhardt:
> That's what I thought but it seems like nobody figured it out.

The log will really help out.

I don't think making this one inconsistent with the others makes sense. The others have the "name of the source". Unless you mean we preprend "Configuration" to `Parameters:{name}`.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
eerhardt: image

PR #3545, opened 2024-04-10
Implement IResourceWithArgs on ProjectResource

### LINKED ISSUE #3306: Implement IResourceWithArgs on ProjectResource
We should also be processing the argument list from the WithProjectDefaults method and supporting passing arguments to projects from the app model.

eerhardt:
Also fix the command line arg parsing to be the same as .NET Process's:
* double quotes around single arguments
* consecutive double quotes inside a quoted argument means one double quote

Fix #3306

"args" on a project resource doesn't work with `azd` yet. It doesn't error, but the project's container doesn't get started with the args. Logged https://github.com/Azure/azure-dev/issues/3678

cc @vhvb1989 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3545)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -12,6 +12,7 @@ using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Dashboard;
 using Aspire.Hosting.Dcp.Model;
 using Aspire.Hosting.Lifecycle;
+using Aspire.Hosting.Utils;
 using k8s;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1031,8 +1032,8 @@ private void PrepareProjectExecutables()
                 var launchProfile = project.GetEffectiveLaunchProfile();
```
davidfowl, line 1032: Why can't we move this logic to the project resource? We can detect the env variable there right and apply this logic?

eerhardt: What value would that add? Then we need to check the env var in 2 places.

eerhardt: It seems like it would be better to co-locate the logic that is dependent on the env var.

davidfowl: The resource logic should be as close to the resource as possible. If anything, I'd prefer if we could influence this logic with annotation as much as possible.

davidfowl: Spoke to @eerhardt offline and agreed this is a fine fix for GA.

davidfowl (approved):

PR #3547, opened 2024-04-10
Enable public API analyzer.

mitchdenny:
This PR brings in the `PublicAPI.*.txt` files for API reviews. This is mostly just enabling the analyzer and setting up the initial baselines. However, the analyzer did highlight a couple of issues that I've suppressed to get things building but we should review and react to before merging this PR.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3547)


src/Aspire.Hosting.Azure/AzureConstructResource.cs
```diff
@@ -109,6 +109,7 @@ public static IResourceBuilder<AzureConstructResource> AddAzureConstruct(this ID
     /// <param name="propertySelector">Property selection expression.</param>
     /// <param name="parameterResourceBuilder">Aspire parameter resource builder.</param>
     /// <param name="parameterName">The name of the parameter to be assigned.</param>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
     public static void AssignProperty<T>(this Resource<T> resource, Expression<Func<T, object?>> propertySelector, IResourceBuilder<ParameterResource> parameterResourceBuilder, string? parameterName = null) where T: notnull
```
mitchdenny, line 113: I'm not sure this is an issue. We have two variants of this overload, both take a different argument type as their second argument.

mitchdenny: .. we don't need to worry too much about breaking this API because its covered by experimental.

src/Aspire.Hosting.Dapr/IDistributedApplicationComponentBuilderExtensions.cs
```diff
@@ -30,6 +30,7 @@ public static class IDistributedApplicationResourceBuilderExtensions
     /// <param name="builder">The resource builder instance.</param>
     /// <param name="options">Options for configuring the Dapr sidecar, if any.</param>
     /// <returns>The resource builder instance.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0027:API with optional parameter(s) should have the most parameters amongst its public overloads", Justification = "<Pending>")]
     public static IResourceBuilder<T> WithDaprSidecar<T>(this IResourceBuilder<T> builder, DaprSidecarOptions? options = null) where T : IResource
```
mitchdenny, line 34: Better to just have another overload that does not take the options?

src/Aspire.Hosting/ApplicationModel/IResourceBuilder.cs
```diff
@@ -25,14 +25,16 @@ public interface IResourceBuilder<out T> where T : IResource
     /// <typeparam name="TAnnotation">The type of the annotation to add.</typeparam>
     /// <param name="behavior">The behavior to use when adding the annotation.</param>
     /// <returns>The resource builder instance.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
     IResourceBuilder<T> WithAnnotation<TAnnotation>(ResourceAnnotationMutationBehavior behavior = ResourceAnnotationMutationBehavior.Append) where TAnnotation : IResourceAnnotation, new() => WithAnnotation(new TAnnotation(), behavior);
```
mitchdenny, line 29: I think that there is an argument that we could get rid of this overload. It isn't used that much because folks almost always want to pass in the annotation object.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -363,6 +363,7 @@ private static void ApplyEndpoints<T>(this IResourceBuilder<T> builder, IResourc
     /// <param name="callback">Callback that modifies the endpoint.</param>
     /// <param name="createIfNotExists">Create endpoint if it does not exist.</param>
     /// <returns></returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
     public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, string endpointName, Action<EndpointAnnotation> callback, bool createIfNotExists = true) where T : IResourceWithEndpoints
```
mitchdenny, line 367: Maybe we turn this into three methods:

```csharp
public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, string endpointName, Action<EndpointAnnotation> callback) where T : IResourceWithEndpoints {}
public static IResourceBuilder<T> WithExistingEndpoint<T>(this IResourceBuilder<T> builder, string endpointName, Action<EndpointAnnotation> callback) where T : IResourceWithEndpoints {}
public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null) where T : IResource {}
```

Basically if you want to use the callback against an existing endpoint you would use `WithExistingEndpoint(...)`.

src/Aspire.Dashboard/Aspire.Dashboard.csproj
```diff
@@ -13,6 +13,10 @@
     <NoWarn>$(NoWarn);CS1591;CS8002</NoWarn>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <AddPublicApiAnalyzers>false</AddPublicApiAnalyzers>
```
drewnoakes, line 17: We have far too many public APIs in the dashboard project. Why disable it here?

mitchdenny: Aspire.Dashboard isn't an API surface though, its a tool that we ship. Do we expect people to be taking a dependency on Aspire.Dashboard in their own projects?

RussKie: Then there should be little reason for the API to be public? ;)

mitchdenny:
@joperezr would appreciate your input on my MSBuild hackery here. I wanted the public API tooling but I'm really just trying to ape what I saw in the _dotnet/aspnetcore_ repository.

src/Aspire.Hosting.Dapr/IDistributedApplicationComponentBuilderExtensions.cs
ReubenBond: Yes, that is the norm in the ASP.NET Core codebase - we did the same for Service Discovery (2 overloads). Make the options non-nullable, too.

src/Aspire.Dashboard/Aspire.Dashboard.csproj
eerhardt: > Do we expect people to be taking a dependency on Aspire.Dashboard in their own projects?

No we don't. It doesn't ship as a package on nuget.org.

> Then there should be little reason for the API to be public?

Some things use reflection that require things to be `public`, it won't reflect on `internal` types. Having public types and methods in an .exe is not an issue.

mitchdenny: You'd also be fighting the default Razor code gen all day long.

RussKie (approved):

RussKie:
> @joperezr would appreciate your input on my MSBuild hackery here. I wanted the public API tooling but I'm really just trying to ape what I saw in the _dotnet/aspnetcore_ repository.

Jose is OOF until next week. Is there anything specific you have questions about? I've looked at the changeset, and it looks sensible to me.

mitchdenny:
Ok I'll merge this once I get a clean build again. I want to make sure I don't break main

PR #3590, opened 2024-04-11
update versions to 8.0.4

danmoseley:
Update to yesterday's 8.0.4 where we shipped one.

The other updates came along for the ride. Used this procedure https://github.com/dotnet/aspire/pull/3512#issue-2232592285
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3590)

PR #3610, opened 2024-04-11
update subset of versions

danmoseley:
Update just versions that seem unlikely to be causing the failure in #3590 as a start at bisecting.

Note - the comments are merge helpers, really so that DCP updates don't conflict with the line below.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3610)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,13 +9,13 @@
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.306.3" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.300.71" />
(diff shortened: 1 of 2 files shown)


danmoseley:
this is not hitting the issue in the larger PR (it's hitting the general test failure). so when that general issue is fixed, we can take this.

eerhardt:
@radical  is going to disable this test. Someone deleted our azure resource and it broke our tests.

radical:
> @radical is going to disable this test. Someone deleted our azure resource and it broke our tests.

Update: @ReubenBond restored the resources, and this should work now. I have triggered a rebuild.

danmoseley:
@eerhardt can you sign off? then I can merge.

radical (approved):

Issue #3614, opened 2024-04-11
Message column duplicates if Attributes contains Message.


vbondaryuk:
During migration from the old code base, we are going to switch for OTEL. We are trying to use Aspire Dashboard for migration steps as a great toll for visualizing our Logs, Metrics, and traces.

We are goiung to keep our logs in the same format we currently have to keep backward compatibility. Everything works well except logging page. We faced the issue that Message column duplicates as a result filtering stopped to work as well as vilualization.

**Repro steps:**
```cs
using System.Diagnostics.Tracing;
using Microsoft.AspNetCore.Mvc;
using OpenTelemetry.Logs;

var builder = WebApplication.CreateBuilder(args);
builder.Logging.AddOpenTelemetry(logging =>
{
    logging.IncludeFormattedMessage = true;
    logging.IncludeScopes = true; 
});
builder.Services.Configure<OpenTelemetryLoggerOptions>(logging => logging.AddOtlpExporter());

var app = builder.Build();

app.MapGet("/sample", ([FromServices] ILogger<Program> logger) =>
{
    const string message = "Hello World";
    IReadOnlyList<KeyValuePair<string, object?>> eventData = new List<KeyValuePair<string, object?>>()
    {
        new KeyValuePair<string, object?>("Message", message),
        new KeyValuePair<string, object?>("ActivityId", 123),
        new KeyValuePair<string, object?>("Level", (int) EventLevel.Informational),
        new KeyValuePair<string, object?>("Tid", Thread.CurrentThread.ManagedThreadId),
        new KeyValuePair<string, object?>("Pid", Environment.ProcessId),
    };

    logger.Log(LogLevel.Information, 0, eventData, null, formatter: (_, _) => null);
    
    return message;
});
```

**Actual result:**
![image](https://github.com/dotnet/aspire/assets/5966463/1d51ab3f-c822-4621-8777-d8751ebc547c)

**Expected result:**
Only one Message column exist. Message from the Attributes can be renamed or added new Column `Body`.

JamesNK:
Hi

I took a look at this issue, and I don't agree with the recommended fix. It's valid to have an attribute in a log message called `Message`. It shouldn't override the log body message.

For example, `log.LogInformation("Error occurred: {Message}", "Test message");` should result in the UI showing `Error occurred: Test message`. It shouldn't just be `Test message`.

You're seeing this log message:
```
[...]

PR #3616, opened 2024-04-11
Fix Message column duplication.

vbondaryuk:
The PR is to fix the issue described here: Message column duplicates if Attributes contains Message. #3614 

If used this overload [ILogger<T>.Log<TState>(LogLevel, EventId, TState, Exception, Func<TState,Exception,String>)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger.log?view=dotnet-plat-ext-8.0#microsoft-extensions-logging-ilogger-log-1(microsoft-extensions-logging-loglevel-microsoft-extensions-logging-eventid-0-system-exception-system-func((-0-system-exception-system-string)))) of `ILogger<T>`  Where the `TState` has `Message` column dashboard incorrectly parse it and shows two Message column as a result filter and visualisation does not work.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3616)

Result with code changes: 
image

### DIFF
--- a/src/Aspire.Dashboard/Otlp/Model/OtlpLogEntry.cs
+++ b/src/Aspire.Dashboard/Otlp/Model/OtlpLogEntry.cs
@@ -27,30 +27,35 @@ public class OtlpLogEntry
     {
         string? originalFormat = null;
         string? parentId = null;
-        Attributes = record.Attributes.ToKeyValuePairs(options, filter: attribute =>
-        {
-            switch (attribute.Key)
+        string? message = null;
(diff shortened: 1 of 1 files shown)


JamesNK:
Closing PR because of reasons in https://github.com/dotnet/aspire/issues/3614#issuecomment-2071406209

https://github.com/dotnet/aspire/pull/3881 addresses the underlying issue.

PR #3618, opened 2024-04-11
Fix Qdrant resource to be publishable

eerhardt:
1. Set the transport to be http2 on the grpc endpoint
2. Use the Url property in the connection string for simplicity

The app still doesn't work end-to-end yet. It is blocked by https://github.com/Azure/azure-dev/issues/3690, which should be fixed soon. cc @ellismg 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3618)

### DIFF
--- a/playground/Qdrant/Qdrant.AppHost/Program.cs
+++ b/playground/Qdrant/Qdrant.AppHost/Program.cs
@@ -4,9 +4,10 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
 var qdrant = builder.AddQdrant("qdrant")
-    .WithDataVolume("qdrant_data");
+    .WithDataVolume("qdrant-data");
 
 builder.AddProject<Projects.Qdrant_ApiService>("apiservice")
+    .WithExternalHttpEndpoints()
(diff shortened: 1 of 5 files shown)


radical:
> The app still doesn't work end-to-end yet. It is blocked by https://github.com/Azure/azure-dev/issues/3690, which should be fixed soon.

Because this is about `azd`, this wouldn't be caught by adding an EndToEnd test, or a workload test, is that correct?

danmoseley:
will meet backport bar

eerhardt:
> > The app still doesn't work end-to-end yet. It is blocked by [Azure/azure-dev#3690](https://github.com/Azure/azure-dev/issues/3690), which should be fixed soon.
> 
> Because this is about `azd`, this wouldn't be caught by adding an EndToEnd test, or a workload test, is that correct?

Correct. This would only be caught when deploying to Azure.

timheuer (approved):

davidfowl (approved):

PR #3630, opened 2024-04-12
Fix configuration key in error messages

drewnoakes:
At some point during development of #3443 the configuration key was changed and these error messages were missed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3630)


tlmii (approved):

davidfowl (approved):

drewnoakes:
@danmoseley would you consider this for 8.0? Seems very safe, and could cause confusion. Not sure it meets the bar though.

PR #3654, opened 2024-04-12
Persist browser token auth cookie across sessions

### LINKED ISSUE #3646: [WebToolsE2E][Aspire] Sign out the vm account and reconnect to vm using the same account, then open the project that has verified the token authentication or create a new aspire project and F5, it still pops up dashboard token authentication again.
**REGRESSION INFO:** This is a new feature on Aspire 8.0 P6

**INSTALL STEPS**

2. Install VS 17.10 Preview 4 FB
3. Install Aspire 8.0 P6 [8.0.0-preview.6.24211.5]
4. Apply dotnet-tools Feed

**REPRO STEPS**

1. Create an Aspire Starter App > give a name 'AspireApp1' > Create
2. F5 with https > popup dashboard token authentication > Input token and Log in
3. Sign out vm's account, then connect to vm again using same account
4. Open the authenticated 'AspireApp1' project, then F5 with https, it will popup dashboard token authentication again

**ACTUAL**

Sign out the vm account and reconnect to vm using the same account, then open the project that has verified the token authentication or create a new aspire project and F5, it still pops up dashboard token authentication again

**Note**

- Disconnecting server and reconnecting vm works fine.

**EXPECTED**

There is no need to reconnect vm with token authentication again.

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/3646

The auth cookie currently has a session lifetime. My experience with the dashboard is you can close and reopen the browser and it is remembered. I thought it was persisted. But it looks like the meaning of session is ambiguous. See https://stackoverflow.com/questions/4132095/when-does-a-cookie-with-expiration-time-at-end-of-session-expire

This PR updates cookie auth to persist the cookie.
* Lifetime of the cookie is the same as the lifetime of the auth token inside the cookie - so 3 days
* Allow it to be refreshed (does this mean it can be extended for sliding experation? @halter73?)

This change means the cookie will be remembered consistently (don't need to worry about the browser deciding the session is done) and between machine restarts.

cc @davidfowl @DamianEdwards What do you think?
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3654)

### DIFF
--- a/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
+++ b/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
@@ -91,7 +91,7 @@ internal sealed class ValidateTokenMiddleware
         await httpContext.SignInAsync(
             CookieAuthenticationDefaults.AuthenticationScheme,
             claims,
-            new AuthenticationProperties { IsPersistent = true, AllowRefresh = true }).ConfigureAwait(false);
+            new AuthenticationProperties { IsPersistent = true }).ConfigureAwait(false);
         return true;
     }
 }
--- a/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
+++ b/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
@@ -88,7 +88,10 @@ internal sealed class ValidateTokenMiddleware
             authenticationType: CookieAuthenticationDefaults.AuthenticationScheme);
         var claims = new ClaimsPrincipal(claimsIdentity);
 
-        await httpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme, claims).ConfigureAwait(false);
+        await httpContext.SignInAsync(
+            CookieAuthenticationDefaults.AuthenticationScheme,
+            claims,
+            new AuthenticationProperties { IsPersistent = true, AllowRefresh = true }).ConfigureAwait(false);
         return true;
     }
 }


kvenkatrajan:
If we are using a persistent cookie, does sliding expiration apply? We may need to explicitly set an expiry time here.

JamesNK:
ASP.NET Core cookie auth has the concept of refreshing the auth token, which in this case should write a new cookie. I don't know the exact logic of how it works.

DamianEdwards (approved):
I'm ok with this if it improves the user experience.

src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
```diff
@@ -88,7 +88,10 @@ public static async Task<bool> TryAuthenticateAsync(string incomingBrowserToken,
             authenticationType: CookieAuthenticationDefaults.AuthenticationScheme);
         var claims = new ClaimsPrincipal(claimsIdentity);
 
-        await httpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme, claims).ConfigureAwait(false);
+        await httpContext.SignInAsync(
+            CookieAuthenticationDefaults.AuthenticationScheme,
+            claims,
+            new AuthenticationProperties { IsPersistent = true, AllowRefresh = true }).ConfigureAwait(false);
```
halter73, line 94: Nit: `AllowRefresh = true` is the [default behavior](https://github.com/dotnet/aspnetcore/blob/7412c71cf881392233b5652310ab2b9e7fdf71eb/src/Security/Authentication/Cookies/src/CookieAuthenticationHandler.cs#L97) if you want to rely on that. I doubt we'd change it.

```suggestion
            new AuthenticationProperties { IsPersistent = true }).ConfigureAwait(false);
```

halter73 (approved):
I'm not sure if there's any UI behind this to make it easier with something like a checkbox, but you could consider setting `IsPersistent` based on a query string parameter like `?rememberMe=true`.

For `MapIdentityApi`, we went with `?useCookie=true` being persistent and `?useSessionCookie=true` as non-persistent. That's a bit different though because the default was to issue and access token in the response body of the /login endpoint.

> - Lifetime of the cookie is the same as the lifetime of the auth token inside the cookie - so 3 days

Out of curiosity, is there a reason you chose three days instead of the default a 14?

> - Allow it to be refreshed (does this mean it can be extended for sliding expiration? halter73?)

Yes, assuming you haven't changed the value of `CookieAuthenticationOptions.SlidingExpiration` away from its [default value of true](https://github.com/dotnet/aspnetcore/blob/7412c71cf881392233b5652310ab2b9e7fdf71eb/src/Security/Authentication/Cookies/src/CookieAuthenticationOptions.cs#L32) and it doesn't look like you have.

You should be able to test this pretty easily by setting a short expiration and watching it refresh after half the expiration window has elapsed.

kvenkatrajan (approved):

davidfowl (approved):

JamesNK:
@halter73

> I'm not sure if there's any UI behind this to make it easier with something like a checkbox, but you could consider setting IsPersistent based on a query string parameter like ?rememberMe=true.

There is UI but I can't think of a reason we wouldn't want a persistent cookie right now.

> You should be able to test this pretty easily by setting a short expiration and watching it refresh after half the expiration window has elapsed.

👍 

> Out of curiosity, is there a reason you chose three days instead of the default a 14?

People wanted a number. We chose one day to be more secure(?). I thought it was too short, so we increased it to 3. We'll see if there is feedback about logging in often. Maybe we'll increase it in the future.

davidfowl:
backport?

JamesNK:
I don't think it meets the bar.

PR #3667, opened 2024-04-12
Update to 8.0.4 of ASP.NET Core

eerhardt:
- Fix Redis output and distributed caching tests in response to new tracing events emitted from https://github.com/dotnet/aspnetcore/pull/54239

FYI - @mgravell 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3667)

### DIFF
--- a/tests/Aspire.Components.Common.Tests/ActivityNotifier.cs
+++ b/tests/Aspire.Components.Common.Tests/ActivityNotifier.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading.Channels;
 using OpenTelemetry;
 
 namespace Aspire.Components.Common.Tests;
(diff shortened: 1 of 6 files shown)


tests/Aspire.Components.Common.Tests/ActivityNotifier.cs
```diff
@@ -11,19 +13,44 @@ namespace Aspire.Components.Common.Tests;
...
+    }
 
     public override void OnEnd(Activity data)
     {
-        ExportedActivities.Add(data);
-        _taskSource.SetResult();
+        _activityChannel.Writer.TryWrite(data);
+    }
+
+    private async IAsyncEnumerable<Activity> WaitAsync([EnumeratorCancellation] CancellationToken cancellationToken)
```
sebastienros, line 39: Why not have `_activityChannel.Reader.ReadAllAsync(cancellationToken).ConfigureAwait(false)` directly in TakeAsync?

sebastienros (approved):

danmoseley (approved):
Didn't review the tests, but yes presumably we need this in release/8.0

Can you link the bug for the underlying issue?

eerhardt:
> Can you link the bug for the underlying issue?

Not sure what you mean. https://github.com/dotnet/aspnetcore/pull/54239 is the PR that changed the behavior. We needed to react to it in our tests.

mgravell:
To confirm: yes, the underlying behaviour change has been backported to 8.0 (to provide better connection-level metadata), so it would be good for this to do likewise.

tests/Aspire.Components.Common.Tests/ActivityNotifier.cs
eerhardt: At first I only had `WaitAsync` public on this class. But when I used it in 4 different places, all consumers needed was to get a collection of the activities and assert on them. So I made the helper `TakeAsync` method and made the `WaitAsync` method private.

Issue #54, opened 2023-10-04
Write JSON schema for Aspire manifest format.


mitchdenny:
We will be using the ```$schema``` attribute to specify the version of the manifest format we are using. That means we need a JSON schema which documents the format. We also need to think about where we host it.

danmoseley:
@mitchdenny what remains here after your PR #3684 ?

mitchdenny:
Just update the manifest writing code to emit $schema. I'll put up a PR. It's a nice to have.

PR #3684, opened 2024-04-14
JSON Schema for manifest.

mitchdenny:
Addresses #54

This is still a work in progress. Building out the validation rules in the schema and adding test cases to cover it. I think we are going to have to use `Newtonsoft.Json.Schema` for the unit tests for now.

Link to schema file: https://github.com/dotnet/aspire/blob/mitchdenny/schema/src/Schema/aspire-8.0.json

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3684)


davidfowl:
How do you expresss the extensibility of the types?

mitchdenny:
Good question. Today that app model allows people to define their own resource types in the manifest. So technically `"myresource.that.no.one.understands.v0"` is technically fine - and other than verifying that it has a value in the `"type"` field no further validation could be done.

I've updated the `oneOf` clause to include a generic resource type that includes any object that has a `type` field as a fallback. However if the resource is captured by one of the earlier conditions with a defined resource type it still validates fine (the test cases cover this).

davidfowl (approved):

Directory.Packages.props
```diff
@@ -158,6 +158,7 @@
     <PackageVersion Include="Microsoft.Orleans.Sdk" Version="8.1.0-nightly.20240307.1" />
     <!-- Pinned version for Component Governance - Remove when root dependencies are updated -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageVersion Include="Newtonsoft.Json.Schema" Version="3.0.15" />
```
eerhardt, line 161: @mitchdenny - why did we choose to `Newtonsoft.Json.Schema` when we are already using `JsonSchema.Net` in this repo?

tests/Aspire.Hosting.Tests/Schema/SchemaTests.cs
```diff
@@ -0,0 +1,296 @@
...
+            var relativePath = Path.Combine("Schema", "aspire-8.0.json");
+            var schemaPath = Path.GetFullPath(relativePath);
+            var schemaText = await File.ReadAllTextAsync(schemaPath);
+            s_schema = JSchema.Parse(schemaText);
+        }
+
+        return s_schema;
+    }
+
+    [SkipOnHelixTheory]
```
eerhardt, line 154: We shouldn't need to skip these tests on Helix. We are doing the same thing in the Components tests, and we can make it work by doing this:

https://github.com/dotnet/aspire/blob/b0e3578fc56cd45cd41897a66a446f2c341c9726/tests/Aspire.Azure.AI.OpenAI.Tests/Aspire.Azure.AI.OpenAI.Tests.csproj#L8

Issue #3199, opened 2024-03-27
Add integration tests for the dashboard's OpenIdConnect auth


drewnoakes:
Following #3033, we need integration test coverage for the OpenIdConnect auth.

This comment from @halter73 will be helpful: https://github.com/dotnet/aspire/pull/3033#issuecomment-2016893891

PR #3692, opened 2024-04-15
Add integration test for OpenIDConnect

drewnoakes:
Contributes to #3199

This test ensures that the server redirects an unauthenticated user to the OpenID authority correctly.

Thanks to @halter73 for [pointing towards a technique](https://github.com/dotnet/aspire/pull/3033#issuecomment-2016893891) for mocking an OpenID authority for testing purposes.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3692)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -62,8 +62,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
     /// Create a new instance of the <see cref="DashboardWebApplication"/> class.
     /// </summary>
     /// <param name="configureBuilder">Configuration the internal app builder. This is for unit testing.</param>
-    /// <param name="requireHttpsMetadataForOpenIdConnect">Allows the OpenIdConnect infrastructure to work without HTTPS, for unit testing purposes only.</param>
-    public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder = null, bool requireHttpsMetadataForOpenIdConnect = true)
+    public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder = null)
     {
         var builder = WebApplication.CreateBuilder();
 
@@ -117,7 +116,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
             builder.Services.Configure<HttpsRedirectionOptions>(options => options.HttpsPort = browserHttpsPort);
         }
 
-        ConfigureAuthentication(builder, dashboardOptions, requireHttpsMetadataForOpenIdConnect);
+        ConfigureAuthentication(builder, dashboardOptions);
 
         // Add services to the container.
         builder.Services.AddRazorComponents().AddInteractiveServerComponents();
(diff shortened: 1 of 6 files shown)


drewnoakes:
The test is failing on Linux:

> IDX20803: Unable to obtain configuration from: 'https://localhost:38813/.well-known/openid-configuration'.

This doesn't happen locally. Pushing a change to capture more information in logs.

davidfowl:
This isn't trying to do oidc on linux with https right?

tests/Aspire.Dashboard.Tests/Integration/FrontendOpenIdConnectAuthTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        await using var app = IntegrationTestHelpers.CreateDashboardWebApplication(
+            testOutputHelper,
+            additionalConfiguration: config =>
+            {
+                // Configure the resource service, as otherwise HTTP requests are redirected to /structuredlogs before OIDC
+                config[DashboardConfigNames.ResourceServiceClientAuthModeName.ConfigKey] = "Unsecured";
+                config[DashboardConfigNames.ResourceServiceUrlName.ConfigKey] = "https://localhost:1234"; // won't actually exist
+
+                // Configure OIDC. We must use an actual IdP because the provider will query "/.well-known/openid-configuration"
+                // at the authority.
```
drewnoakes, line 126:
```suggestion
                // Configure OIDC. It (the RP) will communicate with the mock authority (IdP) spun up for this test.
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -513,6 +514,10 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
 
                     // Avoid "message.State is null or empty" due to use of CallbackPath above.
                     options.SkipUnrecognizedRequests = true;
+
+                    // Allow the requirement of HTTPS communication with the OpenIdConnect authority to be
+                    // relaxed, for unit testing purposes only.
+                    options.RequireHttpsMetadata = requireHttpsMetadataForOpenIdConnect;
```
tlmii, line 520: Does this prevent the user from being able to configure this in appSettings? I guess theoretically that applies to anything above as well. I just noticed this one in particular because it was something I had to set in config when testing OIDC with keycloak

tlmii: I guess a follow-up question would then be, can this be set via config (e.g. around like 31 of `tests/Aspire.Dashboard.Tests/Integration/FrontendOpenIdConnectAuthTests.cs`) and then not need to add this testing-only parameter?

danmoseley:
As with other tests, nice to have in release branch

kvenkatrajan (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: Thank you, this was a very helpful suggestion. Removing test-only code is a nice improvement here.

PR #3708, opened 2024-04-15
Fix environment variable masking being reset on component re-render

### LINKED ISSUE #3693: [AzureTools][Aspire] The environment variable hides the value after dragging the border line at the top of the detail window with the mouse
**VS Version:** VS 17.10.0 Preview 4.0 [34811.107.d17.10]
**Aspire Version:** 8.0.0-preview.5.24201.12
Apply NuGet Feeds
**Podman Desktop:** 1.9.1
**Podman CLI:** 5.0.1

**REPRO STEPS:**
1. Create an Aspire starter project.
2. F5.
3. In the dashboard, click "View" under the "Details" for the project -> Click "Show value" next to an environment variable -> The environment variable shows the value.
4. Drag the border line at the top of the detail window with the mouse.

**Expect:**
The environment variable still shows the value.

**Actual:**
The environment variable hides the value.
Actua

adamint:
Fixes #3693

We were setting the value of environment variable masks `OnParametersSet` to the view's default mask value, which is called after its parent pane's size changes and then re-rendered. This is only needed 1) on initial component render, to set masking to the default value, and 2) when the user toggles that default, which is already handled `ToggleMaskState`.

pre:
environmentvariablesmaskpre

post:
environmentvariablesmaskpost

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3708)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
@@ -106,7 +106,6 @@ public partial class ResourceDetails
 
     private void ResetResourceEnvironmentVariableMasks()
     {
-        _areEnvironmentVariablesMasked = true; // By default, mask environment variables
         foreach (var vm in Resource.Environment.Where(vm => vm.IsValueMasked != _areEnvironmentVariablesMasked))
         {
             vm.IsValueMasked = _areEnvironmentVariablesMasked;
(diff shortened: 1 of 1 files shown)


kvenkatrajan (approved):

JamesNK (changes requested):
This doesn't work when switching between resources because code in OnAfterRender with firstRender check runs once.

Setting the value in `OnParametersSet` is the right place, but it should only happen if the `Resource` parameter changes. Add a private field that is set when it changes, then only update IsValueMasked when the private field and publich property are different.

adamint:
I’m extremely confused why a different component is not rendered for each resource. Thanks, I’ll make that change

JamesNK:
It is re-rendered when the resource changes, but you've placed the code inside firstRender block.

JamesNK (changes requested):

JamesNK:
I don't understand why you're setting this in initialize AND set parameters AND after render. Does just doing this in OnParametersSet like I suggested not work?

adamint:
You're right, sorry.

JamesNK (approved):

Issue #237, opened 2023-09-10
Security Review for Components


danmoseley:

joperezr:
Waiting on sign-off @GrabYourPitchforks

eerhardt:
The work this is tracking is now complete. Closing.

PR #3729, opened 2024-04-16
Move comments on generated password entropy to public API

eerhardt:
This allows for these comments to be seen by callers of the API so they can make educated decisions on what parameters should be set for the password generation.

Contributes to #237
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3729)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ParameterDefault.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ParameterDefault.cs
@@ -30,22 +30,35 @@ public abstract class ParameterDefault
 /// <remarks>
 /// The recommended minimum bits of entropy for a generated password is 128 bits.
 ///
+/// <para>
 /// The general calculation of bits of entropy is:
+/// </para>
 ///
-/// log base 2 (numberPossibleOutputs)
+/// <c>log base 2 (numberPossibleOutputs)</c>
 ///
+/// <para>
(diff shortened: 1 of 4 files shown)


sebastienros (approved):

radical (approved):

src/Aspire.Hosting/ApplicationModel/ParameterDefault.cs
```diff
@@ -27,6 +27,39 @@ public abstract class ParameterDefault
 /// <summary>
 /// Represents that a default value should be generated.
 /// </summary>
+/// <remarks>
+/// The recommended minimum bits of entropy for a generated password is 128 bits.
```
JamesNK, line 31: I'm pretty sure Visual Studio and other tools don't care about whitespace. This content in a tooltip will be one blob of text.

If you want paragraphs, which IMO we should have, then you must add <para> tags.

JamesNK (dismissed):

src/Aspire.Hosting/ApplicationModel/ParameterDefault.cs
JamesNK: Current tooltip:

image

eerhardt: The tooltip now looks like:
image

JamesNK (approved):

PR #3736, opened 2024-04-16
API changed for components

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3736)


danmoseley (approved):
I've no problem with taking for 8.0.

sebastienros:
All the suggested changes were applied. And all tests are currently passing. Lots of little changes though so I hope I haven't missed something.

src/Components/Aspire.Azure.AI.OpenAI/PublicAPI.Unshipped.txt
```diff
@@ -7,6 +7,8 @@ Aspire.Azure.AI.OpenAI.AzureOpenAISettings.Endpoint.get -> System.Uri?
 Aspire.Azure.AI.OpenAI.AzureOpenAISettings.Endpoint.set -> void
 Aspire.Azure.AI.OpenAI.AzureOpenAISettings.Key.get -> string?
 Aspire.Azure.AI.OpenAI.AzureOpenAISettings.Key.set -> void
+Aspire.Azure.AI.OpenAI.AzureOpenAISettings.DisableTracing.get -> bool
+Aspire.Azure.AI.OpenAI.AzureOpenAISettings.DisableTracing.set -> void
 Aspire.Azure.AI.OpenAI.AzureOpenAISettings.Tracing.get -> bool
```
mitchdenny, line 12: Do we need to remove the old APIs from Unshipped across all of these changes?

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
```diff
@@ -43,10 +43,19 @@ public sealed class AzureOpenAISettings : IConnectionStringSettings
     /// Gets or sets a boolean value that indicates whether the OpenTelemetry tracing is enabled or not.
     /// </summary>
     /// <value>
-    /// The default value is <see langword="true"/>.
+    /// The default value is <see langword="true" />.
     /// </value>
+    [Obsolete($"This property is obsolete and will be removed in a future version. Use {nameof(DisableTracing)} instead.")]
```
eerhardt, line 48: I was chatting wtih @davidfowl about this. What if we just renamed the APIs without Obsoleting the old one? We haven't shipped stable yet. Adding a bunch of Obsolete APIs in our final stable build doesn't seem worth it.  We were preview before - we haven't guaranteed we won't break things. Once we ship 8.0.0 we will need to think about API breaking changes. But I don't think we should have to right now. It's more work than necessary IMO.

cc @DamianEdwards

src/Components/Aspire.Seq/SeqSettings.cs
```diff
@@ -11,10 +11,22 @@ namespace Aspire.Seq;
...
+    /// Gets or sets a boolean value that indicates whether the Seq server health check is enabled or not.
     /// </summary>
+    /// <value>
+    /// The default value is <see langword="true"/>.
+    /// </value>
+    [Obsolete($"This property is obsolete and will be removed in a future version. Use {nameof(DisableHealthChecks)} instead.")]
     public bool HealthChecks { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets a boolean value that indicates whether the Seq server health check is enabled or not.
```
eerhardt, line 23:
```suggestion
    /// Gets or sets a boolean value that indicates whether the Seq server health check is disabled or not.
```

tests/Aspire.MongoDB.Driver.Tests/AspireMongoDBDriverExtensionsTests.cs
```diff
@@ -95,7 +95,7 @@ public async Task AddMongoDBDataSource_HealthCheckShouldBeRegisteredWhenEnabled(
 
         builder.AddMongoDBClient(DefaultConnectionName, settings =>
         {
-            settings.HealthChecks = true;
+            settings.DisableHealthChecks = !true;
```
eerhardt, line 98: Let's not do this.

```suggestion
            settings.DisableHealthChecks = false;
```

tests/Shared/Logging/TestLoggerFactory.cs
```diff
@@ -14,9 +14,9 @@ public TestLoggerFactory(ITestSink sink, bool enabled)
         _enabled = enabled;
     }
 
-    public ILogger CreateLogger(string name)
+    public ILogger CreateLogger(string categoryName)
```
eerhardt, line 17: This change doesn't seem necessary.

src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
```diff
@@ -27,7 +27,7 @@ public sealed class AzureMessagingServiceBusSettings : IConnectionStringSettings
     /// <remarks>
     /// Used along with <see cref="Credential"/> to establish the connection.
     /// </remarks>
-    public string? Namespace { get; set; }
+    public string? FullyQualifiedNamespace { get; set; }
```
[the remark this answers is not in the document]
sebastienros: > Was there a reason

My brain that didn't think about this one

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
```diff
@@ -43,10 +43,19 @@ public sealed class AzureOpenAISettings : IConnectionStringSettings
...
     public bool Tracing { get; set; } = true;
 
+    /// <summary>
+    /// Gets or sets a boolean value that indicates whether the OpenTelemetry tracing is disabled or not.
+    /// </summary>
+    /// <value>
+    /// The default value is <see langword="false"/>.
+    /// </value>
+    public bool DisableTracing { get; set; }
```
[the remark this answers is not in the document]
sebastienros: Yes, that would be better, it's obsolete that would still work.

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
sebastienros: We just need to decide about this. Let me know if you prefer to remove all the [Osbsolete] properties. I think @joperezr is also fine with that option.

eerhardt: My vote is to not make things Obsolete. We will do that going forward if we need to break some already "stable" API.

davidfowl: Agree

PR #3755, opened 2024-04-17
[release/8.0] AppModel API omnibus PR for GA

mitchdenny:
### Customer Impact

Overload of WithEnvironment had the wrong constraint. Customers may assume that adding an environment variable these resources has an impact. It would also set them up for a break when we fix this in 8.1 so fixing it in 8.0 since it is a minor low risk change right now.

### Testing

Code itself hasn't changed, just the constraint. Everything compiles including playgrounds.

### Risk

Low.

### Regression?

No.

### Customer Impact

This is a breaking change. We removed the data directory argument on `AddSeq` and instead introduced
two extension methods named `WithDataBindMount` and `WithDataVolume` to be consistent with other container resources. We could not `[Obsolete]` the old method because it has an optional argument and it would introduce ambiguity to the compiler - so we had to remove it outright with this change.

### Testing

Manual testing and unit testing.

### Risk

Low.

### Regression?

No.

### Customer Impact

modifies the arguments for `WithPgAdmin` to take a callback which allows for a greater array of customization. Usage example:

```csharp
builder.AddPostgres("pgslq").WithPgAdmin(c => c.WithImageTag("8.3").WithHostPort(8999));
```

This is a breaking change to the WithPgAdmin API. If someone was using the port argument they'll now need to use the callback syntax. The benefit is that they can do more customizations to the container. Until this change we didn't provide a way for people to bump their container image.

### Testing

Manual testing and unit testing (added a test).

### Risk

Low.

### Regression?

No.

### Customer Impact

Currently there is an extension method for `IResourceBuilder<NatsServerResource>` called `WithJetStream(...)`. With JetStream automatically creates a bind mount when it enables the
JetStream feature on Nats. This change separates that out so that it is consistent with other resources so we have a `WithDataBindMount` and `WithDataVolume` extension methods.

### Testing

Manual testing. Unit test updated to accommodate the change.

### Risk

Low.

### Regression?

No.

### Customer Impact

Breaking changes to the API for WithRedisCommander and WithMongoExpress. Basically adding the container configuration callback the same as was done to Postgres above. Also introduced the WithHostPort methods.

The PR also renames the Cosmos and Storage UseX methods to WithX methods for overall consistency.

### Testing

Manual testing. Unit tests

### Risk

Low.

### Regression?

No.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3755)


danmoseley:
Supportive of these so far. If it would help @mitchdenny we can merge these three now (if reviewed)

danmoseley:
@davidfowl OK with these?

mitchdenny:
I'm OK to merge them now. Better to have them in and being tested by Bala's team than sitting here doing nothing.

mitchdenny:
@danmoseley this is ready now.

RussKie (approved):

danmoseley:
@davidfowl  supportive? If so I'm good with it

mitchdenny:
@davidfowl @danmoseley as discussed, added on the WithMongoExpress and WithRedisCommander container config callbacks as well as a few other tweaks for Cosmos and Storage. These are breaking API changes but these are less frequently used APIs and necessary for overall polish/consistency.

davidfowl (approved):

PR #3758, opened 2024-04-17
Remove generic WithAnnotation.

mitchdenny:
Change to WithAnnotation due to ApiDesign warnings. In hindsight we don't need this to be generic, it doesn't add a lot of value. We almost always want to pass the annotation in.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3758)


src/Aspire.Hosting/DistributedApplicationResourceBuilder.cs
```diff
@@ -24,7 +24,7 @@ internal sealed class DistributedApplicationResourceBuilder<T>(IDistributedAppli
 
         // If the behavior is AddReplace then there should never be more than one annotation present. The following call will result in an exception which
         // allows us to easily spot these bugs.
-        if (behavior == ResourceAnnotationMutationBehavior.Replace && Resource.Annotations.OfType<TAnnotation>().SingleOrDefault() is { } existingAnnotation)
+        if (behavior == ResourceAnnotationMutationBehavior.Replace && Resource.Annotations.SingleOrDefault(a => a.GetType() == annotation.GetType()) is { } existingAnnotation)
```
davidfowl, line 27: This behavior is different.

mitchdenny: Yeah .. will fix this.

mitchdenny: Tested added (found a bug actually). Arguably this API could be tightened up even further by making it that you can only add annotations that are sealed types, to avoid the ambiguity that inheritance introduces here.

src/Aspire.Hosting/DistributedApplicationResourceBuilder.cs
```diff
@@ -24,12 +24,19 @@ internal sealed class DistributedApplicationResourceBuilder<T>(IDistributedAppli
 
         // If the behavior is AddReplace then there should never be more than one annotation present. The following call will result in an exception which
         // allows us to easily spot these bugs.
-        if (behavior == ResourceAnnotationMutationBehavior.Replace && Resource.Annotations.OfType<TAnnotation>().SingleOrDefault() is { } existingAnnotation)
+        if (behavior == ResourceAnnotationMutationBehavior.Replace&& Resource.Annotations.SingleOrDefault(a => IsSameOrSubclassOf(a, annotation)) is { } existingAnnotation)
```
JamesNK, line 27: You can use IsAssignableFrom - https://learn.microsoft.com/en-us/dotnet/api/system.type.isassignablefrom?view=net-8.0

Neither type will ever be an interface so it should have the behavior you want - current class or sub class.

danmoseley:
Is this still proposed for GA, per its milestone?

mitchdenny:
We are going to hold off.

eerhardt:
Should we close this PR? It seems stale. Also, are we going to make API breaking changes after 8.0? I would think we wouldn't.

PR #3763, opened 2024-04-17
Fixed NATS volume mounts

### LINKED ISSUE #3756: [Nats]: WithJetStream method takes a mount path that assumes a bind mount
https://github.com/dotnet/aspire/blob/4829f56034c6ea799743c0550cdbd6c3d64e3ac5/src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs#L36-L47

Seems like we should remove the path from this method and instead follow the pattern we've established for containers that have well known directories for persistence. 

e.g.

https://github.com/dotnet/aspire/blob/4829f56034c6ea799743c0550cdbd6c3d64e3ac5/src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs#L40-L51

cc @mtmk

mtmk:
* resolves #3756 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3763)


mitchdenny:
I did one too #3767.

I know very little about NATS though. I'm assuming that the -js and -sd arguments can be taken in isolation (although I guess -js doesn't make much sense without -sd?).

davidfowl:
😄

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
```diff
@@ -30,19 +31,28 @@ public static IResourceBuilder<NatsServerResource> AddNats(this IDistributedAppl
...
-        if (srcMountPath != null)
-        {
-            args.Add("-sd");
-            args.Add("/data");
-            builder.WithBindMount(srcMountPath, "/data");
-        }
+    /// <param name="builder">The resource builder.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<NatsServerResource> WithJetStream(this IResourceBuilder<NatsServerResource> builder)
+        => builder.WithArgs("-js", "-sd", "/var/lib/nats");
```
davidfowl, line 37: What happens if you call this but you don't call `WithDataVolume`

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
```diff
@@ -30,19 +31,28 @@ public static IResourceBuilder<NatsServerResource> AddNats(this IDistributedAppl
     /// <summary>
     /// Adds JetStream support to the NATS server resource.
     /// </summary>
-    /// <param name="builder">NATS resource builder.</param>
-    /// <param name="srcMountPath">Optional mount path providing persistence between restarts.</param>
-    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<NatsServerResource> WithJetStream(this IResourceBuilder<NatsServerResource> builder, string? srcMountPath = null)
```
eerhardt, line 36: What's the policy on making API breaking changes right now? Is this supposed to be left in and `[Obsolete]`?

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
eerhardt: Seems like a good test to add.

davidfowl:
I think we should keep this one @mitchdenny can close the other one.

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
mtmk: I think data just stays on the ephemeral container storage. Could you give me some ideas about the test. Happy to add them too.

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
mtmk: good point. marked as obsolete. I would've thought api changes are tolerable when in preview so I'd be happy to remove if you're OK with it.

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
mitchdenny: I think we should split out `WithJetStream` and `WithDataBindMount` and `WithDataVolume`. What I mean by that is that `WithJetStream` won't add the `-sd [path]` argument. Lets make that someting that people can do explicitly.

People might want to enable the JetStream feature set but don't care about persistence.

mitchdenny:
I noticed that NATS doesn't have a port argument. If we were to add the port argument (as an optional) we wouldn't be able to retain the obsolete version of the method because of ambiguity. Given that NATS doesn't have a port today we should probably not add it, leave the obsolete method since there is a workaround with the `WithEndpoint(name, callback)` overload and in Aspire 9.0 remove this method.

@davidfowl seem like a reasonable plan?

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
mtmk: where would you keep path value to be used by `WithDataBindMount` and `WithDataVolume`? Just not sure how to implement that.

davidfowl:
Let’s add it now

mitchdenny:
I think I was confused about the port thing. AddNats has the port argument, I thought it would have the same issue as AddSeq but AddSeq had the persistence path on that extension method instead of a seperate one like WithJetstream.

Anyway ... @mtmk I've updated your PR to split out the -sd option from WithJetstream so that they are controlled independently.

eerhardt (approved):

PR #3769, opened 2024-04-17
Remove seqDataDirectory from public API

eerhardt:
Callers should instead follow the pattern we've established for containers that have well known directories for persistence.

Similar to #3756

cc @liammclennan 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3769)


liammclennan:
See https://github.com/dotnet/aspire/pull/3762

mitchdenny:
@eerhardt we should take my PR instead of this one because I add the data volume/bindmount extensions: #3762

Issue #3770, opened 2024-04-17
Update pgAdmin Image from Version 8.3 to 8.5


AClerbois:
### Summary
This issue proposes upgrading the pgAdmin Docker image used in our project from the current version 8.3 to the latest stable release, version 8.5. This update aims to leverage the latest features and improvements offered by pgAdmin 8.5, enhancing our database management capabilities and ensuring better security and performance.

### Current Situation
We are currently using pgAdmin version 8.3. While this version is stable, it lacks several enhancements and bug fixes that have been introduced in subsequent releases.

### Proposed Change
- **Update the pgAdmin Docker image from version 8.3 to 8.5.**
- Review and adapt our configuration and deployment scripts as necessary to accommodate changes introduced in the new version.

### Benefits of Updating
- **Enhanced Features**: Version 8.5 includes new features and UI improvements that can improve user experience and productivity.
- **Bug Fixes and Security Enhancements**: Updating to the latest version helps address known vulnerabilities and bugs present in our current version.
- **Better Performance**: Improved performance features in version 8.5 can enhance the efficiency of our database management tasks.

### Possible Challenges
- Compatibility issues with existing configurations might require adjustments.
- Testing will be necessary to ensure that the new version integrates smoothly with our current systems.

AClerbois:
@davidfowl I totally agree with your idea

PR #3772, opened 2024-04-17
Add parameter to override tagVersion of pgAdmin

AClerbois:
Linked to the issue #3770
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3772)


mitchdenny:
You can already do this using the following:

```csharp
var pgadmin = builder.AddPostgres("pg")
                     .WithPgAdmin(c => c.WithImageTag("8.5"));
```

Hope this helps.

mitchdenny:
Ah I see that you put this PR up before we fixed this in the API. I'll close this PR as this is the approach we decided to take instead of the extra optional parameter on `WithPgAdmin`.

PR #3776, opened 2024-04-17
Make console log page usable on low-res/high-zoom

adamint:
Currently, the console logs page is not usable for low-res/high-zoom users (e.g. 1280x1024, 400%) that we are required to support.

Before:
before

After:
after

After for 100%, higher res (no change)
image
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3776)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.css
+++ b/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.css
@@ -1,7 +1,7 @@
 ::deep.resource-logs-layout {
     display: grid;
     grid-template-rows: auto auto 1fr;
-    height: 100%;
+    min-height: 100%;
     width: 100%;
     grid-template-areas:
         "head"


src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.css
```diff
@@ -1,7 +1,7 @@
 ::deep.resource-logs-layout {
     display: grid;
     grid-template-rows: auto auto 1fr;
-    height: 100%;
+    min-height: 100%;
```
JamesNK, line 4: What about this?
```suggestion
    height: 100%;
    min-height: 300px;
```

JamesNK, line 4: What about this?
```suggestion
    height: 100%;
    min-height: 300px;
```

JamesNK (changes requested):

PR #3777, opened 2024-04-17
Templates: allow to choose between MSTest, NUnit and xUnit

### LINKED ISSUE #2908: Add support for MSTest to Aspire test project templates
Follow up to #2905

Update the option on the starter template and add another standalone template for MSTest-based tests for an Aspire app.

Evangelink:
Fixes #2908 

cc @pavelhorak 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3777)


src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
```diff
@@ -0,0 +1,78 @@
...
+  "description": "A project that contains MSTest integration tests of a .NET Aspire AppHost project.",
+  "shortName": "aspire-mstest",
+  "sourceName": "Aspire.Tests1",
+  "preferNameDirectory": true,
+  "tags": {
+    "language": "C#",
+    "type": "project"
+  },
+  "precedence": "8000",
+  "identity": "Aspire.Tests.MSTest.CSharp.8.0",
```
Evangelink, line 25: Is it ok to update xunit template identity to match the same pattern? This would avoid putting one of the framework above the others.

Evangelink, line 25: It's not clear to me why with a different identity I have some conflict when trying to run the template:

image

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -84,12 +84,26 @@
       "displayName": "_Use Redis for caching (requires a supported container runtime)",
       "description": "Configures whether to setup the application to use Redis for caching. Requires a supported continer runtime to run locally, see https://aka.ms/dotnet/aspire/containers for more details."
     },
-    "CreateTestsProject": {
+    "TestProjectFrameworkKind": {
```
Evangelink, line 87: Any suggestion for a better name? Maybe `SelectedTestFramework`?

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -84,12 +84,26 @@
...
-      "description": "Configures whether to create a project for integration tests using the AppHost project."
+      "description": "Configures whether to create a project for integration tests using the AppHost project.",
+      "choices": [
+        {
+          "choice": "None",
+          "description": "Do not create a test project."
+        },
+        {
+          "choice": "MSTest",
+          "description": "Create a MSTest test project using the AppHost project."
```
Evangelink, line 100:
```suggestion
          "description": "Creates a MSTest test project using the AppHost project."
```

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -84,12 +84,26 @@
...
+        },
+        {
+          "choice": "xUnit",
+          "description": "Create a xUnit test project using the AppHost project."
```
Evangelink, line 104:
```suggestion
          "description": "Creates a xUnit test project using the AppHost project."
```

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -347,7 +361,11 @@
     },
     {
       "path": "AspireStarterApplication.1.Tests\\AspireStarterApplication.1.Tests.csproj",
-      "condition": "(CreateTestsProject)"
+      "condition": "(TestProjectFrameworkKind == MSTest)"
+    },
+    {
+      "path": "AspireStarterApplication.1.Tests\\AspireStarterApplication.1.Tests.csproj",
+      "condition": "(TestProjectFrameworkKind == xUnit)"
```
Evangelink, line 368:
```suggestion
      "condition": "(TestProjectFrameworkKind != None)"
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
```diff
@@ -1,3 +1,28 @@
+<!--#if (TestProjectFrameworkKind == "MSTest") -->
```
Evangelink, line 1: The 2 projects are relatively different so it seems easier to have 2 configurations rather than many conditions.

Evangelink:
I wasn't able to validate the starter project template as there are many packages that cannot be restored (not from the test project) so I am not sure what I am doing wrong.

The call to `build.cmd -pack` was good.

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
```diff
@@ -0,0 +1,78 @@
...
+  "thirdPartyNotices": "https://aka.ms/dotnet/aspire/8.0-third-party-notices",
+  "groupIdentity": "Aspire.Tests",
```
Evangelink, line 27:
```suggestion
  "groupIdentity": "Aspire.Tests.MSTest",
```

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
Evangelink: Solved thanks to @baronfel (group identity needs to differ too).

DamianEdwards: What do you mean by "putting one of the framework above the others"?

@phenning what are the implications of changing a template identity?

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
DamianEdwards: It doesn't really matter, this isn't visible to users. Just make sure it makes sense for us the authors, e.g. `TestFramework`.

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
DamianEdwards: The other option is to completely separate them (different directories) and then conditionally exclude them from the template based on the selected framework. That avoids any in-file conditions like this at all.

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
```diff
@@ -1,3 +1,28 @@
...
+<Project Sdk="MSTest.Sdk/3.3.1">
```
DamianEdwards, line 2: This is very unusual to see for me. I'm not aware of any templates we ship today that use a versioned SDK like this. Are we fully aware of the user experience implications of this?
/Cc @mhutch @dsplaisted

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -84,12 +84,26 @@
...
+          "choice": "None",
+          "description": "Do not create a test project."
+        },
+        {
+          "choice": "MSTest",
+          "description": "Creates a MSTest test project using the AppHost project."
+        },
+        {
+          "choice": "xUnit",
+          "description": "Creates a xUnit test project using the AppHost project."
```
DamianEdwards, line 104: I think the official name is "xUnit.net" so suggest we go with:
```suggestion
          "choice": "xUnit.net",
          "description": "Creates an xUnit.net test project using the AppHost project."
```

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -84,12 +84,26 @@
...
-      "description": "Configures whether to create a project for integration tests using the AppHost project."
+      "description": "Configures whether to create a project for integration tests using the AppHost project.",
+      "choices": [
+        {
+          "choice": "None",
+          "description": "Do not create a test project."
+        },
+        {
+          "choice": "MSTest",
+          "description": "Creates a MSTest test project using the AppHost project."
```
DamianEdwards, line 100:
```suggestion
          "description": "Creates an MSTest test project using the AppHost project."
```

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/dotnetcli.host.json
```diff
@@ -42,13 +42,13 @@
         "longName": "use-redis-cache",
         "shortName": ""
       },
-      "CreateTestsProject": {
-        "longName": "create-tests-project",
+      "TestProjectFrameworkKind": {
+        "longName": "test-project-framework",
```
DamianEdwards, line 46:
```suggestion
        "longName": "test-framework",
```

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/dotnetcli.host.json
```diff
@@ -42,13 +42,13 @@
...
         "shortName": "t"
       }
     },
     "usageExamples": [
       "--use-redis-cache",
-      "--create-tests-project"
+      "--test-project-framework MSTest"
```
DamianEdwards, line 52:
```suggestion
      "--test-framework MSTest"
```

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
phenning: The only impact would be if there could be two versions of the same template installed that had different identities. 

This would likely only happen if the templates were installed by different mechanism, for example, if one version was installed via an optional workload, and a second with the same group id but different identity (but identical targeted frameworks) was installed either as a result of being in the core SDK or was installed via dotnet new install. 

In that case, we would use the one with the highest precedence when that TFM was choosen. So in order to get the "newest" one, the best thing to do would be to update the precedence to say 8001 when changing identities.  

If the group identity changes, that is effectively an entirely different template, so when there were multiples installed after a rename, there would be duplicate templates in NPD.

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
RussKie: That's fresh off the press https://devblogs.microsoft.com/dotnet/introducing-mstest-sdk/

DamianEdwards: Sure but I'm more speaking to the implications of having a version in the Sdk identifier and relying on the NuGet SDK resolver. Those have existed for years now but not in wide use as we don't ship templates using them. Are there any interactions that change the user experience because of that change, e.g. what happens if I'm using a custom NuGet config that doesn't include nuget.org (as per our secure supply chain guidelines)? Does project load fail in a nice way in Visual Studio if the SDK NuGet package cannot be restored? Putting this in the Aspire workload means that many more users will be exposed to these behaviors so I want to be very sure we know what the experience is and are happy with it.

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
Evangelink: I wasn't sure how to avoid leaking the framework name in the project name. Whatever the framework I use, I expect the same project name (usually product.Tests or product.IntegrationTests...) which I couldn't figure with the conditional inclusion/exclusion.

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
thimok: > You are obviously better at English than me, I'm just wondering isn't the rule to have 'a' before consonant and 'an' before voyelle?

I believe that's generally correct, but 'x' is pronounced as 'ex', so that would result in 'an'. I'm not native English however, so I might be mistaken about this exception.

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
Evangelink: > What do you mean by "putting one of the framework above the others"?

We have as identity:

- xUnit: `Aspire.Tests.CSharp.8.0`
- MSTest: `Aspire.Tests.MSTest.CSharp.8.0`
- NUnit: `Aspire.Tests.NUnit.CSharp.8.0`

Which reads to me as we promote mainly xUnit and the others are supported as second class citizen. While with a rename like:

- xUnit: `Aspire.Tests.Xunit.CSharp.8.0`
- MSTest: `Aspire.Tests.MSTest.CSharp.8.0`
- NUnit: `Aspire.Tests.NUnit.CSharp.8.0`

We have the different test frameworks on the same "level"

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
Evangelink: @baronfel What's your view about risks?

@DamianEdwards We have an option without the SDK but if we can clear out any risk, we would prefer to push the SDK forward as it simplifies many things. For example, for v3.4, we can introduce `EnableAspire` property that would add the using of `Aspire.Hosting.Testing` and the package reference `Aspire.Hosting.Testing`.

MarcoRossignoli: > Are there any interactions that change the user experience because of that change, e.g. what happens if I'm using a custom NuGet config that doesn't include nuget.org (as per our secure supply chain guidelines)?

What you mean? The test framework packages and the test sdk comes from nuget, the same issue should happen with the normal configuration. The SDK is only a way to simplify users setup.

mhutch: The problem with resolving MSBuild SDKs from NuGet is that they cause project evaluation to block on resolving the package from NuGet, and AFAIK Visual Studio does not handle this gracefully, so it can cause the UI to hang.

MarcoRossignoli: I had an offline sync with @rainersigwald and VS at the moment does have some suboptimal usage of the MSBuild sdk in some scenario, if we will have users complain about that we will work together with the MSBuild/VS team for improve the situation.

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
DamianEdwards: @thimok is correct. The rule is actually based on the sound of the following word, not the actual letters used: https://www.merriam-webster.com/grammar/is-it-a-or-an

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
DamianEdwards: > if we will have users complain about that we will work together with the MSBuild/VS team to improve the situation.

While I think this is a fine stance for the MSTest team to take, I'm not comfortable right now having the Aspire product be the "guinea pig" to flesh out the experience issues so I'd prefer we stick to the standard Test SDK for these templates. We can revisit this for the next major version of Aspire (releasing with .NET 9).

DamianEdwards: @Evangelink 
> For example, for v3.4, we can introduce EnableAspire property that would add the using of Aspire.Hosting.Testing and the package reference Aspire.Hosting.Testing.

Auto-adding package references from SDKs is not something we typically like doing, as it hides the version from customers and in this case would mean your SDK has to keep track of Aspire package versions which seems like a bad dependency to create. This is why the Aspire workload SDK doesn't auto-add the `Aspire.Hosting.AppHost` package to AppHost projects (those with `<IsAspireHost>true</IsAspireHost>` set in the csproj).

Evangelink: > While I think this is a fine stance for the MSTest team to take, I'm not comfortable right now having the Aspire product be the "guinea pig" to flesh out the experience issues so I'd prefer we stick to the standard Test SDK for these templates. We can revisit this for the next major version of Aspire (releasing with .NET 9).

Fine by me, I'll update the PR.

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
Evangelink: Ping @DamianEdwards

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
Evangelink: Resolving this thread, I am now using MSTest metapackage and no longer the SDK package.

src/Aspire.ProjectTemplates/templates/aspire-mstest/.template.config/template.json
DamianEdwards: I see. These identities aren't actually visible to the user but making them consistent makes sense if there's no downside. I think we actually want to update the group identity too so this will effectively EOL the current xUnit template when we release this change, but I don't think that's a problem.

Evangelink: I'll do the change as part of this PR but please let me know if you would prefer to have this one merged and another one for the xUnit template change.

Also, shall I go ahead and also provide NUnit?

Evangelink: Done for both points.

cc @OsirisTerje

DamianEdwards:
@Evangelink FYI I'm going to pull down your fork and verify the package changes locally before approving

DamianEdwards:
Templates install from locally built package OK:

image

And they show up VS too:

image

Verifying what @phenning said that when we change template groupid for the xunit template, you end up with duplicate templates for xUnit if you have the template from the workload (old groupid) and template from the package (new groupid) both installed. This shouldn't be a concern in the real-world for customers though as installing the templates from a NuGet package isn't a mainstream scenario and the 8.0 and 8.1 workloads won't be side-by-side either (IIUC).

Option for test framework to use on the Starter template shows up in VS OK:

image

And the CLI too:

image

New Starter project with MSTest project works in VS:

image

Works from the CLI too via `dotnet test`:

image

For NUnit it seems we have a missing `using`:

image

But I think we actually want to change to NUnit's new assertion pattern:

image

xUnit option is missing the required `[Fact]` and `Assert` lines completely (and the `PackageReference` in the .csproj file):

image

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
```diff
@@ -8,12 +8,21 @@
...
-    <PackageReference Include="xunit" Version="2.5.3" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" />
+    <PackageReference Include="coverlet.collector" Version="6.0.0" Condition=" $(TestFramework) == 'xUnit' " />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" Condition=" $(TestFramework) != 'MSTest' " />
+    <PackageReference Include="MSTest" Version="3.1.1" Condition=" $(TestFramework) == 'MSTest' " />
+    <PackageReference Include="NUnit" Version="4.0.1" Condition=" $(TestFramework) == 'NUnit' " />
+    <PackageReference Include="NUnit.Analyzers" Version="4.0.1" Condition=" $(TestFramework) == 'NUnit' " />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" Condition=" $(TestFramework) == 'NUnit' " />
+    <PackageReference Include="xunit" Version="2.5.3" Condition=" $(TestFramework) == 'xUnit' " />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" Condition=" $(TestFramework) == 'xUnit' " />
```
DamianEdwards, line 25:
```suggestion
    <PackageReference Include="xunit" Version="2.5.3" Condition=" $(TestFramework) == 'xUnit.net' " />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" Condition=" $(TestFramework) == 'xUnit.net' " />
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
```diff
@@ -8,12 +8,21 @@
...
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Aspire.Hosting.Testing" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
-    <PackageReference Include="coverlet.collector" Version="6.0.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
-    <PackageReference Include="xunit" Version="2.5.3" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" />
+    <PackageReference Include="coverlet.collector" Version="6.0.0" Condition=" $(TestFramework) == 'xUnit' " />
```
DamianEdwards, line 18:
```suggestion
    <PackageReference Include="coverlet.collector" Version="6.0.0" Condition=" $(TestFramework) == 'xUnit.net' " />
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
```diff
@@ -2,9 +2,18 @@
...
+#if (TestFramework == "MSTest")
+[TestClass]
+#endif
 public class WebTests
 {
+#if (TestFramework == "MSTest")
+    [TestMethod]
+#elif (TestFramework == "NUnit")
+    [Test]
+#elif (TestFramework == "xUnit")
```
DamianEdwards, line 14:
```suggestion
#elif (TestFramework == "xUnit.net")
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
```diff
@@ -17,6 +26,12 @@ public async Task GetWebResourceRootReturnsOkStatusCode()
         var response = await httpClient.GetAsync("/");
 
         // Assert
+#if (TestFramework == "MSTest")
+        Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+#elif (TestFramework == "NUnit")
+        ClassicAssert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+#elif (TestFramework == "xUnit")
```
DamianEdwards, line 33:
```suggestion
#elif (TestFramework == "xUnit.net")
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
```diff
@@ -17,6 +26,12 @@ public async Task GetWebResourceRootReturnsOkStatusCode()
         var response = await httpClient.GetAsync("/");
 
         // Assert
+#if (TestFramework == "MSTest")
+        Assert.AreEqual(HttpStatusCode.OK, response.StatusCode);
+#elif (TestFramework == "NUnit")
+        ClassicAssert.AreEqual(HttpStatusCode.OK, response.StatusCode);
```
DamianEdwards, line 32:
```suggestion
        Assert.That(response.StatusCode, Is.EqualTo(HttpStatusCode.OK));
```

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/AspireStarterApplication.1.Tests.csproj
```diff
@@ -22,7 +31,9 @@
 
   <ItemGroup>
     <Using Include="Aspire.Hosting.Testing" />
-    <Using Include="Xunit" />
+    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" Condition=" $(TestFramework) == 'MSTest' " />
+    <Using Include="NUnit.Framework" Condition=" $(TestFramework) == 'NUnit' " />
+    <Using Include="Xunit" Condition=" $(TestFramework) == 'xUnit' " />
```
DamianEdwards, line 36:
```suggestion
    <Using Include="Xunit" Condition=" $(TestFramework) == 'xUnit.net' " />
```

DamianEdwards (changes requested):
These changes should fix the issues with the NUnit and xUnit options in the Starter template.

phenning:
[like]  Phil Henning reacted to your message:
________________________________

Evangelink:
> These changes should fix the issues with the NUnit and xUnit options in the Starter template.

Thanks @DamianEdwards! I totally forgot to redo the manual tests after the last few commits... Sorry and glad you tested it.

DamianEdwards (approved):
Thanks again for this!

PR #3785, opened 2024-04-17
[release/8.0] Start producing stable packages

joperezr:
As part of our preparation for stabilization, we need to start producing stable packages from this branch. This change will make it so that all packages that are not marked as `SuppressFinalPackageVersion` will now produce a stable package version. This will also make it so we have CI validation to ensure that packages that are going stable will not get a new dependency bump with a prerelease version.

This change will have implications going forward, which mainly are:
- Because we are using stable versions, packages from this branch will no longer get published to the dotnet-tools feed, and instead will get a dedicated feed for each build. This means that our installLatestFromREleaseBranch scripts will no longer work for getting latest build, so I'll work with @balachir for new instructions on how to get latest builds for validation.
- For people trying things out locally, you should be aware that because these are stable versions, it means that these would be saved in your machine's global cache. This means that before you can use newer builds that produce the same version, you'll need to make sure you clean up your cache first by running `dotnet nuget locals all -c`, in order to prevent the previous build from the cache to get used.

cc: @eerhardt, @davidfowl, @DamianEdwards, @mitchdenny, @ReubenBond.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3785)


joperezr:
Marking as No-Merge for now given the dogfooding implications. We'll merge this a little later.

joperezr:
This is now ready to go, so removing the NoMerge label.

DamianEdwards:
image

joperezr:
cc @radical if you check my last commit 24a05b8, I had to make a small change given the workload test infrastructure had a bug when testing against stable versions. My change was simple to unblock this PR, but we'll probably need something that actually looks at what the package version for each package should be for main branch.

joperezr:
Merging this in to start preparing for an insertion. I will Address any feedback in a follow up PR.

PR #3800, opened 2024-04-18
Service Discovery: Promptly remove invalid watchers

ReubenBond:
Fixes #3815 

This PR fixes a case where errors are logged to output because an `IChangeToken` changes after an instance is disposed. When the token changes, it fires off a callback which logs an `ObjectDisposedException`. The PR implements disposal of the change token subscription and ensures that endpoint watchers are removed immediately when they become invalid:

* When a watcher is invalid, remove it immediately from the resolver entry collection
* Dispose `IChangeToken` registrations when they are no longer needed or valid
* Check for disposal in watcher and throw where appropriate
* Prevent registration of timers once the watcher has been disposed

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3800)


danmoseley:
I assume this is not required for GA

ReubenBond:
@danmoseley this is required for GA

oising (approved):
LGTM

BrennanConroy (approved):

PR #3801, opened 2024-04-18
Add callback to WithPgAdmin to allow customization.

mitchdenny:
Related: https://github.com/dotnet/aspire/issues/3770

This PR modifies the arguments for `WithPgAdmin` to take a callback which allows for a greater array of customization. Usage example:

```csharp
builder.AddPostgres("pgslq").WithPgAdmin(c => c.WithImageTag("8.3").UseHostPort(8999));
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3801)


davidfowl (approved):

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -68,32 +68,52 @@ public static IResourceBuilder<PostgresDatabaseResource> AddDatabase(this IResou
...
         return builder;
     }
 
+    /// <summary>
+    /// Configures the host port that the PGAdmin resource is exposed on instead of using randomly assigned port.
+    /// </summary>
+    /// <param name="builder">The resource builder for PGAdmin.</param>
+    /// <param name="port">The port to bind on the host. If null is used random port will be assigned.</param>
+    /// <returns>The resource builder for PGAdmin.</returns>
+    public static IResourceBuilder<PgAdminContainerResource> UseHostPort(this IResourceBuilder<PgAdminContainerResource> builder, int? port)
```
DamianEdwards, line 109: This is new right? Are we sure `Use` is the right prefix here? Seems odd that this method is only available for this resource type.

davidfowl: It's what we do with the other azure resources that have emulators:

https://github.com/dotnet/aspire/blob/06d8886d8acf0e0cb4090f15c799545a5514af9c/src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs#L149

PR #3821, opened 2024-04-19
Enable dependabot updates for github actions

danmoseley:
This doesn't update anything currently. We don't use actions much currently here (unlike aspire-samples) but there are some uses and this will keep them up to date.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3821)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -55,3 +55,11 @@ updates:
           - "Grpc.*"
     labels:
       - "area-codeflow"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 5
+    labels:
+      - area-engineering-systems


jeffhandley:
@danmoseley I've not looked into it. For the recent bumps to [locker.yml](https://github.com/dotnet/runtime/blob/main/.github/workflows/locker.yml) though, I don't expect dependabot would have caught that since the reference to node16 was indirect through the [vscode-github-triage-actions](https://github.com/microsoft/vscode-github-triage-actions/blob/stable/locker/action.yml) repo.

I don't see why not though--we can always disable it if the signal-to-noise ratio is poor.

.github/dependabot.yml
```diff
@@ -55,3 +55,11 @@ updates:
           - "Grpc.*"
     labels:
       - "area-codeflow"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
jeffhandley, line 60: Should/could this be `/.github/workflows`?

danmoseley: It runs recursively and I didn't see any need to bake that path in (?). We'll notice if it does something off in a PR.

RussKie (approved):

Issue #3200, opened 2024-03-27
Add integration tests for resource service auth


drewnoakes:
Following #3033 and #3400, we need integration test coverage for auth between the dashboard and a resource service.

This includes both API-key and certificate based auth. We don't currently ship a resource service that uses certificates, so we'd likely need to mock one for test purposes.

- [ ] Unsecured https://github.com/dotnet/aspire/pull/3824
- [ ] ApiKey https://github.com/dotnet/aspire/pull/3824
- [ ] Certificate

kvenkatrajan:
@drewnoakes  - are all the scenarios covered with #3824 ?

drewnoakes:
This was duped in #3739. Closing this as a dupe.

PR #3824, opened 2024-04-19
Add resource service auth tests (Unsecured, ApiKey)

drewnoakes:
Contributes to #3200

Adds tests for `Unsecured` and `ApiKey` auth modes for the resource service, from the perspective of the dashboard client in the dashboard project.

Also:
- Work around race condition between cancellation and connection attempts that leads to a previously-unhandled `ObjectDisposedException`.
- Move resource service code into own namespace, out of model.
- Use a more specific namespace for generated protobuf code.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3824)


tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
```diff
@@ -0,0 +1,177 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using Xunit;
+using DashboardServiceBase = Aspire.ResourceService.Proto.V1.DashboardService.DashboardServiceBase;
+
+namespace Aspire.Dashboard.Tests.Integration;
```
JamesNK, line 20: These tests are ok and don't need to be rewritten, but I think there is a simpler implementation. You could have added support for passing a custom `HttpMessageHandler` to the resource client, then capture the outgoing request when it is sent to the handler and returned a response from the handler. Then assert the outgoing request has headers expected. That saves having a mock server. The integration tests could turn into unit tests.

On the other hand, binary gRPC request/responses aren't easily approachable when working at a bytes level like a message handler does.

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+public sealed class DashboardClientAuthTests
+{
+    private const string ApiKeyHeaderName = "x-resource-service-api-key";
+
+    [Fact]
+    public async Task ConnectsToResourceService_Unsecured()
```
JamesNK, line 27: Should there be HTTPS versions of these tests?

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        Assert.NotNull(call.Request);
+        Assert.NotNull(call.RequestHeaders);
+        Assert.Equal("TestApiKey!", call.RequestHeaders.GetValue(ApiKeyHeaderName));
+    }
+
+    private static async Task<ResourceServiceServer> CreateResourceServiceServerAsync(Action<TestCalls>? configureCalls = null)
+    {
+        var serverAppBuilder = WebApplication.CreateSlimBuilder();
+
+        TestCalls testCalls = new();
```
JamesNK, line 56: nit: we use var when possible

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                AuthMode = authMode,
+                Url = serverAddress
+            }
+        };
+
+        configureOptions?.Invoke(options);
+
+        options.ResourceServiceClient.TryParseOptions(out _);
+
+        DashboardClient client = new(
```
JamesNK, line 105: nit: var

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            loggerFactory: NullLoggerFactory.Instance,
+            configuration: new ConfigurationManager(),
+            dashboardOptions: Options.Create(options));
+
+        var iClient = (IDashboardClient)client;
+
+        await iClient.WhenConnected;
```
JamesNK, line 112: Can WhenConnected stop being explicitly implemented? I don't see the value and it makes using and testing the property a little harder.

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
JamesNK: On that topic, I thought there wasn't much point in testing client cert auth, since it was just adding a certificate to the handler, but certificate loading happens inside the client.

I think we either need to test the end-to-end with cert auth or split the cert loading out into a separate type and test loading cert from configuration independently.

danmoseley:
Presumably this would be good in the 8.0 branch too when it's ready.

kvenkatrajan (approved):

tests/Aspire.Dashboard.Tests/Integration/DashboardClientAuthTests.cs
drewnoakes: > Should there be HTTPS versions of these tests?

https://github.com/dotnet/aspire/pull/4621

PR #3826, opened 2024-04-19
WithRedisCommander and WithMongoExpress configContainer callbacks.

mitchdenny:
Add container configuration callbacks to both `WithMongoExpress` and `WithRedisCommander`. Also add corresponding `UseHostPort` methods to simplify overriding the host port.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3826)


davidfowl:
We need to fix cosmos and storage emulators, and pgadmin in a follow up.

mitchdenny:
I think Cosmos and Storage are in a slightly different category. For Storage there are three different ports, so WithHostPort becomes WithBlobPort, WithQueuePort etc. I guess we could collapse them into one but if you are just using blob storage that seems odd.

In the case of Cosmos I am hopeful that at some point the future we can avoid having to use the Gateway port as the sigular port we use to connect on. At which point we'll have multiple methods to set the gateway and the partition port ranges (I think its ports == partitions .. a good question for the Cosmos team).

davidfowl:
I mean the With vs Use prefix is all.

davidfowl (approved):

PR #3837, opened 2024-04-19
Service Discovery: make scheme selection more intuitive in un-specified case, add more tests

ReubenBond:
### Make scheme selection more intuitive

* If `ServiceDiscoveryOptions.AllowAllSchemes` is TRUE and the service name INCLUDES schemes (eg `https://basket`), we filter out results which do not have a matching scheme, applying the preference. i.e, `https+http` means pick only `https` endpoints if _any_ `https` endpoints are present, `http` only if _no_ `https` endpoints are present.
* If `ServiceDiscoveryOptions.AllowAllSchemes` is TRUE and the service name DOES NOT INCLUDE schemes (eg, `basket`), then we should allow all endpoints.
* If `ServiceDiscoveryOptions.AllowAllSchemes` is FALSE, the service name DOES NOT INCLUDE schemes, and `ServiceDiscoveryOptions.AllowedSchemes` contains some value (eg, `["https"]`), then we should filter to include only endpoints matching those `AllowedSchemes` values.

In pseudo code, the logic is: 

```
IncludeAll = AllowAllSchemes && ServiceName.Schemes.Count == 0
IncludedSchemes = ServiceName.Schemes.Count > 0 ? ServiceName.Schemes.Intersect(AllowedSchemes) : AllowedSchemes
```

### Always check "default" config section first when no endpoint name is specified

* When searching for a configuration section for a service and no endpoint name has been specified (i.e, the common case), we should first check the `"default"` section, then the sections corresponding to schemes. The current behavior is:
  * If the query specified any schemes, check those sections in order (`https+http` -> check `"https"` then `"http"` cfg sections)
  * If the query does not specify any schemes, check the `"default"` section.
 
The change is to always check the default section first if no endpoint name is specified.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3837)

### DIFF
--- a/src/Microsoft.Extensions.ServiceDiscovery/Configuration/ConfigurationServiceEndpointProvider.cs
+++ b/src/Microsoft.Extensions.ServiceDiscovery/Configuration/ConfigurationServiceEndpointProvider.cs
@@ -17,6 +17,7 @@ internal sealed partial class ConfigurationServiceEndpointProvider : IServiceEnd
     private const string DefaultEndpointName = "default";
     private readonly string _serviceName;
     private readonly string? _endpointName;
+    private readonly bool _includeAllSchemes;
     private readonly string[] _schemes;
     private readonly IConfiguration _configuration;
     private readonly ILogger<ConfigurationServiceEndpointProvider> _logger;
@@ -39,6 +40,7 @@ internal sealed partial class ConfigurationServiceEndpointProvider : IServiceEnd
     {
         _serviceName = query.ServiceName;
         _endpointName = query.EndpointName;
+        _includeAllSchemes = serviceDiscoveryOptions.Value.AllowAllSchemes && query.IncludedSchemes.Count == 0;
         _schemes = ServiceDiscoveryOptions.ApplyAllowedSchemes(query.IncludedSchemes, serviceDiscoveryOptions.Value.AllowedSchemes, serviceDiscoveryOptions.Value.AllowAllSchemes);
         _configuration = configuration;
         _logger = logger;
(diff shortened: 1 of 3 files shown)


BrennanConroy (approved):

src/Microsoft.Extensions.ServiceDiscovery/Configuration/ConfigurationServiceEndpointProvider.cs
```diff
@@ -135,46 +127,60 @@ public ValueTask PopulateAsync(IServiceEndpointBuilder endpoints, CancellationTo
...
-                var index = Array.IndexOf(_schemes, scheme);
-                if (index >= 0 && index <= minIndex)
+                if (ep.EndPoint is UriEndPoint uri && uri.Uri.Scheme is { } scheme)
                 {
-                    ++added;
-                    endpoints.Endpoints.Add(ep);
+                    var index = Array.IndexOf(_schemes, scheme);
+                    if (index >= 0 && index < minIndex)
+                    {
+                        minIndex = index;
```
BrennanConroy, line 153: Re-reading this logic, not sure I understand it.
e.g.
`_schemes = [ "https", "http" ]`
`resolved = [ "http://basket", "https://orders" ]`
`minIndex` will be 0 so we'll only add `"https://orders"` to `endpoints` below?

ReubenBond: Yes, that's correct. `_schemes` is in order of preference, lower index = higher preference. We want to pick preferred results only: if an `https` result is found, exclude `http` results, since `https` appears before `http` in `_schemes`.

ReubenBond: The idea is to allow `https+http://foo` so that we can use `http` at dev time and restrict ourselves to `https` in prod using config without changing the code. We changed our policy and opted to use HTTPS everywhere, when possible, but it's still not straightforward/automated on some Linux distros, so there is still some value to this otherwise whacky feature.

ReubenBond: > But it's applying to different endpoints? 

I don't quite understand this, what does it mean?

BrennanConroy: Offline discussion: This isn't a mix of endpoints, so my question doesn't apply.

PR #3845, opened 2024-04-19
Testing: Do not configure HttpClient resilience by default

### LINKED ISSUE #3431: Cannot override default HttpClient resiliency settings when using CreateHttpClient from Aspire.Hosting.Testing
The `CreateHttpClient` extension method on `Aspire.Hosting.Testing.DistributedApplicationHostingTestingExtensions` uses the `IHttpClientFactory` registered in the `DistributedApplication` instance's `IServiceProvider` to create an `HttpClient` for the specified resource (and optionally endpoint name). This `HttpClient` is pre-configured with the standard resiliency handler but this isn't always sufficient for testing scenarios where longer timeouts, etc. might be desired. Unfortunately there doesn't seem to be a way to modify the resiliency handler that's added by default, only add another one (i.e. every call to `IHttpClientBuilder.AddStandardResilencyHandler()` adds another resiliency pipeline to the `HttpClient`).

In the Aspire Samples tests and integration tests in this repo, extra methods have been added to enable creating an `HttpClient` without using the `IHttpClientFactory` that is registered in the `DistributedApplication`'s `IServiceProvider` because of this issue.

We should consider ways to make it easier to create an `HttpClient` instance for a resource/endpoint with custom configuration, or remove the automatic registration of the standard resiliency handler and leave it to the consumer to add the resilience they want.

/Cc @ReubenBond @radical @martincostello @eerhardt 

Related: dotnet/extensions#4814

ReubenBond:
Since it is not straightforward to configure resilience policies after the handler has been added (see https://github.com/dotnet/extensions/issues/4814), we decided to remove these from the testing host so that developers are free to choose how resilience is configured in test projects.

cc @DamianEdwards 

Fixes #3431

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3845)

### DIFF
--- a/src/Aspire.Hosting.Testing/DistributedApplicationFactoryOfT.cs
+++ b/src/Aspire.Hosting.Testing/DistributedApplicationFactoryOfT.cs
@@ -148,11 +148,8 @@ public class DistributedApplicationFactory<TEntryPoint> : IDisposable, IAsyncDis
 
     private void OnBuildingCore(DistributedApplicationBuilder applicationBuilder)
     {
-        // Patch DcpOptions configuration
         var services = applicationBuilder.Services;
-
         services.AddHttpClient();
-        services.ConfigureHttpClientDefaults(http => http.AddStandardResilienceHandler());
 
         InterceptHostCreation(applicationBuilder);
 


DamianEdwards (approved):

radical:
Does this need a corresponding update in docs somewhere?

ReubenBond:
We don't have docs for this area yet, so kind-of

PR #3828, opened 2024-04-19
Aspire.Hosting.* READMEs

mitchdenny:
Partially fixes #2955

READMEs are in this PR, icons are in #3860 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3828)

PR #3860, opened 2024-04-22
Add icons to packages.

mitchdenny:
Adding some icons to packages. I split this out from #3828 because I am not sure if we have permission to use all of these and I didn't want to hold up the READMEs (its critical for them to be in GA).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3860)


danmoseley:
Did you look at the usage policies for each one - usually on the official sites? Maybe link to those if they're handy?

mitchdenny:
OK so here is my research:

* Dapr I think is OK, its covered by this I believe: https://www.linuxfoundation.org/legal/trademark-usage
* AWS ... I'll page @normj  :)
* Kafka: https://kafka.apache.org/trademark
* MongoDB: https://www.mongodb.com/legal/trademark-usage-guidelines
* MySQL: https://www.mysql.com/about/legal/logos.html
* Node.js: https://images.prismic.io/openjsf/69641abf-ac84-44a2-b719-ee4498d07b06_OpenJS-Foundation-Trademark-Policy-2021-01-12.docx.pdf
* Oracle: https://www.oracle.com/legal/logos.html
* RabbitMQ: https://www.rabbitmq.com/trademark-guidelines

Still need to go through and try and interpret them. But I'm also cognitive that this not my specialty and I don't want to get it wrong :)

DamianEdwards:
Please only use logos we are already using for the component packages. We have explicitly sought and been granted permission for each of those already.

DamianEdwards:
@mitchdenny probably best to just log an issue for this and assign to me so that I can coordinate this in the same way we have for the component packages.

PR #3869, opened 2024-04-22
Update docker image tags

radical:
- `confluentinc/confluent-local`: `7.6.0` -> `7.6.1`
- `library/mongo`: `7.0.5` -> `7.0.8`
- `qdrant/qdrant`: `1.8.3` -> `1.8.4`

(cherry picked from commit 7994886
- **Update docker image tags**
- **Update versions in playground, tests, and xml docs**

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3869)

### DIFF
--- a/playground/Qdrant/Qdrant.AppHost/aspire-manifest.json
+++ b/playground/Qdrant/Qdrant.AppHost/aspire-manifest.json
@@ -3,7 +3,7 @@
     "qdrant": {
       "type": "container.v0",
       "connectionString": "Endpoint={qdrant.bindings.grpc.url};Key={qdrant-Key.value}",
-      "image": "docker.io/qdrant/qdrant:v1.8.3",
+      "image": "docker.io/qdrant/qdrant:v1.8.4",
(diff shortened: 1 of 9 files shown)


mitchdenny (approved):

eerhardt (approved):

danmoseley:
For future reference next time we need to do this, how do you find which to update and what to update them to? I assume not by manually browsing the registry. Is there a docker command? is there any alternative to us just periodically repeating this exercise (ie., there's no dependabot like in this context)

mitchdenny:
We could potentially script it by putting marker comments on the lines and then having a script go through and update the files directly? The Azure SDK team did things like this:

https://github.com/Azure/azure-sdk-for-java/blob/3c12cc42b7f29ed7e72c2742297a0519a40d8974/sdk/eventhubs/azure-messaging-eventhubs/docs/pom.xml#L23

PR #3875, opened 2024-04-23
Add non-generic version of DistributedApplicationTestingBuilder.CreateAsync

### LINKED ISSUE #3874: Add non-generic overload of DistributedApplicationTestingBuilder.CreateAsync
There is currently no non-generic overload of `DistributedApplicationTestingBuilder.CreateAsync<TEntrypoint>()`. We should have a non-generic overload that accepts a `Type` parameter. I've had to work around this via [reflection and dynamic invocation in the samples repo](https://github.com/dotnet/aspire-samples/blob/8edd8a56bbacb1580fc567b696fef27e82f7d65b/tests/SamplesIntegrationTests/DistributedApplicationTestFactory.cs#L18-L31) to support running tests against discovered AppHost projects.

DamianEdwards:
~This was a bit harder than I had originally anticipated due to API compat requirements now that we're locked down for GA. Unfortunately that meant I had to perform some less-than-ideal gymnastics to share implementation and preserve API details across the existing generic version and new non-generic version (inner types that override members and forward to outer type, etc.).~

This is a breaking change for GA to unwind the generics and simply have a generic version of the top level factory method.

Fixes #3874
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3875)


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -127,8 +127,8 @@ private void OnBuiltCore(DistributedApplication application)
     private void OnBuilderCreatingCore(DistributedApplicationOptions applicationOptions, HostApplicationBuilderSettings hostBuilderOptions)
     {
         hostBuilderOptions.EnvironmentName = Environments.Development;
-        hostBuilderOptions.ApplicationName = typeof(TEntryPoint).Assembly.GetName().Name ?? string.Empty;
-        applicationOptions.AssemblyName = typeof(TEntryPoint).Assembly.GetName().Name ?? string.Empty;
+        hostBuilderOptions.ApplicationName = entryPoint.Assembly.GetName().Name ?? string.Empty;
+        applicationOptions.AssemblyName = entryPoint.Assembly.GetName().Name ?? string.Empty;
```
eerhardt, line 131:
```suggestion
        hostBuilderOptions.ApplicationName = _entryPoint.Assembly.GetName().Name ?? string.Empty;
        applicationOptions.AssemblyName = _entryPoint.Assembly.GetName().Name ?? string.Empty;
```

Should use the field everywhere.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -1,20 +1,20 @@
...
 
 namespace Aspire.Hosting.Testing;
 
 /// <summary>
 /// Factory for creating a distributed application for testing.
 /// </summary>
-/// <typeparam name="TEntryPoint">
-/// A type in the entry point assembly of the target Aspire AppHost. Typically, the Program class can be used.
-/// </typeparam>
-public class DistributedApplicationFactory<TEntryPoint> : IDisposable, IAsyncDisposable where TEntryPoint : class
```
eerhardt, line 16: Are we going to continue to validate `where TEntryPoint : class`. Was that important?

eerhardt (approved):
2 minor-ish comments. This looks good to me.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
eerhardt: I'm just talking about in this private instance method.

eerhardt: "Should use the field everywhere _possible_"

mitchdenny:
LGTM!

mitchdenny (approved):

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
DamianEdwards: Seems just a carry over from `WebApplicationFactory<TEntryPoint>` which this was based on but I can't see a good reason why it's restricted there too. I don't think it matters too much. We can always loosen the generic type parameter constraint if we ever get feedback on it and folks aren't willing to use the `Type` overload.

PR #3876, opened 2024-04-23
[Cosmos .NET SDK] - Adds User Agent to Identify Usage

kundadebdatta:
### Pull Request Description:

The purpose of this PR is to populate the `ApplicationName` parameter in the `CosmosClientOptions` so that the user agent can be generated with the necessary suffix. This will help the cosmos db team to identify if the request was originated from Aspire. A sample of the user agent will look like the below:

`cosmos-netstandard-sdk/3.39.1|0|X64|Microsoft Windows 10.0.22635|.NET 6.0.29|N|F 00000111|AspireMicrosoftAzureCosmosDBClient|`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3876)

### DIFF
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
@@ -83,7 +83,6 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
         configureSettings?.Invoke(settings);
 
         var clientOptions = new CosmosClientOptions();
-        clientOptions.ApplicationName = CosmosConstants.CosmosApplicationName;
 
         // Needs to be enabled for either logging or tracing to work.
         clientOptions.CosmosClientTelemetryOptions.DisableDistributedTracing = false;
@@ -103,6 +102,13 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
 
         configureClientOptions?.Invoke(clientOptions);
 
+        var cosmosApplicationName = CosmosConstants.CosmosApplicationName;
+        if (!string.IsNullOrEmpty(clientOptions.ApplicationName))
+        {
+            cosmosApplicationName += clientOptions.ApplicationName;
+        }
+
+        clientOptions.ApplicationName = cosmosApplicationName;
         if (serviceKey is null)
         {
             builder.Services.AddSingleton(_ => ConfigureDb());
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
@@ -104,7 +104,7 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
         var cosmosApplicationName = CosmosConstants.CosmosApplicationName;
         if (!string.IsNullOrEmpty(clientOptions.ApplicationName))
         {
-            cosmosApplicationName += clientOptions.ApplicationName;
+            cosmosApplicationName = $"{cosmosApplicationName}|{clientOptions.ApplicationName}";
         }
 
         clientOptions.ApplicationName = cosmosApplicationName;
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
@@ -83,7 +83,6 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
         configureSettings?.Invoke(settings);
 
         var clientOptions = new CosmosClientOptions();
-
         // Needs to be enabled for either logging or tracing to work.
         clientOptions.CosmosClientTelemetryOptions.DisableDistributedTracing = false;
         if (!settings.DisableTracing)
--- a/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
+++ b/src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
@@ -83,6 +83,8 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
         configureSettings?.Invoke(settings);
 
         var clientOptions = new CosmosClientOptions();
+        clientOptions.ApplicationName = CosmosConstants.CosmosApplicationName;
+
         // Needs to be enabled for either logging or tracing to work.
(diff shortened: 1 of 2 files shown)


mitchdenny:
LGTM. @eerhardt

mitchdenny (approved):

kirankumarkolli (approved):

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
```diff
@@ -83,6 +83,8 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
         configureSettings?.Invoke(settings);
 
         var clientOptions = new CosmosClientOptions();
+        clientOptions.ApplicationName = CosmosConstants.CosmosApplicationName;
```
eerhardt, line 86: Just an FYI that users can (and probably will) overwrite this value. We show them how to do this in the README:

https://github.com/dotnet/aspire/blob/87c39e0422f288e2c5a5ac0349ea33f514bc7d8a/src/Components/Aspire.Microsoft.Azure.Cosmos/README.md?plain=1#L105-L109

eerhardt, line 86: Do we want to set the same for the the EntityFramework component: https://github.com/dotnet/aspire/tree/main/src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos?

kundadebdatta: Thanks! This was a good catch. In order to address this, I have prefixed the application name with whatever value the users provide. This will make sure that even if the user doesn't set the application name, it will have a default value when it reaches to out service.

kundadebdatta: I am not sure if we really need to set this. When I look at the definition of `CosmosDbContextOptionsBuilder`, I do not see an option to set the application name in the entity framework library. Cc: @kirankumarkolli .

src/Shared/Cosmos/CosmosConstants.cs
```diff
@@ -5,6 +5,12 @@ namespace Aspire.Hosting.Azure.Cosmos;
 
 internal static class CosmosConstants
 {
+    /// <summary>
+    /// Defines the application name used to interact with the aszure cosmos db. This will be suffixed to the
+    /// cosmos user-agent to include with every Azure Cosmos DB service interaction.
+    /// </summary>
+    internal const string CosmosApplicationName = "AspireMicrosoftAzureCosmosDBClient";
```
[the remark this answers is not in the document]
kundadebdatta: Updated.

kirankumarkolli (approved):

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
```diff
@@ -101,6 +101,13 @@ public static class AspireMicrosoftAzureCosmosDBExtensions
 
         configureClientOptions?.Invoke(clientOptions);
 
+        var cosmosApplicationName = CosmosConstants.CosmosApplicationName;
+        if (!string.IsNullOrEmpty(clientOptions.ApplicationName))
+        {
+            cosmosApplicationName += clientOptions.ApplicationName;
```
kirankumarkolli, line 107: nit: Add a separator

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
kirankumarkolli: https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.infrastructure.cosmosdbcontextoptionsbuilder?view=efcore-8.0

Am also not seeing direct API for it. 

@ajcvickers thoughts?

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
kundadebdatta: Added the `|` separator at the moment.

kirankumarkolli (approved):

src/Shared/Cosmos/CosmosConstants.cs
```diff
@@ -5,6 +5,12 @@ namespace Aspire.Hosting.Azure.Cosmos;
 
 internal static class CosmosConstants
 {
+    /// <summary>
+    /// Defines the application name used to interact with the aszure cosmos db. This will be suffixed to the
+    /// cosmos user-agent to include with every Azure Cosmos DB service interaction.
+    /// </summary>
+    internal const string CosmosApplicationName = "AspireCosmosDBClient";
```
ealsur, line 12: Is there an upper limit on the User Agent? If so, we might want to consider shortening this if possible. Maybe just `Aspire` as the User Agent already contains the fact that it's a Cosmos DB Client.

kundadebdatta: We already have a check in place for validating and trimming the `Suffix` within the `Microsoft.Azure.Documents.UserAgentContainer`:

```
    public string Suffix
    {
        get
        {
            return suffix;
        }
        set
        {
            suffix = value;
            if (suffix.Length > 64)
            {
                suffix = suffix.Substring(0, 64);
            }

            userAgent = BaseUserAgent + suffix;
            userAgentUTF8 = Encoding.UTF8.GetBytes(userAgent);
        }
    }
```

So, in case it's larger than the defined size, we will fit it within the `64` character window, which would still contain the required application name prefix for us to track.

ealsur: That's exactly the point. Because there is a limit, any character we used for the fixed part takes away from the potential user app name.

kundadebdatta: I would agree with @ealsur. Just by using `Aspire` doesn't make much difference, in terms of identification. With this, we would make extra room for the user provided app name, if any.

Addressed with a new commit.

ealsur (approved):

kirankumarkolli (approved):

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosDBExtensions.cs
eerhardt: Is the `|` the correct separator?

https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

Seems to suggest using a slash might be preferred?

eerhardt (approved):

PR #3881, opened 2024-04-23
Improve filtering of log messages when there are duplicate attributes

### LINKED ISSUE #3614: Message column duplicates if Attributes contains Message.

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/3614

Log entries have built-in fields. e.g. the message body, level, category, etc. Log entries also have custom fields. This is from structured data about a message.

It's possible for these to conflict. Structured data could be called `Message` or `Category`. If the user then asks to filter with the `Message` value, which is used? A conflict is created and it's only possible to filter the built-in fields.

This PRs fix is to give custom built-in fields an internal name, e.g. `log.message`. That is much less likely to conflict with real structured data.

Also, I added a divider (disabled select item with text `-`) between built-in fields and structured data fields to make the UI experience a bit better:

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3881)

### DIFF
--- a/src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
+++ b/src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
@@ -27,6 +27,8 @@ public partial class FilterDialog
     public required TelemetryRepository TelemetryRepository { get; set; }
 
     private LogDialogFormModel _formModel = default!;
+    private List<SelectViewModel<string>> _parameters = default!;
+
     public EditContext EditContext { get; private set; } = default!;
 
     protected override void OnInitialized()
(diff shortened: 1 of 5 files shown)


JamesNK:
Review please 🙏

src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
```diff
@@ -41,23 +43,35 @@ protected override void OnInitialized()
...
+    protected override void OnParametersSet()
+    {
+        var names = new List<string> { LogFilter.KnownMessageField, LogFilter.KnownCategoryField, LogFilter.KnownApplicationField, LogFilter.KnownTraceIdField, LogFilter.KnownSpanIdField, LogFilter.KnownOriginalFormatField };
+        var knownFields = names.Select(p => new SelectViewModel<string> { Id = p, Name = LogFilter.ResolveFieldName(p) }).ToList();
+        var customFields = Content.LogPropertyKeys.Select(p => new SelectViewModel<string> { Id = p, Name = LogFilter.ResolveFieldName(p) }).ToList();
+
+        _parameters =
+        [
+            .. knownFields,
+            new SelectViewModel<string> { Id = null, Name = "-" },
```
drewnoakes, line 57: Is there not a better way to create a separator with this control?

drewnoakes (approved):

src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
JamesNK: Not that I know of - https://www.fluentui-blazor.net/Select

PR #3885, opened 2024-04-23
Fix Keyed Redis

### LINKED ISSUE #3873: Keyed Redis component doesn't work correctly
If I add 2 Redis server instances and try to use "keyed" services to talk to each, only the first connection is used for both. The 2nd connection doesn't get used.

```C#

var basketCache = builder.AddRedis("basketcache");
var basketCache2 = builder.AddRedis("basketcache2");

var basketService = builder.AddProject("basketservice", @"..\BasketService\BasketService.csproj")
                           .WithReference(basketCache)
                           .WithReference(basketCache2);
```

And then in BasketService's Program.cs:

```C#

builder.AddKeyedRedisClient("basketcache");
builder.AddKeyedRedisClient("basketcache2");

var app = builder.Build();

var m = app.Services.GetRequiredKeyedService<IConnectionMultiplexer>("basketcache");
var m2 = app.Services.GetRequiredKeyedService<IConnectionMultiplexer>("basketcache2");

Console.WriteLine(m.Configuration);
Console.WriteLine(m2.Configuration);
```

eerhardt:
When adding 2 Redis server instances and try to use "keyed" services to talk to each, only the first connection is used for both. The 2nd connection doesn't get used.

This is because we are using a singleton `OptionsFactory<ConfigurationOptions>` that only knows about the first StackExchangeRedisSettings registered, and always uses its ConnectionString.

The fix is to teach the factory about all the Settings objects being registered, and have it find the correct one given the "options name" when it is being created.

Fix #3873

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3885)

### DIFF
--- a/src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
+++ b/src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
@@ -83,7 +83,7 @@ public static class AspireRedisExtensions
         var optionsName = serviceKey is null ? Options.Options.DefaultName : connectionName;
 
         // see comments on ConfigurationOptionsFactory for why a factory is used here
-        builder.Services.AddTransient(sp => new RedisOptionsSettingsPair { OptionsName = optionsName, Settings = settings });
+        builder.Services.AddKeyedTransient(optionsName, (sp, _) => new RedisSettingsAdapterService { Settings = settings });
         builder.Services.TryAddTransient<IOptionsFactory<ConfigurationOptions>, ConfigurationOptionsFactory>();
 
         builder.Services.Configure<ConfigurationOptions>(
@@ -173,9 +173,12 @@ public static class AspireRedisExtensions
         return options;
     }
 
-    private sealed class RedisOptionsSettingsPair
+    /// <summary>
+    /// Used to pass StackExchangeRedisSettings instances to the ConfigurationOptionsFactory.
+    /// </summary>
+    /// <remarks>Not using StackExchangeRedisSettings itself because it is a public type that someone else could register in DI.</remarks>
+    private sealed class RedisSettingsAdapterService
     {
-        public required string OptionsName { get; init; }
         public required StackExchangeRedisSettings Settings { get; init; }
     }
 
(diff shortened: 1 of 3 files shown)


src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
```diff
@@ -189,17 +192,17 @@ private static ConfigurationOptions BindToConfiguration(ConfigurationOptions opt
...
             : base(setups, postConfigures, validations)
         {
-            _settings = settings;
+            _settingsPairs = settingsPairs;
         }
 
         protected override ConfigurationOptions CreateInstance(string name)
         {
-            var connectionString = _settings.ConnectionString;
+            var connectionString = _settingsPairs.FirstOrDefault(p => p.OptionsName == name)?.Settings.ConnectionString;
```
davidfowl, line 205: What should the error be if there's no setting?

src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
```diff
@@ -80,17 +80,14 @@ public static class AspireRedisExtensions
...
         // see comments on ConfigurationOptionsFactory for why a factory is used here
-        builder.Services.TryAdd(ServiceDescriptor.Transient(typeof(IOptionsFactory<ConfigurationOptions>),
-            sp => new ConfigurationOptionsFactory(
-                settings,
-                sp.GetServices<IConfigureOptions<ConfigurationOptions>>(),
-                sp.GetServices<IPostConfigureOptions<ConfigurationOptions>>(),
-                sp.GetServices<IValidateOptions<ConfigurationOptions>>())));
-
-        string? optionsName = serviceKey is null ? null : connectionName;
+        builder.Services.AddTransient(sp => new RedisOptionsSettingsPair { OptionsName = optionsName, Settings = settings });
```
davidfowl, line 86: This could use keyed di itself

eerhardt: How? The IOptions system doesn't work with keyed DI.

src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
eerhardt: We don't error here, but instead just create a "blank" ConfigurationOptions instance.

If there's no setting it basically means we have a bug, and not that the user did something wrong. I'm not sure how a user could get into that situation.

eerhardt: > I guess someone could try serviceProvider.GetRequiredService<IOptionsMonitor<ConfigurationOptions>>().Get("doesNotExit").

Returning a blank ConfigurationOptions instance is consistent with other uses of named options that don't exist:

image

src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
Zombach: Isn't that how it works?
```csharp
builder.Services.AddKeyedTransient<IOptions<Foo>>("key", (_, _) => Options.Create(new Fo0()));
builder.Services.AddKeyedTransient<IOptions<Foo>>("key2", (_, _) => Options.Create(new Fo0()));
```

eerhardt: OK, I see what you mean. Just the `RedisOptionsSettingsPair` uses keyed DI, but not the OptionsFactory. I thought you meant the whole factory as well.

src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs
eerhardt: I added a comment for this.

radical (approved):

PR #3897, opened 2024-04-23
Fix the casing of AddKeyedAzureCosmosDBClient

eerhardt:
Also fix the casing in the READMEs as well.

Note that the EF Component has `AddCosmosDbContext`, but there it is a "Cosmos" "DbContext", so the casing is correct.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3897)


radical:
Just curious - what is the motivation for using `DB` vs `Db` here, if efcore is already using `Db`?

eerhardt:
> Just curious - what is the motivation for using `DB` vs `Db` here, if efcore is already using `Db`?

To be consistent with all the other places that we say "CosmosDB".

Technically, the .NET naming guidelines say that 2 letter acroynms are upper cased, like System.IO.

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers

radical (approved):

eerhardt:
One other option here would be to drop `DB` and just say `AddAzureCosmosClient`. But we should probably do that everywhere - like in Aspire.Hosting.Azure.CosmosDB.

cc @mitchdenny

DamianEdwards (approved):
I'm happy as long as we're consistent.

mitchdenny:
We should be consistent with https://www.nuget.org/packages/Azure.ResourceManager.CosmosDB/

danmoseley (approved):

eerhardt:
> One other option here would be to drop `DB` and just say `AddAzureCosmosClient`. But we should probably do that everywhere - like in Aspire.Hosting.Azure.CosmosDB.

Closing in favor of #3906, where the Cosmos team decided to drop "DB" from the runtime component APIs.

PR #2781, opened 2024-03-11
Fix docs after client renaming

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2781)

PR #2630, opened 2024-03-05
Update method names for client registrations

sebastienros:
Fixes #1510

We may consider different names based on taste. Sometimes the name matches the client type, and sometimes not.

Examples:

- `AddRabbitMQClient()`, maybe remove the `MQ`. The client name is `IConnection`.
- `AddRedisClient()`, the client name is `IConnectionMultiplexer`, I think it's ok.
- `AddCosmosClient()`, the client name is `CosmosClient` but might sound better with `AddCosmosDBClient`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2630)

PR #3898, opened 2024-04-23
[release/8.0] Fix the casing of AddKeyedAzureCosmosDBClient

eerhardt:
Also fix the casing in the READMEs as well.

Backport of #3897 to release/8.0

/cc @eerhardt

## Customer Impact

Our public API is cased differently:

* AddAzureCosmosDBClient
* AddKeyedAzureCosmosDbClient

If we ship 8.0 this way, we will be forced to keep it forever since it is a stable release.

Also fixing the READMEs to have the correct case, which they were inconsistent on being right or not.

## Testing

Tests updated with the new name. I searched for "CosmosDb" and only found usages for "Cosmos" "DbContext", which is the correct EF casing of DbContext.

## Risk

Low

## Regression?

Yes, we regressed here with:

* https://github.com/dotnet/aspire/pull/2630
* https://github.com/dotnet/aspire/pull/2781

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3898)

### DIFF
--- a/src/Aspire.Hosting.Azure.CosmosDB/README.md
+++ b/src/Aspire.Hosting.Azure.CosmosDB/README.md
@@ -47,7 +47,7 @@ When the AppHost starts up a local container running the Azure CosmosDB will als
 
 ```csharp
 // Service code
-builder.AddAzureCosmosDbClient("cosmos");
+builder.AddAzureCosmosDBClient("cosmos");
 ```
 
 ## Additional documentation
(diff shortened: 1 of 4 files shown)


radical (approved):

danmoseley:
i did the same grep as you and only found EnrichCosmosDbTests, which doesn't matter.

DamianEdwards (approved):

danmoseley (approved):

Pilchie:
Sorry, and thanks for addressing.  I talked to the team, and it turns out we don't typically use `DB` in API names at all.  I recognize it might be too late, but given that the type name is `CosmosClient`, would it be too late to change it just to `AddAzureCosmosClient` and `AddKeyedAzureCosmosClient`? or even `Add[Keyed]CosmosClient` if we don't think Azure is important in the name?

danmoseley:
I'm happy to take such a change if it's, say, in the next couple days. Consistency is important. @eerhardt ?

Pilchie:
Great - @kundadebdatta could you look at changing this?

PR #3903, opened 2024-04-24
Add Encrypt=True to test connection strings

GrabYourPitchforks:
This should silence some CodeQL alerts we're receiving in our backend systems.

Yes, it's test code. There's a weird policy governing when test code is / is not required to be in scope during CodeQL runs. Making these very targeted changes to the test code seemed like the path of least resistance.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3903)

### DIFF
--- a/tests/Aspire.Microsoft.Data.SqlClient.Tests/AspireSqlServerSqlClientExtensionsTests.cs
+++ b/tests/Aspire.Microsoft.Data.SqlClient.Tests/AspireSqlServerSqlClientExtensionsTests.cs
@@ -11,7 +11,7 @@ namespace Aspire.Microsoft.Data.SqlClient.Tests;
 
 public class AspireSqlServerSqlClientExtensionsTests
 {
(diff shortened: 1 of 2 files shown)


eerhardt (approved):
Lgtm

is there anyway to catch these kinds of things during PR validation?

GrabYourPitchforks:
> is there anyway to catch these kinds of things during PR validation?

Not really. CodeQL runs are intentionally asynchronous since they take so long to execute. Otherwise people would be waiting hours - maybe days - for CI to go green. And then you have an angry mob at your doorstep. :)

GrabYourPitchforks:
Following up - confirmed that the raw tool output from the latest runs shows this issue as resolved. It'll take a few days to percolate through the reporting system, but we should consider this fully resolved.

PR #3908, opened 2024-04-24
Oracle data volume following readonly convention

### LINKED ISSUE #3907: Oracle WithDataVolume should not be readonly
The Oracle WithDataVolume implementation is inconsistent with itself and other implementations.

[WithDataVolume](https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.Oracle/OracleDatabaseBuilderExtensions.cs#L63) is set to always mount as readonly, which obviously breaks the Oracle container as it can't open the database. WithDataBindMount in the same file hard codes readonly as false.

Looking at the SQLServer [implementation](https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs#L63) is has an optional parameter with a default of false for both, as does Postgres. Will put in a PR following that convention.

cmdkeen:
Closes #3907 - Oracle needs the data volume to be writeable. Following the convention on other database implementations to use a default false parameter.

Also fixes inconsistency between the readonly states of WithDataVolume and WithDataBindMount 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3908)


cmdkeen:
@dotnet-policy-service agree

src/Aspire.Hosting.Oracle/OracleDatabaseBuilderExtensions.cs
```diff
@@ -58,18 +58,20 @@ public static IResourceBuilder<OracleDatabaseResource> AddDatabase(this IResourc
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<OracleDatabaseServerResource> WithDataVolume(this IResourceBuilder<OracleDatabaseServerResource> builder, string? name = null)
-        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/opt/oracle/oradata", true);
+    public static IResourceBuilder<OracleDatabaseServerResource> WithDataVolume(this IResourceBuilder<OracleDatabaseServerResource> builder, string? name = null, bool isReadOnly = false)
+        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/opt/oracle/oradata", isReadOnly);
```
danmoseley, line 64: This will be a source breaking change in 8.1, right? Default parameters are fine to add, but it changes the default. If this change is important, perhaps we need it in 8.0
@eerhardt

cmdkeen: From my perspective the current default is hard broken - the default Oracle image does not come up if the data volume is mounted as read-only. I'll add a bit more detail on the linked issue.

eerhardt: If the current default is "hard broken", why are we making it an option at all? If someone passes `isReadOnly: true` wouldn't they be just as broken?

src/Aspire.Hosting.Oracle/OracleDatabaseBuilderExtensions.cs
```diff
@@ -58,18 +58,20 @@ public static IResourceBuilder<OracleDatabaseResource> AddDatabase(this IResourc
     /// </summary>
     /// <param name="builder">The resource builder.</param>
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<OracleDatabaseServerResource> WithDataVolume(this IResourceBuilder<OracleDatabaseServerResource> builder, string? name = null)
-        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/opt/oracle/oradata", true);
+    public static IResourceBuilder<OracleDatabaseServerResource> WithDataVolume(this IResourceBuilder<OracleDatabaseServerResource> builder, string? name = null, bool isReadOnly = false)
```
eerhardt, line 63: Are there any tests we can write for these APIs?

davidfowl:
Feels important to get this for GA since it's a breaking change

cc @mitchdenny

src/Aspire.Hosting.Oracle/OracleDatabaseBuilderExtensions.cs
danmoseley: @cmdkeen I have the same question..?

mitchdenny: If it isn't possible for the mount to be read only we should remove the read only option entirely.

@DamianEdwards do you know if the other container types we have any issues with read only volumes?

DamianEdwards: I believe some of them work if the volume already exists and has been initialized by a container of the same image already. The scenario is multiple containers using the same volume but only one has write access. That was the only reason we had an option for read only data volumes.

mitchdenny: OK thanks. Let me just confirm that also works with Oracle here and if it does we should keep the parameter (with `isReadOnly = false` as the default).

mitchdenny:
@cmdkeen I updated your PR to remove the `isReadOnly` flag entirely. There may be some voodoo that we can do to make Oracle support read-only volumes but it is safer for GA to not expose the property and pretend we have a working solution for it.

If/when we figure out readonly volume/bindmount support for Oracle then we can add it back in. And if someone is so inclined they can make this work themselves with the general purpose bind mount/volume methods.

mitchdenny (approved):

src/Aspire.Hosting.Oracle/OracleDatabaseBuilderExtensions.cs
```diff
@@ -60,7 +60,7 @@ public static IResourceBuilder<OracleDatabaseResource> AddDatabase(this IResourc
     /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<OracleDatabaseServerResource> WithDataVolume(this IResourceBuilder<OracleDatabaseServerResource> builder, string? name = null)
-        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/opt/oracle/oradata", true);
+        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/opt/oracle/oradata", false);
```
danmoseley, line 63: Nit, nice to name the parameter at the call site when passing a bool

danmoseley (approved):

cmdkeen:
> @cmdkeen I updated your PR to remove the `isReadOnly` flag entirely. There may be some voodoo that we can do to make Oracle support read-only volumes but it is safer for GA to not expose the property and pretend we have a working solution for it.
> 
> If/when we figure out readonly volume/bindmount support for Oracle then we can add it back in. And if someone is so inclined they can make this work themselves with the general purpose bind mount/volume methods.

Very happy, thanks for getting this in and sorted everyone. Absolutely agree that it is easy enough to do a read only volume via an underlying method if needed.

PR #3921, opened 2024-04-24
[release/8.0] Event Hubs: Update EventProcessorClient component to pull BlobClient from DI (#3293)

eerhardt:
Backport of #3293 to release/8.0

## Customer Impact

On the initial PR of Azure EventHubs there was some usability / composability issues that were addressed in a follow-up PR. Even though the EventHubs component will still be in "preview" when Aspire 8.0.0 is GA, getting feedback on this change is important to ensure we have the right API for the EventHubs component. If we don't take this change, we won't be able to get true feedback on the API.

It also makes porting other changes (like the tests in https://github.com/dotnet/aspire/pull/3919) to reduce conflicts and "drift" between main and release/8.0.

## Testing

The unit tests were updated for the new functionality. playground app is updated as well.

## Risk

Low. EventHubs will still be in preview when 8.0.0 ships. There is some internal refactoring to the common Azure shared code, but it doesn't change behavior.

## Regression?
No
___
* update sample EH project launchsettings; add blob component for processor

* update azurecomponent to expose serviceprovider; update associated components; update processor for DI retrieval of blob client

* throw if blobclient cannot be retrieved from DI (this is a breaking change from P5)

* fix tests

* PR feedback

* More PR feedback

* Rename BlobClientConnectionName to BlobClientServiceKey.

Remove the need to specify a BlobClientServiceKey, and use an unkeyed client if none is provided.

* fix tests

* fixup formatting

---------
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3921)

### DIFF
--- a/playground/AspireEventHub/EventHubs.AppHost/Properties/launchSettings.json
+++ b/playground/AspireEventHub/EventHubs.AppHost/Properties/launchSettings.json
@@ -1,15 +1,44 @@
 {
   "$schema": "http://json.schemastore.org/launchsettings.json",
+
   "profiles": {
+    "https": {
+      "commandName": "Project",
+      "dotnetRunMessages": true,
+      "launchBrowser": true,
+      "applicationUrl": "https://localhost:15887;http://localhost:15888",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "DOTNET_ENVIRONMENT": "Development",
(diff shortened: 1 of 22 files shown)


joperezr (approved):

danmoseley:
@mitchdenny  do you want to be other sign off?

mitchdenny (approved):

PR #3906, opened 2024-04-24
Naming Convention: Refactors `AddAzureCosmosClient` and `AddKeyedAzureCosmosClient` API Names

kundadebdatta:
- Refactors `AddAzureCosmosClient` and `AddKeyedAzureCosmosClient` API Names
- Use `/` delimiter for cosmos user agent, instead of `|`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3906)

PR #3849, opened 2024-04-19
[release/8.0] API changes for components

sebastienros:
Backport of #3736 to release/8.0

/cc @sebastienros

## Customer Impact

Customers will have to change their configuration and code to reflect the new names and defaults of properties. No behavioral changes.

## Testing

Unit tests are covering the API changes.

## Risk

Should be low since only the API names are changing. The new default values however introduce a risk if the test coverage was lacking.

## Regression?

No.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3849)

PR #3934, opened 2024-04-24
[release/8.0] Naming Convention: Refactors AddAzureCosmosClient and AddKeyedAzureCosmosClient API Names

eerhardt:
Backport of #3906 to release/8.0

## Customer Impact

After reviewing with the Cosmos team, they decided that the runtime component APIs for Cosmos shouldn't use "CosmosDB" and instead just be "Cosmos". The Hosting APIs are left as "CosmosDB" to be consistent with the Azure.Provisioning APIs.

Also, they decided to use `/` as the delimiter in the User-Agent instead of `|`.

API Fit and Finish

## Testing

Tests are updated to the new APIs.

## Risk

Low. Just a rename and a delimiter change.

## Regression?
No
_____
- Refactors `AddAzureCosmosClient` and `AddKeyedAzureCosmosClient` API Names
- Use `/` delimiter for cosmos user agent, instead of `|`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3934)

### DIFF
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
@@ -8,7 +8,7 @@ using Newtonsoft.Json;
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
-builder.AddAzureCosmosDBClient("cosmos");
+builder.AddAzureCosmosClient("cosmos");
 builder.AddCosmosDbContext<TestCosmosContext>("cosmos", "ef");
 
 var app = builder.Build();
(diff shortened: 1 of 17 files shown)


Pilchie (approved):
Thanks @eerhardt! :shipit:

mitchdenny:
Do you think we should add obsoletes for the old methods?

eerhardt:
> Do you think we should add obsoletes for the old methods?

No. Adding new "Obsolete" attributes in our `8.0.0` GA release is more work than its worth. We didn't in https://github.com/dotnet/aspire/pull/3849 and that was much more breaking.

mitchdenny (approved):

davidfowl (approved):

Issue #2276, opened 2024-02-16
Review container image versions for .NET Aspire 8.0 preview 6


mitchdenny:
This is a recurring issue for reviewing container image dependencies for upcoming releases. Please follow the following process:

- [x] Review all ContainerImageAnnotation usages and determine whether the container image tag that is being used needs to be updated for the upcoming release.
- [ ] If updates are required update the image tag, doc comments, and test cases.
- [ ] Create a copy of this issue for the next release and assign it to the appropriate milestone.

timheuer:
tbh when I was navigating the code I kinda wanted all these values in some type of constants location :-)

mitchdenny:
Yeah. Not a bad idea. Next time we update them we'll do this. But it'll be marked internal because they are going to change every month ;)

danmoseley:
when closing this please create another issue for next preview.

mitchdenny:
@radical I noticed you were going through extracting some of the hardcoded image tag version strings into a separate static class? If so could you check to see whether the image versions need to be updated at the same time?

danmoseley:
@radical what remains here now https://github.com/dotnet/aspire/pull/3847 was merged?
Well once that is in main  as well.

mitchdenny:
I think this can now be closed for GA.

PR #3948, opened 2024-04-25
Update Docker Image Tags

### LINKED ISSUE #3933: Update docker images to use semantic version tags
The plan in https://github.com/dotnet/aspire/issues/1997#issuecomment-1924592452 was to use a "floating version" tag for docker images where we could take new security/patch updates automatically but not take breaking changes.

If the underlying docker image follows https://semver.org/, this would mean we would ideally depend on the `major` version tag of the docker image, or `major.minor` if available and security/critical bug fixes were released on that tag.

Some of our hard-coded tags use this model today. Some of the underlying docker images don't support `major` or `major.minor` tags.

But there are some docker images that have semantic versioned images, but we are using too specific versioned of a tag, meaning that we won't pick up security and critical bug fixes. The ones I'm aware of are:

https://github.com/dotnet/aspire/blob/bade408eecae2cb080b6de111c97bc290bd2c776/src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs#L10

https://github.com/dotnet/aspire/blob/bade408eecae2cb080b6de111c97bc290bd2c776/src/Aspire.Hosting.MySql/MySqlContainerImageTags.cs#L10

https://github.com/dotnet/aspire/blob/bade408eecae2cb080b6de111c97bc290bd2c776/src/Aspire.Hosting.PostgreSQL/PostgresContainerImageTags.cs#L10

https://github.com/dotnet/aspire/blob/bade408eecae2cb080b6de111c97bc290bd2c776/src/Aspire.Hosting.Redis/RedisContainerImageTags.cs#L10

https://github.com/dotnet/aspire/blob/bade408eecae2cb080b6de111c97bc290bd2c776/src/Aspire.Hosting.Seq/SeqContainerImageTags.cs#L10

We should update these images to follow the above plan.

See also https://github.com/dotnet/aspire/pull/3839#discussion_r1578541000

cc @mitchdenny @timheuer

eerhardt:
Following this policy:

- never 'latest' unless absolutely no other option (and then we should evaluate why that is with the owner)
- major.minor
- if no semver, choose most recent version

Containers that don't follow this pattern:

- CosmosDB Emulator uses 'latest'
- Azure Storage Azurite doesn't have major.minor
- Kafka doesn't have major.minor
- Oracle doesn't have major.minor
- Qdrant doesn't have major.minor
- Redis commander uses 'latest' (hasn't been updated in 3 years)
- SqlServer has `<year>-latest`, which is roughly equivalent to major.minor

Fix #3933
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3948)

### DIFF
--- a/src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs
+++ b/src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs
@@ -7,8 +7,8 @@ internal static class MongoDBContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/mongo";
-    public const string Tag = "7.0.8";
+    public const string Tag = "7.0";
     public const string MongoExpressRegistry = "docker.io";
-    public const string MongoExpressImage = "mongo-express";
-    public const string MongoExpressTag = "1.0.2-20";
+    public const string MongoExpressImage = "library/mongo-express";
+    public const string MongoExpressTag = "1.0";
 }
--- a/src/Aspire.Hosting.MySql/MySqlContainerImageTags.cs
+++ b/src/Aspire.Hosting.MySql/MySqlContainerImageTags.cs
@@ -7,5 +7,5 @@ internal static class MySqlContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/mysql";
-    public const string Tag = "8.3.0";
+    public const string Tag = "8.3";
 }
--- a/src/Aspire.Hosting.Nats/NatsContainerImageTags.cs
+++ b/src/Aspire.Hosting.Nats/NatsContainerImageTags.cs
@@ -7,5 +7,5 @@ internal static class NatsContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/nats";
-    public const string Tag = "2";
+    public const string Tag = "2.10";
 }
--- a/src/Aspire.Hosting.RabbitMQ/RabbitMQContainerImageTags.cs
+++ b/src/Aspire.Hosting.RabbitMQ/RabbitMQContainerImageTags.cs
@@ -7,6 +7,6 @@ internal static class RabbitMQContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/rabbitmq";
-    public const string Tag = "3";
+    public const string Tag = "3.13";
     public const string TagManagement = $"{Tag}-management";
 }
--- a/src/Aspire.Hosting.Redis/RedisContainerImageTags.cs
+++ b/src/Aspire.Hosting.Redis/RedisContainerImageTags.cs
@@ -7,7 +7,7 @@ internal static class RedisContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/redis";
-    public const string Tag = "7.2.4";
+    public const string Tag = "7.2";
     public const string RedisCommanderRegistry = "docker.io";
     public const string RedisCommanderImage = "rediscommander/redis-commander";
     public const string RedisCommanderTag = "latest";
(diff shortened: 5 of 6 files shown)


Zombach:
Shouldn't it be taken into account in tests that the actual version will in most cases be returned as major.minor.revision. Provided that we can set major.minor or just major?

eerhardt:
> Shouldn't it be taken into account in this case for tests that the current version in most cases will be returned as major.minor.revision. Provided we can set major.minor or just major?

I'm not sure I follow your question. What does "in this case for tests" mean?

danmoseley:
Did you review these heuristics with our docker folks like @richlander ?

eerhardt:
> Did you review these heuristics with our docker folks like @richlander ?

I've chatted with @richlander @mthalman and @MichaelSimons about this all separately.

src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs
```diff
@@ -7,8 +7,8 @@ internal static class MongoDBContainerImageTags
 {
     public const string Registry = "docker.io";
     public const string Image = "library/mongo";
-    public const string Tag = "7.0.8";
+    public const string Tag = "7.0";
     public const string MongoExpressRegistry = "docker.io";
-    public const string MongoExpressImage = "mongo-express";
-    public const string MongoExpressTag = "1.0.2-20";
+    public const string MongoExpressImage = "library/mongo-express";
+    public const string MongoExpressTag = "1.0";
```
timheuer, line 13: mongo-express seems to have weird tagging as 1.0 now is NOT 1.0.2-20 FYI here. Not sure we change this based on the principle, but an FYI on how it isn't always mapping

timheuer (approved):

richlander:
I like the direction.

src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs
LaurentGoderre: @eerhardt there are `1.0-20` and `1.0-18` tags that let you choose which version of Node. You can use `1.0` if you don't mind about the underlying Node version. The default now is Node 18 but I suspect that will change in the future.

eerhardt: Thanks for the info. I think sticking to `1.0` makes the most sense for us. As the `1.0` tag gets updated for newer Node versions, we will automatically pick up the new version.

radical:
Should the policy being followed here be documented somewhere for future reference?

eerhardt:
> Should the policy being followed here be documented somewhere for future reference?

You mean in a `.md` file in the repo? Would there be an advantage to that over just referencing this PR and corresponding issue #3933?

radical:
> > Should the policy being followed here be documented somewhere for future reference?
> 
> You mean in a `.md` file in the repo? Would there be an advantage to that over just referencing this PR and corresponding issue #3933?

That issue would get closed eventually though, right? If we follow https://github.com/dotnet/aspire/issues/2276 and create new issues for next releases to track the work, as suggested by @mitchdenny, then this PR+3933 could be referenced in that, so anyone looking at updating the tags would be able to see what was done earlier, and why.

eerhardt:
> That issue would get closed eventually though, right?

Even if it is closed, it can still be referenced as "this is the policy we use".

src/Aspire.Hosting.MongoDB/MongoDBContainerImageTags.cs
mitchdenny: Actually I think there might be a bug with 1.0 around auth. Make sure you test it.

eerhardt: What bug are you thinking? I used the mongo playground app, and was able to hit the mongo-express UI and saw the containers inside of it.

DamianEdwards (approved):

PR #3953, opened 2024-04-25
Update sln templates to special-case CLI instead of VS

### LINKED ISSUE #3951: Aspire solution templates should check for CLI host rather than VS host
The Aspire solution templates currently special case the VS host and don't generate a .sln file in that case. Instead they should special case the CLI host and only generate a .sln file in this case. This will make them aligned with the Blazor solution templates and work correctly in other non-CLI hosts like VS Code C# Dev Kit.

DamianEdwards:
Changes the solution templates so that instead of special-casing VS and not creating the .sln file, they special-case the CLI and only create the .sln file in that case. This matches what the Blazor templates do and makes the templates work correctly in non-VS hosts like C# Dev Kit.

Fixes #3951
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3953)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-empty/.template.config/template.json
+++ b/src/Aspire.ProjectTemplates/templates/aspire-empty/.template.config/template.json
@@ -35,7 +35,7 @@
     {
       "modifiers": [
         {
-          "condition": "(hostIdentifier == \"vs\")",
+          "condition": "(hostIdentifier != \"dotnetcli\" && hostIdentifier != \"dotnetcli-preview\")",
           "exclude": [
             "*.sln"
           ]
@@ -219,7 +219,7 @@
   "primaryOutputs": [
     {
       "path": "AspireApplication.1.sln",
-      "condition": "(hostIdentifier != \"vs\")"
+      "condition": "(hostIdentifier != \"dotnetcli\" && hostIdentifier != \"dotnetcli-preview\")"
     },
     {
       "path": "AspireApplication.1.AppHost\\AspireApplication.1.AppHost.csproj"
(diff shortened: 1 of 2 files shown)


src/Aspire.ProjectTemplates/templates/aspire-empty/.template.config/template.json
```diff
@@ -219,7 +219,7 @@
   "primaryOutputs": [
     {
       "path": "AspireApplication.1.sln",
-      "condition": "(hostIdentifier != \"vs\")"
+      "condition": "(hostIdentifier != \"dotnetcli\" && hostIdentifier != \"dotnetcli-preview\")"
```
smitpatel, line 222: this condition should be equality

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -349,7 +349,7 @@
   "primaryOutputs": [
     {
       "path": "AspireStarterApplication.1.sln",
-      "condition": "(hostIdentifier != \"vs\")"
+      "condition": "(hostIdentifier != \"dotnetcli\" && hostIdentifier != \"dotnetcli-preview\")"
```
smitpatel, line 352: same here, conditions should be equality

radical:
Does anything need to be updated so this case gets verified by CTI?

DamianEdwards:
> Does anything need to be updated so this case gets verified by CTI?

I assume they're already covering C# Dev Kit.

smitpatel (approved):

PR #3959, opened 2024-04-25
Fixes unit test templates for spaces in names

### LINKED ISSUE #3957: Generated test template does not sanitize project name for use as namespace
For example, here `-` is being used as is.
`dotnet new aspire-starter -t xunit.net --use-redis-cache -o asp-starter-with-redis-with-xunit0`

`asp-starter-with-redis-with-xunit0.Tests/WebTests.cs`:
```csharp
namespace asp-starter-with-redis-with-xunit0.Tests;
```

It is correctly changed for other projects: `asp-starter-with-redis-with-xunit0.Web/WeatherApiClient.cs`:
```csharp
namespace asp_starter_with_redis_with_xunit0.Web;
```

cc @ReubenBond @joperezr @DamianEdwards

timheuer:
Fix #3957 and #3974

- Renamed placeholder project name
- Manual validation
- Validated `dotnet new aspire-starter -t xunit.net -o asp-starter-with-redis-with-xunit0`
- Validated `dotnet new` with `aspire-xunit`,`aspire-nunit`,`aspire-mstest` using various hyphenated names
- Follows same pattern as project templates
- Changed to look for new symbol name in SLN file

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3959)


src/Aspire.ProjectTemplates/templates/aspire-mstest/IntegrationTest1.cs
```diff
@@ -1,6 +1,6 @@
 using System.Net;
 
-namespace Aspire.Tests1;
+namespace Aspire.Tests._1;
```
radical, line 3: Why `_1`?

DamianEdwards: Yeah I was going to ask this too.

DamianEdwards: Oh this is required to instruct the template engine which naming convention to use.

timheuer: This is just a pattern that instructs the template engine to do this replacement.  See also: https://github.com/dotnet/aspire/blob/main/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Web/WeatherApiClient.cs#L1

DamianEdwards: I'm curious why it's not needed in the ASP.NET Core templates though https://github.com/dotnet/aspnetcore/blob/eb18cb23cf2be9db70ae799e9b23a74a26e78d0b/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/Controllers/HomeController.cs#L18

DamianEdwards: OK I know what's going on and it's because of something I did way back when we first fixed this issue.

The Starter template uses an odd template `"sourceName"` of `AspireStarterApplication.1` which breaks the template engine's default inference for symbol formatting. We should instead update the `sourceName` to `Aspire.StarterApplication1` and then it should work without the need to specify the `_` explicitly. The standalone test projects don't actually have any issue as they follow the regular naming pattern (like the ASP.NET Core templates do).

I recommend we update all the Aspire templates to follow the `Aspire.TemplateName1` pattern to fix this.

timheuer: I have to claim ignorance here -- because I changed the non-starter ones to be consistent but they aren't actually needed there either as the replacement happens. It must be an aspect of template engine somewhere...I'm asking.

timheuer: https://github.com/dotnet/templating/wiki/Naming-and-default-value-forms

radical: ah `namespace Template._1; //uses `namespace` form`.

Isn't that link suggesting that `AspireStarterApplication.1` would be the correct one, and would allow picking which "form" to use for namespace/classname etc, and `AspireStarterApplication1` would break it?

DamianEdwards: Yeah I made it `AspireStarterApplication.1` when we first hit this issue but it still doesn't make the auto-value forms work. But, `Aspire.StarerApplication1` would work, so we should just change to that to match what ASP.NET Core and others do.

timheuer: > I recommend we update all the Aspire templates to follow the `Aspire.TemplateName1` pattern to fix this.

Want me to make that change here? or cancel this in favor of a PR you're gonna make? Looks like only starter would be the one that would need changing.

timheuer: I think I'd recommend sticking with this change only. The change to the others will affect path names as well and you'll change the csproj reference paths.  Using this convention allows you to use the value forms and keep the 'identity' form for what the user put in.

Example, if we were to make the change to all. and you `dotnet new foo-bar-baz` then the file-paths would not be translated (foo-bar-baz.csproj) but then you'd have sln/project references as foo_bar_baz.csproj.

Honestly it feels safer to look at just this change and keep the others as-is as they are functioning within the template engine value forms (identity + namespace + class).

radical:
`/private/tmp/test/asp-starter-with-redis-with-xunit0/asp-starter-with-redis-with-xunit0.Tests/WebTests.cs(11,131): error CS1525: Invalid expression term ')' [/private/tmp/test/asp-starter-with-redis-with-xunit0/asp-starter-with-redis-with-xunit0.Tests/asp-starter-with-redis-with-xunit0.Tests.csproj]`

line 11: `var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.asp-starter-with-redis-with-xunit0_AppHost>();`

radical:
This works;
```diff
diff --git a/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs b/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
index 26eb9fb66..a0a114dde 100644
--- a/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.Tests/WebTests.cs
@@ -17,7 +17,7 @@ public class WebTests
     public async Task GetWebResourceRootReturnsOkStatusCode()
     {
         // Arrange
-        var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.AspireStarterApplication.1_AppHost>();
+        var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.AspireStarterApplication._1_AppHost>();
         await using var app = await appHost.BuildAsync();
         await app.StartAsync();
```

src/Aspire.ProjectTemplates/templates/aspire-mstest/IntegrationTest1.cs
DamianEdwards: I admit I'm a bit lost now lol.
Are you saying that if you change the form to `Aspire.StarterApplication1` it breaks? Why is it that the ASP.NET Core and Blazor templates work? Also how can you run Explorer with file extensions hidden!!? 😱

DamianEdwards (approved):

timheuer:
@radical I implemented the last one you found...did your local tests find anything else in the template tests?

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.sln
```diff
@@ -8,7 +8,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AspireStarterApplication.1.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AspireStarterApplication.1.ApiService", "AspireStarterApplication.1.ApiService\AspireStarterApplication.1.ApiService.csproj", "{9FEB877E-015D-4E20-AE63-06C596E242E4}"
 EndProject
-#if (CreateTestsProject)
+#if (TestFramework != 'None')
```
DamianEdwards, line 11: Good catch! I assume you verified this fixes it.

DamianEdwards: @radical do we have test coverage the starter template creation? Can we add this as a case?

timheuer: Yep. I assume the conditionals may be case-insensitive, but matched the literal default value in template config and verified 'None' and 'none' both work

DamianEdwards (approved):

DamianEdwards:
@timheuer looks like you'll need to manually backport

src/Aspire.ProjectTemplates/templates/aspire-starter/AspireStarterApplication.1.sln
radical: What's the case that was caught here? Using `none` type as the framework generated a test project when it shouldn't have? I check for non-existence of a test project in case of `none` but I didn't see it failing.

timheuer: In the current shipping templates the option is CreateTestsProject since there was only one type: xunit.  Now that we added unit test framework choice, the option changed to TestFramework and is either none (no arg provided so no tests projects), xunit.net, nunit, or mstest

PR #3961, opened 2024-04-25
Set assembly metadata with build configuration for AppHost and apply to projects in CLI mode

danegsta:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3961)


src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
```diff
@@ -214,6 +214,10 @@ namespace Projects%3B
         <_Parameter1>apphostprojectpath</_Parameter1>
         <_Parameter2>$(MSBuildProjectDirectory)</_Parameter2>
       </AssemblyAttribute>
+      <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
+        <_Parameter1>apphostconfiguration</_Parameter1>
+        <_Parameter2>$(Configuration)</_Parameter2>
+      </AssemblyAttribute>
```
eerhardt, line 220:
```suggestion
```

We don't need to do this, the .NET SDK already does it for us:

https://github.com/dotnet/sdk/blob/cbed36adb0e24a8f3c33ba88a74f101a87410e13/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L81-L83

src/Aspire.Hosting/DistributedApplicationOptions.cs
```diff
@@ -89,6 +93,12 @@ public DistributedApplicationOptions()
         return appHostAssembly;
     }
 
+    private string? ResolveConfiguration()
+    {
+        var assemblyMetadata = Assembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
+        return GetMetadataValue(assemblyMetadata, "AppHostConfiguration");
```
eerhardt, line 99:
```suggestion
        return Assembly?.GetCustomAttribute<AssemblyConfigurationAttribute>()?.Configuration;
```

src/Aspire.Hosting/DistributedApplicationOptions.cs
```diff
@@ -62,8 +66,7 @@ public DistributedApplicationOptions()
 
     private string? ResolveProjectDirectory()
     {
-        var assemblyMetadata = Assembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
-        return GetMetadataValue(assemblyMetadata, "AppHostProjectPath");
+        return Assembly?.GetCustomAttribute<AssemblyConfigurationAttribute>()?.Configuration;
```
eerhardt, line 69: I think this is the wrong method.

danegsta: Yeah, realized just after I posted in chat

tests/Aspire.Hosting.Tests/Dcp/ApplicationExecutorTests.cs
```diff
@@ -88,7 +88,7 @@ public async Task EndpointPortsExecutableNotReplicatedProxiedPortSetNoTargetPort
         Assert.True(dcpExe.TryGetAnnotationAsObjectList<ServiceProducerAnnotation>(CustomResource.ServiceProducerAnnotation, out var spAnnList));
```
eerhardt, line 88: Is there any way to test this change?

joperezr: @radical discovered this issue by writing an end-to-end test with his workload infrastructure. That is not in yet, but he is aiming on pushing a PR up next week.

radical: Tests pass with the changes from this PR added to my workload-tests branch.

joperezr:
@danegsta is this PR stale? should it be closed?

danegsta:
@joperezr this is likely stale at this point; did we get a different workaround for the release mode problem merged?

joperezr:
I believe so but not 100% sure, I thought @mitchdenny came up with a different fix? I can test real quick

joperezr:
Just tested this and it is in fact not yet fixed. Looks like this is one thing we'll want to take for 8.1. There was some conversation in chat around using special attributes, assuming all of the feedback is already incorporated with this PR?

danegsta:
Yeah, it's been updated to honor the built-in `AssemblyConfigurationAttribute` the SDK applies, rather than rolling our own assembly attribute via targets.

eerhardt (approved):
These changes LGTM

Issue #3949, opened 2024-04-25
Removing IVT where possible. To test the product as a user


Zombach:
IVT is used in Redis, MySql, MongoDB and others projects.

https://github.com/dotnet/aspire/pull/3839#discussion_r1578518323

PR #3839, opened 2024-04-19
Add Garnet Resource

Zombach:
Implemented support request from #3428 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3839)

PR #3982, opened 2024-04-27
Removing IVT where possible Attempt 2

Zombach:
https://github.com/dotnet/aspire/issues/3949

Made some minor edits and also cleaned up IVT if possible.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3982)


Zombach:
@eerhardt I hope it will be better this way

davidfowl:
IVT is fine for tests. We don't want many of these APIs to be public. Each of these projects should have their own test project instead of all using the same test project.

@eerhardt we shouldn't be removing IVT to our tests, just between product code.

src/Aspire.Hosting.MySql/PhpMyAdminConfigWriterHook.cs
```diff
@@ -6,8 +6,10 @@
 
 namespace Aspire.Hosting.MySql;
 
-internal sealed class PhpMyAdminConfigWriterHook : IDistributedApplicationLifecycleHook
+/// <inheritdoc/>
+public class PhpMyAdminConfigWriterHook : IDistributedApplicationLifecycleHook
```
eerhardt, line 10: We shouldn't be making things public for this.

https://github.com/dotnet/aspire/pull/3839#discussion_r1579781437

>> Is it worth creating a separate PR to fix this?

> If it is simple, like the case here where it is just trying to reuse constant strings, sure.

eerhardt:
>  we shouldn't be removing IVT to our tests

We don't need to use IVT just to share a couple constant strings, like the docker image name and tag. Many of these tests just need to share the constant strings, and we can just compile them into the tests, like we do elsewhere.

For more complicated things we can leave IVT.

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -256,10 +256,16 @@ public void WithDataVolumeAddsPersistenceAnnotation()
...
-        Assert.Equal(TimeSpan.FromSeconds(60), persistenceAnnotation.Interval);
-        Assert.Equal(1, persistenceAnnotation.KeysChangedThreshold);
+        var args = new List<object>();
+        foreach (var argsAnnotation in argsCallbacks)
+        {
+            Assert.NotNull(argsAnnotation.Callback);
+            argsAnnotation.Callback(new CommandLineArgsCallbackContext(args));
+        }
+
+        Assert.Equal("--save 60 1".Split(" "), args);
```
eerhardt, line 268: I like testing like this. I think we should keep this change, it actually tests more like a customer. Call the API, and observe the change in behavior.

src/Aspire.Hosting.MySql/PhpMyAdminConfigWriterHook.cs
Zombach: Yes OK. I canceled it

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -12,6 +12,7 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class MongoDBBuilderExtensions
 {
+    private const string MongoDBContainerDataDirectory = "/var/mongodb/data";
```
eerhardt, line 15: Please revert the changes to this file.

1. We shouldn't be making unnecessary/unrelated changes in a PR. Keep each PR to a single change.
2. This isn't correct. From https://stackoverflow.com/questions/35400740/how-to-set-docker-mongo-data-volume it needs to stay `/data/db`.

Have you tested if the data volumes still work?

Zombach: Ready. Canceled.
Sorry. This is because I want to do a little more. I fall for this all the time.
Although I understand perfectly well. What one PR needs to do to solve one problem.

Zombach: > Have you tested if the data volumes still work?

Honestly, no. I just saw that the tests were working. I didn’t run it separately to check data volumes

eerhardt (approved):

eerhardt:
Thanks for the contribution!

PR #3992, opened 2024-04-28
#3986 Rename of yarp package to clarify functionality

### LINKED ISSUE #3986: Reworking of Microsoft.Extensions.ServiceDiscoveery packages
Currently there is 2 quite different use cases achieved by the Microsoft.Extensions.ServiceDiscoveery.* packages and to clear things up packages should be split renamed before GA release.

The final structure invisioned is as follows:
-

thompson-tomo:
Move yarp to be under service resolution naming rather than service discovery as it is not actually using yarp for service discovery.

closes #3986 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3992)


danmoseley:
@ReubenBond  thoughts? This is breaking, right?

thompson-tomo:
yes @danmoseley it would be breaking due to namespace & package name change but prior to GA is the best time to do this.

Key thing for me is yarp is not actually being used for ServiceDiscovery with this package.

davidfowl:
Yes, its unlikely we'll take this PR as this code already went through API review.

danmoseley:
Cc @mitchdenny 

Yeah GA branch is now ship stoppers only pretty much. I think this just came too late.

thompson-tomo:
If it is too late perhaps we split this in 2:

- Rename package as 1 PR for GA branch but leave namespace
- In v9 we update the namespace to match the package

This would make the change easier for users if we don't want to change both package and namespace for intial GA (my preference).

davidfowl:
It's too late and I don't know why we'd make this change at all.

thompson-tomo:
Let me phrase it another way, when a user installs a service discovery package the clear expectation is that the provider named ie YARP is the source where services are discovered from. This is how the DNS package works. The yarp package implements a different interface to the DNS package and is actually simply resolving the urls for the destinations specified in the yarp config.

davidfowl:
That doesn’t change what I said. There was an APi review process that was followed. Thats the process you go through again with your valid feedback and the committee decide. This PR and the associated issues doesn’t follow any of the process.

danmoseley:
I'm closing this PR for discussion to occur in the issue.

PR #4013, opened 2024-04-29
Fix safari column header heights

### LINKED ISSUE #3958: Dashboard UI issues with Safari
macOS Sonoma 14.4.1
Safari Version 17.4.1 (19618.1.15.11.14)
 
This is with a build from `main` with `HEAD` at 5fec81174cd305c806f66d4a532188f49bace69c.
1. The headers for the last few columns are at the bottom
2. The urls are not clickable, and neither are the View buttons.

Screenshot 2024-04-25 at 15 45 09

The following screenshot shows that the *cell* for the header row is too tall, and covers the other cells of the same column.

Screenshot 2024-04-25 at 17 22 02
Related issue: https://github.com/dotnet/aspire/issues/2272

cc @timheuer @adamint @JamesNK

adamint:
Fixes #3958 

Taken from the comment I made in the css:

fluent-data-grid-cells have a height of 100%, which should correspond to the height of their parent fluent-data-grid-row.
On Safari only, this is not the case; height actually computes to the grid height (incorrectly). We can force column headers to take the height of their parent by inheriting the height of their row.

**Before, Safari**
image

**Before, Chrome/Edge**
image

**After, Safari**
image

**After, Chrome/Edge**
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4013)

### DIFF
--- a/src/Aspire.Dashboard/wwwroot/css/app.css
+++ b/src/Aspire.Dashboard/wwwroot/css/app.css
@@ -429,3 +429,14 @@ fluent-data-grid-cell.no-ellipsis {
 .endpoint-button::part(control) {
     padding: calc(((var(--design-unit) * 0.5) - var(--stroke-width)) * 1px) calc((var(--design-unit) - var(--stroke-width)) * 1px);
 }
+
+.column-header {
+    /*
+        fluent-data-grid-cells have a height of 100%, which should correspond to
+        the height of their parent fluent-data-grid-row.
+        On Safari only, this is not the case; height actually computes to the grid height.
+        We can force column headers to take the height of their parent by inheriting height
+        of their row.
+     */
+    height: inherit !important;
+}


kvenkatrajan (approved):

vnbaaij (approved):

tlmii:
Testing this here locally on windows with a couple browsers and it seems fine. I can't validate the Safari side. It'd be nice to see if we can find an actual bug logged on Safari about this (both from a tracking perspective, so we can undo the hack if it ever gets fixed, and from the perspective of making sure we fully understand why Safari is behaving the way it is). But this seems to fix it for us right now. Thanks @adamint for digging into this!

tlmii (approved):

kvenkatrajan:
@danmoseley/ @joperezr are you ok with backporting this to to 8.0 for GA?

PR #4015, opened 2024-04-29
ServiceEndpoint.ToString() omit zero port

ReubenBond:
When representing an endpoint as a string, it is better to omit a zero-value port than to include it. A value of zero means that the port was not specified by the provider and therefore the default value should be used. This is typical for DNS, since ports are not specified by DNS (excluding DNS SRV). This is required for YARP when using the DNS resolver.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4015)


danmoseley:
We need to get this in by tomorrow if we take it... can you get a review and port? I don't have domain knowledge to review

davidfowl (approved):

PR #4030, opened 2024-04-30
Add configureContainer argument to WithPhpMyadmin.

mitchdenny:
Adding this API for consistency. This is a breaking change that we want to make now so we don't need to make a breaking change after GA.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4030)


src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -74,16 +74,32 @@ public static IResourceBuilder<MySqlDatabaseResource> AddDatabase(this IResource
...
+        configureContainer?.Invoke(phpMyAdminContainerBuilder);
 
         return builder;
     }
 
+    /// <summary>
+    /// Configures the host port that the PGAdmin resource is exposed on instead of using randomly assigned port.
+    /// </summary>
+    /// <param name="builder">The resource builder for PGAdmin.</param>
+    /// <param name="port">The port to bind on the host. If <see langword="null"/> is used random port will be assigned.</param>
```
eerhardt, line 93:
```suggestion
    /// <param name="port">The port to bind on the host. If <see langword="null"/> is used, a random port will be assigned.</param>
```

eerhardt (approved):
LGTM

Issue #3980, opened 2024-04-26
EventHub component does not have health check


oising:
There is no health check implemented for the Event Hub component.

@ReubenBond @davidfowl Please assign me.

PR #4050, opened 2024-05-01
Add health checks for Event Hub component

oising:
Add healthchecks for eventhub component

closes https://github.com/dotnet/aspire/issues/3980

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4050)

### DIFF
--- a/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
+++ b/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
@@ -47,6 +47,11 @@ public abstract class AzureMessagingEventHubsSettings : IConnectionStringSetting
     public TokenCredential? Credential { get; set; }
 
     /// <summary>
+    /// Gets or sets a boolean value that indicates whether the health checks are disabled or not.
+    /// </summary>
+    public bool DisableHealthChecks { get; set; }
+
+    /// <summary>
     /// Gets or sets a boolean value that indicates whether the OpenTelemetry tracing is disabled or not.
     /// </summary>
     /// <remarks>
--- a/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
+++ b/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
@@ -47,8 +47,11 @@ public abstract class AzureMessagingEventHubsSettings : IConnectionStringSetting
     public TokenCredential? Credential { get; set; }
 
     /// <summary>
-    /// Gets or sets a boolean value that indicates whether the health checks are disabled or not.
+    /// Gets or sets a boolean value that indicates whether the Azure Event Hubs health check is disabled or not.
     /// </summary>
+    /// <value>
+    /// The default value is <see langword="false"/>.
+    /// </value>
     public bool DisableHealthChecks { get; set; }
 
     /// <summary>
--- a/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
+++ b/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
@@ -93,6 +93,8 @@ internal abstract class EventHubsComponent<TSettings, TClient, TClientOptions> :
                         $"A {typeof(TClient).Name} could not be configured. Ensure a valid EventHubName was provided in " +
                         $"the '{configurationSectionName}' configuration section, or include an EntityPath in the ConnectionString.");
                 }
+                // The connection string has an EventHubName, but we'll set this anyway so the health check can use it
+                settings.EventHubName = props.EventHubName;
             }
         }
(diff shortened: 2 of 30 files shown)


src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -101,4 +103,47 @@ protected static string GetNamespaceFromSettings(AzureMessagingEventHubsSettings
                 $"the '{configurationSectionName}' configuration section.");
         }
     }
+    protected sealed class AzureEventHubHealthCheck : IHealthCheck, IAsyncDisposable
```
danmoseley, line 106: nit, newline above

actually also src/Components/** don't seem to be completely consistent but in general seem to prefer only one non-nested class per file. Eg `SeqHealthCheck` is its own file, `NatsHealthCheck` etc.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -101,4 +103,47 @@ protected static string GetNamespaceFromSettings(AzureMessagingEventHubsSettings
...
+        public AzureEventHubHealthCheck(TSettings settings)
+        {
+            var probeOptions = new EventHubProducerClientOptions
+            {
+                Identifier = $"AspireEventHubHealthCheck-{settings.EventHubName}",
+                RetryOptions = new EventHubsRetryOptions()
+                {
+                    Mode = EventHubsRetryMode.Exponential,
+                    MaximumRetries = 3,
+                    Delay = TimeSpan.FromMilliseconds(800),
```
danmoseley, line 119: where is 800 from?

oising: I'll hoist this into Aspire:Settings :)

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
oising: I get you, but it's only used in the derivations of EventHubsComponent itself, so I find private-protected seems to be the right visibility in this case.

oising:
image

If you look at the eventhubs playground API project (it's the only webapp based one capable of exposing /health) you can trace the logic. 

I also added a DisableHealthChecks setting to align with other components.

joperezr:
@oising thanks for the PR and for your contribution. I noticed that your PR is targeting 8.0 branch, which we are no longer taking changes for. Is there a specific reason why you picked `release/8.0` as the target branch and would you be okay with us retargeting to `main` branch instead?

oising:
> @oising thanks for the PR and for your contribution. I noticed that your PR is targeting 8.0 branch, which we are no longer taking changes for. Is there a specific reason why you picked `release/8.0` as the target branch and would you be okay with us retargeting to `main` branch instead?

Retarget as you please!

joperezr:
Actually, I was about to retarget, but because this has changes that are only on release/8.0 branch that we specifically don't want in main, things wouldn't go so well. Due to that, I think the best way would be to reset and cherry-pick the changes directly into main's tip, and then update this PR. Want me to help with that?

src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
```diff
@@ -46,6 +46,11 @@ public abstract class AzureMessagingEventHubsSettings : IConnectionStringSetting
     /// </summary>
     public TokenCredential? Credential { get; set; }
 
+    /// <summary>
+    /// Gets or sets a boolean value that indicates whether the health checks are disabled or not.
+    /// </summary>
```
sebastienros, line 51:
```suggestion
    /// <summary>
    /// Gets or sets a boolean value that indicates whether the Azure Event Hubs health check is disabled or not.
    /// </summary>
    /// <value>
    /// The default value is <see langword="false"/>.
    /// </value>
```

sebastienros: I know it's a lame request but it makes everything the same. Will also require updating the schema files.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
sebastienros: @eerhardt I don't think there is a precedent for health check retry options. I found `MongoDBSettings.HealthCheckTimeout` only.

Would adding a `RetryOptions HealthCheckRetryOptions` property in `AzureMessagingEventHubsSettings` work here?

oising:
Created new PR targeting main (cherry pick) at https://github.com/dotnet/aspire/pull/4139

/cc @joperezr @danmoseley @sebastienros

PR #4052, opened 2024-05-01
Move Cancel call outside of lock and add additional error handling

ReubenBond:
This PR moves a `CancellationTokenSource.Cancel()` call outside of a `lock`, as suggested by @davidfowl, and surrounds two callback invocations with try/catch + logging.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4052)

### DIFF
--- a/src/Microsoft.Extensions.ServiceDiscovery/ServiceEndpointWatcher.cs
+++ b/src/Microsoft.Extensions.ServiceDiscovery/ServiceEndpointWatcher.cs
@@ -77,8 +77,10 @@ internal sealed partial class ServiceEndpointWatcher(
         async ValueTask<ServiceEndpointSource> GetEndpointsInternal(CancellationToken cancellationToken)
         {
             ServiceEndpointSource? result;
+            var disposalToken = _disposalCancellation.Token;
             do
             {
+                disposalToken.ThrowIfCancellationRequested();
                 cancellationToken.ThrowIfCancellationRequested();
                 await RefreshAsync(force: false).WaitAsync(cancellationToken).ConfigureAwait(false);
                 result = _cachedEndpoints;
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

src/Microsoft.Extensions.ServiceDiscovery/ServiceEndpointWatcher.cs
```diff
@@ -244,10 +253,17 @@ private void SchedulePollingTimer()
...
     public async ValueTask DisposeAsync()
     {
-        lock (_lock)
+        try
         {
             _disposalCancellation.Cancel();
+        }
+        catch (Exception exception)
+        {
+            _logger.LogError(exception, "Error cancelling disposal cancellation token.");
```
danmoseley, line 262: Just curious, what can cause this error

ReubenBond: @danmoseley When you call `CancellationTokenSource.Cancel()`, exceptions thrown by the callbacks are propagated to the caller. See this doc comment fragment:
```xml
/// <exception cref="AggregateException">An aggregate exception containing all the exceptions thrown
/// by the registered callbacks on the associated <see cref="CancellationToken"/>.</exception>
```

The online docs also have this remark:
> We recommend that cancelable operations and callbacks registered with [CancellationToken](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-8.0) not throw exceptions.

But, of course, we cannot guarantee that.

ReubenBond: The log line could say "error thrown by cancellation callback", but that would be presumptuous

davidfowl: Is it internal? The token?

ReubenBond: We pass the cancellation token to providers, and we don't know what they will do since that is an extensibility point.

PR #4064, opened 2024-05-02
Adding extra constraint to WithEndpoint

### LINKED ISSUE #4059: WithEndpoint shows up on all resources and isn't constrained to IResourceWithEndpoints
This shows an azure storage resource which doesn't support adding endpoints:

image

mitchdenny:
Fixes #4059 

Some of the `WithEndpoint` overloads were not constrained. This PR adds the appropriate constraint. There is some splash damage to the Orleans-specific resource types here. It seems strange to require `IResourceWithEndpoints` for something that is being configured as being an orleans client, although practically anything that is a client would be a resource that has this interface, so I don't think the constraint is a problem here.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4064)


davidfowl:
This is a breaking change. Do we obsolete?

mitchdenny:
I don't think we can in this case because they would be ambiguous.

src/Aspire.Hosting.Orleans/OrleansServiceClientExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static class OrleansServiceClientExtensions
     public static IResourceBuilder<T> WithReference<T>(
         this IResourceBuilder<T> builder,
         OrleansServiceClient orleansServiceClient)
-        where T : IResourceWithEnvironment
+        where T : IResourceWithEnvironment, IResourceWithEndpoints
```
ReubenBond, line 24: Why does referencing a resource imply that the receiver exposes endpoints? It seems overly constrained. Am I missing something obvious?

ReubenBond: If anything, it seems inverted and that the resource being referenced should have endpoints, but even that is overly constrained: resources should define what it means to be referenced. In the case of Orleans, that means that they can receive the configuration required to join an Orleans cluster.

mitchdenny: I agree that this is not ideal ... but here is the WHYs :)

1. `WithReference(IResourceBuilder<T>, OrleansServiceClient)` is constrained with `T: IResourceWithEndpoints` because it calls `WithOrleansReference(IResourceBuilder<T>` which is constrained with `T: IResourceWithEndpoints`.
2. `WithOrleansReference(IResourceBuilder<T>)` is constrained with `T: IResourceWithEndpoints` because it calls `WithEndpoint(IResourceBuilder<T>, ...) which is constrained with `T: IResourceWithEndpoints`.

How do you you think we should evolve `WithOrleansReference` here?

ReubenBond: We should split the WithOrleansReference method into two, one for the client (which doesn't have endpoints) and one for the silos (which do)

mitchdenny: I think that is a good suggestion. Although this would be a breaking change. So we probably need two methods and obsolete this one rather than retaining it.

mitchdenny: I'm thinking we make that Orleans refactoring a different PR and add the constraint for now in this one.

mitchdenny:
Closing in favor of #4071

src/Aspire.Hosting.Orleans/OrleansServiceClientExtensions.cs
mitchdenny: Created an issue to track this: https://github.com/dotnet/aspire/issues/4116

I think it is low pri for now since we need to stretch pretty hard to come up with a scenario where we are using WithReference on something that isn't a .NET project to add orleans config.

PR #4053, opened 2024-05-01
[release/8.0] Port translations from main

joperezr:
cc: @danmoseley 

These are selectively brought from https://github.com/dotnet/aspire/pull/3924
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4053)

PR #4068, opened 2024-05-02
Port translations from main (#4053)

danmoseley:
This is part of https://github.com/dotnet/aspire/pull/3924 that seems correct (it's already in 8.0, where I got this from). But not all of that PR, which seems to have bugs eg dropping the term Aspire and reversing the localization of tr to English.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4068)


joperezr (approved):
LGTM and matches what we backported to release/8.0

PR #4069, opened 2024-05-02
YARP: add special-case for localhost when setting Host value

### LINKED ISSUE #3991: Https with YARP doesn't work for https and the ASP.NET Core Dev certificate
The [ServiceDiscoveryDestinationResolver](https://github.com/dotnet/aspire/blob/104a1004681eaa9958ca74ed7d6a3aceeb31ec6c/src/Microsoft.Extensions.ServiceDiscovery.Yarp/ServiceDiscoveryDestinationResolver.cs#L87) sets the host name to the original host in local dev scenarios and this results in a `RemoteCertificateNameMismatch` since the dev cert expects localhost as the domain name (for SNI).

This can be worked around by forcing the host name to be localhost in dev scenarios but that's not a great solution:

https://github.com/dotnet/aspire/issues/836#issuecomment-2075656462

cc @ReubenBond

ReubenBond:
Fixes #3991

This special-cases the logic used by the YARP Service Discovery integration when an address resolves to localhost:

* If the service resolves to localhost, we *do not* set the Host property
* Add an option to opt-out and *always* respect YARP's Host value (null or not)

The core logic is this:

<details>
<summary>Original PR logic</summary>

```csharp
string? resolvedHost;
if (_yarpOptions.AlwaysUseConfiguredHostValue) // defaults to false
{
    // Always use the configured host value only.
    resolvedHost = originalConfig.Host;
}
else if (IsLocalhost(endpoint.EndPoint))
{
    // Suppress the host value for localhost
    resolvedHost = null;
}
else
{
    // Use the configured Host value if set and fall back to the authority from the input URI.
    resolvedHost = !string.IsNullOrWhiteSpace(originalConfig.Host) ? originalConfig.Host : originalUri.Authority;
}
```

</details>

```csharp
// Use the configured 'Host' value if it is provided.
if (!string.IsNullOrEmpty(originalConfig.Host))
{
    resolvedHost = originalConfig.Host;
}
else if (uri.IsLoopback)
{
    // If there is no configured 'Host' value and the address resolves to localhost, do not set a host.
    // This is to account for non-wildcard development certificate.
    resolvedHost = null;
}
else
{
    // Excerpt from RFC 9110 Section 7.2: The "Host" header field in a request provides the host and port information from the target URI [...]
    // See: https://www.rfc-editor.org/rfc/rfc9110.html#field.host
    // i.e, use Authority and not Host.
    resolvedHost = originalUri.Authority;
}
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4069)

### DIFF
--- a/src/Microsoft.Extensions.ServiceDiscovery.Yarp/ServiceDiscoveryDestinationResolver.cs
+++ b/src/Microsoft.Extensions.ServiceDiscovery.Yarp/ServiceDiscoveryDestinationResolver.cs
@@ -55,7 +55,6 @@ internal sealed class ServiceDiscoveryDestinationResolver(ServiceEndpointResolve
         CancellationToken cancellationToken)
     {
         var originalUri = new Uri(originalConfig.Address);
-        var originalHost = originalConfig.Host is { Length: > 0 } h ? h : originalUri.Authority;
         var serviceName = originalUri.GetLeftPart(UriPartial.Authority);
 
         var result = await resolver.GetEndpointsAsync(serviceName, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.ServiceDiscovery.Yarp/ServiceDiscoveryDestinationResolver.cs
```diff
@@ -118,4 +137,22 @@ private string GetDefaultScheme(Uri originalUri)
             return originalUri.Scheme;
         }
     }
+
+    private static bool IsLocalhost(EndPoint endpoint)
```
davidfowl, line 141: Can this ever be `UriEndpoint("http://127.0.0.1")`

ReubenBond: Fixed using `Uri.IsLoopback`. I added more test coverage to ensure we treat the following as localhost:
* 127.0.0.1
* localhost
* ::1

davidfowl (approved):

PR #4071, opened 2024-05-03
Ensure endpoint names are unique to avoid failures later in execution.

mitchdenny:
Builds upon changes in #4064 

Moves the uniqueness check until after the endpoint is constructed and name is derived (to share the various bits of fallback logic).

Added a test case to cover this.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4071)

### DIFF
--- a/src/Aspire.Hosting/ResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ResourceBuilderExtensions.cs
@@ -423,7 +423,7 @@ public static class ResourceBuilderExtensions
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     /// <exception cref="DistributedApplicationException">Throws an exception if an endpoint with the same name already exists on the specified resource.</exception>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
-    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null) where T : IResource
+    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null) where T : IResourceWithEndpoints
     {
         if (builder.Resource.Annotations.OfType<EndpointAnnotation>().Any(sb => string.Equals(sb.Name, name, StringComparisons.EndpointAnnotationName)))
         {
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -439,6 +434,11 @@ private static void ApplyEndpoints<T>(this IResourceBuilder<T> builder, IResourc
             isExternal: isExternal,
             isProxied: isProxied);
 
+        if (builder.Resource.Annotations.OfType<EndpointAnnotation>().Any(sb => string.Equals(sb.Name, annotation.Name, StringComparisons.EndpointAnnotationName)))
+        {
+            throw new DistributedApplicationException($"Endpoint with name '{annotation.Name}' already exists");
```
davidfowl, line 439: I think we should improve the error message. We default the scheme to the name so they may not even know the endpoint has a name. We should try out best to tell them that it does and that the if no name is specified, the scheme will be used as the name. Then we should recommend that they specify a unique name per endpoint.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -439,6 +434,11 @@ private static void ApplyEndpoints<T>(this IResourceBuilder<T> builder, IResourc
             isExternal: isExternal,
             isProxied: isProxied);
 
+        if (builder.Resource.Annotations.OfType<EndpointAnnotation>().Any(sb => string.Equals(sb.Name, annotation.Name, StringComparisons.EndpointAnnotationName)))
+        {
+            throw new DistributedApplicationException($"Endpoint with name '{annotation.Name}' already exists. Endpoint name may not have been explicitly specified and was derived automatically from scheme argument (e.g. 'http', 'https', or 'tcp;). Multiple calls to WithEndpoint (and related methods) may result in a conflict if name argument is not specified. Each endpoint must have a unique name. For more information on networking in .NET Aspire see: https://aka.ms/dotnet/aspire/networking");
```
timheuer, line 439: typo: 'tcp; -- i assume to be 'tcp'

tests/Aspire.Hosting.Tests/WithEndpointTests.cs
```diff
@@ -94,7 +94,22 @@ public void EndpointsWithTwoPortsSameNameThrows()
                     .WithHttpsEndpoint(3000, 2000, name: "mybinding");
         });
 
-        Assert.Equal("Endpoint with name 'mybinding' already exists", ex.Message);
+        Assert.Equal("Endpoint with name 'mybinding' already exists. Endpoint name may not have been explicitly specified and was derived automatically from scheme argument (e.g. 'http', 'https', or 'tcp;). Multiple calls to WithEndpoint (and related methods) may result in a conflict if name argument is not specified. Each endpoint must have a unique name. For more information on networking in .NET Aspire see: https://aka.ms/dotnet/aspire/networking", ex.Message);
```
timheuer, line 97: Similar tcp; typos in the asserts.

timheuer (changes requested):
typos

mitchdenny:
@timheuer thanks for spotting those!

mitchdenny:
@timheuer can you reset your vote?

timheuer (approved):

Issue #3161, opened 2024-03-26
[tests] `Oracle` test for E2E fails with `ORA-12514: TNS:listener does not currently know of service requested in connect descriptor`


radical:
[Log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-3113-merge-0b98931593fc4882b9/Aspire.EndToEnd.Tests-oracle/1/console.6a5b6981.log?helixlogtype=result). Hit on https://github.com/dotnet/aspire/pull/3113/checks?check_run_id=23082675262 . This is only on the PR since the change hasn't been merged yet.

`[xUnit.net 00:08:30.67] Aspire.EndToEnd.Tests: Retry #9 for 'http://localhost:5281/oracledatabase/verify' due to StatusCode: 500 ReasonPhrase: 'Internal Server Error' Content:`
```json
{
   "detail" : "Oracle.ManagedDataAccess.Client.OracleException (0x80004005): Oracle Communication: Failed to connect to server or failed to parse connect string
 ---> OracleInternal.Network.NetworkException (0x80004005): Oracle Communication: Failed to connect to server or failed to parse connect string
 ---> OracleInternal.Network.NetworkException (0x000030E2): ORA-12514: TNS:listener does not currently know of service requested in connect descriptor
   at OracleInternal.Network.OracleCommunication.SendConnectPacketAndProcessResponse(AddressResolution addrRes)
   at OracleInternal.Network.OracleCommunication.ConnectViaCO(ConnectionOption connOption, AddressResolution addrRes)
   at OracleInternal.Network.OracleCommunication.DoConnect(String tnsDescriptor)
   at OracleInternal.Network.OracleCommunication.Connect(String tnsDescriptor, Boolean doNAHandshake, String IName, ConnectionOption CO)
   at OracleInternal.ServiceObjects.OracleConnectionImpl.Connect(ConnectionString cs, Boolean bOpenEndUserSession, OracleConnection connRefForCriteria, String instanceName)
   at OracleInternal.ConnectionPool.PoolManager`3.Get(ConnectionString csWithDiffOrNewPwd, Boolean bGetForApp, OracleConnection connRefForCriteria, String affinityInstanceName, Boolean bForceMatch)
   at OracleInternal.ConnectionPool.OraclePoolManager.Get(ConnectionString csWithNewPassword, Boolean bGetForApp, OracleConnection connRefForCriteria, String affinityInstanceName, Boolean bForceMatch)
[...]

Issue #3063, opened 2024-03-21
[tests] Move component tests to use testcontainers


radical:
- [x] mongodb
- [x] mysql
- [x] postgresql
- [x] Redis
- [ ] Kafka
- [ ] Seq
- [ ] SqlServer
- [ ] Oracle
- [ ] Cosmos

cc @eerhardt

davidfowl:
@eerhardt @radical did we do this?

eerhardt:
Looks like 2 left from the above list. I'm not sure this is worth keeping open.

radical:
`Seq` doesn't have testcontainers yet. Maybe we can get copilot to do update for Cosmos.

radical:
Stale.

PR #4078, opened 2024-05-03
Oracle OpenTelemetry support

### LINKED ISSUE #4072: Add Oracle OpenTelemetry
I can't comment on https://github.com/dotnet/aspire/issues/2814 as it's locked. ODP.Net 23c has just gone GA, including Oracle.ManagedDataAccess.OpenTelemetry. Waiting on a GA version was blocking adding OTel support for Oracle, would be great to pick this up now it's out. Happy to help / try stuff out.

https://www.nuget.org/packages/Oracle.ManagedDataAccess.Core/#versions-body-tab
https://www.nuget.org/packages/Oracle.ManagedDataAccess.OpenTelemetry#versions-body-tab

cmdkeen:
Closes #4072 by updating to the new ODP.Net version and adding OTel support via Oracle.ManagedDataAccess.OpenTelemetry. As mentioned in the issue this will need a new package and bumped version adding to the NuGet feeds.

Used the tracing settings taken from the [Oracle sample](https://github.com/oracle/dotnet-db-samples/blob/master/samples/opentelemetry/opentelemetry.cs) to ensure that EF Core SQL text shows is provided to the traces.

Added an EndToEnd test using the Npgsql one as a base.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4078)


cmdkeen:
@mitchdenny or @danmoseley are you able to take a look at this?

mitchdenny:
Adding @eerhardt for the Otel side of things.

mitchdenny:
Importing the Otel package to internal feed so we can build this.

playground/OracleEndToEnd/OracleEndToEnd.AppHost/aspire-manifest.json
mitchdenny, line 1: Need to update the manifest.

src/Components/Aspire.Oracle.EntityFrameworkCore/AspireOracleEFCoreExtensions.cs
```diff
@@ -163,6 +164,20 @@ void ConfigureRetry()
...
+        if (!settings.DisableTracing)
+        {
+            builder.Services.AddOpenTelemetry().WithTracing(tracerProviderBuilder =>
+            {
+                tracerProviderBuilder.AddOracleDataProviderInstrumentation(o =>
+                {
+                    o.EnableConnectionLevelAttributes = true;
+                    o.RecordException = true;
+                    o.InstrumentOracleDataReaderRead = true;
+                    o.SetDbStatementForText = true;
```
eerhardt, line 176: This might be a silly question, but why aren't these set to true by default in `AddOracleDataProviderInstrumentation`? Are there other properties that should be set?

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests.cs
```diff
@@ -66,7 +66,7 @@ protected override void SetHealthCheck(OracleEntityFrameworkCoreSettings options
         => options.DisableHealthChecks = !enabled;
```
eerhardt, line 66: Can you add tests for the tracing. They need to be declared at each component level. For example:

https://github.com/dotnet/aspire/blob/9fab2f6bd66866fb210f3b6c1f6347a73c9ab57a/tests/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.Tests/ConformanceTests.cs#L134-L137

src/Components/Aspire.Oracle.EntityFrameworkCore/AspireOracleEFCoreExtensions.cs
cmdkeen: One for the package authors as to why they're not the defaults. I did try originally with just an empty `AddOracleDataProviderInstrumentation()` constructor call but then didn't see any database calls being traced from the E2E playground. This is the same set of properties as used in the [Oracle OTel samples](https://github.com/oracle/dotnet-db-samples/blob/master/samples/opentelemetry/opentelemetry.cs).

Potentially it is because the Oracle OTel library isn't specifically aimed at EF so it might be the way that the Oracle EF package is setting the SQL statement, but hard to tell as it isn't OSS.

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests.cs
cmdkeen: Will do

src/Components/Aspire.Oracle.EntityFrameworkCore/AspireOracleEFCoreExtensions.cs
eerhardt: Do you know any of the authors? Can you tag them here?

cmdkeen: Can't see Alex Keh on here, have raised https://github.com/oracle/dotnet-db-samples/issues/378 to ask

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests.cs
cmdkeen: From what I can see this will require adding the TestContainers.Oracle nuget package to the internal package feed as https://github.com/dotnet/aspire/issues/3063 hasn't covered Oracle yet. Happy to give that a go.

cmdkeen: It also seems that the current .Net Testcontainers is using the XE image, there isn't the equivalent of the Java one having the newer Oracle free module. It also locks some of the config behind private methods that would swapping out the image to support the same image as Hosting uses. A bit of pondering required on the best approach.

src/Components/Aspire.Oracle.EntityFrameworkCore/AspireOracleEFCoreExtensions.cs
alexkeh: @eerhardt @cmdkeen Oracle ODP.NET sets these properties to false by default as a security precaution. We didn't want developers to incorporate the ODP.NET OpenTelemetry and then inadvertently publish sensitive DB info into the trace. We want to make sure there was an explicit intent to do so.

Are other DB OpenTelemetry providers enabling these settings by default? I forget whether SqlClient has the options class members automatically enabled (and I couldn't locate the API doc).

cmdkeen: Hmm - I'm no expert, from looking at the playground output the Postgres and MySQL ones do emit at least some of this data (SQL command, passwordless connection string and db info) by default whereas it looks like the SqlServer one does not.
From my perspective having that level of detail is incredibly helpful. I can see at least four options:

1. Turning on this detail by default, mentioning in the docs
2. Switch back to the library defaults, allow fine grained control of the instrumentation options via OracleEntityFrameworkCoreSettings
3. Switch back to the library defaults, add a "DetailedTracing" property to OracleEntityFrameworkCoreSettings that sets these options with a suitable comment
4. Switch back to the library defaults - [not sure](https://docs.oracle.com/en/database/oracle/oracle-database/23/odpnt/OracleDataProviderInstrumentationOptionsClass.html) if you can change the options easily, otherwise users might need to disable tracing and manually add

I can't see any other examples of tracing configuration being passed through the constructors, and appreciate the extensions are trying to set up sensible defaults. My preference is for either 1 or 3 from above, but happy to take a steer.

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests.cs
cmdkeen: Added and also enabled Testcontainers for Oracle as part of it. Turns out the base Testcontainers instance works fine.

mitchdenny:
> @mitchdenny or @eerharor @eerhardt are you able to add Testcontainers.Oracle to the internal feed as well to get this compiling again.

Importing that package now. Should be there in 30 mins.

cmdkeen:
@eerhardt @sebastienros would you be able to take a look with the failing tests? I've got them working on my machine, and made some progress on working out what's needed on the them running on the build agent (adding the Helix reference helped) but it looks like the ones that use Docker are still all failing. 
I did adjust the TriggerActivity to try and be obvious about connection issues to the database, but it feels like this is something else.

sebastienros:
@radical The [Helix tests](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-4078-merge-ebd6a4e975da420eb1/Aspire.Oracle.EntityFrameworkCore.Tests/1/console.83c564db.log?helixlogtype=result) are failing on things related to this PR, can you sport something wrong with how the new tests are written?

playground/OracleEndToEnd/OracleEndToEnd.ApiService/OracleEndToEnd.ApiService.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    
+    <ProjectReference Include="..\..\..\src\Components\Aspire.Oracle.EntityFrameworkCore\Aspire.Oracle.EntityFrameworkCore.csproj" />
+
```
radical, line 12:
```suggestion
    <ProjectReference Include="$(RepoRoot)src\Components\Aspire.Oracle.EntityFrameworkCore\Aspire.Oracle.EntityFrameworkCore.csproj" />   
```

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests.cs
```diff
@@ -100,4 +118,12 @@ public void DbContextCanBeAlwaysResolved()
 
         Assert.NotNull(dbContext);
     }
+
+    [RequiresDockerFact]
+    public void TracingEnablesTheRightActivitySource()
+    => RemoteExecutor.Invoke(static connectionStringToUse => RunWithConnectionString(connectionStringToUse, obj => obj.ActivitySourceTest(key: null)),
```
radical, line 124:
```suggestion
        => RemoteExecutor.Invoke(static connectionStringToUse => RunWithConnectionString(connectionStringToUse, obj => obj.ActivitySourceTest(key: null)),
```

tests/Aspire.Oracle.EntityFrameworkCore.Tests/ConformanceTests_TypeSpecificConfig.cs
```diff
@@ -8,6 +8,11 @@ namespace Aspire.Oracle.EntityFrameworkCore.Tests;
 
 public class ConformanceTests_TypeSpecificConfig : ConformanceTests
 {
+    public ConformanceTests_TypeSpecificConfig(OracleContainerFixture containerFixture) : base(containerFixture)
+    {
+
```
radical, line 13:
```suggestion
```

radical:
Looking at the test results the first test fails with:
Screenshot 2024-05-10 at 14 39 40

Oracle might be taking a longer time to start up, or failing at some point? I would try to add some waits to confirm that the *db* is up, and accessible. [This](https://dotnet.testcontainers.org/api/wait_strategies/) might be helpful too.

Also, there is https://github.com/dotnet/aspire/issues/3161 but it doesn't use testcontainers, and instead depends on the AppHost starting the container.

playground/OracleEndToEnd/OracleEndToEnd.AppHost/aspire-manifest.json
```diff
@@ -0,0 +1,294 @@
+{
+  "$schema": "https://json.schemastore.org/aspire-8.0.json",
+  "resources": {
+    "pg1": {
```
eerhardt, line 4: This doesn't look correct. Did you just copy-paste this?

src/Components/Aspire.Oracle.EntityFrameworkCore/AspireOracleEFCoreExtensions.cs
eerhardt: The intent of .NET Aspire components is to codify best practices. So developers will fall into a "pit of success" with using the defaults.

I don't think we should be enabling any setting by default if there are security concerns with that setting being on.

playground/OracleEndToEnd/OracleEndToEnd.AppHost/aspire-manifest.json
cmdkeen: Yes, and you're right definitely not accurate. Do I need to generate one or can I just delete the file?

mitchdenny: Regenerate it.

eerhardt:
It looks like we aren't getting the database icon to show up in the dashboard:

image

compare that to SqlServerEndToEnd:
image

The reason appears to be because there is no `peer.service` attribute/tag on the span:

image

vs.

image

@JamesNK - is there a reason why the span has to have `peer.service` for the DB icon to show up? Can't it having `db.system` be enough?

JamesNK:
`peer.service` indicates the span goes to an external app. The value is used to map to a resource with the same address.

JamesNK:
server.address and server.port also cause the icon to be displayed:
https://github.com/dotnet/aspire/blob/09d5efb49e258fe54e15578ff3b37cbbabe48e79/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs#L195-L224

They're both recommend attributes in database trace semantic conventions - https://opentelemetry.io/docs/specs/semconv/database/database-spans/

playground/OracleEndToEnd/OracleEndToEnd.AppHost/aspire-manifest.json
cmdkeen: Done

cmdkeen:
I went through the different non-default settings of the OracleDataProviderInstrumentationOptions, and setting `EnableConnectionLevelAttributes = true` does provide:

image

That results in the addition of db.name, db.user, server.address and server.port being set on the spans.

cmdkeen:
> Looking at the test results the first test fails with: Screenshot 2024-05-10 at 14 39 40
> 
> Oracle might be taking a longer time to start up, or failing at some point? I would try to add some waits to confirm that the _db_ is up, and accessible. [This](https://dotnet.testcontainers.org/api/wait_strategies/) might be helpful too.
> 
> Also, there is #3161 but it doesn't use testcontainers, and instead depends on the AppHost starting the container.

Hi @radical after much bashing away at this I managed to unearth that the tests are failing with:
```
Oracle.ManagedDataAccess.Client.OracleException (0x80004005): ORA-50201: Oracle Communication: Failed to connect to server or failed to parse connect string
 ---> OracleInternal.Network.NetworkException (0x80004005): ORA-50201: Oracle Communication: Failed to connect to server or failed to parse connect string
 ---> OracleInternal.Network.NetworkException (0x80004005): ORA-12514: Cannot connect to database. Service XEPDB1 is not registered with the listener at host 127.0.0.1/127.0.0.1 port 5432. (CONNECTION_ID=c4G0SvajqUmwqX6okJZmQA==)
```

That ORA-12514 code led me to https://github.com/gvenzl/oci-oracle-xe/issues/225 where it seems there might be issues with different networking setups. I don't know anything about Helix - is it potentially doing something along the lines of network=host?

That ORA-12514 code led me to https://github.com/gvenzl/oci-oracle-xe/issues/225 where it seems there might be issues with different networking setups. I don't know anything about Helix - is it potentially doing something along the lines of network=host?

The wait strategy in place on the container and the output logs at the point of the exception show that the container db has reached a point where it should work. In previous runs I've tried adding infinite waits on fail that went all the way up to the test timeout with no luck either, so it shouldn't be a case of the database still coming up.

JamesNK:
> That results in the addition of db.name, db.user, server.address and server.port being set on the spans.

db.name and db.user aren't standard parameters. Making them opt-in makes sense.

server.address and server.port are part of the standard. I think the underlying Oracle library should include them by default. Until that happens, updating configuration to include them is the next best option.

cmdkeen:
> > That results in the addition of db.name, db.user, server.address and server.port being set on the spans.
> 
> `db.name` and `db.user` aren't standard parameters. Making them opt-in makes sense.
> 
> `server.address` and `server.port` are part of the standard and is important information. I think the underlying Oracle library should include them by default. Until that happens, updating Aspire component to set configuration to include them is the next best option.

@alexkeh is adding the two server properties something you're able to raise?

eerhardt:
@cmdkeen - Is setting `EnableConnectionLevelAttributes = true` enough to get all the OTel semantic convention attributes added to the trace? If yes, and if there are no security concerns with what information this setting enables, then setting that by default for now makes sense to me. We can remove it when the Oracle provider adds the OTel semantic convention attributes by default.

cmdkeen:
@eerhardt looks to be, from the playground with that property turned on results in this for the sendExecuteRequest span:
image

image

I'll make that the default for now, with a comment that it can be removed if/when the Oracle library supports.

In other happy news it looks like the Helix Docker tests are finally passing!

alexkeh:
> > > That results in the addition of db.name, db.user, server.address and server.port being set on the spans.
> > 
> > 
> > `db.name` and `db.user` aren't standard parameters. Making them opt-in makes sense.
> > `server.address` and `server.port` are part of the standard and is important information. I think the underlying Oracle library should include them by default. Until that happens, updating Aspire component to set configuration to include them is the next best option.
> 
> @alexkeh is adding the two server properties something you're able to raise?

@cmdkeen Let me check with my dev team.

cmdkeen:
@mitchdenny  and @eerhardt  anything more you want doing on this?

alexkeh:
> > > > That results in the addition of db.name, db.user, server.address and server.port being set on the spans.
> > > 
> > > 
> > > `db.name` and `db.user` aren't standard parameters. Making them opt-in makes sense.
> > > `server.address` and `server.port` are part of the standard and is important information. I think the underlying Oracle library should include them by default. Until that happens, updating Aspire component to set configuration to include them is the next best option.
> > 
> > 
> > @alexkeh is adding the two server properties something you're able to raise?
> 
> @cmdkeen Let me check with my dev team.

@cmdkeen I got feedback from my dev team. There are a couple reasons ODP.NET OTel does not expose `server.address` and `server.port` by default.

1) Security - Organizations can be sensitive to have these two property values written to traces inadvertently and unknowingly. Having the user set `EnableConnectionLevelAttributes` makes it unambiguous.

2) Other OTel DB provider behavior - We observed that Microsoft SqlClient provider doesn’t trace the `server.address` and `server.port` tags by default either. It requires `EnableConnectionLevelAttributes` set to true as well. Its default value is false.

eerhardt:
> Other OTel DB provider behavior - We observed that [Microsoft SqlClient provider](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/README.md#enableconnectionlevelattributes) doesn’t trace the equivalent server.address and server.port tags by default either. It requires EnableConnectionLevelAttributes set to true as well. Its default value is false.

When `EnableConnectionLevelAttributes` is not enabled, Microsoft.SqlClient will emit `peer.service` though. See [this code](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/efd10777672707def2be999dd2d87bcf8eaccb9f/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs#L258-L288).

```C#
    internal void AddConnectionLevelDetailsToActivity(string dataSource, Activity sqlActivity)
    {
        if (!this.EnableConnectionLevelAttributes)
        {
            sqlActivity.SetTag(SemanticConventions.AttributePeerService, dataSource);
        }
        else
        {
...
```

eerhardt:
> That applies to the [MySQL provider](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.MySqlData/README.md#enableconnectionlevelattributes) as well.

For the MySqlConnector library (which we have an Aspire component for), the reason the icon shows up is because MySqlConnector [adds the `net.peer.name` and `net.peer.port` tags to the Activity](https://github.com/mysql-net/MySqlConnector/blob/5a377426aa049f90689e3fffa9894ecd7e2f8e7b/src/MySqlConnector/Core/ServerSession.cs#L964-L1020).

And then the OTLP exporter [appends `peer.service` to the tag list](https://github.com/open-telemetry/opentelemetry-dotnet/blob/4d02fe1a0803daaf6e5db1cf6f45f5d8d89c99a4/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs#L148-L157), using the above 2 tags.

alexkeh:
> > That applies to the [MySQL provider](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.MySqlData/README.md#enableconnectionlevelattributes) as well.
> 
> For the MySqlConnector library (which we have an Aspire component for), the reason the icon shows up is because MySqlConnector [adds the `net.peer.name` and `net.peer.port` tags to the Activity](https://github.com/mysql-net/MySqlConnector/blob/5a377426aa049f90689e3fffa9894ecd7e2f8e7b/src/MySqlConnector/Core/ServerSession.cs#L964-L1020).
> 
> And then the OTLP exporter [appends `peer.service` to the tag list](https://github.com/open-telemetry/opentelemetry-dotnet/blob/4d02fe1a0803daaf6e5db1cf6f45f5d8d89c99a4/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ActivityExtensions.cs#L148-L157), using the above 2 tags.

The Oracle provider doesn't have the `peer.service` attribute. If we support this attribute, we'll plan to emit it by default.

cmdkeen:
@eerhardt given the clarification above do you have a steer on what needs doing to get this merged?

1. Back to the default `EnableConnectionLevelAttributes` - for now devs will have to manually choose whether to enable. In time the Oracle library might add `peer.service` support.
2. Stick with `EnableConnectionLevelAttributes=true` by default and document that it is configurable noting the security concern

Happy either way - suspect given the security element that 1 is the sensible choice.

eerhardt:
> suspect given the security element that 1 is the sensible choice.

Yeah, that is where I'm leaning. There are reasons this is the default from Oracle, following suite is the better choice.

radical:
> > Oracle might be taking a longer time to start up, or failing at some point? I would try to add some waits to confirm that the _db_ is up, and accessible. [This](https://dotnet.testcontainers.org/api/wait_strategies/) might be helpful too.
> > Also, there is #3161 but it doesn't use testcontainers, and instead depends on the AppHost starting the container.
> 
> Hi @radical after much bashing away at this I managed to unearth that the tests are failing with:
> 
> ```
> Oracle.ManagedDataAccess.Client.OracleException (0x80004005): ORA-50201: Oracle Communication: Failed to connect to server or failed to parse connect string
>  ---> OracleInternal.Network.NetworkException (0x80004005): ORA-50201: Oracle Communication: Failed to connect to server or failed to parse connect string
>  ---> OracleInternal.Network.NetworkException (0x80004005): ORA-12514: Cannot connect to database. Service XEPDB1 is not registered with the listener at host 127.0.0.1/127.0.0.1 port 5432. (CONNECTION_ID=c4G0SvajqUmwqX6okJZmQA==)
> ```
> 
> That ORA-12514 code led me to [gvenzl/oci-oracle-xe#225](https://github.com/gvenzl/oci-oracle-xe/issues/225) where it seems there might be issues with different networking setups. I don't know anything about Helix - is it potentially doing something along the lines of network=host?

Sorry, I missed your message earlier. Do IIUC that the issue is fixed for the testcontainers case, but you are still seeing the above with `EndToEnd` tests which depend on the test app starting the containers?

> Is it potentially doing something along the lines of network=host?

And is this something that would hinder it? I'm not sure if this is used when starting the container. https://docs.docker.com/network/drivers/host/ says that `network=host` would mean the container's network is not isolated from the host itself. I would *guess* that it shouldn't be using this, but I'm not sure.
Is there some way to figure that out from within the container?

cmdkeen:
Hi @radical  - thanks for getting back. That networking issue never popped up for the `EndToEnd` tests in playground as they run locally. I managed to fix it for the Helix tests by using `.WithHostname("localhost")` so the tests are working. Without a detailed understanding of how Helix is configuring the networking settings I didn't want to dig further into what the problem might be. TBH if it's working now perhaps best to let it lie, at least as long as it seems to be a problem specific to Oracle TestContainer.

cmdkeen:
Hi @sebastienros and @eerhardt  - apologies for leaving this, it's holiday season in Europe, though no-one has told the weather here...
Have reverted back to the default ODP.Net tracing options, while providing the option to specify if users do want the extra detail and are happy with the security trade-off in their environment. Assuming the builds all go green this should be in a position for a code review.

eerhardt:
@sebastienros - any chance you can take a look at this PR?

@cmdkeen - looks like the build is failing. Can you merge main and get the build green?

radical:
Resolved the merge conflict, and started the pipeline to mirror `Testcontainers.oracle` `3.9.0` nuget.

radical:
Two more packages need to be mirrored - started that.

cmdkeen:
Hi @sebastienros - anything you need from me?

tests/Aspire.Oracle.EntityFrameworkCore.Tests/OracleContainerFixture.cs
```diff
@@ -0,0 +1,56 @@
...
+    public async Task InitializeAsync()
+    {
+        if (RequiresDockerAttribute.IsSupported)
+        {
+            Container = new OracleBuilder()
+                .WithPortBinding(1521, true)
+                .WithHostname("localhost")
+                .WithWaitStrategy(Wait
+                    .ForUnixContainer()
+                    .UntilMessageIsLogged("Pluggable database XEPDB1 opened read write")
```
sebastienros, line 35: @eerhardt Would be awesome to have this on TestDistributedApplicationBuilder to wait for containers ready state. Will check how this can be done, probably the same way ResourceLoggerForwarderService works.

sebastienros:
LGTM, and tried it locally with and without tracing disabled without issue.

mitchdenny:
@eerhardt ready to merge?

eerhardt (approved):
Thanks for the contribution here. This looks good to me.

I will merge it once CI is green.

PR #4081, opened 2024-05-04
Check for Dapr CLI on PATH

### LINKED ISSUE #2219: Aspire.Hosting.Dapr should allow specifying CLI executable location
To allow consuming the dapr CLI in a GitHub Actions workflow, I believe we should support picking up the CLI from an env var value or allow some other way to specify it. Currently, it only checks some fixed locations. For context, see this comment: https://github.com/dotnet/aspire-samples/pull/108#issuecomment-1942916314

cc @philliphoff

philliphoff:
Currently the Dapr CLI is looked for at several very specific locations, depending on the platform. With this change, Aspire will default to the Dapr CLI found on PATH, if any. If none is found, falls back to one of those specific locations.

Resolves #2219
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4081)

### DIFF
--- a/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
+++ b/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
@@ -294,26 +294,31 @@ internal sealed class DaprDistributedApplicationLifecycleHook : IDistributedAppl
             {
                 if (_defaultDaprPath is null)
                 {
-                    var process = Process.Start(
-                        new ProcessStartInfo
-                        {
-                            Arguments = "version",
-                            CreateNoWindow = true,
-                            FileName = "dapr",
-                            UseShellExecute = true,
-                            WindowStyle = ProcessWindowStyle.Hidden
-                        });
+                    Process? process = null;
 
-                    if (process is null)
+                    try
                     {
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -281,17 +284,68 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+            try
             {
-                return path;
+                if (_defaultDaprPath is null)
+                {
+                    Process? process = null;
+
+                    try
+                    {
+                        process = Process.Start(
```
davidfowl, line 301: Can you do this instead?

https://github.com/dotnet/aspire/blob/31ce95ba73ec2b526e90a7ddef795bb79627c95f/src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs#L398-L420

mitchdenny:
@philliphoff any time to come back and look at @davidfowl's question above: 

https://github.com/dotnet/aspire/pull/4081#discussion_r1590002713

PR #4086, opened 2024-05-04
Add Dapr sidecar port environment variables

### LINKED ISSUE #4077: Consider setting DAPR_HTTP_PORT and DAPR_GRPC_PORT in DaprDistributedApplicationLifecycleHook
Currently, when adding a Dapr sidecar, the DAPR_HTTP_ENDPOINT and DAPR_GRPC_ENDPOINT environment variables are set:

https://github.com/dotnet/aspire/blob/31ce95ba73ec2b526e90a7ddef795bb79627c95f/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs#L148-L152

Some apps may instead rely on DAPR_HTTP_PORT and/or DAPR_GRPC_PORT and will therefore fail.

See also:

https://github.com/dapr/dotnet-sdk/blob/ba1341510bfc81f0e1ba4687539ec49aa90abc4e/src/Shared/DaprDefaults.cs#L63-L76

frankbuckley:
Resolves #4077 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4086)

### DIFF
--- a/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
+++ b/src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
@@ -9,6 +9,7 @@ using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Net.Sockets;
 using static Aspire.Hosting.Dapr.CommandLineArgs;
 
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -148,6 +149,9 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
                         var http = daprCli.GetEndpoint("http");
                         var grpc = daprCli.GetEndpoint("grpc");
 
+                        context.EnvironmentVariables.TryAdd("DAPR_HTTP_PORT", grpc.Port.ToString(CultureInfo.InvariantCulture));
```
davidfowl, line 152: You don’t need ToString

src/Aspire.Hosting.Dapr/DaprDistributedApplicationLifecycleHook.cs
```diff
@@ -148,6 +149,9 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
                         var http = daprCli.GetEndpoint("http");
                         var grpc = daprCli.GetEndpoint("grpc");
 
+                        context.EnvironmentVariables.TryAdd("DAPR_HTTP_PORT", http.Port.ToString(CultureInfo.InvariantCulture));
+                        context.EnvironmentVariables.TryAdd("DAPR_GRPC_PORT", grpc.Port.ToString(CultureInfo.InvariantCulture));
```
davidfowl, line 153: I think this is a fine change to make, but it will not work if the resource is a container. This might just be something we add to documentation.

davidfowl (approved):

PR #4093, opened 2024-05-06
[tests] Misc improvements for workload based tests

radical:
This contains changes extracted from the workload-tests branch (https://github.com/dotnet/aspire/pull/3270) which are not specific to workload testing. It will help make the main workload tests PR be more focused.

- Fixes to deploy per-test project `.runsettings` as some tests have custom settings
- Add bits to `BuildEnvironment` to support different TFMs (net8.0/net9.0/..)
- Add support for installing dev certs on helix so `ASPIRE_ALLOW_UNSECURED_TRANSPORT=true` is not required.
- A new `tests/workloads.proj` which can be used to directly install sdk+workload

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4093)


radical:
This extracts the changes that are not specific to workload tests, so the main PR can be more focused.

tests/Shared/WorkloadTesting/ToolCommand.cs
```diff
@@ -59,6 +57,15 @@ public ToolCommand WithEnvironmentVariables(IDictionary<string, string>? extraEn
         return this;
     }
 
+    public ToolCommand WithEnvironmentVariableRemoved(string key)
```
eerhardt, line 60: Is this method called?

tests/Shared/WorkloadTesting/BuildEnvironment.cs
```diff
@@ -76,25 +84,16 @@ public BuildEnvironment(bool expectSdkWithWorkload = true, Func<string, string,
...
-            }
-            else
-            {
-                TestAssetsPath = Path.Combine(solutionRoot!.FullName, "tests");
-            }
+            BuiltNuGetsPath = Path.Combine(SolutionRoot.FullName, "artifacts", "packages", EnvironmentVariables.BuildConfiguration, "Shipping");
+            TestAssetsPath = Path.Combine(AppContext.BaseDirectory, "testassets");
             if (!Directory.Exists(TestAssetsPath))
             {
                 throw new ArgumentException($"Cannot find TestAssetsPath={TestAssetsPath}");
```
eerhardt, line 91: When/how does this folder get created? I'm not seeing where it gets created.

tests/Aspire.EndToEnd.Tests/Directory.Build.props
```diff
@@ -2,8 +2,11 @@
   <PropertyGroup>
     <IsHelixOnlyTestProject>true</IsHelixOnlyTestProject>
     <IsWorkloadTestProject>true</IsWorkloadTestProject>
-    <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
   </PropertyGroup>
 
   <Import Project="..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
```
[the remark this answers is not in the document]
radical: It's because of the properties at the top:
```xml
  <PropertyGroup>
    <IsWorkloadTestProject>true</IsWorkloadTestProject>
  </PropertyGroup>
```

.. which we want to set before `tests/Directory.Build.props` so it imports the workload testing targets.

tests/Shared/WorkloadTesting/AspireProject.cs
```diff
@@ -33,17 +37,23 @@ public AspireProject(string id, string baseDir, ITestOutputHelper testOutput, Bu
...
+            throw new InvalidOperationException("Project is already running");
+        }
+
         object outputLock = new();
         var output = new StringBuilder();
         var projectsParsed = new TaskCompletionSource();
         var appRunning = new TaskCompletionSource();
         var stdoutComplete = new TaskCompletionSource();
         var stderrComplete = new TaskCompletionSource();
+        AppExited = new();
```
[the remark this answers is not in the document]
radical: This is actually useful for when the same project is used to run processes more than once. For example, run the `AppHost` to validate the main app. then call `StopAsync`, and then run the tests project.

I'll change this to make the field nullable. But lemme know you if you don't agree.

tests/Shared/WorkloadTesting/AspireProject.cs
```diff
@@ -172,19 +182,39 @@ public async Task StartAsync(string[]? extraArgs = default, CancellationToken to
...
     {
-        using var restoreCmd = new DotNetCommand(_buildEnv, _testOutput, label: "restore")
+        using var restoreCmd = new DotNetCommand(_testOutput, buildEnv: _buildEnv, label: "restore")
                                     .WithWorkingDirectory(Path.Combine(RootDir, $"{Id}.AppHost"));
-        var res = await restoreCmd.ExecuteAsync($"restore -bl:{Path.Combine(LogPath!, $"{Id}-restore.binlog")} /p:TreatWarningsAsErrors=true");
+        var res = await restoreCmd.ExecuteAsync($"restore \"-bl:{Path.Combine(LogPath!, $"{Id}-restore.binlog")}\" /p:TreatWarningsAsErrors=true");
         res.EnsureSuccessful();
 
-        using var buildCmd = new DotNetCommand(_buildEnv, _testOutput, label: "build")
+        var buildArgs = $"build \"-bl:{Path.Combine(LogPath!, $"{Id}-build.binlog")}\" /p:TreatWarningsAsErrors=true";
```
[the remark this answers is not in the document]
radical: Either works.

tests/Shared/WorkloadTesting/AspireProject.cs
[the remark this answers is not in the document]
radical: Ideally, it should be reset to `null`. I can explore that in the workload tests PR which will be exercising this more.

tests/Shared/WorkloadTesting/ToolCommand.cs
radical: Oops, this is from the workload tests PR.

radical: Should I drop it for this PR?

tests/Shared/WorkloadTesting/BuildEnvironment.cs
radical: It contains a copy of the test project:
https://github.com/dotnet/aspire/blob/9990c950c11d0b9b240c94651e48827ad9c26ff1/tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj#L27

tests/Shared/WorkloadTesting/ToolCommand.cs
eerhardt: > Should I drop it for this PR?

Yes, I would only add it when it is called.

tests/Shared/WorkloadTesting/AspireProject.cs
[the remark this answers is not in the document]
eerhardt: > This is actually useful for when the same project is used to run processes more than once.

When do we do this? Why wouldn't we create a new `AspireProject` every time we wanted to run it?

tests/helix/send-to-helix-endtoendtests.targets
```diff
@@ -1,44 +1,39 @@
...
     <ItemGroup>
+      <HelixPreCommand Condition="'$(OS)' == 'Windows_NT'" Include="set TestsRunningOutsideOfRepo=true" />
+      <HelixPreCommand Condition="'$(OS)' != 'Windows_NT'" Include="export TestsRunningOutsideOfRepo=true" />
+
       <_E2ETestScenarios Include="basicservices" />
       <_E2ETestScenarios Include="cosmos" />
 
-      <HelixCorrelationPayload Include="$(ArtifactsBinDir)dotnet-latest" Destination="dotnet-latest" />
-      <HelixCorrelationPayload Include="$(ArtifactsShippingPackagesDir)" Destination="built-nugets" />
+      <!--<HelixCorrelationPayload Condition="'$(OS)' != 'Windows_NT'" Include="$(_CreateDotNetDevCertsDirectory)" Destination="create-dotnet-devcert" />-->
```
eerhardt, line 36: Why is this commented out?

tests/helix/send-to-helix-ci.proj
```diff
@@ -4,11 +4,22 @@
...
     </ItemGroup>
 
+    <!-- Invoke 'PrepareDependencies' targets with BuildInParallel=false -->
+    <MSBuild Projects="@(_ProjectsToBuild)" Targets="PrepareDependencies" BuildInParallel="false" StopOnFirstFailure="true" />
+
+    <ItemGroup>
+      <!-- Change the set of properties in the evaluated project, so msbuild will build the
+           following projects in parallel, instead of combining with the previous one to build
+           sequentially. -->
+      <_ProjectsToBuild AdditionalProperties="%(AdditionalProperties);__RandomPropertyForDifferentiatingBuilds=$([System.Guid]::NewGuid())" />
```
eerhardt, line 21: This is weird. Why exactly is this needed?

eerhardt (approved):

tests/Shared/WorkloadTesting/AspireProject.cs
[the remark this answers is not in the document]
radical: This is useful if you are trying to run a project more than once in a single test. But thinking about it more, the project has only a a few fields that would need to be copied over to create a fresh instance, so I can do that. Or the other way would be to create a class to represent just the *running* aspireProject.

```
var project = new AspireProject(...);
var runningProject = await project.StartAsync(..);
.. validate ..
runningProject.StopAsync();

.. for a second run ..
var runningProject2 = await project.StartAsync(..);
.. validate ..
runningProject2.StopAsync();
```

I'm not sure what to name this new class though. `AspireProjectProcess`?

tests/helix/send-to-helix-endtoendtests.targets
radical: This can be dropped.

tests/Shared/WorkloadTesting/AspireProject.cs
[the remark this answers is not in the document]
eerhardt: Do we actually want to run the project multiple times in a single test?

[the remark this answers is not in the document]
radical: Yes, for example:

```csharp
    [Theory]
    [MemberData(nameof(ProjectNamesWithTestType_TestData))]
    public async Task StarterTemplateWithTest_ProjectNames(string prefix, string testType)
    {
        string id = $"{prefix}-{testType}";
        string config = "Debug";

        await using var project = await AspireProject.CreateNewTemplateProjectAsync(
            id,
            "aspire-starter",
            _testOutput,
            BuildEnvironment.ForDefaultFramework,
            $"-t {testType}").ConfigureAwait(false);

        _testOutput.WriteLine($"Checking the starter template project");
        await AssertStarterTemplateRunAsync(project, config, _testOutput).ConfigureAwait(false);

        _testOutput.WriteLine($"Checking the starter template project tests");
        await AssertTestProjectRunAsync(project.TestsProjectDirectory, testType, _testOutput, config).ConfigureAwait(false);
    }
```

The test above runs the `AppHost` first and then runs the Tests project.
There could be other tests which need to run the `AppHost` more than once, hm though I can't think of a concrete example right now. The repeat *run* seems to be mostly for Tests project (at least as of now).

The simplest solution would be to have a `public AspireProject(AspireProject other) {..}`. If we need more in future then we could add it at that point.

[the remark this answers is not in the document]
eerhardt: Ok, then what you have is fine. Let's not create the `AppExited` TCS until `Start` is called.

tests/helix/send-to-helix-ci.proj
radical: If I have:
```xml
      <_ProjectsToBuild Include="send-to-helix-inner.proj"
                        Condition="'%(TestCategory.Identity)' != ''"
                        AdditionalProperties="TestCategory=%(TestCategory.Identity);@(AdditionalProperties, ';')" />
 
<!-- meant to not run in parallel -->
 <MSBuild Projects="@(_ProjectsToBuild)" Targets="PrepareDependencies" BuildInParallel="false" StopOnFirstFailure="true" />

<!-- meant to run in parallel -->
<MSBuild Projects="@(_ProjectsToBuild)" Targets="Test" BuildInParallel="true" StopOnFirstFailure="false" />
```

With two test categories `basictests;endtoendtests` then msbuild runs:

1. build `PrepareDependencies` for the first category; then build `Test` for the same, and wait till it is done
2. Then repeat (1) but for the second category

It's possibly some optimization in msbuild that combines these two `msbuild` task invocation, and does not run them in parallel. But adding a new property to `AdditionalProperties` means that the two runs cannot be combined. The value of the property can be anything.

radical:
Please let me know if I missed any of the suggestions.

radical:
Merging this based on the approval. But there were a couple of minor changes after that. Please let me know if there is any feedback to be addressed for those.

Issue #4099, opened 2024-05-07
`UserNameParameter` should be populated like `PasswordParameter` on Postgres resource


mitchdenny:
... and possibly other resources, I haven't checked.

At the moment when I do this:

```csharp
var postgres = builder.AddPostgres("pgsql");
```

The value of `postgres.Resource.UserNameParameter` is `null` but the value of `postgres.Resource.PasswordParameter` is populated. We should treat the username and password the same way as it makes writing expressions which reference the parameters easier ... example:

https://github.com/mitchdenny/nocodb-aspire/blob/90ddebde4362508c7c4c8358b110d778baa80446/src/NocoDB/NocoDB.Hosting/NocoDBExtensions.cs#L28

mitchdenny:
> See my feedback on #4102. One concern I have is that users shouldn't be prompted for a username when they deploy via `azd`.

Looks like @vhvb1989 has a solution for that. When we provide a default value, they don't prompt but it can be overridden.

mitchdenny:
related #4100

davidfowl:
Do we still want to do this?

PR #4102, opened 2024-05-07
Make username parameter always be injected.

### LINKED ISSUE #4100: `WithPgAdmin` does not use username parameter value when provided.
Currently the lifecycle hook that is registered via `WithPgAdmin` does not use the `UserNameParameter` and uses a hardcoded username of `postgres`. This means that if someone overrides the username it won't work.

https://github.com/dotnet/aspire/blob/c6b9f8e195f5b1cd72951f447d35aae527b381b4/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs#L40

mitchdenny:
Fixes #4100 and #4099

This PR does a few things:

1. Changes the PgAdmin config writer hook to use the password from `UserNameParameter` on the Postgres server resource.
2. Changes the `UserNameParameter` on the postgres server resource to be non-nullable and adjust in the constructor as well.

The reason for this change is that if you want to use the Postgres database resource from a container that is not built around .NET, the connection string format that the resource exposes won't necessarily be in the format that you want. For that reason you want to be able to get at the username and password parameters and know that they are always there. The password parameter is fine, but the fact that the username parameter is optional is problematic and makes consuming code more complex.

This represents a breaking change ... see unshipped API diff.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4102)

### DIFF
--- a/src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
@@ -19,6 +19,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <param name="password">A parameter that contains the PostgreSQL server password.</param>
     public PostgresServerResource(string name, ParameterResource userName, ParameterResource password) : base(name)
     {
+        ArgumentNullException.ThrowIfNull(userName);
         ArgumentNullException.ThrowIfNull(password);
 
         PrimaryEndpoint = new(this, PrimaryEndpointName);
--- a/src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
@@ -17,7 +17,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <param name="name">The name of the resource.</param>
     /// <param name="userName">A parameter that contains the PostgreSQL server user name, or <see langword="null"/> to use a default value.</param>
     /// <param name="password">A parameter that contains the PostgreSQL server password.</param>
-    public PostgresServerResource(string name, ParameterResource? userName, ParameterResource password) : base(name)
+    public PostgresServerResource(string name, ParameterResource userName, ParameterResource password) : base(name)
     {
         ArgumentNullException.ThrowIfNull(password);
 
@@ -34,7 +34,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <summary>
     /// Gets the parameter that contains the PostgreSQL server user name.
     /// </summary>
-    public ParameterResource? UserNameParameter { get; }
+    public ParameterResource UserNameParameter { get; }
 
     internal ReferenceExpression UserNameReference =>
         UserNameParameter is not null ?
--- a/tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
+++ b/tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
@@ -580,12 +580,12 @@ public class ManifestGenerationTests
                 },
                 "postgres": {
                   "type": "container.v0",
-                  "connectionString": "Host={postgres.bindings.tcp.host};Port={postgres.bindings.tcp.port};Username=postgres;Password={postgres-password.value}",
+                  "connectionString": "Host={postgres.bindings.tcp.host};Port={postgres.bindings.tcp.port};Username={postgres-username.value};Password={postgres-password.value}",
                   "image": "{{PostgresContainerImageTags.Registry}}/{{PostgresContainerImageTags.Image}}:{{PostgresContainerImageTags.Tag}}",
                   "env": {
                     "POSTGRES_HOST_AUTH_METHOD": "scram-sha-256",
                     "POSTGRES_INITDB_ARGS": "--auth-host=scram-sha-256 --auth-local=scram-sha-256",
-                    "POSTGRES_USER": "postgres",
+                    "POSTGRES_USER": "{postgres-username.value}",
                     "POSTGRES_PASSWORD": "{postgres-password.value}",
                     "POSTGRES_DB": "postgresdb"
                   },
@@ -718,6 +718,15 @@ public class ManifestGenerationTests
                           "minLength": 22
                         }
                       }
+                    }
+                  }
+                },
+                "postgres-username": {
+                  "type": "parameter.v0",
+                  "value": "{postgres-username.inputs.value}",
+                  "inputs": {
+                    "value": {
+                      "type": "string"
                     }
                   }
                 },
--- a/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
@@ -37,7 +37,7 @@ internal sealed class PgAdminConfigWriterHook : IDistributedApplicationLifecycle
                 writer.WriteString("Group", "Servers");
                 writer.WriteString("Host", endpoint.ContainerHost);
                 writer.WriteNumber("Port", endpoint.Port);
-                writer.WriteString("Username", "postgres");
+                writer.WriteString("Username", $"{postgresInstance.UserNameParameter.Value}");
                 writer.WriteString("SSLMode", "prefer");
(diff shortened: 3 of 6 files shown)


src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
```diff
@@ -17,7 +17,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <param name="name">The name of the resource.</param>
     /// <param name="userName">A parameter that contains the PostgreSQL server user name, or <see langword="null"/> to use a default value.</param>
     /// <param name="password">A parameter that contains the PostgreSQL server password.</param>
-    public PostgresServerResource(string name, ParameterResource? userName, ParameterResource password) : base(name)
+    public PostgresServerResource(string name, ParameterResource userName, ParameterResource password) : base(name)
```
eerhardt, line 20: We should be consistent across all resources that take a user name.

https://github.com/dotnet/aspire/blob/9fab2f6bd66866fb210f3b6c1f6347a73c9ab57a/src/Aspire.Hosting.RabbitMQ/RabbitMQServerResource.cs#L21

src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
```diff
@@ -17,7 +17,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <param name="name">The name of the resource.</param>
     /// <param name="userName">A parameter that contains the PostgreSQL server user name, or <see langword="null"/> to use a default value.</param>
```
eerhardt, line 18: This doc should be updated.

src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
```diff
@@ -17,7 +17,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
...
     /// <param name="password">A parameter that contains the PostgreSQL server password.</param>
-    public PostgresServerResource(string name, ParameterResource? userName, ParameterResource password) : base(name)
+    public PostgresServerResource(string name, ParameterResource userName, ParameterResource password) : base(name)
     {
         ArgumentNullException.ThrowIfNull(password);
```
eerhardt, line 22:
```suggestion
        ArgumentNullException.ThrowIfNull(userName);
        ArgumentNullException.ThrowIfNull(password);
```

src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
```diff
@@ -34,7 +34,7 @@ public PostgresServerResource(string name, ParameterResource? userName, Paramete
     /// <summary>
     /// Gets the parameter that contains the PostgreSQL server user name.
     /// </summary>
-    public ParameterResource? UserNameParameter { get; }
+    public ParameterResource UserNameParameter { get; }
 
     internal ReferenceExpression UserNameReference =>
```
eerhardt, line 39: This can be removed now that the UserNameParameter can't be null.

src/Aspire.Hosting.PostgreSQL/PostgresServerResource.cs
```diff
@@ -17,7 +17,7 @@ public class PostgresServerResource : ContainerResource, IResourceWithConnection
     /// <param name="name">The name of the resource.</param>
```
eerhardt, line 17: Should be able to remove line 12:

```
private const string DefaultUserName = "postgres";
```

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -721,6 +721,15 @@ public void VerifyTestProgramFullManifest()
...
                 },
+                "postgres-username": {
+                  "type": "parameter.v0",
+                  "value": "{postgres-username.inputs.value}",
+                  "inputs": {
+                    "value": {
+                      "type": "string"
+                    }
+                  }
+                },
```
eerhardt, line 732: Does this mean that users will always be prompted for a user name when they deploy via `azd`?

eerhardt (commented):
I think this will cause `azd` to always prompt for a user name now.

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
davidfowl: Is there a way to have the default value be a constant?

eerhardt: I don't think that was ever implemented in azd, or Aspire.

The way we do it today is we just put the constant directly where it needs to go. i.e. it isn't a "parameter".

davidfowl: 8.1

mitchdenny: What we could do is expand the the manifest so you can do this:

```json
{
  "inputs": {
    "value": {
      "type": "string",
      "default": {
        "value": "postgres"
      }
    }
  }
}
```

This would be a trigger for an `aspire-8.1.json` schema. If `value` is provided then `generate` isn't allowed. `azd` would prompt and show the default value (the user just hits ENTER). If it is secret it wouldn't show the value but indicate somehow that it should accept the default.

/cc @vhvb1989

vhvb1989: For parameters with a default value, azd does not prompt the user for a value.  The strategy is usually just to tell folks what env-var they can set to override the default value.

In your example above, azd would create a bicep param like

```bicep
param postgres-username string = "postgres"
```

then, azd will also create a mapping to an env-var `AZURE_POSTGRES_USERNAME` (I think we do camelCase to SNAKE_UPPER) which users can set (either with `azd env set AZURE_POSTGRES_USERNAME otherValue` or by setting the env var to their system (or a variable or secret on CI).

This strategy allows the overriding w/o too much prompting.  And the mapping to env-vars is already created for all aspire parameters (even for the auto-geneated, IIRC).

mitchdenny: So do we have enough to decide what we want to go ahead with this? We would need:

1. A change to AZD to support using default values if they are present (in accordance with the schema above).
2. A change to our schema (introduce an 8.1 version of the schema)
3. Introduce the necessary API changes to allow for setting the default value.

davidfowl: Can we make this change in a non breaking way that doesn't require a schema update? AZD can prompt in older versions.

mitchdenny: Well the value field is additive, so we can add it without rev'ing the schema. And tools use use the value if they understand it. We might be able to get away with that?

davidfowl: Added an API proposal with the manifest schema change above https://github.com/dotnet/aspire/issues/4429

davidfowl: I now think we shouldn't version the schema here. On an older azd you'll get the prompting behavior.

davidfowl: We're adding the schema for default values here https://github.com/dotnet/aspire/pull/5586

cc @davidebbo

davidebbo: Yes, this seems to describe exactly what we have in that PR. What's a bit more vague is exactly how azd will behave, prompting wise.

> If it is secret it wouldn't show the value but indicate somehow that it should accept the default.

@mitchdenny this part is maybe questionable (iiuc), as it suggests that the secret would be stored in the manifest, but just not shown to the user at prompt time. It's best to say that secrets can never be in the manifest, which relates to https://github.com/dotnet/aspire/issues/5237.

davidfowl:
We should be able to finish this now.

davidfowl:
@mitchdenny I think this is worth doing but not sure it's critical for 9. There's another PR from a contributor here https://github.com/dotnet/aspire/pull/5805

mitchdenny:
Yeah can punt this one.

PR #4106, opened 2024-05-07
Add AWS Console urls in the dashboard for CloudFormation stacks and resources

djonser:
This PR adds URLs in the Aspire Dashboard to the AWS Console for CloudFormation Stacks and a selection of AWS resource types that are commonly used in local development. The resource types are: SQS Queues, SNS Topics, S3 Buckets, Cognito UserPools and DynamoDB Tables/GlobalTables.

It could be that the CloudFormation URL is sufficient but the CloudFormation-panel does not provide URLs for SQS Queues or Global Tables. It's also an extra navigation step. I will let review decide that.

aspire-dashboard

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4106)


src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public static string MapCloudFormationUrl(string stackId)
+    {
+        var region = stackId.Split(":")[3];
+
+        return
+            $"https://{region}.console.aws.amazon.com/cloudformation/home?region={region}#/stacks/resources?stackId={stackId}";
+    }
+
+    public static string? MapResourceUrl(string stackId, string resourceType, string physicalResourceId)
```
normj, line 16: Keeping this up to date will be an impossible task and this is one of those features that if you give it to users they are going to expect it for all of the resources. I would stick with having the URL to the stack since there isn't a way to programmatically determine the URL for any arbitrary resource. I think have the URL to the stack is a really cool addition and gives users 90 percent of what they need.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Aspire.Hosting.AWS.CloudFormation;
+
+internal static class CloudFormationAWSConsoleUrlMapper
+{
+    public static string MapCloudFormationUrl(string stackId)
+    {
+        var region = stackId.Split(":")[3];
+
+        return
+            $"https://{region}.console.aws.amazon.com/cloudformation/home?region={region}#/stacks/resources?stackId={stackId}";
```
normj, line 13: There are some corner cases for knowing the console host. For example the China regions go to separate host. It is also possible users could configure the service client being used here to point to Localstack in which case you wouldn't have a console to go to.

I suggest passing in the `IAmazonCloudFormation` service client into this method. Then call the following method to determine the endpoint the service client is going to send the request to.
```
var cloudFormationUrl = client.DetermineServiceOperationEndpoint(new DescribeStacksRequest { StackName = stackId }).URL;
```

Then in the host of that URL look for `.amazonaws.`. If you don't find that token then don't add a url to the dashboard. If you do find one then take from the service url from the `amazonaws.` to the end to create the suffix of the console endpoint.

normj (changes requested):
Cool feature adding the console links to the stacks.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
```diff
@@ -0,0 +1,35 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.AWS.CloudFormation;
+
+internal static class CloudFormationAWSConsoleUrlMapper
+{
+    public static string MapCloudFormationUrl(string stackId)
+    {
+        var region = stackId.Split(":")[3];
```
davidfowl, line 10: Can we make this more robust? I'd hate to see random exceptions because of format changes or parsing errors.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
djonser: 90% is enough. Removed service specific urls.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
djonser: Using `DetermineServiceOperationEndpoint` I received the following URL: `https://cloudformation.eu-west-1.amazonaws.com/`. Filtering for `.amazonaws.` as suggested is not a problem. However, I'm a unsure what part of this can be used to build the full console URL. 

The URL should be `https://console.aws.amazon.com/cloudformation/home?region={region}#/stacks/resources?stackId={stackId}` right?

If I'm passing the client that has just been used successfully, then I should be able to use region from `client.Config.RegionEndpoint.SystemName` and avoid string parsing. I also opted to move the single mapping method into same class instead.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
djonser: Now that the client is used to resolve information I resolve this via a the client configuration instead.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationProvisioner.cs
```diff
@@ -177,4 +182,27 @@ private static IAmazonCloudFormation GetCloudFormationClient(ICloudFormationReso
...
+        {
+            var endpointUrl = client.DetermineServiceOperationEndpoint(new DescribeStacksRequest { StackName = stackId })?.URL;
+            if (endpointUrl == null || !endpointUrl.Contains(".amazonaws."))
+            {
+                return null;
+            }
+
+            var url = $"https://console.aws.amazon.com/cloudformation/home?region={client.Config.RegionEndpoint.SystemName}#/stacks/resources?stackId={stackId}";
+
+            return
```
normj, line 198: Once we know the region run it through the `^(us|eu|ap|sa|ca|me|af|il)\\-\\w+\\-\\d+$` regex. That will ensure the region is valid for the `console.aws.amazon.com` console link. If the region is not a match then skip adding the URL. The SDK has this regex to figure out what AWS partition the region is in. https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Core/partitions.json#L12C22-L12C63

normj (changes requested):

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
normj: Funny I hadn't noticed before that main console host doesn't use the same pattern as service endpoints. The console for China does follow the service endpoint. I'm going to guess some legacy issue there. 

I added a new comment on how you can validate the region for the stack is valid for the `https://console.aws.amazon.com/` console endpoint.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationAWSConsoleUrlMapper.cs
normj: I added a new comment on how you can parse the stack arn using the SDK's `Arn` class.

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationProvisioner.cs
djonser: Done. Should there be support for other partitions? I have no familiarity with their URLs so would need some guidance with that.

normj (approved):
Looks good!

djonser:
@davidfowl an e2e test is failing. Should I push a commit to re-trigger?

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationProvisioner.cs
```diff
@@ -177,4 +184,34 @@ private static IAmazonCloudFormation GetCloudFormationClient(ICloudFormationReso
             throw new AWSProvisioningException("Failed to construct AWS CloudFormation service client to provision AWS resources.", e);
         }
     }
+
+    private static readonly Regex s_awsRegionRegex = new("^(us|eu|ap|sa|ca|me|af|il)-\\w+-\\d+$", RegexOptions.Singleline);
```
sebastienros, line 188: Missed opportunity to use regex source generators

https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators?pivots=dotnet-8-0

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationProvisioner.cs
```diff
@@ -177,4 +184,34 @@ private static IAmazonCloudFormation GetCloudFormationClient(ICloudFormationReso
...
+
+    internal static ImmutableArray<UrlSnapshot>? MapCloudFormationStackUrl(IAmazonCloudFormation client, string stackId)
+    {
+        try
+        {
+            var endpointUrl = client.DetermineServiceOperationEndpoint(new DescribeStacksRequest { StackName = stackId })?.URL;
+            if (endpointUrl == null || !endpointUrl.Contains(".amazonaws."))
```
sebastienros, line 195: StringComparison.OrdinalIgnoreCase ?

src/Aspire.Hosting.AWS/CloudFormation/CloudFormationProvisioner.cs
```diff
@@ -177,4 +184,34 @@ private static IAmazonCloudFormation GetCloudFormationClient(ICloudFormationReso
...
+            {
+                return null;
+            }
+
+            if (!Arn.TryParse(stackId, out var arn) || !s_awsRegionRegex.IsMatch(arn.Region))
+            {
+                return null;
+            }
+
+            var url = $"https://console.aws.amazon.com/cloudformation/home?region={arn.Region}#/stacks/resources?stackId={stackId}";
```
sebastienros, line 205: UrlEncode arguments?

sebastienros (approved):

normj:
@davidfowl @sebastienros Can we merge this PR? I'm worried other PRs are going to make this PR go stale and would rather the other in progress PR incorporate this change.

eerhardt:
I've enabled auto-merge on this PR. It should merge as soon as CI passes.

davidfowl:
This needs to be rebased on main.

davidfowl:
Thanks @vlesierse, @normj and @djonser !

PR #4113, opened 2024-05-07
Update machine-requirements

DamianEdwards:
Update VS preview version and mention Podman as supported alternative to Docker Desktop.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4113)


joperezr (approved):
Thanks for fixing

danmoseley:
this file covers (is linked from) "using daily build" and "contributing". But as of GA, there's a 3rd scenario -- "using supported build". Eg. , for the latter, we presumably put the latest version of 17.X that is GA'd whatever that is. 

I guess when we get to GA in 2 weeks, we should update to add that info.

Issue #1403, opened 2023-12-15
Redesign EF components public API


eerhardt:
We've gotten feedback about missing EF scenarios - it started with [Npgsql component doesn't allow configuration of the NpgsqlDbContextOptionsBuilder](https://github.com/dotnet/aspire/issues/438). We we fixed this by allowing to configure the `DbContextOptionsBuilder`, but this had the problem that the user needed to call `UseNpgsql` again to configure Npgsql specific options. Npgsql and SqlServer work this way, but CosmosDB doesn't - the [UseCosmos API](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.cosmosdbcontextoptionsextensions.usecosmos) forces callers to pass a connection string, or other connection information.

Then more feedback came in that we don't support all the options that EF allows w.r.t. DependencyInjection:

* https://github.com/dotnet/aspire/issues/1276
* https://github.com/dotnet/aspire/issues/1241
* https://github.com/dotnet/aspire/issues/1194
* https://github.com/dotnet/aspire/issues/744 (the error happens because the scaffolding injects EF code into Program.cs)

Given all this feedback, and diving into the EF APIs, we think we need to redesign the Aspire EF component's public APIs. Today we are wrapping the EF APIs for Pooling or regular DbContext.

But instead, we should have the dev user call directly into the EF DependencyInjection APIs, so they have full control over what they want to add to the app. Then the Aspire EF Component can just have an extension method to add the other things:

* Health Checks
* Tracing
* Metrics
* Tracing
* Logging
* etc

The hard part will be coming up with the naming of that Aspire extension method. It can't be called `AddNpgsqlDbContext` anymore, since it doesn't add the DbContext.

eerhardt:
Some other options I thought of. Not that I like any of them, but listing them here for completeness:

### Add Settings properties to control behavior

Following the current design, which has 

https://github.com/dotnet/aspire/blob/cce39149529f3388d5ab9fbc36f3d5e10feac3ab/src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/NpgsqlEntityFrameworkCorePostgreSQLSettings.cs#L23

that controls how we add the DbContext to the ServiceCollection

[...]

PR #4115, opened 2024-05-08
Add support for specifying DbContext service in Aspire.Npgsql.EntityFrameworkCore et al.

### LINKED ISSUE #4114: Allow specifying service (interface) and implementation when calling AddNpgsqlDbContext
Currently the method `AddNpgsqlDbContext` only allows for a concrete context type to be specified. This makes it difficult to use with both unit testing and Clean Architecture without jumping through a few additional hoops.

It would be nice of there was an overload to `AddNpgsqlDbContext<TContext>()` with the signature `AddNpgsqlDbContext<TContextService, TContextImplementation>()` as there is for most of the existing EF extension methods for DI.

nwoolls:
This update introduces the ability to register a DbContext with a specified context service interface in the Aspire.Npgsql.EntityFrameworkCore. Extension and test methods have been added to support this functionality.

Resolves https://github.com/dotnet/aspire/issues/4114.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4115)


src/Components/Aspire.Npgsql.EntityFrameworkCore.PostgreSQL/AspireEFPostgreSqlExtensions.cs
```diff
@@ -64,14 +94,14 @@ public static partial class AspireEFPostgreSqlExtensions
 
         configureSettings?.Invoke(settings);
 
-        builder.Services.AddDbContextPool<TContext>(ConfigureDbContext);
+        builder.Services.AddDbContextPool<TContextService, TContextImplementation>(ConfigureDbContext);
```
jbroumels, line 97: Should pooling be used by default ?

nwoolls: Good question but I would defer to the OG author. This PR does not alter or introduce that behavior.

jbroumels: Thanks for that info.
I just wondered, because the normal EF AddDbContext of course is a non-pooled setup.

The previous PR could be introducing certain "problems"

nwoolls: @jbroumels if I had to pick I'd prefer restoring the previous functionality where this is configurable. In our own projects, over time, we've adopted a lot of opinionated code very similar to Aspire, where we take safe, sane, defaults and the rest is taken from `appsettings.json` (including pooling).

jbroumels: Agree, the defaults should always be the safe manner.
I overlooked it didn't came from this PR directly, so it doesn't belong here.

Will look into the referenced commit and PR.

Closing this comment section here for now :)

sebastienros: You may file an issue about why pooling or not should be the default, even for consistency in the APIs, I would loop some EF people too.

sebastienros:
Looking at how `AddDbContextPool<TService, TImplementation>` [is implemented in EF Core](https://github.com/dotnet/efcore/blob/main/src/EFCore/Extensions/EntityFrameworkServiceCollectionExtensions.cs#L163-L169) I think we could just have the implementation for `AddNpgsqlDbContext<TContextService, TContextImplementation>()` and have `AddNpgsqlDbContext<TContext>()` invoke `AddNpgsqlDbContext<TContext, TContext>()`.

Then we also need to do it for all other EF components (Cosmos, MySQL, Oracle, SQL Server). Can you do that?

/cc @AndriySvyryd

nwoolls:
> Looking at how `AddDbContextPool<TService, TImplementation>` [is implemented in EF Core](https://github.com/dotnet/efcore/blob/main/src/EFCore/Extensions/EntityFrameworkServiceCollectionExtensions.cs#L163-L169) I think we could just have the implementation for `AddNpgsqlDbContext<TContextService, TContextImplementation>()` and have `AddNpgsqlDbContext<TContext>()` invoke `AddNpgsqlDbContext<TContext, TContext>()`.
> 
> Then we also need to do it for all other EF components (Cosmos, MySQL, Oracle, SQL Server). Can you do that?
> 
> /cc @AndriySvyryd

If I understand your comment that is what this PR does.

sebastienros:
@nwoolls oh well, the big green blurb tricked me, sorry, we agree on the solution then ;)

sebastienros (approved):

AndriySvyryd:
> Then we also need to do it for all other EF components (Cosmos, MySQL, Oracle, SQL Server).

This part is still missing

sebastienros:
@AndriySvyryd good point, that's why I had not provided my approval yet, we should merge when we have the other providers implemented.

tests/Aspire.Pomelo.EntityFrameworkCore.MySql.Tests/AspireEFMySqlExtensionsTests.cs
```diff
@@ -247,6 +247,25 @@ public void CommandTimeoutFromSettingsWinsOverOthers(bool useSettings)
...
+    /// </summary>
+    [Fact]
+    public void CanSpecifyContextService()
+    {
+        var builder = Host.CreateEmptyApplicationBuilder(null);
+        builder.Configuration.AddInMemoryCollection([
+            new KeyValuePair<string, string?>("ConnectionStrings:mysql", ConnectionString),
+        ]);
+
+        builder.AddMySqlDbContext<ITestDbContext, TestDbContext>("mysql");
```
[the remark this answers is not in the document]
AndriySvyryd: You can try it, but now I am not sure that it'll work. That test shouldn't be connecting to the database at all, you might need to debug it to see why that's happening. It might be passing locally just because it successfully connects to the database.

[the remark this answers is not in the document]
nwoolls: Unfortunately I'm not sure where to go from here. I only needed this for Postgres, and all of the tests for the other DB providers work.

If I stop Docker, I get 14 test failures for MySQL with:

```
Docker is either not running or misconfigured. Please ensure that Docker is running and that the endpoint is properly configured. You can customize your configuration using either the environment variables or the ~/.testcontainers.properties file. For more information, visit:
https://dotnet.testcontainers.org/custom_configuration/
```

If I start Docker, the tests all pass locally.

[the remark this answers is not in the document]
nwoolls: If I'm reading the failed actions correctly, these same tests succeed on Linux. This is only failing on Windows, which may explain why I cannot reproduce it. I am not on Windows, nor do I have access to a Windows device. @AndriySvyryd @sebastienros are you able to run this failing test on a Windows machine?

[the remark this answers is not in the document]
nwoolls: I took one more stab based on existing tests. All passing now.

eerhardt:
@sebastienros @AndriySvyryd - I thought the design with https://github.com/dotnet/aspire/issues/1403 was to not add all these APIs to Aspire. And instead, if the "simple" API doesn't work for you, drop to using the EF API directly to register your DbContext, and then call `EnrichNpgsqlDbContext` to add any Aspire services over top?

Is this no longer the case?

AndriySvyryd:
@eerhardt It depends on where we define the line for the "simple" API, I don't have a strong opinion one way or the other. Perhaps @DamianEdwards can weigh in?

nwoolls:
@eerhardt I'm clearly a bit biased having introduced this, but I don't think this is equivalent to the requests in #1403, nor do I think this is particularly pushing in that direction. This only introduces the ability to specify the service context service in addition to the implementation. Without that, how do we unit test Aspire applications that have database contexts? As soon as someone wants to inject an interface instead of a concrete class, they have to drop back to using the `Enrich` syntax. Sure you could go full repository pattern, or use in-memory providers, or SQLite, or dev containers, but with existing .NET EF it's possible to do what this PR allows to start getting test coverage with a simple DB context and interface.

DamianEdwards:
We were concerned that there will always be pressure and expectation to add more overloads and/or methods to the Aspire EF components to support all the possible ways one can configure EF Core. Each seems reasonable in isolation but the approach isn't tenable on the whole IMO. I think we should just stick with what we landed with in #1403 for now.

nwoolls:
@DamianEdwards agreed on that, but this isn't about introducing overloads to configure EF core. This is simply allowing someone to specify both the interface and implementation with the DI container, something I would think is **extremely** common with DI in .NET.

DamianEdwards:
@nwoolls while I appreciate this space is full of many opinions, my reading of https://learn.microsoft.com/ef/core/testing/ is that the EF Core team themselves discourage mocking or `DbContext` substitution in tests and instead promote the use of SQLite in-memory mode or introducing an abstraction layer higher than the `DbContext` (e.g. a repository).

At this point I'm not convinced that these methods belong in the Aspire components. They're easy enough to add as extensions in your own projects or as a separate package (perhaps even in [aspirant](https://github.com/aspirant-project/aspirant)).

sebastienros:
> They're easy enough to add as extensions in your own projects or as a separate package

In this specific case I am not sure that this is the case. The one we currently expose only takes one generic type, the one added takes two.

I am not disagreeing with the other arguments though.

DamianEdwards:
@sebastienros is it not as simple as?

```csharp
public static class DbContextSqlServerExtensions
{
    public static void AddSqlServerDbContext<TContextService, TContextImplementation>(this IHostApplicationBuilder builder, string name)
        where TContextImplementation : DbContext, TContextService
        where TContextService : class
    {
        var connectionString = builder.Configuration.GetConnectionString(name)
            ?? throw new InvalidOperationException("Connection string not found");

        builder.Services.AddDbContextPool<TContextService, TContextImplementation>(options =>
            options.UseSqlServer(connectionString));

        builder.EnrichSqlServerDbContext<TContextImplementation>();
    }
}
```

sebastienros:
@DamianEdwards yes, perfect, was afraid Enrich would do something different but I checked the doc and it should be equivalent.

PR #4141, opened 2024-05-10
Fix starter template's http launch profile

afscrome:
The aspire sample template currently ships with a broken http launch profile.  Running the template `dotnet run --launch-profile http` results in an error.

```
The 'applicationUrl' setting must be an https address unless the 'ASPIRE_ALLOW_UNSECURED_TRANSPORT' environment variable is set to true. This configuration is commonly set in the launch profile. See https://aka.ms/dotnet/aspire/allowunsecuredtransport for more details
```

This PR fixes this by specifying the `ASPIRE_ALLOW_UNSECURED_TRANSPORT` by default to make this work.

An alternative solution could be to remove the `http` profile completely as it feels wrong to ship with a profile that is broken by default.  
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4141)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-apphost/Properties/launchSettings.json
+++ b/src/Aspire.ProjectTemplates/templates/aspire-apphost/Properties/launchSettings.json
@@ -24,7 +24,8 @@
         "ASPNETCORE_ENVIRONMENT": "Development",
         "DOTNET_ENVIRONMENT": "Development",
         "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "http://localhost:19000",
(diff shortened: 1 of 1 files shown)


davidfowl:
@DamianEdwards thoughts? I think this is a good change, but you could argue that it makes the user less aware of this setting. Maybe we can log something instead to raise awareness.

DamianEdwards:
We did this intentionally so that the template cannot be run in the unsecured mode without user intervention. I think we should leave it the way it is.

Issue #2760, opened 2024-03-10
Python Support in Aspire


dorgrin6:
Hey team,

I’ve been exploring Aspire, and it’s impressive! But here’s the thing: Python support is a must for us. Many data science teams, including ours, rely on Python containers. Since [Visual Studio already supports Python](https://visualstudio.microsoft.com/vs/features/python/), adding it to Aspire would be a game-changer.

Cheers.
@davidfowl  - FYI

En3Tho:
@dorgrin6 What do you mean by supporting Python?

There could be multiple things:
1) Running python based apps natively like with Node support?

2) ServiceDiscovery package for Python?

3) A whole separate Aspire.Python package which imitates behavior of Aspire but for Python?

I think you can easily hack together 1 and 2 and see how it goes.

As for 3 I don't think there is a need for this kind of thing - Python community can easily do the thing themselves.

I guess Node native support is a must because .Net backend + js/ts + node frontend is standard thing. I can't say the same for Python + .Net honestly.

dorgrin6:
I meant 1 & 2.
I'll try to hack this together as proposed.
Would node integration be a good example for this?

mitchdenny:
@dorgrin6 in terms of 1 & 2 I'd be interested in seeing a PR for Aspire. I think item 3 is probably a different open source project entirely.

wmeints:
I'm working on a separate package for Python support. https://github.com/wmeints/aspire-python

wmeints:
Turns out there are a couple of limitations for extending the hosting bits outside of the main repo.
Testing is harder because you need to build your own testing infrastructure. Which involves copying over quite a lot of code.

So instead of building my own package, I propose to use my PR that I submitted.

mitchdenny:
Python support is almost certainly something we'd want to support in-tree anyway since its pretty common in things like AI scenarios.

mitchdenny:
There is a meta conversation here which is not Python specific about how we want to handle closely related application frameworks:

https://github.com/dotnet/aspire/issues/4163

Interesting for Node.js we don't have APIs which are specific to say ExpressJS?

Meir017:
*fixed by #4142

PR #4142, opened 2024-05-10
Add support for running python projects

wmeints:
This PR introduces support for running Python projects as part of an Aspire solution. I took the NodeJS extension as inspiration but had to add a couple of extra pieces of logic since Python is a lot less standardized when it comes to how projects are structured.

Initially I created this project: https://github.com/wmeints/aspire-python but I feel that this PR is a much more useful way to add support for Python. I think it's a great resolution for #2760.

## Assumptions made by the extension

- Your python project has a virtual environment located under `.venv`. Although you can change that.
- You have a Dockerfile for publishing the project to test/acceptance/production environments.


Aspire.sln
```diff
@@ -383,7 +384,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aspire.Hosting.RabbitMQ", "
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aspire.Hosting.Redis", "src\Aspire.Hosting.Redis\Aspire.Hosting.Redis.csproj", "{7D57BF77-AD99-48C5-BC4D-AB133A155D97}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Aspire.Hosting.Garnet", "src\Aspire.Hosting.Garnet\Aspire.Hosting.Garnet.csproj", "{96D36588-B223-4381-B774-FB1A3F8BC269}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aspire.Hosting.Garnet", "src\Aspire.Hosting.Garnet\Aspire.Hosting.Garnet.csproj", "{96D36588-B223-4381-B774-FB1A3F8BC269}"
```
mitchdenny, line 387: Why did this change?

playground/python/Python.AppHost/Program.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.Python;
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+builder.AddPythonProjectWithVirtualEnvironment("script-only", "../script_only", "main.py");
```
mitchdenny, line 8: This is a pretty long `Add...` method. Can we make it more concise?

src/Aspire.Hosting.Python/FlaskProjectResourceExtensionBuilder.cs
```diff
@@ -0,0 +1,80 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Utils;
+
+namespace Aspire.Hosting.Python;
+
+/// <summary>
+/// Provides extension methods for adding Python Flask applications to an <see cref="IDistributedApplicationBuilder"/>.
+/// </summary>
+public static class FlaskProjectResourceExtensionBuilder
```
mitchdenny, line 12: I think this could probably collapsed into a single extension class with all the python-related extension methods in it.

Aspire.sln
mitchdenny: It might be a good idea to revert the sln file change and rebase on main and then add the projects back in. I'm not entirely sure why these lines have been updated. But sln file merges can be interesting :)

mitchdenny:
@davidfowl I've been thinking about things like Node.js (and now Python). I'm wondering if whether to make things easier for people we should create a boilerplate Dockerfile if one doesn't already exist. The idea being that it gives people a starting point.

So if someone does `AddPython...`) and points at a particular directory we'll drop down a Dockerfile and .dockerignore file if one doesn't already exist. Then the can tweak it from there customizing the base image etc.

davidfowl:
I don't think we should be writing files to disk using these APIs, but I like the idea that they would publish as a docker file by default. The publish operation would throw because it couldn't find one. That would instruct the user to add one for their language.

playground/python/Python.AppHost/Program.cs
wmeints: I can remove the virtual environment part and make that implied. I don't think anyone wants to work on python projects without a virtual env.

Aspire.sln
wmeints: Yeah, that's absolutely true. I don't know what changed that particular line. I used `dotnet sln add` to include the python project. I'll revert the file, rebase, and try again.

src/Aspire.Hosting.Python/FlaskProjectResourceExtensionBuilder.cs
wmeints: For the Flask extension that's certainly the case. I'm also working on Django right now, and that's a little more complicated. For Django to run well, you'll need to have two things:

- A super user which you can create with a specific python command
- Applied migrations which requires another command

I would like to make it so that you can use `WithSuperUser(username, secret)` and `WithApplyMigrations()` so that people don't have to fiddle with the commands.

I think I can do this by using a `DjangoProjectResource` and have two extension methods to extend it. What do you think?

playground/python/Python.AppHost/Program.cs
davidfowl: Make it a bool? Add a doc comment?

src/Aspire.Hosting.Python/FlaskProjectResourceExtensionBuilder.cs
```diff
@@ -0,0 +1,83 @@
...
+                context.Args.Add(flaskExecutable!);
+            }
+
+            context.Args.Add("--app");
+            context.Args.Add(entrypoint);
+            context.Args.Add("run");
+        });
+
+        // Expose the port on the FLASK_RUN_PORT to allow the Flask app to bind to the correct port.
+        resourceBuilder.WithHttpEndpoint(targetPort: 5000, name: "http", env: "FLASK_RUN_PORT");
```
davidfowl, line 66: Remove the target port.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,54 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.Tests.Helpers;
+using Xunit;
+using Microsoft.Extensions.DependencyInjection;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Tests.Python;
+
+public class AddPythonProjectTests
```
davidfowl, line 11: We need tests that also assert the resource definition and values.

src/Aspire.Hosting.Python/PythonProjectResource.cs
```diff
@@ -0,0 +1,42 @@
...
+/// <param name="executablePath">The path to the executable used to run the python project.</param>
+/// <param name="projectDirectory">The path to the directory containing the python project.</param>
+public class PythonProjectResource(string name, string executablePath, string projectDirectory)
+    : ExecutableResource(name, executablePath, projectDirectory), IResourceWithServiceDiscovery
+{
+    /// <summary>
+    /// Writes out the publishing manifest for the python project.
+    /// </summary>
+    /// <param name="context">The manifest publishing context to extend.</param>
+    internal async Task WriteDockerFileManifestAsync(ManifestPublishingContext context)
```
davidfowl, line 22: How is this different from calling PublishAsDockerFile?

src/Aspire.Hosting.Python/FlaskProjectResourceExtensionBuilder.cs
mitchdenny: I'm not sure about flask and django at this point. Lets get python + virtual environments sorted out first.

wmeints: Agreed, I'll focus on that first.

src/Aspire.Hosting.Python/PythonProjectResource.cs
wmeints: It's not, I haven't seen the method ;-) I'll fix it.

Aspire.sln
wmeints: I redid the change, at looks better now.

wmeints:
I can see I broke the build because I forgot to update the playground apphost. I will do that tomorrow morning (GMT+1)

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        var app = appBuilder.Build();
+        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
+        var executableResources = appModel.GetExecutableResources();
+
+        var pythonProjectResource = Assert.Single(executableResources);
+
+        Assert.Equal("python", pythonProjectResource.Name);
+        Assert.Equal(pythonProjectDirectory, pythonProjectResource.WorkingDirectory);
+    }
```
davidfowl, line 33: I would love to see some pure unit tests like this https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Tests/Garnet/AddGarnetTests.cs

src/Aspire.Hosting.Python/VirtualEnvironment.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Locates an executable in the virtual environment.
+    /// </summary>
+    /// <param name="name">The name of the executable.</param>
+    /// <returns>Returns the path to the executable if it exists in the virtual environment.</returns>
+    public string? GetExecutable(string name)
+    {
+        string[] allowedExtensions = [".exe", ".cmd", ".bat", ""];
+        string[] executableDirectories = ["bin", "Scripts"];
```
davidfowl, line 20: Filter this based on operating system?

src/Aspire.Hosting.Python/PythonProjectResource.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Aspire.Hosting.Python;
+
+/// <summary>
+/// A resource that represents a Python project.
+/// </summary>
+/// <param name="name">The name of the resource.</param>
+/// <param name="executablePath">The path to the executable used to run the python project.</param>
+/// <param name="projectDirectory">The path to the directory containing the python project.</param>
+public class PythonProjectResource(string name, string executablePath, string projectDirectory)
+    : ExecutableResource(name, executablePath, projectDirectory), IResourceWithServiceDiscovery
```
davidfowl, line 15: Do we want this to implement `IResourceWithServiceDiscovery`?

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
wmeints: I need a little advice on this. The python virtual env layout varies across platforms, on Windows there's a `scripts` directory while on linux and mac there's a `bin` directory for the executables. I could add OS detection and assume that the scripts/binaries are in the right spot. But I felt that it would be nicer to actually check on disk if the binaries are there.

Because of this choice I figured it would make sense to have local facts with an actual python environment instead of writing pure unit-tests.

I can make purer unit-tests by either: Making the assumption about the virtual env layout and not checking the disk. Or I can write temp dummy files to disk to make the unit-test pass. 

What do you prefer? Or do you have other ideas?

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+using Aspire.Hosting.Utils;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Provides extension methods for adding Python applications to an <see cref="IDistributedApplicationBuilder"/>.
+/// </summary>
+public static class PythonProjectResourceBuilderExtensions
+{
+    /// <summary>
```
mitchdenny, line 15: The `<summary>` should be a short one sentence thing. Move the rest into `<remarks>` (after `<returns>`).

Would be good to add an `<example>` as well for each of the extension methods.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace Aspire.Hosting.Tests.Python;
+
+public class AddPythonProjectTests
+{
+    private static readonly Regex s_pythonExecutablePattern  = new Regex("python(\\.(exe|bat|cmd|sh|ps1))?$", RegexOptions.IgnoreCase);
+    private static readonly Regex s_telemetryExecutablePattern = new Regex("opentelemetry-instrument(\\.(exe|bat|cmd|sh|ps1))?$", RegexOptions.IgnoreCase);
+    private static readonly string s_playgroundDirectory = Path.GetFullPath(Path.Combine(AppContext.BaseDirectory, "../../../../../playground/python/"));
+
+    [LocalOnlyFact("python")]
+    public void AddPythonProject_SetsResourcePropertiesCorrectly()
```
mitchdenny, line 18: @radical heads up on this coming into the repo ... when we start running these tests in C# we are going to need Python available.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        var pythonProjectDirectory = Path.Combine(s_playgroundDirectory, "script_only");
```
mitchdenny, line 20: You are going to need to use the testbuilder here (look at some of the other tests). Otherwise this will leak file watch handles.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
mitchdenny: What I'm doing with the incoming WithDockerfile(...) extension is creating a temp directory and dropping the minimum number of files I need in there. If the tests are going to actually spin up Python that is what you'd need to do.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+                AddOpenTelemetryArguments(context);
+
+                // Add the python executable as the next argument so we can run the project.
+                context.Args.Add(pythonExecutable!);
+            }
+
+            AddProjectArguments(scriptPath, scriptArgs, context);
+        });
+
+        if(!string.IsNullOrEmpty(instrumentationExecutable))
```
davidfowl, line 60: nit: Formatting

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/> to add the resource to.</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="projectDirectory">The path to the directory containing the python project files.</param>
+    /// <param name="scriptPath">The path to the script relative to the project directory to run.</param>
+    /// <param name="virtualEnvironmentPath">Path to the virtual environment.</param>
+    /// <param name="scriptArgs">The arguments for the script.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<PythonProjectResource> AddPythonProject(
+        this IDistributedApplicationBuilder builder, string name, string projectDirectory, string scriptPath, string virtualEnvironmentPath = ".venv", params string[] scriptArgs)
```
mitchdenny, line 33: Can you make the default value for `virtualEnvironmentPath` `null` and then check for it and replace it with the default inside the method. Also note the earlier comments I made about making `<summary>` shorter and more concise and adding a `<remarks>` section (where you can have your paragraphs.

ReubenBond, line 33: Should the default `virtualEnvironmentPath` value be taken from the `VIRTUAL_ENV` env var if it's set?

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        projectDirectory = PathNormalizer.NormalizePathForCurrentPlatform(Path.Combine(builder.AppHostDirectory, projectDirectory));
+        var virtualEnvironment = new VirtualEnvironment(Path.IsPathRooted(virtualEnvironmentPath)
+            ? virtualEnvironmentPath
+            : Path.Join(projectDirectory, virtualEnvironmentPath));
+
+        var instrumentationExecutable = virtualEnvironment.GetExecutable("opentelemetry-instrument");
+        var pythonExecutable = virtualEnvironment.GetRequiredExecutable("python");
```
ReubenBond, line 41: I wonder how best to handle something like [hatch](https://hatch.pypa.io/latest/environment/#command-execution), which runs projects using the `hatch run ...` command.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
mitchdenny: @wmeints just wondering if you've had a chance to come back and look at the <remarks> and <example> on this PR?

mitchdenny:
@wmeints this is 95% of the way there.

wmeints:
@mitchdenny I've had a few crazy weeks but I haven't forgotten about things! I'm going to take a look at the feedback to address issues. Thanks for reviewing :-)

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
wmeints: Good question, I didn't envision running commands other than the regular `python` command. I may add that after this PR is finished.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
wmeints: @ReubenBond I didn't go for this solution since I expect people to have multiple virtual environments, one for each project. That's at least how I've seen people around me use Aspire.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -14,21 +14,29 @@ public static class PythonProjectResourceBuilderExtensions
...
     /// Adds a python application with a virtual environment to the application model.
-    /// The Python project should have a virtual environment set up in the project directory under the name ".venv".
-    /// <para>
-    /// The python script in the project is automatically instrumented with opentelemetry when the virtual environment
-    /// contains the opentelemetry-instrument executable. You can get this by adding the opentelemtry-distro package
-    /// to your python project. In addition to the opentelemetry-distro package, you need to add the opentelemetry-exporter-otlp
-    /// for the traces, logs, and metrics to be exported to the Aspire observability platform.
-    /// </para>
     /// </summary>
+    /// <example>
```
mitchdenny, line 18: nit: Can you move the `<example>` below `<remarks>`. In the description for the example could you provide a little bit of detail about the layout of the project to help people understand what "PythonProject" refers to. Like the fact that it is relative to the AppHost directory.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        Assert.Equal("--traces_exporter", commandArguments[0]);
+        Assert.Equal("otlp", commandArguments[1]);
+        Assert.Equal("--logs_exporter", commandArguments[2]);
+        Assert.Equal("console,otlp", commandArguments[3]);
+        Assert.Equal("--metrics_exporter", commandArguments[4]);
+        Assert.Equal("otlp", commandArguments[5]);
+        Assert.Equal(pythonExecutable, commandArguments[6]);
+        Assert.Equal(scriptName, commandArguments[7]);
```
radical, line 77: nit: This will also check the length, and avoid typos with the indices.

```suggestion
        Assert.Collection(commandArguments,
            arg => Assert.Equal("--traces_exporter", arg),
            arg => Assert.Equal("otlp", arg),
            arg => Assert.Equal("--logs_exporter", arg),
            arg => Assert.Equal("console,otlp", arg),
            arg => Assert.Equal("--metrics_exporter", arg),
            arg => Assert.Equal("otlp", arg),
            arg => Assert.Equal(pythonExecutable, arg),
            arg => Assert.Equal(scriptName, arg));
```

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            Assert.Equal(Path.Join(projectDirectory, ".venv", "Scripts", "python.exe"), pythonProjectResource.Command);
+        }
+        else
+        {
+            Assert.Equal(Path.Join(projectDirectory, ".venv", "bin", "python"), pythonProjectResource.Command);
+        }
+
+        var commandArguments = await ArgumentEvaluator.GetArgumentListAsync(pythonProjectResource);
+
+        Assert.Equal(scriptName, commandArguments[0]);
```
radical, line 42: nit:
```suggestion
        Assert.Collection(commandArguments,
            arg => Assert.Equal(scriptName, arg));
```

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,159 @@
...
+        if (OperatingSystem.IsWindows())
+        {
+            var scriptsDirectory = Path.Join(virtualEnvDirectory, "Scripts");
+            Directory.CreateDirectory(scriptsDirectory);
+
+            File.WriteAllText(pythonExecutable, "");
+
+            if(instrument)
+            {
+                File.WriteAllText(Path.Join(scriptsDirectory, "opentelemetry-instrument.exe"), "");
```
radical, line 135: This path can also be returned from the method, which would simplify the checks for this path.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            Assert.Equal(Path.Join(projectDirectory, ".venv", "bin", "python"), pythonProjectResource.Command);
+        }
+
+        var commandArguments = await ArgumentEvaluator.GetArgumentListAsync(pythonProjectResource);
+
+        Assert.Equal(scriptName, commandArguments[0]);
+    }
+
+    [Fact]
+    public async Task AddPythonProjectWithInstrumentation_SwitchesExecutableToInstrumentationExecutable()
```
radical, line 46: This should have `instrument: true`, and `instrument: false` cases.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Python;
+using Aspire.Hosting.Utils;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Provides extension methods for adding Python applications to an <see cref="IDistributedApplicationBuilder"/>.
+/// </summary>
+public static class PythonProjectResourceBuilderExtensions
```
radical, line 13: It would be useful to add tests that *run* the python app too. See:
https://github.com/dotnet/aspire/blob/0af8505ff294c4b90ee8cccb5639bf73ec9077ba/tests/Aspire.Hosting.Tests/Node/NodeFunctionalTests.cs#L20-L32

https://github.com/dotnet/aspire/blob/0af8505ff294c4b90ee8cccb5639bf73ec9077ba/tests/Aspire.Hosting.Tests/TestProgramFixture.cs#L83-L92

https://github.com/dotnet/aspire/blob/0af8505ff294c4b90ee8cccb5639bf73ec9077ba/tests/testproject/TestProject.AppHost/TestProgram.cs#L64-L76

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+{
```
radical, line 14: Also, see the `node` related tests in https://github.com/dotnet/aspire/blob/0af8505ff294c4b90ee8cccb5639bf73ec9077ba/tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs#L336-L378

We could have similar for python too.

mitchdenny:
@davidfowl @wmeints so I've been thinking about what `AddPythonProject` does around defaulting to using PublishAsDockerfile. I think that for the initial merge we should leave it to emit an `executable.v0` resource and leave it up to developers to call `PublishAsDockerfile` if they want to.

My reasoning is that I want to add an extension method to `IResourceBuilder<ExecutableResource>` for `WithDockerfile` as the replacement to PublishAsDockerfile.

I think this will lead to less confusion and greater consistency.

src/Aspire.Hosting.Python/Aspire.Hosting.Python.csproj
```diff
@@ -0,0 +1,29 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Remove="PublicAPI.Shipped.txt" />
+    <None Remove="PublicAPI.Unshipped.txt" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="PublicAPI.Shipped.txt" />
+    <AdditionalFiles Include="PublicAPI.Unshipped.txt" />
```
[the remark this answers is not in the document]
wmeints: This was added by VS2022 and I wasn't aware that this was taken care of somewhere else. Removed the extra file entries.

mitchdenny:
@wmeints looks like we are both working on the branch. I switched you project generation code to use real python and actually setup a virtual environment. We need this so we can actually spin up the project in CI to verify things. @radical hopefully we have Python in the environment ;)

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
mitchdenny: Changed this up a bit to support running real Python.

wmeints:
@mitchdenny, I noticed that you were changing files. I'm going to watch things unfold for a bit. I love how this works on GitHub. I never expected to learn this much about testing a product like Aspire :)

mitchdenny:
@radical I think we need to kick the tests in `Aspire.Hosting.Tests` that were recently enabled for CI back to local only until we can fix the issue of the port conflicts.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        if (!createVirtualEnvironmentProcessResult)
+        {
+            createVirtualEnvironmentProcess.Kill(true);
+            throw new InvalidOperationException("Failed to create virtual environment.");
+        }
+
+        var relativePipPath = Path.Combine(
+            ".venv",
+            "Scripts",
```
radical, line 260:
```suggestion
            OperatingSystem.IsWindows() ? "Scripts" : "bin",
```

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        var prepareVirtualEnvironmentStartInfo = new ProcessStartInfo()
+        {
+            FileName = "python",
+            Arguments = $"-m venv .venv",
+            WorkingDirectory = projectDirectory
+        };
+
+        var createVirtualEnvironmentProcess = Process.Start(prepareVirtualEnvironmentStartInfo);
+        var createVirtualEnvironmentProcessResult = createVirtualEnvironmentProcess!.WaitForExit(10000);
```
radical, line 250: It would be very useful to capture the output too, incase the command fails, especially on CI.

tests/Aspire.Hosting.Tests/Python/AddPythonProjectTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        var installRequirementsStartInfo = new ProcessStartInfo()
+        {
+            FileName = pipPath,
+            Arguments = $"install -r requirements.txt",
+            WorkingDirectory = projectDirectory
+        };
+
+        var installRequirementsProcess = Process.Start(installRequirementsStartInfo);
+        var installRequirementsProcessResult = installRequirementsProcess!.WaitForExit(20000);
```
radical, line 273: Should the timeout be longer, for slower CI machines?

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// var builder = DistributedApplication.CreateBuilder(args);
+    ///
+    /// builder.AddPythonProject("python-project", "PythonProject", "main.py");
+    /// 
+    /// builder.Build().Run(); 
+    /// </code>
+    /// </example>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/> to add the resource to.</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="projectDirectory">The path to the directory containing the python project files.</param>
```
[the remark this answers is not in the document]
mitchdenny: Covered in remarks.

mitchdenny:
@wmeints thanks so much for your contribution to .NET Aspire. We'll be shipping your work in .NET Aspire 8.1!

davidfowl (approved):

mitchdenny (approved):

wmeints:
Excited to see this appear in 8.1! And thanks again for all the reviews and patience. I learned a lot from this experience. I'm coming back for more 😄

mitchdenny:
@wmeints I'm writing up the blog post for .NET Aspire 8.1 are you happy for me to link to your GitHub profile for this contribution?

wmeints:
@mitchdenny Feel free to link my profile in the blogpost :-)

PR #4151, opened 2024-05-10
Supporting health probes on resources

### LINKED ISSUE #890: Support modeling health checks into the AppHost model for use in dashboard, deployment manifest, etc.
We should consider adding support for modeling health checks endpoints into the AppHost app model so that resources' health checks endpoints can be integrated into the dashboard (#889) and emitted into the deployment manifest for potential auto-configuration by deployment tools, e.g. `azd` could configure [HTTP health probes when deploying Aspire apps to ACA](https://learn.microsoft.com/en-us/azure/container-apps/health-probes?tabs=arm-template#http-probes).

GMouaad:
Added first draft of supporting health probes on resources

Contributes to #890 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4151)


davidfowl:
Thanks for opening the PR to start the discussion. We need to decide what reads these annotations and how they appear in the manifest.

GMouaad:
@davidfowl Thanks for the quick feedback, you were quicker then my comment in the issue :D. 

> We need to decide what reads these annotations and how they appear in the manifest.

Definitly, I mentioned this in the open points on my [comment no the issue](https://github.com/dotnet/aspire/issues/890#issuecomment-2105027885). AFAIK, the only think that the platforms would need is to know the **endpoint type** (plain TCP or HTTP Request) and on which port it should call the resource (the "internal" port and not the exposed one I would imagine). 

Maybe some input from the DCP folks is needed (?)

My initial idea is to just make as simple as possible and reflect the common field needed. It would look something like the follwing:
```json
{
  "resources": {
    "HelloWorldApp": {
      "type": "container.v0",
      "image": "docker.io/mgsair/hello-world-java:latest",
      "env": {
        "SERVER_PORT": "8000"
      },
      "bindings": {
        "endpoint": {
          "scheme": "http",
          "protocol": "tcp",
          "transport": "http",
          "containerPort": 8000
        }
      },
      "probes": [
        {
          "type": "liveness",
          "httpGet": {
            "path": "/health",
            "port": 8000
          },
          "initialDelaySeconds": 5,
          "periodSeconds": 5
        }
      ]
    },
    "apiservice": {
      "type": "project.v0",
      "path": "../AspireWithContainers.ApiService/AspireWithContainers.ApiService.csproj",
      "env": {
        "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES": "true",
        "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES": "true",
        "services__HelloWorldApp__0": "{HelloWorldApp.bindings.endpoint.url}"
      },
      "bindings": {
        "http": {
          "scheme": "http",
          "protocol": "tcp",
          "transport": "http"
        },
        "https": {
          "scheme": "https",
          "protocol": "tcp",
          "transport": "http"
        }
      }
    },
    "webfrontend": {
      "type": "project.v0",
      "path": "../AspireWithContainers.Web/AspireWithContainers.Web.csproj",
      "env": {
        "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES": "true",
        "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES": "true",
        "services__apiservice__0": "{apiservice.bindings.http.url}",
        "services__apiservice__1": "{apiservice.bindings.https.url}"
      },
      "bindings": {
        "http": {
          "scheme": "http",
          "protocol": "tcp",
          "transport": "http"
        },
        "https": {
          "scheme": "https",
          "protocol": "tcp",
          "transport": "http"
        }
      }
    }
  }
}
```
I'll try to include this in my draft so other parties can play around with it maybe.

Any thoughts?

GMouaad:
After giving more thoughts, I think it's not trivial to figure out which port should be used, from the example: `8000`or `{HelloWorldApp.bindings.endpoint.port}`, i.e. `endpointReference.Port` or `endpointReference.GetExpression(EndpointProperty.Port)`.

mitchdenny:
Like @davidfowl said thanks for opening up the PR.

I'm wondering whether we should model probes as a sub-field on the endpoint? The other thing to think about here is whether there is some kind of default we apply to the probe path based on the defaults we have in Aspire application libraries.

GMouaad:
Thank you for the feedback @mitchdenny !

> The other thing to think about here is whether there is some kind of default we apply to the probe path based on the defaults we have in Aspire application libraries.

I think that would be either something we can add in the service defaults or as supplementary extensions that uses the `WithProbe(..)` under the hood?

src/Aspire.Hosting/ApplicationModel/IResourceWithProbes.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents a resource that has endpoints associated with it.
+/// </summary>
+public interface IResourceWithProbes : IResource
```
davidfowl, line 9: It's unclear why we need this.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +549,27 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    /// </summary>
+    /// <typeparam name="T">Type of resource.</typeparam>
+    /// <param name="builder">Resource builder.</param>
+    /// <param name="type">The type of the health probe.</param>
+    /// <param name="endpoint">The endpoint reference to be used for the probe.</param>
+    /// <param name="path">The path to be used in case of a http endpoint.</param>
+    /// <param name="initialDelaySeconds">The initial delay before calling the probe endpoint for the first time.</param>
+    /// <param name="periodSeconds">The period between each probe.</param>
+    /// <returns></returns>
+    public static IResourceBuilder<T> WithProbe<T>(this IResourceBuilder<T> builder, ProbeType type,
```
davidfowl, line 564: This should just take an endpoint name.

davidfowl:
@GMouaad sorry we didn't get around to going into the design here for the release. The team is quite heads down right now on the Aspire 9 release. I think you've got a pretty good start with this PR, the APIs look great (I added some comments).

The bigger changes come around adding runtime and deployment support for these. For the runtime support, we just added a health checking system to the apphost, this might just be plugged into that. cc @mitchdenny 

For deployment we have more options to figure out how this maps

mitchdenny:
Now that we have `WaitFor` in which ties into health checks I think that there is a fairly clear path here now of how probes (especially readiness probes) might work for local development. Consider the following code:

```csharp
builder.AddProject<Projects.Project1>("project1")
       .WithReadinessProbe("endpointName", "/healthz", ...);
```

This would add a readiness probe annotation to the model which could be synthesised into a health check annotation which could then be used to do things like this:

```csharp
var p1 = builder.AddProject<Projects.Project1>("project1")
                .WithReadinessProbe("endpointName", "/healthz", ...);
builder.AddProject<Projects.Projec2>("p2")
       .WithReference(p1.GetEndpoint("https")).WaitFor(p1);
```

GMouaad:
@davidfowl I thought also that you'll be busy with the release topics. Thank you for the feedback. 
Regarding the health checking system, great news! Thanks @mitchdenny for the sample and hints. 

I'm currently abroad, I'll resolve the issues and comments and play arround with the new APIs as soon as come back!

GMouaad:
@davidfowl @mitchdenny I messed I'm up my branch while trying it from another IDE. I reopened the PR with the change from your feedback in #6379. Sorry for the inconvenience..

I'll close this one for the moment.

PR #4153, opened 2024-05-10
Add response compression for static assets to the dashboard

### LINKED ISSUE #4150: Dashboard should configure response compression & caching for static assets
The dashboard currently doesn't configure compression or caching headers for the static assets it delivers, including libraries like plotly (which is [3.5 MB](https://github.com/dotnet/aspire/issues/4149)).

We should configure the response compression middleware to ensure that at least static assets are compressed (and potentially the HTML output too), and investigate setting cache headers for static assets with versioned paths to help improve the client-side experience.

DamianEdwards:
Fixes #4150
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4153)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -124,8 +124,9 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         builder.Services.AddResponseCompression(options =>
         {
             options.EnableForHttps = true;
-            // Avoid compressing HTML to mitigate user supplied data being compressed over HTTPS
-            options.ExcludedMimeTypes = ["text/html"];
+            // Limit to compressing static text assets to mitigate user supplied data being compressed over HTTPS
+            // See https://learn.microsoft.com/aspnet/core/performance/response-compression#compression-with-https for more information
+            options.MimeTypes = ["text/javascript", "application/javascript", "text/css", "image/svg+xml"];
         });
 
         // Data from the server.
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -121,6 +121,12 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         // Add services to the container.
         builder.Services.AddRazorComponents().AddInteractiveServerComponents();
         builder.Services.AddCascadingAuthenticationState();
+        builder.Services.AddResponseCompression(options =>
+        {
+            options.EnableForHttps = true;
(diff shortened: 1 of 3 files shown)


drewnoakes (approved):

DamianEdwards:
@davidfowl @blowdart any comments on this one? It's pretty straightforward but there's nuance in where in the middleware pipeline it's enabled and enabling compression over HTTPS but not for HTML, to mitigate any security issues stemming from being able to force user content to be compressed over HTTPS.

blowdart:
Funnily enough Javier is doing the same thing as a feature - you should sync with him for his new idea which might mitigate the need to do this in a custom manner. (But generally, no, no concerns for this type of thing as it's static assets)

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -121,6 +121,12 @@ public DashboardWebApplication(Action<WebApplicationBuilder>? configureBuilder =
         // Add services to the container.
         builder.Services.AddRazorComponents().AddInteractiveServerComponents();
         builder.Services.AddCascadingAuthenticationState();
+        builder.Services.AddResponseCompression(options =>
+        {
+            options.EnableForHttps = true;
+            // Avoid compressing HTML to mitigate user supplied data being compressed over HTTPS
```
danmoseley, line 127: Should the comment say because of side channel attack? Or maybe that's obvious.

DamianEdwards: Maybe just link to [our docs about it](https://learn.microsoft.com/aspnet/core/performance/response-compression?view=aspnetcore-8.0#compression-with-https)?

danmoseley: Seems reasonable

JamesNK: What are the rules when it comes to Blazor compression? Are they different from regular server rendered HTML?

We only want static resources compressed:
* JavaScript
* CSS
* Images
* SVG

Everything else should be excluded:
* Communication over SignalR websockets
* gRPC calls (remember the dashboard has some gRPC endpoints)
* Re-rendered HTML (I think that's disabled)

JamesNK: I'd feel safer with an allow list than a deny list.

DamianEdwards: @JamesNK changed it to use an allow list

JamesNK (approved):

PR #4154, opened 2024-05-10
Add playwright test for changing theme

adamint:
Aspire.Hosting.Tests/Dashboard may not be the best place to put this test long-term. I made it local only for now.

I was unable to access the dashboard frontend from a browser when trying the approach in Aspire.Dashboard.Tests of just creating the dashboard frontend app - it would not connect. If anyone has an idea why that might be, I'm all ears.

I also ran into gRPC issues after orchestrating. Also seeking feedback on that, but in the meantime I injected a mock dashboard service so that there is no grpc connection.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4154)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,6 +15,7 @@
     <PackageVersion Include="AWSSDK.Core" Version="3.7.303.13" />
     <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.300" />
     <PackageVersion Include="AWS.Messaging" Version="0.9.0" />
+    <PackageVersion Include="Microsoft.Playwright" Version="1.43.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.3" />
     <PackageVersion Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
     <!-- Azure SDK for .NET dependencies -->
--- a/tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
+++ b/tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Aspire.Hosting.Tests.Helpers;
 using Microsoft.Playwright;
 using Xunit;
 
@@ -15,7 +16,7 @@ public class AppbarTests : IClassFixture<PlaywrightFixture>
         _playwrightFixture = playwrightFixture;
     }
 
-    [Fact]
+    [LocalOnlyFact]
     public async Task Dashboard_Test_Change_Theme()
     {
         var (_, page) = await _playwrightFixture.SetupDashboardForPlaywrightAsync();
--- a/tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
+++ b/tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Playwright;
+using Xunit;
+
+namespace Aspire.Hosting.Tests.Dashboard;
+
+public class AppbarTests : IClassFixture<PlaywrightFixture>
+{
+    private readonly PlaywrightFixture _playwrightFixture;
+
+    public AppbarTests(PlaywrightFixture playwrightFixture)
+    {
+        _playwrightFixture = playwrightFixture;
+    }
+
+    [Fact]
+    public async Task Dashboard_Test_Change_Theme()
+    {
+        var (_, page) = await _playwrightFixture.SetupDashboardForPlaywrightAsync();
+        var settingsButton = page.GetByRole(AriaRole.Button, new()
+        {
+            Name = "Launch settings"
+        });
+        await settingsButton.ClickAsync();
+        var darkThemeCheckbox = page.GetByRole(AriaRole.Radio).And(page.GetByText("Dark")).First;
+        await darkThemeCheckbox.ClickAsync();
+        Assert.Equal("dark", await page.Locator("html").First.GetAttributeAsync("data-theme"));
+    }
+}
--- a/tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
+++ b/tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
@@ -1,0 +1,84 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Dashboard.Model;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Playwright;
+using Xunit;
+
+namespace Aspire.Hosting.Tests.Dashboard;
+
+public class PlaywrightFixture : IAsyncLifetime
+{
+    public async Task InitializeAsync()
+    {
+        PlaywrightInstance = await Playwright.CreateAsync();
+        Browser = await PlaywrightInstance.Chromium.LaunchAsync(new() { Headless = false });
+    }
+
+    public IBrowser Browser { get; set; } = null!;
+    private IPlaywright PlaywrightInstance { get; set; } = null!;
+
+    public async Task DisposeAsync()
+    {
+        await Browser.DisposeAsync();
+        PlaywrightInstance.Dispose();
+    }
+
+    private sealed class MockDashboardClient : IDashboardClient
+    {
+        public bool IsEnabled => true;
+        public Task WhenConnected => Task.CompletedTask;
+        public string ApplicationName => "<marquee>An HTML title!</marquee>";
+        public ValueTask DisposeAsync() => ValueTask.CompletedTask;
+        public Task<ResourceCommandResponseViewModel> ExecuteResourceCommandAsync(string resourceName, string resourceType, CommandViewModel command, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public IAsyncEnumerable<IReadOnlyList<ResourceLogLine>>? SubscribeConsoleLogs(string resourceName, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public Task<ResourceViewModelSubscription> SubscribeResourcesAsync(CancellationToken cancellationToken) => throw new NotImplementedException();
+    }
+
+    public async Task<(string DashboardUrl, IPage Page)> SetupDashboardForPlaywrightAsync()
+    {
+
+        var url = "http://localhost:1234";
+
+        var args = new string[] {
+            "ASPNETCORE_ENVIRONMENT=Development",
+            "DOTNET_ENVIRONMENT=Development",
+            $"ASPNETCORE_URLS={url}",
+            $"DOTNET_DASHBOARD_OTLP_ENDPOINT_URL={url}5",
+            "DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS=true",
+            "DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES=true"
+        };
+
+        using var testProgram = TestProgram.Create<PlaywrightFixture>(
+            args,
+            includeIntegrationServices: false,
+            disableDashboard: false);
+        var services = testProgram.AppBuilder.Services;
+        services.AddSingleton<ILoggerFactory>(NullLoggerFactory.Instance);
+        services.AddSingleton<IDashboardClient, MockDashboardClient>();
+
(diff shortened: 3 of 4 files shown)


Directory.Packages.props
```diff
@@ -161,4 +162,4 @@
     <PackageVersion Include="Newtonsoft.Json.Schema" Version="3.0.15" />
     <!-- Introduced by Microsoft.Azure.Cosmos, https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4302 -->
   </ItemGroup>
-</Project>
```
tlmii, line 164: Inadvertent change?

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        public bool IsEnabled => true;
+        public Task WhenConnected => Task.CompletedTask;
+        public string ApplicationName => "<marquee>An HTML title!</marquee>";
+        public ValueTask DisposeAsync() => ValueTask.CompletedTask;
+        public Task<ResourceCommandResponseViewModel> ExecuteResourceCommandAsync(string resourceName, string resourceType, CommandViewModel command, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public IAsyncEnumerable<IReadOnlyList<ResourceLogLine>>? SubscribeConsoleLogs(string resourceName, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public Task<ResourceViewModelSubscription> SubscribeResourcesAsync(CancellationToken cancellationToken) => throw new NotImplementedException();
+    }
+
+    public async Task<(string DashboardUrl, IPage Page)> SetupDashboardForPlaywrightAsync()
```
tlmii, line 41: Not a huge fan of tuples in a public surface. But it's test code so probably OK.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        async Task WaitUntilDashboardReadyAsync()
+        {
+            try
+            {
+                await page.GotoAsync(url);
+            }
+            catch (Exception)
+            {
+                await Task.Delay(1000, cts.Token);
+                await WaitUntilDashboardReadyAsync();
```
tlmii, line 80: Should there be a point at which this bails out?

tlmii (commented):
Agree on the confusion of why it didn't work the way you originally tried it. It'd also be nice to get some feedback from the Hosting folks about pulling in the dashboard ref here. Just test code, so probably fine, but an ack would be nice.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
danmoseley: also probably a loop rather than recursion

radical:
FYI, I'm adding some playwright based tests in https://github.com/dotnet/aspire/pull/3270 . But in that I'm using `dotnet run` to start up a real app, and then interact with it.

[This](https://github.com/radical/aspire/blob/5a900f1f44018a84e6741c45c4d632a1b6e35090/tests/Shared/WorkloadTesting/AspireProject.cs#L295-L305) waits for the dashboard to become available, and uses resiliency to handle retries - https://github.com/radical/aspire/blob/5a900f1f44018a84e6741c45c4d632a1b6e35090/tests/Shared/WorkloadTesting/AspireProject.cs#L408-L427 .


And [this](https://github.com/radical/aspire/blob/5a900f1f44018a84e6741c45c4d632a1b6e35090/tests/Aspire.Workload.Tests/WorkloadTestsBase.cs#L26-L138) connects with playwright, and verifies that the expected resources show up on the dashboard.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        return (url, page);
+
+        async Task WaitUntilDashboardReadyAsync()
+        {
+            try
+            {
+                await page.GotoAsync(url);
+            }
+            catch (Exception)
```
danmoseley, line 77: do we want to retry on any exception type other than PlaywrightException ?

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
danmoseley: oh, looks like that's the 1 minute timeout on the cts. still, feels like a loop

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                await Task.Delay(1000, cts.Token);
```
danmoseley, line 79: is 1000 reasonable? say, pages typically take 1100 ms to be ready, then it's adding 900ms delay on every test, and perhaps we end with 100 tests.

sometimes in these cases you see a backoff strategy. So starting with brief delays (or just calling GotoAsync again immediately) so the happy path on a machine where it's available fast has no delay. Then backing off progressively the delay for slower machines to release resources for the pages to actually start.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        PlaywrightInstance = await Playwright.CreateAsync();
+        Browser = await PlaywrightInstance.Chromium.LaunchAsync(new() { Headless = false });
+    }
+
+    public IBrowser Browser { get; set; } = null!;
+    private IPlaywright PlaywrightInstance { get; set; } = null!;
+
+    public async Task DisposeAsync()
+    {
+        await Browser.DisposeAsync();
```
danmoseley, line 26: Browser.CloseAsync() seems canonical in their examples

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        PlaywrightInstance.Dispose();
```
danmoseley, line 27: Does it need disposing? I can't find in the docs.

I think in general you don't dispose `Task<T>` without a reason
https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.dispose?view=net-8.0#system-threading-tasks-task-dispose

tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.Tests.Helpers;
+using Microsoft.Playwright;
+using Xunit;
+
+namespace Aspire.Hosting.Tests.Dashboard;
+
+public class AppbarTests : IClassFixture<PlaywrightFixture>
```
danmoseley, line 10: is this the Fluent AppBar? In which case it should be capitalized AppBar

tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        var (_, page) = await _playwrightFixture.SetupDashboardForPlaywrightAsync();
+        var settingsButton = page.GetByRole(AriaRole.Button, new()
+        {
+            Name = "Launch settings"
+        });
+        await settingsButton.ClickAsync();
+        var darkThemeCheckbox = page.GetByRole(AriaRole.Radio).And(page.GetByText("Dark")).First;
+        await darkThemeCheckbox.ClickAsync();
+        Assert.Equal("dark", await page.Locator("html").First.GetAttributeAsync("data-theme"));
```
danmoseley, line 30: Assert it's not dark before you click the box?

danmoseley: and click again and assert it's light again?

danmoseley: also will this fail if the OS is set to dark theme up front (prefers-color-scheme: dark) ? if so it should instead test for what it is, click the box, check it changed, click the box, check it changed again.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        _playwrightFixture = playwrightFixture;
+    }
+
+    [LocalOnlyFact]
+    public async Task Dashboard_Test_Change_Theme()
+    {
+        var (_, page) = await _playwrightFixture.SetupDashboardForPlaywrightAsync();
+        var settingsButton = page.GetByRole(AriaRole.Button, new()
+        {
+            Name = "Launch settings"
```
danmoseley, line 25: This is matching the English string, right? Is there a way to find this button on any locale, as we want to test all localized builds. Perhaps via a `class` or `id` ?

tests/Aspire.Hosting.Tests/Dashboard/Playwright/AppbarTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        });
+        await settingsButton.ClickAsync();
+        var darkThemeCheckbox = page.GetByRole(AriaRole.Radio).And(page.GetByText("Dark")).First;
```
danmoseley, line 28: same issue with assuming English.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        public ValueTask DisposeAsync() => ValueTask.CompletedTask;
+        public Task<ResourceCommandResponseViewModel> ExecuteResourceCommandAsync(string resourceName, string resourceType, CommandViewModel command, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public IAsyncEnumerable<IReadOnlyList<ResourceLogLine>>? SubscribeConsoleLogs(string resourceName, CancellationToken cancellationToken) => throw new NotImplementedException();
+        public Task<ResourceViewModelSubscription> SubscribeResourcesAsync(CancellationToken cancellationToken) => throw new NotImplementedException();
+    }
+
+    public async Task<(string DashboardUrl, IPage Page)> SetupDashboardForPlaywrightAsync()
+    {
+
+        var url = "http://localhost:1234";
```
danmoseley, line 44: does it need a port at all?

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+            $"ASPNETCORE_URLS={url}",
+            $"DOTNET_DASHBOARD_OTLP_ENDPOINT_URL={url}5",
+            "DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS=true",
+            "DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES=true"
+        };
+
+        using var testProgram = TestProgram.Create<PlaywrightFixture>(
+            args,
+            includeIntegrationServices: false,
+            disableDashboard: false);
```
danmoseley, line 58: is this includeNodeApp actually?
https://github.com/dotnet/aspire/blob/612429cfc9efa22a143e124b10c4e6a29821448f/tests/testproject/TestProject.AppHost/TestProgram.cs#L151-L157

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        var page = await Browser.NewPageAsync();
+        await WaitUntilDashboardReadyAsync();
+
+        return (url, page);
+
+        async Task WaitUntilDashboardReadyAsync()
+        {
+            try
+            {
+                await page.GotoAsync(url);
```
danmoseley, line 75: I was looking to see whether there's a way to pass in the cts. But one can only pass a timeout (via PageGotoOptions) which is otherwise 30 seconds .. so I guess no reason to change that.

tests/Aspire.Hosting.Tests/Dashboard/Playwright/PlaywrightFixture.cs
danmoseley: hmm but if that's so how is the named param compiling? I must be missing an overload

kvenkatrajan:
Was this closed by accident?

adamint:
@kvenkatrajan we are going to use Ankit’s e2e infrastructure so 90% of this pr content will be replaced :)

danmoseley:
thanks for working together on this, sounds like we'll end up with a single infrastructure and it will be easier for everyone.

(when you have your new PR could you check whether any of the feedback above still applies?)

PR #4168, opened 2024-05-13
[tests] Workload tests

### LINKED ISSUE #47: Add tests for templates
The templates currently have no unit test coverage. We should look at porting some of the template testing infrastructure from dotnet/aspnetcore to enable us to get some template test coverage.

radical:
This adds `Aspire.Workload.Tests` which are meant to exercise the `aspire` workload. The individual tests create projects from templates just like a user would, and then run, and validate them.

It uses the existing infrastructure from `Aspire.EndToEnd.Tests` to install sdk+workload for testing.

- Enables tests on Linux, and Windows
- Because we don't have docker support on helix/Windows yet, the docker dependent tests are skipped on windows

## Tests

Two kinds of tests are being added here:

1. Tests that create+build a `aspire-starter` template project, and then some tests interact with that running app to validate it
2. Individual tests exercise the various templates, for example to create `aspire-starter` with [xunit,nunit,mstest] options

- The tests also connect to the dashboard and validate that the expected services show up as expected using `Playwright`.
- Some tests also connect to the webapp in `aspire-starter` template to validate some parts of it.

Fixes https://github.com/dotnet/aspire/issues/47 .

Follow up work is tracked in https://github.com/dotnet/aspire/issues/4222

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4168)


dogfood.sh
```diff
@@ -0,0 +1,28 @@
+if [ -z "$ZSH_VERSION" ]; then
```
joperezr, line 1: NIT: should probably have `#!/usr/bin/env bash` at the top line of this script.

joperezr: Also, I don't see in other scripts in the repo us checking if you are using Bash or Zsh, is there a reason why we need that here?

radical: This is meant to be used like `source dogfood.sh`, so that wouldn't be appropriate AFAIU.

dogfood.sh
```diff
@@ -0,0 +1,28 @@
...
+  done
+  scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
+else
+  # :A will resolve all symlinks, :h will truncate last path component leaving you with a directory name
+  scriptroot=${0:A:h}
+fi
+
+REPO_ROOT=$(cd "${scriptroot}";pwd)
+
+PKG_DIR=${REPO_ROOT}/artifacts/packages/Release/Shipping
```
joperezr, line 20: Should the script instead take configuration parameter?

dogfood.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ ! -d "$PKG_DIR" ]; then
+  PKG_DIR=${REPO_ROOT}/artifacts/packages/Debug/Shipping
+fi
+
+export PATH=$REPO_ROOT/artifacts/bin/dotnet-latest:$PATH
+echo Setting export BUILT_NUGETS_PATH=$PKG_DIR to resolve locally built packages
+export BUILT_NUGETS_PATH=$PKG_DIR
+echo "Use $REPO_ROOT/tests/Aspire.Workload.Tests/data/nuget8.config as your local nuget.config to ensure the packages can be restored"
```
joperezr, line 28: Do we need a dogfood.ps1 for Windows too?

dogfood.sh
radical: `${BASH_SOURCE[0]}` is not supported in `zsh`.

dogfood.sh
radical: Sure, we can do that. Auto-detect seemed easier in use.

dogfood.sh
radical: Yes. Coming soon in the next PR 😬

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -21,26 +21,14 @@ parameters:
...
-              -pack
-              -configuration ${{ parameters.buildConfig }}
-              /p:ArchiveTests=true
-              /bl:${{ parameters.repoLogPath }}/build.binlog
-              $(extraBuildArgs)
-              $(_OfficialBuildIdArgs)
-      displayName: Build
+  - script: ${{ parameters.buildScript }}
+            -restore -build
+            -pack
```
joperezr, line 26: Since we are now calling pack here always, we should probably start calling sign here too and remove the extra step we have later in this file called "Pack, Sign, and Publish". The original thinking of the split was that packacging and signing took longer so those should happen only after tests have validated things are good, but today they mostly just redo a bunch of stuff that we are already doing here, so might as well just merge the two.

tests/Directory.Build.props
```diff
@@ -6,6 +6,10 @@
     <TestArchiveTestsDir Condition="'$(TestArchiveTestsDir)' == ''">$([MSBuild]::NormalizeDirectory($(ArtifactsDir), 'helix', 'tests'))</TestArchiveTestsDir>
     <TestArchiveTestsDirForWorkloadTests Condition="'$(TestArchiveTestsDirForWorkloadTests)' == ''">$([MSBuild]::NormalizeDirectory($(ArtifactsDir), 'helix', 'workload-tests'))</TestArchiveTestsDirForWorkloadTests>
     <TestArchiveTestsDirForEndToEndTests Condition="'$(TestArchiveTestsDirForEndToEndTests)' == ''">$([MSBuild]::NormalizeDirectory($(ArtifactsDir), 'helix', 'e2e-tests'))</TestArchiveTestsDirForEndToEndTests>
+    <PlaywrightDependenciesDirectory>$(ArtifactsBinDir)playwright-deps</PlaywrightDependenciesDirectory>
+
+    <!-- This is useful for local runs -->
+    <VSTestResultsDirectory>$(ArtifactsTestResultsDir)</VSTestResultsDirectory>
```
joperezr, line 12: I assume they were still saved before, just inside the project directory?

tests/Aspire.Workload.Tests/data/nuget8.config
```diff
@@ -0,0 +1,17 @@
...
+  <!-- Don't use any higher level config files. -->
+  <fallbackPackageFolders>
+    <clear />
+  </fallbackPackageFolders>
+  <packageSources>
+    <clear />
+    <!-- TEST_RESTORE_SOURCES_INSERTION_LINE -->
+    <add key="built-local" value="%BUILT_NUGETS_PATH%" />
+    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
+    <add key="nuget.org"  value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
```
joperezr, line 12: I may be wrong, but we may get flagged in compliance for having nuget.org as a feed (even when just used for testing.)

dogfood.sh
joperezr: I see, I think that having it be a `.sh` will be misleading then if the intent is to do source dogfood.sh. Just curious, why not make it an actual bash script that can be executed? Here is precedent of a very similar use cas in the sdk repo: https://github.com/dotnet/sdk/blob/main/eng/dogfood.sh

eng/pipelines/templates/BuildAndTest.yml
radical: I want to tackle this one in a separate PR. Because for the tests I'm using a more constrained project `tests/workloads.proj` which results in faster installs. In a follow up I will check how to merge these two steps. Or probably be able to re-use parts of it.

joperezr: Sounds good to me, let's just make sure that we have an issue or something to track the de-duplication as that would save a bunch of build time.

tests/Directory.Build.props
radical: Yes.

joperezr (commented):
I think it would be really great if we could add some docs somewhere that explain with detail how are people supposed to run these tests locally. Same goes for the rest of the infra and docker tests that we have added in the past, I don't believe I've seen those docs anywhere. I'm fine if we do that as a follow up change but just let's make sure we do get to it.

radical:
> I think it would be really great if we could add some docs somewhere that explain with detail how are people supposed to run these tests locally. Same goes for the rest of the infra and docker tests that we have added in the past, I don't believe I've seen those docs anywhere. I'm fine if we do that as a follow up change but just let's make sure we do get to it.

See `README.md` in `tests/Aspire.Workload.Tests`, and `tests/Aspire.EndToEnd.Tests`.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,59 @@
...
+  <Target Name="_ExtractTestClassNames"
+          Condition="'$(ExtractTestClassNamesForHelix)' == 'true'"
+          BeforeTargets="ZipTestArchive">
+
+    <Exec Command="&quot;$(DotNetTool)&quot; test --no-build -c $(Configuration) -s $(RunSettingsFilePath) -t --nologo -v:q -p:VsTestUseMSBuildOutput=false" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" ItemName="_ListOfTestsLines" />
+    </Exec>
+
+    <PropertyGroup>
+      <_Regex>^ *(Aspire.Workload.Tests[^\($]+)</_Regex>
```
joperezr, line 46: NIT: `^\s*(Aspire\.Workload\.Tests[^\($]+)`

tests/Aspire.Workload.Tests/Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <PropertyGroup>
+    <IsWorkloadTestProject>true</IsWorkloadTestProject>
```
joperezr, line 3: Small NIT: it is sometimes helpful to add comments above properties like this to just explain what this will do. For instance, this will import workload testing props and targets.

dogfood.sh
radical: If the script is run as `/foo/dogfood.sh` then the changes to the environment will be limited to that script. The documentation in sdk also suggests using `source dogfood.sh`:

https://github.com/dotnet/sdk/blob/9e277c9ce470d35ee4ebe751a5629be8688f5037/documentation/project-docs/developer-guide.md?plain=1#L108

When using `source ..` we are using the current shell to run the commands so the shebang should be not be needed, AFAIU.

joperezr:
> > I think it would be really great if we could add some docs somewhere that explain with detail how are people supposed to run these tests locally. Same goes for the rest of the infra and docker tests that we have added in the past, I don't believe I've seen those docs anywhere. I'm fine if we do that as a follow up change but just let's make sure we do get to it.
> 
> See `README.md` in `tests/Aspire.Workload.Tests`, and `tests/Aspire.EndToEnd.Tests`.

Oh, missed that! Would be great to link those somewhere from our Contributing docs. Essentially, imagine you are a brand new contributor to the repo and want to contribute a change and want to run all tests locally, but don't know where to start. Having that top level documentation on how to build, and how to run each type of test is really helpful.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,252 @@
...
+        var t = Task.Run(async () =>
+        {
+            var playwright = await Playwright.CreateAsync();
+            BrowserTypeLaunchOptions? options = null;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                options = new BrowserTypeLaunchOptions
+                {
+                    Headless = false,
+                    ExecutablePath = "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
joperezr, line 35: NIT: should this be configurable (via env var or something) or is it always guaranteed to be there?

Directory.Packages.props
```diff
@@ -159,5 +159,6 @@
     <!-- Pinned version for Component Governance - Remove when root dependencies are updated -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="Newtonsoft.Json.Schema" Version="3.0.16" />
+    <PackageVersion Include="Microsoft.Playwright" Version="1.42.0" />
```
eerhardt, line 162: Can you move this to be under "unit test dependencies"?

tests/PlaywrightWrapper/Program.cs
```diff
@@ -0,0 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+//
+sealed class Program
+{
+    public static int Main(string[] args)
+        => Microsoft.Playwright.Program.Main(args);
```
eerhardt, line 7: Why do we need this wrapper? Why can't we just call Microsoft.Playwright directly?

(It would be good to put the answer in a comment in the code)

tests/helix/send-to-helix-endtoendtests.targets
```diff
@@ -32,11 +32,16 @@
...
       <_E2ETestScenarios Include="cosmos" />
+      <!-- needed for batching below -->
+      <_E2ETestScenarios PreCommands="" />
 
       <HelixWorkItem Include="@(_E2ETestScenarios -> '$(_WorkItemName)-%(Identity)')">
         <PayloadArchive>$(_E2ETestsArchivePath)</PayloadArchive>
-        <PreCommands Condition="'$(OS)' == 'Windows_NT'">set &quot;TEST_NAME=$(_WorkItemName)&quot; &amp; set TEST_SCENARIO=%(Identity) &amp; set &quot;CODE_COV_FILE_SUFFIX=-%(Identity)&quot;</PreCommands>
-        <PreCommands Condition="'$(OS)' != 'Windows_NT'">export &quot;TEST_NAME=$(_WorkItemName)&quot; &amp;&amp; export TEST_SCENARIO=%(Identity) &amp;&amp; export &quot;CODE_COV_FILE_SUFFIX=-%(Identity)&quot;</PreCommands>
+
+        <PreCommands                                         >$(_EnvVarSetKeyword) &quot;TEST_NAME=$(_WorkItemName)&quot;</PreCommands>
```
eerhardt, line 41: This doesn't look readable/maintainable. I would remove the whitespace.

tests/Aspire.EndToEnd.Tests/Directory.Build.props
```diff
@@ -1,7 +1,12 @@
 <Project>
   <PropertyGroup>
+    <!-- This triggers the import of targets needed for workload testing -->
     <IsWorkloadTestProject>true</IsWorkloadTestProject>
   </PropertyGroup>
 
   <Import Project="..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
```
eerhardt, line 10: Isn't this already set in the .csproj?

tests/Shared/WorkloadTesting/AspireProject.cs
```diff
@@ -36,13 +47,69 @@ public AspireProject(string id, string baseDir, ITestOutputHelper testOutput, Bu
...
+    }
+
     public async Task StartAppHostAsync(string[]? extraArgs = default, Action<ProcessStartInfo>? configureProcess = null, bool noBuild = true, CancellationToken token = default)
     {
         if (IsRunning)
         {
             throw new InvalidOperationException("Project is already running");
         }
 
+        Stopwatch runTimeStopwatch = new();
```
eerhardt, line 112: Can you move this closer to where it is used?

tests/Shared/WorkloadTesting/AspireProject.cs
```diff
@@ -199,6 +286,51 @@ public async Task BuildAsync(string[]? extraBuildArgs = default, CancellationTok
...
+        cts.CancelAfter(TimeSpan.FromSeconds(DashboardAvailabilityTimeoutSecs));
+        await WaitForDashboardToBeAvailableAsync(dashboardUrlToUse, _testOutput, cts.Token).ConfigureAwait(false);
+
+        var dashboardPage = await context.NewPageWithLoggingAsync(_testOutput);
+        await dashboardPage.GotoAsync(dashboardUrlToUse);
+
+        return dashboardPage;
+    }
+
+    public Task WaitForDashboardToBeAvailableAsync(CancellationToken token = default)
```
eerhardt, line 315:
```suggestion
    public Task WaitForDashboardToBeAvailableAsync(CancellationToken cancellationToken = default)
```

This is a better name than `token`.

tests/Aspire.Workload.Tests/data/Endpoint_cs
```diff
@@ -0,0 +1,42 @@
+using Aspire.Hosting.Lifecycle;
```
eerhardt, line 1: Can you rename this file to match the class name?

tests/Aspire.EndToEnd.Tests/Directory.Build.props
radical: oops, yes!

tests/Aspire.Workload.Tests/data/nuget8.config
eerhardt: Why do we need nuget.org? We should have everything we need in our normal nuget.config (plus the local path)

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,59 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+
+    <RunTestsOnHelix>true</RunTestsOnHelix>
+    <TestUsingWorkloads>true</TestUsingWorkloads>
+    <InstallWorkloadForTesting>$(TestUsingWorkloads)</InstallWorkloadForTesting>
```
eerhardt, line 8: This is kind of weird. Should this just be the default if `TestusingWorkloads` is true?

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,59 @@
...
+
+    <BuiltNuGetsDir>$(ArtifactsShippingPackagesDir)</BuiltNuGetsDir>
```
eerhardt, line 10: Why does this need to be set in each .csproj? Can't it be the default?

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,59 @@
...
+    <PackageVersionForWorkloadManifests>$(PackageVersion)</PackageVersionForWorkloadManifests>
```
eerhardt, line 11: Same for this - should be the default and not needed in the .csproj.

tests/Aspire.Workload.Tests/data/nuget8.config
radical: Hm I'll try without that for the tests. I found it useful when dogfooding the sdk+workload.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            var res = (await cmd.ExecuteAsync($"test -c {config}"))
+                                .EnsureSuccessful();
+
+            Assert.Matches("Passed! * - Failed: *0, Passed: *1, Skipped: *0, Total: *1", res.Output);
+            return res;
+        }
+    }
+
+    internal static async Task AssertStarterTemplateRunAsync(IBrowserContext context, AspireProject project, string config, ITestOutputHelper _testOutput)
```
eerhardt, line 208: This is a bit odd to have in the WorkloadTestBase class. It seems like it is specific to certain tests.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: I could move these two to `tests/Directory.Build.props`.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+        // fluent-toolbar/h1 resources
+
+        testOutput.WriteLine($"Waiting for resources to appear on the dashboard");
+        await Task.Delay(500);
+
+        Dictionary<string, ResourceRow> expectedRowsTable = expectedResources.ToDictionary(r => r.Name);
+        HashSet<string> foundNames = [];
+        List<ResourceRow> foundRows = [];
+
+        var timeout = TimeSpan.FromSeconds(timeoutSecs);
```
eerhardt, line 79: Can we use a CancellationToken here?

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+                var probePath = "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome";
+                if (File.Exists(probePath))
+                {
+                    options.ExecutablePath = probePath;
+                }
+            }
+            return await playwright.Chromium.LaunchAsync(options).ConfigureAwait(false);
+        });
+
+        // default timeout for LaunchAsync is 30secs, so pick something larger than that
```
eerhardt, line 52: comment seems off from the code

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+
+                int matchingEndpoints = 0;
+                var expectedEndpoints = expectedRow.Endpoints;
+
+                string[] endpointsFound =
+                    (await rowLoc.Locator("//div[@class='fluent-overflow-item']").AllAsync())
+                        .Select(async e => await e.InnerTextAsync())
+                        .Select(t => t.Result.Trim(','))
+                        .ToArray();
+                // FIXME: this could still return "+2"
```
eerhardt, line 137: What's the plan to address this?

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+        string url = resourceRows.First(r => r.Name == "webfrontend").Endpoints[0];
+        await StarterTemplateRunTestsBase<StarterTemplateFixture>.CheckWebFrontendWorksAsync(context, url, _testOutput, project.LogPath);
+        await project.StopAppHostAsync();
+    }
+
+    private static readonly char[] s_charsToReplace = ['.', '-', '+'];
+
+    public static string FixupSymbolName(string name)
+    {
+        UTF8Encoding utf8 = new();
```
eerhardt, line 238: We shouldn't need to encode the name to UTF8 just to do this fix up / normalizing.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -0,0 +1,263 @@
...
+        {
+            Assert.Fail($"Expected rows not found: {string.Join(", ", expectedRowsTable.Keys.Except(foundNames))}");
+        }
+
+        return foundRows.ToArray();
+    }
+
+    // Don't fixup the prefix so it can have characters meant for testing, like spaces
+    public static string GetNewProjectId(string? prefix = null)
+        => (prefix is null ? "" : $"{prefix}_") + FixupSymbolName(Path.GetRandomFileName());
```
eerhardt, line 183: GetRandomFileName returns an 8.3 file name:

[https://github.com/dotnet/runtime/blob/0b10f4ba0a9174937829ec529c197dd42a569e36/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L275-L279](https://github.com/dotnet/runtime/blob/0b10f4ba0a9174937829ec529c197dd42a569e36/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L275-L279)

We should only need to replace the `.` (but even then - why? `.` are fine in project names)

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
eerhardt: Why not to Workloads.Shared.targets? Isn't that what uses them?

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
```diff
@@ -0,0 +1,143 @@
...
+
+    public static List<ResourceRow> GetExpectedResources(AspireProject project, bool hasRedisCache)
+    {
+        List<ResourceRow> expectedResources = new()
+        {
+            new ResourceRow(
+                Type: "Project",
+                Name: "apiservice",
+                State: "Running",
+                Source: $"{project!.Id}.ApiService.csproj",
```
eerhardt, line 119: Why the `!`?

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: Sure. It just shows the dependence between the two properties. In EndToEnd.Tests `TestUsingWorkloads` is `false` on CI/windows.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
radical: Updated the comment. Please lemme know if it still seems off.

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
radical: good catch!

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
radical: This is originally from https://github.com/dotnet/runtime/commit/26990158d9725aa55d0af81f050a7a8e90c35f58 . But I changed this up to not fixup the whole name, but didn't get to deciding whether it could be dropped completely.

I think you are correct, and we can drop the fixup as our template generators are now to handle anything odd.

radical: I tested this with some more tests I have for project names, and they passed with the fixup removed.

tests/Aspire.Workload.Tests/data/nuget8.config
radical:
```
ServiceDefaults.csproj : error NU1101:

Unable to find package OpenTelemetry.Exporter.OpenTelemetryProtocol. No packages exist with this id in source(s)
Unable to find package OpenTelemetry.Extensions.Hosting. No packages exist with this id in source(s)
Unable to find package OpenTelemetry.Instrumentation.AspNetCore. No packages exist with this id in source(s)
Unable to find package OpenTelemetry.Instrumentation.Http. No packages exist with this id in source(s)
Unable to find package OpenTelemetry.Instrumentation.Runtime. No packages exist with this id in source(s)

Unable to find package Microsoft.Extensions.Http with version (>= 8.0.0)
Unable to find package Microsoft.Extensions.Configuration.Abstractions with version (>= 8.0.0)
Unable to find package Microsoft.Extensions.DependencyInjection.Abstractions with version (>= 8.0.1)
Unable to find package Microsoft.Extensions.Logging.Abstractions with version (>= 8.0.1)
Unable to find package Microsoft.Extensions.Options with version (>= 8.0.2)
Unable to find package Microsoft.Extensions.Primitives with version (>= 8.0.0)
```

radical:
```
Unable to find package Microsoft.Extensions.Http with version (>= 8.0.0)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
Unable to find package Microsoft.Extensions.Configuration.Abstractions with version (>= 8.0.0)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
Unable to find package Microsoft.Extensions.DependencyInjection.Abstractions with version (>= 8.0.1)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
Unable to find package Microsoft.Extensions.Logging.Abstractions with version (>= 8.0.1)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
Unable to find package Microsoft.Extensions.Options with version (>= 8.0.2)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
Unable to find package Microsoft.Extensions.Primitives with version (>= 8.0.0)
  - Found 1459 version(s) in dotnet8 [ Nearest version: 8.0.0-rtm.23523.3 ]
  - Found 0 version(s) in /Users/ankj/dev/aspire/artifacts/bin/dotnet-latest/library-packs
```

tests/PlaywrightWrapper/Program.cs
radical: Looks like creating a simple runtimeconfig.json, and deploying it next to the library is enough to run that directly:

```json
{
  "runtimeOptions": {
    "tfm": "net8.0",
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "8.0.0"
    },
    "configProperties": {
      "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization": false
    }
  }
}
```

radical: Dropped the project, and directly running the assembly now. It works locally.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
radical: um I don't have any immediate plan on fixing this. I haven't hit this yet, but it is a possibility. And the comment might help somebody 🤷

dogfood.sh
radical: updated.

eng/pipelines/templates/BuildAndTest.yml
radical: https://github.com/dotnet/aspire/issues/4222

dogfood.sh
radical: https://github.com/dotnet/aspire/issues/4222

tests/Aspire.Workload.Tests/data/nuget8.config
eerhardt: Why don't you have a `<add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />` entry?

tests/Aspire.Workload.Tests/data/nuget8.config
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <!-- Don't use any higher level config files. -->
+  <fallbackPackageFolders>
+    <clear />
+  </fallbackPackageFolders>
+  <packageSources>
+    <clear />
+    <!-- TEST_RESTORE_SOURCES_INSERTION_LINE -->
```
eerhardt, line 9: Is this necessary? What's it for?

tests/Aspire.Workload.Tests/data/nuget8.config
```diff
@@ -0,0 +1,17 @@
...
+    <add key="built-local" value="%BUILT_NUGETS_PATH%" />
+    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
+    <add key="nuget.org"  value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+  </packageSources>
+    <disabledPackageSources>
+    <clear />
+  </disabledPackageSources>
```
eerhardt, line 16:
```suggestion
  <disabledPackageSources>
    <clear />
  </disabledPackageSources>
```

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,56 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+
+    <RunTestsOnHelix>true</RunTestsOnHelix>
+    <TestUsingWorkloads>true</TestUsingWorkloads>
+    <InstallWorkloadForTesting>true</InstallWorkloadForTesting>
```
eerhardt, line 8:
```suggestion
```

I don't think this is needed, since it defaults to `TestUsingWorkloads`

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,56 @@
...
+    <PackageReference Include="Microsoft.DotNet.XUnitExtensions" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" />
+  </ItemGroup>
+
+  <!-- Used for running one helix job per test class -->
+  <Target Name="_ExtractTestClassNames"
+          Condition="'$(ExtractTestClassNamesForHelix)' == 'true'"
+          BeforeTargets="ZipTestArchive">
+
+    <Exec Command="&quot;$(DotNetTool)&quot; test --no-build -c $(Configuration) -s $(RunSettingsFilePath) -t --nologo -v:q -p:VsTestUseMSBuildOutput=false" ConsoleToMSBuild="true">
```
eerhardt, line 38:
```suggestion
    <Exec Command="&quot;$(DotNetTool)&quot; test --no-build -c $(Configuration) -s $(RunSettingsFilePath) --list-tests --nologo -v:q -p:VsTestUseMSBuildOutput=false" ConsoleToMSBuild="true">
```

Makes it a little more understandable/readable.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -0,0 +1,56 @@
...
+      <Output TaskParameter="ConsoleOutput" ItemName="_ListOfTestsLines" />
+    </Exec>
+
+    <PropertyGroup>
+      <_Regex>^\s*(Aspire.Workload.Tests[^\($]+)</_Regex>
```
eerhardt, line 43: What happens if a new test gets added that doesn't follow this naming? Does the test just get skipped and never executed?

tests/Shared/WorkloadTesting/TestUtils.cs
```diff
@@ -0,0 +1,55 @@
+
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using Microsoft.Playwright;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Aspire.Workload.Tests;
+
+public static class TestExtensions
```
eerhardt, line 10: Can the file name match the class name?

tests/Shared/WorkloadTesting/RequiresDockerDiscoverer.cs
```diff
@@ -0,0 +1,20 @@
...
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Aspire.Workload.Tests;
+
+public class RequiresDockerDiscoverer : ITraitDiscoverer
+{
+    public IEnumerable<KeyValuePair<string, string>> GetTraits(IAttributeInfo traitAttribute)
+    {
+        if (!RequiresDockerTheoryAttribute.IsSupported)
```
eerhardt, line 15: (not necessarily for this PR) - it would be nice if we had a consolidated "RequiresDocker" story across all our tests. This feels pretty spread out - some in `Aspire.Components.Common.Tests`, some here, some in `Aspire.Hosting.Tests`.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: It doesn't default to `TestUsingWorkloads`.

The workload testing targets install two sdks - `dotnet-latest`, and `dotnet-none`. The former has the manifest+workload installed, but that latter has only the manifest installed. It is useful for running tests that want to check the case the user does not have the workload installed.

The `TestUsingWorkloads` property essentially controls that. I'll open a PR in `dotnet/runtime` to rename this property so its purpose is clearer. And possibly clean this up a bit.

tests/Aspire.Workload.Tests/data/nuget8.config
radical: This was used for inserting a source for the locally built nugets. But I changed the way we do that, so this line can be dropped.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: Yeah :/ I'll open an issue to make this more robust.

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: Tracking in https://github.com/dotnet/aspire/issues/4222 .

tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
radical: Tracking in https://github.com/dotnet/aspire/issues/4222 .

tests/Shared/WorkloadTesting/RequiresDockerDiscoverer.cs
radical: Tracking in https://github.com/dotnet/aspire/issues/4222 .

tests/Aspire.Workload.Tests/data/nuget8.config
radical: Dropped.

tests/Aspire.Workload.Tests/data/nuget8.config
radical: Fixed.

eerhardt (approved):
Nice work here @radical! Let's get this in and we can iterate on it as we go.

Issue #4145, opened 2024-05-10
[WebToolsE2E][Aspire] With SDK 8.0.300 + Aspire 8.0 GA, dotnet new aspire-starter does not work due to manifest provider returned a duplicate manifest ID 'microsoft.net.sdk.android'.


v-elenafeng:
**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2 ENU

1. Install SDK 8.0.300
    - Apply feed for SDK 8.0.300
    
1. Install Aspire 8.0 GA

   - Apply the "darc-pub-dotnet-aspire" feed for Aspire 8.0 GA
   
1. Install cred helper (run from Windows PowerShell admin prompt) 

    `iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) }" `

**REPRO STEPS**

1. Run `dotnet --info`  -- works fine
 
    ![image](https://github.com/dotnet/aspire/assets/78580199/fd0c4901-f1c9-4b0a-8da8-a3294cf8f9b4)

1. Run `dotnet workload list`

    ![image](https://github.com/dotnet/aspire/assets/78580199/b807e2f7-04e9-487b-9f3d-d50bc4386fee)

1. Run `dotnet new aspire-starter -o Aspire-StarterApp01`

 
**ACTUAL**
Template creation failed

![image](https://github.com/dotnet/aspire/assets/78580199/f1ea57b1-a435-47e9-a385-61f18a0e6f11)

**EXPECTED**
Template creation should succeed

v-elenafeng:
This issue can also be reproduced with a clean install of VS 17.10 GA (with Aspire 8.0 GA)+ Web workload when creating a new template using the CLI, but works fine using VS.

balachir:
@joperezr are we missing an additional feed? It seems that 'dotnet new aspire-starter' doesn't work using CLI with above steps.

balachir:
The error seems similar to an issue that was reported earlier as well - internal Bug 2048871.

balachir:
Elena and I investigated today and it seems like a possible feed permissions issue. But still, we have two machines that Elena and I setup - she can still reproduce the issue on her machine and I cannot reproduce the issue on my machine. We'll continue doing more investigation on Monday.

v-elenafeng:
I installed a new machine today and can still reproduce this issue, it happens randomly and the failure rate is about 50%.

[...]

PR #4176, opened 2024-05-14
Removing constraint from templates to avoid SDK multithreading issue

joperezr:
Works around https://github.com/dotnet/aspire/issues/4145.

@dsplaisted has done a full diagnosis of the issue [here](https://github.com/dotnet/templating/issues/7946#issuecomment-2108853692), but the TL;DR is that there seems to be a multi-threading issue inside the SDK, particularly in their logic that tries to check for installed workloads. Aspire templates today have a constraint to show these templates only when the aspire workload is installed, which is technically unrequired as the templates get installed via the workload anyway. By adding this constraint, we are forcing the piece of code in the SDK with the race condition, which will make it so that around 50% of the time, aspire template creation will fail. The SDK team will fix this bug in servicing, but in the meantime we are working around the bug by removing the unnecessary constraint from our templates.

We expect a lot of people will hit this issue (essentially anyone creating aspire projects from the command line), so we'll want to service this to 8.0.1 release. (FYI: @danmoseley )

cc: @DamianEdwards @phenning 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4176)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-apphost/.template.config/template.json
+++ b/src/Aspire.ProjectTemplates/templates/aspire-apphost/.template.config/template.json
@@ -23,12 +23,6 @@
   "guids": [
     "98048C9C-BF28-46BA-A98E-63767EE5E3A8"
   ],
-  "constraints": {
-    "aspire": {
-      "type": "workload",
-      "args": [ "aspire" ]
-    }
-  },
   "symbols": {
     "Framework": {
       "type": "parameter",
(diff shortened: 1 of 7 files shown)


phenning (approved):
Looks good. As long as we aren't expecting customers to install templates directly from dotnet new, which I don't believe is the case. Worse case there is that the created template may not function as intended.

Forgind (approved):

DamianEdwards (approved):

Issue #3735, opened 2024-04-16
Support replacing the container host within ReferenceExpression


davidfowl:
Today we automatically replace localhost with container host in a [couple of cases](https://github.com/dotnet/aspire/blob/09b783a418f2c778105f07dc927eeed2ee5c3199/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L1299-L1302), but we need to manually evaluate reference expressions in order to properly evaluate and replace the container host. Each of these sub expressions might be an EndpointReference or ConnectionStringReference.

davidfowl:
Reverted and re-opening this issue.

PR #4179, opened 2024-05-15
Adds Milvus to the Aspire hosting/component packages

timheuer:
Adds the Milvus vector database to Aspire. 

- [x] Hosting component using container standalone
- [x] Client component using C# Milvus.Client
- [x] Hosting/Component initial tests
- [x] Added playground app (use api.http to run the steps) - uses code samples from C# SDK
- [x] Adds the Attu web admin experience (excluded from publish manifest)

Requires that a parameter exists for the root auth api key
No healthchecks/tracing/metrics available in the C# Milvus.Client

https://github.com/dotnet/aspire/assets/4821/b763b107-b216-4202-965d-5d9dc8997b3f

/cc @roji @luisquintanilla 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4179)


src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/var/lib/milvus", isReadOnly);
+
+    /// <summary>
+    /// Adds a bind mount for the configuration of a Milvus container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="source">The source directory on the host to mount into the container.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<MilvusServerResource> WithConfiguration(this IResourceBuilder<MilvusServerResource> builder, string source, bool isReadOnly = false)
```
davidfowl, line 94:
```suggestion
    public static IResourceBuilder<MilvusServerResource> WithConfiguration(this IResourceBuilder<MilvusServerResource> builder, string configurationFilePath)
```

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            .WithHttpEndpoint(port: grpcPort, targetPort: MilvusPortGrpc, name: MilvusServerResource.PrimaryEndpointName)
+            .WithEndpoint(MilvusServerResource.PrimaryEndpointName, endpoint =>
+            {
+                endpoint.Transport = "http2";
+            })
+            .WithEnvironment("COMMON_STORAGETYPE", "local")
+            .WithEnvironment("ETCD_USE_EMBED", "true")
+            .WithEnvironment("ETCD_DATA_DIR", "/var/lib/milvus/etcd")
+            .WithEnvironment("COMMON_SECURITY_AUTHORIZATIONENABLED", "true")
+            .WithArgs("milvus","run","standalone");
```
davidfowl, line 50:
```suggestion
            .WithArgs("milvus", "run", "standalone");
```

Aspire.sln
```diff
@@ -451,26 +451,34 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OrleansContracts", "playgro
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.ServiceDiscovery.Yarp.Tests", "tests\Microsoft.Extensions.ServiceDiscovery.Yarp.Tests\Microsoft.Extensions.ServiceDiscovery.Yarp.Tests.csproj", "{B5D637F3-7E44-4C7D-AE4B-D56D400671CD}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "kafka", "kafka", "{920BB263-E68F-4FA2-93FC-2E385EEA405B}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "KafkaBasic.AppHost", "playground\kafka\KafkaBasic.AppHost\KafkaBasic.AppHost.csproj", "{A39389A0-E780-4B97-808B-DC95CF59B35C}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "KafkaBasic.AppHost", "playground\kafka\KafkaBasic.AppHost\KafkaBasic.AppHost.csproj", "{A39389A0-E780-4B97-808B-DC95CF59B35C}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Consumer", "playground\kafka\Consumer\Consumer.csproj", "{7AA4C56C-3BB2-4FF0-BB03-F3F0D6A4FDAB}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Consumer", "playground\kafka\Consumer\Consumer.csproj", "{7AA4C56C-3BB2-4FF0-BB03-F3F0D6A4FDAB}"
```
mitchdenny, line 460: Any idea why these keep getting updated? Seen it in other PRs too. Result of a merge or something?

NuGet.config
```diff
@@ -19,6 +19,7 @@
     <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="orleans-nightly" value="https://orleans.pkgs.visualstudio.com/orleans-public/_packaging/orleans-nightly/nuget/v3/index.json" />
     <add key="dotnet9-transport" value="https://dnceng.pkgs.visualstudio.com/public/_packaging/dotnet9-transport/nuget/v3/index.json" />
+    <add key="dotnet8-early" value="https://pkgs.dev.azure.com/dnceng/internal/_packaging/dotnet8-early-access/nuget/v3/index.json" />
```
mitchdenny, line 22: Remove I think.

playground/milvus/MilvusPlayground.ApiService/Properties/launchSettings.json
```diff
@@ -0,0 +1,25 @@
...
+      "commandName": "Project",
+      "dotnetRunMessages": true,
+      "launchBrowser": true,
+      "launchUrl": "",
+      "applicationUrl": "http://localhost:5456",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    },
+    "https": {
```
mitchdenny, line 14: add `generate-manifest` profile and make sure you check-in the manifest.

playground/milvus/MilvusPlayground.AppHost/Properties/launchSettings.json
```diff
@@ -0,0 +1,29 @@
...
+      "launchBrowser": true,
+      "applicationUrl": "https://localhost:17115;http://localhost:15293",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "DOTNET_ENVIRONMENT": "Development",
+        "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "https://localhost:21286",
+        "DOTNET_RESOURCE_SERVICE_ENDPOINT_URL": "https://localhost:22004"
+      }
+    },
+    "http": {
```
mitchdenny, line 16: `generate-manifest` profile here instead of where I said above :)

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        IResourceBuilder<ParameterResource>? apiKey = null,
+        int? grpcPort = null)
+    {
+        var tokenParameter = apiKey?.Resource ??
+            ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-Key", special: false);
+        var milvus = new MilvusServerResource(name, tokenParameter);
+        return builder.AddResource(milvus)
+            .WithImage(MilvusContainerImageTags.Image, MilvusContainerImageTags.Tag)
+            .WithImageRegistry(MilvusContainerImageTags.Registry)
+            .WithHttpEndpoint(port: grpcPort, targetPort: MilvusPortGrpc, name: MilvusServerResource.PrimaryEndpointName)
```
mitchdenny, line 41: Why not `WithEndpoint("name", callback, create: true)`? Are you just trying to get the http defaults and doing a minor change?

davidfowl: we dont let you set the transport in these calls. Missed it...

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Milvus;
+
+internal static class MilvusContainerImageTags
+{
+    public const string Registry = "docker.io";
+    public const string Image = "milvusdb/milvus";
+    public const string Tag = "v2.3.15";
```
mitchdenny, line 10: Based on our policy here you can probably use `2.3-latest` here.

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// <remarks>
+    /// This version the package defaults to the v2.4.0 tag of the milvusdb/milvus container image.
+    /// The .NET client library uses the gRPC port by default to communicate and this resource exposes that endpoint.
+    /// </remarks>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency</param>
+    /// <param name="apiKey">The parameter used to provide the auth key/token user for the Milvus resource.</param>
+    /// <param name="grpcPort">The host port of gRPC endpoint of Milvus database.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{MilvusServerResource}"/>.</returns>
+    public static IResourceBuilder<MilvusServerResource> AddMilvus(this IDistributedApplicationBuilder builder,
```
mitchdenny, line 30: Would be good to include `<example />` elements where. I'd probably have one example that shows with and without explicit auth keys. You might also `<see />` for `WithAttu` to and add an example of its wire-up there to make it more discoverable.

And yes we have to do this everywhere but trying to stop the bleeding :)

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
mitchdenny, line 94: Should this be `WithConfigurationBindMount`?

mitchdenny:
This is looking pretty good. Mostly just some minor things and questions. @eerhardt for a once over for the component side of things. I gave it a once over but he is more attuned to that side of things.

mitchdenny:
One thing that I'm thinking about is allowing unauthenticated access to Attu?

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        string name,
+        IResourceBuilder<ParameterResource>? apiKey = null,
+        int? grpcPort = null)
+    {
+        var tokenParameter = apiKey?.Resource ??
+            ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-Key", special: false);
+        var milvus = new MilvusServerResource(name, tokenParameter);
+        return builder.AddResource(milvus)
+            .WithImage(MilvusContainerImageTags.Image, MilvusContainerImageTags.Tag)
```
roji, line 39: Should it be possible for the user to set the Milvus version? As things are, it seems like it's hard-coded to a specific version (unless I'm missing something).

roji (commented):
Super nice to see this!

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
davidfowl: They can use WithImageTag to override it

Aspire.sln
timheuer: I've got a note in to the project team -- I noticed this a while back...I can't explain either :-(

NuGet.config
timheuer: would remove in final PR yes...needed just during WIP

timheuer:
> One thing that I'm thinking about is allowing unauthenticated access to Attu?

Attu requires authenticated access to the Milvus instance -- much like things like PgAdmin have a login prompt as well.  Is that what you're referring to? Attu starts with a login requirement to an instance of Milvus, it isn't open-ended.

playground/milvus/MilvusPlayground.ApiService/Properties/launchSettings.json
eerhardt: Can we automate this somehow?

Do we really need a `generate-manifest` profile? Why can't we just have a script that `dotnet run -- ...` all the playground projects with the correct args? Then we don't need to duplicate the args in every playground app.

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
eerhardt: Looks like there is a `2.4-latest` now. https://hub.docker.com/r/milvusdb/milvus/tags

Directory.Packages.props
```diff
@@ -101,6 +101,7 @@
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.2.0" />
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.7.2" />
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.7.2" />
+    <PackageVersion Include="Milvus.Client" Version="2.3.0-preview.1"/>
```
eerhardt, line 104: I see there isn't a stable version of this yet. Do we know when it plans on going stable?

For now, we will need to add 

https://github.com/dotnet/aspire/blob/756dfe663d0759aa0d2a0e1208502a403a398c32/src/Components/Aspire.Azure.AI.OpenAI/Aspire.Azure.AI.OpenAI.csproj#L6-L7

to the component .csproj since we can't ship stable until all dependencies are stable.

NuGet.config
eerhardt: What's it needed for? The 2 new packages are on nuget.org. We can just get the mirrored internally. https://github.com/dotnet/arcade/blob/main/Documentation/MirroringPackages.md

playground/milvus/MilvusPlayground.ApiService/Program.cs
```diff
@@ -0,0 +1,131 @@
...
+                FieldSchema.Create<long>("word_count"),
+                FieldSchema.CreateVarchar("book_name", 256),
+                FieldSchema.CreateFloatVector("book_intro", 2)
+                }
+            );
+    logger.LogInformation("Collection created: book");
+    return Results.Ok("Collection created");
+});
+
+app.MapGet("/insert", async (MilvusClient milvusClient, ILogger<Program> logger) =>
```
eerhardt, line 47: Would it make sense to combine this with `/create`? That way I only need to hit 1 endpoint to set the thing up, and then `/search` to ensure it is working. Like how you did in

https://github.com/dotnet/aspire/blob/756dfe663d0759aa0d2a0e1208502a403a398c32/playground/Qdrant/Qdrant.ApiService/Program.cs#L19

playground/milvus/MilvusPlayground.AppHost/Program.cs
```diff
@@ -0,0 +1,12 @@
+var builder = DistributedApplication.CreateBuilder(args);
+
+var token = builder.AddParameter("milvusauth", true);
```
eerhardt, line 3: Is this necessary? Can we just generate an auth token instead?

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
```diff
@@ -0,0 +1,15 @@
...
+
+    public const string AttuImage = "zilliz/attu";
+    public const string AttuTag = "v2.3.10";
```
eerhardt, line 13: Can we ask the owner of this container to start making `vX.Y` versioned tags?

src/Aspire.Hosting.Milvus/MilvusServerResource.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="apiKey">A <see cref="ParameterResource"/> that contains the authentication apiKey/token</param>
+    public MilvusServerResource(string name, ParameterResource apiKey) : base(name)
+    {
+        ArgumentNullException.ThrowIfNull(apiKey);
+        ApiKeyParameter = apiKey;
+    }
+
+    internal const string PrimaryEndpointName = "grpc";
```
eerhardt, line 24: (nit) for consistency, it is customary to put these const strings at the top of the class.

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// <summary>
+    /// Adds a named volume for the data folder to a Milvus container resource.
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the resource name.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<MilvusServerResource> WithDataVolume(this IResourceBuilder<MilvusServerResource> builder, string? name = null, bool isReadOnly = false)
```
eerhardt, line 84: Why don't we have a `WithDataBindMount` as well?

src/Components/Aspire.Milvus.Client/Aspire.Milvus.Client.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <PackageTags>$(ComponentCommonPackageTags) milvus</PackageTags>
```
eerhardt, line 6: Do we also want `database vector search` here? That's what https://www.nuget.org/packages/Qdrant.Client has. 

Note https://www.nuget.org/packages/Milvus.Client just has `milvus vector database`.

src/Components/Aspire.Milvus.Client/README.md
```diff
@@ -0,0 +1,105 @@
...
+```csharp
+builder.AddMilvusClient("milvus");
+```
+
+And then the connection string will be retrieved from the `ConnectionStrings` configuration section:
+
+```json
+{
+  "ConnectionStrings": {
+    "milvus": "Endpoint=http://localhost:19530;Key=root:123456!@#$%"
```
eerhardt, line 44: Why does the `Key` have `root:` in it? But it doesn't below in the json example. And then it does again in the C# example.

src/Components/Aspire.Milvus.Client/ConfigurationSchema.json
```diff
@@ -0,0 +1,39 @@
+{
```
eerhardt, line 1: Are you missing the assembly level attribute for `ConfigurationSchema` in a AssemblyInfo.cs file?

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
timheuer: Only reason for now I've kept it is the client versioning in the docs mentions staying in sync with the server versioning.  e.g., the client is 2.3.x targets the 2.3.x version of the server.  It says it is compatible, but that's the only reason so far.  IMO the Milvus tagging is all over the place.

NuGet.config
timheuer: Hmm, at least for me locally I was failing on some 8.0.5 packages and it wasn't finding the 8.0.0 Aspire ones

playground/milvus/MilvusPlayground.AppHost/Program.cs
timheuer: No. there is only one user in a default Milvus container (root:Milvus). This cannot be configured presently via environment variables or anything and can only be configured after the fact (or additional users after the fact). See: https://github.com/milvus-io/milvus/issues/33058 and https://github.com/milvus-io/milvus/discussions/33032#discussioncomment-9427837

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
timheuer: https://github.com/zilliztech/attu/issues/502

src/Components/Aspire.Milvus.Client/README.md
timheuer: In Milvus terms the "apiKey" is a user:pass name (root:Milvus being default)...rather than a traditional apiKey aspect (which may exist for some hosted/managed instances of Milvus).  Can make this consistent for sure, but also trying not to have our readme document the milvus default -- it _can_ be configured after the fact so someone could have a separate u/p if they have a script that does the update.

tests/Aspire.Milvus.Client.Tests/AspireMilvusHelpers.cs
```diff
@@ -0,0 +1,38 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.DotNet.XUnitExtensions;
+using Milvus.Client;
+
+namespace Aspire.Milvus.Client.Tests;
+public class AspireMilvusHelpers
+{
+    public const string TestingEndpoint = "http://localhost:19530";
```
eerhardt, line 10: Let's not hard code this into the tests. Instead, get the `MilvusContainer` (test container) instance and ask it for the connection string.

tests/Aspire.Milvus.Client.Tests/MilvusContainerFixture.cs
```diff
@@ -0,0 +1,37 @@
...
+    public string GetConnectionString() => Container?.GetEndpoint().AbsoluteUri ??
+        throw new InvalidOperationException("The test container was not initialized.");
+
+    public async Task InitializeAsync()
+    {
+        if (RequiresDockerTheoryAttribute.IsSupported)
+        {
+            Container = new MilvusBuilder()
+                .WithImage($"{MilvusContainerImageTags.Image}:{MilvusContainerImageTags.Tag}")
+                .WithPortBinding(19530)
```
eerhardt, line 24:
```suggestion
```

Why do we need this? We should just let the default happen. That way we can do 2 at once, if necessary.

tests/Aspire.Milvus.Client.Tests/ConformanceTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        => configuration.AddInMemoryCollection(new KeyValuePair<string, string?>[2]
+        {
+            new KeyValuePair<string, string?>(CreateConfigKey("Aspire:Milvus:Client", key, "Endpoint"), "http://localhost:19530"),
+            new KeyValuePair<string, string?>($"ConnectionStrings:{key}","Endpoint=http://localhost:19530;Key=pass")
+        });
+
+    protected override void TriggerActivity(MilvusClient service)
+    {
+        service.GetVersionAsync().Wait();
+    }
```
eerhardt, line 50: Do we have any tracing with Milvus?

What is the tracing, metrics, and health checks plan for this component?

src/Components/Aspire.Milvus.Client/Aspire.Milvus.Client.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: Can you also update:

* https://github.com/dotnet/aspire/blob/main/src/Components/Aspire_Components_Progress.md
* https://github.com/dotnet/aspire/blob/main/src/Components/Telemetry.md

src/Components/Aspire.Milvus.Client/AspireMilvusExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+        else
+        {
+            builder.Services.AddKeyedSingleton(serviceKey, (sp, key) => ConfigureMilvus(sp));
+        }
+
+        MilvusClient ConfigureMilvus(IServiceProvider serviceProvider)
+        {
+            if (settings.Endpoint is not null)
+            {
+                return new MilvusClient(settings.Endpoint, apiKey: settings.Key!, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());
```
eerhardt, line 85: Should we also be checking for the Key being null, and throwing a meaningful exception if it is null?

playground/milvus/MilvusPlayground.AppHost/Properties/launchSettings.json
timheuer: Will add this, but curious what's the point for doing so in the playground source here? Convenient launching point to generate the manifest?

src/Aspire.Hosting.Milvus/MilvusContainerImageTags.cs
timheuer: Changing to 2.3-latest

playground/milvus/MilvusPlayground.AppHost/Program.cs
eerhardt: How's this code going to work then?

https://github.com/dotnet/aspire/blob/0a3cbecc3a8754a8b45e3ab6d5548854e6779d05/src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs#L35-L37

NuGet.config
eerhardt: > was failing on some 8.0.5 packages 

that must have been a point in time thing as 8.0.5 was just released yesterday.

src/Components/Aspire.Milvus.Client/Aspire.Milvus.Client.csproj
timheuer: We didn't add those tags to the Aspire components, could add them no problem

tests/Aspire.Milvus.Client.Tests/AspireMilvusHelpers.cs
timheuer: File no longer needed now that i added testcontainers...will delete.

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
timheuer: Yes @mitchdenny exactly that -- else I'd be setting all the defaults in WithEndpoint -- this felt better

tests/Aspire.Milvus.Client.Tests/ConformanceTests.cs
timheuer: I was planning on logging issues on the component to add some of these. @roji do you know if these may already be in the works?

playground/milvus/MilvusPlayground.AppHost/Program.cs
timheuer: Me thinking aloud why I kept this in...

It *is* possible to reset the root token.  But right now it has to be done after the fact. So in theory someone could do a data volume, have a script (or some other client operation) that changes the root login (or adds another user), and then this would need to be changed (they would populate the parameter.  Correct here that the CreateDefaultPasswordParameter wouldn't really work in this instance.  I was also trying to avoid hard-coding the default root creds here.

To me the ideal would look like `apiKey?.Resource ?? DefaultMilvusApiKeyParameter` - per discussion in teams channels, there is no easy way to create a ParameterResource with a string value default.

Thoughts welcome here...

eerhardt:
Can you also add some end-to-end tests?

https://github.com/dotnet/aspire/blob/5f2af96cba6ee7764b36e38d7340e530bdb5cd66/tests/Aspire.EndToEnd.Tests/IntegrationServicesTests.cs#L26-L35

See https://github.com/dotnet/aspire/pull/3839 for an example.

tests/Aspire.Milvus.Client.Tests/ConformanceTests.cs
roji: The tracking issue on the Milvus SDK side is https://github.com/milvus-io/milvus-sdk-csharp/issues/85.

I'm planning on adding tracing support there, but it may take me a bit of time before I get around to it... Re metrics, I **think** it's more appropriate to get them from Milvus itself (the server), which emits them - see the issue above for some discussion on that. We may want to have a more general discussion on the client vs. server preference for metrics.

src/Components/Aspire.Milvus.Client/AspireMilvusExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+                timeout: default));
+        }
+
+        MilvusClient ConfigureMilvus(IServiceProvider serviceProvider)
+        {
+            if (settings.Endpoint is not null && settings.Key is not null)
+            {
+                return settings.Database is not null
+                    ? new MilvusClient(settings.Endpoint, apiKey: settings.Key!, database: settings.Database, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>())
+                    : new MilvusClient(settings.Endpoint, apiKey: settings.Key!, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());
```
eerhardt, line 101:
```suggestion
                return new MilvusClient(settings.Endpoint, apiKey: settings.Key, database: settings.Database, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());
```

From reading the code, I don't think it throws if you pass in `null` for `database`.

Also, you verified `settings.Key is not null` above, you shouldn't need to `!` it.

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <example>
+    /// Use in application host
+    /// <code>
+    /// var builder = DistributedApplication.CreateBuilder(args);
+    ///
+    /// var booksdb = builder.AddMilvus("milvus");
+    ///   .AddDatabase("booksdb");
+    /// 
+    /// var api = builder.AddProject&lt;Projects.Api&gt;("api")
+    ///                  .WithReference(booksdb);
```
eerhardt, line 79: (nit) - it would be nice to use the same whitespace pattern in an example.

```suggestion
    /// var booksdb = builder.AddMilvus("milvus");
    ///   .AddDatabase("booksdb");
    /// 
    /// var api = builder.AddProject&lt;Projects.Api&gt;("api")
    ///   .WithReference(booksdb);
```

src/Components/Aspire.Milvus.Client/AspireMilvusExtensions.cs
timheuer: It doesn't need to throw, because it would hit the other conditional and NOT create a client with a database arg.  Good catch on the ! (removed)

timheuer: In Miluvs there is an implied default database, so if you don't specify one (which is NOT a requirement) you write to the default space.

eerhardt: What's the functional difference between the code:

```C#
                return settings.Database is not null
                    ? new MilvusClient(settings.Endpoint, apiKey: settings.Key, database: settings.Database, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>())
                    : new MilvusClient(settings.Endpoint, apiKey: settings.Key, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());
```

and 

```C#
                return new MilvusClient(settings.Endpoint, apiKey: settings.Key, database: settings.Database, loggerFactory: serviceProvider.GetRequiredService<ILoggerFactory>());
```

?

Decompiling your current code with ILSpy shows:

image

Why check for `null` at all? Just pass `settings.Database` into the ctor.

timheuer: Ah, I missed the ctor that Milvus.Client handled this already...got it, will change/simplify.

eerhardt (approved):
Thanks for the contribution!

playground/milvus/MilvusPlayground.ApiService/Program.cs
```diff
@@ -0,0 +1,125 @@
...
+    // Check result
+    logger.LogInformation("Insert status: {0},", result.ToString());
+
+    return Results.Ok("Collection created");
+});
+
+app.MapGet("/search", async (MilvusClient milvusClient, ILogger<Program> logger) =>
+{
+    MilvusCollection collection = milvusClient.GetCollection("book");
+    await collection.CreateIndexAsync(
```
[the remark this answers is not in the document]
timheuer: No, I just combined sample code -- this was previously in a diff block before combining...can move to /create so index is created and loaded there.

playground/milvus/MilvusPlayground.ApiService/Program.cs
```diff
@@ -0,0 +1,125 @@
...
+        logger.LogInformation("FieldName:{0}, IndexName:{1}, IndexId:{2}", info.FieldName, info.IndexName, info.IndexId);
+    }
+
+    // Then load it
+    await collection.LoadAsync();
+
+    // Search
+    List<string> search_output_fields = new() { "book_id" };
+    List<List<float>> search_vectors = new() { new() { 0.1f, 0.2f } };
+    SearchResults searchResult = await collection.SearchAsync(
```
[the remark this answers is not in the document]
timheuer: Can simplify this by only using one query call -- just meant to use the API...but will remove search and keep query

src/Components/Aspire.Milvus.Client/Aspire.Milvus.Client.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <PackageTags>$(ComponentCommonPackageTags) milvus database vector search</PackageTags>
+    <Description>A Milvus client that integrates with Aspire, including logging.</Description>
+    <PackageIconFullPath>$(SharedDir)Milvus_256x.png</PackageIconFullPath>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    <NoWarn>$(NoWarn);CS8002</NoWarn> <!-- Milvus.Client packages are not signed -->
```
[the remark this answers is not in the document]
timheuer: Logged: https://github.com/milvus-io/milvus-sdk-csharp/issues/87

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// </summary>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="configurationFilePath">The source directory on the host to mount into the container.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<MilvusServerResource> WithConfigurationBindMount(this IResourceBuilder<MilvusServerResource> builder, string configurationFilePath)
+        => builder.WithBindMount(configurationFilePath, "/milvus/configs/milvus.yaml");
+
+    private static void ConfigureAttuContainer(EnvironmentCallbackContext context, MilvusServerResource resource)
+    {
+        context.EnvironmentVariables.Add("MILVUS_URL", $"{resource.PrimaryEndpoint.Scheme}://{resource.PrimaryEndpoint.ContainerHost}:{resource.PrimaryEndpoint.Port}");
```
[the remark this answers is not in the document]
timheuer: I don't think this is a correct edit @eerhardt.  See: https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.MySql/PhpMyAdminConfigWriterHook.cs#L37 (and Postgres as well) -- for the local dev experience and the admin tools, they are containers needing to access containers.  The proxy ports don't help and need the container host network.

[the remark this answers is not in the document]
eerhardt: @davidfowl @mitchdenny - would it make sense to make a "ContainerHostUrl" property on an Endpoint? That way would could simplify code like this?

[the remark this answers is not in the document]
mitchdenny: I think that we will be able to use the .Url property when this is addressed:

https://github.com/dotnet/aspire/issues/3735

timheuer:
@eerhardt any other feedback to get this over the last mile?

src/Components/Aspire.Milvus.Client/Aspire.Milvus.Client.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions">
+      <HintPath>..\..\..\..\..\AppCacheFiles\nuget\microsoft.extensions.diagnostics.healthchecks.abstractions\8.0.4\lib\net8.0\Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
eerhardt, line 29:
```suggestion
```

eerhardt (approved):
LGTM. Thanks for the contribution, @timheuer!

playground/milvus/MilvusPlayground.AppHost/Program.cs
eerhardt: @timheuer - thinking more about this - how is this going to work in a deployed/production environment? If I use `azd` to deploy an app that has a Milvus resource, will the ApiKey always be `root:Milvus` even in a deployed environment? That doesn't seem secure.

Issue #2957, opened 2024-03-16
Incorrect file permission for /pgadmin4/servers.json using WithPgAdmin


miceiken:
I'm running into an issue on Windows 10 using WSL 2 with Ubuntu 22.04.4 LTS where the pgAdmin container is unable to read the bind mounted tmp file mapped to `/pgadmin4/servers.json`

System information:
```
➜  ~ dotnet --version
8.0.201
➜  ~ dotnet workload --info
 Workload version: 8.0.200-manifests.a7f084b6
 [aspire]
   Installation Source: SDK 8.0.200
   Manifest Version:    8.0.0-preview.4.24156.9/8.0.100
   Manifest Path:       /home/miceiken/.dotnet/sdk-manifests/8.0.100/microsoft.net.sdk.aspire/8.0.0-preview.4.24156.9/WorkloadManifest.json
   Install Type:        FileBased
```

AppHost:
```
var postgres = builder
    .AddPostgres("postgres")
    .WithVolumeMount("volume.postgres", "/var/lib/postgresql/data")
    .WithPgAdmin();
```

Logs from pgAdmin container:
```
2024-03-16 17:33:23 NOTE: Configuring authentication for DESKTOP mode.
2024-03-16 17:33:23 pgAdmin 4 - Application Initialisation
2024-03-16 17:33:23 ======================================
2024-03-16 17:33:23 
2024-03-16 17:33:32 ----------
2024-03-16 17:33:32 Loading servers with:
2024-03-16 17:33:32 User: pgadmin4@pgadmin.org
2024-03-16 17:33:32 SQLite pgAdmin config: /var/lib/pgadmin/pgadmin4.db
2024-03-16 17:33:32 ----------
2024-03-16 17:33:36 Error reading input file /pgadmin4/servers.json: [13] Permission denied
```

Executed in the container shell:
```
/pgadmin4 $ cat servers.json 
cat: can't open 'servers.json': Permission denied
/pgadmin4 $ ls -la servers.json 
-rw-------    1 1000     1000           232 Mar 16 16:33 servers.json
```

On host:
```
$ ls -la /tmp/tmphpJFaH.tmp
-rw------- 1 miceiken miceiken 232 Mar 16 17:33 /tmp/tmphpJFaH.tmp
$ cat /tmp/tmphpJFaH.tmp
[...]

PR #4191, opened 2024-05-15
Set read permissions to allow pgadmin container to read the generated config file

danegsta:
Grants read permissions to the generated pgadmin config file on unix systems such that it can be read by the user that runs in the container which may not match the local user ID.

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4191)

### DIFF
--- a/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
@@ -1,7 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.Json;
 using Aspire.Hosting.ApplicationModel;
@@ -22,7 +21,7 @@ internal sealed class PgAdminConfigWriterHook : IDistributedApplicationLifecycle
         using var stream = new FileStream(serverFileMount.Source!, FileMode.Create);
         using var writer = new Utf8JsonWriter(stream);
         // Need to grant read access to the config file on unix like systems.
-        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        if (!OperatingSystem.IsWindows())
         {
             File.SetUnixFileMode(serverFileMount.Source!, UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.GroupRead | UnixFileMode.OtherRead);
         }
--- a/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Text.Json;
 using Aspire.Hosting.ApplicationModel;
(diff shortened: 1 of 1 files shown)


danegsta:
This is intended to resolve #2957

src/Aspire.Hosting.PostgreSQL/PgAdminConfigWriterHook.cs
```diff
@@ -20,6 +21,11 @@ public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, C
 
         using var stream = new FileStream(serverFileMount.Source!, FileMode.Create);
         using var writer = new Utf8JsonWriter(stream);
+        // Need to grant read access to the config file on unix like systems.
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
davidfowl, line 25: OperationSystem.IsLinux

oskardudycz:
@danegsta @davidfowl I'm not sure how much effort it'd require, but I think that it'd be worth trying to add an integration test that'd either 
- scan the pgAdmin docker instance logs for the error,
- or exec into the docker instance to try to access the file (e.g., with cat).
(so essentially, what I did manually and described in the GH Issue).

Of course, that's not a perfect test, but if it is run on an Ubuntu-based runner, then it could at least verify the basic regression.

oskardudycz:
@danegsta @davidfowl I did a test, and put a breakpoint in the PgAdminConfigWriterHook in the line where new change is located, took the filepath and try to patch it manually as:

```bash
$ chown 644 /tmp/tmpJLoUtQ.tmp
chown: changing ownership of '/tmp/tmpJLoUtQ.tmp': Operation not permitted
$ sudo chown 644 /tmp/tmpJLoUtQ.tmp
```

So firstly it failed without `sudo` (which is potentially bad), but even with sudo, I still got the same error as in the issue: `Error reading input file /pgadmin4/servers.json: [13] Permission denied`

oskardudycz:
@danegsta @davidfowl I did a test, and put a breakpoint in the PgAdminConfigWriterHook in the line where new change is located, took the filepath from serverFileMount.Source and patched it manually as:

$ chmod 644 /tmp/tmpklaU5k.tmp
$ ls -l /tmp/tmpklaU5k.tmp    
-rw-r--r-- 1 oskar oskar 0 May 15 22:10 /tmp/tmpklaU5k.tmp


And succeeded! Almost 😄


The server appears, but I cannot login.
image

```bash
$ cat  /tmp/tmpklaU5k.tmp
{"Servers":{"1":{"Name":"postgres-db","Group":"Servers","Host":"host.docker.internal","Port":41961,"Username":"postgres","SSLMode":"prefer","MaintenanceDB":"postgres","PasswordExecCommand":"echo \u0027z~PekxykhkB}h71C9GdT3J\u0027"}}}% 
```

Something is wrong with networking between containers. When I got the IP of PG Container by

```bash
$ docker inspect  -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' postgres-db-c0qfrto
172.17.0.2
````

And used it as the host name instead of `host.docker.internal` that was provided then it worked well.

**Comment by danegsta** [comment]
@davidfowl it might be worth considering adding a custom default host for Aspire containers `--add-host=host.aspire.internal:host-gateway` and using that domain instead of `host.docker.internal`/`host.podman.internal` domains. It works for both Docker and Podman (with recent Podman versions); the only caveat for Docker is that, if you specify a custom host that way, Docker Desktop won't automatically create a `host.docker.internal` binding unless explicitly added.

**Comment by oskardudycz** [comment]
@danegsta I'm not sure if that helps, but you can specify multiple hosts mapping with `--add-host` by passing this param multiple times, e.g.:

```
docker run --rm=True --add-host=1.example.com:10.0.0.1 --add-host=2.example2.com:10.0.0.2 --add-host=3.example.com:10.0.0.3 ubuntu cat /etc/hosts
```

mitchdenny (approved):

Issue #3688, opened 2024-04-15
Add support for http OTLP endpoint in the dashboard


davidfowl:
Right now the dashboard supports the grpc OTLP endpoint, we could support http as well but it's unclear how much demand or need there is for it. Filing this issue so others can upvote.

JamesNK:
Work involved:

* Add new endpoint for OTLP http. See https://opentelemetry.io/docs/specs/otlp/#otlphttp-default-port. HTTP protocol will be Http1AndHttp2 (default's to HTTP/1.1 when there isn't TLS).
* Protobuf over HTTP. Existing Protobuf data over HTTP instead of gRPC.
* JSON over HTTP. New JSON DTOs plus mapping data.

kvenkatrajan:
@davidfowl / @JamesNK  this is for post GA correct?

davidfowl:
Correct. This can stay in the backlog

JamesNK:
Related: Documentation to make it clear only OTLP/grpc is supported: https://github.com/dotnet/docs-aspire/pull/678

thangchung:
+1 for this feature

ThomasPiskol:
Support for http is required in case you want to use OTLP from the browser. Currently the OTLP SDK doesn't support gRPC from the browser:

![image](https://github.com/dotnet/aspire/assets/38790465/5db62c27-9607-417c-a1ea-bb58544170aa)


https://opentelemetry.io/docs/languages/js/exporters/#usage-in-the-browser

leslierichardson95:
Keeping this in the backlog because it's needed for non-Aspire services.

pksorensen:
Nor AzureMonitorTraceExporter or OTLPTraceExporter works in the browser, so its a bit of sad face when using some hours getting into otel and finally getting things up and running, then to realize there is no real path atm to export from browser frontend :D Would be cool to have this implementet so we can use http protocol, or alternative fix those exports to use grpc :)

martinjt:
For those wanting this, especially those looking at it from a Browser perspective. In production, you'll be using the collector as an entrypoint for browser based telemetry from a security perspective etc.

I built a library that should allow you to do this right now by adding an Otel Collector to aspire.

https://www.nuget.org/packages/PracticalOtel.OtelCollector.Aspire/0.9.3-rc

davidfowl:
@sharpSteff sent a PR!

JamesNK:
[...]

PR #4197, opened 2024-05-16
 Add support for HTTP(S) OTLP endpoint 

sharpSteff:
Hi Aspire-Team,

Great work with the Aspire-Dashboard. 
I'm currently looking for a replacement for current OTLP-Backend (https://github.com/SigNoz/signoz), and Aspire Dashboard seems to be a lightweight sulution.

However we are relying in the HTTP OTLP Receiver (#3688). Therefore i tried my best to add HTTP OTLP endpoint support and it works quite will in our usecase.

With this changes Aspire Dashboard can consume OTLP Grpc and http(s) simultaneously. Due to the http/protobuf I could reuse the existing grpc-otlps models and services

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4197)


src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    public static WebApplication ConfigureHttpOtlp(this WebApplication app, Uri httpEndpoint)
+    {
+        app.MapPost("/v1/metrics",
+            async ([FromServices] OtlpMetricsService service, HttpContext httpContext) =>
+            {
+                var body = await ReadFullyAsync(httpContext).ConfigureAwait(false);
+                var metrics = ExportMetricsServiceRequest.Parser.ParseFrom(body);
+                await service.Export(metrics, null!).ConfigureAwait(false);
+            }
+        ).AllowAnonymous()
```
davidfowl, line 23: Shouldn't this use the otlp auth policy?

https://github.com/dotnet/aspire/blob/03c6a503930f0a37a089a05fa82b49e636b261dd/src/Aspire.Dashboard/Otlp/Grpc/OtlpMetricsService.cs#L14

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        return app;
+    }
+
+    private static async Task<byte[]> ReadFullyAsync(HttpContext context)
+    {
+        using var ms = new MemoryStream();
+        await context.Request.Body.CopyToAsync(ms).ConfigureAwait(false);
+        return ms.ToArray();
+    }
```
davidfowl, line 54: This should be using pipelines to buffer without allocating.

davidfowl:
The formatting changes make it really hard to review. I don't know if you can undo those but with a big change like this it really makes the diff bigger than it needs to be.

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: Good catch!
changed to 
`.RequireAuthorization(OtlpAuthorization.PolicyName)`

sharpSteff:
> The formatting changes make it really hard to review. I don't know if you can undo those but with a big change like this it really makes the diff bigger than it needs to be.

I reverted the format changes. Sorry for that

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: How do I achieve this?
`Google.Protobuf.MessageParser<T>` wants either `ReadOnlySequence<byte> `, `ReadOnlySpan<byte>` or a `Stream`. but the HttpContext.Request.Body-Stream can not be passed directly

sharpSteff: like this?
```
var reader = httpContext.Request.BodyReader;
var readResult = await reader.ReadAsync().ConfigureAwait(false);
var readResultBuffer = readResult.Buffer;
var metrics = ExportMetricsServiceRequest.Parser.ParseFrom(readResultBuffer);
```

sharpSteff: @davidfowl I got rid of the byte[] and stream allocation

eerhardt:
Are there tests that can be written for this new functionality?

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+        var readResult = await reader.ReadAsync().ConfigureAwait(false);
+        var readResultBuffer = readResult.Buffer;
+        if (readResultBuffer.IsEmpty && readResult.IsCompleted)
+        {
+            return;
+        }
+
+        await exporter(readResultBuffer).ConfigureAwait(false);
+        reader.AdvanceTo(readResultBuffer.Start, readResultBuffer.End);
+        await reader.CompleteAsync().ConfigureAwait(false);
```
davidfowl, line 57: I would suggest something like this:

https://github.com/DamianEdwards/MinimalApis.Extensions/blob/f1a6ab59edadd7625d12770adbede0569ed90244/src/MinimalApis.Extensions/Binding/BodyOfT.cs#L150-L171

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -73,10 +74,18 @@ public sealed class OtlpOptions
     public OtlpAuthMode? AuthMode { get; set; }
     public string? EndpointUrl { get; set; }
```
drewnoakes, line 75: Can we rename this to `GrpcEndpointUrl` for consistency?

drewnoakes (commented):
Looks like a good addition, thanks. Left a few comments.

It'd be great to add a playground application that exercised these options, so there was an F5-able test bed, for manual testing.

There are integration tests for the current gRPC-based OTLP interactions. Having similar tests for the HTTP interactions would be valuable to ensure this scenario doesn't break.

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: done

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
sharpSteff: done

sharpSteff:
@eerhardt @drewnoakes 
> Are there tests that can be written for this new functionality?
I added for all endpoints IntegrationTests

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -43,6 +44,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
     private readonly IReadOnlyList<string> _validationFailures;
     private Func<EndpointInfo>? _frontendEndPointAccessor;
     private Func<EndpointInfo>? _otlpServiceEndPointAccessor;
```
JamesNK, line 46: Can you change:

* _otlpServiceEndPointAccessor -> _otlpServiceGrpcEndPointAccessor
* OtlpServiceEndPointAccessor -> OtlpServiceGrpcEndPointAccessor

If there are two OTLP endpoint types then the variables should always say whether it's the grpc or http.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -339,7 +359,12 @@ private void ConfigureKestrelEndpoints(WebApplicationBuilder builder, DashboardO
         {
             // Translate high-level config settings such as DOTNET_DASHBOARD_OTLP_ENDPOINT_URL and ASPNETCORE_URLS
             // to Kestrel's schema for loading endpoints from configuration.
-            AddEndpointConfiguration(initialValues, "Otlp", otlpUri.OriginalString, HttpProtocols.Http2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
+            AddEndpointConfiguration(initialValues, "Otlp", otlpGrpcUri.OriginalString, HttpProtocols.Http2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
```
JamesNK, line 362:
```suggestion
            AddEndpointConfiguration(initialValues, "OtlpGrpc", otlpGrpcUri.OriginalString, HttpProtocols.Http2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
```

There might be some tests that depend on setting the name. Technically this is a breaking change from external config perspective, but we don't document anywhere that endpoints have names so I don't consider it part of the API.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -357,7 +382,7 @@ private void ConfigureKestrelEndpoints(WebApplicationBuilder builder, DashboardO
         }
         else
         {
-            AddEndpointConfiguration(initialValues, "Otlp", otlpUri.OriginalString, HttpProtocols.Http1AndHttp2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
+            AddEndpointConfiguration(initialValues, "Otlp", otlpGrpcUri.OriginalString, HttpProtocols.Http1AndHttp2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
```
JamesNK, line 385:
```suggestion
            AddEndpointConfiguration(initialValues, "OtlpGrpc", otlpGrpcUri.OriginalString, HttpProtocols.Http1AndHttp2, requiredClientCertificate: dashboardOptions.Otlp.AuthMode == OtlpAuthMode.ClientCertificate);
```

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+
+public static class WebApplicationExtensions
+{
+    // https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/large-object-heap
+    private const int MaxSizeLessThanLOH = 84999;
+
+    public static WebApplication ConfigureHttpOtlp(this WebApplication app, Uri httpEndpoint)
+    {
+        app.MapPost("/v1/metrics",
+                async ([FromServices] OtlpMetricsService service, HttpContext httpContext) =>
```
JamesNK, line 22: Is this injecting the OTLP gRPC service into the OTLP HTTP API? I haven't looked closely into what this means, but on the surface I'm not sure I like that coupling.

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    return await ExportOtlpData(httpContext,
+                            sequence => service.Export(ExportMetricsServiceRequest.Parser.ParseFrom(sequence), null!))
+                        .ConfigureAwait(false);
```
JamesNK, line 26: Is this API accepting a protobuf request message and then returning a JSON response message? I couldn't see any Protobuf serialization code so I assume that the MapPost func is returning a value and ASP.NET Core is doing the serialization, which is always JSON.

I need to read the specification for OTLP+HTTP. I know it supports both Protobuf and JSON, but both in one request seems very weird. I think to support both the API method would have to look at the request content-type and do some content negotiation. I've yet to do that sort of thing inside minimal API methods so I don't know how hard/easy it is.

Also, I'm surprised the Protobuf response message just works when serialized to JSON if the .NET serializer is just accepting a plain old C# class and serializing it. Protobuf has rules about how it maps Protobuf messages to JSON (and unfortunately OTLP+HTTP/JSON breaks some of them). Maybe it works because the export response message is so simple.

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+                }
+            ).RequireAuthorization(OtlpAuthorization.PolicyName)
+            .RequireHost($"*:{httpEndpoint.Port}");
```
JamesNK, line 29: I used connection middleware to ensure OTLP + gRPC request came in on the right endpoint. I'd like OTLP + HTTP to be consistent.

There are probably some relatively modest changes/improvements to reuse the same code for both.

JamesNK (changes requested):
Some small nits but also big issues. I need to read the OTLP HTTP spec to understand what exactly we need to support.

I'm on leave at the moment, but when I get back (beginning of June) are you ok with me making changes to your PR? I think communicating everything via comments will be too much overhead.

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
JamesNK: I think it will be much easier to initially only support OTLP+HTTP with Protobuf. Supporting JSON will require a lot of custom JSON serialization.

sharpSteff:
@JamesNK It's a big pleasure to receive a review of Newtonsoft.Json's author. Thanks for all your work!

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: The OTLP .NET SDK  supports only **HTTP/Protobuf** and **grpc** (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md). I suggest to add HTTP/JSON support when OLTP .NET does also support it.

sharpSteff: Regarding the return value, you a probably right. I have to check the spec here.
The OTLP .NET HTTP Exporter pretty much just checks the status code of the response and stores the headers in bad case:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/043d8a31c768d05544c06629e072492238a0069b/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/BaseOtlpHttpExportClient.cs#L43-L57

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: I'm with you. I would like to have a protocol indipendent service to export otlp data

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
sharpSteff: @JamesNK any idea how to make this work elegantly ?

src/Aspire.Dashboard/Otlp/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+using OpenTelemetry.Proto.Collector.Trace.V1;
+
+namespace Aspire.Dashboard.Otlp;
+
+public static class WebApplicationExtensions
+{
+    // https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/large-object-heap
+    private const int MaxSizeLessThanLOH = 84999;
+
+    public static WebApplication ConfigureHttpOtlp(this WebApplication app, Uri httpEndpoint)
```
wertzui, line 18: I think, you have to configure CORS too, so browsers can send data to this endpoint.

davidfowl: Lets avoid the cors configuration until we understand what it means to do this securely.

JamesNK:
I made a bunch of changes.

* Locked the endpoint down to Protobuf payloads
* Make OTLP/gRPC endpoint optional. Either OTLP/gRPC or OTLP/HTTP is required.
* More tests

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,160 @@
...
+
+    public static void ConfigureHttpOtlp(this IEndpointRouteBuilder endpoints, Uri httpEndpoint)
+    {
+        var logsService = endpoints.ServiceProvider.GetRequiredService<OtlpLogsService>();
+        var traceService = endpoints.ServiceProvider.GetRequiredService<OtlpTraceService>();
+        var metricsService = endpoints.ServiceProvider.GetRequiredService<OtlpMetricsService>();
+
+        endpoints.MapPost("/v1/logs", LogsEndpoint).AddOtlpHttpMetadata();
+        endpoints.MapPost("/v1/traces", TracesEndpoint).AddOtlpHttpMetadata();
+        endpoints.MapPost("/v1/metrics", MetricsEndpoint).AddOtlpHttpMetadata();
```
davidfowl, line 31:
```suggestion
        var group =  endpoints.MapGroup("/v1").AddOtlpHttpMetadata();
        
        group.MapPost("logs", LogsEndpoint);
        group.MapPost("traces", TracesEndpoint);
        group.MapPost("metrics", MetricsEndpoint);
```

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,160 @@
...
+                    break;
+                case KnownContentType.Json:
+                default:
+                    context.Response.StatusCode = StatusCodes.Status415UnsupportedMediaType;
+                    break;
+            }
+        }
+    }
+
+    private static async Task WriteOtlpResponse<T>(HttpContext context, T response) where T : IMessage
```
davidfowl, line 83: These are small responses right?

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,160 @@
...
+    // https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/large-object-heap
+    private const int MaxSizeLessThanLOH = 84999;
+    public const string ProtobufContentType = "application/x-protobuf";
+    public const string JsonContentType = "application/json";
+
+    public static void ConfigureHttpOtlp(this IEndpointRouteBuilder endpoints, Uri httpEndpoint)
+    {
+        var logsService = endpoints.ServiceProvider.GetRequiredService<OtlpLogsService>();
+        var traceService = endpoints.ServiceProvider.GetRequiredService<OtlpTraceService>();
+        var metricsService = endpoints.ServiceProvider.GetRequiredService<OtlpMetricsService>();
```
davidfowl, line 27: These should be injected into the endpoints.

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
JamesNK: Yes. Usually just a number. Maybe also a string if there was an error.

JamesNK:
I think everything is done and this is mergable.

A follow-up piece of work is changing the app host templates to include `DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL` env var. Will do that in its own PR as I think it merits its own discussion without distraction.

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,200 @@
...
+using OpenTelemetry.Proto.Collector.Trace.V1;
+
+namespace Aspire.Dashboard.Otlp.Http;
+
+public static class OtlpHttpEndpointsBuilder
+{
+    public const string ProtobufContentType = "application/x-protobuf";
+    public const string JsonContentType = "application/json";
+
+    public static void ConfigureHttpOtlp(this IEndpointRouteBuilder endpoints)
```
davidfowl, line 22: nit: 
```suggestion
    public static void MapOtlpEndpoints(this IEndpointRouteBuilder endpoints)
```

Map* is the verb we use for adding routes. Configure is for options.

JamesNK: I think I came up with the `Map*` verb back in .NET Core 3 😋

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,200 @@
...
+            {
+                return await next(context).ConfigureAwait(false);
+            }
+            catch (BadHttpRequestException ex)
+            {
+                _logger.LogError(ex, "Bad HTTP request when receiving OTLP data.");
+                return Results.BadRequest(ex.Message);
+            }
+        }
+    }
```
davidfowl, line 134:
```suggestion
    private sealed class ErrorHandlerEndpointFilter(ILogger<ErrorHandlerEndpointFilter> logger) : IEndpointFilter
    {
        public async ValueTask<object?> InvokeAsync(EndpointFilterInvocationContext context, EndpointFilterDelegate next)
        {
            try
            {
                return await next(context).ConfigureAwait(false);
            }
            catch (BadHttpRequestException ex)
            {
                logger.LogError(ex, "Bad HTTP request when receiving OTLP data.");
                return Results.BadRequest(ex.Message);
            }
        }
    }
```

src/Aspire.Dashboard/Otlp/Grpc/OtlpGrpcLogsService.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.AspNetCore.Mvc;
+using OpenTelemetry.Proto.Collector.Logs.V1;
+
+namespace Aspire.Dashboard.Otlp.Grpc;
+
+[Authorize(Policy = OtlpAuthorization.PolicyName)]
+[SkipStatusCodePages]
+public class OtlpGrpcLogsService : LogsService.LogsServiceBase
+{
+    private readonly ILogger<OtlpGrpcLogsService> _logger;
```
drewnoakes, line 16: `_logger` is unused.

This class is a good candidate for a primary constructor.

Ditto: `OtlpGrpcMetricsService` and `OtlpGrpcTraceService`.

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+        None,
+        Protobuf,
+        Json
+    }
+
+    private static KnownContentType GetKnownContentType(string? contentType, out StringSegment charSet)
+    {
+        if (contentType != null && MediaTypeHeaderValue.TryParse(contentType, out var mt))
+        {
+            if (string.Equals(mt.MediaType, JsonContentType, StringComparison.OrdinalIgnoreCase))
```
drewnoakes, line 66: Could introduce `StringComparisons.MediaType` for this and the comparison below.

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
```diff
@@ -78,6 +79,30 @@ public static class IntegrationTestHelpers
         return dashboardWebApplication;
     }
 
+    public static HttpClient CreateHttpClient(
+        string address,
+        Action<X509Certificate2?>? validationCallback = null,
+        X509CertificateCollection? clientCertificates = null)
+    {
+        var handler = new SocketsHttpHandler
```
drewnoakes, line 87: For the resource service client, we allow configuration of SSL options via `configuration.Bind("Dashboard:ResourceServiceClient:Ssl", httpHandler.SslOptions);`. Do we want something similar here?

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -89,18 +96,22 @@ public byte[] GetPrimaryApiKeyBytes()
 
     internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
     {
-        if (string.IsNullOrEmpty(EndpointUrl))
+        if (string.IsNullOrEmpty(GrpcEndpointUrl) && string.IsNullOrEmpty(HttpEndpointUrl))
         {
-            errorMessage = $"OTLP endpoint URL is not configured. Specify a {DashboardConfigNames.DashboardOtlpUrlName.EnvVarName} value.";
+            errorMessage = $"Both OTLP gRPC endpoint URL and OTLP HTTP endpoint URL are not configured. Specify either a {DashboardConfigNames.DashboardOtlpGrpcUrlName.EnvVarName} or {DashboardConfigNames.DashboardOtlpHttpUrlName.EnvVarName} value.";
```
drewnoakes, line 101: I found this a bit hard to parse. What about this?

```suggestion
            errorMessage = $"Neither gRPC or HTTP endpoint URL are configured for OTLP. Specify either a {DashboardConfigNames.DashboardOtlpGrpcUrlName.EnvVarName} or {DashboardConfigNames.DashboardOtlpHttpUrlName.EnvVarName} value.";
```

drewnoakes (approved):

tests/Aspire.Dashboard.Tests/Integration/IntegrationTestHelpers.cs
JamesNK: This is just test code. In the real world, the client in this situation is the OTEL SDK.

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
JamesNK: I think this is a one off usage.

JamesNK (approved):

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+using OpenTelemetry.Proto.Collector.Trace.V1;
+
+namespace Aspire.Dashboard.Otlp.Http;
+
+public static class OtlpHttpEndpointsBuilder
+{
+    public const string ProtobufContentType = "application/x-protobuf";
+    public const string JsonContentType = "application/json";
+
+    public static void MapHttpOtlpApi(this IEndpointRouteBuilder endpoints)
```
captainsafia, line 23: It's typically recommended that `MapX` implementations like this one return the `IEndpoitRouteBuilder` instance passed in in order to support chaining invocations onto the call (e.g. adding metadata via `WithMetadta`). Is there a particular reason to prohibit chaining on this API?

davidfowl: This is only called in on place. It’s not a library, just part of an app. It’s less important in that context

src/Aspire.Dashboard/Otlp/Http/OtlpHttpEndpointsBuilder.cs
```diff
@@ -0,0 +1,205 @@
...
+                    }
+                case KnownContentType.Json:
+                default:
+                    context.Response.StatusCode = StatusCodes.Status415UnsupportedMediaType;
+                    return Empty;
+            }
+        }
+    }
+
+    private sealed class OtlpResult<T> : IResult where T : IMessage
```
captainsafia, line 126: Is it helpful for this result type to implement `IEndpointMetadataProvider` so that information about the response produced from these endpoints can surface to documentation tools like OpenAPI?

JamesNK: Right now, the dashboard doesn't produce OpenAPI. This can improve later if that changes.

JamesNK:
Merged! Thanks for your work on this @sharpSteff.

PR #4203, opened 2024-05-16
Updating dependencies to latest versions in NuGet

joperezr:
Since I'm not sure what the state of dependabot is, I manually did a run to check all packages that have new versions in NuGet and the following were the ones found. @eerhardt I'd appreciate a look to ensure that the updates seem reasonable.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4203)

PR #4209, opened 2024-05-16
Add Azure Event Hub emulator support

### LINKED ISSUE #3412: Add Azure EventHub emulator support
According to https://feedback.azure.com/d365community/idea/39679808-7626-ec11-b6e6-000d3a4f032c the Azure Event Hub emulator is currently being developed and a public preview is planned for Q1 of 2024.

We should have support in Aspire for using the emulator locally, similar to CosmosDB and Storage.

```C#
var eventHub = builder.AddAzureEventHubs("eventhubns")
    .AddEventHub("hub")
    .RunAsEmulator();
```

That would spin the emulator up locally and hook up a connection to it for anyone who `.WithReference(eventHub)`.

cc @sebastienros

sebastienros:
Fixes #3412

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4209)


src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,108 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+            ;
+
+        if (storageResource == null)
+        {
+            storageResource = builder.ApplicationBuilder
+                .AddAzureStorage($"{builder.Resource.Name}-storage")
+                .RunAsEmulator();
+        }
+
+        if (storageResource.Resource is not AzureStorageResource storage)
```
davidfowl, line 145: Why the runtime check?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,108 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+
+            if (blobEndpoint == null || tableEndpoint == null)
+            {
+                throw new InvalidOperationException($"A Blob and Table endpoints could not be found is the provided Azure Storage resource.");
+            }
+
+            context.EnvironmentVariables.Add("ACCEPT_EULA", "Y");
+            context.EnvironmentVariables.Add("BLOB_SERVER", $"{blobEndpoint.ContainerHost}:{blobEndpoint.Port}");
+            context.EnvironmentVariables.Add("METADATA_SERVER", $"{tableEndpoint.ContainerHost}:{tableEndpoint.Port}");
+        }));
```
davidfowl, line 163:
```suggestion
        builder.WithEnvironment(context =>
        {
            var blobEndpoint = storage.GetEndpoint("blob");
            var tableEndpoint = storage.GetEndpoint("table");

            if (blobEndpoint == null || tableEndpoint == null)
            {
                throw new InvalidOperationException($"A Blob and Table endpoints could not be found is the provided Azure Storage resource.");
            }

            context.EnvironmentVariables.Add("ACCEPT_EULA", "Y");
            context.EnvironmentVariables.Add("BLOB_SERVER", $"{blobEndpoint.ContainerHost}:{blobEndpoint.Port}");
            context.EnvironmentVariables.Add("METADATA_SERVER", $"{tableEndpoint.ContainerHost}:{tableEndpoint.Port}");
        });
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,108 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+        builder
+            .WithEndpoint(name: "emulator", targetPort: 5672)
+            .WithAnnotation(new ContainerImageAnnotation
+            {
+                Registry = "messagingemulators.azurecr.io",
+                Image = "microsoft/azure/eventhubs/emulator",
+                Tag = "latest"
+            })
+            .WithAnnotation(new ContainerMountAnnotation(
+                Path.GetFullPath(Path.GetTempFileName(), builder.ApplicationBuilder.AppHostDirectory),
```
davidfowl, line 132: This already returns a full path
```suggestion
                Path.GetTempFileName(),
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
```diff
@@ -23,9 +25,18 @@ public class AzureEventHubsResource(string name, Action<ResourceModuleConstruct>
...
+    /// </summary>
+    public bool IsEmulator => this.IsContainer();
+
     /// <summary>
     /// Gets the connection string template for the manifest for the Azure Event Hubs endpoint.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
-        ReferenceExpression.Create($"{EventHubsEndpoint}");
+        IsEmulator
+        ? ReferenceExpression.Create($"Endpoint=sb://{EmulatorEndpoint.Host}:{EmulatorEndpoint.Property(EndpointProperty.Port)};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;EntityPath={Hubs.First().Name}")
```
davidfowl, line 40: Hubs.First() seems problematic

Directory.Packages.props
```diff
@@ -22,7 +22,7 @@
     <PackageVersion Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
     <PackageVersion Include="Azure.Identity" Version="1.11.3" />
-    <PackageVersion Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
+    <PackageVersion Include="Azure.Messaging.EventHubs.Processor" Version="5.11.3" />
```
joperezr, line 25: FYI: I updated this yesterday in https://github.com/dotnet/aspire/pull/4203

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
sebastienros: I think we might not require the `EntityPath` actually if it's provided in the settings, so I will try to remove it.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -46,17 +46,21 @@ protected static string GetNamespaceFromSettings(AzureMessagingEventHubsSettings
...
-                EventHubsConnectionStringProperties.Parse(settings.ConnectionString).FullyQualifiedNamespace :
-                new Uri(settings.FullyQualifiedNamespace).Host;
+            // Extract the namespace from the connection string or qualified namespace
+            ns = string.IsNullOrWhiteSpace(settings.FullyQualifiedNamespace)
+                ? EventHubsConnectionStringProperties.Parse(settings.ConnectionString).Endpoint.Host
+                : new Uri(settings.FullyQualifiedNamespace).Host;
 
-            ns = fullyQualifiedNs[..fullyQualifiedNs.IndexOf('.')];
+            // This is likely to be similar to {yournamespace}.servicebus.windows.net.
+            if (ns.EndsWith(".servicebus.windows.net", StringComparison.OrdinalIgnoreCase))
```
eerhardt, line 55: What about other clouds like China and government? Will this work for them?

cc @jsquire

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
```diff
@@ -0,0 +1,67 @@
...
+using Aspire.Hosting.Lifecycle;
+using System.Text.Json;
+
+namespace Aspire.Hosting.Azure;
+
+internal sealed class AzureEventHubsEmulatorConfigWriterHook : IDistributedApplicationLifecycleHook
+{
+    public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
+    {
+        if (appModel.Resources.OfType<AzureEventHubsResource>().SingleOrDefault() is not { } emulatorResource || !emulatorResource.IsEmulator)
```
eerhardt, line 14: > SingleOrDefault

Why do we think there will be only one?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            // The default consumer group ('$default') is automatically created, and this
+            // is also the one used by a consumer client if no ConsumerGroup value
+            // is provided.
+            writer.WriteStartObject();                  //           {
+            writer.WriteString("Name", hub.Name);       //             "Name": "hub",
+            writer.WriteString("PartitionCount", "2");  //             "PartitionCount": "2",
+            writer.WriteStartArray("ConsumerGroups");   //              "ConsumerGroups": [
+            writer.WriteStartObject();                  //                {
+            writer.WriteString("Name", "cg1");          //                  "Name": "cg1"
```
eerhardt, line 49: What does hard-coding `cg1` here mean? Do you have to have a consumer group?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        writer.WriteEndArray();                         //          ] (/Entities)
+        writer.WriteEndObject();                        //        } 
+        writer.WriteEndArray();                         //      ], (/NamespaceConfig)
+        writer.WriteStartObject("LoggingConfig");       //      "LoggingConfig": {
+        writer.WriteString("Type", "File");             //        "Type": "File"
+        writer.WriteEndObject();                        //        } (/LoggingConfig)
+
+        writer.WriteEndObject();                        //        } (/UserConfig)
+        writer.WriteEndObject();                        //      } (/Root)
```
eerhardt, line 63: (nit) the nesting of `}` in the comments is off

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
jsquire: This will differ between clouds.   I believe China is `.servicebus.chinacloudapi.cn` and US Gov is `servicebus.usgovcloudapi.net`, though I'm not 100% on either.   We don't test in some of the more specialized clouds, so we may see even greater variation.

jsquire: Is the suffix important here?  If we're just trying to trim off the namespace name from the FQNS, couldn't we just do a `Contains('.')` check and, if so, trim?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: There was a bug I reported that if would throw if the array was empty. I just want the default here. I will update since it just got fixed and published.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
sebastienros: The code before that was getting the first `.` but if you try using an IP address then you get the first number. And if you had "localhost" then it would simply fail because there was no ``. This code is used to create a Blob container

```c#
            // Do we have a container name provided in the settings?
            if (string.IsNullOrWhiteSpace(settings.BlobContainerName))
            {
                // If not, we'll create a container name based on the namespace, event hub name and consumer group
                settings.BlobContainerName = $"{ns}-{settings.EventHubName}-{consumerGroup}";
                shouldTryCreateIfNotExists = true;
            }
```

Maybe we could just omit this part from the name, and use some random value to prevent collisions:

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: Will fix.

PS: The `Utf8JsonWriter` is almost overkill here. A template might feel nicer, but I still preferred this construct to something like `""" ... """ + loop + """ ... """`

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
jsquire: ah... fair point.   I hadn't considered the IP address scenario.    If we're looking to create a blob container, maybe we could just strip punctuation and force lowercase?   I don't recall exactly what the character limit is for those, but I think its enough that may work.     If not, maybe using the first 10-15 characters and adding a random suffiX

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
sebastienros: My bad, I think I was not sure I could reference the storage package when I wrote that. Then you confirmed I could, will fix.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
eerhardt: > The Utf8JsonWriter is almost overkill here. 

I had that comment written up... and then deleted it. 😄

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: Got this code from Redis and PhpMyAdmin. I suggest to replace by `FirstOrDefault` in all places, seems good?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: Done

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: Done

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
sebastienros: Done

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
sebastienros: Done

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
eerhardt: Why not a loop? Can't I use multiple EventHubs in my app?

sebastienros: Yes, it's a loop now, refresh.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
sebastienros: Done

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
sebastienros: Done

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
sebastienros: I ended up checking for `.servicebus` and take the first part, like before but better (cn). Then in all other cases I generate a random 12 chars prefix, and it will still be`{random12}-{settings.EventHubName}-{consumerGroup}`. If I were to use "localhost" for instance I am worried that it could conflict with other instance on different ports.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -46,17 +47,26 @@ protected static string GetNamespaceFromSettings(AzureMessagingEventHubsSettings
...
-            ns = fullyQualifiedNs[..fullyQualifiedNs.IndexOf('.')];
+            // This is likely to be similar to {yournamespace}.servicebus.windows.net or {yournamespace}.servicebus.chinacloudapi.cn
+            if (ns.Contains(".servicebus", StringComparison.OrdinalIgnoreCase))
+            {
+                ns = ns[..ns.IndexOf(".servicebus")];
+            }
+            else
+            {
+                // Use a random prefix if no meaningful name is found e.g., "localhost", "127.0.0.1"
+                RandomNumberGenerator.GetHexString(12, true);
```
davidfowl, line 63: Comment on why?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,105 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
     }
+
+    /// <summary>
+    /// Configures an Azure Event Hubs resource to be emulated. This resource requires an <see cref="AzureEventHubsResource"/> to be added to the application model.
+    /// </summary>
+    /// <param name="builder">The Azure Event Hubs resource builder.</param>
+    /// <param name="storageResource">An optional Azure Storage resource.</param>
+    /// <param name="configureContainer">Callback that exposes underlying container used for emulation to allow for customization.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResourceBuilder<AzureEventHubsResource> builder, IResourceBuilder<AzureStorageResource>? storageResource = null, Action<IResourceBuilder<AzureEventHubsEmulatorResource>>? configureContainer = null)
```
davidfowl, line 112: How do you feel about this @mitchdenny? Allowing the storage resource to be passed in? Should we just always add one instead?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,105 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+
+        builder.WithAnnotation(new EnvironmentCallbackAnnotation((EnvironmentCallbackContext context) =>
+        {
+            var blobEndpoint = storage.GetEndpoint("blob");
+            var tableEndpoint = storage.GetEndpoint("table");
+
+            if (blobEndpoint == null || tableEndpoint == null)
+            {
+                throw new InvalidOperationException($"A Blob and Table endpoints could not be found is the provided Azure Storage resource.");
+            }
```
davidfowl, line 155: This is never null.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
sebastienros: Since we are trying to run the emulator locally I don't feel like it matters to pass a well-known instance, almost like an implementation detail and creating separate container should not be much of a burden, not much more resources are taken I believe. Also this can always be added later if users ask for it.

davidfowl: Lets remove it.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubProducerClientComponent.cs
```diff
@@ -32,9 +32,21 @@ protected override void BindSettingsToConfiguration(AzureMessagingEventHubsProdu
         {
             EnsureConnectionStringOrNamespaceProvided(settings, connectionName, configurationSectionName);
 
-            return !string.IsNullOrEmpty(settings.ConnectionString) ?
-                new EventHubProducerClient(settings.ConnectionString, options) :
-                new EventHubProducerClient(settings.FullyQualifiedNamespace, settings.EventHubName, cred, options);
+            if (!string.IsNullOrEmpty(settings.ConnectionString))
```
IEvangelist, line 35: Maybe invert both negations for readability?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,97 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+
+    /// <summary>
+    /// Adds a named volume for the data folder to an Azure Event Hubs emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureEventHubsEmulatorResource"/>.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
+    /// <returns>A builder for the <see cref="AzureEventHubsEmulatorResource"/>.</returns>
+    public static IResourceBuilder<AzureEventHubsEmulatorResource> WithDataVolume(this IResourceBuilder<AzureEventHubsEmulatorResource> builder, string? name = null, bool isReadOnly = false)
+        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/data", isReadOnly);
```
mitchdenny, line 182: Wouldn't these need to do something with the underlying storage emulator as well to make sure data is persisted correctly?

src/Aspire.Hosting.MySql/PhpMyAdminConfigWriterHook.cs
```diff
@@ -14,7 +14,7 @@ public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, C
         var serverFileMount = adminResource.Annotations.OfType<ContainerMountAnnotation>().Single(v => v.Target == "/etc/phpmyadmin/config.user.inc.php");
         var mySqlInstances = appModel.Resources.OfType<MySqlServerResource>();
 
-        if (appModel.Resources.OfType<PhpMyAdminContainerResource>().SingleOrDefault() is not { } myAdminResource)
+        if (appModel.Resources.OfType<PhpMyAdminContainerResource>().FirstOrDefault() is not { } myAdminResource)
```
mitchdenny, line 17: This should be a separate PR.

src/Aspire.Hosting.Redis/RedisCommanderConfigWriterHook.cs
```diff
@@ -11,7 +11,7 @@ internal sealed class RedisCommanderConfigWriterHook : IDistributedApplicationLi
 {
     public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
     {
-        if (appModel.Resources.OfType<RedisCommanderResource>().SingleOrDefault() is not { } commanderResource)
+        if (appModel.Resources.OfType<RedisCommanderResource>().FirstOrDefault() is not { } commanderResource)
```
mitchdenny, line 14: This should be a separate PR.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubProducerClientComponent.cs
mitchdenny: Maybe we could use some pattern matching here instead of nested if clauses :)

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,97 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
         builder.Resource.Hubs.Add((name, configureHub));
         return builder;
     }
+
+    /// <summary>
+    /// Configures an Azure Event Hubs resource to be emulated. This resource requires an <see cref="AzureEventHubsResource"/> to be added to the application model.
+    /// </summary>
+    /// <param name="builder">The Azure Event Hubs resource builder.</param>
+    /// <param name="configureContainer">Callback that exposes underlying container used for emulation to allow for customization.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
```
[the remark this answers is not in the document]
sebastienros: That's fine, I will try to do it in the corresponding places in other components then.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Aspire.Hosting.Azure;
+
+internal sealed class AzureEventHubsEmulatorConfigWriterHook : IDistributedApplicationLifecycleHook
+{
+    public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
+    {
+        var eventHubsEmulatorResources = appModel.Resources.OfType<AzureEventHubsResource>().Where(x => x is { } eventHubsResource && eventHubsResource.IsEmulator);
+
+        if (!eventHubsEmulatorResources.Any())
```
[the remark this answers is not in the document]
sebastienros: This class is not registered if we are in PublishMode, see `RunAsEmulator`.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            using var stream = new FileStream(configFileMount.Source!, FileMode.Create);
+            using var writer = new Utf8JsonWriter(stream);
+
+            writer.WriteStartObject();                      // {
+            writer.WriteStartObject("UserConfig");          //   "UserConfig": {
+            writer.WriteStartArray("NamespaceConfig");      //     "NamespaceConfig": [
+            writer.WriteStartObject();                      //       {
+            writer.WriteString("Type", "EventHub");         //         "Type": "EventHub",
+            writer.WriteString("Name", "emulatorNs1");      //         "Name": "emulatorNs1"
```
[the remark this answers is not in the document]
sebastienros: There can be multiple hubs in this file.

@krishankumar_microsoft how should this resource be named? Or can we remove the name?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,97 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+
+        builder.ApplicationBuilder.Services.TryAddLifecycleHook<AzureEventHubsEmulatorConfigWriterHook>();
+
+        // Add emulator container
+
+        builder
+            .WithEndpoint(name: "emulator", targetPort: 5672)
+            .WithAnnotation(new ContainerImageAnnotation
+            {
+                Registry = "messagingemulators.azurecr.io",
```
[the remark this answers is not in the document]
sebastienros: > see some of the other container resources

I can make a custom class with that, but I couldn't find any example you seem to suggest. Aren't all these in the test project?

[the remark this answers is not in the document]
sebastienros: Disregard, I understand now.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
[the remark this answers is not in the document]
mitchdenny: Yeah what I'm getting at is that for each evenhub namespace in the config file ... it corresponds to a call to `AddAzureEventHubs` in the app host - that resource has a name so we should use that name here.

tests/Aspire.Hosting.Tests/Azure/AzureResourceExtensionsTests.cs
```diff
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Aspire.Hosting.Nats;
```
mitchdenny, line 4: Remove.

tests/Aspire.Hosting.Tests/Azure/AzureResourceExtensionsTests.cs
```diff
@@ -172,4 +173,36 @@ public void AddAzureCosmosDBWithEmulatorGetsExpectedImageTag(string imageTag)
         var actualTag = containerImageAnnotation.Tag;
         Assert.Equal(imageTag ?? "latest", actualTag);
     }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData(8081)]
+    [InlineData(9007)]
+    public void AzureEventHubsWithEmulatorGetsExpectedPort(int? port = null)
```
mitchdenny, line 181: Do we need to add integration tests that actually spin up the emulator?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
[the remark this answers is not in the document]
sebastienros: I get it now, thanks.

tests/Aspire.Hosting.Tests/Azure/AzureResourceExtensionsTests.cs
sebastienros: Yes, I want to do that (started with the easy ones) from existing emulators.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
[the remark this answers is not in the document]
sebastienros: I will have to revert this change, 

```
Expected string to be "emulatorns1" with a length of 11 because NamespaceName is non-modifiable.Only supported emulatorns1, but "eventhubns" has a length of 10, differs near "ven" (index 1).
```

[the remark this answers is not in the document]
davidfowl: odd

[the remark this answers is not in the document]
mitchdenny: wow

src/Aspire.Hosting.Azure.EventHubs/EventHubsEmulatorContainerImageTags.cs
```diff
@@ -1,11 +1,11 @@
...
-namespace Aspire.Hosting.Nats;
+namespace Aspire.Hosting.Azure.EventHubs;
 
 internal static class EventHubsEmulatorContainerImageTags
 {
-    public const string Registry = "messagingemulators.azurecr.io";
-    public const string Image = "microsoft/azure/eventhubs/emulator";
+    public const string Registry = "mcr.microsoft.com";
+    public const string Image = "azure-messaging/eventhubs-emulator";
     public const string Tag = "latest";
```
mitchdenny, line 10: Looks like there isn't a stable version number for this?

mitchdenny (approved):
LGTM!

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,103 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+                Registry = EventHubsEmulatorContainerImageTags.Registry,
+                Image = EventHubsEmulatorContainerImageTags.Image,
+                Tag = EventHubsEmulatorContainerImageTags.Tag
+            })
+            .WithAnnotation(new ContainerMountAnnotation(
+                Path.GetTempFileName(),
+                AzureEventHubsEmulatorResource.EmulatorConfigJsonPath,
+                ContainerMountType.BindMount,
+                isReadOnly: false))
+            ;
```
IEvangelist, line 151:
```suggestion
                isReadOnly: false));
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsEmulatorConfigWriterHook.cs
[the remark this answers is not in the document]
davidfowl: Yea, the team says this is hardcoded for now. So proceed as you have it.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
```diff
@@ -23,9 +25,18 @@ public class AzureEventHubsResource(string name, Action<ResourceModuleConstruct>
...
+    /// </summary>
+    public bool IsEmulator => this.IsContainer();
+
     /// <summary>
     /// Gets the connection string template for the manifest for the Azure Event Hubs endpoint.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
-        ReferenceExpression.Create($"{EventHubsEndpoint}");
+        IsEmulator
+        ? ReferenceExpression.Create($"Endpoint=sb://{EmulatorEndpoint.Host}:{EmulatorEndpoint.Property(EndpointProperty.Port)};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;")
```
davidfowl, line 40:
```suggestion
        ? ReferenceExpression.Create($"Endpoint=sb://{EmulatorEndpoint.Property(EndpointProperty.Host)}:{EmulatorEndpoint.Property(EndpointProperty.Port)};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;")
```

src/Aspire.Hosting.Azure.EventHubs/EventHubsEmulatorContainerImageTags.cs
sebastienros: Not yet https://mcr.microsoft.com/en-us/product/azure-messaging/eventhubs-emulator/tags

```
latest
1.2.3-preview-arm64 
1.2.3-preview-amd64 
```

davidfowl:
Make sure this still works after that connection string change.

sebastienros:
@davidfowl E2E test is passing and tried the playground app locally too after your change which spits out the expected `Endpoint=sb://localhost:54297;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;`

It is mergeable now IMO.

davidfowl (approved):

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -99,4 +101,102 @@ public static IResourceBuilder<AzureEventHubsResource> AddEventHub(this IResourc
...
+    }
+
+    /// <summary>
+    /// Adds a bind mount for the data folder to an Azure Event Hubs emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureEventHubsEmulatorResource"/>.</param>
+    /// <param name="path">Relative path to the AppHost where emulator storage is persisted between runs. Defaults to the path '.eventhubs/{builder.Resource.Name}'</param>
+    /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
+    /// <returns>A builder for the <see cref="AzureEventHubsEmulatorResource"/>.</returns>
+    public static IResourceBuilder<AzureEventHubsEmulatorResource> WithDataBindMount(this IResourceBuilder<AzureEventHubsEmulatorResource> builder, string? path = null, bool isReadOnly = false)
```
eerhardt, line 186: Why don't we have a `WithDataVolume`?

eerhardt: Also - what does a data folder do for Azure EventHubs? Doesn't it store all its necessary data in the Storage blobs we gave it?

oising: > Also - what does a data folder do for Azure EventHubs? Doesn't it store all its necessary data in the Storage blobs we gave it?

The EH uses the storage blobs for managing partition leases only, not for general data (messages.) The EH client uses a storage account to hold checkpointing info.

sebastienros: @eerhardt The Azure Storage emulator is using `/data` for its data folder (at least that what our storage component assume, so I'll trust it). I assume this is correct here to do the same since we use the azure emulator too.

But I agree we should do the same with a volume, I will create the PR.

(NB: Was catching up on inclusive code reviews I missed and heard your comment about comments on closed PRs, then remembered this comment 🙇)

PR #4226, opened 2024-05-18
Fix port/targetport fallback logic in proxyless path

### LINKED ISSUE #4225: Proxyless endpoint does not default the TargetPort when using the callback ctor
Suppose you have something like:

```c#
                    builder.WithEndpoint(schemeAsEndpointName ?? endpoint.EndpointName, e =>
                    {
                        e.Port = 5000;
                        e.UriScheme = "http";
                        e.IsProxied = false;
                    });
```

The intent is to have the TargetPort defaults to the Port, per https://github.com/dotnet/aspire/blob/5f2af96cba6ee7764b36e38d7340e530bdb5cd66/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs#L51-L68

But this breaks because that logic happens in the ctor, before the callback is called.

davidebbo:
Fixes #4225
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4226)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -17,7 +17,9 @@ public sealed class EndpointAnnotation : IResourceAnnotation
 {
     private string? _transport;
     private int? _port;
+    private bool _portSetToNull;
     private int? _targetPort;
+    private bool _targetPortSetToNull;
     /// <summary>
     /// Initializes a new instance of <see cref="EndpointAnnotation"/>.
     /// </summary>
@@ -45,8 +47,8 @@ public sealed class EndpointAnnotation : IResourceAnnotation
         UriScheme = uriScheme;
         _transport = transport;
         Name = name;
-        Port = port;
-        TargetPort = targetPort;
+        _port = port;
+        _targetPort = targetPort;
         IsExternal = isExternal ?? false;
         IsProxied = isProxied;
     }
@@ -72,8 +74,12 @@ public sealed class EndpointAnnotation : IResourceAnnotation
         // It also depends on what the EndpointAnnotation is applied to.
         // In the Container case the TargetPort is the port that the process listens on inside the container,
         //  and the Port is the host interface port, so it is fine for them to be different.
-        get => _port ?? (IsProxied ? null : _targetPort);
-        set => _port = value;
+        get => _port ?? (IsProxied || _portSetToNull ? null : _targetPort);
+        set
+        {
+            _port = value;
+            _portSetToNull = value == null;
+        }
     }
 
     /// <summary>
@@ -85,8 +91,12 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     public int? TargetPort
     {
         // See comment on the Port setter, as this is the reciprocal logic
-        get => _targetPort ?? (IsProxied ? null : _port);
-        set => _targetPort = value;
+        get => _targetPort ?? (IsProxied || _targetPortSetToNull ? null : _port);
+        set
+        {
+            _targetPort = value;
+            _targetPortSetToNull = value == null;
+        }
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -16,6 +16,8 @@ namespace Aspire.Hosting.ApplicationModel;
 public sealed class EndpointAnnotation : IResourceAnnotation
 {
     private string? _transport;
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -81,15 +64,30 @@ public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, strin
...
-    public int? Port { get; set; }
+    public int? Port {
+        // For proxy-less Endpoints the client port and target port should be the same.
+        // Note that this is just a "sensible default"--the consumer of the EndpointAnnotation is free
+        // to change Port and TargetPort after the annotation is created, but if the final values are inconsistent,
+        // the associated resource may fail to run.
+        // It also depends on what the EndpointAnnotation is applied to.
+        // In the Container case the TargetPort is the port that the process listens on inside the container,
+        //  and the Port is the host interface port, so it is fine for them to be different.
+        get => _port ?? (IsProxied ? null : _targetPort);
```
karolz-ms, line 75: I am not sure about this change... this is going to result in pretty weird behavior for a **settable** property:

```csharp
var e = new EndpointAnnotation(ProtocolType.Tcp, port: 1234, targetPort: null, isProxied: false);
Console.WriteLine(e.TargetPort); // prints 1234
e.TargetPort = 4567;
Console.WriteLine(e.TargetPort); // prints 4567
e.TargetPort = null;
Console.WriteLine(e.TargetPort); // prints 1234
e.IsProxied = true;
Console.WriteLine(e.TargetPort); // prints null
```

davidfowl: I was thinking the same thing. This is basically a way to get torn state. Maybe we need a gross internal method that used in the WithEndpoint callback...

davidebbo: Internal method was my first thought, but it felt a bit dirty to have logic outside of `EndpointAnnotation`. And it opens the possibility that some new scenario creates an `EndpointAnnotation` and forgets to call it, leading to inconsistent fallback behavior.

An alternative is to add `_portSettingCalled`/`_targetPortSettingCalled` fields to disable the fallback if the setter was ever called.

Let me know which way you prefer, I can do either.

davidebbo: I made a change along those lines. It's a bit more convoluted than I would have liked, but maybe still better than an internal method called from the outside. I also extended the tests to cover setting to null. PTAL.

karolz-ms (approved):
I can live with this, thanks @davidebbo

mitchdenny:
I don't think we can easily change this now, but one alternative we could have taken here is have `ProxiedEndpointAnnotation` and `EndpointAnnotation` since they can have slightly different shapes and then use the type system to define the behavior.

mitchdenny:
Its ugly but I can't think of a way that is necessarily better.

mitchdenny (approved):

davidebbo:
> I don't think we can easily change this now, but one alternative we could have taken here is have `ProxiedEndpointAnnotation` and `EndpointAnnotation` since they can have slightly different shapes and then use the type system to define the behavior.

That would mean not allowing `IsProxied` to be settable, though that would probably not be a bad thing. It seems strange to allow it to change on a given instance. Anyway, as you say it seems too drastic a change now.

davidebbo:
So it turns out that this change breaks the test `ProxylessContainerWithoutPortThrows`. This test has:

```c#
        var redis = builder.AddRedis("redis").WithEndpoint("tcp", endpoint =>
        {
            endpoint.IsProxied = false;
        });
```

Note that it sets a TargetPort in `AddRedis`, but no Port. With the old logic, the fallback from TargetPort to Port would not happen, because `IsProxied` is set later. With the new logic, the fallback happens, causing the test to *not* throw, and hence fail.

My take is that the new behavior is more correct, and that we should just adjust the test to have:

```
        var redis = builder.AddRedis("redis").WithEndpoint("tcp", endpoint =>
        {
            endpoint.IsProxied = false;
            endpoint.Port = null;
        });
```

Since clearly the intent of the test is to have no Port.

@davidfowl @karolz-ms @mitchdenny do you agree?

Also, kind of concerning that this was not caught earlier. I guess this test doesn't really run anywhere in an automated way?

davidfowl:
> Also, kind of concerning that this was not caught earlier. I guess this test doesn't really run anywhere in an automated way?

This is what we're working on 😄.

davidfowl:
> My take is that the new behavior is more correct, and that we should just adjust the test to have:

This sounds about right.

PR #4240, opened 2024-05-21
[tests] workload tests for test project templates

radical:
- Add tests for test framework specific templates - `aspire-xunit`, `aspire-nunit`, and `aspire-mstest`.
- Also, test these with different project names to validate the generated code
- And add a test for validating a `aspire-starter` template with the various project names
- The tests revealed a missing `using` in the test cs file, fixed here

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4240)


radical:
@DamianEdwards the xunit test file in the template seems to be broken. I think this should fix it.

[`edd5812` (#4240)](https://github.com/dotnet/aspire/pull/4240/commits/edd581268b64a1868d72ddcc0b673ccf55e63c30)

The test failures: https://dev.azure.com/dnceng-public/public/_build/results?buildId=708682&view=ms.vss-test-web.build-test-results-tab

tests/Shared/Aspire.Workload.Testing.props
```diff
@@ -2,6 +2,7 @@
   <Import Project="$(RepoRoot)src\Shared\Workload.props" />
 
   <PropertyGroup>
+    <InstallWorkloadUsingArtifactsAfterThisTarget />
```
joperezr, line 5: NIT: Can we add a comment here explaining why this property needs to get reset.

tests/Shared/Aspire.Workload.Testing.targets
```diff
@@ -169,4 +169,68 @@
...
+                return false;
+            }
+            if (!Directory.Exists(PackagesDirectory))
+            {
+                Log.LogError($"Could not find {nameof(PackagesDirectory)}={PackagesDirectory}");
+                return false;
+            }
+
+            var files = Directory.EnumerateFiles(SourceDirectory, "*.csproj", SearchOption.AllDirectories);
+            PackageNames = files.Where(f => File.ReadAllText(f).Contains("<IsPackable>true</IsPackable>"))
```
joperezr, line 227: This would break if we move the IsPackable to a Directory.Build.props file, right? I wonder if instead of using an inline task if it's best to just use MSBuild task for all of these projects and call a target that just returns the value of IsPackable. We don't have to do it that way but I just think that would probably be more future proof and resilient against refactorings.

joperezr (approved):
Left couple of suggestions, but looks good otherwise.

tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            foreach (var line in File.ReadAllLines(testCsPath))
+            {
+                if (!inTest && line.Contains(marker))
+                {
+                    inTest = true;
+                }
+
+                if (inTest && CommentLineRegex().IsMatch(line))
+                {
+                    sb.AppendLine(CommentLineRegex().Replace(line, "    "));
```
eerhardt, line 145: Hopefully we don't ever add real comments to the code, or else this would start failing.

tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    public static TheoryData<string> ProjectNamesWithTestTemplate_TestData()
+    {
+        var data = new TheoryData<string>();
+        foreach (var name in GetProjectNamesForTest())
+        {
+            data.Add(name);
+        }
+        return data;
+    }
```
eerhardt, line 39: This could be simplified:

```suggestion
    public static IEnumerable<object[]> ProjectNamesWithTestTemplate_TestData()
    {
        foreach (var name in GetProjectNamesForTest())
        {
            yield return [name];
        }
    }
```

eerhardt (approved):

tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs
radical: Lol! Yeah, but that would cause a failure on the PR, and then that person gets to figure it out :/

tests/Shared/Aspire.Workload.Testing.targets
radical: Thanks for the suggestion! Changed.

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -180,6 +182,53 @@ protected static async Task<ResourceRow[]> CheckDashboardHasResourcesAsync(IPage
...
+        }
+
+        // ActiveIssue: https://github.com/dotnet/aspire/issues/4550
+        // yield return "aspire  sta-rter.test"; // Issue: two spaces
+
+        yield return "aspire_starter.1period then.34letters";
+        yield return "aspire-starter & with.1";
+
+        // ActiveIssue: https://github.com/dotnet/aspnetcore/issues/56277
+        // yield return "aspire_😀";
```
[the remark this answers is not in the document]
DamianEdwards: I think falls under "naming is one of the hardest problems in computer science" right? 😉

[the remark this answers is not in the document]
radical: For the curious this might be a bug in the template engine - https://github.com/dotnet/sdk/issues/41667 !

PR #4245, opened 2024-05-21
Improve summary details view / toolbars / navbar experience on desktop for low-vision users, as well as mobile users 

adamint:
Makes page layout / toolbars / detail views accessible to users with low viewport widths and heights (including mobile and desktop low-vision users, both of which I'll refer to as mobile), but **does not yet make individual pages accessible**.

This is done by 
1. Standardizing pages to use [AspirePageContentLayout](https://github.com/dotnet/aspire/pull/4245/files#diff-2ef5849dc9ceeab4bf7bdaa592dd9af81e4e8e48c37f55c69d7be5fc3ecf708d) to lay out their content
2. Add specific DesktopLayout and [MobileLayout](https://github.com/dotnet/aspire/pull/4245/files#diff-be4204cb8a284299f19b66fbf13fddafbeddeae5a376a537341899d94e897559)
3. Make the entire page content scrollable at very low page height for desktop users (as otherwise the page title will significantly obscure page content). In the below screen recording you can see how page scrolling works for 1) very low page height, 2) typical mobile viewports, and 3) typical desktop viewports

https://github.com/dotnet/aspire/assets/20359921/2b1eb691-4b25-4c23-9022-da8d31687c3e

4. Handle browser resizes (and block until we can initially size the viewport) with [BrowserDimensionWatcher.razor.cs](https://github.com/dotnet/aspire/pull/4245/files#diff-8322dddbabbee602ba7c9e12b154de462a914c560a7caf0380ff232fd501c34f)

5. Making toolbars appear as buttons at the bottom of the screen for mobile users 
image
image

6. Expanding details views on mobile to be the full page height, minus the navbar
image

7. Removing the AppBar and adding a navigation menu similar to microsoft.com
image

| Name | Desktop before  | Desktop after | Low vw/vh desktop before  | Low vw/vh desktop after| Mobile before | Mobile after |
| ------------- | ------------- |  ------------- | ------------- | ------------- | ------------- | ------------- |
| Resources page  | image | image | image | [video](https://github.com/dotnet/aspire/assets/20359921/580ef3da-4a15-4382-90ee-469b7bd12133) | image | image |
| Console logs page | image | image | image | [video](https://github.com/dotnet/aspire/assets/20359921/62844df7-99d1-4ce0-a0f6-3f0a9b965b93) | image | image |
| Structured logs page  | image | image | image | image | image | image |
| Trace details summary detail view  | image | image | image (unable to even get details on screen) | image | image | image |

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4245)


src/Aspire.Dashboard/Components/Controls/SummaryDetailsView.razor.cs
```diff
@@ -17,7 +17,7 @@ public partial class SummaryDetailsView<T> : IGlobalKeydownListener, IDisposable
     public RenderFragment? Summary { get; set; }
 
     [Parameter]
-    public RenderFragment? Details { get; set; }
+    public RenderFragment<T>? Details { get; set; }
```
adamint, line 20: As summary details view uses the same selected item, we can just pass it to the render fragment

leslierichardson95 (dismissed):
Everything looks good to me for the most part but for #5, would it still be accessible if the toolbar / filter button was moved to the top of the screen while just using a filter (or funnel) icon?  Having it take up the entire bottom of the screen feels a little like overkill to me, but I'm cool with it if that's the more accessible option.

adamint:
> Everything looks good to me for the most part but for #5, would it still be accessible if the toolbar / filter button was moved to the top of the screen while just using a filter (or funnel) icon? Having it take up the entire bottom of the screen feels a little like overkill to me, but I'm cool with it if that's the more accessible option.

A toolbar button / something on the top of the page would also work. My only concern is around discoverability in that case? I can mock it up

adamint:
@leslierichardson95 what do you think of this?
image

leslierichardson95:
I like that a lot!

tlmii (approved):
Overall I think this looks really good. A couple small comments a long the way but nothing else really jumps out at me.

I'm conflicted on the filter button. I think the bigger button at the bottom looks nicer (and is more easily accessible one-handed on mobile). But it also draws the eye a little too much. I wouldn't mind seeing an alternative where its not accent, but stands out some other way. But I also don't want to delay this PR too much for that as you have tow workable options and a subsequent PR could tweak/change that specific item.

Approving it for the overall work as it looks great and I'm fine with whichever way we land on the filter for now.

JamesNK:
From testing:

image

1. In light mode the text in the badge is black instead of white
2. The state icon and text isn't properly vertically middle aligned

JamesNK:
top-menu-horizontal-scroll

The top menu in the small view adds a horizontal scroll bar when open.

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -10,25 +10,75 @@
...
+                            Label="@resourceType"
+                            @bind-Value:get="isChecked"
+                            @bind-Value:set="c => OnResourceTypeVisibilityChangedAsync(resourceType, c)"/>
+                    }
+                </FluentStack>
+            </div>
+        }
+
+
+            @if (ViewportInformation.IsDesktop)
```
JamesNK, line 57: This indenting doesn't seem correct

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -10,25 +10,75 @@
...
+            }
+            else
+            {
+                <div>
+                    <FluentSearch Placeholder="@ControlsStringsLoc[nameof(ControlsStrings.FilterPlaceholder)]"
+                                  Immediate="true"
+                                  @bind-Value="_filter"
+                                  slot="end"
+                                  Label="@ControlsStringsLoc[nameof(ControlsStrings.FilterPlaceholder)]"
+                                  @bind-Value:after="HandleSearchFilterChangedAsync"/>
```
JamesNK, line 73: Can this duplication be avoided?

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -10,25 +10,75 @@
...
+                        @bind-CheckState="AreAllTypesVisible"/>
+                    @foreach (var (resourceType, _) in _allResourceTypes)
+                    {
+                        var isChecked = _visibleResourceTypes.ContainsKey(resourceType);
+                        <FluentCheckbox
+                            Label="@resourceType"
+                            @bind-Value:get="isChecked"
+                            @bind-Value:set="c => OnResourceTypeVisibilityChangedAsync(resourceType, c)"/>
+                    }
+                </FluentStack>
```
JamesNK, line 52: Can this duplication be avoided (it's copy paste from the content in the popup below)

JamesNK, line 52: Changing these checkboxes isn't reflected in the UI

I would expect the h4 text to change and the tri-state checkboxes to change like the non-mobile view.

type-filter-not-updated

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -10,25 +10,75 @@
...
+            </div>
+        }
+
+
+            @if (ViewportInformation.IsDesktop)
+            {
+                <FluentSearch Placeholder="@ControlsStringsLoc[nameof(ControlsStrings.FilterPlaceholder)]"
```
JamesNK, line 59: State is lost when resizing and switching between views.

Losing filtering state because you resized the browser is very unexpected.

state-lost-on-resize

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
```diff
@@ -8,15 +8,28 @@
 
 <PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsPageTitle)" Loc="@Loc" /></PageTitle>
 
-<div class="resource-logs-layout">
-    <h1 class="page-header">@Loc[nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsHeader)]</h1>
-    <FluentToolbar Orientation="Orientation.Horizontal">
+<AspirePageContentLayout AddNewlineOnDesktopToolbar="true">
```
JamesNK, line 11: When I resize this page the console logs completely break.

I expect the page state (log selected, logs scroll position) to be unaffected.

consolelogs-resize

JamesNK, line 11: Mobile view has multiple scroll bars

image

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
```diff
@@ -16,9 +16,15 @@
 
 <PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.StructuredLogs.StructuredLogsPageTitle)" Loc="@Loc" /></PageTitle>
 
-<div class="logs-layout">
-    <h1 class="page-header">@Loc[nameof(Dashboard.Resources.StructuredLogs.StructuredLogsHeader)]</h1>
-    <FluentToolbar Orientation="Orientation.Horizontal">
+<AspirePageContentLayout
```
JamesNK, line 19: I noticed that any filter change that changes the page URL causes the filter UI to close.

For example, changing the filter textbox doesn't close the filter UI. Changing the application, level or filters does.

filter-close

JamesNK: Also, clicking add filter causes the filter view to close when the sidebar opens.

JamesNK, line 19: The bottom padding/margin is quite large and causes the scroll bar to appear when it looks like it doesn't need to:

image

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
```diff
@@ -30,19 +36,36 @@
...
+
+                <FluentSelect TOption="SelectViewModel<LogLevel?>"
+                              Items="@_logLevels"
+                              Position="SelectPosition.Below"
+                              OptionText="@(c => c.Name)"
+                              @bind-SelectedOption="PageViewModel.SelectedLogLevel"
+                              @bind-SelectedOption:after="HandleSelectedLogLevelChangedAsync"
+                              Width="120px"
+                              Style="min-width: auto;"
+                              AriaLabel="@Loc[nameof(Dashboard.Resources.StructuredLogs.StructuredLogsSelectMinimumLogLevel)]"/>
```
JamesNK, line 53: Can this duplication be removed? This sort of copy and paste feels like a bug waiting to happen. For example, one drop down is changed, and the other isn't. This goes for most duplication in this PR.

src/Aspire.Dashboard/Components/Pages/Traces.razor
```diff
@@ -14,9 +13,11 @@
...
 <PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.Traces.TracesPageTitle)" Loc="@Loc" /></PageTitle>
 
-<div class="traces-layout">
-    <h1 class="page-header">@Loc[nameof(Dashboard.Resources.Traces.TracesHeader)]</h1>
-    <FluentToolbar Orientation="Orientation.Horizontal">
+<AspirePageContentLayout AddNewlineOnDesktopToolbar="true">
+    <PageTitleSection>
+        <h1 class="page-header">@Loc[nameof(Dashboard.Resources.Traces.TracesHeader)]</h1>
+    </PageTitleSection>
+    <ToolbarSection>
```
JamesNK, line 20: Filter text is lost when switching between views.

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
```diff
@@ -12,12 +12,17 @@
 
 @if (_trace is { } trace)
 {
-    <div class="trace-detail-layout">
-        <div class="trace-detail-header page-header">
-            <h1 style="display:inline-block">@GetResourceName(trace.FirstSpan.Source): @trace.FirstSpan.Name</h1>
-            <span class="trace-id">@OtlpHelpers.ToShortenedId(trace.TraceId)</span>
-        </div>
-        <FluentToolbar Orientation="Orientation.Horizontal">
+    <AspirePageContentLayout
```
JamesNK, line 15: This is obviously broken:

image

JamesNK, line 15: The "View trace details" button shows content that says "Filters" at the top. There are no filters.

view-trace-details

JamesNK, line 15: The open span details is lost when switching views. When this is fixed, make sure the state of the span details stays the same when switching between views.

JamesNK, line 15: The horizontal/vertical button is displayed in mobile view. It doesn't do anything.

image

src/Aspire.Dashboard/Components/Pages/Metrics.razor
```diff
@@ -12,9 +12,11 @@
 
 <PageTitle><ApplicationName ResourceName="@nameof(Dashboard.Resources.Metrics.MetricsPageTitle)" Loc="@Loc" /></PageTitle>
 
-<div class="metrics-layout">
-    <h1 class="page-header">@Loc[nameof(Dashboard.Resources.Metrics.MetricsHeader)]</h1>
-    <FluentToolbar Orientation="Orientation.Horizontal">
+<AspirePageContentLayout AddNewlineOnDesktopToolbar="true">
```
JamesNK, line 15: The meter/instrument area scrollbar isn't displayed. Instead, the whole page because scrollable.

image

JamesNK: Also, the horizontal scrollbar for the graph area (it required because of the browser size) isn't visible unless you scroll the page to the bottom.

JamesNK (changes requested):
Initial testing. Lots of things to fix.

src/Aspire.Dashboard/wwwroot/js/app.js
```diff
@@ -435,3 +435,20 @@ window.focusElement = function(selector) {
...
+    }
+}
+
+window.listenToWindowResize = function(dotnetHelper) {
+    window.addEventListener('load', () => {
+        dotnetHelper.invokeMethodAsync('OnResizeAsync', { width: window.innerWidth, height: window.innerHeight });
+    });
+
+    window.addEventListener('resize', () => {
+        dotnetHelper.invokeMethodAsync('OnResizeAsync', { width: window.innerWidth, height: window.innerHeight });
```
JamesNK, line 452: The resize event fires very frequently when the browser is being resized with a mouse. We don't want to raise hundreds of events. This should be debounced.

For example, https://stackoverflow.com/a/52256801

src/Aspire.Dashboard/Components/Pages/Resources.razor
JamesNK: nit: `Type filter: xxx` text is bigger than the filter header and doesn't look good.

JamesNK:
Help doesn't do anything in mobile view:

help-button

JamesNK:
The normal navigation sidebar shows which page is open. The mobile navigation menu doesn't give any indication which page is open. It seems like it should.

For example, the traces page in the video below:

menu-selected-page

src/Aspire.Dashboard/Components/Routes.razor
```diff
@@ -1,15 +1,31 @@
...
+<BrowserDimensionWatcher @bind-ViewportInformation="@_viewportInformation" />
+
+@if (_viewportInformation is null)
+{
+    // prevent render until we've determined the browser viewport so that we don't have to re-render
+    // if we guess wrong
+    return;
+}
+
+<CascadingValue Value="@_viewportInformation">
```
JamesNK, line 14: What happens when this changes? Does it cause everything to re-render? I noticed weird behavior on the console logs page with everything reseting on resize. We should be certain about what is happening.

Right now, I believe the value changes for every pixel resize. `ViewportInformation` has height and width, and then IsDesktop and IsUltraLowHeight are calculated from the dimensions.

I think this value should only change if either of these properties changes. That means the type only has IsDesktop and IsUltraLowHeight properties on it that are calculated when it is created. The value only changes if either of these properties changes.

Does that make sense?

src/Aspire.Dashboard/Model/ViewportInformation.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Dashboard.Model;
+
+public record ViewportInformation(int Height, int Width)
```
JamesNK, line 6: readonly struct record?

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: The fact that one control was changed was intentional, but I can do that inline

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: fixed

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
adamint: That is not what appears for me
image

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
adamint: The padding/margin comes from fluentui-blazor. I checked this dialog and the only case where this appears is on very low height, not even on mobile devices with 100% viewport sizes. ie:

image
image

This does not seem to be an issue.

src/Aspire.Dashboard/Components/Pages/Resources.razor
JamesNK: I know the difference here is intentional. I mean future changes.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
JamesNK: [It works on my machine](https://blog.codinghorror.com/the-works-on-my-machine-certification-program/) and marking this as resolved... isn't the best response here.

I'm using Chrome. I don't recognize what browser you're using. Does that explain the difference.

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: We can fix state being lost easily. See the below comment for additional information

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
adamint: It is fairly simple to fix this for console logs, but for the pages with data grids it is not. The component is re-rendered as we are in a new layout, and the data grid does not support persisting its existing state. However, @vnbaaij noted that the next version of fluentui-blazor will have this functionality, so I will create an issue as a follow-up

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor.cs
```diff
@@ -54,6 +55,9 @@ public partial class StructuredLogs : IPageWithSessionAndUrlState<StructuredLogs
     [Inject]
     public required BrowserTimeProvider TimeProvider { get; set; }
 
+    [CascadingParameter]
+    public required ViewportInformation ViewportInformation { get; set; }
```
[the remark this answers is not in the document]
adamint: We could additionally inject the `NavigationManager` in it, since that's also on basically every page. But I'm guessing not a huge priority

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: fixed

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: This has been fixed

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
adamint: > I noticed that any filter change that changes the page URL causes the filter UI to close. For example, changing the filter textbox doesn't close the filter UI. Changing the application, level or filters does.

Dialogs automatically close without any callback when the URL is changed (@vnbaaij I'm guessing that's not intentional), so I have changed it to apply non-navigation changes only when you close the dialog. But it will be expected to close when switching application

> Also, clicking add filter causes the filter view to close when the sidebar opens.

This was on purpose, since focus is getting messed up when multiple panels are open at the same time.

src/Aspire.Dashboard/Components/Pages/Traces.razor
adamint: This has been fixed

src/Aspire.Dashboard/wwwroot/js/app.js
adamint: I added a 100ms debounce time

src/Aspire.Dashboard/Components/Routes.razor
adamint: >  Right now, I believe the viewport value changes for every pixel resize. 
> That's pretty heavy weight overhead

Yep, you're right.

> I think this value should only change if either of these properties changes.

I agree with you, and have updated the type to only contain those two properties, as we don't really need width or height

src/Aspire.Dashboard/Model/ViewportInformation.cs
adamint: changed

src/Aspire.Dashboard/Components/Layout/MobileLayout.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        yield return new(
+            LayoutLoc[nameof(Resources.Layout.NavMenuMetricsTab)],
+            () => NavigateToAsync(DashboardUrls.MetricsUrl()),
+            NavMenu.MetricsIcon()
+        );
+
+        yield return new(
+            LayoutLoc[nameof(Resources.Layout.MainLayoutAspireRepoLink)],
+            () => NavigateToAsync("https://aka.ms/dotnet/aspire/repo"),
```
JamesNK, line 55: This should open in a new window like it does in the desktop view

src/Aspire.Dashboard/Components/Layout/MobileLayout.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+        );
+
+        yield return new(
+            LayoutLoc[nameof(Resources.Layout.MainLayoutAspireDashboardHelpLink)],
+            LaunchHelpAsync,
+            new Icons.Regular.Size24.QuestionCircle()
+        );
+
+        yield return new(
+            LayoutLoc[nameof(Resources.Layout.MainLayoutLaunchSettings)],
```
JamesNK, line 66: nit: I think the text here should just be "Settings".

"Launch settings" is different from all the other menu items. For example, there is "Help", not "Launch Help"

src/Aspire.Dashboard/wwwroot/js/app.js
JamesNK: I tried this out and I think this needs to be tweaked. Debouncing delays the event until resize is finished. That means you keep the old view until resize is finished. Not a great experience.

Instead, throttling the event would be better: https://hackernoon.com/optimizing-performance-with-throttling-in-javascript. The browser can tell the server it is resizing every 150ms.

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
JamesNK: Is the workaround complicated? I'd rather wait for fixes to controls that make the dashboard code more complicated.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
JamesNK: I also get this in Edge:

image

It appears to only happen in desktop view now.

JamesNK: The console logs page needs changes to not lose and rewatch logs on render.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -35,43 +38,58 @@ public partial class Resources : ComponentBase, IAsyncDisposable
...
     public required IJSRuntime JS { get; init; }
+    [Inject]
+    public required ProtectedSessionStorage SessionStorage { get; init; }
+
+    [CascadingParameter]
+    public required ViewportInformation ViewportInformation { get; init; }
+
+    [Parameter]
+    [SupplyParameterFromQuery]
+    public string? Filter { get; set; }
```
JamesNK, line 49: Text fields like this shouldn't go in the query string. Every character change updates the query string. This ruins the back button, because you have to back through each character change.

adamint: The best option I can then think of to avoid it being lost on layout change is to inject a scoped service that contains this field. And then to clear the filter when leaving the page.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
adamint: There will be another log viewer component created on layout change. That cannot be avoided. We may be able to reduce the delay of re-showing the logs by storing previously written logs in a scoped service, but it will not be eliminated.

adamint: I am still unable to reproduce this on Edge. Please share the exact screen resolution you are using

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
adamint: It's not very complicated, consisting of a block in `IPageWithSessionAndUrlState` that prevents re-render and navigation until after the dialog is closed. But ideally there would be a library solution

adamint: @vnbaaij additionally, `OnDialogClosing` is not called when calling `IDialogReference.OnCloseAsync`

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
JamesNK: An unnecessary scrollbar is always visible for me. Browser size doesn't matter.

console-logs-scrollbars

adamint: The log viewer will be disposed and re-created during layout change. I don't see a way around that. We can shorten the amount of time before logs re-appear on screen but they won't stay static.

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
adamint: This has been fixed

adamint: removed

src/Aspire.Dashboard/Components/Layout/MobileLayout.razor.cs
adamint: Changed

src/Aspire.Dashboard/Components/Layout/MobileLayout.razor.cs
adamint: Changed

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
adamint: Fixed

adamint: Blazor deterministic styles were breaking even though they should not have - the solution posed [here](https://khalidabuhakmeh.com/blazors-css-isolation-deep-issue-and-solution) worked.

src/Aspire.Dashboard/Components/Pages/Metrics.razor
adamint: fixed

src/Aspire.Dashboard/wwwroot/js/app.js
adamint: This is much better, thanks. Changed to throttling

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
adamint: Interestingly that doesn't happen on mac, but does on windows. Fixed - resolving. Please reopen if you find other issues.

image

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.css
```diff
@@ -28,6 +28,7 @@
     width: 100%;
     overflow: auto;
     background-color: var(--console-background-color);
+    margin-top: 10px;
```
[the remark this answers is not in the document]
adamint: It could be, but we do use 10px as a margin/padding elsewhere

src/Aspire.Dashboard/wwwroot/css/app.css
```diff
@@ -440,3 +442,7 @@ fluent-data-grid-cell.no-ellipsis {
      */
     height: inherit !important;
 }
+
+.single-line-toolbar-button {
+    margin-top: 5px;
```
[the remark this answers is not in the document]
adamint: see above comment, I'm totally fine with changing it

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
adamint: Made this one component now

adamint:
I removed the duplicated code and added an active color to the icon for the current page. @JamesNK

kvenkatrajan:
Looks great..thanks for putting this together @adamint. Just focussed on the testing for navbar/toolbar and details at the moment since the PR is scoped for the same.

Here are a few issues noticed during testing:

- Logo not present on filter page
image
- Toggling between desktop viewport and mobile view port on the filter page 
image
image
image
- clicking on .NET Aspire repo from the menu  - navigates to the repo and displays the page navigated to prior to the Menu page - should this just stay at the menu page?
- Metrics - filters - should that be appearing on a similar filters page?
image


Usability/Discoverability
- Once a filter is applied there is no context in the view of what filter is applied
- Search for resource is embedded in the Filters page - which requires 2 clicks to get to 
image
Consider adding the search box in the main view and the filters for specific resource types in filters
- No explicit Apply Filter button - kinda confusing to the user whether their filters/search is applied.
image


Validated that:
- shortcuts work
- settings for theme is preserved between the close of settings panel and the individual pages
- help menu works 
- Context between help and other screens are maintained (traces > help > close help > back to traces)
- Show only value updates is preserved across metrics
- Search count reflects the # of rows displayed on page based on search
- Context between last search filter applied remains for example - structured logs > apply filter error > help > console > structured logs displays error

JamesNK:
Should an open mobile view filter close when the view switches to desktop? Right now it stays open.

resize-filter-open

I think it should probably close, but curious about whether other people disagree. Also, it if would be technically difficult to close then don't worry about it.

JamesNK:
The selected menu item in dark mode is hard to see:

image

I think `--accent-foreground-active` looks good on both light and dark.

JamesNK:
The close button when a detail view is open shouldn't have a black background. In all other places the button background is transparent (or the same as the background color to make it look like it is transparent)

image

JamesNK:
I don't think this has to change in this PR, but I'd like the metrics graph to scale down in size if the browser is small:

image

I'll create a follow up issue.

JamesNK:
I noticed a wierd thing that the duration icon scales down and disappears based on available space.

image

If it isn't difficult, could you make the icon hidden when in mobile view?

JamesNK:
> Logo not present on filter page

I think this is fine. It's a popup.

> Metrics - filters - should that be appearing on a similar filters page?

The metrics page experience could be improved (see earlier comment I had about resizing graph), but I think that can be done later. This PR is big enough as it is.

> No explicit Apply Filter button - kinda confusing to the user whether their filters/search is applied.

There isn't an apply button when filtering in desktop view. I guess the difference with desktop view is you can see data immediately change which isn't available in mobile view.

JamesNK:
Console logs aren't automatically updating:

console-logs-watch-broken

New lines are only added after I refresh the page.

JamesNK:
Structured logs automatic scrolling isn't working after switching view. After switching to mobile view:

1. The scrollbar is at the top of the results instead of at the bottom.
2. If scrolled to the bottom, new results should cause automatic scrolling. That isn't happening.

structurelogs-mobile-view

Same issues on the traces page.

JamesNK:
image

I think the console logs page should say what resource it is watching. The filter on this page feels different to other filters because there is only one selected resource, and it feels like the page should tell you it.

Changes:

* Move the filter button to the right on the main title. That makes it consistent with all the other pages.
* Add the name of the resource, followed by the status. e.g. `catalogservice      Watching logs...`

adamint:
> Should an open mobile view filter close when the view switches to desktop? Right now it stays open.
> 
> I think it should probably close, but curious about whether other people disagree. Also, it if would be technically difficult to close then don't worry about it.

I agree with you. The same thing goes for the mobile navigation menu. I've made it so that both now close when layout is switched to desktop.

adamint:
> The selected menu item in dark mode is hard to see:
> 
> 
> I think `--accent-foreground-active` looks good on both light and dark. Alternatively, the icon could be the filled variant.

Changed to accent-foreground-active.

adamint:
> I don't think this has to change in this PR, but I'd like the metrics graph to scale down in size if the browser is small:
> 
> 
> I'll create a follow up issue.

You're welcome to create an issue, but the requested change is already scheduled to happen after this PR is merged. This PR does not change any page content.

adamint:
> I noticed a weird thing that the duration icon scales down and disappears based on available space.
> 
> 
> If it isn't difficult, could you make the icon hidden when in mobile view?
Done.

adamint:
> Structured logs automatic scrolling isn't working after switching view. After switching to mobile view:
> 
> 1. The scrollbar is at the top of the results instead of at the bottom.

This is not possible to fix until we are able to restore data grid state after re-render, which is a feature that is coming soon in fluentui-blazor.

> 2. If scrolled to the bottom, new results should cause automatic scrolling. That isn't happening.
> 
> 
> Same issues on the traces page.

This actually is also the same issue that was happening to the console logs page, which is now fixed.

adamint:
> image
> 
> I think the console logs page should say what resource it is watching. The filter on this page feels different to other filters because there is only one selected resource, and it feels like the page should tell you it.
> 
> Changes:
> 
> * Move the filter button to the right on the main title. That makes it consistent with all the other pages.
> * Add the name of the resource, followed by the status. e.g. `catalogservice      Watching logs...`

Changed.

adamint:
per offline discussion, re-added active filter button in the mobile footer
image

JamesNK:
This branch is old and has conflicts. Please rebase.

JamesNK:
Console logs stop updating when the layout changes:

console-logs-stop-updating

leslierichardson95 (approved):

JamesNK:
What do you think about moving the metrics instrument section from a tree to the filter popup? It's basically a filter, just displayed in a tree in a sidebar instead of dropdown at the top of the page.

image

Could be done in a follow up PR.

JamesNK:
Metrics filters and state like showing count is being lost when switching between views.

metrics-filters-lost

JamesNK:
On the metrics page it looks kind of odd that there are two places for filters. I wonder if graph filters should move into the view filters popup.

I don't think we have to resolve that now.

image

src/Aspire.Dashboard/Components/Layout/MobileNavMenu.razor.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    private IEnumerable<MobileNavMenuEntry> GetMobileNavMenuEntries()
+    {
+        if (DashboardClient.IsEnabled)
+        {
+            yield return new MobileNavMenuEntry(
+                Loc[nameof(Resources.Layout.NavMenuResourcesTab)],
+                () => NavigateToAsync(DashboardUrls.ResourcesUrl()),
+                DesktopNavMenu.ResourcesIcon(),
+                LinkMatchRegex: new Regex($"^{DashboardUrls.ResourcesUrl()}$")
```
JamesNK, line 40: The regexes here should ignore case and use invariant culture.

src/Aspire.Dashboard/Components/Layout/MobileNavMenu.razor.cs
```diff
@@ -0,0 +1,99 @@
...
+        yield return new MobileNavMenuEntry(
+            Loc[nameof(Resources.Layout.MainLayoutLaunchSettings)],
+            LaunchSettingsAsync,
+            new Icons.Regular.Size24.Settings()
+        );
+    }
+
+    private static Regex GetNonIndexPageRegex(string pageRelativeBasePath)
+    {
+        return new Regex($"^({pageRelativeBasePath}|{pageRelativeBasePath}/.+)$");
```
JamesNK, line 96: For safety, I think [Regex.Escape(String)](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.escape?view=net-8.0) should be used on the path.
```suggestion
        pageRelativeBasePath = Regex.Escape(pageRelativeBasePath);
        return new Regex($"^({pageRelativeBasePath}|{pageRelativeBasePath}/.+)$");
```

src/Aspire.Dashboard/Components/Pages/IPageWithSessionAndUrlState.cs
```diff
@@ -56,13 +57,24 @@ public static class PageExtensions
     /// Called after a change in the view model that will affect the url associated with new page state
     /// to navigate to the new url and save new state in localstorage.
     /// </summary>
-    public static async Task AfterViewModelChangedAsync<TViewModel, TSerializableViewModel>(this IPageWithSessionAndUrlState<TViewModel, TSerializableViewModel> page) where TSerializableViewModel : class
+    public static async Task AfterViewModelChangedAsync<TViewModel, TSerializableViewModel>(this IPageWithSessionAndUrlState<TViewModel, TSerializableViewModel> page, AspirePageContentLayout? layout, bool isChangeInToolbar) where TSerializableViewModel : class
     {
-        var serializableViewModel = page.ConvertViewModelToSerializable();
-        var pathWithParameters = page.GetUrlFromSerializableViewModel(serializableViewModel).ToString();
+        if (layout is not null && !layout.ViewportInformation.IsDesktop && isChangeInToolbar)
```
JamesNK, line 62: Add a comment about what this does

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -85,7 +91,7 @@ static bool SetEqualsKeys(ConcurrentDictionary<string, bool> left, ConcurrentDic
                 var keysLeft = left.Keys;
                 var keysRight = right.Keys;
 
-                return keysLeft.Count == keysRight.Count && keysLeft.SequenceEqual(keysRight, StringComparers.ResourceType);
+                return keysLeft.Count == keysRight.Count && keysLeft.OrderBy(key => key).SequenceEqual(keysRight.OrderBy(key => key), StringComparers.ResourceType);
```
JamesNK, line 94: Should the OrderBy also use StringComparers.ResourceType?

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor.cs
```diff
@@ -335,7 +375,7 @@ public StructuredLogsPageState ConvertViewModelToSerializable()
 
     public void UpdateViewModelFromQuery(StructuredLogsPageViewModel viewModel)
     {
-        PageViewModel.SelectedApplication = _applicationViewModels.GetApplication(Logger, ApplicationName, _allApplication);
+        viewModel.SelectedApplication = _applicationViewModels.GetApplication(Logger, ApplicationName,  _allApplication);
```
JamesNK, line 378:
```suggestion
        viewModel.SelectedApplication = _applicationViewModels.GetApplication(Logger, ApplicationName, _allApplication);
```

src/Aspire.Dashboard/Components/Resize/BrowserDimensionWatcher.razor.cs
```diff
@@ -0,0 +1,57 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.AspNetCore.Components;
+using Microsoft.JSInterop;
+
+namespace Aspire.Dashboard.Components.Resize;
+
+public class BrowserDimensionWatcher : ComponentBase
+{
+    [Parameter] public ViewportInformation? ViewportInformation { get; set; }
```
JamesNK, line 11: nit: Repo convention is attribute on the line above the method/property.

src/Aspire.Dashboard/Model/LogViewerViewModel.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using Aspire.Dashboard.ConsoleLogs;
+
+namespace Aspire.Dashboard.Model;
+
+public class LogViewerViewModel
+{
+    public LogEntries LogEntries { get; }= new();
+    public string? ResourceName { get; set; }
+
```
JamesNK, line 12:
```suggestion
    public LogEntries LogEntries { get; } = new();
    public string? ResourceName { get; set; }
```

src/Aspire.Dashboard/Components/Layout/AspirePageContentLayout.razor.cs
```diff
@@ -0,0 +1,106 @@
...
+using Microsoft.FluentUI.AspNetCore.Components;
+
+namespace Aspire.Dashboard.Components.Layout;
+
+public partial class AspirePageContentLayout : ComponentBase
+{
+    [CascadingParameter]
+    public required ViewportInformation ViewportInformation { get; init; }
+
+    [Parameter] public required RenderFragment PageTitleSection { get; set; }
```
JamesNK, line 16: Repo convention is attribute on line above

src/Aspire.Dashboard/Components/Pages/IPageWithSessionAndUrlState.cs
JamesNK: Isn't AfterViewModelChangedAsync potentially called multiple times on a page? If so, won't this func be added to DialogCloseListeners multiple times?

JamesNK:
Getting close 👍

adamint:
> What do you think about moving the metrics instrument section from a tree to the filter popup? It's basically a filter, just displayed in a tree in a sidebar instead of dropdown at the top of the page.
> 
> image
> 
> Could be done in a follow up PR.

> On the metrics page it looks kind of odd that there are two places for filters. I wonder if graph filters should move into the view filters popup.

Yes, those are both good ideas. Modifying the metric page is out of scope of this PR, I'll include it in follow-up work on this page.

src/Aspire.Dashboard/Components/Resize/BrowserDimensionWatcher.razor.cs
adamint: changed

src/Aspire.Dashboard/Components/Layout/AspirePageContentLayout.razor.cs
adamint: changed

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
adamint: Yea, that's a good idea

src/Aspire.Dashboard/Components/Pages/IPageWithSessionAndUrlState.cs
adamint: Yep, thank you for catching this. I have made it so that only the last listener here will be added

adamint:
> Metrics filters and state like showing count is being lost when switching between views.
> 
> metrics-filters-lost [ ](https://private-user-images.githubusercontent.com/303201/344831731-c6a0904b-902f-4ceb-b9ef-fde7200a9812.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjA2Mjg2NTEsIm5iZiI6MTcyMDYyODM1MSwicGF0aCI6Ii8zMDMyMDEvMzQ0ODMxNzMxLWM2YTA5MDRiLTkwMmYtNGNlYi1iOWVmLWZkZTcyMDBhOTgxMi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNzEwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDcxMFQxNjE5MTFaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT03N2Y3ZTc5N2U3YWYyYTkxMjNhNWUyOWY2NjhkYjVjMzgzNjQyOWFhYzYyYTk1NjE4OWU5ZTFjY2ZlYzAwMTJjJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.OyHitA7IpHYMMBDOI4tcwN5DXO759-o2Umj7Wz-8pMM) metrics-filters-lost [ ](https://private-user-images.githubusercontent.com/303201/344831731-c6a0904b-902f-4ceb-b9ef-fde7200a9812.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjA2Mjg2NTEsIm5iZiI6MTcyMDYyODM1MSwicGF0aCI6Ii8zMDMyMDEvMzQ0ODMxNzMxLWM2YTA5MDRiLTkwMmYtNGNlYi1iOWVmLWZkZTcyMDBhOTgxMi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNzEwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDcxMFQxNjE5MTFaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT03N2Y3ZTc5N2U3YWYyYTkxMjNhNWUyOWY2NjhkYjVjMzgzNjQyOWFhYzYyYTk1NjE4OWU5ZTFjY2ZlYzAwMTJjJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.OyHitA7IpHYMMBDOI4tcwN5DXO759-o2Umj7Wz-8pMM)

This has been fixed by moving that state out of the chart container and into a separate service, since the chart gets re-created.

JamesNK:
> This has been fixed by moving that state out of the chart container and into a separate service, since the chart gets re-created.

@SteveSandersonMS Hi, Blazor question for you. We're adding two views to Aspire: desktop view and mobile view.

Right now, we have page content embedded in each view, which means switching view, causes by the browser window resizing, loses the current page content state. To work around this, we have to store the state somewhere externally and pass it to the page content when it's re-initialized in the new view.

Is there a way to "move" a Blazor component? We want to remove it from one place (the old view) and add it to the other place (the new view).

SteveSandersonMS:
@JamesNK As much as possible, the web-native way to do this is with CSS - a combination of media queries and flex/grid makes it possible to have a single markup representation for very different-looking layouts on different form factors. From the animated GIFs above it looks to me like CSS would comfortably handle this (and would be a lot less code to deal with) but perhaps I'm missing something.

If you must have two physically separate components for the two cases, then you're already doing the right thing of keeping the state external. There isn't a way to move a Blazor component from one parent to another (many parts of the developer programming model, and the framework internals, assume a lifecycle in which components are attached to the hierarchy one time only).

adamint:
> @JamesNK As much as possible, the web-native way to do this is with CSS - a combination of media queries and flex/grid makes it possible to have a single markup representation for very different-looking layouts on different form factors. From the animated GIFs above it looks to me like CSS would comfortably handle this (and would be a lot less code to deal with) but perhaps I'm missing something.
> 
> If you must have two physically separate components for the two cases, then you're already doing the right thing of keeping the state external. There isn't a way to move a Blazor component from one parent to another (many parts of the developer programming model, and the framework internals, assume a lifecycle in which components are attached to the hierarchy one time only).

It is notable that this is a Blazor-specific problem, where there is no facility to avoid disposing/re-creating the main "Body" component of a page when the `MainLayout` is changed. Responsive component rendering is an established, working pattern in some other frameworks (such as React + Chakra UI). The Blazor behavior appears like a bug. @SteveSandersonMS

adamint:
@JamesNK I fixed the issue of console logs not updating after layout. I don't think there were any other requested changes

SteveSandersonMS:
> Responsive component rendering is an established, working pattern in some other frameworks (such as React + Chakra UI)

Can you point to a doc showing what you mean exactly? Normally "responsive" within the context of front-end web apps refers to CSS functionality that does work equally with Blazor.

I don't know whether reparenting is really supported in React - an initial search suggests it's [not a native feature](https://www.reddit.com/r/reactjs/comments/prn3zt/how_to_move_a_component_in_the_dom_without/) as of a few years ago, and I didn't see any clear indication this has changed. But perhaps I'm missing something.

Blazor's nonsupport for reparenting is an intentional design choice, not a bug :)

JamesNK:
> @JamesNK As much as possible, the web-native way to do this is with CSS - a combination of media queries and flex/grid makes it possible to have a single markup representation for very different-looking layouts on different form factors.

Yeah, I have a feeling that the right CSS might have gotten a lot of the way there. But this PR is pretty far advanced, and I don't think we should restart it. And server control over HTML does give ultimate control compared to CSS.

In the future we could look at whether `AspirePageContentLayout.razor` could be refactored to make the overall page layout changes using CSS. That would eliminate the issue with the `@MainContent` being thrown away and recreated, which is the source of most state issues.

JamesNK:
> @JamesNK I fixed the issue of console logs not updating after layout. I don't think there were any other requested changes

console-logs-line-number

Something is wrong with console line numbers. Every time I refresh the page, it increases by about 1000.

And when switching to other resources, their logs also start from a number that is way too high instead of 1.

adamint:
never mind, it's been filed already in #4893 - should be routed to AppHost

JamesNK (approved):

Issue #4246, opened 2024-05-21
Internal pipeline build broken on Windows


radical:
Update: the workload tests have been disabled on Windows/internal pipeline right now to avoid this issue. When this issue is fixed, they can be re-enabled.

The manifest package version on the internal pipeline is not `8.1.0-ci`.

[Log](https://dev.azure.com/dnceng/internal/_build/results?buildId=2456718&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=e2f678a7-b875-5c17-f40e-3e55bcefed52):
```
##[error].packages\microsoft.net.runtime.workloadtesting.internal\9.0.0-preview.4.24208.7\Sdk\WorkloadTesting.Core.targets(201,5): error : (NETCORE_ENGINEERING_TELEMETRY=Build) Restoring packages failed with exit code: 1. Output:
  Determining projects to restore...
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102: Unable to find package Microsoft.NET.Sdk.Aspire.Manifest-8.0.100 with version (= 8.1.0-ci)
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102:   - Found 786 version(s) in dotnet8 [ Nearest version: 8.1.0-preview.1.24178.6 ]
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102:   - Found 8 version(s) in dotnet-public [ Nearest version: 8.0.0-preview.7.24251.11 ]
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102:   - Found 1 version(s) in nuget-local [ Nearest version: 8.1.0-preview.1.24270.2 ]
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102:   - Found 0 version(s) in dotnet9-transport
D:\a\_work\1\s\artifacts\obj\workload-q3zh33t1.fms\surfuxtl.vly\restore\Restore.csproj : error NU1102:   - Found 0 version(s) in dotnet-eng
[...]

PR #4248, opened 2024-05-21
CI: Disable workload tests on Windows for internal builds

radical:
Disable the workload tests on Windows for internal builds.

Issue: https://github.com/dotnet/aspire/issues/4246

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4248)


radical:
cc @joperezr @eerhardt 
Testing internal build - https://dev.azure.com/dnceng/internal/_build/results?buildId=2457459&view=results

eerhardt:
Why does this need to be disabled?

radical:
Because installing the manifest on Windows/internal builds fails as the version string is like `-preview1...` instead of `-ci` like on public pipelines. And I need to get that preview version to the workloads project, and I might just have to unify the existing `build workloads step` with this, which could take a bit of time and I don't want the official builds to be blocked because of that.

This will still run the workload tests on Windows for PRs though.

eerhardt (approved):

PR #4260, opened 2024-05-22
[tests] Workload tests follow up

radical:
- Remove unused workload install target
- BuildEnvironment: always clean the local nuget cache on CI
- Move all the RequiresDocker attribute classes to the Common directory
- Improve failure message for workload tests when sdk+workload is not installed
- Handle IOExceptions when deleting `testroot` to avoid failing the tests if something is in use by a running process
- Install playwright dependencies on windows

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4260)


tests/Aspire.Components.Common.Tests/RequiresDockerAttribute.cs
```diff
@@ -1,12 +1,13 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Xunit;
 using Xunit.Sdk;
 
-namespace Aspire.Workload.Tests;
+namespace Aspire.Components.Common.Tests;
 
-[TraitDiscoverer("Aspire.Workload.Tests.RequiresDockerDiscoverer", "Aspire.Workload.Tests")]
+// class level attribute
```
eerhardt, line 8: What is this comment for?

tests/Shared/WorkloadTesting/BuildEnvironment.cs
```diff
@@ -166,19 +168,66 @@ public BuildEnvironment(bool useSystemDotNet = false, TestTargetFramework target
...
+
+            try
+            {
+                Directory.Delete(TestRootPath, recursive: true);
+            }
+            catch (IOException)
+            {
+                if (EnvironmentVariables.IsRunningOnCI)
+                {
+                    // Investigate failures on CI, so don't guard against exceptions
```
eerhardt, line 210: What does this comment mean?

tests/Shared/WorkloadTesting/BuildEnvironment.cs
```diff
@@ -166,19 +168,66 @@ public BuildEnvironment(bool useSystemDotNet = false, TestTargetFramework target
             LogRootPath = Path.Combine(AppContext.BaseDirectory, "logs");
         }
 
-        if (Directory.Exists(TestRootPath))
+        Console.WriteLine($"*** [{TargetFramework}] Using path for projects: {TestRootPath}");
+        CleanupTestRootPath();
+        if (!Directory.Exists(TestRootPath))
```
eerhardt, line 173: You shouldn't need to check before. You can just call `Directory.CreateDirectory`.

eerhardt (approved):

tests/Shared/WorkloadTesting/BuildEnvironment.cs
radical: Basically that we want to fail on CI if deleting the `testroot` throws an exception, as it would be something that should be checked manually, like some lingering process from a previous run holding onto a file handle in `testroot`.

tests/Aspire.Components.Common.Tests/RequiresDockerAttribute.cs
radical: `RequiresDockerAttribute` - the name by itself seemed unclear on what it was meant for. Maybe I should make this applicable to methods also.

tests/Shared/WorkloadTesting/BuildEnvironment.cs
eerhardt: >  that we want to fail on CI if deleting the testroot throws an exception

How does swallowing the exception cause the test to fail?

radical: OOPS!!

tests/Aspire.Components.Common.Tests/RequiresDockerAttribute.cs
eerhardt: It says it applies to Class and Method in the AttributeUsage.

Can we just use this method instead of `RequiresDockerFact` and `RequiresDockerTheory`? Would you just annotate the method as:
```C#
[Fact]
[RequiresDocker]
```

radical: Yeah, that would work.

eerhardt: Don't need to fix it here, but that would be nice to consolidate on a single way to do this.

PR #4269, opened 2024-05-23
Update references to VS 2022 Preview to use version 17.11

DamianEdwards:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4269)


docs/machine-requirements.md
```diff
@@ -6,7 +6,7 @@ Whether you want to contribute to Aspire, or just use the latest build of Aspire
 
 On Windows, Visual Studio contains special tooling support for .NET Aspire that you will want to have.
 
-[Visual Studio 2022 version 17.10 Preview](https://visualstudio.microsoft.com/vs/preview/)
+[Visual Studio 2022 version 17.11 Preview](https://visualstudio.microsoft.com/vs/preview/)
```
joperezr, line 9: mmm this is not actually a requirement, as you should be able to use Aspire with 17.10. Should we make a note saying that 17.10 is the minimum requirement (and add the right link to it of course), and then say that 17.11 is preferred for newer features?

docs/using-latest-daily.md
```diff
@@ -1,6 +1,6 @@
 # Set up your machine to use the latest Aspire builds
 
-These instructions will get you set up with the latest build of Aspire. If you just want the last preview release of .NET Aspire, the packages are on nuget.org, and install the latest [Visual Studio 2022 version 17.9 Preview](https://visualstudio.microsoft.com/vs/preview/) for the tooling.
+These instructions will get you set up with the latest build of Aspire. If you just want the last preview release of .NET Aspire, the packages are on nuget.org, and install the latest [Visual Studio 2022 version 17.11 Preview](https://visualstudio.microsoft.com/vs/preview/) for the tooling.
```
joperezr, line 3: Same here, not sure if we should be making the claim that 17.11 is required.

docs/machine-requirements.md
DamianEdwards: This is to build Aspire, not use it. Are we confident we've got good coverage of using non-preview bits to build the repo? If so, seems fine to change to your suggestion.

docs/using-latest-daily.md
DamianEdwards: Given the ties between DCP and VS, any new feature that involves VS behavior is by definition going to require the preview bits, even if building the repo doesn't. I figured it was just easier to say that preview is required.

docs/machine-requirements.md
joperezr: I believe so, at least I've been working with 17.10 GA and building the repo and it has worked fine for me. In any case, I did originally believed that this was for consumers of Aspire (as opposed to contributors) so I'm fine with what you prefer.

docs/using-latest-daily.md
joperezr: I just think that there are still some customers out there that have more strict limitations on which versions of tooling they can use for their projects, so I'd probably lean towards saying that the minimum requirement is 17.10 and 17.11 encouraged, until we find that 17.10 is actually not enough. Just my 2 cents though.

DamianEdwards:
Seems we should update this to latest requirements. I'm not aware of us requiring VS Preview at all anymore right? We could just change this to say VS 2022 17.12 or higher?

danmoseley:
I changed it to 17.11 or later -- we can merge if @joperezr confirms that.

BTW, this refers to 17.10 -- still correct?
https://github.com/dotnet/aspire/blob/0767aa11ae86126db7ed1a39820a5dd46b164ab3/docs/list-of-diagnostics.md?plain=1#L9

joperezr (approved):

joperezr:
17.11 is out of support I believe so 17.12 is right.

PR #4273, opened 2024-05-24
Restore sln project guids

sebastienros:
Otherwise this is polluting every PR when a new solution folder is added and VS tries to update the guids.

https://github.com/dotnet/project-system/blob/main/docs/opening-with-new-project-system.md

@drewnoakes will definitely have a good explanation how it happened (dotnet cli?)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4273)


drewnoakes:
No concerns with the change here, though we generally advocate for use of the legacy GUID throughout, even for SDK-style projects. There's a very quick parse of the XML that delegates to the correct project factory when a project is identified as using an SDK.

There is a bug where in some cases the project type GUID changes in the solution file, though we don't have a repro or clear picture of why this happens. I've only seen it once myself, when adding an xUnit project to my solution. Do you have any information about how the GUIDs changed here, prior to your PR?

drewnoakes (approved):

sebastienros:
Only suppositions, from other stackoverflow questions about this exact thing (which you also commented on) it could be an issue when creating the projects from `dotnet new`, and maybe using our templates for creating a component. Will be easy to test locally. Then try the same thing from VS.

PR #4194, opened 2024-05-16
Bumping npgsql dependencies because of CVE-2024-32655.

mitchdenny:
Bumping dependencies due to CVE-2024-32655.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4194)

PR #4283, opened 2024-05-24
Upgrade npgsql to 8.0.3

amcasey:
Effectively a backport of #4194.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4283)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -107,7 +107,7 @@
     <PackageVersion Include="MySqlConnector.Logging.Microsoft.Extensions.Logging" Version="2.1.0" />
     <PackageVersion Include="NATS.Net" Version="2.2.1" />
     <PackageVersion Include="Npgsql.DependencyInjection" Version="8.0.3" />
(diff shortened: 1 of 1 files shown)


danmoseley (approved):

danmoseley:
This assumes we will release an 8.0.1 (8.0.2 I think)? As 8.1 is still in main.

danmoseley:
> This assumes we will release an 8.0.1 (8.0.2 I think)? As 8.1 is still in main.

@joperezr ?

sebastienros:
I think we should put it in main and then decide to backport it or not.  

NB: 8.0.1 was already released (at the same time as 8.0.0), next patch version 8.0.2 is not yet decided (AFAIK).

joperezr:
Correct, it is not yet decided if we'll ship a patch next month or not. If we do, it'll be from this branch. If we don't then this branch is essentially dead-ended. My suggestion is to keep this PR open in case we do end up deciding to ship a patch, and if so we merge.

eerhardt:
Closing as we are shipping 8.1 this week which already contains this change.

PR #4287, opened 2024-05-24
Testing: set launch profile name based on target AppHost

### LINKED ISSUE #4274: Hello, Integration testing problem, hoping to get technical support.
In the integration test, `DistributedApplicationTestingBuilder` was used to create a test environment. As a result, the service discovery could not be automatically injected. The debugging found that the HTTP endpoint of the service was not injected into the application's microservices through environment variables.

### Test

```csharp
[Fact]
public async Task ProductServiceGetProductReturnsProductDetails()
{
    // Arrange
    IDistributedApplicationTestingBuilder appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.HelloWorld_AppHost>();

    await using DistributedApplication app = await appHost.BuildAsync();
    await app.StartAsync();

    // Act
    HttpClient identityServiceHttpClient = app.CreateHttpClient("identityservice");
    HttpResponseMessage loginResponse = await identityServiceHttpClient.PostAsJsonAsync("api/Account/Login", new
    {
        UserName = "admin",
        Password = "admin"
    });

    AccessTokenResponse? accessTokenResponse = await loginResponse.Content.ReadFromJsonAsync<AccessTokenResponse>();

    HttpClient productServiceHttpClient = app.CreateHttpClient("productservice");
    productServiceHttpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", accessTokenResponse?.AccessToken);

    HttpResponseMessage productDetailsResponse = await productServiceHttpClient.GetAsync("api/Products/1");

    JsonNode? result = await productDetailsResponse.Content.ReadFromJsonAsync<JsonNode>();

    string? productName = result?["Name"]?.GetValue<string>();

    // Assert
    Assert.NotNull(productName);
    Assert.Equal("Product 1", productName);
}

```

### Host

```csharp

var identityService = builder.AddProject<Projects.HelloWorld_IdentityService>("identityservice");

var productService = builder.AddProject<Projects.HelloWorld_ProductService>("productservice")
    .WithReference(identityService);

builder.Build().Run();
```

### Problems

Because the productService. WithReference (identityService), it can be found in the productService service through httpclient.GetAsync ("http://IdentityService") It can work normally.

However, in the integrated test environment, use the DistributedApplicationTestingBuilder to start. productService cannot find the identityService service found. Debug found that the HTTP end -point environment variables of the service discovered were not automatically configured in the productService, so the test could not be performed.

Integrated testing may require Unicom test, and service1->service2-> service3 is performed throughout the link test.

### LINKED ISSUE #4282: Error when trying to use DistributedApplicationTestingBuilder with resource with unproxied ports
Trying to use `DistributedApplicationTestingBuilder` in `Aspire.Hosting.Testing` with a resource that uses unproxied ports results in an exception:

```
System.InvalidOperationException : Service 'webfrontend_https_66317ca1' needs to specify a port for endpoint 'https' since it isn't using a proxy.
ApplicationExecutor.AddAllocatedEndpointInfo(IEnumerable`1 resources) line 867
ApplicationExecutor.CreateContainersAndExecutablesAsync(CancellationToken cancellationToken) line 836
ApplicationExecutor.RunApplicationAsync(CancellationToken cancellationToken) line 121
DcpHostService.StartAsync(CancellationToken cancellationToken) line 70
Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
Host.StartAsync(CancellationToken cancellationToken)
ObservedHost.StartAsync(CancellationToken cancellationToken) line 369
DistributedApplicationFactory.StartAsync(CancellationToken cancellationToken) line 52
DelegatedHost.StartAsync(CancellationToken cancellationToken) line 158
DelegatedDistributedApplication.StartAsync(CancellationToken cancellationToken) line 133
WebTests.GetWebResourceRootReturnsOkStatusCode() line 13
WebTests.GetWebResourceRootReturnsOkStatusCode() line 20
```

To reproduce, create a new Aspire Starter App with a tests project and disable the endpoint proxies for the webfrontend resource:

```csharp

var apiService = builder.AddProject<Projects.AspireApp7_ApiService>("apiservice");

builder.AddProject<Projects.AspireApp7_Web>("webfrontend")
    .WithEndpoint("http", ea => ea.IsProxied = false)
    .WithEndpoint("https", ea => ea.IsProxied = false)
    .WithExternalHttpEndpoints()
    .WithReference(apiService);

builder.Build().Run();
```

Run the application and note that it works just fine.

Run the test it will fail with the exception above.

### LINKED ISSUE #4286: ContentRootPath for AppHost project is the test project output directory when using DistrubtedApplicationTestBuilder
When using `DistrubtedApplicationTestBuilder` in a test project to load and test an AppHost project, the `ContentRootPath` for the AppHost project ends up being the test project's output directory. This breaks scenarios that are relying on loading configuration from files like **appsettings.json** as they won't be found when running the AppHost in the context of the test.

We likely need to update `DistrubtedApplicationTestBuilder` to set the `ContentRootPath` for the AppHost project to the project directory.

ReubenBond:
Also: set AppHost `ContentRootPath`.

This adds a `DefaultLaunchProfileAnnotation` to all projects where the launch profile is not explicitly set, which lets us override it via config (not just process-wide env var). This lets us set this value on a per-app-host basis in tests, which we cannot do using env vars (they are global state).

Fixes #4286
Fixes #4282

Possibly fixes #4274

cc @DamianEdwards 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4287)


src/Aspire.Hosting/LaunchProfileExtensions.cs
```diff
@@ -105,31 +106,30 @@ private static bool TrySelectLaunchProfileByOrder(ProjectResource projectResourc
         return true;
     }
 
-    private static bool TrySelectLaunchProfileFromEnvironment(ProjectResource projectResource, [NotNullWhen(true)] out string? launchProfileName)
+    private static bool TrySelectLaunchProfileFromDefaultAnnotation(ProjectResource projectResource, [NotNullWhen(true)] out string? launchProfileName)
     {
-        var launchProfileEnvironmentVariable = Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE");
```
ReubenBond, line 110: Moved to `WithProjectDefaults` via `DefaultLaunchProfileAnnotation`. See https://github.com/dotnet/aspire/pull/4287/files#diff-9e18fbbbb697e4ee2da32654325b13f2d6b8b8659e5a34747a63c8a81781aaedR227

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -131,16 +134,66 @@ private void OnBuilderCreatingCore(DistributedApplicationOptions applicationOpti
...
+        var appHostProjectPath = ResolveProjectPath(_entryPoint.Assembly);
+        if (!string.IsNullOrEmpty(appHostProjectPath))
+        {
+            hostBuilderOptions.ContentRootPath = appHostProjectPath;
+        }
+
+        var appHostLaunchSettings = GetLaunchSettings(appHostProjectPath);
+        if (appHostLaunchSettings?.Profiles.FirstOrDefault().Key is { } profileName)
+        {
+            additionalConfig["AppHost:DefaultLaunchProfileName"] = profileName;
```
mitchdenny, line 153: Wonder whether we should start putting these keys in constants somewhere.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -378,3 +431,40 @@ public async Task StartAsync(CancellationToken cancellationToken = default)
...
     }
 }
+
+internal sealed class LaunchSettings
+{
+    [JsonPropertyName("profiles")]
+    public Dictionary<string, LaunchProfile> Profiles { get; set; } = new Dictionary<string, LaunchProfile>();
+}
+
+internal sealed class LaunchProfile
```
mitchdenny, line 441: We have identical code in `Aspire.Hosting` ... shared code?

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -78,6 +80,41 @@ public async Task GetHttpClientBeforeStart(bool genericEntryPoint)
...
+        await using var app = await appHost.BuildAsync();
+        await app.StartAsync();
+        var hostEnvironment = app.Services.GetRequiredService<IHostEnvironment>();
+        Assert.Contains("TestingAppHost1", hostEnvironment.ContentRootPath);
+    }
+
+    [LocalOnlyTheory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task SelectsFirstLaunchProfile(bool genericEntryPoint)
```
mitchdenny, line 100: Should we be adding an API for selecting the launch profile when using the test builder?

src/Aspire.Hosting/ApplicationModel/DefaultLaunchProfileAnnotation.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+
+namespace Aspire.Hosting.ApplicationModel;
+
+[DebuggerDisplay("Type = {GetType().Name,nq}, LaunchProfileName = {LaunchProfileName}")]
+internal sealed class DefaultLaunchProfileAnnotation : IResourceAnnotation
```
mitchdenny, line 9: Does this need to be public? We've got behavior that sets this annotation when calling `CreateAsync(...)` but we don't have a mechanism to override it, and we can't modify the annotations because we don't have this type public.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -131,16 +134,66 @@ private void OnBuilderCreatingCore(DistributedApplicationOptions applicationOpti
...
+    {
+        if (appHostPath is null || !Directory.Exists(appHostPath))
+        {
+            return null;
+        }
+
+        var projectFileInfo = new DirectoryInfo(appHostPath);
+        var launchSettingsFilePath = projectFileInfo.FullName switch
+        {
+            null => Path.Combine("Properties", "launchSettings.json"),
```
DamianEdwards, line 182: We might want to check if casing for the path is enforced in other places that use it, e.g. `dotnet run`. Would be a pain if `properties/launchsettings.json` was read just fine by `dotnet run` on Linux, but not here.

ReubenBond: `dotnet run` uses the same logic, although they have a case for VB which I think we're ok to ignore: https://github.com/dotnet/sdk/blob/2581c9040605e81811d5d7b4257684c04f1fb382/src/Cli/dotnet/commands/dotnet-run/RunCommand.cs#L135-L140

src/Aspire.Hosting/ApplicationModel/DefaultLaunchProfileAnnotation.cs
ReubenBond: > Does this need to be public? 

What's the use case? I followed the pattern set by `LaunchProfileAnnotation` btw, which is internal.

> We've got behavior that sets this annotation when calling `CreateAsync(...)` but we don't have a mechanism to override it, and we can't modify the annotations because we don't have this type public.

Other than via configuration & env var?

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
ReubenBond: Good idea. Done

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
ReubenBond: We probably should do a sweep of them at some point

mitchdenny (approved):

PR #4308, opened 2024-05-27
Escape ReferenceExpression Format curly braces not used for parameters

### LINKED ISSUE #3787: Lliterals with {} break ReferenceExpression
Repro:

```C#
var s = "{x}";
var expr = ReferenceExpression.Create($"{s}").ValueExpression
```

This explodes. We should handle this.

jeffl8n:
Fixes #3787 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4308)


tests/Aspire.Hosting.Tests/ApplicationModel/ReferenceExpressionTests.cs
```diff
@@ -0,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Xunit;
+
+namespace Aspire.Hosting.Tests.ApplicationModel;
+public class ReferenceExpressionTests {
```
davidfowl, line 7: Can you make the formatting match the rest of the project?

@eerhardt @joperezr Do we have an editor config?

src/Aspire.Hosting/ApplicationModel/ReferenceExpression.cs
```diff
@@ -69,6 +71,15 @@ private ReferenceExpression(string format, IValueProvider[] valueProviders, stri
         return string.Format(CultureInfo.InvariantCulture, Format, args);
     }
 
+    internal static string ParseFormat(string format)
+    {
+        // Escape curly braces which aren't used for a parameter.
+        var parsedFormat = Regex.Replace(format, @"{(?!\d)", "{{");
+        parsedFormat = Regex.Replace(parsedFormat, @"(?<!\d)}", "}}");
```
davidfowl, line 78: Do we need to use a regex?

joperezr: This expression doesn't have loops and writing the iterative code would probably not be much different, but can we at least use the regex source generator for this?

tests/Aspire.Hosting.Tests/ApplicationModel/ReferenceExpressionTests.cs
jeffl8n: There is one in the root. I was making these changes on VSCode on Linux and missed the brace not on the newline. I'll check it out on VS on Windows and make sure all of the editorconfig rules are met.

jeffl8n: Updated with correct formatting.

src/Aspire.Hosting/ApplicationModel/ReferenceExpression.cs
jeffl8n: Created GeneratedRegex and moved to new `ReferenceExpressionParser` partial class.

src/Aspire.Hosting/Utils/ReferenceExpressionParser.cs
```diff
@@ -0,0 +1,22 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Utils;
+internal static partial class ReferenceExpressionParser
+{
+    [GeneratedRegex(@"{(?!\d)")]
+    private static partial Regex LeftNonParameterBraceRegex();
+    [GeneratedRegex(@"(?<!\d)}")]
```
eerhardt, line 11: How does this handle something that ends in a digit like `myVar1`?

jeffl8n: It doesn't handle these cases currently. Thinking about potential use cases more, it also doesn't handle valid parameter formatting correctly - ex: `{0:D}`, `{1:N}`, etc., so this does need an update to not cause an issue if formatting is used.

How far should this logic go to handle if a poorly formatted string is passed in as the Format value? The error here is really bubbling up from `string.Format(...)` when evaluating `ValueExpression` - https://github.com/dotnet/aspire/blob/ff9732fafed47140ac38226acdfef4ecd8eb5942/src/Aspire.Hosting/ApplicationModel/ReferenceExpression.cs#L49

jeffl8n: Modified parsing to use similar logic as is used in StringBuilder.AppendFormat (https://github.com/dotnet/runtime/blob/3393b4d564660e76178c442766f92158f04cfea2/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs#L1518)

src/Aspire.Hosting/ApplicationModel/ReferenceExpression.cs
jeffl8n: Removed RegEx and replaced with logic similar to StringBuilder.AppendFormat.

tests/Aspire.Hosting.Tests/ApplicationModel/ReferenceExpressionTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        var expr = ReferenceExpression.Create($"{input}", [new HostUrl("test")], [parameterValue]).ValueExpression;
+        Assert.Equal(expected, expr);
+    }
+
+    [Theory]
+    [InlineData("{0} {x}", "abc123", "abc123 {x}")]
+    [InlineData("{x} {0}", "abc123", "{x} abc123")]
+    [InlineData("{0} test {x}", "abc123", "abc123 test {x}")]
+    [InlineData("{x} test {0}", "abc123", "{x} test abc123")]
+    public void ReferenceExpressionHandlesValueWithBothParameterAndNonParameterBrackets(string input, string parameterValue, string expected)
```
davidebbo, line 54: Based on my understanding, these 4 test cases are not valid and should actually fail. Unlike with the other `Create` override, this one does not treat the input string as a `ExpressionInterpolatedStringHandler`. It's just a string, and is expected to be a valid format string. The presence of the `{x}` in the format string makes it blow up with "Input string was not in a correct format. Failure to parse near offset 5. Expected an ASCII digit.", which I think is fully expected.

jeffl8n: My understanding of #3787 was to be able to handle when the string had formatting like this and handle / accept it without throwing the exception.

davidebbo: Yes, but I think that only applies to the `ReferenceExpression.Create($"{s}")`, which uses `ExpressionInterpolatedStringHandler`, and not for this overload, which takes a plain format string. In this case, there is no difference between passing `$"{input}"` as you do, and just passing `input`.

That being said, I very well might be wrong, hence waiting for feedback from Aspire folks like @mitchdenny and @davidfowl.

jeffl8n:
Superseeded by #4869 and #5342

Issue #1852, opened 2024-01-25
Add support for a Dockerfile resource


raffaeler:
As for #1326 
> That said, having support in Aspire.Hosting to add a resource via a Dockerfile to simplify using custom containers seems like a reasonable suggestion. Feel free to log that issue

The scenario of the linked issue is about Keycloak container whose image is almost always customized depending on the deployment environment (for example to add certificates to the java keystore inside the container image).

That said, there are many other examples for enterprise images that makes sense to define via a dockerfile.

DamianEdwards:
@mitchdenny @davidfowl thoughts on this one? We could do this all in the `Aspire.Hosting` layer couldn't we, calling `docker build` when running locally and just emitting the path to the Dockerfile in the manifest when publishing?

mitchdenny:
We kinda have some of this already on the deployment side with AsDockerfile but I agree this would be useful.

josephaw1022:
is the AsDockerfile just for the node apps or for project resources as well? @mitchdenny

mitchdenny:
It's for executable resources. What we are thinking here is a resource in its own right that we could build and run.

davidfowl:
We don't support building docker images today for running, we only support deploying them.

mitchdenny:
This is bubbling up towards the top of the list for the 8.1 milestone. There are a few different approaches we could take here. We could do something like this:

```diff
+ public IResourceBuilder<DockerfileResource> AddDockerfile(
+    this IDistributedApplicationBuilder builder,
+    string name,
+    string contextPath,
+    string? dockerfilePath = null)

+ public class DockerfileResource : IResourceWithEnvironment, IResourceWithEndpoints
+ {
+ }
```

This would have corresponding changes in ApplicationExecutor that look for Dockerfile resources specifically and start them as a container, but with the new DCP `build` settings.

Or alternatively we could just leverage the existing container resource and do something like this:

```diff
+ public IResourceBuilder<ContainerResource> FromDockerfile(
+    this IResourceBuilder<ContainerResource> builder,
+    string contextPath,
+    string? dockerfilePath= null)

+ public class DockerfileBuildAnnotation : IAnnotation {}
```

In this scenario, a container built from a Dockerfile would be treated as a container. The only thing that would change is that the application executor would look for the presence of the `DockerfileBuildAnnotation` and inject the appropriate values into the DCP resource when creating it so that DCP can take over building the container.

[...]

PR #4312, opened 2024-05-27
WithDockerfile

mitchdenny:
Related #1852

This PR adds a `FromDockerfile` extenson method to `IResourceBuilder<T> where T: ContainerResource`. It allows developers to point to a context path and optionally a dockerfile which will be built during launch of the AppHost. DCP manages handing off the build operation to the appropriate container runtime.

Status:

- [x] `FromDockerfile` adds metadata to resource and results in build context properties being sent to DCP.
- [x] Basic unit tests for FromDockerfile.
- [x] Manifest generation logic
- [x] Handling of build args.
- [ ] Logging of docker build to output.

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4312)


src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
```diff
@@ -234,6 +234,64 @@ public static IResourceBuilder<ContainerResource> AddContainer(this IDistributed
...
+    /// <example>
+    /// Creates a container called <c>mycontainer</c> with an image called <c>myimage</c>.
+    /// <code language="csharp">
+    /// var builder = DistributedApplication.CreateBuilder(args);
+    /// builder.AddContainer("mycontainer", "myimage")
+    ///        .FromDockerfile("path/to/context");
+    /// builder.Build().Run();
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> FromDockerfile<T>(this IResourceBuilder<T> builder, string contextPath, string? dockerfilePath = null) where T : ContainerResource
```
davidfowl, line 272: We should make a context object that has all of the arguments that can be passed to build. Like the docker build args.

playground/dockerfile/Dockerfile.AppHost/qots/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Stage 1: Build the Go program
+FROM golang:1.16 AS builder
```
karolz-ms, line 2:
```suggestion
FROM golang:1.22 AS builder
```
1.16 is prehistoric

playground/dockerfile/Dockerfile.AppHost/qots/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR /app
+
+COPY . .
+
+RUN go build qots.go
+
+# Stage 2: Run the Go program
+FROM golang:1.16
```
karolz-ms, line 11:
```suggestion
FROM golang:1.22
```

playground/dockerfile/Dockerfile.AppHost/qots/Dockerfile
karolz-ms: (not that it matters that much with this program TBH)

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
mitchdenny: That's kind of what the annotation is. My plan was to put the build args on the annotation.

playground/dockerfile/Dockerfile.AppHost/qots/Dockerfile
mitchdenny: A good test case for parameter support!

playground/dockerfile/Dockerfile.AppHost/aspire-manifest.json
```diff
@@ -0,0 +1,25 @@
...
+      "inputs": {
+        "value": {
+          "type": "string"
+        }
+      }
+    },
+    "mycontainer": {
+      "type": "container.v0",
+      "image": "myimage:latest",
+      "build": {
```
mitchdenny, line 16: @vhvb1989 @ellismg I wanted to point out some upcoming schema changes in support for allowing us to build container images. We are looking at deprecating `dockerfile.v0` in favor of unifying it with `container.v0` and adding extra properties to `container.v0` to provide enough information for deployment tools to perform a build.

This is off the back of some work that @danegsta and @karolz-ms have been doing around adding support for Docker builds in DCP. This PR is getting this working on the app model side of things.

We are targeting this for 8.1. Wanted to get your early feedback here. If we can agree on the schema I'll update the official schema.

We need to decide whether we want to publish a 8.1 schema which adds this or just add to the 8.0 schema.

ellismg: @mitchdenny What does it mean in this case since there is an `image` property on the container in addition to the `build` block?  I would have expected one or the other?  Is this trying to influence the naming of the image that is going to be pushed to the remote ACR after we build the dockerfile from the `build` section?

I think this format would work fine for us - FWIW - we already support doing a `docker build` for `dockerfile.v0` resources.  We'd treat this the same way, and I have a pretty good view about how our manifest loader could just rewrite this into the old form so we could get privates quickly for you to test.

playground/dockerfile/Dockerfile.AppHost/qots/qots.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"math/rand"
+    "runtime"
```
ellismg, line 6: NIT: I think there are spaces here and not a tab and the go toolchain is going to yell during `go build` but maybe it will be cool about this?

playground/dockerfile/Dockerfile.AppHost/aspire-manifest.json
mitchdenny: @ellismg yeah I'm thinking that the image name is just there to influence the name of the image that is pushed. The other aspect of this is the image version label. What do you do with dockerfile.v0 labels today?

mitchdenny:
@radical I'm adding some end to end integration style tests for the new `FromDockerfile` functionality and I'd like to make sure they can run in CI rather than just disabling them (but I still want them to run locally).

What is the best way to do that today?

radical:
> @radical I'm adding some end to end integration style tests for the new `FromDockerfile` functionality and I'd like to make sure they can run in CI rather than just disabling them (but I still want them to run locally).
> 
> What is the best way to do that today?

The `Aspire.Hosting.Tests` don't run on CI yet. The ones that run with docker on CI have `<RunTestsOnHelix ..` in the project files. I think for your case you can:

1.  create a separate tests project for now, and add the following to your csproj:
```xml
    <!-- no docker support on helix/windows yet -->
    <RunTestsOnHelix Condition="'$(OS)' != 'Windows_NT'">true</RunTestsOnHelix>
    <SkipTests Condition="'$(OS)' == 'Windows_NT'">true</SkipTests>
```

2. And add `[RequiresDocker]` attribute on your test methods like https://github.com/dotnet/aspire/blob/58e22a8e7e8298d73f96cccb20637be7d56b3931/tests/Aspire.MySqlConnector.Tests/ConformanceTests.cs#L124-L126 

3. and make sure to reference `<ProjectReference Include="..\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" />` in your tests project.

mitchdenny:
Oh OK so its just Docker support on Windows that is the issue. I am seeing the tests that launch containers working on the Linux build agents:

image

So really this is just about restructuring things so that the the tests on Linux machines in Helix. In this case however Helix won't give us much more in terms of validation than just running the end to test on the agent like we are today. I just need to specify that the agent needs docker.

mitchdenny:
@danegsta @karolz-ms is there any chance we can get the build outputs prepended to the log stream for the containers if a build is performed? We are getting console output but it would be good to see it under the resources in the dashboard in the log stream.

karolz-ms:
> @danegsta @karolz-ms is there any chance we can get the build outputs prepended to the log stream for the containers if a build is performed? We are getting console output but it would be good to see it under the resources in the dashboard in the log stream.

We have a log stream (two of them in fact: stdout and stderr) for the "build and startup" of the container. These are separate from log stream(s) of the running container. AFAIK the dashboard side hasn't made a decision how to show the container startup logs and whether/how to combine them with the container run logs.

playground/dockerfile/Dockerfile.AppHost/aspire-manifest.json
ellismg: @mitchdenny We have been using the current UTC Unix in the generation of the name and label, so we got a unique container image per `deploy`, instead of having a fixed image name that we overwrote each time.  I'm not sure how I should think about pushing to same image over and over again.  A possible nice thing about using a unique image each time is that in something like ACA it does mean the previous revisions point to the older docker container images, which could be useful in some rollback scenario or later on if we want to start exploring something like blue/green deployments or other stagged rollout type operations?

danegsta: One option would be to apply and push more than one tag for the image. A unique deployment tag gets generated using the current pattern, but if `image` is specified, also apply and push that tag.

ellismg: > A unique deployment tag gets generated using the current pattern, but if image is specified, also apply and push that tag.

That makes sense to me - How does a user in code reflect this intention?  I assume they won't be able to do something like:

```csharp
builder.AddContainer()
       .FromDockerfile("qots")
       .WithBuildArg("GO_VERSION", goVersion);
```

(i.e. not passing any arguments to `AddContainer`, because it's not obvious from the API that if you call the "simpler" overload, you need to get on a `.FromDockerfile()` plan.

Would you imagine something like:

```csharp
builder.AddContainer("doesnt", "matter")
       .FromDockerfile("qots")
       .WithBuildArg("GO_VERSION", goVersion)
       .WithoutDefaultTag()
```
?

danegsta: What updating AddContainer to make `image` optional and add a new optional `dockerfile` field as well?

`public static IResourceBuilder<ContainerResource> AddContainer(this IDistributedApplicationBuilder builder, string name, string? image = default, string tag = "latest", string? dockerfile = default)`

```
builder.AddContainer(dockerfile: "quots")
    .WithBuildArg("GO_VERSION", goVersion);
```

It could apply `ContainerImageAnnotation`, `DockerfileBuildAnnotation`, or both depending on the arguments. And if both image, and dockerfile are all null/empty, throw an exception.

`IsContainer` and `GetContainerResources` would both need to be updated to return resources that have either `ContainerImageAnnotation` or `DockerfileBuildAnnotation`.

danegsta: It'd mean `AddContainer` always returns a "valid" container resource, but it might have just a `ContainerImageAnnotation`, just a `DockerfileBuildAnnotation`, or both. In either case `WithImageTag` or `FromDockerfile` could still be appended to modify the annotations

mitchdenny: Hrm ... I think I'm more comfortable with the way it is now. The way I see it is that `AddContainer(...)` is all about telling Aspire that there is a container resource in the model ... for that you need 3 essential bits of information ... the resource name, the image, and optionally the tag (with `latest`) being the default.

The purpose of `FromDockerfile` is all about telling DCP and manifest publishing how that image is sourced. By default we just try to pull it, but if we provide some additional instructions (the new annotation) the deployment tools and DCP will use to build the image locally.

When the container image is built via DCP, we probably want the tag to be unique per run. Here is why ... when we run locally if we had two AppHost projects open with the same name which happened to have two containers with the same name (not unreasonable) we wouldn't want the build/run loop to interfere with one another (so they both couldn't use latest because you might get some weird results).

For deployment tool based builds we'd probably let the tool decide and be consistent with whatever they use for project resource containers.

mitchdenny:
I think for the sake of simplicity you'd want to just combine them into the one stream? Being able to see the Docker pull output etc when you first spin up an AppHost on a dev workstation would be good visual confirmation that something is happening.

We could have some kind of log stream selector but I'm not sure it does anything for usability?

/cc @kvenkatrajan @jamesnk and @drewnoakes for their thoughts on this one.

playground/dockerfile/Dockerfile.AppHost/aspire-manifest.json
danegsta: It feels a little weird to force users specify an image (and potentially a tag) if we aren't going to honor it, but I definitely agree that applying a random tag per run makes sense to avoid conflicts. I think I'd expect to be able to find an image with the specified tag after an F5. I'm wondering if it makes sense for me to add a `tags` property to DCP to support more than one tag during build?

danegsta:
> I think for the sake of simplicity you'd want to just combine them into the one stream? Being able to see the Docker pull output etc when you first spin up an AppHost on a dev workstation would be good visual confirmation that something is happening.
> 
> We could have some kind of log stream selector but I'm not sure it does anything for usability?
> 
> /cc @kvenkatrajan @JamesNK and @drewnoakes for their thoughts on this one.

We designed the new streams so that they'd be easy to combine with the existing streams; even if follow is requested for startup log streams, we'll still terminate the stream once startup is "done". So the logs could be combined simply by reading to the end of the startup logs and then switching to the standard streams.

mitchdenny:
@radical I'm having a problem with CI. I split up the projects the way you suggested. And I'm referencing the common tests project to get the RequiresDocker attribute ... it compiles and tests fine locally but its failing to build:

Are we doing something funky with the project references that might be causing this. I feel like I must be missing something obvious its really got me stumped.

playground/dockerfile/Dockerfile.AppHost/aspire-manifest.json
mitchdenny: I think we honour everything (right now) - which is fine except when the image names conflict and you get two versions of "myfavoritecontainerthatiuseinallmyprojects" both tagged latest and you can a random one loaded as part of a race condition.

Another alternative might be that we always apply the tag as specified, but when the container is launched we use the SHA that DCP would know from doing the build. That would avoid any conflicts.

No need to invent another mechanism to uniquely identify an image from the build. Seem fair?

mitchdenny: Using a sha is potentially a good idea for deployment tools too.

mitchdenny:
@radical in the end I've copied the attribute into `Aspire.Hosting.Containers.Tests` ... I've got no idea why the project reference wasn't working (worked locally so it has to be a CI specific thing that is doing something funny with the the project references).

We are getting quite a collection of custom fact attributes. I'm wondering if we should start to consolidate them all somewhere that can be used across hosting test projects and component test projects.

radical:
We dropped `RequiresDockerFact`, and `RequiresDockerTheory` in favor of a single `RequiresDocker` attribute:

```csharp
[Fact]
[RequiresDocker]
...
```

mitchdenny:
Oh wait ... its a difference from `main` 🤦

davidfowl:
> AFAIK the dashboard side hasn't made a decision how to show the container startup logs and whether/how to combine them with the container run logs.

The apphost

mitchdenny:
@radical I think I need some help getting these tests running in Helix. I tried to copy what was done in endtoend tests but that only succeeded in stopping them running altogether. Do we have any documentation/guidance on how to have a test written that runs in helix that depends on DCP + Docker to be present so we can test things end to end?

radical:
Oh, this is a `Aspire.Hosting.Testing` based test which doesn't work on CI/helix just yet. I'm working on fixing that. The error that you are seeing:
`The Aspire orchestration component is not installed at "/mnt/vss/_work/1/s/.packages/microsoft.developercontrolplane.linux-amd64/0.4.2/tools/dcp".`

.. is because the test assembly was built on the build machine and has the `dcp` path hardcoded, but that path isn't valid on helix.

mitchdenny:
OK so I should really just make this run on the build machine (when its Linux) for now.

danmoseley:
How do we think of API review for new public API going forward? What is the right process ?

mitchdenny:
> How do we think of API review for new public API going forward? What is the right process ?

We should. The process was started in the linked issue but doesn't necessarily follow a set template like we use in aspnetcore.

https://github.com/dotnet/aspire/issues/1852

mitchdenny:
@danmoseley I ended up creating a proper API Review issue based on the ASP.NET Core template: #4346

davidfowl:
> How do we think of API review for new public API going forward? What is the right process ?

Within the team, but formalizing would be great.

danegsta:
Syntax highlighting in VSCode for the new playground .csproj is defaulting to `Dockerfile` syntax since the file starts with `Dockerfile.`. It might be worth renaming the new samples to `FromDockerfile.AppHost` (or `WithDockerfile.AppHost`) to avoid confusing the highlighter.

mitchdenny:
@ellismg @vhvb1989 we've got a new `container.v1` resource type that adds the `build` field. We'll need support in AZD for this before we can ship this feature. IF the container resource does not have a build field it will still emit a container.v0 resource ... and a build field is explicitly disallowed for a `container.v0`

src/Schema/aspire-8.0.json
```diff
@@ -73,6 +73,47 @@
...
+                "type": "string",
+                "description": "A string representing the container image to be used."
+              },
+              "entrypoint": {
+                "type": "string",
+                "description": "The entrypoint to use for the container image when executed."
+              },
+              "args": {
+                "$ref": "#/definitions/args"
+              },
```
vhvb1989, line 101: azd doesn't support `args` for container.v0!!  (only for project and dockerfile).  

Guessing it might not be relevant, as no one has complained so far :)

src/Schema/aspire-8.0.json
```diff
@@ -73,6 +73,47 @@
...
+            "description": "A generic container resource.",
+            "required": [ "type", "image", "build" ],
+            "properties": {
+              "type": {
+                "const": "container.v1"
+              },
+              "image": {
+                "type": "string",
+                "description": "A string representing the container image to be used."
+              },
```
vhvb1989, line 94: is it still valid to have `image` for container.v1?  It is marked as required, so, I'm a little confused about it.

src/Schema/aspire-8.0.json
davidfowl: It is, we should add it. That's a gap.

src/Schema/aspire-8.0.json
vhvb1989: @mitchdenny , can you confirm if `image` is actually used to give a name for the image that is created from the `build` field?  For example

```json
{
   "image": "regystry/my-image:v1",
   "build": {
        "path": "some/docker/context/path"
   }
}
```

azd would build the dockerfile in the `path` mentioned on the `build` field and use the name/tag `regystry/my-image:v1` for it?   Which means that, the same image is overridden on each deploy until the `image` name is updated in the AppHost to, say, `regystry/my-image:v2` ??

mitchdenny: @vhvb1989 the `image` field is still required and determines the name of the built image. When it comes to deployment using AZD I think it might make sense to push the image to the registry with the repo name being the image name but for the container app to reference the image via SHA?

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -0,0 +1,358 @@
...
+        Assert.Equal(expectedManifest, manifest.ToString());
+    }
+
+    [Fact]
+    public async Task WithDockerfileWithParameterLaunchesContainerSuccessfully()
+    {
+        if (!IsDockerAvailable())
+        {
+            return;
+        }
```
radical, line 105: You should be able to use `[RequiresDocker]` with the test. Though it assumes that `docker` is available unless running on Windows+Helix. And the reason for that is to make sure that tests don't get silently skipped on CI if docker isn't failing somehow.

src/Schema/aspire-8.0.json
mitchdenny: @vhvb1989 following the meeting that we had today I've changed the schema so that `image` and `build` are exclusive fields on `container.v1`. The plan is to reuse the container image naming logic that you use for `project.v0` resources.

src/Schema/aspire-8.0.json
mitchdenny: Noted in the `container.v1` issue on the AZD side so hopefully we can get that fixed up.

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -0,0 +1,358 @@
...
+namespace Aspire.Hosting.Containers.Tests;
+
+public class WithDockerfileTests
+{
+    [Fact]
+    public async Task WithDockerfileLaunchesContainerSuccessfully()
+    {
+        if (!IsDockerAvailable())
+        {
+            return;
```
radical, line 24: If you don't want to use `[RequiresDocker]` then I would suggest making this `[ConditionalFact]` and `throw new SkipException("..")`, so the test correctly shows up as `Skipped`.

mitchdenny: It doesn't seem to work that way. It just results in the test failing on Windows. Are we missing some infrastructure to make this work?

radical: I can take a look later this week.

mitchdenny:
@danegsta I rebased on main with the latest build of DCP but I'm hitting an issue with secrets.

**What is working**
Using a file to pass in a secret is working. I've verified this by building the container and verifying that the contents from the secret file were copied where I expected them to be (not normal usage but needed to verify).

**What is not working**
Environment variable based secrets. If I populate the `Value/value` field it seemds to break the `docker build` command. If I remove the `value` and just keep the `id,type, source` fields on the spec it builds successfully, but the value isn't populated (not surprising).

In this PR I took at stab at updating the `BuildContextSecret` type with the additional fields:

```csharp
internal sealed class BuildContextSecret
{
    // The ID of the secret (a secret can be used in a Dockerfile with `RUN --mount-type=secret,id=<id>,target=<targetpath>`)
    [JsonPropertyName("id")]
    public string? Id { get; set; }

    // Type of the secret, can be "env" or "file".
    [JsonPropertyName("type")]
    public string? Type { get; set; }

    // Value of the secret to be used in the build when the type of the secret is "env".
    [JsonPropertyName("value")]
    public string? Value { get; set; }

    // Path to secret file/folder that will be mounted as a build secret using --secret
    [JsonPropertyName("source")]
    public string? Source { get; set; }
}
```

mitchdenny:
Doc request: https://github.com/dotnet/docs-aspire/issues/1063

mitchdenny:
@vhvb1989 @prom3theu5

This is pretty much done now on the app model side. Just waiting for a bug fix to come in from DCP and it'll be ready to merge. The final manifest format looks like this:

```json
{
  "$schema": "https://json.schemastore.org/aspire-8.0.json",
  "resources": {
    "goversion": {
      "type": "parameter.v0",
      "value": "{goversion.inputs.value}",
      "inputs": {
        "value": {
          "type": "string"
        }
      }
    },
    "secret": {
      "type": "parameter.v0",
      "value": "{secret.inputs.value}",
      "inputs": {
        "value": {
          "type": "string",
          "secret": true
        }
      }
    },
    "mycontainer": {
      "type": "container.v1",
      "build": {
        "context": "qots",
        "dockerfile": "qots/Dockerfile",
        "args": {
          "GO_VERSION": "{goversion.value}"
        },
        "secrets": {
          "SECRET_ASFILE": {
            "type": "file",
            "source": "Program.cs"
          },
          "SECRET_ASENV": {
            "type": "env",
            "value": "{secret.value}"
          }
        }
      }
    }
  }
}
```

We now need to get this working with both AZD and Aspir8.

src/Aspire.Hosting/ApplicationModel/DockerfileBuildAnnotation.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+internal class DockerfileBuildAnnotation(string contextPath, string dockerfilePath, string? stage) : IResourceAnnotation
+{
+    public string ContextPath => contextPath;
+    public string DockerfilePath = dockerfilePath;
+    public string? Stage => stage;
+    public Dictionary<string, object> BuildArguments { get; } = new();
```
IEvangelist, line 11: Why not use the existing `DockerBuildArg` object?

danegsta:
@mitchdenny DCP build with the fix has been merged into main

src/Aspire.Hosting/ApplicationModel/DockerfileBuildAnnotation.cs
mitchdenny: `DockerBuildArg` doesn't give us much (anything) beyond `KeyValuePair<TKey, TValue>`. If this annotation type was public I might be tempted but in this case it would just result in spinning up a bunch of unnecessary objects.

`DockerBuildArg` is also interesting in the context of `PublishAsDockerfile` because you could pass in duplicate args with that API.

`PublishAsDockerfile` is due some TLC after this change goes in.

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -0,0 +1,449 @@
...
+
+        if (createDockerfile)
+        {
+            await File.WriteAllTextAsync(tempDockerfilePath, HelloWorldDockerfile);
+        }
+
+        return (tempContextPath, tempDockerfilePath);
+    }
+
+    private static bool IsDockerAvailable()
```
davidfowl, line 400: Don't we have something better for this? .NET must have something for this.

cc @joperezr @eerhardt

joperezr: We have a `[RequiresDocker]` attribute which can be applied to tests. Looking at both places where this is used, I think it would be better to just use that as opposed to this. The implementation of that attribute is far simpler as it isn't actually checking if docker itself is running, but instead just makes the assumption based on the fact that we are running on Helix. https://github.com/dotnet/aspire/blob/b891c21433a6acf07ead8725fbd99825d7a21ea6/tests/Aspire.Components.Common.Tests/RequiresDockerAttribute.cs#L12

radical: @mitchdenny was trying this but ran into some issues. I will take a look at this to check what might be missing.

mitchdenny: So, I've tried `[RequiresDocker]`. It didn't work. Here are my requirements:

1. Must allow the test to run locally.
2. Must allow the test to run on the build agent on Linux
3. Must allow the test to run on the helix machines on Linux
4. Must not require a PhD in MSBuild to make it work (we want a pattern that others can follow for other resource types).

I don't want to bog this feature down in making this work because this is a long standing issue that is blocking a lot of other tests running in CI as well. But I didn't want to merge this feature without at least some coverage which is why I made it work this way.

radical: I'm trying out a fix. The `IsSupported` condition did not expect the tests to be running on the build machine. It came from `RequiresDockerTheoryAttribute` which was only for helix, but then we had merged the attributes. I'll push a fix for that.

mitchdenny: @radical I accidentally blew away your change after a force push (I was merging in some changes from main and accidentally assumed that I had a conflicting change that I might have already main upstream).

Feel free to force push over my change because your updates would have included what I was doing anyway.

mitchdenny: OK I've put your changes back in. State of git-fu zen achieved.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1614,10 +1615,8 @@ private static async Task ApplyBuildArgumentsAsync(Container dcpContainerResourc
                 var valueString = buildSecret.Value switch
                 {
                     FileInfo filePath => filePath.FullName,
-                    string stringValue => stringValue,
```
davidfowl: ?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1614,10 +1615,8 @@ private static async Task ApplyBuildArgumentsAsync(Container dcpContainerResourc
...
                     IValueProvider valueProvider => await valueProvider.GetValueAsync(cancellationToken).ConfigureAwait(false),
-                    bool boolValue => boolValue ? "true" : "false",
-                    _ => buildSecret.Value.ToString()
+                    _ => throw new InvalidOperationException("Build secret can only be a parameter or a file.")
```
davidfowl, line 1619: Oh that's nice

mitchdenny: "opinionated" to me means going out of our way to make sure people don't shoot themselves in their feet :)

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: I was thinking ... wait ... I thought I just deleted that code ... and I did :)

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: Makes https://github.com/dotnet/aspire/issues/4429 more important.

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -790,6 +787,30 @@ ARG MESSAGE
         COPY --from=builder /app/static/aspire.html /app/static
         RUN --mount=type=secret,id=FILE_SECRET cp /run/secrets/FILE_SECRET /app/static/FILE_SECRET.txt
         RUN --mount=type=secret,id=ENV_SECRET cp /run/secrets/ENV_SECRET /app/static/ENV_SECRET.txt
+        RUN sleep 40
         """;
 }
 
+internal static class RetryExtensions
+{
+    private static ResiliencePipeline? s_pipeline;
```
davidfowl, line 795: static?

mitchdenny: Yeah, my understanding is that these are reusable.

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -18,7 +19,8 @@ namespace Aspire.Hosting.Containers.Tests;
 
 public class WithDockerfileTests(ITestOutputHelper testOutputHelper)
 {
-    [Fact]
+    // Currently we can only run this locally because the CI agents don't have buildkit support enabled.
+    [LocalOnlyFact]
```
davidfowl, line 23: :|

mitchdenny: Temporary until we can get buildkit working. Only impacts secret scenarios.

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
```diff
@@ -0,0 +1,863 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
radical, line 1: The file should be renamed to `WithDockerfileTests.cs`

tests/Aspire.Hosting.Containers.Tests/FromDockerfileTests.cs
mitchdenny: Ok it's working again through the power of environment variables.

mitchdenny:
@davidfowl this is ready to merge now.

davidfowl:
Finally!!!

davidfowl:
>  Logging of docker build to output.

Is this done?

Mrxx99:
Why does the endpoint have to be specified like this when it is already defined in the docker file? Can't it be used from there directly?

```csharp
       .WithEndpoint("tcp", (endpoint) => 
        { 
            endpoint.TargetPort = 6380; // This fixes the app model so it still works.
```

playground/withdockerfile/WithDockerfile.AppHost/qots/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR /app
+COPY . .
+RUN go build qots.go
+
+# Stage 2: Run the Go program
+ARG GO_VERSION
+FROM golang:${GO_VERSION}
+WORKDIR /app
+RUN --mount=type=secret,id=SECRET_ASENV cp /run/secrets/SECRET_ASENV /app/SECRET_ASENV
+RUN --mount=type=secret,id=SECRET_ASFILE cp /run/secrets/SECRET_ASFILE /app/SECRET_ASFILE
```
davidfowl, line 13: Do these get read anywhere

tests/Aspire.Hosting.Tests/Utils/ArgumentEvaluator.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Aspire.Hosting.Tests.Utils;
 
-internal sealed class ArgumentEvaluator
+public sealed class ArgumentEvaluator
```
davidfowl, line 6: Make this shared source.

davidfowl (approved):
Small nits to fix

playground/withdockerfile/WithDockerfile.AppHost/qots/Dockerfile
mitchdenny: Not in the playground. I just use docker desktop to inspect the values to understand what is going on.

mitchdenny:
> Why does the endpoint have to be specified like this when it is already defined in the docker file? Can't it be used from there directly?
> 
> ```cs
>        .WithEndpoint("tcp", (endpoint) => 
>         { 
>             endpoint.TargetPort = 6380; // This fixes the app model so it still works.
> ```

In this particular case the AddRedis(...) command is abstracting the endpoints that we add (although they aren't inferred from the container image). But for this playground sample I'm just testing the ability to mutate the container via a Dockerfile, and that mutation was changing the port that the container listens on.

Even if we did automatically generate endpoints based on the container image we'd still need to add extra information to the endpoints from an Aspire point of view to help with the magic that Aspire does behind the scenes to hook things up seamlessly - so we may as well just deal with it in one place.

tests/Aspire.Hosting.Tests/Utils/ArgumentEvaluator.cs
mitchdenny: We need a bunch of stuff from Aspire.Hosting.Tests, making this shared source would make things more difficult.

joperezr:
As FYI, this change broke internal mirroring as the playground app that was getting added here introduced a string that was flagged as a secret from credscan (a false positive). I'll take a look and fix it.

playground/withdockerfile/WithDockerfile.AppHost/spicyredis/redis.conf
```diff
@@ -0,0 +1,2321 @@
+# Redis configuration file example.
```
eerhardt, line 1: Do we really need this 2k line file in our repo? Can this be slimmed down?

src/Schema/aspire-8.0.json
```diff
@@ -73,6 +73,54 @@
               "bindMounts": {
                 "$ref": "#/definitions/bindMounts"
               },
+              "volumes": {
```
eerhardt, line 76: Should we update this file name? `aspire-8.0.json` seems like what was supported in Aspire 8.0. But now that we have new content in here, revving the version seems appropriate.

tests/Aspire.Hosting.Tests/Schema/SchemaTests.cs
```diff
@@ -266,6 +311,93 @@ public async Task ManifestWithContainerResourceAndNoEnvOrBindingsIsAccepted()
         Assert.True(manifestJson.IsValid(schema));
     }
 
+    [SkipOnHelixFact]
```
eerhardt, line 314: Why are these skipped? Is it because we need the json schema file? If so, we can solve that like we do in the components tests, by copying the json file to the test output directory.

For exampe, see

https://github.com/dotnet/aspire/blob/c294706ce6276d96a0271d9f7b1bb496c72fe165/tests/Aspire.Npgsql.Tests/Aspire.Npgsql.Tests.csproj#L10

PR #4313, opened 2024-05-27
Baseline code coverage

### LINKED ISSUE #4293: Look into adding gates to enforce CodeCoverage in the repo
As discussed offline, we already have Code coverage enabled in the repo, but no-one is looking at it as it isn't really acting as a gate to changes that drop coverage. @RussKie volunteered in bringing the existing infrastructure that lives in dotnet/extensions into the aspire repo, to help ensure that our code coverage levels keep improving to ensure the quality of the incoming PRs. We agreed on doing this as an experiment to see how blocking it would be for new changes, and reassess in the future in case we start thinking this is too drastic.

RussKie:
Resolves #4293
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4313)

### DIFF
--- a/eng/scripts/ValidateProjectCoverage.ps1
+++ b/eng/scripts/ValidateProjectCoverage.ps1
@@ -1,0 +1,205 @@
+﻿#!/usr/bin/env pwsh
+<#
+.SYNOPSIS
+    Validates the code coverage policy for each project.
+.DESCRIPTION
+    This script compares code coverage with thresholds given in "MinCodeCoverage" property in each project.
+    The script writes an error for each project that does not comply with the policy.
+.PARAMETER CoberturaReportXml
+    Path to the XML file to read the code coverage report from in Cobertura format
+.EXAMPLE
+    PS> .\ValidatePerProjectCoverage.ps1 -CoberturaReportXml .\Cobertura.xml
+#>
+
+param (
+    [Parameter(Mandatory = $true, HelpMessage="Path to the XML file to read the code coverage report from")]
+    [string]$CoberturaReportXml
+)
+
+function Write-Header {
+    param($message, [bool]$isError);
+    $color = if ($isError) { 'Red' } else { 'Green' };
+    Write-Host $message -ForegroundColor $color;
+    Write-Host ("=" * 80)
+ }
+
+ function Write-GitHubComment {
+    param($message);
+    $gitHubCommentVar = '##vso[task.setvariable variable=GITHUB_COMMENT]' + $message.Replace("`r`n","`n").Replace("`n","%0D%0A")
+    Write-Host $gitHubCommentVar
+ }
+
+function Get-XmlValue { param($X, $Y); return $X.SelectSingleNode($Y).'#text' }
+
+Write-Verbose "Reading cobertura report..."
+[xml]$CoberturaReport = Get-Content $CoberturaReportXml
+if ($null -eq $CoberturaReport.coverage -or 
+    $null -eq $CoberturaReport.coverage.packages -or
+    $null -eq $CoberturaReport.coverage.packages.package -or 
+    0 -eq $CoberturaReport.coverage.packages.package.count)
+{
+    Write-GitHubComment -message ":bangbang: Code coverage information is missing or invalid."
+    exit -1;
+}
+
+$ProjectToMinCoverageMap = @{}
+
+Get-ChildItem -Path src -Include '*.*sproj' -Recurse | ForEach-Object {
+    if ($_.FullName.Contains('Aspire.ProjectTemplates')) {
+        return
+    }
+
+    $csProjXml = [xml](Get-Content $_)
+    $AssemblyName = Get-XmlValue $csProjXml "//Project/PropertyGroup/AssemblyName"
+    $MinCodeCoverage = Get-XmlValue $csProjXml "//Project/PropertyGroup/MinCodeCoverage"
+
+    if ([string]::IsNullOrWhiteSpace($AssemblyName)) {
+        $AssemblyName = $_.BaseName
+    }
+
+    if ([string]::IsNullOrWhiteSpace($MinCodeCoverage)) {
+        # Test projects may not legitimely have min code coverage set.
+        Write-Warning "$AssemblyName doesn't declare 'MinCodeCoverage' property or value is unset."
+        return
+    }
+
+    $ProjectToMinCoverageMap[$AssemblyName] = $MinCodeCoverage
+}
+
+$esc = [char]27
+$Errors = New-Object System.Collections.ArrayList
+$Kudos = New-Object System.Collections.ArrayList
+$ErrorsMarkdown = @();
+$KudosMarkdown = @();
+
+Write-Verbose "Collecting projects from code coverage report..."
+$CoberturaReport.coverage.packages.package | ForEach-Object {
+    $Name = $_.name
+
+    # Code coverage information is in % (i.e., it's in fractions 0.123456) - bring it back to human readable format, e.g.: 12.34
+    $LineCoverage = [math]::Round([double]$_.'line-rate' * 100, 2)
+    $BranchCoverage = [math]::Round([double]$_.'branch-rate' * 100, 2)
+    $IsFailed = $false
+
+    Write-Verbose "Project $Name with line coverage $LineCoverage and branch coverage $BranchCoverage"
+
+    if ($ProjectToMinCoverageMap.ContainsKey($Name)) {
+        if ($ProjectToMinCoverageMap[$Name] -eq 'n/a')
+        {
+            Write-Host "$Name ...code coverage is not applicable"
+            return
+        }
+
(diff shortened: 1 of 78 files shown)


danmoseley:
Just curious how does increasing thresholds work - we do it sometimes manually when adding tests?

RussKie:
> Just curious how does increasing thresholds work - we do it sometimes manually when adding tests?

Yes, it's a manual process.

In case of a regressed code coverage a build will fail, and a user will see a message as one you can see above. E.g., [this one](https://github.com/dotnet/aspire/pull/4313#issuecomment-2164209657).
However, if the user increased the code coverage and didn't adjust the threasholds before pushing the code (i.e., didn't run the [code coverage locally](https://github.com/dotnet/aspire/blob/68740d44642e80f542513ab2ee08e9d939665599/eng/build.ps1#L118)), then the user will see something a note the following, and the user will need to up the threshold numbers in the affected projects:
image

src/Directory.Build.targets
```diff
@@ -0,0 +1,4 @@
+<Project>
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+  <Import Project="$(MSBuildThisFileDirectory)/../eng/ProjectStaging.targets" />
```
radical, line 3: Would it better to add the target here directly?

src/Microsoft.Extensions.ServiceDiscovery.Abstractions/Microsoft.Extensions.ServiceDiscovery.Abstractions.csproj
```diff
@@ -9,6 +9,10 @@
     <RootNamespace>Microsoft.Extensions.ServiceDiscovery</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MinCodeCoverage>82</MinCodeCoverage>
```
radical, line 13: nit: this can be in the `PropertyGroup` above.

src/Aspire.Hosting.Nats/Aspire.Hosting.Nats.csproj
```diff
@@ -8,6 +8,14 @@
     <PackageIconFullPath>$(SharedDir)nats-icon.png</PackageIconFullPath>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MinCodeCoverage>7</MinCodeCoverage>
```
radical, line 12: This can be removed as it gets overridden below.

eng/scripts/ValidateProjectCoverage.ps1
```diff
@@ -0,0 +1,199 @@
...
+function Get-XmlValue { param($X, $Y); return $X.SelectSingleNode($Y).'#text' }
+
+Write-Verbose "Reading cobertura report..."
+[xml]$CoberturaReport = Get-Content $CoberturaReportXml
+if ($null -eq $CoberturaReport.coverage -or 
+    $null -eq $CoberturaReport.coverage.packages -or
+    $null -eq $CoberturaReport.coverage.packages.package -or 
+    0 -eq $CoberturaReport.coverage.packages.package.count)
+{
+    return
```
radical, line 34: When can this happen? Should this also generate a comment if it means that code coverage did not get generated for some reason?

eng/scripts/ValidateProjectCoverage.ps1
```diff
@@ -0,0 +1,199 @@
...
+}
+
+$ProjectToMinCoverageMap = @{}
+
+Get-ChildItem -Path src -Include '*.*sproj' -Recurse | ForEach-Object {
+    if ($_.FullName.Contains('Aspire.ProjectTemplates')) {
+        return
+    }
+
+    $XmlDoc = [xml](Get-Content $_)
```
radical, line 44: super nit: maybe `projectDoc` or something to indicate that this is the msbuild project file.

eng/scripts/ValidateProjectCoverage.ps1
```diff
@@ -0,0 +1,199 @@
...
+
+    if ([string]::IsNullOrWhiteSpace($AssemblyName)) {
+        $AssemblyName = $_.BaseName
+    }
+
+    if ([string]::IsNullOrWhiteSpace($MinCodeCoverage)) {
+        # Test projects may not legitimely have min code coverage set.
+        Write-Warning "$AssemblyName doesn't declare 'MinCodeCoverage' property"
+        return
+    }
```
radical, line 56: Can this happen if we are failing the build when `$(MinCodeCoverage)==''`?

src/Microsoft.Extensions.ServiceDiscovery.Abstractions/Microsoft.Extensions.ServiceDiscovery.Abstractions.csproj
RussKie: I wanted to keep `MinCodeCoverage` in own group for ease of rebase. I already had to perform few of those and deal with few merge conflicts. If I added the new property to the existing groups, the conflicts would be much harded to deal with.

src/Aspire.Hosting.Nats/Aspire.Hosting.Nats.csproj
RussKie: Oh! Thanks for this, I totally missed it.

eng/scripts/ValidateProjectCoverage.ps1
RussKie: Hard to say. Some of this infra originated from R9, I can speculated that there may have been some weird situations where something like that happened... 
How likely this is to happen, probably unlikely but again, we don't control the subsystem that generates cobertura files ¯\\_(ツ)\_/¯

We can certainly emit a comment at this point, and fail a build (when we'll be failing a build). Good idea 👍

eng/scripts/ValidateProjectCoverage.ps1
```diff
@@ -0,0 +1,199 @@
...
+    $null -eq $CoberturaReport.coverage.packages.package -or 
+    0 -eq $CoberturaReport.coverage.packages.package.count)
+{
+    return
+}
+
+$ProjectToMinCoverageMap = @{}
+
+Get-ChildItem -Path src -Include '*.*sproj' -Recurse | ForEach-Object {
+    if ($_.FullName.Contains('Aspire.ProjectTemplates')) {
```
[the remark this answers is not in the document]
RussKie: I don't quite uderstand what that would mean. Happy to chat to explore this topic further.

eng/scripts/ValidateProjectCoverage.ps1
RussKie: That's precisely what we're checking here - a ptoject must declare a value for `MinCodeCoverage`.

src/Directory.Build.targets
RussKie: I'm not quite sure what you mean... Could you please clarify?

src/Aspire.Hosting.Nats/Aspire.Hosting.Nats.csproj
RussKie: FWIW, the top one is the one we want - that's the one the validation script reads. The botton one is the one redundant.
Since these aren't used by MSBuild, there's no overriding involved.

eng/scripts/ValidateProjectCoverage.ps1
RussKie: How does [this look](https://github.com/dotnet/aspire/pull/4313#issuecomment-2172308578)?
image

eng/scripts/ValidateProjectCoverage.ps1
RussKie: I tweaked the wording a little. Please have a look

eng/scripts/ValidateProjectCoverage.ps1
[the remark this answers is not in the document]
radical: Ensuring that we have coverage for the code emitted in the template projects.

eng/scripts/ValidateProjectCoverage.ps1
radical: This would catch the case where the code coverage report generation did not *fail* but emitted an empty file.

RussKie: We don't control coverage reporting, it's essentailly a 3P tool as far as this CI is concerned, so anything can happen. If the coverage generation fails it would likely fail the build at the previous step. If the coverage file was produced and we got to this point, then we'd fail here, if the content was unexpected.
I think that's good enough for now, we can always tweak it later, if desired.

mitchdenny (approved):
Would be great to get this in. Given this code doesn't ship I don't think it needs to be perfect, its more valuable to have it giving us the feedback on the PRs. We can refine it if it needs improvements.

radical (approved):

PR #4318, opened 2024-05-28
Add support for custom port to RabbitMQ management plugin

### LINKED ISSUE #4316: [RabbitMQ] Provide custom port to management plugin
Currently the `.WithManagementPlugin()` doesn't allow custom configuration and internally resource will be setup with `builder.WithHttpEndpoint(targetPort: 15672, name: RabbitMQServerResource.ManagementEndpointName);`. During development it might be useful to have always the same port of the management UI in order to test, for example, failures across restart.

Dona278:
Fixes #4316
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4318)


Dona278:
@dotnet-policy-service agree company="Hexesoft"

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -78,9 +78,10 @@ public static IResourceBuilder<RabbitMQServerResource> WithDataBindMount(this IR
     /// Calling this method on a resource configured with an unrecognized image registry, name, or tag will result in a <see cref="DistributedApplicationException"/> being thrown.
     /// </remarks>
     /// <param name="builder">The resource builder.</param>
+    /// <param name="port">The host port that can be used to access the management UI page when running locally.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     /// <exception cref="DistributedApplicationException">Thrown when the current container image and tag do not match the defaults for <see cref="RabbitMQServerResource"/>.</exception>
-    public static IResourceBuilder<RabbitMQServerResource> WithManagementPlugin(this IResourceBuilder<RabbitMQServerResource> builder)
+    public static IResourceBuilder<RabbitMQServerResource> WithManagementPlugin(this IResourceBuilder<RabbitMQServerResource> builder, int? port = null)
```
davidfowl, line 84: This needs to be an overload to avoid the breaking change.

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -81,6 +81,20 @@ public static IResourceBuilder<RabbitMQServerResource> WithDataBindMount(this IR
...
+    /// </summary>
+    /// <remarks>
+    /// This method only supports the default RabbitMQ container image and tags, e.g. <c>3</c>, <c>3.12-alpine</c>, <c>3.12.13-management-alpine</c>, etc.<br />
+    /// Calling this method on a resource configured with an unrecognized image registry, name, or tag will result in a <see cref="DistributedApplicationException"/> being thrown.
+    /// </remarks>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="port">The host port that can be used to access the management UI page when running locally.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    /// <exception cref="DistributedApplicationException">Thrown when the current container image and tag do not match the defaults for <see cref="RabbitMQServerResource"/>.</exception>
+    public static IResourceBuilder<RabbitMQServerResource> WithManagementPlugin(this IResourceBuilder<RabbitMQServerResource> builder, int? port)
```
eerhardt, line 97: Can you add a unit test for this new overload?

src/Aspire.Hosting.RabbitMQ/RabbitMQBuilderExtensions.cs
```diff
@@ -81,6 +81,20 @@ public static IResourceBuilder<RabbitMQServerResource> WithDataBindMount(this IR
...
+    /// Configures the RabbitMQ container resource to enable the RabbitMQ management plugin.
+    /// </summary>
+    /// <remarks>
+    /// This method only supports the default RabbitMQ container image and tags, e.g. <c>3</c>, <c>3.12-alpine</c>, <c>3.12.13-management-alpine</c>, etc.<br />
+    /// Calling this method on a resource configured with an unrecognized image registry, name, or tag will result in a <see cref="DistributedApplicationException"/> being thrown.
+    /// </remarks>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="port">The host port that can be used to access the management UI page when running locally.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    /// <exception cref="DistributedApplicationException">Thrown when the current container image and tag do not match the defaults for <see cref="RabbitMQServerResource"/>.</exception>
```
eerhardt, line 96: Can this be simplified to:

```suggestion
    /// <inheritdoc cref="WithManagementPlugin(IResourceBuilder{RabbitMQServerResource})" />
    /// <param name="builder">The resource builder.</param>
    /// <param name="port">The host port that can be used to access the management UI page when running locally.</param>
```

mitchdenny: Would also be good to add an extra `<example>` in the docs for this.

mitchdenny:
@davidfowl are you sure you want to do this? We don't do this on other things like `WithPgAdmin` or `WithRedisCommander` (for example). If we do it here should we do it for the others?

Dona278:
Hi @mitchdenny! When I added the overload I took a look at `WithPgAdmin` and `WithRedisCommander` methods and I noticed a simply difference: another resource (for example "pgadmin4" container / "redis-commander" container) is added to the builder, instead in this case we have a different images ("rabbitmq" and "rabbitmq-management") for a single resource. I initially thought to a `RabbitMQManagementResource` with `WithHostPort` but I opted for an overload method due this reason.

mitchdenny:
@Dona278 ah! good point, I keep forgetting that.

mitchdenny:
This PR will require unit tests, and XML doc comments with examples before we can move forward.

Dona278:
@mitchdenny I added a second parameter (nullable) to an existing test because it already checks all the things, port included. It's ok or do you want a specific test only for that case?

mitchdenny (approved):

jbogard:
Will this be in the 8.0.3 release? I didn't see it in the 8.0.2 drop from last week.

eerhardt:
> Will this be in the 8.0.3 release? I didn't see it in the 8.0.2 drop from last week.

No it didn't make it into 8.0.2.

We plan on releasing an 8.1 soon (likely the end of this month). This will make it into that release.

PR #4329, opened 2024-05-29
Localize "Attributes", fix nested data grid swallowing events issue

adamint:
There's a weird issue where if you have a data grid with a non-primitive item type nested inside another data grid also with a non-primitive type, the value rows (but not header) swallow click events. This manifests in being unable to copy values in the event attribute data grid. I was unable to find a root cause in fluentui-blazor or the fluent webcomponents in a reasonable amount of time. To work around this issue, we can use attribute index as the grid item.

cc @vnbaaij because I still don't quite understand why this happens

Screen recording of this branch with the workaround (working)

https://github.com/dotnet/aspire/assets/20359921/dd3e3d40-a8ca-4930-8a19-829a52a26342


tlmii (approved):
That does seem like a weird side effect of that. @vnbaaij any ideas what might be happening? But the workaround seems to work and the comment explains why we're doing it in a non-obvious way. So works for me.

vnbaaij:
Sorry, was a bit busy. No idea why that is happening. Maybe there is a `stopPropogation` in play somewhere?

PR #4331, opened 2024-05-29
Improve error handling in DistributedApplicationFactory

### LINKED ISSUE #4330: Tests hang when using xUnit and DistributedApplicationTestingBuilder if an exception is thrown in the AppHost project before the host is built
`DistributedApplicationTestingBuilder.CreateAsync` hangs forever if there's an exception thrown in the AppHost project before `Build()` is called, e.g. when attempting to add multiple resources with the same name:

```csharp

var pg = builder.AddPostgres("pg")
var db = pg.AddDatabase("pg"); // <-- Same name as the server resource above it

builder.Build().Run();
```

The Output window in Visual Studio in the "Tests" source shows just that an exception was thrown but no detail:

image

This is pretty frustrating when trying to do anything "test first" as the exception is just not visible. We should try and make the call to `CreateAsync` fail if the AppHost throws.

/Cc @ReubenBond

ReubenBond:
Fixes #4330

Please note the public API change to add `string[] args` to the test app builders.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4331)

### DIFF
--- a/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
+++ b/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
@@ -12,7 +12,10 @@ namespace Aspire.Hosting.Testing;
 /// Factory for creating a distributed application for testing.
 /// </summary>
 /// <param name="entryPoint">A type in the entry point assembly of the target Aspire AppHost. Typically, the Program class can be used.</param>
-public class DistributedApplicationFactory(Type entryPoint) : IDisposable, IAsyncDisposable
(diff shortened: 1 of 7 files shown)


DamianEdwards (approved):
Looks good from a consumption API point of view!

ReubenBond:
@mitchdenny PTAL

mitchdenny (approved):

Issue #3052, opened 2024-03-20
Add support for PgVector extension to postgres resource


davidfowl:
PGVector is a popular extension for exposing vector search capabilities in a postgres database. Today eShop hack around this by using [code like this](https://github.com/dotnet/eShop/blob/d19323582ff0f1fcea869ea1f90bb8bb69d7b780/src/eShop.AppHost/Program.cs#L9-L11)

We should build a more first class API for adding/enabling PGVector on the postgres resource.

aaronpowell:
It's a little more complex than, as you need to add the `vector` extension to the database that you're going to be storing the vectors in.

I believe that eShop does it here: https://github.com/dotnet/eShop/blob/72445fd19c85e46420a4ae4c33eb46245c970c9b/src/Catalog.API/Infrastructure/Migrations/20231009153249_Initial.cs#L15

But if you're using a non-ef database connector, say the memory pipeline in Semantic Kernel, you need to add the extension another way. My approach is to bind mount a SQL script into the init path of the Postgres container (see [this code](https://github.com/aaronpowell/HanselminutesBot/blob/main/HanselminutesBot.AppHost/Program.cs#L45)).

From a deployment perspective, we need to then ensure that we're surfacing up the metadata required for the IaC provider, with Azure and AWS we need to enable the right feature on the DB resource being deployed rather than using the container image.

davidfowl:
Isn't that automagic when you use https://github.com/dotnet/eShop/blob/72445fd19c85e46420a4ae4c33eb46245c970c9b/src/Catalog.API/Extensions/Extensions.cs#L11?

> From a deployment perspective, we need to then ensure that we're surfacing up the metadata required for the IaC provider, with Azure and AWS we need to enable the right feature on the DB resource being deployed rather than using the container image.

That should be as simple as representing it in the model is some first class way. The azure/aws postgres extension method should be able to look at the model the glean that it needs the pgvector extension (I'm not sure we need to make this super generic until we have more concrete use cases).

aaronpowell:
No, that extension method is used to enable the Vector extension in the npgsql driver for .NET - https://github.com/pgvector/pgvector-dotnet/blob/master/src/Pgvector/Npgsql/VectorExtensions.cs#L10

[...]

PR #4335, opened 2024-05-30
Enabling vector support on Azure PostgreSQL Flexible Server

aaronpowell:
Initial prototype for #3052

Since the CDK doesn't currently support configuring extensions on the PG Server resource (see [this issue](https://github.com/Azure/azure-sdk-for-net/issues/43175)), we have to generate the Bicep on the fly to do that. There's no easy hook that can be used after the original call to PublishAsAzurePostgresFlexibleServer to have it output values from the bicep, so instead we have to provide our own configureResource method.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4335)


davidfowl:
We would change the CDK to support this.

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -171,4 +171,42 @@ public static class AzurePostgresExtensions
...
+      properties: {
+        value: 'VECTOR'
+        source: 'user-override'
+      }
+    }
+    """);
+
+        builder.AsAzurePostgresFlexibleServer((resource, construct, server) =>
+        {
+            construct.AddOutput(server.AddOutput("name", data => data.Name));
```
davidfowl, line 206: This does what you want:

```suggestion
            server.AddOutput("name", data => data.Name);
```

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -171,4 +171,42 @@ public static class AzurePostgresExtensions
...
+            template.WithParameter("postgresServerName", resource.GetOutput("name"));
```
davidfowl, line 207: This being in the callback is quite strange but I see why you have to do it that way. @mitchdenny another strike against this pattern.

mitchdenny: out params :) /jk

aaronpowell:
> We would change the CDK to support this.

I do have an open PR for this on the CDK - https://github.com/Azure/azure-sdk-for-net/pull/44315

But I also wanted to look at how to unblock people who might want it sooner than those pieces become available.

davidfowl:
I understand but we're not going to take the non CDK version in this repo. That can be a sample.

eerhardt:
Is this PR stale? Can it be closed? Or are we going to take this change?

aaronpowell:
Yep, I'll close it.

I have a PR open on the CDK to get it supported in the .NET API - https://github.com/Azure/azure-sdk-for-net/pull/44315

Issue #4348, opened 2024-05-31
MongoDB Authentication


Crazyht:
For the MongoDB package, I need to be able to set an initial Username and Password to enable authentication.

This can be done by using MONGO_INITDB_ROOT_USERNAME and MONGO_INITDB_ROOT_PASSWORD environment variables.

This parameter must also be passed to mongo-express and used in the connection string.

davidfowl:
This isn't supported today for mongo, but if you're interested feel free to send a PR with this support. You can model it after the postgres resource.

Crazyht:
PR request build failed, but I can see why here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=696252&view=logs&j=22fea640-1099-5f32-ec5d-316ad83f359a&t=25461519-4e76-58ab-2938-2b66c09712e4&l=23

Aspire.Azure.Data.Tables are not touched by my commit, so this test failed for probably a good reason, but I don't see which one :(

Issue #3117, opened 2024-03-23
Usernames and passwords in connection strings need to be escaped


DamianEdwards:
While investigating #2247 I used a parameter to create a stable password for a RabbitMQ resource, only to find that the password I'd generated contained a character that broke the connection string. RabbitMQ connection strings are actually URIs, so if the username or password contains restricted characters (e.g. `@`) then the RabbitMQ client fails due to the URI being invalid.

I updated `ReferenceExpression` to support custom value escaping and a convenience method for getting a URI escaped version of an existing `ReferenceExpression` and that solves the issue for runtime, but this issue will occur during manifest processing too. Resolving that is a much bigger challenge that will need some thought.

eerhardt:
I think in order to be able to solve this generically, and in all the places we need to fix it, we will need to:

1. Have a set of "well known" escape strategies/algorithms.
2. Encode which escape strategy to use inside the `ReferenceExpression`, so it gets written to the manifest.

RabbitMQ uses URIs, but PostgreSQL uses DbConnectionString. To properly escape any string in a DbConnectionString, we would need to add single quotes `'` around the value, and then escape any real single quotes in the value by doubling them: `''`.

So at a minimum we would need 2 strategies: URI escaping and single-quote escaping.

As you pointed out above, we would need to put this in the manifest and have any deployment tools (like `azd`) understand these strategies.

cc @davidfowl @mitchdenny @ellismg @vhvb1989

DamianEdwards:
So we'd represent this in the manifest something like this?
`"amqp://{urlencode(resources.rabbitusr.value)}:{urlencode(resources.rabbitpw.value)}@{resources.rabbitmq.bindings.tcp.host}:{resources.rabbitmq.bindings.tcp.port}"`

eerhardt:
> So we'd represent this in the manifest something like this? `"amqp://{urlencode(resources.rabbitusr.value)}:{urlencode(resources.rabbitpw.value)}@{resources.rabbitmq.bindings.tcp.host}:{resources.rabbitmq.bindings.tcp.port}"`

Yep - something like that is what I was thinking.

DamianEdwards:
@davidfowl we thinking punt this until post-GA now? Or do we think this will be common enough when considering values in URIs and connection strings that we need to mitigate it earlier?

davidfowl:
[...]

PR #4370, opened 2024-06-03
Add authentication (username & password) to AddMongoDB

Crazyht:
Please ensure to include optional authentication parameters, specifically a username and password, when utilizing the AddMongoDB function. In situations where these parameters are not explicitly defined, the default username "admin" will be utilized, and a password will be automatically generated. #4348 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4370)


Crazyht:
How I can solve this test failure, in connection string password should be data-encoded, so I can't directly check "{mongodb-password.value}". 
Have you already a mechanism somewhere I can inspire of ? The only possibility I see is to force password like "aspire123" in test suit. Do you think is a good way to handle it ?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -15,22 +17,39 @@ public static class MongoDBBuilderExtensions
     // Internal port is always 27017.
     private const int DefaultContainerPort = 27017;
 
+    private const string UserEnvVarName = "MONGO_INITDB_ROOT_USERNAME";
+    private const string PasswordEnvVarName = "MONGO_INITDB_ROOT_PASSWORD";
+    //MONGO_INITDB_DATABASE ???
```
davidfowl, line 22: Delete?
```suggestion
```

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -6,24 +6,53 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// A resource that represents a MongoDB container.
 /// </summary>
-/// <param name="name">The name of the resource.</param>
-public class MongoDBServerResource(string name) : ContainerResource(name), IResourceWithConnectionString
+public class MongoDBServerResource : ContainerResource, IResourceWithConnectionString
```
davidfowl, line 9: This is a breaking API change, we need to preserve the existing overload that takes a name.

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -6,24 +6,53 @@ namespace Aspire.Hosting.ApplicationModel;
...
+            ReferenceExpression.Create($"{UserNameParameter}") :
+            ReferenceExpression.Create($"{DefaultUserName}");
+
     /// <summary>
     /// Gets the connection string for the MongoDB server.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
         ReferenceExpression.Create(
-            $"mongodb://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+            $"mongodb://{UserNameReference}:{Uri.EscapeDataString(PasswordParameter.Value)}@{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
```
davidfowl, line 55: Remove the `Uri.EscapeDataString`. This needs to be solved here https://github.com/dotnet/aspire/issues/3117

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -634,10 +634,23 @@ public void VerifyTestProgramFullManifest()
...
+                  "inputs": {
+                    "value": {
+                      "type": "string"
+                    }
+                  }
+                },
                 "mongodb": {
                   "type": "container.v0",
-                  "connectionString": "mongodb://{mongodb.bindings.tcp.host}:{mongodb.bindings.tcp.port}",
+                  "connectionString": "mongodb://admin:aspire123@{mongodb.bindings.tcp.host}:{mongodb.bindings.tcp.port}",
```
mitchdenny, line 648: This should be referencing the parameter ... we should never have a literal secret here.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -15,22 +17,39 @@ public static class MongoDBBuilderExtensions
...
     /// Adds a MongoDB resource to the application model. A container is used for local development. This version the package defaults to the 7.0.8 tag of the mongo container image.
     /// </summary>
     /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
     /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
     /// <param name="port">The host port for MongoDB.</param>
+    /// <param name="userName">A parameter that contains the MongoDb server user name, or <see langword="null"/> to use a default value.</param>
+    /// <param name="password">A parameter that contains the MongoDb server password, or <see langword="null"/> to use a generated password.</param>
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<MongoDBServerResource> AddMongoDB(this IDistributedApplicationBuilder builder, string name, int? port = null)
+    public static IResourceBuilder<MongoDBServerResource> AddMongoDB(this IDistributedApplicationBuilder builder,
```
eerhardt, line 33: This is also a binary breaking change.

tests/testproject/TestProject.AppHost/TestProgram.cs
```diff
@@ -120,8 +121,9 @@ public class TestProgram : IDisposable
             }
             if (!resourcesToSkip.HasFlag(TestResourceNames.mongodb))
             {
+                var mongoPassword = builder.AddParameter("mongodb-password");
                 var mongoDbName = "mymongodb";
-                var mongodb = AppBuilder.AddMongoDB("mongodb")
+                var mongodb = AppBuilder.AddMongoDB("mongodb", password: mongoPassword)
```
eerhardt, line 126: Existing tests shouldn't be modified. They should stay working after your change.

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
Crazyht: Is it better to overload the constructor or add new parameters with default to null?

tests/testproject/TestProject.AppHost/aspire-manifest.json
```diff
@@ -262,7 +262,7 @@
     },
     "mongodb": {
       "type": "container.v0",
-      "connectionString": "mongodb://{mongodb.bindings.tcp.host}:{mongodb.bindings.tcp.port}",
+      "connectionString": "mongodb://admin:aspire123@{mongodb.bindings.tcp.host}:{mongodb.bindings.tcp.port}",
```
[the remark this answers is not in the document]
Crazyht: Resolved by removing the Uri.EscapeDataString

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
Crazyht: Is it better to add a new overload with new parameters?

tests/testproject/TestProject.AppHost/TestProgram.cs
Crazyht: Resolved by removing the Uri.EscapeDataString

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -6,53 +6,48 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// A resource that represents a MongoDB container.
 /// </summary>
-public class MongoDBServerResource : ContainerResource, IResourceWithConnectionString
+/// <param name="name">The name of the resource.</param>
+/// /// <param name="userNameParameter">A parameter that contains the MongoDb server user name, or <see langword="null"/> to use a default value.</param>
+/// <param name="passwordParameter">A parameter that contains the MongoDb server password.</param>
+public class MongoDBServerResource(string name, ParameterResource? userNameParameter = null, ParameterResource? passwordParameter = null) : ContainerResource(name), IResourceWithConnectionString
```
davidfowl, line 12: Is this also a breaking change?

Crazyht: So, must I create an overload with new parameters instead of named parameters with defaults?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
mitchdenny: I'm not sure if we can avoid the breaking change here. If we add another overload with optional arguments how would you disambiguate between it and the new overload with the optional password arguments?

Crazyht: I agree; otherwise, maybe the best option will be to get back to the original implementation and add an extension method of IResourceBuilder<MongoDBServerResource> (like "WithAuthentication(user?, pwd?)" ) to enable authentication for the MongoDB container. With this approach, we don't have any break changes, and we activate authentication only when the method is called (even if I wonder if running db without authentication is part of good practice:) ). Do you think this approach is more aligned?

davidfowl: I think we should switch mongodb to always have a password (runtime breaking change in the name of security). The key here is to avoid removing the original method that will result in a method missing exception for those that compiled against the old one. To add a new method, we need the old method to make all of the other arguments non-optional and a new method with all of the parameters.

```C#
public static IResourceBuilder< MongoDBServerResource> AddMongoDb(this IDistributedApplicationBuilder builder, string name, int? port);
public static IResourceBuilder< MongoDBServerResource> AddMongoDb(this IDistributedApplicationBuilder builder, string name, int? port = null, IResourceBuilder<ParameterResource>? userName = null, IResourceBuilder<ParameterResource>? password = null);
```

Crazyht: @davidfowl I'm okay with switching MongoBd secure by default, but in this case, I can't be fully secure without breaking change on MongoDBServerResource. Otherwise, someone can use the primary constructor with name only and default to MongoDBServerResource without username and password, even if I think nobody tries to create a MongoDBServerResource instance by hand. Let me know if you prefer I secure MongoDBServerResource also or if I code the version without breaking changes.

Crazyht:
May I put a specific comment on `var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password", special: false);` to specify the "special: false" should be removed/reviewed when #3117 is done?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
davidfowl: If you manually create the resource, yes you're right. We can obsolete the constructor with no arguments and remove it in a later release. These 2 methods will always set the password.

Crazyht:
I need to mark the MongoDBServerResource(string name) constructor as obsolete. I'm not happy with the Connection string management (due to the database connection string needing to insert the DB name in the middle of the URL).

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -10,20 +10,95 @@ namespace Aspire.Hosting.ApplicationModel;
...
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression =>
-        ReferenceExpression.Create(
-            $"mongodb://{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{ConnectionStringWithoutOptionsExpression}{ConnectionStringOptionsExpression}");
+
+    /// <summary>
+    /// Gets the connection string for the MongoDB server without options parameters.
+    /// </summary>
+    public ReferenceExpression ConnectionStringWithoutOptionsExpression
```
davidfowl, line 70: Do these need to be public?

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -10,20 +10,95 @@ namespace Aspire.Hosting.ApplicationModel;
...
+            {
+                return ReferenceExpression.Create($"mongodb://{UserNameReference}:{PasswordParameter}@{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+            }
+        }
+    }
+
+    /// <summary>
+    /// Gets the options parameters for connection string of the MongoDB server.
+    /// </summary>
+    public ReferenceExpression ConnectionStringOptionsExpression
```
davidfowl, line 88: Does this need to be public?

tests/testproject/TestProject.AppHost/TestProgram.cs
```diff
@@ -51,7 +51,7 @@ public class TestProgram : IDisposable
         builder.Configuration["DcpPublisher:DeleteResourcesOnShutdown"] = "true";
         builder.Configuration["DcpPublisher:ResourceNameSuffix"] = $"{Random.Shared.Next():x}";
         builder.Configuration["DcpPublisher:RandomizePorts"] = randomizePorts.ToString(CultureInfo.InvariantCulture);
-
+
```
davidfowl, line 54:
```suggestion
```

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -10,20 +10,95 @@ namespace Aspire.Hosting.ApplicationModel;
...
     private EndpointReference? _primaryEndpoint;
 
+    /// <summary>
+    /// Initialize a resource that represents a MongoDB container.
+    /// </summary>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="userNameParameter">A parameter that contains the MongoDb server user name, or <see langword="null"/> to use a default value.</param>
+    /// <param name="passwordParameter">A parameter that contains the MongoDb server password.</param>
+    /// <param name="authenticationMechanism">A parameter that contains the MongoDb server authentication mechanism, or <see langword="null"/> to use a default value.</param>
+    public MongoDBServerResource(string name, ParameterResource? userNameParameter, ParameterResource? passwordParameter, ParameterResource? authenticationMechanism) : this(name)
```
eerhardt, line 26:
```suggestion
    public MongoDBServerResource(string name, ParameterResource? userNameParameter, ParameterResource passwordParameter, ParameterResource? authenticationMechanism) : this(name)
```

The rest of the resources that take a password parameter aren't nullable.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -15,22 +17,57 @@ public static class MongoDBBuilderExtensions
...
+    /// <param name="userName">A parameter that contains the MongoDb server user name, or <see langword="null"/> to use a default value.</param>
+    /// <param name="password">A parameter that contains the MongoDb server password, or <see langword="null"/> to use a generated password.</param>
+    /// <param name="authenticationMechanism">A parameter that contains the MongoDb server authentication mechanism, or <see langword="null"/> to use a default value.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<MongoDBServerResource> AddMongoDB(this IDistributedApplicationBuilder builder,
+        string name,
+        int? port = null,
+        IResourceBuilder<ParameterResource>? userName = null,
+        IResourceBuilder<ParameterResource>? password = null,
+        IResourceBuilder<ParameterResource>? authenticationMechanism = null)
```
eerhardt, line 50: Do we need this authentication mechanism as a public option? I don't see it in other resources - ex. PostgreSQL.

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -10,20 +10,95 @@ namespace Aspire.Hosting.ApplicationModel;
...
+    /// Gets the parameter that contains the MongoDb server password.
+    /// </summary>
+    public ParameterResource? PasswordParameter { get; }
+
+    /// <summary>
+    /// Gets the parameter that contains the MongoDb server authentication mechanism.
+    /// </summary>
+    public ParameterResource? AuthenticationMechanismParameter { get; }
+
+    internal ReferenceExpression AuthenticationMechanismReference =>
```
eerhardt, line 48:
```suggestion
    private ReferenceExpression AuthenticationMechanismReference =>
```

tests/Aspire.Hosting.Tests/MongoDB/AddMongoDBTests.cs
```diff
@@ -2,10 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Net.Sockets;
+
 using Aspire.Hosting.MongoDB;
 using Aspire.Hosting.Tests.Utils;
 using Aspire.Hosting.Utils;
+
 using Microsoft.Extensions.DependencyInjection;
+
```
eerhardt, line 11:
```suggestion
using Aspire.Hosting.MongoDB;
using Aspire.Hosting.Tests.Utils;
using Aspire.Hosting.Utils;
using Microsoft.Extensions.DependencyInjection;
```

These newlines are unnecessary.

tests/testproject/TestProject.AppHost/TestProgram.cs
```diff
@@ -51,7 +51,6 @@ public class TestProgram : IDisposable
         builder.Configuration["DcpPublisher:DeleteResourcesOnShutdown"] = "true";
         builder.Configuration["DcpPublisher:ResourceNameSuffix"] = $"{Random.Shared.Next():x}";
         builder.Configuration["DcpPublisher:RandomizePorts"] = randomizePorts.ToString(CultureInfo.InvariantCulture);
-
```
eerhardt, line 54: This can be reverted as it is unnecessary.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
davidfowl: We don't

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -15,22 +17,57 @@ public static class MongoDBBuilderExtensions
...
+        var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password", special:false);
+
+        var mongoDBContainer = new MongoDBServerResource(name, userName?.Resource, passwordParameter, authenticationMechanism?.Resource);
+
+        var resBuilder = builder
             .AddResource(mongoDBContainer)
             .WithEndpoint(port: port, targetPort: DefaultContainerPort, name: MongoDBServerResource.PrimaryEndpointName)
             .WithImage(MongoDBContainerImageTags.Image, MongoDBContainerImageTags.Tag)
             .WithImageRegistry(MongoDBContainerImageTags.Registry);
+        if (mongoDBContainer.PasswordParameter is not null)
```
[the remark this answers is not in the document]
Crazyht: Not in a manual crafting scenario, as I only mark MongoDBServerResource ctor obsolete, so we can't completely avoid null for extension. This test should be removed only when the obsolete MongoDBServerResource constructor is also removed.

src/Aspire.Hosting.MongoDB/PublicAPI.Shipped.txt
```diff
@@ -8,12 +8,16 @@ Aspire.Hosting.ApplicationModel.MongoDBServerResource
 Aspire.Hosting.ApplicationModel.MongoDBServerResource.ConnectionStringExpression.get -> Aspire.Hosting.ApplicationModel.ReferenceExpression!
 Aspire.Hosting.ApplicationModel.MongoDBServerResource.Databases.get -> System.Collections.Generic.IReadOnlyDictionary<string!, string!>!
 Aspire.Hosting.ApplicationModel.MongoDBServerResource.MongoDBServerResource(string! name) -> void
+Aspire.Hosting.ApplicationModel.MongoDBServerResource.MongoDBServerResource(string! name, Aspire.Hosting.ApplicationModel.ParameterResource? userNameParameter, Aspire.Hosting.ApplicationModel.ParameterResource? passwordParameter) -> void
```
eerhardt, line 11: We haven't shipped these APIs yet, so they should go in the Unshipped.txt file.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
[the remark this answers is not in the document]
eerhardt: > Not in a manual crafting scenario

This isn't the manual crafting scenario. We directly create the object on line 52 above and we always pass in a non-null `passwordParameter`.

tests/testproject/TestProject.AppHost/TestProgram.cs
eerhardt, line 54: Can you revert this change so it doesn't go into source control history?

davidfowl:
@eerhardt this should follow the persistence pattern we just added

eerhardt:
> @eerhardt this should follow the persistence pattern we just added

It does already.

davidfowl:
I think we need to test a deployment to ACA as well to make sure it still works.

eerhardt:
Thanks @Crazyht for your contribution. I'd like to get this change in for .NET Aspire 9.0, but I'm unable to push to this branch because it is from a protected branch (your `main` branch).

So instead I created https://github.com/dotnet/aspire/pull/5788, which takes these changes, resolves the conflicts with the current code, and fixes the tests. Please take a look over there.

Closing in favor of https://github.com/dotnet/aspire/pull/5788.

PR #4374, opened 2024-06-04
Add new DCP build context schema

danegsta:
Adds the schema for the newly added "secrets" and "tags" DCP image build context properties.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4374)

### DIFF
--- a/src/Aspire.Hosting/Dcp/Model/Container.cs
+++ b/src/Aspire.Hosting/Dcp/Model/Container.cs
@@ -68,9 +68,26 @@ internal sealed class BuildContext
     [JsonPropertyName("args")]
     public List<EnvVar>? Args { get; set; }
 
+    // Optional build secret mounts to pass to the build command
+    [JsonPropertyName("secrets")]
+    public List<BuildContextSecret>? Secrets { get; set; }
+
     // Optional specific stage to use when building a multiple stage Dockerfile
     [JsonPropertyName("stage")]
     public string? Stage { get; set; }
+
+    // Optional additional tags to apply to the built image
+    [JsonPropertyName("tags")]
+    public List<string>? Tags { get; set; }
+}
+
+internal sealed class BuildContextSecret
+{
+    [JsonPropertyName("id")]
+    public string? Id { get; set; }
+
+    [JsonPropertyName("source")]
+    public string? Source { get; set; }
 }
 
 internal static class VolumeMountType
--- a/src/Aspire.Hosting/Dcp/Model/Container.cs
+++ b/src/Aspire.Hosting/Dcp/Model/Container.cs
@@ -83,9 +83,11 @@ internal sealed class BuildContext
 
 internal sealed class BuildContextSecret
 {
+    // The ID of the secret (a secret can be used in a Dockerfile with `RUN --mount-type=secret,id=<id>,target=<targetpath>`)
     [JsonPropertyName("id")]
     public string? Id { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting/Dcp/Model/Container.cs
```diff
@@ -68,9 +68,26 @@ internal sealed class BuildContext
...
+    public List<string>? Tags { get; set; }
+}
+
+internal sealed class BuildContextSecret
+{
+    [JsonPropertyName("id")]
+    public string? Id { get; set; }
+
+    [JsonPropertyName("source")]
+    public string? Source { get; set; }
```
mitchdenny, line 90: What do we put in the `Source` field? Do we include the type?

src/Aspire.Hosting/Dcp/Model/Container.cs
```diff
@@ -68,9 +68,26 @@ internal sealed class BuildContext
...
     public string? Stage { get; set; }
+
+    // Optional additional tags to apply to the built image
+    [JsonPropertyName("tags")]
+    public List<string>? Tags { get; set; }
+}
+
+internal sealed class BuildContextSecret
+{
+    [JsonPropertyName("id")]
```
karolz-ms, line 86:
```suggestion
    // The ID of the secret
    [JsonPropertyName("id")]
```

src/Aspire.Hosting/Dcp/Model/Container.cs
```diff
@@ -68,9 +68,26 @@ internal sealed class BuildContext
...
+    public string? Id { get; set; }
+
+    [JsonPropertyName("source")]
```
karolz-ms, line 89:
```suggestion
    // The source filepath of the secret
    [JsonPropertyName("source")]
```

karolz-ms (approved):

src/Aspire.Hosting/Dcp/Model/Container.cs
danegsta: Added a comment

src/Aspire.Hosting/Dcp/Model/Container.cs
danegsta: Added an informative comment

src/Aspire.Hosting/Dcp/Model/Container.cs
danegsta: Added a comment to explain that this is a source filepath (same as source in a bind volume mount).

mitchdenny: @danegsta I think we need to talk through this bit a little bit more because I'm not quite sure how it is going to practically work. Consider the following code:

```csharp

var nugettoken = builder.AddParameter("nugettoken", secret: true);

var container = builder.AddContainer("mycontainer", "myimage")
                       .WithDockerfile(contextPath)
                       .WithSecretBuildArg("nameOfSecret", nugettoken);
```

It seems to me that the source in the DCP Container model above is expecting a path to a file.  I think that we'll probably just want to give DCP the literal value and have it manage whether it either sets an environment variable to use with `--secret` or temporarily writes the secret value to the file.

mitchdenny: I think setting an ENV when you launch docker build would be preferrable to avoid any potential leakage to the filesystem.

Issue #958, opened 2023-11-20
UI tests for the dashboard 


Meir017:
I think we should start adding UI tests for the dashboard. 
We could use Microsoft.Playwright and integrate with the TestServer by implementing a route fir redirecting the browser's traffic to the in memory TestServer

leslierichardson95:
@Meir017 are you referring to UI tests for the dashboard itself or UI tests for the actual UI components that make up your Aspire app?

danmoseley:
@kvenkatrajan shall we use this issue for getting 1 Playwright test running and checked in this week?

kvenkatrajan:
#4386 - in review

PR #4386, opened 2024-06-04
Add sample dashboard playwright test

adamint:
Runs dashboard with a mocked dashboard client, adds example playwright test, as we will be adding many more dashboard playwright tests using the same infrastructure in the future.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4386)


Meir017:
related to #958

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,69 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Playwright;
+using Polly;
+using Polly.Retry;
+using Xunit;
+
+namespace Aspire.Dashboard.Tests.Integration.Playwright;
+
+public class PlaywrightFixture : IAsyncLifetime
```
Meir017, line 11: https://playwright.dev/dotnet/docs/test-runners#xunit-support It might be better use one of the officially supported testing-libraries for playwright (MSTest or NUnit https://playwright.dev/dotnet/docs/test-runners#xunit-support)

adamint: I would definitely prefer to use MSTest or NUnit, but I do want to balance that against other concerns that
- this would be the only test project not using xUnit, and
- there are other playwright tests using xUnit

@danmoseley Could we get your opinion on this?

Meir017: @mxschmitt whats your opinion here?

radical: > 
> > there are other playwright tests using xUnit
> 
> Is this referring to [that](https://github.com/dotnet/aspire/blob/521a2c63cbcc3dcbc5d93c99d8c6fbc3a0933b39/tests/Aspire.Workload.Tests/WorkloadTestsBase.cs#L26)?

Yes. And the Workload tests are using Playwright in (b) mode from your comment. We `PackageReference` `"Microsoft.Playwright"` nuget package, and use that.

adamint: Yes, the concern is whether the benefits of switching to MSTest or nUnit outweigh the cost of supporting a singular test project not built with xUnit.

danmoseley: > @danmoseley Could we get your opinion on this?

I don't have much to add: Aspire builds with reusable infrastructure from dotnet/arcade, which supports xunit, as the .NET platform repos picked xunit long ago. I do not see anything in there for mstest (as a test platform) or nunit so use of those would presumably have to be set up custom in this repo.

I defer to @radical/@joperezr

radical: I'm not sure what additional work would be needed on the CI to support that - on the build machine, and helix. Also, arcade provides many test attributes for xunit like `ActiveIssue` which we use to skip tests conditionally. We also have `RequiresDocker` which is xunit based too.

This would also mean additional documentation, and providing ways to have similar functionality for skipping tests etc.

IMHO, sticking to xunit would be better as it would utilize all the existing infrastructure.

mxschmitt: Sounds reasonable! As long as you provide a base class which gives them a browser context and test per test it should not be too bad of a compromise. Also make sure to add an expect method like [here](https://github.com/microsoft/playwright-dotnet/blob/86ad6c744350f97b01eba68dbb3b4c59efc9caf4/src/Playwright.MSTest/PlaywrightTest.cs#L129).

adamint:
Could I get a review on this?

radical:
Build failing with: `tests/Aspire.Dashboard.Tests/Integration/Playwright/DashboardServerFixture.cs(30,35): error CS0117: (NETCORE_ENGINEERING_TELEMETRY=Build) 'DashboardConfigNames' does not contain a definition for 'DashboardOtlpUrlName'`

adamint:
> DashboardOtlpUrlName

Sorry, @radical. That property name was seemingly changed and I did not notice after merge. It builds now 😄

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    public async Task InitializeAsync()
+    {
+        var installExitCode = Microsoft.Playwright.Program.Main(new[] {"install"});
+        if (installExitCode != 0)
+        {
+            throw new PlaywrightException($"Playwright exited with code {installExitCode}");
+        }
+
+        var installDepsExitCode = Microsoft.Playwright.Program.Main(new[] {"install-deps"});
```
radical, line 26: On the linux (ubuntu) build machine `install-deps` will install a whole bunch of packages with `apt-get`. Because this is a shared machine, we want to avoid installing packages on it.

The existing infrastructure for installing playwright available is partially broken right now due to package installation issues (https://github.com/dotnet/aspire/issues/4623). And these tests are likely to run into the same issue randomly.

So, I would suggest not running the Playwright based tests on linux for now. You can add this to your test - `[ActiveIssue("https://github.com/dotnet/aspire/issues/4623", TestPlatforms.Linux)]`.

Going forward:
1. https://github.com/dotnet/aspire/pull/4689 - this will make the existing playwright targets reusable
2. Once https://github.com/dotnet/dnceng/issues/3122 is resolved, we can drop even this code.
Neither of these need to block your PR though.

radical: Also, could you please open an issue to track that the code here needs to be adjusted once https://github.com/dotnet/aspire/pull/4689 is merged, and https://github.com/dotnet/dnceng/issues/3122 is fixed?

radical (dismissed):

tests/Aspire.Dashboard.Tests/Integration/Playwright/DashboardServerFixture.cs
```diff
@@ -0,0 +1,63 @@
...
+                builder.Services.RemoveAll<IDashboardClient>();
+                builder.Services.AddSingleton<IDashboardClient, MockDashboardClient>();
+            });
+
+        await DashboardApp.StartAsync();
+    }
+
+    public async Task DisposeAsync()
+    {
+        await DashboardApp.DisposeAsync();
```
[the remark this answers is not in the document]
adamint: that returns a ValueTask, I changed to return `DashboardApp.DisposeAsync().AsTask()`

tests/Aspire.Dashboard.Tests/Integration/Playwright/AppBarTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        await SetAndVerifyTheme(darkThemeCheckbox, "dark");
+        await SetAndVerifyTheme(lightThemeCheckbox, "light");
+
+        return;
+
+        async Task SetAndVerifyTheme(ILocator locator, string expected)
+        {
+            await locator.ClickAsync();
+            await Task.Delay(500);
```
mxschmitt, line 40: Remove the delay and use our assertion which retries? https://playwright.dev/dotnet/docs/api/class-locatorassertions#locator-assertions-to-have-attribute

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
mxschmitt: Usually we strongly recommend doing install and --with-deps from the outside. Since these require sudo and modify the system. These should not delay the test execution and skipping it on Linux seems like showing the drawbacks for it.

lets call it from the CI yaml instead?

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public async Task DisposeAsync()
+    {
+        await Browser.CloseAsync();
+    }
+
+    public async Task GoToHomeAndWaitForDataGridLoad(IPage page)
+    {
+        await page.GotoAsync("/");
```
mxschmitt, line 45: I recommend page -> Page like we have in our base classes.

```suggestion
        await Page.GotoAsync("/");
```

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        await _resiliencePipeline.ExecuteAsync(async _ =>
```
mxschmitt, line 47: Is this like a custom retry strategy? I recommend to e.g. stick with ToBeVisible which will retry and end up in better reporting in e.g. Trace Viewer or logs.

https://playwright.dev/dotnet/docs/api/class-locatorassertions#locator-assertions-to-be-visible

adamint:
> Usually we strongly recommend doing install and --with-deps from the outside. Since these require sudo and modify the system. These should not delay the test execution and skipping it on Linux seems like showing the drawbacks for it.

> lets call it from the CI yaml instead?

This code was removed.

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
adamint: As a method parameter, that does not follow naming conventions. If `page` were a property, I would agree

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
adamint: Thanks, I've made this change 😄

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
mxschmitt: I see, thought its a property on the class where page gets created in TestInitialize etc. Feel free to ignore then.

danmoseley:
Thanks @adamint . Does any of my feedback from your original PR apply here? (IDK)

tests/Aspire.Dashboard.Tests/Aspire.Dashboard.Tests.csproj
```diff
@@ -6,12 +6,15 @@
...
+
     <ProjectReference Include="..\..\src\Aspire.Dashboard\Aspire.Dashboard.csproj" />
 
     <Compile Include="$(TestsSharedDir)TestCertificateLoader.cs" Link="shared/TestCertificateLoader.cs" />
     <Compile Include="$(TestsSharedDir)Logging\*.cs" LinkBase="shared/Logging" />
     <Content Include="$(TestsSharedDir)TestCertificates\*.pfx" LinkBase="shared/TestCertificates" CopyToOutputDirectory="PreserveNewest" />
 
     <Content Include="..\..\src\Aspire.Dashboard\wwwroot\**\*.*" LinkBase="wwwroot" CopyToOutputDirectory="PreserveNewest" />
+
+    <ProjectReference Include="..\Aspire.EndToEnd.Tests\Aspire.EndToEnd.Tests.csproj" />
```
radical, line 18: If this was added for getting `PlaywrightProvider`, then instead use `<Import Project="..\Shared\Playwright\Playwright.targets" />`

tests/Aspire.Dashboard.Tests/Integration/Playwright/AppBarTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    public AppBarTests(DashboardServerFixture dashboardServerFixture, PlaywrightFixture playwrightFixture)
+        : base(dashboardServerFixture, playwrightFixture)
+    {
+    }
+
+    [Fact]
+    public async Task AppBar_Change_Theme()
+    {
+        // Arrange
+        await RunTestAsync(async page =>
```
radical, line 23: nit: this can return `RunTestAsync(..)` directly.

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightTestsBase.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+        finally
+        {
+            await page.CloseAsync();
+        }
+    }
+
+    private async Task<IPage> CreateNewPageAsync()
+    {
+        _context ??= await PlaywrightFixture.Browser.NewContextAsync(new BrowserNewContextOptions
```
radical, line 37: This will be using a single context for all the tests in a class. Is that intentional?

radical: nvm, that makes sense at least for a single class at least.

tests/Aspire.Dashboard.Tests/Integration/Playwright/PlaywrightFixture.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Aspire.Dashboard.Tests.Integration.Playwright;
+
+public class PlaywrightFixture : IAsyncLifetime
+{
+    public IBrowser Browser { get; set; } = null!;
+
+    public async Task InitializeAsync()
+    {
+        PlaywrightProvider.DetectAndSetInstalledPlaywrightDependenciesPath();
+        Browser = await PlaywrightProvider.CreateBrowserAsync();
```
radical, line 17: This would create the browser instance, and launch the browser for every test class using this fixture. For workload tests we use:
https://github.com/adamint/aspire/blob/64ea4ab15976fbf903c9663aa6982205be8b1e2f/tests/Aspire.Workload.Tests/WorkloadTestsBase.cs#L22-L34
and `private static Lazy<IBrowser> Browser => new(CreateBrowser);`

radical:
Screenshot 2024-06-28 at 20 43 46

And the test was skipped was linux.

radical:
@adamint I merged the base playwright changes in a separate PR, and update this one to fix the merge conflicts.

adamint:
> @adamint I merged the base playwright changes in a separate PR, and update this one to fix the merge conflicts.

Is anything else needed here? @radical 
thank you 😄 

@danmoseley not that I'm aware of. The recommendation is to avoid assuming the structure of the page, so locating elements by text instead of css classes is preferred

radical (approved):
The infra bits LGTM 👍

PR #4404, opened 2024-06-06
Set HTTP/HTTPS_PORTS on projects in publish mode

### LINKED ISSUE #3749: Set HTTP/HTTPS_PORTS on projects in publish mode by default
Today if you define multiple endpoints on a project resource, none will be bound to ASPNETCORE_URLs in publish mode. We're currently relying on the base container image to do this. This causes a difference in behavior in run mode vs publish mode that creates confusion.

davidebbo:
- Only do this for explicit endpoints, not from profile
- Fixes #3749
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4404)

### DIFF
--- a/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
@@ -349,6 +349,13 @@ public static class ProjectResourceBuilderExtensions
                     {
                         e.UriScheme = scheme;
                         e.Transport = adjustTransport(e);
+
+                        // In the https case, we don't want this default endpoint to end up in the HTTPS_PORTS env var,
+                        // because the container likely won't be set up to listen on https (e.g. ACA case)
+                        if (scheme == "https")
+                        {
+                            e.ExcludeFromPortEnvironment = true;
+                        }
                     },
                     createIfNotExists: true);
                 }
@@ -514,7 +521,7 @@ public static class ProjectResourceBuilderExtensions
         // Turn endpoint ports into a single environment variable
         foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
         {
-            if (e.EndpointAnnotation.UriScheme == scheme)
+            if (e.EndpointAnnotation.UriScheme == scheme && !e.EndpointAnnotation.ExcludeFromPortEnvironment)
             {
                 Debug.Assert(!e.EndpointAnnotation.FromLaunchProfile, "Endpoints from launch profile should never make it here");
 
--- a/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
@@ -514,8 +514,7 @@ public static class ProjectResourceBuilderExtensions
         // Turn endpoint ports into a single environment variable
         foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
         {
-            var port = e.EndpointAnnotation.TargetPort ?? e.EndpointAnnotation.Port;
-            if (e.EndpointAnnotation.UriScheme == scheme && port != null)
+            if (e.EndpointAnnotation.UriScheme == scheme && e.EndpointAnnotation.TargetPort != null)
             {
                 Debug.Assert(!e.EndpointAnnotation.FromLaunchProfile, "Endpoints from launch profile should never make it here");
 
--- a/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics;
-using System.Text;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Dashboard;
 using Aspire.Hosting.Utils;
@@ -445,6 +444,9 @@ public static class ProjectResourceBuilderExtensions
         return configBuilder.Build();
     }
 
+    static bool IsValidAspNetCoreUrl(EndpointAnnotation e) =>
+        e.UriScheme is "http" or "https" && e.TargetPortEnvironmentVariable is null;
+
     private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> builder)
     {
         builder.WithEnvironment(context =>
@@ -459,9 +461,6 @@ public static class ProjectResourceBuilderExtensions
 
             var processedHttpsPort = false;
             var first = true;
-
-            static bool IsValidAspNetCoreUrl(EndpointAnnotation e) =>
-                e.UriScheme is "http" or "https" && e.TargetPortEnvironmentVariable is null;
 
             // Turn http and https endpoints into a single ASPNETCORE_URLS environment variable.
             foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
@@ -509,30 +508,30 @@ public static class ProjectResourceBuilderExtensions
             return;
         }
 
-        var ports = new StringBuilder();
+        var ports = new ReferenceExpressionBuilder();
         var firstPort = true;
 
         // Turn endpoint ports into a single environment variable
-        foreach (var annotation in builder.Resource.Annotations.OfType<EndpointAnnotation>())
+        foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
         {
-            var port = annotation.TargetPort ?? annotation.Port;
-            if (annotation.UriScheme == scheme && port != null)
+            var port = e.EndpointAnnotation.TargetPort ?? e.EndpointAnnotation.Port;
+            if (e.EndpointAnnotation.UriScheme == scheme && port != null)
             {
-                Debug.Assert(!annotation.FromLaunchProfile, "Endpoints from launch profile should never make it here");
+                Debug.Assert(!e.EndpointAnnotation.FromLaunchProfile, "Endpoints from launch profile should never make it here");
 
                 if (!firstPort)
                 {
-                    ports.Append(';');
+                    ports.AppendLiteral(";");
                 }
-                ports.Append(port);
 
+                ports.Append($"{e.Property(EndpointProperty.TargetPort)}");
                 firstPort = false;
             }
         }
 
         if (!firstPort)
         {
-            context.EnvironmentVariables[portEnvVariable] = ports.ToString();
+            context.EnvironmentVariables[portEnvVariable] = ports.Build();
         }
     }
 }
--- a/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics;
+using System.Text;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Dashboard;
 using Aspire.Hosting.Utils;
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -324,30 +326,33 @@ private static IResourceBuilder<ProjectResource> WithProjectDefaults(this IResou
...
-                {
-                    e.UriScheme = "http";
-                    e.Transport = adjustTransport(e);
-                },
-                createIfNotExists: true);
+                return builder;
             }
 
-            if (!projectResource.Annotations.OfType<EndpointAnnotation>().Any(sb => sb.UriScheme == "https" || string.Equals(sb.Name, "https", StringComparisons.EndpointAnnotationName)))
+            string[] schemes = ["http", "https"];
```
davidebbo, line 344: This part is just a little refactoring to avoid repeating code, and not directly related to this issue

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -324,30 +326,33 @@ private static IResourceBuilder<ProjectResource> WithProjectDefaults(this IResou
         }
         else
         {
-            // If we aren't a web project (looking at both launch profile and Kestrel config) we don't automatically add bindings.
-            if (launchProfile?.ApplicationUrl == null && !kestrelEndpointsByScheme.Any())
+            // Set HTTP_PORTS/HTTPS_PORTS in publish mode, to override the default port set in the base image. Note that:
+            // - We don't set them if we have Kestrel endpoints configured, as Kestrel will get everything from its config.
+            // - We only do that for endpoint set explicitely (.WithHttpEndpoint), not for the ones coming from launch profile.
```
davidebbo, line 331: This is what we discussed with @davidfowl and @DamianEdwards, but let's make sure it's the right behavior.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -442,11 +447,6 @@ private static IConfiguration GetConfiguration(ProjectResource projectResource)
 
     private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> builder)
     {
-        if (builder.ApplicationBuilder.ExecutionContext.IsPublishMode)
```
davidebbo, line 445: This was redundant code from my last change, not related to this issue.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -491,4 +491,48 @@ private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> bui
...
+            return;
+        }
+
+        var ports = new StringBuilder();
+        var firstPort = true;
+
+        // Turn endpoint ports into a single environment variable.
+        foreach (var annotation in builder.Resource.Annotations.OfType<EndpointAnnotation>())
+        {
+            var port = annotation.TargetPort ?? annotation.Port;
```
davidebbo, line 518: Is this correct? Or should we only be looking at the TargetPort?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -491,4 +491,48 @@ private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> bui
...
+
+    private static void SetOnePortsEnvVariable(this IResourceBuilder<ProjectResource> builder, EnvironmentCallbackContext context, string portEnvVariable, string scheme)
+    {
+        if (context.EnvironmentVariables.ContainsKey(portEnvVariable))
+        {
+            // If the user has already set that variable, we don't want to override it.
+            return;
+        }
+
+        var ports = new StringBuilder();
```
davidebbo, line 512: I started out using a `ReferenceExpressionBuilder`, but found that it wasn't working in Publish mode, because it was never getting evaluated. Is there any drawback to using strings directly here, since there is no port allocation going on in publish mode (iiuc)?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Also, should we include 8080 (default container port) on the list for http, in addition to the explicit ports? As in @DamianEdwards's https://github.com/dotnet/aspire-samples/commit/b4d24f87f01d644cefb750dfc431fd02de994ad5

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: This should use `ReferenceExpressionBuilder`, we don't want ports in the manifest in the env. These should point to bindings endpoint references, not concrete ports.

tests/Aspire.Hosting.Tests/WithEndpointTests.cs
```diff
@@ -490,6 +490,66 @@ public async Task VerifyManifestProjectWithHttpEndpointDoesNotAllocatePort()
...
+            {
+              "type": "project.v0",
+              "path": "projectpath",
+              "env": {
+                "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES": "true",
+                "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES": "true",
+                "OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY": "in_memory",
+                "ASPNETCORE_FORWARDEDHEADERS_ENABLED": "true",
+                "HTTP_PORTS": "5001;5003",
+                "HTTPS_PORTS": "7001;7003"
```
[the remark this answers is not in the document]
davidebbo: Ah, that's indeed what I was getting, and I didn't realize it was desirable! I'll go back to that then...

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Done.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -491,4 +490,48 @@ private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> bui
...
+            return;
+        }
+
+        var ports = new ReferenceExpressionBuilder();
+        var firstPort = true;
+
+        // Turn endpoint ports into a single environment variable
+        foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
+        {
+            var port = e.EndpointAnnotation.TargetPort ?? e.EndpointAnnotation.Port;
```
davidfowl, line 517: delete this.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: This causes `AspNetCoreUrlsNotInjectedInPublishMode` to fail, because it calls `EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync()`, which expects endpoints to be allocated:

```
System.InvalidOperationException : The endpoint `http` is not allocated for the resource `projectName`.

EndpointReference.get_AllocatedEndpoint() line 105
EndpointReferenceExpression.ComputeTargetPort() line 194
EndpointReferenceExpression.GetValueAsync(CancellationToken cancellationToken) line 179
ReferenceExpression.GetValueAsync(CancellationToken cancellationToken) line 66
EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(IResource resource, DistributedApplicationOperation applicationOperation) line 28
ProjectResourceTests.AspNetCoreUrlsNotInjectedInPublishMode() line 276
```

But allocation only happens as part of `ApplicationExecutor.CreateContainersAndExecutablesAsync()`, which is not supposed to happen at Publish time.

One possible take is that this is not a real problem, but that it is wrong for the Publish test to try to call `EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(resource);`. It should instead generated the manifest and check off of that.

WDYT?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: So we only care about TargetPort here, and don't want to fall back to Port, right?

davidfowl: That's right.

davidebbo: Done.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -491,4 +490,47 @@ private static void SetAspNetCoreUrls(this IResourceBuilder<ProjectResource> bui
...
+            return;
+        }
+
+        var ports = new ReferenceExpressionBuilder();
+        var firstPort = true;
+
+        // Turn endpoint ports into a single environment variable
+        foreach (var e in builder.Resource.GetEndpoints().Where(e => IsValidAspNetCoreUrl(e.EndpointAnnotation)))
+        {
+            if (e.EndpointAnnotation.UriScheme == scheme && e.EndpointAnnotation.TargetPort != null)
```
davidfowl, line 517: I don't think we need a port check at all.

davidebbo: Interesting. So even if someone has `.WithHttpEndpoint(name: "hp1", port: 5001)` without a TargetPort, that should turn into an `HTTP_PORTS` entry. So I guess we're really collecting all named bindings in getting them in there? I don't have my head fully wrapped around this, but I'll take your word for it!

davidfowl: Yep.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: You shouldn't need to look at the target port value so it should work now right?

davidebbo: See the stack. The call to `EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync` results in looking up the `AllocatedEndpoint`, which blows up.

davidebbo: To be clear, this is not caused by the logic I had that was looking up TargetPort (and is now gone). It's specific to this one test.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Per @davidfowl on Discord: "No we shouldn't include 8080, null endpoint = 8080". So we're good here.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Fixed by calling `GetEnvironmentVariablesAsync` in Publish mode in the test.

tests/Aspire.Hosting.Tests/WithEndpointTests.cs
```diff
@@ -490,6 +492,39 @@ public async Task VerifyManifestProjectWithHttpEndpointDoesNotAllocatePort()
         Assert.Equal(expectedManifest, manifest.ToString());
     }
 
+    [Fact]
+    public async Task VerifyManifestProjectWithEndpointsSetsPortsEnvVariables()
```
davidfowl, line 496: I would move this to project tests

davidebbo: But then shouldn't the one above it (`VerifyManifestProjectWithHttpEndpointDoesNotAllocatePort`) be moved too? They seem similar in spirit, and they share the `TestProject` inner class.

davidfowl: Yep they should both be moved.

davidfowl: Spoke offline, I'll move these in a later refactoring.

davidfowl (approved):

PR #4407, opened 2024-06-06
Make it possible to have a different connectionName than serviceKey

egil:
It would be nice if it were possible to separate the service key name from the connection name.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4407)


src/Components/Aspire.RabbitMQ.Client/AspireRabbitMQExtensions.cs
```diff
@@ -48,18 +48,20 @@ public static class AspireRabbitMQExtensions
...
     /// <param name="name">The name of the component, which is used as the <see cref="ServiceDescriptor.ServiceKey"/> of the service and also to retrieve the connection string from the ConnectionStrings configuration section.</param>
+    /// <param name="connectionName">A name used to retrieve the connection string from the ConnectionStrings configuration section. If not specified, the <paramref name="name"/> will be used.</param>
     /// <param name="configureSettings">An optional method that can be used for customizing the <see cref="RabbitMQClientSettings"/>. It's invoked after the settings are read from the configuration.</param>
     /// <param name="configureConnectionFactory">An optional method that can be used for customizing the <see cref="ConnectionFactory"/>. It's invoked after the options are read from the configuration.</param>
     /// <remarks>Reads the configuration from "Aspire:RabbitMQ:Client:{name}" section.</remarks>
     public static void AddKeyedRabbitMQClient(
         this IHostApplicationBuilder builder,
         string name,
+        string? connectionName = null,
```
davidfowl, line 58: This is a breaking change. You need to add a new method.

eerhardt:
What's the scenario for the difference? I don't think we can just do this for just one component, we would have to do it across them all.

Instead, can you wrap it with the different serviceKey? For example with Redis:

```csharp
builder.AddKeyedRedisClient("connectionName");
builder.Services.AddKeyedSingleton("serviceKey", (sp, _) => sp.GetRequiredKeyedService<IConnectionMultiplexer>("connectionName"));
```

And then later:

```csharp
var connection = app.Services.GetKeyedService<IConnectionMultiplexer>("serviceKey");
```

egil:
> What's the scenario for the difference? I don't think we can just do this for just one component, we would have to do it across them all.

Sorry, should have specified the reason too. 

It's for readability, essentially. The name has dual meaning which can be confusing, especially if the string is coming from a constant that is referenced in different contexts, e.g.:

```c#
internal sealed class MyService
{
    internal const string ServiceKey = "MyServiceKey";
    internal const string ConnectionStringKey = "MyService";

    // const is used both when registering service and when getting the keyed service
    public MyService([FromKeyedServices(ServiceKey)] IConnectionFactory connectionFactory)
    { 
      // ...
    }
}
```

Then, in my program.cs:

```c#
// this is more readable IMO
builder.AddKeyedRabbitMQClient(
        name: MyService.ServiceKey,
        connectionName: MyService.ConnectionStringKey);

// vs. this, that requires knowledge of aspire/browsing the docs.
builder.AddKeyedRabbitMQClient(
        name: MyService.ServiceKey);
```

src/Components/Aspire.RabbitMQ.Client/AspireRabbitMQExtensions.cs
egil: fixed

egil:
> Instead, can you wrap it with the different serviceKey? For example with Redis:
> 
> ```cs
> builder.AddKeyedRedisClient("connectionName");
> builder.Services.AddKeyedSingleton("serviceKey", (sp, _) => sp.GetRequiredKeyedService<IConnectionMultiplexer>("connectionName"));
> ```
> 
> And then later:
> 
> ```cs
> var connection = app.Services.GetKeyedService<IConnectionMultiplexer>("serviceKey");
> ```

Sure, that will work, but now the service is available on two different keys, not great for maintainability.

eerhardt:
I'm not sure this has enough value to add a new overload to every component. There is an easy workaround (1 line of code). We can't just make this for 1 component and not have the same pattern across all ~30 of them.

Yes, the service is available on two different keys - but you don't have to use the connectionName as the service key in your app - just use the service key. If you really can't have two different service keys, you could the mapping on the configuration side - map the `ConfigurationStrings:connectionName` => `ConfigurationStrings:serviceKey`, but this would be more work.

If we get more feedback that using the same name for the connection name and the service key isn't understandable, we can add these overloads. But unles we get that feedback, I am not sure we will add these overloads at this time.

egil:
Another argument for adding it to all client libraries is that the hosting libraries offer the same functionality, a split between "name" and "connectionName". It seems to be that client and hosting library should match in this regard. E.g.:

```c#
var commandQueue = builder
    .AddRabbitMQ("MessageQueue")
    .WithManagementPlugin()

builder.AddProject<Projects.ProcessingService>("ProcessingService")
    .WithReference(source: commandQueue, connectionName: "Queue");
```

eerhardt:
For the above reasons, I'm not convinced that we need to take this change. I'm going to close this PR, for now, so it doesn't remain inactive while we decide if it should be done or not.

@egil - Can you open an issue for this scenario? And if we get enough feedback from others that this is necessary, we can consider making this change in the future.

egil:
> For the above reasons, I'm not convinced that we need to take this change. I'm going to close this PR, for now, so it doesn't remain inactive while we decide if it should be done or not.
> 
> @egil - Can you open an issue for this scenario? And if we get enough feedback from others that this is necessary, we can consider making this change in the future.

I can make the issue, but its almost as quick as just submitting a PR that does this for all packages, if you are declining this because it is not consistent across all packages.

PR #4410, opened 2024-06-06
Fix tests for proxyless containers

karolz-ms:
Tests for proxyless container endpoints got stale and are failing. Fixed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4410)

### DIFF
--- a/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
+++ b/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
@@ -713,8 +713,15 @@ public class DistributedApplicationTests
         {
             endpoint.IsProxied = false;
         });
+
+        // Since port is not specified, this instance will use the container target port (6379) as the host port.
+        var redisNoPort = builder.AddRedis("redisNoPort").WithEndpoint("tcp", endpoint =>
+        {
+            endpoint.IsProxied = false;
+        });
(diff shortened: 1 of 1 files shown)


tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
```diff
@@ -744,7 +757,7 @@ public async Task ProxylessContainerWithoutPortThrows()
     {
         using var builder = TestDistributedApplicationBuilder.Create();
 
-        var redis = builder.AddRedis("redis").WithEndpoint("tcp", endpoint =>
+        var redis = builder.AddContainer("dummyRedis", "redis").WithEndpoint("tcp", endpoint =>
```
mitchdenny, line 760: When change from `AddRedis` to `AddContainer` here?

karolz-ms: Because `AddRedis` sets the TargetPort

karolz-ms: So it succeeds and I cannot test the error condition

ReubenBond (approved):

PR #4415, opened 2024-06-07
Allow WithArgs to accept ReferenceExpressions

### LINKED ISSUE #4237: Update `WithArgs` to accept object array
I'd like to be able to do the following

```cs
builder
  .AddExecutable(...)
  .WithEndpoint(name: "endpoint")
  .WithArgs("-p", builder.Resource.GetEndpoint("endpoint").Property(EndpointProperty.Port));
```

But this doesn't work as the `WithArgs` callback only accepts string array params - https://github.com/dotnet/aspire/blob/d4acce5704b8f086cd38c2e29b2766669d8a1277/src/Aspire.Hosting/ResourceBuilderExtensions.cs#L162-L165

For now, can work around this with the callback overload that does accept objects
```cs
return builder
  .AddExecutable(...)
  .WithArgs(context =>{
      context.Args.Add("-p");
      context.Args.Add(builder.Resource.GetEndpoint(endpointName).Property(EndpointProperty.TargetPort));
    });
```

afscrome:
Update the `WithArgs` extension method to accept `object[]` rather than `string[]`.  In particular, this now allows for the use of reference expressions inside `WithArgs`
```cs
.WithArgs(
   "-p",
   resource.PrimaryEndpoint.Property(EndpointProperty.TargetPort)
);
```

Fixes #4237
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4415)

### DIFF
--- a/src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
@@ -47,7 +47,7 @@ public static class ExecutableResourceBuilderExtensions
                           {
                               context.Args.AddRange(args);
                           }
-                      });        
+                      });
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
@@ -23,6 +23,20 @@ public static class ExecutableResourceBuilderExtensions
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistributedApplicationBuilder builder, string name, string command, string workingDirectory, params string[]? args)
     {
+        return AddExecutable(builder, name, command, workingDirectory, (object[]?)args);
+    }
+
+    /// <summary>
+    /// Adds an executable resource to the application model.
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="command">The executable path. This can be a fully qualified path or a executable to run from the shell/command line.</param>
+    /// <param name="workingDirectory">The working directory of the executable.</param>
+    /// <param name="args">The arguments to the executable.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistributedApplicationBuilder builder, string name, string command, string workingDirectory, params object[]? args)
+    {
         workingDirectory = PathNormalizer.NormalizePathForCurrentPlatform(Path.Combine(builder.AppHostDirectory, workingDirectory));
 
         var executable = new ExecutableResource(name, command, workingDirectory);
@@ -33,7 +47,7 @@ public static class ExecutableResourceBuilderExtensions
                           {
                               context.Args.AddRange(args);
                           }
-                      });
+                      });        
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -159,7 +159,7 @@ public static IResourceBuilder<T> WithEnvironment<T>(this IResourceBuilder<T> bu
     /// <param name="builder">The resource builder.</param>
     /// <param name="args">The arguments to be passed to the container when it is started.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<T> WithArgs<T>(this IResourceBuilder<T> builder, params string[] args) where T : IResourceWithArgs
+    public static IResourceBuilder<T> WithArgs<T>(this IResourceBuilder<T> builder, params object[] args) where T : IResourceWithArgs
```
eerhardt, line 162: This is an API breaking change. We can't change existing APIs, we can only add new ones.

davidfowl:
@afscrome are you going to fix this PR?

afscrome:
Hoping to pick this up over the weekend unless someone else wants to take it before then.

Also spotted earlier today that a similar addition is probably warranted for `public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistributedApplicationBuilder builder, string name, string command, string workingDirectory, params string[]? args)`

afscrome:
There is a similar issue with `WithContainerRuntimeArgs`.  Whilst the underlying `ContainerRuntimeArgsCallbackContext` has arguments as a list of objects
https://github.com/dotnet/aspire/blob/a891b99472315073888103ab66dc1223a0bea9d0/src/Aspire.Hosting/ApplicationModel/ContainerRuntimeArgsCallbackAnnotation.cs#L56

ContainerResourceBuilderExtensions.WithContainerRuntimeArgs only accepts a string array
https://github.com/dotnet/aspire/blob/a891b99472315073888103ab66dc1223a0bea9d0/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L183-L186

Shall I fix that as well?

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -33,7 +47,7 @@ public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistribut
                           {
                               context.Args.AddRange(args);
                           }
-                      });
+                      });
```
davidfowl, line 50:
```suggestion
                      });
```

afscrome:
Any ideas on how to fix the ambiguity this change introduces between 
```cs
public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistributedApplicationBuilder builder, string name, string command, string workingDirectory, params string[]? args)
    public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistributedApplicationBuilder builder, string name, string command, string workingDirectory, params object[]? args)
```
I could remove the `params` from the `string[]` version, but according to https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#signatures that is a breaking change, or does the addition of the `params object[]` version mitigate that break?

But equally if I remove the `params` from the `object[]` version, that would somewhat defeat the point.

eerhardt:
@afscrome - it looks like we were using an older SDK - 8.0.200. I updated to the latest .NET SDK 8.0.302 and the build error should be fixed now. You can overload methods on `params` arrays of different types. So we should be good.

afscrome:
@eerhardt Can this be merged now?  I've resolved the merge conflicts and checks are all passing.

eerhardt:
> @eerhardt Can this be merged now? I've resolved the merge conflicts and checks are all passing.

Can you add some tests for the new functionality?

davidfowl:
@afscrome Are you going to finish this one?

davidfowl (approved):

PR #4438, opened 2024-06-10
Improve handling of Kestrel endpoints

### LINKED ISSUE #4436: Improve handling of endpoints defined in Kestrel config
My previous change #4205 added some initial handling, but it has limitations.

In Run mode:
- It doesn't support proxy mode

In Publish mode:
- Kestrel endpoints don't work in a container if they use `localhost` instead of `*`

In both modes:
- If you have additional endpoints defined via `.WithHttpEndpoint()`, they don't work because Kestrel config takes over everything

davidebbo:
- Support using Kestrel endpoints in Proxy mode
- Allow Kestrel endpoints specified as localhost to work in containers
- Allow endpoints defined via .WithHttpEndpoint() to work alongside Kestrel endpoints
- Fixes #4436
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4438)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -9,26 +9,6 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// Used to represent some special endpoints that require different handling
 /// </summary>
-internal enum EndpointSource
-{
-    /// <summary>
-    /// Regular endpoint.
-    /// </summary>
-    Normal,
-    /// <summary>
-    /// The endpoint came from a launch profile
-    /// </summary>
-    LaunchProfile,
-    /// <summary>
-    /// The endpoint came from the Kestrel configuration
-    /// </summary>
-    KestrelConfig,
-    /// <summary>
-    /// For these, we skip adding the endpoint from HTTPS_PORTS env var
-    /// </summary>
-    DefaultHttps
-}
-
 /// <summary>
 /// Represents an endpoint annotation that describes how a service should be bound to a network.
 /// </summary>
@@ -149,9 +129,9 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     public bool IsProxied { get; set; } = true;
 
     /// <summary>
-    /// Used to represent some special endpoints that require different handling
+    /// Gets or sets a value indicating whether the endpoint is from a launch profile.
     /// </summary>
-    internal EndpointSource Source { get; set; }
+    internal bool FromLaunchProfile { get; set; }
 
     /// <summary>
     /// The environment variable that contains the target port. Setting prevents a variable from flowing into ASPNETCORE_URLS for project resources.
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -7,6 +7,29 @@ using System.Net.Sockets;
 namespace Aspire.Hosting.ApplicationModel;
 
 /// <summary>
+/// Used to represent some special endpoints that require different handling
+/// </summary>
+internal enum EndpointType
+{
+    /// <summary>
+    /// Regular endpoint.
+    /// </summary>
+    Normal,
+    /// <summary>
+    /// The endpoint came from a launch profile
+    /// </summary>
+    FromLaunchProfile,
+    /// <summary>
+    /// The endpoint came from the Kestrel configuration
+    /// </summary>
+    FromKestrelConfig,
+    /// <summary>
+    /// For these, we skip adding the endpoint from HTTPS_PORTS env var
+    /// </summary>
+    ExcludedFromPortEnvironment
+}
+
+/// <summary>
 /// Represents an endpoint annotation that describes how a service should be bound to a network.
 /// </summary>
 /// <remarks>
@@ -126,14 +149,9 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     public bool IsProxied { get; set; } = true;
 
     /// <summary>
-    /// Gets or sets a value indicating whether the endpoint is from a launch profile.
+    /// Used to represent some special endpoints that require different handling
     /// </summary>
-    internal bool FromLaunchProfile { get; set; }
-
-    /// <summary>
-    /// Gets or sets a value indicating whether to skip adding the endpoint from HTTPS_PORTS env var
-    /// </summary>
-    internal bool ExcludeFromPortEnvironment { get; set; }
+    internal EndpointType Type { get; set; }
 
     /// <summary>
     /// The environment variable that contains the target port. Setting prevents a variable from flowing into ASPNETCORE_URLS for project resources.
--- a/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
@@ -9,7 +9,7 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// Used to represent some special endpoints that require different handling
 /// </summary>
-internal enum EndpointType
+internal enum EndpointSource
 {
     /// <summary>
     /// Regular endpoint.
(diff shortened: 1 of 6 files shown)


tests/Aspire.Hosting.Tests/KestrelConfigTests.cs
```diff
@@ -72,28 +85,72 @@ public void MultipleKestrelHttpEndpointsKeepTheirNames()
...
               "type": "project.v0",
               "path": "another-path",
               "env": {
                 "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES": "true",
                 "OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES": "true",
                 "OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY": "in_memory",
-                "ASPNETCORE_FORWARDEDHEADERS_ENABLED": "true"
+                "ASPNETCORE_FORWARDEDHEADERS_ENABLED": "true",
+                "Kestrel__Endpoints__http__Url": "http://*:{projectName.bindings.http.targetPort}",
+                "Kestrel__Endpoints__https__Url": "https://*:{projectName.bindings.https.targetPort}"
```
davidebbo, line 146: I'm not entirely sure whether it's correct to have https here. It happens because we now use this Kestrel override mechanism for all bindings. This particular binding is a default one that we generate because there are no explicit https bindings. @davidfowl for comments.

davidebbo: Actually, I'm sure it shouldn't be there. Fixed.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -540,4 +544,37 @@ private static void SetOnePortsEnvVariable(this IResourceBuilder<ProjectResource
             context.EnvironmentVariables[portEnvVariable] = ports.Build();
         }
     }
+
+    private static void SetKestrelUrlOverrideEnvVariables(this IResourceBuilder<ProjectResource> builder)
+    {
+        builder.WithEnvironment(context =>
+        {
+            // If there are any Kestrel endpoints, we need to override all endpoints, even if they
+            // don't come from Kestrel. This is because having Kestrel endpoints overrides eveything
```
davidfowl, line 553:
```suggestion
            // don't come from Kestrel. This is because having Kestrel endpoints overrides everything
```

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -540,4 +544,37 @@ private static void SetOnePortsEnvVariable(this IResourceBuilder<ProjectResource
...
+
+                    var url = new ReferenceExpressionBuilder();
+
+                    // In Run mode, we keep the original Kestrel config host.
+                    // In Publish mode, we always use *, so it can work in a container (where localhost wouldn't work).
+                    var host = builder.ApplicationBuilder.ExecutionContext.IsRunMode &&
+                        builder.Resource.KestrelEndpointAnnotationHosts.TryGetValue(e.EndpointAnnotation, out var kestrelHost) ? kestrelHost : "*";
+
+                    url.AppendLiteral($"{e.EndpointAnnotation.UriScheme}://{host}:");
+                    url.Append($"{e.Property(EndpointProperty.TargetPort)}");
```
davidfowl, line 572:
```suggestion
                    var url = ReferenceExpression.Create($"{e.EndpointAnnotation.UriScheme}://{host}:{e.Property(EndpointProperty.TargetPort)}");
```

tests/Aspire.Hosting.Tests/KestrelConfigTests.cs
```diff
@@ -23,13 +28,21 @@ public async Task SingleKestrelHttpEndpointIsNamedHttpAndOverridesProfile()
...
+            },
+            a =>
+            {
+                Assert.Equal("ExplicitHttp", a.Name);
             }
             );
 
-        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(resource);
+        // Even though we're in Run mode, we need to get the env in Publish mode since ports don't get allocated
+        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(resource, DistributedApplicationOperation.Publish);
```
davidfowl, line 39: Do we have a test for both publish and run modes?

tests/Aspire.Hosting.Tests/KestrelConfigTests.cs
```diff
@@ -72,28 +85,71 @@ public void MultipleKestrelHttpEndpointsKeepTheirNames()
...
+        var resource = CreateTestProjectResource<ProjectWithMultipleHttpKestrelEndpoints>(
+            operation: DistributedApplicationOperation.Run,
+            callback: builder =>
+            {
+                builder.WithHttpEndpoint(5017, name: "ExplicitProxiedHttp");
+                builder.WithHttpEndpoint(5018, name: "ExplicitNoProxyHttp", isProxied: false);
+            });
+
+        // Even though we're in Run mode, we need to get the env in Publish mode since ports don't get allocated
+        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(resource, DistributedApplicationOperation.Publish);
```
davidfowl, line 100: Is this testing the runtime behavior or publish time behavior? You can manually set allocated ports to force that code path to execute.

davidebbo: The intent here is to test Publish, but I didn't know how to allocate the port. As it stands, it's kind of hybrid, in the sense that the ExecutionContext is in RunMode, when we get to `SetKestrelUrlOverrideEnvVariables`, but we're passing Publish here. I'll look at allocating ports to be in full run mode.

davidebbo: I added code to allocate the ports, but I'm not sure this is correct. Is it normal for the env variables to still contain unevaluated expressions like `{projectName.bindings.FirstHttpEndpoint.targetPort}`? Clearly, by the time they get injected in the running process, everything must be evaluated.

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -6,6 +6,9 @@
 
 namespace Aspire.Hosting.ApplicationModel;
 
+/// <summary>
+/// Used to represent some special endpoints that require different handling
+/// </summary>
```
davidfowl, line 11: Remove?

```suggestion
```

davidfowl (approved):
I'll verify deploying works before merging

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -252,27 +252,31 @@ private static IResourceBuilder<ProjectResource> WithProjectDefaults(this IResou
...
-                    builder.WithEndpoint(schemeAsEndpointName ?? endpoint.EndpointName, e =>
-                    {
-                        e.Port = endpoint.BindingAddress.Port;
-                        e.UriScheme = endpoint.BindingAddress.Scheme;
-                        e.IsProxied = false; // turn off the proxy, as we cannot easily override Kestrel bindings
-                        e.Transport = adjustTransport(e);
-                    },
-                    createIfNotExists: true);
-                }
+                    e.Port = endpoint.BindingAddress.Port;
```
davidfowl, line 265: This is an issue for publish mode, we only want to set the TargetPort not the Port. Settings a port does not work for the default http and https endpoint as those are 80 and 443.

davidfowl (changes requested):
Deployment doesn't work.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Hmmm, I'm not crazy about having this difference between Publish and Run.

Stepping back, how is this Kestrel situation different from having a user manually define an external endpoint with some hard coded `Port`? It's going to generate the same manifest, and it's going to blow up on ACA.

It feels quirky to have the Kestrel ports treated as external as runtime, and as a target at Publish time. Instead, can we do this:
- There is exactly one Kestrel endpoint that is used for external traffic. Probably the one named `http`
- We always treat that port as a **target port**, both at runtime and Publish
- Aspire comes up with a proper external facing Port for it:
  - At runtime it can be a random port, unless the user somehow tells us what to use
  - At publish time, we always map both http 80 and https 443 to it in the manifest.

davidebbo: Ok so that's not going to work, because in local Run mode, you can't have replicas if you set the target port. So we have two very different worlds:

Local run:
- Pros: Can have many external ports
- Cons: Target ports must be generated to support proxies and replicas, since they must be unique across the box
Result: the user can pick the Port but not the Target Port

ACA:
- Cons: Can only have one http and one https port
- Pros: Target ports are scoped within the container, so all the replicas can use the same one
Result: the user can pick the Target Port but not the Port

davidfowl (approved):

Issue #3309, opened 2024-03-31
Spin off configuration schema generator as a standalone package


pedropaulovc:
Autocomplete of Aspire configs in appsettings.json is fantastic as my project is growing. I would love to have this capability in my projects as well. Also, I'm using keyed component resolution and the default Aspire schema doesn't know what's the key I'm using in my code. Looks like this capability is handled by the configuration schema generator introduced in #1383 . The team should consider spinning off this capability as an extension package for 3P usage.

davidfowl:
This is the plan right @eerhardt?

eerhardt:
> This is the plan right @eerhardt?

Yes, we will need to do something here if we want to build Aspire Components outside of the dotnet/aspire repo.

aaronpowell:
We're wanting to use this in the Community Toolkit (tracking https://github.com/CommunityToolkit/Aspire/issues/112), what would be required to ship this as a .NET CLI tool that we could use, even if it was on a package feed other than nuget.org.

Alirexaa:
@davidfowl @eerhardt
Is there any plan for this in the near future?
Maybe an aspire cli command?

I would like to resolve this issue if you can provide some information about how this should be designed.

maddymontaquila:
The configuration schema generator has been spun off and improved via PRs #4441, #1318, and #6312. Closing as implemented.

KalleOlaviNiemitalo:
> The configuration schema generator has been spun off and improved via PRs #4441, #1318, and #6312. Closing as implemented.

#1318 is not a pull request and does not look related.

What is the name of the package?  It is not clear to me from <https://github.com/microsoft/aspire/blob/5c8aa3df3eba94f3d71f7ab4b276f03e8233955e/src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj> how this is being packaged.

davidfowl:
@eerhardt are we gonna do this still? It's not really a priorty anymore is it.

Meir017:
could this tool move into https://github.com/dotnet/extensions ?

bart-vmware:
Steeltoe still has an internal copy of this tool. It would be great if we could just reference a NuGet package instead.

davidfowl:
This just isn't a priority for the team. If someone is inspired to move it to extesnions and get it merged that would be great.

PR #2663, opened 2024-03-06
Update a bunch of dependencies to latest

eerhardt:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/2663)

PR #4441, opened 2024-06-10
Fixes and enhancements in JSON Configuration schema generator

bart-vmware:
This PR contains bugfixes and enhancements to the JSON Configuration schema generator, which I'm trying to use outside of Aspire. I've tested with VS 2022 v17.10.1. Please see the individual commit messages for change details. Hopefully, this PR brings the tool a step closer to achieving #3309.

I'm curious as to why JSON properties are sorted alphabetically. I found it makes it harder to compare against the original source code it was generated from.

> [!NOTE]
> This includes a fix in the RuntimeSource directory to prevent a `StackOverflowException`. It should probably be synced with the original runtime source code. Tracked at https://github.com/dotnet/runtime/issues/103802.

/cc @eerhardt 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4441)


eerhardt:
> I'm curious as to why JSON properties are sorted alphabetically.

It makes it "stable", so the .json file can be checked in. It makes version-to-version updates of a library easier, so when new properties are added, you can see what properties are added, and they don't become all mixed up. See https://github.com/dotnet/aspire/pull/2663/ for an example, a new property was added and we can easily diff and see the new property (the properties aren't in a new order after the update).

src/Components/Aspire.Azure.Storage.Blobs/ConfigurationSchema.json
```diff
@@ -75,15 +90,15 @@
                         "GeoRedundantSecondaryUri": {
                           "type": "string",
                           "format": "uri",
-                          "description": "Gets or sets the secondary storage 'System.Uri' that can be read from for the storage account if the\naccount is enabled for RA-GRS.\n\nIf this property is set, the secondary Uri will be used for GET or HEAD requests during retries.\nIf the status of the response from the secondary Uri is a 404, then subsequent retries for\nthe request will not use the secondary Uri again, as this indicates that the resource\nmay not have propagated there yet. Otherwise, subsequent retries will alternate back and forth\nbetween primary and secondary Uri."
```
eerhardt, line 78: We left these newlines in purposefully so when they are in the docs, they are in the UI as well.

Today:
image

After this change:

image

This makes it less readable.

cc @sebastienros

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
```diff
@@ -20,8 +20,13 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
     <PackageReference Include="System.CommandLine" />
+    <PackageReference Include="IgnoresAccessChecksToGenerator" PrivateAssets="all" />
```
eerhardt, line 23: I don't think we should be using this. We should only be using public APIs.

Which internal APIs are you using here?

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaGenerator.cs
```diff
@@ -68,7 +68,12 @@ private static PortableExecutableReference CreateMetadataReference(string path)
                 }
 
                 var path = (string)args[0].Value;
-                (configurationPaths ??= new()).Add((string)args[0].Value);
+                if (string.IsNullOrEmpty(path))
```
eerhardt, line 71: I don't think path should ever be null or empty. What's the scenario here?

tests/ConfigurationSchemaGenerator.Tests/RecordWithPrimitives.cs
```diff
@@ -1,50 +0,0 @@
...
-
-using System.Globalization;
-using Aspire;
-using ConfigurationSchemaGenerator.Tests;
-
-[assembly: ConfigurationSchema("RecordWithPrimitives", typeof(RecordWithPrimitives))]
-
-namespace ConfigurationSchemaGenerator.Tests;
-
-public record RecordWithPrimitives
```
eerhardt, line 12: Can you leave the existing tests in place? You can add new ones, but removing existing tests makes it hard to see what changed.

We are losing some coverage here.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -84,94 +81,241 @@ private JsonObject GenerateGraph()
                 var type = spec.ConfigurationTypes[i];
                 var path = spec.ConfigurationPaths[i];
 
-                GenerateProperties(root, type, path);
+                var pathSegments = ImmutableQueue<string>.Empty;
```
sebastienros, line 84: Does the immutability matter?
I couldn't find any place where it would.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaGenerator.cs
bart-vmware: This is new: an empty path allows to define properties in the configuration root. See [ShouldAllowEmptyComponentPath](https://github.com/dotnet/aspire/blob/bce9dd1249c73123eb260362ea873d822ebf2493/tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs#L719-L754).

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
bart-vmware: `Microsoft.CodeAnalysis.Shared.Extensions.ISymbolExtensions.GetDocumentationComment`, this is explained in https://github.com/dotnet/aspire/pull/4441/commits/218d43de70c75637b767198d6211f1e11828466f.

tests/ConfigurationSchemaGenerator.Tests/RecordWithPrimitives.cs
bart-vmware: I'm convinced that no coverage is lost. I moved the existing tests for primitive types from `RecordWithPrimitiveTypes` to [ShouldConvertTypes](https://github.com/dotnet/aspire/blob/bce9dd1249c73123eb260362ea873d822ebf2493/tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs#L158-L418). The [diff](https://github.com/dotnet/aspire/compare/ace04ef35b8c08e302e6c6dc34a486c6c76447a5...bart-vmware:aspire:config-schema-types#diff-5231539b68734fdabae038ece33d44afd5377e7810b7fd9962382a31b9d53306) between the first and last commit shows how this PR changes them. The only change in output is that `object?` gets converted to `type:object`, which used to be `type:string`.

The reason I moved these is because I had to insert more types. Then the numbering in property names got in the way, so I kept renumbering them, which messes up the PR diff. At that point, I realized it makes more sense to have unit tests to cover all the primitive types, than to stash everything into one very large integration test.

But my move did not cover associated doc-comments. So I picked a few basic types with interesting JSON structures and included doc-comments for them in the new integration test.

Another way of looking at the diff is [here](https://github.com/dotnet/aspire/compare/main...bart-vmware:aspire:config-schema-types-old-tests#diff-ba37d348ecc89e724d88d88a3e2dc1f78c95c76c456d1a04bd359f1a7c3f2a9c), where I added an extra commit on top of this PR that restores the original `RecordWithPrimitives.cs` file. It shows the same change as mentioned above.

This PR greatly improves coverage, and I think the new structure makes it easier to keep it high. Please let me know if you still want me to re-add `RecordWithPrimitives.cs`.

src/Components/Aspire.Azure.Storage.Blobs/ConfigurationSchema.json
bart-vmware: That's interesting. The _duplicate_ line break at https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlobClientOptions.cs#L171-L180 shows up at https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobclientoptions.georedundantsecondaryuri?view=azure-dotnet#azure-storage-blobs-blobclientoptions-georedundantsecondaryuri, but the _singular_ line breaks do not. Are you okay with preserving only the duplicate line breaks?

bart-vmware:
> > I'm curious as to why JSON properties are sorted alphabetically.
> 
> It makes it "stable", so the .json file can be checked in. It makes version-to-version updates of a library easier, so when new properties are added, you can see what properties are added, and they don't become all mixed up. See #2663 for an example, a new property was added and we can easily diff and see the new property (the properties aren't in a new order after the update).

Thanks for the clarification. I suppose it depends on whether to optimize for the development of the generator versus its consumption. Although member order in assemblies is already deterministic, according to the Remarks section at https://learn.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=net-8.0#system-type-getproperties:

> In .NET 6 and earlier versions, the [GetProperties](https://learn.microsoft.com/en-us/dotnet/api/system.type.getproperties?view=net-8.0) method does not return properties in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which properties are returned, because that order varies. However, starting with .NET 7, the ordering is deterministic based upon the metadata ordering in the assembly.

Because https://github.com/dotnet/aspire/pull/2663/files#diff-ba29f362bde4fa8dfaf456726c5d33b10aec019ce5485384d75974e2c98b9277 did not shuffle the order of existing properties, the diff in Aspire would still remain minimal.

bart-vmware:
Regarding the failed checks: I've followed the steps at https://github.com/dotnet/aspire/tree/main/tests/Aspire.Workload.Tests#readme and can't reproduce the failure locally. Here's what's in the failure log:
```
   Failed Aspire.Workload.Tests.StarterTemplateWithWithRedisCacheTests.WebFrontendWorks(urlPrefix: "https://") [13 s]
  Error Message:
   Microsoft.Playwright.PlaywrightException : net::ERR_NETWORK_CHANGED at https://localhost:17003/login?t=d2e8dbf1418a5877c9cd259dff91a759
```
I don't suspect this is related to the changes in this PR, but please let me know if there's anything I can do to get a green build.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -268,98 +430,58 @@ private static void GenerateDocCommentsProperties(JsonObject propertyNode, strin
         }
     }
 
-    private void GenerateTypeDocCommentsProperties(JsonObject? currentNode, TypeSpec type)
+    internal static string FormatDescription(XElement element)
```
eerhardt, line 433: This algorithm is a bit surprising. We inject `<br/><br/>` in ConvertBlankLines and then here turn them to `\n\n\n\n`, and then `\n\n`.

Can you add some comments here to explain it?

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -369,36 +491,46 @@ internal static void AppendUnindentedValue(StringBuilder builder, string value)
...
-            AppendParsableFromString(propertyNode, parsable);
+            return $"<br/><br/>{element.Value}<br/><br/>";
         }
-        else if (propertyTypeSpec is ObjectSpec)
-        {
-            propertyNode["type"] = "object";
-        }
-        else if (propertyTypeSpec is EnumerableSpec)
-        {
-            // TODO: support enumerables correctly
```
eerhardt, line 386: It looks like we lost this TODO and handling for `EnumerableSpec`. Can we add it back?

eerhardt (commented):
Sorry for the delay here @bart-vmware, I was on vacation for a while. But I'm back now.

First off, this is really great! Thank you for the contribution here. These fixes and enhancements really help.

I pushed some changes that I could (mostly to get rid of the extra dependency and to just use normal reflection). I left some other comments. Once those are addressed, I can merge this.

Thanks again.

tests/ConfigurationSchemaGenerator.Tests/RecordWithPrimitives.cs
eerhardt: Thanks for the explanation. I agree this coverage isn't lost (and is greatly expanded in the new code).

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
eerhardt: I removed this and just used regular reflection.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
bart-vmware: There's no need for that. The behavior is the same for its base type, `CollectionSpec`:

image

The test coverage is at `GeneratorTests.ShouldConvertTypes`:
```c#
[InlineData("IEnumerable<string>?",
    """
    "type": "array",
    "items": {
      "type": "string"
    }
    """)]
```

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -84,96 +86,259 @@ private JsonObject GenerateGraph()
...
+
+                GenerateComponent(rootNode, type, pathSegments);
             }
         }
-
-        return root;
     }
 
-    private void GenerateProperties(JsonObject parent, TypeSpec type, string path)
+    private bool GenerateComponent(JsonObject currentNode, TypeSpec type, Queue<string> pathSegments)
```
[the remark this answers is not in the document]
bart-vmware: In:
```c#
[assembly: ConfigurationSchema("One.Two.Three", typeof(ExampleSettings))]
```
"One", "Two" and "Three" are components.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
```diff
@@ -51,56 +118,1110 @@ public void ShouldRemoveIndentation(string value, string expected)
...
+
+            public class JsonFormattingOptions
+            {
+                public bool WriteIndented { get; set; }
+            }
+            """;
+
+        var schema = GenerateSchemaFromCode(source, []);
+
+        Assert.Contains("FormatSettings", schema);
```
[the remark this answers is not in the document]
bart-vmware: In general, get-only properties are excluded in the schema. But collections and non-primitive types are exceptions to that. This test verifies the property gets emitted, even if the declared type is get-only.

I didn't assert on the full tree, because most of its content isn't relevant here. This way, future refactorings of other tree parts won't require "fixing" this test.

src/Tools/ConfigurationSchemaGenerator/RuntimeSource/Configuration.Binder/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -528,8 +528,17 @@ private bool IsAssignableTo(ITypeSymbol source, ITypeSymbol dest)
                 return conversion.IsReference && conversion.IsImplicit;
             }
 
-            private bool IsUnsupportedType(ITypeSymbol type)
+            private bool IsUnsupportedType(ITypeSymbol type, int recursionDepth = 0)
```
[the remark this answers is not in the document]
bart-vmware: Created https://github.com/dotnet/runtime/issues/103802.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -84,96 +86,259 @@ private JsonObject GenerateGraph()
...
-        foreach (var segment in pathSegments)
+        if (pathSegments.Count == 0)
         {
-            if (currentNode[segment] is JsonObject child)
+            return GenerateType(currentNode, type);
+        }
+
+        var pathSegment = pathSegments.Dequeue();
+
+        var backupTypeNode = currentNode["type"];
```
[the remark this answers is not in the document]
bart-vmware: I've pushed a commit that explains it.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -31,15 +35,14 @@ internal sealed partial class ConfigSchemaEmitter(SchemaGenerationSpec spec, Com
     private readonly Stack<TypeSpec> _visitedTypes = new();
     private readonly string[] _exclusionPaths = CreateExclusionPaths(spec.ExclusionPaths);
 
-    [GeneratedRegex(@"( *)\r?\n( *)")]
-    private static partial Regex Indentation();
+    [GeneratedRegex(@"( *)(?:\r?\n\s*\r?\n)( *)")]
```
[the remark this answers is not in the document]
bart-vmware: Good catch! I took this from the original regex without thinking about it. Pushed a commit to address this.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
bart-vmware: I've pushed a commit to explain. Tests exist for all whitespace replacements around `<br/>`.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
eerhardt: Sounds good. Thanks for the explanation.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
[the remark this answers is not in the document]
eerhardt: Can we use a different term? "Components" is kind of an overloaded term in the .NET Aspire world. How about "GeneratePathSegment"?

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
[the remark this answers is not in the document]
eerhardt: But it doesn't catch other bugs. It also doesn't validate that "FormatSettings" gets put in the right location. Since we are validating the full tree everywhere else, it feels like we should just do the same here. It gets us better test coverage.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
```diff
@@ -51,56 +118,1110 @@ public void ShouldRemoveIndentation(string value, string expected)
...
+        //   "TestComponent": {
+        //     "TestProperty": {
+        //       "UserName": "john.doe@email.com",
+        //       "Password": "P@ssw0rd!",
+        //       "Options": {
+        //         "RequireSSL": "false"
+        //       }
+        //     }
+        //   }
+        // }
```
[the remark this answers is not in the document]
eerhardt: My biggest concern is that this comment doesn't match the code below. I don't see it using `TestProperty1` or `TestProperty2`. I think the comment is confusing since it doesn't match.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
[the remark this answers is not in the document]
eerhardt: Thanks!

src/Tools/ConfigurationSchemaGenerator/RuntimeSource/Configuration.Binder/ConfigurationBindingGenerator.Parser.cs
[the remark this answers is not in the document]
eerhardt: Awesome. Thank you!

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
```diff
@@ -51,56 +118,1110 @@ public void ShouldRemoveIndentation(string value, string expected)
...
-        Assert.Equal(expected, first.ToString().Trim());
+        Assert.Equal(expectedText, actualText);
     }
 
     [Fact]
     public void IntegrationTest()
     {
         // the 'refs' folder is populated by PreserveCompilationContext in the .csproj
-        var referenceAssemblies = Directory.GetFiles(Path.Combine(Directory.GetCurrentDirectory(), "refs"), "*.dll", SearchOption.AllDirectories)
+        var referenceAssemblies = Directory.GetFiles(Directory.GetCurrentDirectory(), "*.dll", SearchOption.AllDirectories)
```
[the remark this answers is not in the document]
eerhardt: > For example, leaving out <PreserveCompilationContext>true</PreserveCompilationContext> in the test project fails for reasons I don't fully understand.

Check out https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#preservecompilationcontext for an explanation of what PreserveCompilationContext.

We use this in the tests so we can get all the "ref" assemblies that were used at build-time. We can load the "ref" assemblies up when the test runs and pass them to Roslyn, so Roslyn can return valid information.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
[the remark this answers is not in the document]
bart-vmware: Makes sense. I've replaced the comment with compilable code.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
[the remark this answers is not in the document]
bart-vmware: Existing tests already cover all other concerns. It's _duplicating_ test coverage, not making it better. But I've added it anyway.

bart-vmware:
@eerhardt I've addressed all comments. Can you take another look? Please don't hesitate to resolve the open conversations if they've been addressed properly.

I don't think the failing checks are related to these changes, but I can't rerun them.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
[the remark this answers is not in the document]
eerhardt: > Do you understand why that happens?

yes, this happens because not all the "ref" assemblies are available when you remove PreserveCompilationContext.

> The integration test doesn't load reference assemblies from the "ref" directory anymore

This isn't true because the code recurses to all directories under the current directory, which includes the `ref` directory.

eerhardt (approved):
Thanks again for this great contribution, @bart-vmware!

The changes look good to me. I'm ready to merge once CI is green. We had some flaky failures in the past, but we should be in a better state now.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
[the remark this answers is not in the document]
bart-vmware: Thanks for the explanation, makes sense.

PR #4443, opened 2024-06-10
update dependency versions

danmoseley:
```
dotnet tool install --global dotnet-outdated-tool
dotnet outdated -vl Major -u
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4443)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -21,7 +21,7 @@
     <PackageVersion Include="Azure.AI.OpenAI" Version="1.0.0-beta.17" />
     <PackageVersion Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
-    <PackageVersion Include="Azure.Identity" Version="1.11.3" />
+    <PackageVersion Include="Azure.Identity" Version="1.11.4" />
     <PackageVersion Include="Azure.Messaging.EventHubs.Processor" Version="5.11.3" />
     <PackageVersion Include="Azure.Messaging.ServiceBus" Version="7.17.5" />
     <PackageVersion Include="Azure.Search.Documents" Version="11.5.1" />
@@ -131,7 +131,7 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.8.0" />
     <!-- build dependencies -->
     <PackageVersion Include="MicroBuild.Plugins.SwixBuild.Dotnet" Version="1.1.87-gba258badda" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.9.2" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4" />
     <PackageVersion Include="Microsoft.DotNet.Build.Tasks.Workloads" Version="8.0.0-beta.23564.4" />
     <PackageVersion Include="Microsoft.Signed.Wix" Version="$(MicrosoftSignedWixVersion)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,11 +9,11 @@
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.307.17" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.301.4" />
-    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.301.39" />
-    <PackageVersion Include="AWSSDK.Core" Version="3.7.304.1" />
-    <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.300" />
+    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.308.9" />
+    <PackageVersion Include="AWSSDK.SQS" Version="3.7.301.17" />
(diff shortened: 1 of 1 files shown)


joperezr:
Thanks for doing this @danmoseley, change looks good to me. However, we may want to add one thing to your script before calling outdated tool: `dotnet new nugetconfig --force`. This will make sure that the feed used for the updates is nuget.org, as that will be the only way of knowing if a dependency actually has a new version.

danmoseley:
Yes, you're quite right, I did this for aspire-samples and forgot this one's different. Let me see.

joperezr:
Also, I don't think we want to do version lock on Major, at least not for the Aspire Component dependencies, as we said that we would be always on latest stable, regardless of it being a major version bump.

danmoseley:
You shared your script somewhere last time and I should have just used that...

danmoseley:
Looks like the Azure OpenAI package has breaking changes in 2.x. Reverting that.

joperezr:
> Looks like the Azure OpenAI package has breaking changes in 2.x. Reverting that.

Ooh, first interesting one then. I guess for this particular case it is okay because our component that references that isn't stable yet, so we just have to update and address the breaks. cc: @eerhardt @sebastienros

Directory.Packages.props
```diff
@@ -128,24 +128,24 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.GrpcNetClient" Version="1.8.0-beta.1" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.8.1" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.8.1" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.8.0" />
+    <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.8.1" />
     <!-- build dependencies -->
     <PackageVersion Include="MicroBuild.Plugins.SwixBuild.Dotnet" Version="1.1.87-gba258badda" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.9.2" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
```
joperezr, line 134: I think I ask this every time (will write a note this time for myself) but just to triple check, this isn't going to mean that in order to use Aspire you need to use a newer version of VS, right @eerhardt?

joperezr (approved):

danmoseley:
opened https://github.com/dotnet/aspire/issues/4466

danmoseley:
aaand Dependabot comes to life and opens PR's for many of these. @eerhardt if you can answer the question above (or we can snip that update) I suggest we merge this one as it will likely avoid merge conflicts. Any remaining ones we can merge after. @joperezr 

Also I think this should be the last time we do this so manually

joperezr:
I'm pretty sure the CodeAnalysis update is okay to take too as that source generator we only use internally, but totally okay to update that later too. Nice to hear that dependabot decided to work again :)

danmoseley:
```

Aspire.Workload.Tests.StarterTemplateRunTests.ResourcesShowUpOnDashboad [1 s]
  Error Message:
   System.AggregateException : One or more errors occurred. (
╔══════════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers.     ║
║ Please install them with the following command:          ║
║                                                          ║
║     sudo pwsh bin/Debug/netX/playwright.ps1 install-deps ║
║                                                          ║
║ Alternatively, use apt:                                  ║
║     sudo apt-get install libgbm1\                        ║
║         libxkbcommon0                                    ║
``` 
@radical ?

radical:
> ```
> 
> Aspire.Workload.Tests.StarterTemplateRunTests.ResourcesShowUpOnDashboad [1 s]
>   Error Message:
>    System.AggregateException : One or more errors occurred. (
> ╔══════════════════════════════════════════════════════════╗
> ║ Host system is missing dependencies to run browsers.     ║
> ║ Please install them with the following command:          ║
> ║                                                          ║
> ║     sudo pwsh bin/Debug/netX/playwright.ps1 install-deps ║
> ║                                                          ║
> ║ Alternatively, use apt:                                  ║
> ║     sudo apt-get install libgbm1\                        ║
> ║         libxkbcommon0                                    ║
> ```
> 
> @radical ?

Ah, I see it..

```
+ sudo apt-get install -y libgbm1 libxkbcommon0
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 1669 (dpkg)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

`apt-get` failed to install the dependencies, but we don't stop the test run because of that. And so it goes on to fail later in the run. Re-run should "fix" it for this PR.

In a separate one I can make it fail immediately if `apt-get` fails. Additionally, we should have some retries around this, and I remember looking at it for `dotnet/runtime`, and I can try to add something here.

radical:
Tracking it in https://github.com/dotnet/aspire/issues/4487 .

danmoseley:
Thanks

eerhardt:
> > Looks like the Azure OpenAI package has breaking changes in 2.x. Reverting that.
> 
> Ooh, first interesting one then. I guess for this particular case it is okay because our component that references that isn't stable yet, so we just have to update and address the breaks. cc: @eerhardt @sebastienros

This is known. The Azure SDK team rewrote the OpenAI library. See https://devblogs.microsoft.com/dotnet/openai-dotnet-library/.

Thanks for opening https://github.com/dotnet/aspire/issues/4466 to track this. I've also opened https://github.com/dotnet/aspire/issues/4567 to track adding a plain "OpenAI" component given this new library.

Directory.Packages.props
eerhardt: > this isn't going to mean that in order to use Aspire you need to use a newer version of VS

That is correct. Notice the comment above this: `<!-- build dependencies -->`. We only use this version in our JSON schema generator.

PR #4444, opened 2024-06-11
Updates to Aspire.Hosting.Testing

### LINKED ISSUE #2790: Testing: allow logs to flow into test output
When developers use `Aspire.Hosting.Testing`, they would benefit from seeing logs in test output. We should add this. Logs are accessible via `Aspire.Hosting.ResourceLoggerService`. See also `ResourceNotificationService`

### LINKED ISSUE #4445: Add ResourceNotificationService.WaitForResourceAsync
## Background and Motivation

When writing tests with `Aspire.Hosting.Testing` against AppHost projects, it's beneficial to have a programmatic way to wait on a resource to transition to a known state from within a test, so that tests don't needlessly attempt to connect to a resource when it's known to be in a state where that attempt will fail, e.g. before it transitions to the "Running" state as shown in the dashboard, or reported by `ResourceNotificationService`.

## Proposed API

We should consider adding a new method `Aspire.Hosting.ApplicationModel.ResourceNotificationService.WaitForResourceAsync` to aid in these scenarios, along with adding more common states to `Aspire.Hosting.ApplicationModel.KnownResourceStates`.

```diff
namespace Aspire.Hosting.ApplicationModel;

public class ResourceNotificationService
{
    public ResourceNotificationService(ILogger<ResourceNotificationService> logger);
+   public ResourceNotificationService(ILogger<ResourceNotificationService> logger, IHostApplicationLifetime hostApplicationLifetime);
+   public Task WaitForResourceAsync(string resourceName, string targetState = KnownResourceStates.Running, CancellationToken? cancellationToken = null);
+   public Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates, CancellationToken? cancellationToken = null);
}

public static class KnownResourceStates
{
    public static readonly string Hidden = "Hidden";
+   public static readonly string Starting = "Starting";
+   public static readonly string Running = "Running";
+   public static readonly string FailedToStart = "FailedToStart";
+   public static readonly string Stopping = "Stopping";
+   public static readonly string Exited = "Exited";
+   public static readonly string Finished = "Finished";
}
```

## Usage Examples

```csharp
[Fact]
{
    // Arrange
    var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.MyAspireApp_AppHost>();
    appHost.Services.ConfigureHttpClientDefaults(client =>
    {
        client.AddStandardResilienceHandler();
    });

    var resourceNotification = app.Services.GetRequiredService<ResourceNotificationService>();
    await app.StartAsync();

    // Act
    await resourceNotification.WaitForResourceAsync("webfrontend", KnownResourceStates.Running);
    var httpClient = app.CreateHttpClient("webfrontend");
    var response = await httpClient.GetAsync("/");

    // Assert
    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
}
```

## Alternative Designs

We initially thought about adding a new public type in `Aspire.Hosting.Testing` that could be resolved from DI manually and have the `WaitForResourceAsync` method on that but given the target state is aligned with the state text reported by `RequestNotificationService` it seems appropriate to add it directly there.

Perhaps `WaitForResourceAsync` could just be `WaitForAsync` given it's already on the `RequestNotificationService`.

## Risks

The resource states are just string values and aren't enforced anywhere, or even used to set states by `Aspire.Hosting`, as the states generally come from DCP. It seems reasonable that we should have a set of common known states though.

The overload `WaitForResource("myresource")` where the `targetState` parameter is defaulted could be something we'd prefer to map to a different behavior in the future, e.g. once health checks are added it could wait for health checks to be healthy before returning rather than just status update text matching.

DamianEdwards:
This change enables capturing resource logs and forwarding them to `ILogger` during testing when using `Aspire.Hosting.Testing`, along with an update to `ResourceNotificationService` to enable waiting for a resource to transition to a given state text.

- Forward resource logs to `ILogger` in `Aspire.Hosting.Testing`
- Add `ResourceNotificationService.WaitForResourceAsync`
- Added more states to `KnownResourceStates`
- Updates test templates to use `ResourceNotificationService.WaitForResourceAsync`
- Updates xUnit templates with a comment about packages that map `ILogger` to `ITestOutputHelper`

Some open questions:
- Should we have an option/method to control whether resource logs are forwarded to `ILogger` or not in `Aspire.Hosting.Testing`?
- Should we have an option to control the log level the resource logs are logged at (stderr vs. stdout, etc.)?
- Should we update `ResourceNotificationService` to cancel `WatchAsync` streams when `IHostApplicationLifetime.ApplicationStopping` is triggered?

Fixes #4445
Fixes #2790

File -> New -> Aspire Starter App with xUnit.net test project with these changes:
image

File -> New -> Aspire Starter app with MSTest test project:
image

File -> New -> Aspire Starter app with NUnit test project:
image


src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,47 @@ namespace Aspire.Hosting.ApplicationModel;
...
+    /// If the resource never reaches the target state, the method will wait for the passed <see cref="CancellationToken"/> to be cancelled, or 60 seconds if <c>null</c>, and then throw <see cref="TaskCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resouce.</param>
+    /// <param name="targetState">The state to wait for the resource to transition to. See <see cref="KnownResourceStates"/> for common states. Defaults to <see cref="KnownResourceStates.Running"/>.</param>
+    /// <param name="cancellationToken">The cancellation token to monitor to cancel the wait operation. If <c>null</c> defaults to waiting for 60 seconds.</param>
+    /// <returns></returns>
+    public Task WaitForResourceAsync(string resourceName, string? targetState = null, CancellationToken? cancellationToken = null)
+    {
+        var stateToWaitFor = targetState ?? KnownResourceStates.Running;
+        var token = cancellationToken ?? new CancellationTokenSource(s_defaultWaitForResourceStateTimeout).Token;
```
davidfowl, line 54: you should dispose this CTS if you create it

DamianEdwards: Updated.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,71 @@ namespace Aspire.Hosting.ApplicationModel;
...
+        if (cancellationToken is { } ct)
+        {
+            token = ct;
+        }
+        else
+        {
+            timeoutCts = new(s_defaultWaitForResourceStateTimeout);
+            token = timeoutCts.Token;
+        }
+        var cts = CancellationTokenSource.CreateLinkedTokenSource(_applicationStopping, token);
```
DamianEdwards, line 63: Should we update `WatchAsync` to do this too so that it cancels when the `_applicationStopping` token is signalled?

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    private async Task WatchResourceLogs(IResource resource, string resourceId, CancellationToken cancellationToken)
+    {
+        var applicationName = hostEnvironment.ApplicationName;
+        var logger = loggerFactory.CreateLogger($"{applicationName}.Resources.{resource.Name}");
+        await foreach (var logEvent in resourceLoggerService.WatchAsync(resourceId).WithCancellation(cancellationToken))
+        {
+            foreach (var line in logEvent)
+            {
+                var logLevel = line.IsErrorMessage ? LogLevel.Error : LogLevel.Information;
```
DamianEdwards, line 54: Should we make the log level the resource `LogLine`s are logged at an option?

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        await Task.WhenAll(logWatchTasks).ConfigureAwait(false);
+    }
+
+    private async Task WatchResourceLogs(IResource resource, string resourceId, CancellationToken cancellationToken)
+    {
+        var applicationName = hostEnvironment.ApplicationName;
+        var logger = loggerFactory.CreateLogger($"{applicationName}.Resources.{resource.Name}");
```
DamianEdwards, line 49: Log category will be like "MyAspireApp.AppHost.Resources.webfrontend" which should informative and facilitate prefix filtering.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,125 @@ namespace Aspire.Hosting.ApplicationModel;
...
+            token = ct;
+        }
+        else
+        {
+            timeoutCts = new CancellationTokenSource(s_defaultWaitForResourceStateTimeout);
+            token = timeoutCts.Token;
+        }
+        try
+        {
+            var targetStates = !string.IsNullOrEmpty(targetState) ? new[] { targetState } : [KnownResourceStates.Running];
```
davidfowl, line 68: nit:
```suggestion
            string[] targetStates = !string.IsNullOrEmpty(targetState) ? [targetState] : [KnownResourceStates.Running];
```

DamianEdwards: Ah I wish the fixer suggested that.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,125 @@ namespace Aspire.Hosting.ApplicationModel;
...
+    /// </summary>
+    /// <remarks>
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.<br/>
+    /// If the resource doesn't reach one of the target states within 60 seconds, it will throw <see cref="TaskCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
+    public Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates)
```
davidfowl, line 88: We don't want a default timeout.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,125 @@ namespace Aspire.Hosting.ApplicationModel;
 /// </summary>
 public class ResourceNotificationService(ILogger<ResourceNotificationService> logger)
 {
+    private static readonly TimeSpan s_defaultWaitForResourceStateTimeout = TimeSpan.FromSeconds(60);
```
davidfowl, line 17: Lets not have a default timeout.

DamianEdwards: Really? Hmmm. I feal uneasy about having methods that just hang forever so easily.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,125 @@ namespace Aspire.Hosting.ApplicationModel;
...
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <param name="cancellationToken">A cancellation token that cancels the wait operation when signalled.</param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
+    public async Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates, CancellationToken cancellationToken)
+    {
+        var watchCts = CancellationTokenSource.CreateLinkedTokenSource(_applicationStopping, cancellationToken);
```
davidfowl, line 114:
```suggestion
        using var watchCts = CancellationTokenSource.CreateLinkedTokenSource(_applicationStopping, cancellationToken);
```

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -13,12 +14,125 @@ namespace Aspire.Hosting.ApplicationModel;
...
+        var watchToken = watchCts.Token;
+        try
+        {
+            await foreach (var resourceEvent in WatchAsync(watchToken).WithCancellation(watchToken).ConfigureAwait(false))
```
davidfowl, line 118: Why do we pass the token twice

DamianEdwards: You're right, that's not required. It was related to the other comment about how WatchAsync doesn't combine with the app stopping token by default but this is doing it anyway.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: See all of the methods on this class, pass a cancellation token if you want a timeout 😄

DamianEdwards: The other methods on this class are async enumerable though so they're consumption pattern is a bit different. But if you feel strongly about it I'm happy to remove it and put the timeout in the test templates instead.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -16,9 +17,89 @@ public class ResourceNotificationService(ILogger<ResourceNotificationService> lo
...
+    private readonly CancellationToken _applicationStopping;
 
     private Action<ResourceEvent>? OnResourceUpdated { get; set; }
 
+    /// <summary>
+    /// Creates a new instance of <see cref="ResourceNotificationService"/>.
+    /// </summary>
+    /// <param name="logger">The logger.</param>
+    /// <param name="hostApplicationLifetime">The host application lifetime.</param>
+    public ResourceNotificationService(ILogger<ResourceNotificationService> logger, IHostApplicationLifetime hostApplicationLifetime)
```
DamianEdwards, line 29: Do we do anything with the existing constructor when we add a new one with more parameters, e.g. obsolete it?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -16,9 +17,89 @@ public class ResourceNotificationService(ILogger<ResourceNotificationService> lo
...
+    /// <remarks>
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.<br/>
+    /// Note the returned task will not complete if the resource doesn't reach one of the target states.
+    /// Use <see cref="WaitForResourceAsync(string, IEnumerable{string}, CancellationToken)"/> to pass a <see cref="CancellationToken"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
+    public Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates)
```
davidfowl, line 70: Do you need this method?

DamianEdwards: Yeah without a default it's kinda dangerous :)
I'll remove

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/WebTests.cs
```diff
@@ -23,10 +24,16 @@ public async Task GetWebResourceRootReturnsOkStatusCode()
         {
             clientBuilder.AddStandardResilienceHandler();
         });
+#if (TestFramework == "xUnit.net")
+        appHost.Services.AddLogging(logging => logging.AddXunit(output));
```
DamianEdwards, line 28: If we don't want to include an `XunitLogger` implementation in the templates, we could instead add a comment pointing folks to https://www.nuget.org/packages?q=logging+xunit

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
mitchdenny: I was on the fence, but then I thought we might want to tune things so we only get errors/warnings to help with memory consumption in the test case (for longer running tests).

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
mitchdenny: Yeah ... obsolete it.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -16,9 +17,74 @@ public class ResourceNotificationService(ILogger<ResourceNotificationService> lo
...
+    /// is already in the target state, the method will return immediately.<br/>
+    /// If the resource doesn't reach one of the target states before <paramref name="cancellationToken"/> is signalled, this method
+    /// will throw <see cref="OperationCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resouce.</param>
+    /// <param name="targetState"></param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> </param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
+#pragma warning disable RS0027 // API with optional parameter(s) should have the most parameters amongst its public overloads
+    public Task WaitForResourceAsync(string resourceName, string? targetState = null, CancellationToken? cancellationToken = null)
```
mitchdenny, line 51: Should this return `Task<bool>`?

mitchdenny: Scratch that ... in this context throwing is probably better - we wouldn't want people to miss this.

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/WebTests.cs
mitchdenny: XunitContext?

mitchdenny: I feel like it is a pretty big chunk of code to include. But I think that there are some benefits to having it here just to make things simple for people. I'm surprised this isn't in a package somewhere.

mitchdenny:
To me the outstanding question regarding this is how this intersects with WaitFor behavior that you might have in your apphost. I really wish we could make this one and the same thing.

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
DamianEdwards: That could be done already by filtering the logs to error or higher in the test setup I think. But it might be desirable to be able to change the level the non-errors are logged at? Not sure, might not be worth the complexity.

DamianEdwards:
> To me the outstanding question regarding this is how this intersects with WaitFor behavior that you might have in your apphost. I really wish we could make this one and the same thing.

Perhaps we rename the method to `WaitForResourceStatus` to try and separate it from any future WaitFor concept at the AppHost level?

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/WebTests.cs
DamianEdwards: There's lots of them! Just none "official".

davidfowl: This amount of code in the template exceeds my comfort level.

DamianEdwards: Yeah it's you I was thinking of 😄 
Let's remove it and add a comment pointing to a nuget query instead.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,103 @@
...
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.<br/>
+    /// If the resource doesn't reach one of the target states before <paramref name="cancellationToken"/> is signalled, this method
+    /// will throw <see cref="OperationCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <param name="cancellationToken">A cancellation token that cancels the wait operation when signalled.</param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
+    public async Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates, CancellationToken cancellationToken)
```
davidfowl, line 87:
```suggestion
    public async Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates, CancellationToken? cancellationToken = null)
```

timheuer:
Can do a validation these work as expected within test explorer in VS Code as well?

DamianEdwards:
I had a thought this morning... should we put the request logs aggregator in Aspire.Hosting instead but just not enable it there by default, and then enable it in Aspire.Hosting.Testing, rather than it being exclusive to Aspire.Hosting.Testing? That way, if you want resource logs in the AppHost logs, you can do it easily, e.g. in tests that don't use the `DistributedApplicationTestingBuilder`.

Thoughts? @davidfowl @ReubenBond @mitchdenny

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
DamianEdwards: @davidfowl thoughts on this? Currently, the `IAsyncEnumerable` returned from `WatchAsync` does not get cancelled when the application stopping token is triggered. Should we fix that?

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
```diff
@@ -0,0 +1,67 @@
...
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting.Testing;
+
+/// <summary>
+/// A background service that aggregates resource logs and forwards them to the ILogger infrastructure.
+/// </summary>
+internal sealed class ResourceLogAggregatorBackgroundService(
```
DamianEdwards, line 13: I had a thought this morning... should we put the request logs aggregator in Aspire.Hosting instead but just not enable it there by default, and then enable it in Aspire.Hosting.Testing, rather than it being exclusive to Aspire.Hosting.Testing? That way, if you want resource logs in the AppHost logs, you can do it easily, e.g. in tests that don't use the DistributedApplicationTestingBuilder.

Thoughts? @davidfowl @ReubenBond @mitchdenny

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,105 @@
...
+
+    /// <summary>
+    /// Creates a new instance of <see cref="ResourceNotificationService"/>.
+    /// </summary>
+    /// <param name="logger">The logger.</param>
+    /// <param name="hostApplicationLifetime">The host application lifetime.</param>
+    public ResourceNotificationService(ILogger<ResourceNotificationService> logger, IHostApplicationLifetime hostApplicationLifetime)
+    {
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        _applicationStopping = hostApplicationLifetime?.ApplicationStopping ?? throw new ArgumentNullException(nameof(hostApplicationLifetime));
```
davidfowl, line 50: When is this ever null?

DamianEdwards: It was previously throwing the ANE for the logger parameter in the original constructor so I made it consistent when adding the new one. You want to remove bother ANEs?

davidfowl: It's fine

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -108,16 +108,45 @@ public static class KnownResourceStateStyles
...
 {
     /// <summary>
     /// The hidden state. Useful for hiding the resource.
     /// </summary>
     public static readonly string Hidden = "Hidden";
+
+    /// <summary>
+    /// The starting state. Useful for showing the resource is starting.
+    /// </summary>
+    public static readonly string Starting = "Starting";
```
radical, line 126: Maybe `public static readonly string Starting = nameof(Starting);`?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,105 @@
...
+                                                     Justification = "targetState(s) parameters are mutually exclusive.")]
+    public async Task<string> WaitForResourceAsync(string resourceName, IEnumerable<string> targetStates, CancellationToken cancellationToken = default)
+    {
+        using var watchCts = CancellationTokenSource.CreateLinkedTokenSource(_applicationStopping, cancellationToken);
+        var watchToken = watchCts.Token;
+        await foreach (var resourceEvent in WatchAsync(watchToken).ConfigureAwait(false))
+        {
+            if (string.Equals(resourceName, resourceEvent.Resource.Name, StringComparisons.ResourceName)
+                && resourceEvent.Snapshot.State?.Text is { Length: > 0 } statusText
+                && targetStates.Contains(statusText, StringComparers.ResourceState))
```
radical, line 97: Should `targetStates` be converted to an array or some thing else, to avoid iterating over the `IEnumerable` repeatedly? I'm not sure if that is warranted here or necessarily helps a lot.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,105 @@
...
+    /// <remarks>
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.<br/>
+    /// If the resource doesn't reach one of the target states before <paramref name="cancellationToken"/> is signalled, this method
+    /// will throw <see cref="OperationCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <param name="cancellationToken">A cancellation token that cancels the wait operation when signalled.</param>
+    /// <returns>A <see cref="Task"/> representing the wait operation.</returns>
```
radical, line 86: Should this also express that the task returns a `string`, and what that is?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,105 @@
...
+        {
+            if (string.Equals(resourceName, resourceEvent.Resource.Name, StringComparisons.ResourceName)
+                && resourceEvent.Snapshot.State?.Text is { Length: > 0 } statusText
+                && targetStates.Contains(statusText, StringComparers.ResourceState))
+            {
+                return statusText;
+            }
+        }
+
+        throw new OperationCanceledException("The operation was cancelled before the resource reached one of the target states.");
```
radical, line 103: Would it be useful to list the expected target states here?

tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
```diff
@@ -147,6 +148,149 @@ await foreach (var item in notificationService.WatchAsync(cancellation))
...
+        var waitTask = notificationService.WaitForResourceAsync("myResource1", targetState: null, cancellationToken: cts.Token);
+
+        await notificationService.PublishUpdateAsync(resource1, snapshot => snapshot with { State = KnownResourceStates.Running });
+        await waitTask;
+
+        Assert.True(waitTask.IsCompletedSuccessfully);
+    }
+
+    [Fact]
+    public async Task WaitingOnResourceReturnsCorrectStateWhenResourceReachesOneOfTargetStatesBeforeCancellation()
```
radical, line 216: `ReachesOneOfTargetStatesBeforeCancellation` - what is this checking for exactly? When should the cancellation be happening for the test? Relying only on the timeout would mean that the cancellation could happen before/after the publish, or before the new state gets read, IIUC.

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
radical: Does this need to be documented somewhere?

radical:
`Updates XUnit templates to include an XunitLogger` - should be dropped from the PR description.

radical:
In a follow up PR, we should use some of this in the template tests, and the EndToEnd tests.

tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
DamianEdwards: It's simply testing the happy-path case of passing multiple target states and ensuring the task returns with the first state that was reached. The timeout CTS is only there to ensure the test doesn't hang if there's an issue.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
DamianEdwards: I don't think it matters much here and it'd just be calling `Contains` anyway.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: Leave it where it is for now. We can always move it.

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
davidfowl: Yep. @DamianEdwards file a doc issue

radical:
> > In a follow up PR, we should use some of this in the template tests, and the EndToEnd tests.
> 
> @radical are you doing this?

Opened an issue for it - https://github.com/dotnet/aspire/issues/4549 .

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
DamianEdwards: I think this is worthwhile for the flexibility it adds.

DamianEdwards: Sure, it's just "fresh" right now and trivial to do. But I can log a follow-up issue isntead.

src/Aspire.Hosting.Testing/ResourceLogAggregatorBackgroundService.cs
DamianEdwards: Logged dotnet/aspire-docs#1114

src/Aspire.ProjectTemplates/templates/aspire-starter/.template.config/template.json
```diff
@@ -49,6 +49,12 @@
           "exclude": [
             "Aspire-StarterApplication.1.Tests/**/*.*"
           ]
+        },
+        {
+          "condition": "(TestFramework != \"xUnit.net\")",
+          "exclude": [
+            "Aspire-StarterApplication.1.Tests/XunitLogger.cs"
```
radical, line 56: Where is this file coming from?

DamianEdwards: ah it's no longer there, will update this

DamianEdwards:
Hmm looks like there's a race somewhere as tests are sometimes failing with duplicate logs. I can repro locally if I run 10,000 iterations. Not sure if the bug is in the `FakeLogger`, `ResourceNotificationService`, or `ResourceNotificationService` though.

mitchdenny:
> Hmm looks like there's a race somewhere as tests are sometimes failing with duplicate logs. I can repro locally if I run 10,000 iterations. Not sure if the bug is in the `FakeLogger`, `ResourceNotificationService`, or `ResourceNotificationService` though.

I can't repro it myself. Must be super tight timing to make it appear.

DamianEdwards:
> I can't repro it myself. Must be super tight timing to make it appear.

I can repro if I update the test to do 10K iterations. Also repros when not using the new type and just using `RequestNotificationService` directly.

DamianEdwards:
Update to the failing test: it's a race in `ResourceLoggerService`. Latest commit includes a change to manually de-dupe log lines being returned due to the race.

DamianEdwards:
OK this is ready for review again @davidfowl @radical @mitchdenny @eerhardt

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -244,9 +249,28 @@ void Log(LogLine log)
...
+                    {
+                        if (lineCount > backlogSnapshot.Length)
+                        {
+                            // We're beyond the number of lines in the backlog snapshot so no need to check for dupes anymore
+                            batch.Add(log);
+                        }
+                        else
+                        {
+                            // We need to check to ensure we haven't already returned this line via the backlog snapshot we took
+                            if (!backlogSnapshot.Contains(log))
```
radical, line 266: Do IIUC that even if the same exact message string would be logged the `LogLine` instance here would be different because it would always have a monotonically increasing `lineNumber`? And so it wouldn't skip repeated message strings when they are in the snapshot, and on the reader?

Would it be useful to add a test for repeated strings being logged?

I don't know much here so my question might be non-sensical!

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -244,9 +249,28 @@ void Log(LogLine log)
...
             {
-                await foreach (var entry in channel.GetBatchesAsync(cancellationToken: cancellationToken).ConfigureAwait(false))
+                var lineCount = 0;
+                while (await channel.Reader.WaitToReadAsync(cancellationToken).ConfigureAwait(false))
                 {
-                    yield return entry;
+                    var batch = new List<LogLine>();
+                    while (channel.Reader.TryRead(out var log))
+                    {
+                        if (lineCount > backlogSnapshot.Length)
```
radical, line 258: Do IIUC that each logged line, even if it showing up in the snapshot or the channel here, would have the same `log.LineNumber`? And can we expect these to be in order? If so, maybe we 
can use that fact to decide when we are "past", like if the `log.LineNumber > backlogSnapshot[^1].LineNumber`, and thus avoid `backlogSnapshot.Contains`?

I'm super new to the code, jfyi! 😬

DamianEdwards: Yeah I attempted an approach like that first but had issues with it while this worked. Admittedly I was still tracking down the root cause at that point and then @eerhardt made the breakthrough and this was the logic we landed on.

radical (approved):
Whatever bits that I could review LGTM! 👍

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
eerhardt: > Do IIUC that even if the same exact message string would be logged the LogLine instance here would be different because it would always have a monotonically increasing lineNumber? And so it wouldn't skip repeated message strings when they are in the snapshot, and on the reader?

Correct. Because we are using `record` for LogLine, it automatically gets an `Equals` override that compares all of its properties for equality. So if the line numbers are different, they won't equal.

eerhardt: If we thought we needed a faster check here, instead of `Contains`, I guess we could just check for `LineNumber`, or even check if the `LineNumber` is greater than the `backlogSnapshot.Length`. But that might be premature optimization. The current code seems sufficient.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
eerhardt, line 258: We might want to capture `backlogSnapshot.Length` in a variable so `backlogSnapshot` can get GC'd earlier. As it stands, it won't be available for collection until the log stops.

src/Aspire.ProjectTemplates/templates/aspire-mstest/IntegrationTest1.cs
```diff
@@ -25,9 +25,12 @@ public class IntegrationTest1
     //         clientBuilder.AddStandardResilienceHandler();
     //     });
     //     await using var app = await appHost.BuildAsync();
+    //     var resourceNotificationService = app.Services.GetRequiredService<ResourceNotificationService>();
     //     await app.StartAsync();
 
     //     // Act
+    //     using var waitForResourceCts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+    //     await resourceNotificationService.WaitForResourceAsync("webfrontend", KnownResourceStates.Running, waitForResourceCts.Token);
```
eerhardt, line 33:
```suggestion
    //     await resourceNotificationService.WaitForResourceAsync("webfrontend", KnownResourceStates.Running).WaitAsync(TimeSpan.FromSeconds(30));
```

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/Aspire-StarterApplication.1.Tests.csproj
```diff
@@ -30,10 +30,14 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Using Include="System.Net" />
```
eerhardt, line 33: What's this for?

tests/Aspire.Hosting.Testing.Tests/Aspire.Hosting.Testing.Tests.csproj
```diff
@@ -23,4 +23,8 @@
     <ProjectReference Include="..\TestingAppHost1\TestingAppHost1.AppHost\TestingAppHost1.AppHost.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" />
```
eerhardt, line 27: Are we using this?

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
DamianEdwards: Doesn't it remain in scope due to the `Contains` call anyway?

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/Aspire-StarterApplication.1.Tests.csproj
DamianEdwards: `HttpStatusCode.OK` in the test. Not sure why it wasn't there before, I'll recheck.

tests/Aspire.Hosting.Testing.Tests/Aspire.Hosting.Testing.Tests.csproj
DamianEdwards: Not anymore, will remove.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
davidfowl: This code bothers me a lot 😅. The only other thing that came to mind was to populate the channel from the backlog (doing de-duping there). Also deduping based on content not id bothers me

src/Aspire.ProjectTemplates/templates/aspire-starter/Aspire-StarterApplication.1.Tests/Aspire-StarterApplication.1.Tests.csproj
DamianEdwards: Ah adding it meant I could remove all the `using` statements from the test file. Given app hosts will often be using `HttpClient` and asserting against related types it felt worthwhile.

radical:
The `aspire-{testFramework}` template tests are [failing](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-4444-merge-f35314e7d1f4434e98/Aspire.Workload.Tests.Xunit_PerTestFrameworkTemplatesTests/1/console.c56bdd4b.log?helixlogtype=result) because those template test projects fail:
```
 [test-aspire-xunit] Test run for C:\h\w\AA680980\t\testroot\aspire-starter & with.1-aspire-xunit\aspire-starter & with.1-aspire-xunit.aspire-xunitTests\bin\Debug\net8.0\aspire-starter & with.1-aspire-xunit.aspire-xunitTests.dll (.NETCoreApp,Version=v8.0)
 [test-aspire-xunit] Microsoft (R) Test Execution Command Line Tool Version 17.9.0 (x64)
 [test-aspire-xunit] Copyright (c) Microsoft Corporation.  All rights reserved.
 [test-aspire-xunit] 
 [test-aspire-xunit] Starting test execution, please wait...
 [test-aspire-xunit] A total of 1 test files matched the specified pattern.
 [test-aspire-xunit] info: Aspire.Hosting.DistributedApplication[0]
 [test-aspire-xunit]       Aspire version: 8.1.0-ci
 [test-aspire-xunit] info: Aspire.Hosting.DistributedApplication[0]
 [test-aspire-xunit]       Distributed application starting.
 [test-aspire-xunit] info: Aspire.Hosting.DistributedApplication[0]
 [test-aspire-xunit]       Application host directory is: C:\h\w\AA680980\t\testroot\aspire-starter & with.1-aspire-xunit\aspire-starter & with.1-aspire-xunit.AppHost
 [test-aspire-xunit] $ENDPOINTS: {}
 [test-aspire-xunit] info: Aspire.Hosting.DistributedApplication[0]
 [test-aspire-xunit]       Distributed application started. Press Ctrl+C to shut down.
 [test-aspire-xunit] [xUnit.net 00:00:31.75]     aspire_starter___with._1_aspire_xunit.aspire_xunitTests.Tests.IntegrationTest1.GetWebResourceRootReturnsOkStatusCode [FAIL]
 [test-aspire-xunit]   Failed aspire_starter___with._1_aspire_xunit.aspire_xunitTests.Tests.IntegrationTest1.GetWebResourceRootReturnsOkStatusCode [31 s]
 [test-aspire-xunit]   Error Message:
 [test-aspire-xunit]    System.OperationCanceledException : The operation was canceled.
 [test-aspire-xunit]   Stack Trace:
 [test-aspire-xunit]      at System.Threading.Channels.AsyncOperation`1.GetResult(Int16 token)
 [test-aspire-xunit]    at System.Threading.Channels.ChannelReader`1.ReadAllAsync(CancellationToken cancellationToken)+MoveNext()
 [test-aspire-xunit]    at Aspire.Hosting.ApplicationModel.ResourceNotificationService.WatchAsync(CancellationToken cancellationToken)+MoveNext() in /_/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs:line 132
 [test-aspire-xunit]    at Aspire.Hosting.ApplicationModel.ResourceNotificationService.WatchAsync(CancellationToken cancellationToken)+MoveNext() in /_/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs:line 132
 [test-aspire-xunit]    at Aspire.Hosting.ApplicationModel.ResourceNotificationService.WatchAsync(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
 [test-aspire-xunit]    at Aspire.Hosting.ApplicationModel.ResourceNotificationService.WaitForResourceAsync(String resourceName, IEnumerable`1 targetStates, CancellationToken cancellationToken) in /_/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs:line 93
 [test-aspire-xunit]    at Aspire.Hosting.ApplicationModel.ResourceNotificationService.WaitForResourceAsync(String resourceName, IEnumerable`1 targetStates, CancellationToken cancellationToken) in /_/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs:line 93
 [test-aspire-xunit]    at aspire_starter___with._1_aspire_xunit.aspire_xunitTests.Tests.IntegrationTest1.GetWebResourceRootReturnsOkStatusCode() in C:\h\w\AA680980\t\testroot\aspire-starter & with.1-aspire-xunit\aspire-starter & with.1-aspire-xunit.aspire-xunitTests\IntegrationTest1.cs:line 34
 [test-aspire-xunit]    at aspire_starter___with._1_aspire_xunit.aspire_xunitTests.Tests.IntegrationTest1.GetWebResourceRootReturnsOkStatusCode() in C:\h\w\AA680980\t\testroot\aspire-starter & with.1-aspire-xunit\aspire-starter & with.1-aspire-xunit.aspire-xunitTests\IntegrationTest1.cs:line 39
 [test-aspire-xunit] --- End of stack trace from previous location ---
 [test-aspire-xunit] 
 [test-aspire-xunit] Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1, Duration: < 1 ms - aspire-starter & with.1-aspire-xunit.aspire-xunitTests.dll (net8.0)
```

They were expected to fail but with `Assert.Matches("System.ArgumentException.*Resource 'webfrontend' not found.", res.Output)`.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
DamianEdwards: > The only other thing that came to mind was to populate the channel from the backlog (doing de-duping there).

Not quite following your thinking on this one. Ultimately with the current order of things we need to subscribe first before taking the snapshot, so when would populate the channel from the backlog in a way that allows to de-dupe?

> Also deduping based on content not id bothers me

What id? The line number?

DamianEdwards:
@radical which test is that?

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
eerhardt: > Doesn't it remain in scope due to the `Contains` call anyway?

Yeah, I missed that. Good call.

tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs
```diff
@@ -79,7 +79,7 @@ await using (var context = await CreateNewBrowserContextAsync())
         res = await cmd.ExecuteAsync($"test -c {config}");
 
         Assert.True(res.ExitCode != 0, $"Expected the tests project build to fail");
-        Assert.Matches("System.ArgumentException.*Resource 'webfrontend' not found.", res.Output);
+        Assert.Matches("System.InvalidOperationException.*Sequence contains no matching element", res.Output);
```
eerhardt, line 82: 1. (In the existing code) @radical - Why are we expecting the tests to fail?
2. For the updated code, isn't the original exception better? Seems like we are making this worse.

radical: > 1. (In the existing code) @radical - Why are we expecting the tests to fail?

The test creates an `aspire` project, then a `aspire-xunit` project, uncomments the code in the generated test file, and runs that. And that causes the test to fail because the test is expecting a `webfrontend` to exist. So, this what the user would see.

We could add a `webfrontend` to the solution, and patch that up so it gets used in this test. Do we want to do that?

DamianEdwards: We could move the `CreateHttpClient` call to before the `WaitForResourceAsync` call to retain the original exception message and not bother with finding the resource in the `DistributedApplicationModel`.

DamianEdwards: I went ahead and adjusted the order of lines in the test templates so that the original exception is thrown. @radical I do think it would be better to update these tests to add a resource though so the tests actually pass. Perhaps we could just base it on the starter template instead (without the tests project option) in order to get the web project by default?

eerhardt (approved):
LGTM.

Might want to wait to see if @davidfowl doesn't approve of the ResourceLoggerService change.

tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs
eerhardt: I was under the understanding we were validating the scenario @radical describes above. What is the experience when a dev adds the test project, and just runs it. Do they get an understandable error?

DamianEdwards: If you just add the test project and run it you don't get an error. The test here actually uncomments the sample code and then runs the test, which fails because the sample code assumes there's a resource to test against (which the comments say).

ReubenBond (approved):

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
DamianEdwards: > This code bothers me a lot 😅. The only other thing that came to mind was to populate the channel from the backlog (doing de-duping there). Also deduping based on content not id bothers me

OK @davidfowl this is reimplemented based on our discussion now. LMK what you think.

src/Aspire.Hosting.Dapr/IDistributedApplicationComponentBuilderExtensions.cs
```diff
@@ -30,7 +30,6 @@ public static class IDistributedApplicationResourceBuilderExtensions
     /// <param name="builder">The resource builder instance.</param>
     /// <param name="options">Options for configuring the Dapr sidecar, if any.</param>
     /// <returns>The resource builder instance.</returns>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0027:API with optional parameter(s) should have the most parameters amongst its public overloads", Justification = "<Pending>")]
```
mitchdenny, line 33: Seems unrelated?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -4,21 +4,105 @@
...
+    /// This method returns a task that will complete when the resource reaches one of the specified target states. If the resource
+    /// is already in the target state, the method will return immediately.<br/>
+    /// If the resource doesn't reach one of the target states before <paramref name="cancellationToken"/> is signalled, this method
+    /// will throw <see cref="OperationCanceledException"/>.
+    /// </remarks>
+    /// <param name="resourceName">The name of the resource.</param>
+    /// <param name="targetStates">The set of states to wait for the resource to transition to one of. See <see cref="KnownResourceStates"/> for common states.</param>
+    /// <param name="cancellationToken">A cancellation token that cancels the wait operation when signalled.</param>
+    /// <returns>A <see cref="Task{String}"/> representing the wait operation and which of the target states the resource reached.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters",
```
mitchdenny, line 87: Is there a way we can do this without violating these rules? What are the long term implications?

src/Aspire.Hosting.Dapr/IDistributedApplicationComponentBuilderExtensions.cs
DamianEdwards: Hmm not sure. Possibly I got a "suppression not required" diagnostic and acted on it? Will have to check

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
DamianEdwards: I was told to just suppress these but we can choose to expand to a bazillion overloads instead 😁

mitchdenny: OK seems like you've chosen the lesser of two evils.

afscrome:
@DamianEdwards Wonder if the logger ra

mitchdenny (approved):
Looking forward to this going in. I think I've got an immediate use case for it in some tests I'm writing!

src/Aspire.Hosting.Dapr/IDistributedApplicationComponentBuilderExtensions.cs
DamianEdwards: Yep: 
image

radical:
You can merge `main` to get https://github.com/dotnet/aspire/pull/4614 which disables the failing `WithDockerfileTests`.

DamianEdwards:
@davidfowl @ReubenBond @eerhardt @mitchdenny I've reimplemented the `ResourceLoggerService.WatchAsync` logic based on a discussion I had with @davidfowl but it changes the behavior in a fundamental way, i.e. new subscribers won't get the backlog until after the next log is sent since they subscribed, as flushing the backlog was moved to the OnLog subscription method itself. I've updated the tests to account for this but I'd appreciate folks' thoughts on whether this is actually what we want. It definitely makes testing a bit harder (need to coordinate waiting for subs to change before publishing logs and need to publish extra logs to get the backlog) and could lead to situations where no historical logs are shown until the next log is sent, or never shown if another log is never sent (e.g. if the resource has stopped).

radical:
Green!!

mitchdenny:
kicking build again because it failed because of a code coverage upload issue.

davidfowl (approved):

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -224,19 +224,31 @@ public async IAsyncEnumerable<IReadOnlyList<LogLine>> WatchAsync([EnumeratorCanc
...
-                channel.Writer.TryWrite(log);
+                if (flushBacklogSync.IsSet)
+                {
+                    channel.Writer.TryWrite(log);
+                    return;
+                }
+
+                flushBacklogSync.Wait(cancellationToken);
+                // We need to ensure we don't write this log to the channel if it was already in the backlog
+                if (backlogSnapshot?.Contains(log) == false)
```
eerhardt, line 239: Can `backlogSnapshot` ever be `null` here? `flushBacklogSync.Set();` is called after `backlogSnapshot` is set. So if we are waiting for it to be set, `backlogSnapshot` should always be non-null.

DamianEdwards: Correct. I had this throwing when null previously but that seemed extreme. Are you suggesting this is a case we should just bang the nullability warning away?

eerhardt: I guess I personally would just do:

```csharp
Debug.Assert(backlogSnapshot is not null, "backlogSnapshot is always set before flushBacklogSync is set");
if (!backlogSnapshot.Contains(log))
```

(Note that this is really just a nit on the readability of the method. Not critical to change.)

DamianEdwards: Yeah I personally like the `Debug.Assert` approach too but this seems like another example of where we're not super consistent in the repo.

PR #4463, opened 2024-06-11
Filter additionals records from dns srv response

### LINKED ISSUE #4494: Service discovery fails if the SRV response also contains TXT entries
A DNS server can also specify TXT entries in the additionals area in an SRV response. In this case, the resolution of the service will fail. This happens, for example, if you use Consul as a DNS server.

vha-schleupen:
In addition to address-records and cname-records, there can also be other record types in the additional field. This types must be ignored, otherwise the service can not be resolved. Fixes #4494
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4463)


vha-schleupen:
@dotnet-policy-service agree company="Schleupen SE"

sja-schleupen:
@davidfowl @ReubenBond any chance getting this reviewed and merged?

ReubenBond (approved):

PR #4467, opened 2024-06-11
Mark shipped public api as such and add infrastructure for PublicApiAnalyzers

### LINKED ISSUE #4332: Setup a process for how to use PublicApiAnalyzers as part of the release process or remove it
Currently, we use the PublicApiAnalyzers Roslyn analyzer package which generates a bunch of PublicAPI.[Un]Shipped files as part of the build. We currently don't have a process to mark which one of the ones marked unshipped should go to shipped as part of the release. This issue is to track the work of determining whether or not they are useful for our use-case, and if so to incorporate updating them as part of the release process. If we decide we don't want them, then this issue will track the work to remove them.

joperezr:
fixes #4332

cc: @timheuer @eerhardt @mitchdenny 

This is updating all of the PublicAPI.*.txt files to reflect what was shipped in 8.0.1, as well as what has been added since then. This also enables the analyzers on all shipping projects. Finally, it adds some scripts written by @RussKie (thanks!!) that should be run after each release, which will take care of marking all unshipped API as shipped.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4467)


joperezr:
I see @eerhardt suggested in a PR if we could also add package validation, which I agree would be very valuable. Given the changes are around the same area and within the same spirit, I'll add a commit in a sec here that will also enable package validation to ensure we don't break APIs vs a previously shipped package.

joperezr:
@eerhardt please take a look at last commit 5f01a31 as that should take care of enabling package validation.

.gitignore
```diff
@@ -144,3 +144,4 @@ node_modules/
 /playground/**/.gitignore
 /playground/**/azure.yaml
 /playground/**/next-steps.md
+*.svclog
```
mitchdenny, line 147: Do we know where this file comes from?

mitchdenny:
LGTM...thanks @joperezr and @RussKie

mitchdenny (approved):

.gitignore
RussKie: These are generated by the new VS IntPre, I think.

eng/Versions.props
```diff
@@ -6,6 +6,10 @@
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>8.1.0</VersionPrefix>
     <PreReleaseVersionLabel>preview.1</PreReleaseVersionLabel>
+    <!-- When running package validation as part of the build, we want to ensure we didn't break the API against the previous
+    version of each package. The following property points to the package version that should be used as baseline. This should
+    get updated each time packages are shipped. -->
```
RussKie, line 11: The formatting sets off a perfectionists in me
```suggestion
    <!--
      When running package validation as part of the build, we want to ensure we didn't break the API against the previous
      version of each package. The following property points to the package version that should be used as baseline.
      This should get updated each time packages are shipped.
    -->
```

RussKie (approved):
With a minor suggestion

RussKie:
> @eerhardt please take a look at last commit [5f01a31](https://github.com/dotnet/aspire/commit/5f01a31b931c3afc03012c2a417e00a34b9a3041) as that should take care of enabling package validation.

My gut feel is that, whilst the implementation is technically correct, it may present a discoverability challenge - i.e., it could be too easy to miss/forget to remove or update one the "non-standard" definitions.
I almost feel like perhaps we should centralise all package validation-related configuration in one place (or two, it's props and targets files).

src/Components/Directory.Build.props
```diff
@@ -5,6 +5,8 @@
   <PropertyGroup>
```
RussKie, line 5: Could you please update line:3 above to read:
```cs
<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```

joperezr: I don't think that would be equivalent. This file seems like it is intentionally skipping the Directory.Build.props from the parent directory, and instead going to the one two levels above (not sure if that is intentional). That is why I needed to update both this file and the one in one directory above. I'll take your suggestion but point to two directories above to not change meaning here, and we can revisit this in case this was unintentional. cc @eerhardt  in case he knows.

.gitignore
RussKie: image

src/Aspire.Dashboard/Aspire.Dashboard.csproj
```diff
@@ -14,7 +14,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <AddPublicApiAnalyzers>false</AddPublicApiAnalyzers>
+    <UsePublicApiAnalyzers>false</UsePublicApiAnalyzers>
+    <EnablePackageValidation>false</EnablePackageValidation>
```
eerhardt, line 18: @joperezr - should we only EnablePackageValidation on `IsPackable=true` projects?

src/Aspire.Hosting.Azure.WebPubSub/Aspire.Hosting.Azure.WebPubSub.csproj
```diff
@@ -6,6 +6,8 @@
     <PackageTags>aspire hosting azure webpubsub pubsub websocket</PackageTags>
     <Description>Azure WebPubSub resource types for .NET Aspire.</Description>
     <PackageIconFullPath>$(SharedDir)AzureWebPubSub_256x.png</PackageIconFullPath>
+    <!-- Disable package validation as this package hasn't shipped yet. -->
+    <EnablePackageValidation>false</EnablePackageValidation>
```
eerhardt, line 10: How do we track enabling this once these packages do ship?

src/Aspire.Dashboard/Aspire.Dashboard.csproj
joperezr: Yup, that's what we are doing, I think this is needed because the dashboard project is actually packable, but I'll double check.

src/Aspire.Hosting.Azure.WebPubSub/Aspire.Hosting.Azure.WebPubSub.csproj
joperezr: Part of the branding activities for each time we ship new packages will need to be A) to update the baseline package version to point to the package versions that we just released and B) to enable validation on any of the packages that just shipped for the first time.

So essentially just like we have to create a GH release, do branding bumps to use a new minor/patch version, and all those other activities, now we will also want to make sure we do this. I'll document this process and add the checklist of branding things to do per release into our docs in the repo.

src/Components/Directory.Build.props
eerhardt: It wasn't intentional. This file was created before `src/Components/Directory.Build.props` was created, and it used the "old" import pattern of directly reference the parent file.

https://github.com/dotnet/aspire/pull/4573 fixes this.

src/Aspire.Dashboard/Aspire.Dashboard.csproj
eerhardt: It shouldn't be packable. We haven't shipped it since preview2 - https://www.nuget.org/packages/Aspire.Dashboard/.  (we should probably deprecate that package).

It is a standalone web app, so it doesn't make sense to make a nupkg out of it.

Issue #4294, opened 2024-05-24
Break up the TestProject to individual projects per Resource/component


joperezr:
Today, we have a single project which we use for several different component's tests which stands up several containers and services. As discussed offline, we want to break this project apart into individual ones per Resource, which will also allow us to add resource-specific tests to ensure things are working as expected. One example of resource-specific tests, would be to validate that resources that support volumes are working as expected.

@davidfowl has volunteered in taking a first stab at this and separate one resource project to see how it looks like, and we can then parallelize the effort from there.

# Components:
- [x] [Aspire.Hosting.Azure.Tests](https://github.com/dotnet/aspire/pull/5013)
- [x] [Aspire.Hosting.Garnet.Tests](https://github.com/dotnet/aspire/pull/4871)
- [x] [Aspire.Hosting.Kafka.Tests](https://github.com/dotnet/aspire/pull/4910)
- [x] [Aspire.Hosting.MongoDB.Tests](https://github.com/dotnet/aspire/pull/5017)
- [x] [Aspire.Hosting.MySql.Tests](https://github.com/dotnet/aspire/pull/4925)
- [x] [Aspire.Hosting.Nats.Tests](https://github.com/dotnet/aspire/pull/5002)
- [x] [Aspire.Hosting.NodeJs.Tests](https://github.com/dotnet/aspire/pull/5142)
- [x] [Aspire.Hosting.Oracle.Tests](https://github.com/dotnet/aspire/pull/4943)
- [x] [Aspire.Hosting.Postgres.Tests](https://github.com/dotnet/aspire/pull/4862)
- [x] [Aspire.Hosting.Python.Tests](https://github.com/dotnet/aspire/pull/4915)
- [x] [Aspire.Hosting.Qdrant.Tests](https://github.com/dotnet/aspire/pull/4879)
- [x] [Aspire.Hosting.RabbitMQ.Tests](https://github.com/dotnet/aspire/pull/5130)
- [x] [Aspire.Hosting.Redis.Tests](https://github.com/dotnet/aspire/pull/4468)
- [x] [Aspire.Hosting.SqlServer.Tests](https://github.com/dotnet/aspire/pull/5056)
- [x] [Aspire.Hosting.Valkey.Tests](https://github.com/dotnet/aspire/pull/4930)
- [x] [Aspire.Hosting.Elasticsearch.Tests](https://github.com/dotnet/aspire/pull/4430)
[...]

PR #4468, opened 2024-06-11
Split redis tests into separate test assembly

davidfowl:
Initial stab at https://github.com/dotnet/aspire/issues/4294.

I added solution folders for tests:

image

TODO:
- I need to make to move the functional tests for redis.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4468)


src/Aspire.Hosting/Aspire.Hosting.csproj
```diff
@@ -62,6 +62,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="Aspire.Hosting.Tests" />
+    <InternalsVisibleTo Include="Aspire.Hosting.Redis.Tests" />
```
davidfowl, line 65: We should aim to remove this

src/Aspire.Hosting.Redis/Aspire.Hosting.Redis.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="Aspire.Hosting.Tests" />
```
davidfowl, line 20: We should aim to remove this.

src/Aspire.Hosting/Aspire.Hosting.csproj
joperezr: Do we have an issue for tracking that?

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -97,15 +97,15 @@ public async Task WithContainerRegistryUpdatesContainerImageAnnotationsDuringPub
...
+        var redis = builder.AddContainer("redis", "redis");
         builder.Build().Run();
 
         var redisManifest = await ManifestUtils.GetManifest(redis.Resource);
         var expectedManifest = $$"""
             {
               "type": "container.v0",
               "connectionString": "{redis.bindings.tcp.host}:{redis.bindings.tcp.port}",
-              "image": "myprivateregistry.company.com/{{RedisContainerImageTags.Image}}:{{RedisContainerImageTags.Tag}}",
+              "image": "myprivateregistry.company.com/redis:latest",
```
joperezr, line 108: Not sure I'm following this change. Do we build that image as part of our build? If so, what is wrong with having it be strongly typed?

tests/Aspire.Hosting.Tests/Containers/ContainerResourceBuilderTests.cs
```diff
@@ -80,7 +79,7 @@ public void WithImageTagThrowsIfNoImageAnnotation()
         using var builder = TestDistributedApplicationBuilder.Create();
         var container = builder.AddResource(new TestContainerResource("testcontainer"));
 
-        var exception = Assert.Throws<InvalidOperationException>(() => container.WithImageTag(RedisContainerImageTags.Tag));
+        var exception = Assert.Throws<InvalidOperationException>(() => container.WithImageTag("7.1"));
```
joperezr, line 82: Not a fan of having the same number in several places, as that makes it hard to update. What was wrong with the previous approach where the version was in a single place?

tests/Aspire.Hosting.Redis.Tests/XunitAttributes.cs
```diff
@@ -0,0 +1,6 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
joperezr, line 6: NIT: This should be a shared file referenced by all projects as opposed to a file in each test project.

tests/Aspire.Hosting.Tests/Containers/ContainerResourceBuilderTests.cs
davidfowl: This test doesn't need to use redis. I can make it a variable, it doesn't really matter.

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
davidfowl: This isn't a redis test. All of the tests that aren't redis specific I am going to replace.

tests/Aspire.Hosting.Redis.Tests/XunitAttributes.cs
davidfowl: Done

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Extensions.Hosting;
+using StackExchange.Redis;
+using Xunit;
+
+namespace Aspire.Hosting.Redis.Tests;
+
+public class RedisFunctionalTests
+{
+    [Fact]
+    public async Task VerifyRedisResource()
```
davidfowl, line 16: cc @eerhardt @radical @ReubenBond @mitchdenny thoughts?

src/Aspire.Hosting/Aspire.Hosting.csproj
davidfowl: We dont

tests/Aspire.Hosting.Redis.Tests/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+
+  <!-- NOTE: This line is only required because we are using P2P references, not NuGet. It will not exist in real apps. -->
+  <Import Project="../../src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.props" />
```
eerhardt, line 6: Are these Build props/targets files really required? It would be nice if we didn't have to have them in every single hosting test project.

tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <Compile Include="..\Aspire.Hosting.Tests\Utils\ArgumentEvaluator.cs" Link="Utils\ArgumentEvaluator.cs" />
+    <Compile Include="..\Aspire.Hosting.Tests\Utils\EnvironmentVariableEvaluator.cs" Link="Utils\EnvironmentVariableEvaluator.cs" />
+    <Compile Include="..\Aspire.Hosting.Tests\Utils\FileUtil.cs" Link="Utils\FileUtil.cs" />
+    <Compile Include="..\Aspire.Hosting.Tests\Utils\ManifestUtils.cs" Link="Utils\ManifestUtils.cs" />
+    <Compile Include="..\Aspire.Hosting.Tests\Utils\TestDistributedApplicationBuilder.cs" Link="Utils\TestDistributedApplicationBuilder.cs" />
+    <Compile Include="..\Aspire.Hosting.Tests\XunitAttributes.cs" Link="XunitAttributes.cs" />
+    
+    <Content Include="$(RepoRoot)src\Schema\aspire-8.0.json" Link="Schema\aspire-8.0.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
```
eerhardt, line 31: Is this needed?

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
eerhardt: I think this test adds value over top of the other Redis Hosting tests, since it actually connects a client to it.

It seems like a "middle of the ground" test. Not a unit test, but not a full end-to-end functional test like 

https://github.com/dotnet/aspire/blob/8e311f3892178d441ae2f4f55763b2837fd50d2a/tests/Aspire.EndToEnd.Tests/IntegrationServicesTests.cs#L32-L53

It is definitely simpler than the end-to-end tests.

mitchdenny: I like these tests, they are a good smoke test. I do think they need to be broken up though. That way if I've made some changes in "redis" land I can run everything related to Redis just by pointing at a particular test assembly.

tests/Aspire.Hosting.Redis.Tests/Directory.Build.props
mitchdenny: Didn't need it for my live tests for the WithDockerfile changes:

https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Containers.Tests/Aspire.Hosting.Containers.Tests.csproj

tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
mitchdenny: I don't think so, I think this might be copy-pasta. We have schema tests in Aspire.Hosting.Tests that does need this but those tests aren't really Redis specific (although they do call AddRedis to verify that the manifest output complies to the schema).

If we wanted to move those schema tests over then we would need this file.

tests/Aspire.Hosting.Redis.Tests/Directory.Build.props
sebastienros: I changed it locally to remove all these imports and in the end it looks like @mitchdenny 's Containters.Tests project. One thing that changed too is 

```
var volumeAnnotation = redis.Resource.Annotations.OfType<ContainerMountAnnotation>().Single();
Assert.Equal("Aspire.Hosting.Tests-myRedis-data", volumeAnnotation.Source);
```

Where in @davidfowl 's branch `TestDistributedApplicationBuilder` would come from an `Include` and the test would check for `"Aspire.Hosting.Redis.Tests-myRedis-data"`. I believe the include was not really required for this PR and this doesn't matter, but please confirm in case I missed something.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
sebastienros: I don't think it's testing less than the current E2E tests, at least when looking from the Redis POV. The E2E is only creating a distinct web api project to use the component, that's the difference with this local functional test.

Do we still need the Redis portion of the E2E test if so? Or should we extract the current portion of the E2E test into a separate project?

For instance, could we not add more volumes testing with this new functional test project already?

sebastienros:
Added some tests for `WithDataVolume()` for now. I don't know yet how we can clean these once the test is done.

eerhardt:
@sebastienros - it looks like the tests are still failing

src/Aspire.Hosting.Redis/Aspire.Hosting.Redis.csproj
```diff
@@ -22,6 +22,7 @@
...
+    <InternalsVisibleTo Include="Aspire.Hosting.Redis.Tests" />
```
eerhardt, line 25: @sebastienros - can we remove this? What is it needed for?

src/Aspire.Hosting/Aspire.Hosting.csproj
eerhardt: This one seems worse. We shouldn't be using internal stuff from `Aspire.Hosting` in our extension library's tests. What is this needed for?

tests/Aspire.Hosting.Containers.Tests/ContainerResourceBuilderTests.cs
```diff
@@ -54,8 +54,8 @@ public void WithImageMutatesImageNameOfLastAnnotation()
     public void WithImageTagMutatesImageTag()
     {
         using var builder = TestDistributedApplicationBuilder.Create();
-        var redis = builder.AddContainer("redis", "redis").WithImageTag("foo");
-        Assert.Equal("foo", redis.Resource.Annotations.OfType<ContainerImageAnnotation>().Single().Tag);
+        var redis = builder.AddRedis("redis").WithImageTag("7.1");
+        Assert.Equal("7.1", redis.Resource.Annotations.OfType<ContainerImageAnnotation>().Single().Tag);
```
eerhardt, line 58: Should this be reverted? Why it is using Redis in the core "Hosting.Containers" tests?

radical:
`Aspire.Hosting.Redis.Tests` logs show:

`could not create the container	{"Container": {"name":"redis-mcqtmvnk-64491c08"}, "Reconciliation": 3, "error": "docker command 'CreateContainer' returned with non-zero exit code 1: command output: Stdout: '' Stderr: 'Unable to find image 'redis:7.2' locally\nError response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit\n'"}`

tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" IsAspireProjectResource="false" />
+    <ProjectReference Include="..\..\src\Aspire.Hosting.Redis\Aspire.Hosting.Redis.csproj" IsAspireProjectResource="false" />
```
eerhardt, line 9:
```suggestion
    <ProjectReference Include="..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" />
    <ProjectReference Include="..\..\src\Aspire.Hosting.Redis\Aspire.Hosting.Redis.csproj" />
```

Are these needed? This project doesn't have `<IsAspireHost>true</IsAspireHost>`

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -0,0 +1,227 @@
...
+        var db = redisClient.GetDatabase();
+
+        await db.StringSetAsync("key", "value");
+
+        var value = await db.StringGetAsync("key");
+
+        Assert.Equal("value", value);
+    }
+
+    [Fact(Skip = "Debugging")]
```
eerhardt, line 54: Seems like this "Skip"s should be removed.

tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj
```diff
@@ -74,4 +74,8 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Aspire.Hosting.Redis.Tests" />
```
eerhardt, line 78: Do we really need this? It is just tests - why not just make things `public` that we need to be public?

radical:
At least that was the case with the *first* attempt.

I'm not sure about the second one because [this](https://dev.azure.com/dnceng-public/public/_build/results?buildId=728907&view=logs&j=527818ac-28d2-597c-50c9-0c4fc550bd35&t=30520e81-a87b-5c77-38ee-323a50a27929) shows `##[error]Artifact Logs_Build_Linux_Release already exists for build 728907.` for that. That file name needs to append the attempt number.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -0,0 +1,227 @@
...
+    public async Task WithDataVolumeWithCustomPersistenceInterval()
+    {
+        var volumeName = "myvolume2";
+        var snapshotInterval = TimeSpan.FromSeconds(1);
+
+        // Use a volume to do a snapshot save with a custom interval
+
+        await VerifyDataPersistence(
+            options => options.WithDataVolume(volumeName).WithPersistence(snapshotInterval),
+            async redisClient => await Task.Delay(snapshotInterval + TimeSpan.FromSeconds(1))
```
eerhardt, line 79: This seems like it will be flaky - especially on slow machines.

eerhardt (commented):
Maybe we should split this apart and add the Volume/BindMount tests in a separate PR?

src/Aspire.Hosting.Redis/Aspire.Hosting.Redis.csproj
sebastienros: RedisContainerImageTags
RedisCommanderConfigWriterHook

Are use in these tests and internal.
This is why `Aspire.Hosting.Tests` was already listed because these tests are coming from (moved to this new `Aspire.Hosting.Redis.Tests`)

src/Aspire.Hosting/Aspire.Hosting.csproj
sebastienros: I removed it and it's fine now.

tests/Aspire.Hosting.Containers.Tests/ContainerResourceBuilderTests.cs
sebastienros: Reverted

tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
sebastienros: Works as you suggest

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
sebastienros: I was trying to debug why the tests were not working on the CI. Ankit made it clear now (docker registry throttling).

tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj
sebastienros: Done as suggested

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
sebastienros: I am ok to remove it, as long as we agree that the other explicit SAVE call is sufficient to test the functionality since we already have a unit test that ensures the correct `--save` argument is passed with a custom timespan. This test is only validating that the feature actually works in Redis. Or we can add more that "1 more second after the time that Redis should have saved it".

tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs
```diff
@@ -169,7 +171,10 @@ public async Task SingleRedisInstanceProducesCorrectRedisHostsVariable(string co
 
         var commander = builder.Resources.Single(r => r.Name.EndsWith("-commander"));
 
-        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(commander, DistributedApplicationOperation.Run, TestServiceProvider.Instance);
+        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(
```
sebastienros, line 174: Line too long couldn't read the code.

src/Aspire.Hosting/Aspire.Hosting.csproj
sebastienros: Also removed the other one

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
eerhardt: > I don't think it's testing less than the current E2E tests

The part that the current E2E tests cover that aren't covered here is the "WithReference" portion that connects a separate project with a resource. This is being simulated in the current test with:

```csharp
        hb.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
        {
            ["ConnectionStrings:redis"] = await redis.Resource.GetConnectionStringAsync()
        });
```

> Do we still need the Redis portion of the E2E test if so? Or should we extract the current portion of the E2E test into a separate project?

I think once we have all the AppHost resource tests split into separate test assemblies, and FunctionalTests like this one for each resource, we can remove the bulk of the resources from the E2E test, and only keep maybe ~5 core resources being tested there. That way when a new resource/component gets added we don't need to update the E2E test and have it continually grow, like it is doing today.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                var value = await db.StringGetAsync("key");
+
+                Assert.True(value.IsNull);
+            }
+
+            await app.StopAsync();
+        }
+    }
+
+    private static void AttemptDeleteDockerVolume(string volumeName)
```
sebastienros, line 297: We will probably need this in other test projects. @eerhardt Any preference where to move it? Copy it for now?

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -0,0 +1,306 @@
...
+                var redisClient = host.Services.GetRequiredService<IConnectionMultiplexer>();
+
+                var db = redisClient.GetDatabase();
+
+                var value = await db.StringGetAsync("key");
+
+                Assert.True(value.IsNull);
+            }
+
+            await app.StopAsync();
```
sebastienros, line 293: This required with `AttemptDeleteDockerVolume` as otherwise the container is still running until the test is done, and holds a ref to it. Also the two containers exist concurrently without it.

I would expect `app.Dispose` to call `StopAsync` but it's not the case.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
eerhardt: Probably can move it into https://github.com/dotnet/aspire/tree/main/tests/Aspire.Hosting.Tests/Utils. We already are using other Utils in there, so I assume all the other separate test assemblies will use it.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -115,7 +115,7 @@ public static IResourceBuilder<RedisResource> WithDataVolume(this IResourceBuild
     /// Use <see cref="WithPersistence(IResourceBuilder{RedisResource}, TimeSpan?, long)"/> to adjust Redis persistence configuration, e.g.:
     /// <code lang="csharp">
     /// var cache = builder.AddRedis("cache")
-    ///                    .WithDataBindMount()
+    ///                    .WithDataBindMount("myredisdata")
```
eerhardt, line 118: We should do this for the other 2 places where this is copied.

https://github.com/dotnet/aspire/blob/2249f7d4d114784ed4f117a60b1b2963d971bd92/src/Aspire.Hosting.Valkey/ValkeyBuilderExtensions.cs#L105-L106

https://github.com/dotnet/aspire/blob/2249f7d4d114784ed4f117a60b1b2963d971bd92/src/Aspire.Hosting.Garnet/GarnetBuilderExtensions.cs#L98-L99

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
eerhardt: > I would expect app.Dispose to call StopAsync but it's not the case.

I would expect the same thing. @mitchdenny - is this a bug in DistributedApplication?

eerhardt (approved):

mitchdenny:
@sebastienros looks like the restructured tests are failing.

sebastienros:
@mitchdenny are you aware of https://github.com/dotnet/aspire/issues/4786 ?
This is hitting us hard, we are working on it.

mitchdenny:
I'm aware of that, although I think these test failures aren't related (I could be wrong). If they are then feel free to ignore, and obviously #4786 takes priority.

sebastienros:
I think it's related, last time I had the same issue I asked @radical and he showed me the throttling issue. Also these tests worked earlier when Eric triggered a new build (when the quotas got renewed apparently). But I'll check in the log to see if I can find a confirmation.

sebastienros:
Confirmed in the logs:

```
      could not create the container	{"Container": {"name":"redis-fexhgyze-5dedf3af"}, "Reconciliation": 3, "error": "docker command 'CreateContainer' returned with non-zero exit code 1: command output: Stdout: '' Stderr: 'Unable to find image 'redis:7.2' locally\nError response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: 
```

It's showing up first because it's a job before Helix runs, it's the first tests that try to download docker images. Then the Job is terminated.

PR #4484, opened 2024-06-12
Fix Grafana dashboard metric name

NAvanDijk:
The Grafana dashboards were referencing an incorrect metric name. The name should be `http_server_request_duration` instead of `http_server_request_duration_seconds` according to the [documentation](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-aspnetcore#metric-httpserverrequestduration).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4484)

### DIFF
--- a/src/Grafana/dashboards/aspnetcore-endpoint.json
+++ b/src/Grafana/dashboards/aspnetcore-endpoint.json
@@ -210,7 +210,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "histogram_quantile(0.50, sum(rate(http_server_request_duration_seconds_bucket{job=\"$job\", instance=\"$instance\", http_route=\"$route\", http_request_method=\"$method\"}[5m])) by (le))",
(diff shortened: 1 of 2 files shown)


niels013:
@dotnet-policy-service agree company="Pandora Intelligence B.V."

adamint:
@JamesNK

JamesNK:
The names here are based on what the OTEL Prometheus library exporter provides. I assume you are using something else to get metrics?

If we change to the format you propose then metrics via OTEL and Prometheus will break. I think they should stay as they are.

niels013:
@JamesNK 
I see now. Is it then an option to have different dashboard files? For example one for OTEL Prometheus and one using the .NET metric name? I am currently exporting the metrics via the OTEL exporter and collecting it via an OTEL collector. In Grafana I can then choose the metrics with a name that is described in the .NET documentation.

JamesNK:
No, I don't want to duplicate them. It's harder to maintain and I don't know how many people want different names, and what the possible names are. If more people asking for this then we'll re-evaluate based on demand.

PR #4500, opened 2024-06-13
[tests] Enable `Aspire.Hosting.Testing`, and `Aspire.Hosting` tests that use docker to run on the build machine

radical:
There is work in progress to get tests like `Aspire.Hosting.Testing`, and `Aspire.Hosting` running on helix. Meanwhile, we can have the tests run on the build machine, at least for linux which has `docker` installed, by simply removing/updating the test attributes blocking them.

- Remove `[LocalOnlyFact/Theory]` attributes in favor of `[RequiresDocker]`
- Remove unncessary `SkipOnHelix` attributes from `SchemaTests`
- Install devcerts on Linux/build machine
- Fix path construction for `nodeapp` so it correctly works on linux
- Split, and fix node/npm tests to work on CI whenever the relevant tool is available.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4500)


tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
```diff
@@ -28,31 +29,36 @@ public async Task HasEndPoints()
...
         Assert.Contains(appModel.GetProjectResources(), p => p.Name == "myworker1");
     }
 
-    [LocalOnlyFact]
+    [Fact]
+    [RequiresDocker]
     public async Task HttpClientGetTest()
     {
         var httpClient = _app.CreateHttpClient("mywebapp1");
+        await Task.Delay(3000);
```
radical, line 46: @ReubenBond @DamianEdwards This PR enables the hosting.testing tests to run on CI. But two tests fail - `HttpClientGetTest`, and `SelectsFirstLaunchProfile` with connection refused. And it seems to be due to webapp not being ready. I added a `Task.delay` to confirm that, and it does work. But what is the correct way to have the test run only once the webapp is ready?

tests/testproject/TestProject.AppHost/TestProgram.cs
```diff
@@ -65,7 +65,7 @@ public class TestProgram : IDisposable
         {
             // Relative to this project so that it doesn't changed based on
             // where this code is referenced from.
-            var path = Path.Combine(Projects.TestProject_AppHost.ProjectPath, @"..\nodeapp");
+            var path = Path.Combine(Projects.TestProject_AppHost.ProjectPath, "..", "nodeapp");
```
radical, line 68: If the nodeapp fails to start for some reason, then no error is being surfaced. @ReubenBond

tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
```diff
@@ -666,8 +682,11 @@ public async Task ProxylessAndProxiedEndpointBothWorkOnSameResource()
 
         var token = new CancellationTokenSource(TimeSpan.FromMinutes(1)).Token;
 
+        await Task.Delay(5000);
```
radical, line 685: @ReubenBond I added a delay here because the first `client.GetStringAsync($"{httpEndPoint}urls"` returned only the `http` endpoint. But the second invocation got both `http`, and `https`. Does this need to wait for all the endpoints to get allocated?

tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
DamianEdwards: You should configure the `HttpClient` with resilience so that it retries. Once #4444 goes in you'll be able to at least wait for the resource to transition to the "Running" state before attempting to send the request, but you still need the retries.

radical: Should that be added to the `Aspire.Hosting.Testing` api? I added that as an extension method in the test right now.

tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
radical: Should the API surface something (event?) for the endpoints getting allocated?
For this test, we could either use `IDistributedApplicationLifecycleHook.AfterEndpointsAllocatedAsync` or retry a few times in a loop.

tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
DamianEdwards: `HttpClient` default resiliency pipelines aren't composable so we don't add it by default as there'd be no way to remove or modify it. The test templates have code in them that adds resiliency in the test itself.

radical: Look at the template, I added this to the `TestingAppHost1.AppHost/Program.cs`:
```csharp
builder.Services.ConfigureHttpClientDefaults(http =>
{
    http.AddStandardResilienceHandler();
});
```

tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
radical: The tests have an existing class for the hook which I tried to use here - https://gist.github.com/radical/7ef257229178c36f6b72ffd09f7e0b79#file-test-cs-L22-L27

But running the test with CWLs added to print the results from `/urls`:
```
Got urls from http://localhost:1234/urls: ["http://localhost:1234"]

Got urls from https://localhost:57048/urls: ["http://localhost:1234","https://localhost:57051"]
```

The first call is missing the `https` endpoint. Did the `AfterEndpointsAllocatedAsync` get invoked too early? Or am I missing something else?
@ReubenBond @DamianEdwards

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -31,6 +31,15 @@ steps:
     displayName: Build
 
   - ${{ if ne(parameters.skipTests, 'true') }}:
+    - ${{ if ne(parameters.isWindows, 'true') }}:
+      - script: mkdir ${{ parameters.repoArtifactsPath }}/devcert-scripts &&
+                cd ${{ parameters.repoArtifactsPath }}/devcert-scripts &&
+                wget https://raw.githubusercontent.com/BorisWilhelms/create-dotnet-devcert/main/scripts/ubuntu-create-dotnet-devcert.sh &&
+                wget https://raw.githubusercontent.com/BorisWilhelms/create-dotnet-devcert/main/scripts/common.sh &&
+                chmod +x ubuntu-create-dotnet-devcert.sh &&
+                ./ubuntu-create-dotnet-devcert.sh
```
eerhardt, line 40: This seems risky as we can get broken (or compromised) by updates to this repo that we don't own.

tests/TestingAppHost1/TestingAppHost1.AppHost/Program.cs
```diff
@@ -1,10 +1,16 @@
...
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 
 var builder = DistributedApplication.CreateBuilder(args);
 
+builder.Services.ConfigureHttpClientDefaults(http =>
+{
+    http.AddStandardResilienceHandler();
+});
+
```
eerhardt, line 13: What HTTP requests is the AppHost making?

eng/pipelines/templates/BuildAndTest.yml
radical: I opened https://github.com/dotnet/dnceng/issues/3123 to move this to the image itself, so it gets done only once. We have been using this for the helix tests already.

tests/TestingAppHost1/TestingAppHost1.AppHost/Program.cs
radical: Not the AppHost itself, but the tests which are creating the `HttpClient` using that - see https://github.com/dotnet/aspire/pull/4500#discussion_r1638986530

eerhardt: Can we just add this in the tests then? That way the "real app" isn't affected?

radical: Yeah, I'll try that. I wasn't sure what the idiomatic way of doing it was.

eerhardt: In a "normal" test, you would do it here:

```csharp
    {
        var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.AspireApp17_AppHost>();
        // HERE
        await app.StartAsync();
```

tests/Aspire.Hosting.Testing.Tests/DistributedApplicationHttpClientExtensionsForTests.cs
```diff
@@ -0,0 +1,30 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Aspire.Hosting.Testing.Tests;
+
+internal static class DistributedApplicationHttpClientExtensions
```
eerhardt, line 8: (nit) the file name and the class name should match.

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
```
eerhardt, line 2: I can't leave a comment on the empty file above, so I'm doing it here. Looks like the `Extensions.cs` file above is empty.

tests/TestingAppHost1/TestingAppHost1.AppHost/TestingAppHost1.AppHost.csproj
```diff
@@ -9,6 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" />
+
```
eerhardt, line 13:
```suggestion
```

I don't think this is used anymore.

tests/Aspire.Components.Common.Tests/RequiresDockerAttribute.cs
```diff
@@ -22,8 +22,7 @@ public class RequiresDockerAttribute : Attribute, ITraitAttribute
     // - Linux - Local, or CI: always assume that docker is installed
     public static bool IsSupported =>
         !OperatingSystem.IsWindows() ||
-        (Environment.GetEnvironmentVariable("BUILD_BUILDID") is null && // NOT CI - build machine or helix
-            Environment.GetEnvironmentVariable("HELIX_WORKITEM_ROOT") is null);
+        (!PlatformDetection.IsRunningOnBuildMachine && !PlatformDetection.IsRunningOnHelix);
```
eerhardt, line 25:
```suggestion
        !PlatformDetection.IsRunningOnCI;
```

eerhardt (approved):
LGTM. Thanks!

radical:
Added some debug bits, so don't merge before that is removed.

radical:
How did the coverage go down here?!

> ‼️ **Found issues** ‼️
> 
> Project	Coverage Type	Expected	Actual
> Aspire.RabbitMQ.Client	Branch	**80**	77.5 🔻
> Full code coverage report: [dev.azure.com/dnceng-public/public/_build/results?buildId=714974&view=codecoverage-tab](https://dev.azure.com/dnceng-public/public/_build/results?buildId=714974&view=codecoverage-tab)

How did the coverage go down here?!

radical:
The number from the previous `main` [build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=714972&view=codecoverage-tab) is `75%`, so this should be an increase. Though, again, I'm not sure why with this PR. Maybe some extra bits are getting covered in the docker dependent tests.

cc @RussKie

RussKie:
> The number from the previous `main` [build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=714972&view=codecoverage-tab) is `75%`, so this should be an increase. Though, again, I'm not sure why with this PR. Maybe some extra bits are getting covered in the docker dependent tests.

It could be due to the code coverage reporter not correctly reporting the numbers... ¯\\_(ツ)\_/¯ You'd need to look into the actual cobertura.xml.

https://dev.azure.com/dnceng-public/public/_build/results?buildId=714972&view=logs&j=8cf0092d-1f6c-522b-3c27-9f632d33f4f3&t=b92e38fc-0165-5b68-b32b-2eef48c1a225&l=9
image

The thing is - the new version was broken and unusable, and I made numerous attempt to contact the team responsible in the past two years but they wouldn't listen....

PR #4507, opened 2024-06-14
Save password parameter default values to user secrets

### LINKED ISSUE #1151: Automatically persist generated password for all containers with user names and passwords

DamianEdwards:
Adds `UserSecretsParameterDefault` which wraps a `ParameterDefault` and saves the value returned by its `GetDefaultValue` method to the AppHost user secrets.

Updated all resources that use `GeneratedParameterDefault` to generate a default password/apikey to wrap it in `UserSecretsParameterDefault` so that the default value is saved to user secrets.

Fixes #1151

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4507)


davidfowl:
This feels like an opportunity to design a better API for parameter defaults.

DamianEdwards:
What did you have in mind?

playground/TestShop/AppHost/ParameterExtensions.cs
DamianEdwards, line 1: This file is no longer needed as it previously contained the hacked workaround for the lack of this feature.

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
```diff
@@ -24,7 +25,7 @@ public override string GetDefaultValue()
         var configurationKey = $"{ParameterResourceBuilderExtensions.ConfigurationSectionKey}:{parameterName}";
         if (!TrySetUserSecret(applicationName, configurationKey, value))
         {
-            throw new DistributedApplicationException($"Failed to set value for parameter '{parameterName}' in application '{applicationName}' to user secrets.");
+            Debug.WriteLine($"Failed to set value for parameter '{parameterName}' in application '{applicationName}' to user secrets.");
```
davidfowl, line 31: Why?

DamianEdwards: To make it a bit easier to see if it's doesn't succeed during dev. There's no logging at this stage and swallowing blindly just seems bad mmm/'k

mitchdenny:
Couple of things playing around with this locally.

1. Seems to have broken pgadmin ... is it working for you?
2. CatalogDbApp is crashing with an error saying it can't authenticate.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -28,7 +29,7 @@ public static IResourceBuilder<MySqlServerResource> AddMySql(this IDistributedAp
         var passwordParameterName = $"{name}-password";
         var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, passwordParameterName);
 
-        if (passwordParameter.Default is not null)
+        if (builder.Environment.IsDevelopment() && builder.ExecutionContext.IsRunMode && passwordParameter.Default is not null)
```
mitchdenny, line 32: Why does the environment matter?

mitchdenny:
> This feels like an opportunity to design a better API for parameter defaults.

Are you talking about supporting binding to KeyVault etc?

mitchdenny:
@DamianEdwards I'm pretty happy with this. I made a minor change so that the secrets file is formatted when we save it.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
DamianEdwards: User secrets is only added as a configuration provider when the environment is Development

davidfowl:
@eerhardt can you review please

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
DamianEdwards: I removed the environment check @mitchdenny

playground/TestShop/AppHost/ParameterExtensions.cs
```diff
@@ -1,117 +0,0 @@
...
-
-internal static class ParameterExtensions
-{
-    /// <summary>
-    /// Creates a password parameter that in the development environment is generated once and stored in the user secrets store.
-    /// </summary>
-    /// <remarks>
-    /// The password is only stable when in the development environment and the application is running. In all other cases, the password is generated each time.
-    /// </remarks>
-    public static IResourceBuilder<ParameterResource> CreateStablePassword(this IDistributedApplicationBuilder builder, string name,
```
eerhardt, line 18: 🥳

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -25,7 +25,13 @@ public static class MySqlBuilderExtensions
...
     public static IResourceBuilder<MySqlServerResource> AddMySql(this IDistributedApplicationBuilder builder, string name, IResourceBuilder<ParameterResource>? password = null, int? port = null)
     {
-        var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password");
+        var passwordParameterName = $"{name}-password";
+        var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, passwordParameterName);
+
+        if (builder.ExecutionContext.IsRunMode && passwordParameter.Default is not null)
+        {
+            passwordParameter.Default = new UserSecretsParameterDefault(builder.AppHostAssembly, builder.Environment.ApplicationName, passwordParameterName, passwordParameter.Default);
+        }
```
eerhardt, line 34: Would it make sense to extract this into a shared method, so it doesn't need to be copied everywhere?

Maybe we could just make `ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter` do this?

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
eerhardt: Would it be possible for us to get the ability to log here? (obviously not in this PR) It seems like a big hole that we aren't able to log things until the app is built. Especially as we add more functionality earlier in the app. Our only options are throw or eat...

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Wraps a <see cref="ParameterDefault"/> such that the default value is saved to the project's user secrets store.
+/// </summary>
+/// <param name="appHostAssembly">The app host assembly.</param>
+/// <param name="applicationName">The application name.</param>
+/// <param name="parameterName">The parameter name.</param>
+/// <param name="parameterDefault">The parameter with a default value.</param>
+public sealed class UserSecretsParameterDefault(Assembly? appHostAssembly, string applicationName, string parameterName, ParameterDefault parameterDefault)
```
eerhardt, line 19: Since this is public API, do we want to do any argument checking that things like `parameterDefault` are not null?

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Wraps a <see cref="ParameterDefault"/> such that the default value is saved to the project's user secrets store.
+/// </summary>
+/// <param name="appHostAssembly">The app host assembly.</param>
+/// <param name="applicationName">The application name.</param>
+/// <param name="parameterName">The parameter name.</param>
+/// <param name="parameterDefault">The parameter with a default value.</param>
```
eerhardt, line 18:
```suggestion
/// <param name="parameterDefault">The <see cref="ParameterDefault"/> that will produce the default value when it isn't found in the project's user secrets store.</param>
```

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
eerhardt: Another cause for this to not succeed is if `appHostAssembly is null`. Do we need to support that scenario? It feels odd that we would Debug.WriteLine if someone passed null (which we claim to support). We basically wouldn't do anything.

src/Shared/SecretsStore.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Text.Json.Nodes;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.UserSecrets;
+
+namespace Microsoft.Extensions.SecretManager.Tools.Internal;
+
+/// <summary>
+/// Adapted from dotnet user-secrets at https://github.com/dotnet/aspnetcore/blob/482730a4c773ee4b3ae9525186d10999c89b556d/src/Tools/dotnet-user-secrets/src/Internal/SecretsStore.cs
+/// </summary>
+internal class SecretsStore
```
eerhardt, line 15: Since we are putting this in the `Shared` folder, it would be best if we shared it with the other places that manipulate the user secrets:

https://github.com/dotnet/aspire/blob/9dfbea4e3201fc6de8fb42b14d013ac4fca70517/src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs#L164-L216

Maybe in a follow up PR, or log an issue?

src/Shared/SecretsStore.cs
```diff
@@ -0,0 +1,94 @@
...
+        if (_secrets is not null)
+        {
+            foreach (var secret in _secrets.AsEnumerable())
+            {
+                contents[secret.Key] = secret.Value;
+            }
+        }
+
+        // Create a temp file with the correct Unix file mode before moving it to the expected _filePath.
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
eerhardt, line 62:
```suggestion
        if (!OperatingSystem.IsWindows())
```

src/Shared/SecretsStore.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private static Dictionary<string, string?> Load(string secretsFilePath)
+    {
+        return new ConfigurationBuilder()
+            .AddJsonFile(secretsFilePath, optional: true)
+            .Build()
+            .AsEnumerable()
+            .Where(i => i.Value != null)
+            .ToDictionary(i => i.Key, i => i.Value, StringComparer.OrdinalIgnoreCase);
```
eerhardt, line 92: How much does this blow away any comments users have put in the file?

Does it blow up when there are comments in the file?

tests/Aspire.Hosting.Tests/AddParameterTests.cs
```diff
@@ -13,7 +15,7 @@ public class AddParameterTests
     public void ParametersAreHiddenByDefault()
     {
         var appBuilder = DistributedApplication.CreateBuilder();
-        appBuilder.Configuration["Parameters:pass"] = "pass1";
+        appBuilder.Configuration[$"Parameters:pass"] = "pass1";
```
eerhardt, line 18: I'm not sure I understand why these changes were made. This doesn't do anything, right?

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
DamianEdwards: We're super inconsistent with this in the repo right now and last time I brought it up the decision was not to bother adding it 'cos we have nullable annotations. I'm happy to add it but we really should decide and update it everywhere 😄

tests/Aspire.Hosting.Tests/AddParameterTests.cs
DamianEdwards: That's a shadow of a previous commit that was undone. I'll fix.

src/Shared/SecretsStore.cs
DamianEdwards: This is the same code used by `dotnet user-secrets` so I don't think it's worth trying to preserve anything extra 'cos the tool won't.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
DamianEdwards: I thought there was a reason I didn't do that but I'll take another look because I can't remember now :smile:

DamianEdwards: Ah it was because all that API surface is public and based on `GenerateParameterDefault` rather than `ParameterDefault` so it wasn't immediately clean to do. I'll take another look.

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
DamianEdwards: Now I've moved the caller of this down to `ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter` we could choose to make it internal. Thoughts?

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
DamianEdwards: OK I moved this down to inside the call chain for `ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter`. Please take another look.

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
eerhardt: > not to bother adding it 'cos we have nullable annotations

At least in dotnet/runtime, our stance is basically you should never allow your library to throw a NullReferenceException. A big reason is because it is harder to figure out how did the `null` get there. Whereas an ArgNullException catches it earlier.

It's definitely a drawback to primary constructors that you can't check for null, and I can see why the push for `!!` was so great.

> Now I've moved the caller of this down to ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter we could choose to make it internal. Thoughts?

Yeah, let's make it internal for now. We can always expose it later. (can't do the opposite)

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -25,7 +25,8 @@ public static class MySqlBuilderExtensions
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<MySqlServerResource> AddMySql(this IDistributedApplicationBuilder builder, string name, IResourceBuilder<ParameterResource>? password = null, int? port = null)
     {
-        var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password");
+        var passwordParameterName = $"{name}-password";
+        var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, passwordParameterName);
```
eerhardt, line 29: (nit) Can we revert these changes now? That way they don't show up in source control history, since they don't really add value after the latest changes.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -41,6 +42,8 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
     private const string BuilderConstructingEventName = "DistributedApplicationBuilderConstructing";
     private const string BuilderConstructedEventName = "DistributedApplicationBuilderConstructed";
 
+    private readonly Lazy<Assembly?> _appHostAssembly;
```
eerhardt, line 45: Do we need a Lazy here? Would it be better to just have a reference to the `DistributedApplicationOptions` instance as a private field, and then below:

```csharp
    /// <inheritdoc />
    public Assembly? AppHostAssembly => _options.Assembly;
```

tests/Aspire.Hosting.Tests/RabbitMQ/AddRabbitMQTests.cs
```diff
@@ -11,6 +11,27 @@ namespace Aspire.Hosting.Tests.RabbitMQ;
...
+        var rmq = appBuilder.AddRabbitMQ("rmq");
+
+        Assert.IsType<UserSecretsParameterDefault>(rmq.Resource.PasswordParameter.Default);
+    }
+
+    [Fact]
+    public void AddRabbitMQDoesNotAddGeneratedPasswordParameterWithUserSecretsParameterDefaultInPublishMode()
+    {
+        using var appBuilder = TestDistributedApplicationBuilder.Create(DistributedApplicationOperation.Publish);
+        appBuilder.Environment.EnvironmentName = "Test";
```
eerhardt, line 28: (nit) The test is named `InPublishMode()` but the environment name is `Test`. Can we either change the test name to be "NonRunMode()" or change the environment name to "Publish"?

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -184,9 +190,31 @@ public static void ConfigureConnectionStringManifestPublisher(IResourceBuilder<I
...
+    /// <summary>
+    /// Creates a new <see cref="ParameterResource"/> that has a generated value using the <paramref name="parameterDefault"/>.
+    /// </summary>
+    /// <param name="configuration">The application configuration.</param>
+    /// <param name="name">Name of parameter resource.</param>
+    /// <param name="secret">Flag indicating whether the parameter should be regarded as secret.</param>
+    /// <param name="parameterDefault">The <see cref="GenerateParameterDefault"/> that describes how the parameter's value should be generated.</param>
+    /// <returns>The created <see cref="ParameterResource"/>.</returns>
+    public static ParameterResource CreateGeneratedParameter(
+        IConfiguration configuration, string name, bool secret, GenerateParameterDefault parameterDefault)
```
eerhardt, line 212: This is kind of surprising for the overload that takes an `IDistributedApplicationBuilder` to save the value to user secrets, but the overload that just takes an `IConfiguration` doesn't.

Should/can we just make this private until we need to expose it?

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
eerhardt: Nice!

src/Shared/SecretsStore.cs
eerhardt: OK, as long as we don't throw if there are comments in the file, that should be OK. Note that this is probably going to get more usage than the `dotnet user-secrets` gets, since it will be used for all password resources no matter what editor/IDE you are using. (I tend to modify user secrets via VS.)

src/Shared/SecretsStore.cs
DamianEdwards: Logged #4596

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
DamianEdwards: These aren't new methods though. Previously the overload that accepted the builder just forwarded to the one that takes configuration. This was one of the reasons why I avoided doing it here originally as this method chain is all public and adding new optional parameters or overloads wasn't super appealing.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
DamianEdwards: I started out trying to plumb the options that way and landed on this instead for some reason but I can take another look.

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
DamianEdwards: OK I made it internal

src/Aspire.Hosting/ApplicationModel/UserSecretsParameterDefault.cs
DamianEdwards: Logged #4597 to track enabling logging pre-host build.

DamianEdwards: > Another cause for this to not succeed is if appHostAssembly is null. Do we need to support that scenario? 

@eerhardt I made it non-nullable and don't call in now if the appHostAssembly is null for some reason.

eerhardt: Now that it's internal, it is less of a concern. But this sounds good.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
DamianEdwards: ok let me take another look lol

DamianEdwards: OK I removed the overload that takes `IConfiguration`

eerhardt (approved):
LGTM.

Nice work! This is going to solve a lot of problems with container persistence.

src/Shared/SecretsStore.cs
DamianEdwards: I added a test to ensure that it handles a secrets JSON file with a comment in it.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
DamianEdwards: OK did what you suggested.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -55,6 +58,9 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
     /// <inheritdoc />
     public string AppHostDirectory { get; }
 
+    /// <inheritdoc />
+    public Assembly? AppHostAssembly => _options.Assembly;
```
davidfowl, line 62: Do we really need to expose this as public API?

DamianEdwards: We can revert but it's kinda useful to have. I think originally it was required when the user secrets stuff was public and each resource extension method was doing the logic to add the user secrets parameter default, but it could all be internal now.

DamianEdwards: Logged #4612

PR #4519, opened 2024-06-14
Add redis commander for garnet

Zombach:
Continuation of [Add Garnet #3839](https://github.com/dotnet/aspire/pull/3839)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4519)


davidfowl:
Can we do more code sharing here?

Zombach:
> Can we do more code sharing here?
Initially, I wanted to create a library project that could add a redis command and a couple of extension methods, and add it to Redis, Garnet, Valkey.
But I didn't quite understand where exactly (the path) to create this library. If you are satisfied with this option, then I will redo it. And I still need a hint with the location of the new library

timheuer:
How much validation have we done with Redis Commander against Garnet? Despite the claim of being Postgres compliant, at least initially when I tried, there were several things Commander (and other redis tools) weren't able to achieve (exhibited by failed connections, errors on commands, etc.). I didn't document them as I assumed Garnet's claim of 'mostly' compliant was taken to be a bit 'your mileage may vary' -- but perhaps latest versions have improved protocol support?

Zombach:
> How much validation have we done with Redis Commander against Garnet? Despite the claim of being Postgres compliant, at least initially when I tried, there were several things Commander (and other redis tools) weren't able to achieve (exhibited by failed connections, errors on commands, etc.). I didn't document them as I assumed Garnet's claim of 'mostly' compliant was taken to be a bit 'your mileage may vary' -- but perhaps latest versions have improved protocol support?

Well, I made myself such a docker-compose and tested it on it
docker-compose.yml
```yml
version: "3.9"
services:
  redis:
    hostname: redis
    container_name: redis
    image: redis:latest
    ports:
      - 6379:6379

  garnet:
    hostname: garnet
    container_name: garnet
    image: ghcr.io/microsoft/garnet:latest
    ports:
      - 6380:6379

  valkey:
    hostname: valkey
    container_name: valkey
    image: valkey/valkey:latest
    ports:
      - 6381:6379
      
  redis-commander:
    image: rediscommander/redis-commander:latest
    restart: always
    environment:
      REDIS_HOSTS: redis:redis,garnet:garnet,valkey:valkey
    ports:
      - 8081:8081
```

image

There was only one mistake with Garnet. I was unable to add "Stream" via Redis-Commander, it gave me this error
image

And when I used redis-commander even before adding Garnet to Aspire. I couldn't connect at all.

But now after many updates. Almost everything works

mitchdenny:
> Can we do more code sharing here?

I'm not sure we would want to create an interdependency between Aspire.Hosting.Redis and Aspire.Hosting.Garnet. So shared source would be the most logical option with if defs?

Maybe the cure is worse than the disease?

davidfowl:
We can share source, we do this today for other things in the apphost (the volume generator), seems like a cheaper way to share this as well.

PR #4529, opened 2024-06-15
Treat the first endpoint for each scheme as default

### LINKED ISSUE #3786: Any http or https endpoint added to a project resource without a target port map to the default port when published
The manifest publisher [special cases http or https endpoints for project resources](https://github.com/dotnet/aspire/blob/bafcacd4e5e571a7551381e3a734138dd64bf8bc/src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs#L293) and does not attempt to allocate a target port for them. This behavior differs from runtime behavior as each endpoint is unique.

We have this logic to let the tool set the default the port, but instead we should have the ability to mark those endpoints specially so other http endpoints get allocated target ports.

davidebbo:
The idea is that all non-default endpoints should get a target port allocate in the manifest. Only the default gets it omitted.

Fixes #3786.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4529)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ProjectResource.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ProjectResource.cs
@@ -17,4 +17,18 @@ public class ProjectResource(string name) : Resource(name), IResourceWithEnviron
 
     // Track the https endpoint that was added as a default, and should be excluded from the port & kestrel environment
     internal EndpointAnnotation? DefaultHttpsEndpoint { get; set; }
+
+    internal bool IsDefaultEndpoint(EndpointAnnotation endpoint)
+    {
+        // Determine if the endpoint should be treated as the Default endpoint.
+        // Endpoints can come from 3 different sources (in this order):
+        // 1. Kestrel configuration
+        // 2. Default endpoints added by the framework
+        // 3. Explicitly added endpoints
+        // But wherever they come from, we treat the first one as Default (for each scheme).
+        // NOTE: the implementation is a bit inefficient as it iterates over the endpoints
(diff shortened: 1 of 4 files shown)


davidfowl (approved):

src/Aspire.Hosting/ApplicationModel/ProjectResource.cs
```diff
@@ -17,4 +17,18 @@ public class ProjectResource(string name) : Resource(name), IResourceWithEnviron
...
+    internal bool IsDefaultEndpoint(EndpointAnnotation endpoint)
+    {
+        // Determine if the endpoint should be treated as the Default endpoint.
+        // Endpoints can come from 3 different sources (in this order):
+        // 1. Kestrel configuration
+        // 2. Default endpoints added by the framework
+        // 3. Explicitly added endpoints
+        // But wherever they come from, we treat the first one as Default (for each scheme).
+        // NOTE: the implementation is a bit inefficient as it iterates over the endpoints
+        // for each check, but it's not a performance critical path.
```
davidfowl, line 30: This makes me a little uncomfortable. The manifest writer already has the endpoint list, can we pass in both the list of endpoints and the current endpoint. At least we wouldn't be look at all annotations per endpoint.

davidebbo: Ok, optimized it

davidfowl (approved):

davidfowl:
There's a regression here, we only want this logic for http and https schemed endpoints.

PR #4530, opened 2024-06-15
Add ```GetEnvironmentVariableValuesAsync``` method to  ResourceExtensions

### LINKED ISSUE #4464: Add an API to easily extract a resource's environment variables as a dictionary
Resources are configured by way of setting environment variables in AppHost projects, e.g. calling `WithEnvironment`. In some scenarios it would be beneficial to be able to easily extract all the environment variable details for a resource as an `IReadOnlyDictionary<string, string?>` or similar, allowing for the values to be used in other contexts, e.g. injecting into the configuration of a `WebApplicationFactory` instance in a test.

We should add a method that makes it easy to get environment variable values for a resource, e.g.:

```diff
public static class ResourceExtensions
{
    public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullWhen(true)] out IEnumerable<EnvironmentCallbackAnnotation>? environmentVariables)
+   public static Task<IReadOnlyDictionary<string, string?>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource);
}
```

Alirexaa:
close: #4464 

This change is based on this [EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync](https://github.com/dotnet/aspire/blob/3a974afc395f8f144ebb0cf8a116bd7038970297/tests/Aspire.Hosting.Tests/Utils/EnvironmentVariableEvaluator.cs#L8) method
with little change.

the difference between the issue Proposed API and this commit:

The issue Proposed API :
```
public static class ResourceExtensions
{
   public static Task<IReadOnlyDictionary<string, string?>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource);
}
```

The commit Proposed API :
```
public static class ResourceExtensions
{
  public static async ValueTask<IReadOnlyDictionary<string, string>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource,
    DistributedApplicationOperation applicationOperation = DistributedApplicationOperation.Run);
}
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4530)


src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,49 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
         return TryGetAnnotationsOfType(resource, out environmentVariables);
     }
 
+    /// <summary>
+    ///  Get the environment variables from the given resource.
+    /// </summary>
+    /// <param name="resource">The resource to get the environment variables from.</param>
+    /// <param name="applicationOperation">The context in which the AppHost is being executed</param>
+    /// <returns>The environment variables retrieved from the resource.</returns>
```
mitchdenny, line 73: Can we get some `<remarks>` added which cover off where this method is useful and considerations around the sync nature. As well as one or two `<example>` sections which cover off its usage.

I'm not sure this needs its own conceptual document, but the following could be expanded:

https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/testing

Could you create a docs issue to capture that?

mitchdenny:
@Alirexaa this is looking good. Just want to make sure we capture the doc requirement and make sure we expand the XML docs a bit for this change.

We need some covering tests as well.

mitchdenny:
@Alirexaa just to be clear we want the XML doc comments in this PR :)

Alirexaa:
> @Alirexaa just to be clear we want the XML doc comments in this PR :)

sure :smile:

> Could you create a docs issue to capture that?

@mitchdenny, what did you mean by that?

DamianEdwards:
@Alirexaa 

> @mitchdenny, what did you mean by that?

Just that can you please log an issue in dotnet/docs-aspire to ensure the new documentation requirement is captured.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -68,11 +68,40 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
     /// <summary>
     ///  Get the environment variables from the given resource.
     /// </summary>
+    /// <remarks>
+    /// This method is useful when you want to make sure the environment variables are added properly to resources, mostly in test situations.
+    /// This method has asynchronous behavior if <paramref name = "applicationOperation" /> be <see cref="DistributedApplicationOperation.Run"/>
```
DamianEdwards, line 73:
```suggestion
    /// This method has asynchronous behavior when <paramref name = "applicationOperation" /> is <see cref="DistributedApplicationOperation.Run"/>
```

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -68,11 +68,40 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
     /// <returns>The environment variables retrieved from the resource.</returns>
+    /// <example>
+    /// <code>
+    /// using var builder = TestDistributedApplicationBuilder.Create();
+    ///
+    /// builder.AddMongoDB("mongo")
+    ///      .WithEndpoint("tcp", e => e.AllocatedEndpoint = new AllocatedEndpoint(e, "localhost", 3000, containerHost))
+    ///      .WithMongoExpress();
+    ///
+    ///   var env = await mongoExpress.GetEnvironmentVariableValuesAsync();
```
DamianEdwards, line 87: nit: seems indenting got messed up here

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -68,11 +68,40 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    /// This method is useful when you want to make sure the environment variables are added properly to resources, mostly in test situations.
+    /// This method has asynchronous behavior if <paramref name = "applicationOperation" /> be <see cref="DistributedApplicationOperation.Run"/>
+    /// and environment variables were provided from <see cref="IValueProvider"/> otherwise it will be synchronous.
+    /// </remarks>
     /// <param name="resource">The resource to get the environment variables from.</param>
     /// <param name="applicationOperation">The context in which the AppHost is being executed</param>
     /// <returns>The environment variables retrieved from the resource.</returns>
+    /// <example>
+    /// <code>
+    /// using var builder = TestDistributedApplicationBuilder.Create();
```
DamianEdwards, line 81: The `TestDistributedApplicationBuilder` type is in this repo only. For the example here you should use `DistributedApplicationTestingBuilder.CreateAsync()`

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,78 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+            }
+
+            foreach (var (key, expr) in config)
+            {
+                var value = (applicationOperation, expr) switch
+                {
+                    (_, string s) => s,
+                    (DistributedApplicationOperation.Run, IValueProvider provider) => await provider.GetValueAsync().ConfigureAwait(false),
+                    (DistributedApplicationOperation.Publish, IManifestExpressionProvider provider) => provider.ValueExpression,
+                    (_, null) => null,
```
radical, line 126: Tests needed for these cases. From code coverage:
Screenshot 2024-06-27 at 16 00 23

JamesNK:
@Alirexaa We have a deadline for 8.1 coming in a couple of weeks. Do you have time to look at the changes? If not, I can help out by addressing some feedback on your branch.

Alirexaa:
> @Alirexaa We have a deadline for 8.1 coming in a couple of weeks and we'd like to include this feature. Do you have time to look at the changes soon? If not, I can help out by addressing some feedback on your branch.

Yes, I have time, but I need some examples to write tests for the following cases :

Alirexaa:
@JamesNK, please take a look.

mitchdenny:
@DamianEdwards @Alirexaa @JamesNK @ReubenBond I've been thinking about this method. At the moment it is in `Aspire.Hosting` but I think given its most common usage scenario it should probably be in `Aspire.Hosting.Testing`.

If we keep it in `Aspire.Hosting` we'd probably want it to extend `IResourceBuilder<T> where T: IResourceWithEnviroment` rather `IResourceBuilderWithEnvironment` directly. Additionally, we'd probably want to remove the operation argument since this would be controlled by the app builder.

@Alirexaa looking at the code coverage on my local run it looks like you figured out how to hit those branches (except the null case and unsupported type case ...)

Alirexaa:
> @Alirexaa looking at the code coverage on my local run it looks like you figured out how to hit those branches (except the null case and unsupported type case ...)

Do we need code coverage for null case and unsupported type case?

JamesNK:
> @DamianEdwards @Alirexaa @JamesNK @ReubenBond I've been thinking about this method. At the moment it is in `Aspire.Hosting` but I think given its most common usage scenario it should probably be in `Aspire.Hosting.Testing`.

Doesn't this basically already exist in `Aspire.Hosting.Testing`? [EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync](https://github.com/dotnet/aspire/blob/3a974afc395f8f144ebb0cf8a116bd7038970297/tests/Aspire.Hosting.Tests/Utils/EnvironmentVariableEvaluator.cs#L8)

mitchdenny:
This is a good point!

mitchdenny:
Upon reflection I think that this API is fine the way it is.

Its most likely to be used in something like a lifecycle hook where you don't have access to a builder, but you can resolve the execution context from DI. So, you would pass the operation in.

mitchdenny (approved):

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,78 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    /// <returns>The environment variables retrieved from the resource.</returns>
+    /// <example>
+    /// <code>
+    /// using var builder = await DistributedApplicationTestingBuilder.CreateAsync&lt;SampleShop_AppHost&gt;();
+    ///
+    /// builder.AddMongoDB("mongo")
+    ///     .WithEndpoint("tcp", e => e.AllocatedEndpoint = new AllocatedEndpoint(e, "localhost", 3000, containerHost))
+    ///     .WithMongoExpress();
+    ///
+    /// var env = await mongoExpress.GetEnvironmentVariableValuesAsync();
```
eerhardt, line 87: Does this code build?

1. What is the `mongoExpress` variable? I don't see it defined in the sample.
2. The result of `DistributedApplicationTestingBuilder.CreateAsync` is an `IDistributedApplicationTestingBuilder` which doesn't have an `.AddMongoDB` extension method.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,78 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    ///         },
+    ///         e =>
+    ///         {
+    ///             Assert.Equal("ME_CONFIG_BASICAUTH", e.Key);
+    ///             Assert.Equal("false", e.Value);
+    ///         });
+    /// </code>
+    /// </example>
+
+    public static async ValueTask<IReadOnlyDictionary<string, string>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource,
```
eerhardt, line 103: Does this need to return an IReadOnlyDictionary? Can it just be a Dictionary?

Since this is a new object every time the method is called, and no one else references it, what is the point of returning a read-only object? If the caller wants to modify/add/remove from the dictionary, they will need to make a copy of it. This seems more like a `.ToDictionary()` method, which returns a copy.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
mitchdenny: Yeah, I think a more realistic scenario for this API usage is actually inside a lifecycle hook. It's hard to show that without a much more detailed sample.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,88 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    ///         },
+    ///         env =>
+    ///         {
+    ///             Assert.Equal("ELASTIC_PASSWORD", env.Key);
+    ///             Assert.False(string.IsNullOrEmpty(env.Value));
+    ///         });
+/// </code>
+/// </example>
+
+public static async ValueTask<Dictionary<string, string>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource,
```
radical, line 113:
```suggestion
    public static async ValueTask<Dictionary<string, string>> GetEnvironmentVariableValuesAsync(this IResourceWithEnvironment resource,
```

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,88 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    /// Using <see cref="GetEnvironmentVariableValuesAsync(IResourceWithEnvironment, DistributedApplicationOperation)"/> inside
+    /// a unit test to validate environment variable values.
+    /// <code>
+    /// var builder = DistributedApplication.CreateBuilder();
+    /// var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
+    ///  .WithEnvironment("discovery.type", "single-node")
+    ///  .WithEnvironment("xpack.security.enabled", "true")
+    ///  .WithEnvironment(context =>
+    ///  {
+    ///      context.EnvironmentVariables["ELASTIC_PASSWORD"] = "123456";
```
radical, line 89: Maybe this so credscan does not complain:
```suggestion
    ///      context.EnvironmentVariables["ELASTIC_PASSWORD"] = "p@ssw0rd1";
```

Or we can add this to `./.config/CredScanSuppressions.json`.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -65,6 +65,88 @@ public static bool TryGetEnvironmentVariables(this IResource resource, [NotNullW
...
+    ///             Assert.Equal("xpack.security.enabled", env.Key);
+    ///             Assert.Equal("true", env.Value);
+    ///         },
+    ///         env =>
+    ///         {
+    ///             Assert.Equal("ELASTIC_PASSWORD", env.Key);
+    ///             Assert.False(string.IsNullOrEmpty(env.Value));
+    ///         });
+/// </code>
+/// </example>
```
radical, line 111:
```suggestion
    /// </code>
    /// </example>
```

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+            },
+            env =>
+            {
+                Assert.Equal("xpack.security.enabled", env.Key);
+                Assert.Equal("true", env.Value);
+            },
+            env =>
+            {
+                Assert.Equal("{ElasticPassword.value}", env.Value);
+                Assert.False(string.IsNullOrEmpty(env.Value));
```
eerhardt, line 123: We should verify the Value is the correct parameter reference in publish mode, not just that it isn't null or empty.

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+            },
+            env =>
+            {
+                Assert.Equal("xpack.security.enabled", env.Key);
+                Assert.Equal("true", env.Value);
+            },
+            env =>
+            {
+                Assert.Equal("ELASTIC_PASSWORD", env.Key);
+                Assert.False(string.IsNullOrEmpty(env.Value));
```
eerhardt, line 57: Why isn't this

```suggestion
                Assert.Equal("123456", env.Value);
```

?

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+    [Fact]
+    public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesInRunMode()
+    {
+        var builder = DistributedApplication.CreateBuilder();
+        var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
+         .WithEnvironment("discovery.type", "single-node")
+         .WithEnvironment("xpack.security.enabled", "true")
+         .WithEnvironment(context =>
+         {
+             context.EnvironmentVariables["ELASTIC_PASSWORD"] = "123456";
```
radical, line 38: To avoid credscan failure:
```suggestion
             context.EnvironmentVariables["ELASTIC_PASSWORD"] = "p@ssw0rd1";
```

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
radical, line 57: Should this check for the contents of the string too?

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+                Assert.Equal("ELASTIC_PASSWORD", env.Key);
+                Assert.False(string.IsNullOrEmpty(env.Value));
+            });
+    }
+
+    [Fact]
+    public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesUsingValueProviderInRunMode()
+    {
+        var builder = DistributedApplication.CreateBuilder();
+        builder.Configuration["Parameters:ElasticPassword"] = "123456";
```
radical, line 65:
```suggestion
        builder.Configuration["Parameters:ElasticPassword"] = "p@ssw0rd1";
```

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+                Assert.Equal("ELASTIC_PASSWORD", env.Key);
+                Assert.False(string.IsNullOrEmpty(env.Value));
+            });
+    }
+
+    [Fact]
+    public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesUsingManifestExpressionProviderInPublishMode()
+    {
+        var builder = DistributedApplication.CreateBuilder();
+        builder.Configuration["Parameters:ElasticPassword"] = "123456";
```
radical, line 98:
```suggestion
        builder.Configuration["Parameters:ElasticPassword"] = "p@ssw0rd1";
```

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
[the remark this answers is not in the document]
mitchdenny: Fair, good point.

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
mitchdenny: Removed password entirely. Its not important what values we assert for the example.

eerhardt: These test comments aren't resolved yet.

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
eerhardt: Reactivating. This hasn't been resolved yet.

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
mitchdenny: Fixed.

mitchdenny:
@joperezr we are going to want to get this one in for 8.1

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -25,4 +25,102 @@ public void TryGetContainerImageNameReturnsCorrectFormatWhenShaNotSupplied()
...
+    [Fact]
+    public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesInRunMode()
+    {
+        var builder = DistributedApplication.CreateBuilder();
+        var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
+         .WithEnvironment("discovery.type", "single-node")
+         .WithEnvironment("xpack.security.enabled", "true")
+         .WithEnvironment(context =>
+         {
+             context.EnvironmentVariables["ELASTIC_PASSWORD"] = "p@ssw0rd1";
```
eerhardt, line 38: I think we are still going to get cred scan failures here, but we can try it.

eerhardt (approved):
LGTM

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
mitchdenny: Will see. @radical was saying that they were in a suppression list.

eerhardt: There are 2 cred scan checks, and my understanding is that only one of them respects the suppression list.

Issue #4522, opened 2024-06-15
Route `docker build` log stream to dashboard console for resource.


mitchdenny:
To help make `docker build` failures more digestible we want to route them through to the dashboard console (in addition to showing in the log outputs for the AppHost).

mitchdenny:
Follow on from https://github.com/dotnet/aspire/pull/4312

mitchdenny:
This is currently working however we have some out of order logs being displayed. Trying to figure out where the the sequence is being broken.

PR #4535, opened 2024-06-17
Log stream the build and pull.

mitchdenny:
This PR addresses #4522 

This is still a draft because we have a few issues:

1. `docker build/pull` output is showing up as stderr.
2. The output isn't actually streaming ... seems to arrive all at once.
3. We probably want to tweak the way we process the streams to make sure it is impossible to interleave `build/pull` output with regular logs.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4535)


mitchdenny:
@danegsta it seems like the output from `docker build` is coming through stderr ... and it looks like it isn't streaming.

Based on what I'm observing it looks like you are observing the error code and piping into stderr the entire output?

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
```diff
@@ -25,6 +25,8 @@ public async IAsyncEnumerator<LogEntryList> GetAsyncEnumerator(CancellationToken
         }
 
         var timestamps = resource is Container; // Timestamps are available only for Containers as of Aspire P5.
+        var startupStdoutStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdOut, follow: true, timestamps: timestamps, cancellationToken).ConfigureAwait(false);
```
karolz-ms, line 28: There are no startup logs for Executables at the moment. Asking for them (for Executables) just slows things down a bit.

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
```diff
@@ -35,13 +37,19 @@ public async IAsyncEnumerator<LogEntryList> GetAsyncEnumerator(CancellationToken
...
+        var startupStdoutStreamTask = Task.Run(() => StreamLogsAsync(startupStdoutStream, isError: false), cancellationToken);
+        var startupStderrStreamTask = Task.Run(() => StreamLogsAsync(startupStderrStream, isError: true), cancellationToken);
         var stdoutStreamTask = Task.Run(() => StreamLogsAsync(stdoutStream, isError: false), cancellationToken);
         var stderrStreamTask = Task.Run(() => StreamLogsAsync(stderrStream, isError: true), cancellationToken);
 
         // End the enumeration when both streams have been read to completion.
 
         async Task WaitForStreamsToCompleteAsync()
         {
+            // We want all the startup streams to be completed before we render anything
```
karolz-ms, line 49: This method just waits for the tasks to complete, it does not affect how output is consumed in any way, so it is completely fine to add the startup tasks to the list and do a single `WaitAll()` call.

karolz-ms (changes requested):

karolz-ms:
> @danegsta it seems like the output from `docker build` is coming through stderr ... and it looks like it isn't streaming.

I will let David investigate but it might be a limitation of Docker/Podman CLI in the sense that if they detect they do not write to actual console, with its control character capabilities, they buffer the output until the build process is done. Just a theory. Because we do stream whatever is coming from the build command, at least that is the intent.

`stderr` is not for "errors only". Stderr is for human-readable content. Stdout is for machine-readable content. That is the original Unix design. Sadly it has not been followed much. But sending diagnostic/informational output to stderr is the right thing to do.

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
mitchdenny: So it is possible to get interleaved output. I probably need to not start the streaming of the stdout and stderr streams until the startup streams have completed.

In practice it seems to only happen if you are setting breakpoints in the code here to debug things. But the fact that interleaving can occur makes me think we should fix it :)

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
mitchdenny: Fixed

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
mitchdenny: I've reverted it back to where it was. I've decided that whilst this is a problem in theory, its not a problem in practice. My main goal in this PR is to get the build/pull logs showing.

It would be great if you and @danegsta can get at getting the build logs streaming so folks can get some kind of visual queue as to what is happening (so it doesn't look like it has hung).

mitchdenny:
@karolz-ms could you reset your vote?

karolz-ms (approved):

playground/withdockerfile/WithDockerfile.AppHost/Program.cs
```diff
@@ -12,15 +12,6 @@
...
 
-builder.AddRedis("vanillaredis").WithRedisCommander();
-
-builder.AddRedis("spicyredis")
-       .WithDockerfile("spicyredis") // This overrides the port that we listen on.
-       .WithEndpoint("tcp", (endpoint) =>
-       {
-           endpoint.TargetPort = 6380; // This fixes the app model so it still works.
-       }).WithRedisCommander();
-
```
eerhardt, line 23: Why are we removing this?

eerhardt (commented):
Are there any tests that can be written for this?

danegsta:
@mitchdenny I've got a small change to DCP that'll land soon to pass `--progress plain` to `docker build`. That should improve the verbosity of build output vs. the default setting when running without a TTY session.

mitchdenny:
> @mitchdenny I've got a small change to DCP that'll land soon to pass `--progress plain` to `docker build`. That should improve the verbosity of build output vs. the default setting when running without a TTY session.

Will that also help with streaming.

danegsta:
> Will that also help with streaming.

It should, but I'll need to test with your branch to make sure.

radical:
> Are there any tests that can be written for this?

Would testing with a simple custom dockerfile be useful here? 1. purge the image; 2. WithDockerFile with some custom messages(?); 3. start; 4. check the logs?
I'm suggesting a custom dockerfile so we could maybe have one that would result in a smaller image, and wouldn't affect any of the other real ones.

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
```diff
@@ -24,25 +24,42 @@ public async IAsyncEnumerator<LogEntryList> GetAsyncEnumerator(CancellationToken
...
             AllowSynchronousContinuations = false,
             SingleReader = true,
             SingleWriter = false
         });
 
+        var streamTasks = new List<Task>();
+
+        var timestamps = resource is Container; // Timestamps are available only for Containers as of Aspire P5.
+
+        if (resource is Container)
```
radical, line 38: nit:
```suggestion
        if (timestamps)
```

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
```diff
@@ -24,25 +24,42 @@ public async IAsyncEnumerator<LogEntryList> GetAsyncEnumerator(CancellationToken
...
+        {
+            var startupStderrStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdErr, follow: true, timestamps: timestamps, cancellationToken).ConfigureAwait(false);
+            var startupStdoutStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdOut, follow: true, timestamps: timestamps, cancellationToken).ConfigureAwait(false);
+
+            var startupStdoutStreamTask = Task.Run(() => StreamLogsAsync(startupStdoutStream, isError: false), cancellationToken);
+            streamTasks.Add(startupStdoutStreamTask);
+
+            var startupStderrStreamTask = Task.Run(() => StreamLogsAsync(startupStderrStream, isError: false), cancellationToken);
```
radical, line 46: Why is this using `isError: false`?

mitchdenny:
> Are there any tests that can be written for this?

Eric I accidentally ignored this comment. I'll look at what our coverage for this change looks like. Mostly this is prepending existing logs so existing test coverage should make sure its working.

But I'll have a think about what if anything extra we can do.

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
mitchdenny: Will look at this as a follow up. You would be surprised how much I agonized over that.

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
mitchdenny: Deliberate. `docker build` outputs vritually everything to stderr which @karolz-ms points out is the right thing for it to do from a unix philosophy point of view. But given it prepends the `err` text to ever line its quite noisy.

I think we might want to have a chat to the dashboard folks about a slightly different visual treatment here.

mitchdenny:
> > Are there any tests that can be written for this?
> 
> Would testing with a simple custom dockerfile be useful here? 1. purge the image; 2. WithDockerFile with some custom messages(?); 3. start; 4. check the logs? I'm suggesting a custom dockerfile so we could maybe have one that would result in a smaller image, and wouldn't affect any of the other real ones.

We do have live tests already which take an image from MCR (which is pretty small) inject some custom files and then fetch the files from the web-server built into the image just to confirm the build args and build secrets are working.

mitchdenny:
https://github.com/dotnet/aspire/issues/4579

PR #4541, opened 2024-06-17
Don't show success status for FailedToStart state, use static helpers for determining resource state

### LINKED ISSUE #1406: Improve code around resource status
Resource status is currently a string. I'm guessing that gives custom resources flexibility to report their own unique statuses. cc @drewnoakes

However, there is code in a number of places that checks the status using string comparisons, e.g. `if (resource.Status == "Ready") { ... }`. This is brittle if statuses ever change, and we don't have an easy way to find out exactly what statuses we care about.

The dashboard should have strongly typed helper methods for checking the status when needed, e.g. `if (resource.GetKnownStatus() == Status.Ready) { ... }`. Or `IsReady(resource.Status)`, etc.

### LINKED ISSUE #4498: Failed to start without an exit code shows up with a success status
See https://github.com/dotnet/aspire/pull/4365#issuecomment-2165866039. This is a regression.

adamint:
Fix #4498 and fix #1406 

This PR converts the static class used to hold known resource states into an enum, adds static helper methods for determining state kind, and fixes #4498 by only showing success for `Finished` state.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4541)

### DIFF
--- a/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
+++ b/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
@@ -1,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Dashboard.Model;
+
+namespace Aspire.Dashboard.Extensions;
+
+public static class ResourceViewModelExtensions
+{
+    internal static KnownResourceState? GetKnownState(this ResourceViewModel resource)
+    {
+        if (Enum.TryParse(resource.State, out KnownResourceState knownState))
+        {
+            return knownState;
+        }
+
+        return null;
+    }
+
+    internal static bool IsHiddenState(this ResourceViewModel resource)
+    {
+        return resource.GetKnownState() == KnownResourceState.Hidden;
+    }
+
+    internal static bool IsRunningState(this ResourceViewModel resource)
+    {
+        return resource.GetKnownState() == KnownResourceState.Running;
+    }
+
+    internal static bool IsStopped(this ResourceViewModel resource)
+    {
+        return resource.GetKnownState() is KnownResourceState.Exited or KnownResourceState.Finished or KnownResourceState.FailedToStart;
+    }
+
+    internal static bool IsStartingOrBuilding(this ResourceViewModel resource)
+    {
+        return resource.GetKnownState() is KnownResourceState.Starting or KnownResourceState.Building;
+    }
+
+    internal static bool HasNoState(this ResourceViewModel resource) => string.IsNullOrEmpty(resource.State);
+}
--- a/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
+++ b/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
@@ -27,6 +27,11 @@ public static class ResourceViewModelExtensions
         return resource.GetKnownState() == KnownResourceState.Running;
     }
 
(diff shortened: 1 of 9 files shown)


src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Dashboard.Model;
+
+namespace Aspire.Dashboard.Extensions;
+
+public static class ResourceViewModelExtensions
+{
+    internal static KnownResourceState? GetKnownState(this ResourceViewModel resource)
```
davidfowl, line 10: Can we put this on the view model instead of parsing it all the time?

adamint: done

src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Dashboard.Model;
+
+namespace Aspire.Dashboard.Extensions;
+
+public static class ResourceViewModelExtensions
```
drewnoakes, line 8: All these methods could be extensions of KnownResourceState rather than the view model.

drewnoakes, line 8: Looks like all members are internal, so consider making the class internal too (or even instead).

tests/Aspire.Dashboard.Tests/ResourceOutgoingPeerResolverTests.cs
```diff
@@ -25,6 +25,7 @@ private static ResourceViewModel CreateResource(string name, string? serviceAddr
             Properties = FrozenDictionary<string, Value>.Empty,
             Urls = servicePort is null || servicePort is null ? [] : [new UrlViewModel(name, new($"http://{serviceAddress}:{servicePort}"), isInternal: false)],
             State = null,
```
drewnoakes, line 27: Should we rename State to StateName?

drewnoakes (approved):

src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
adamint: You're absolutely right, the reason I kept it on ResourceViewModel was because putting an extension on KnownResourceState results in some additional code because of the nullability of `KnownState`. I'd like to keep it for now, but I'd be happy to chat offline about this too!

ie: `r.Value.KnownState?.IsHiddenState() is false` instead of `!r.Value.IsHiddenState()`

adamint: Good call, thank you. I will make this change

tests/Aspire.Dashboard.Tests/ResourceOutgoingPeerResolverTests.cs
adamint: I have no strong opinion on this, but I'm happy to make the change if you would like - I will merge now but please respond if so. The attribute name from gRPC is state

PR #4548, opened 2024-06-17
Fix CredScan false positive.

joperezr:
Repo mirroring is broken currently because CredScan is flagging a false positive secret in the recent changes for FromDockerfile support. This PR fixes that by using a placeholder that is more clearly fake, as well as adding it to the suppressions file. All of the existing suppressions are mostly credentials, which is why I went with defining a new one instead of using an existing one.

cc: @mitchdenny @mmitche 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4548)

### DIFF
--- a/.config/CredScanSuppressions.json
+++ b/.config/CredScanSuppressions.json
@@ -20,6 +20,10 @@
     {
       "placeholder": "p@ssw0rd1",
       "_justification": "This is fake password, used in integration tests."
+    },
+    {
+      "placeholder": "thisIsAFakeSecret",
+      "_justification": "This isn't a real secret, it's used in one of the playground applications for testing purposes."
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


radical (approved):

danmoseley:
I believe the value "placeholder" Is automatically ignored, but this is fine too of course.

joperezr:
> I believe the value "placeholder" Is automatically ignored, but this is fine too of course.

I didn't know that! That said, I think I slightly prefer being more explicit to note that something is fake, so I think I'll keep as is. If you feel strongly though, I'm happy to change.

joperezr:
Merging to unblock 8.1 builds.

danmoseley:
One problem with avoidable suppressions it is it defeats code reviews because I can't tell if something's real without reading the suppression file. Also the suppressions aren't location specific either so if I go add a real secret p@ssw0rd1 into the code, credscan won't flag it because it's suppressed for some place it's used as a fake one.
Not suggesting you change this just making the observation.

PR #4566, opened 2024-06-18
Update resource names throughout app to combine name and UID

### LINKED ISSUE #4302: Trace span names not showing useful titles
I'm on the latest version of Aspire, and when I look at traces, they have the guids as the span service names like this:
image

In demos, I typically see a nice service name here instead. Is there something I need to do so that I have nice service names here?

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/4302

The problem is the dashboard uses the `service.instance.id` for the name when there are replicas. This is great for apps created by Aspire, which have a friendly `service.instance.id` value, such as `catalogservice-mbrpbvo`. It is a problem for apps that use a GUID as a `service.instance.id` value. Most non-Aspire apps use a GUID.

Changes:

* Aspire uses a GUID for `service.instance.id` value.
* The `Resource.Uid` property is fixed to be the UID from DCP.
* Resources and telemetry pages now build a name that combines the `server.name` and the first 7 characters of the `service.instance.id`.

Resources page:
image

Structured logs page:
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4566)


JamesNK:
WIP: The value in the querystring needs some work on console logs and telemetry pages.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -45,7 +46,14 @@ public static string GetResourceName(ResourceViewModel resource, ConcurrentDicti
...
                 {
-                    return resource.Name;
+                    // There are multiple resources with the same display name so they're part of a replica set.
+                    // Need to change the name so it includes an ID to tell them apart.
+                    // DCP automatically includes a suffix on a kind of resource name, however we don't want it
+                    // because we can't share it between resource pages and telemetry pages.
+                    // Instead, combine resource name with the UID.
+                    // Before: catalogservice-mbrpbvo
+                    // After: catalogservice (026d19d)
+                    return $"{resource.DisplayName} ({OtlpHelpers.TruncateString(resource.Uid, maxLength: 7)})";
```
adamint, line 56: why a max length of 7?

src/Aspire.Dashboard/Otlp/Model/OtlpApplication.cs
```diff
@@ -198,7 +198,7 @@ public static string GetResourceName(OtlpApplication app, List<OtlpApplication>
                 count++;
                 if (count >= 2)
                 {
-                    return app.InstanceId;
+                    return $"{item.ApplicationName} ({OtlpHelpers.TruncateString(app.InstanceId, maxLength: 7)})";
```
kvenkatrajan, line 201: nit : Can we add the length of 7 to constants since the ids are referenced in multiple files?

src/Aspire.Dashboard/ResourceService/DashboardClient.cs
```diff
@@ -227,9 +227,17 @@ private void EnsureInitialized()
...
+            {
+                await ConnectAsync().ConfigureAwait(false);
 
-            await WatchResourcesWithRecoveryAsync().ConfigureAwait(false);
+                await WatchResourcesWithRecoveryAsync().ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error load data from the resource service.");
+                throw;
```
kvenkatrajan, line 239: thanks for adding logerror logic 👍

kvenkatrajan (approved):

JamesNK:
Thanks, that's good feedback. I think for now I'll make it simple and concat the service name and service instance id together when there are duplicate service names.

JamesNK:
> We have various applications that make use of the service.namepsace and the Aspire front end dashboard currently has no way of filtering on that.

I think that is a new issue. Create a feature request if this is important to you.

JamesNK:
Replaced by https://github.com/dotnet/aspire/pull/4661

PR #4572, opened 2024-06-18
Support PublishAsDockerFile using custom Dockerfile

### LINKED ISSUE #3825: Support PublishAsDockerFile on project resources
Today [PublishAsDockerFile](https://github.com/dotnet/aspire/blob/ffd52d9a853c820438a6b420acc43d4cc53ca38f/src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs#L48) only supports ExecutableResources, but we should support docker files in projects.

mehrshadshams:
Adding support to PublishAsDockerFile on project resources using default Dockerfile or custom Dockerfile

Resolves #3825 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4572)


mitchdenny:
We just recently merged `WithDockerfile<T>(this IResourceBuilder<T> builder, ...) where T: ContainerResource`. 

I mention it because we are considering deprecating the `dockerfile.v0` resource type in favor of just using the recently added `build` field on the `container.v1` resource type. This means that `PublishAsDockerfile(...)` which currently applies to `ExecutableResource` builders will probably change how it works (and the same would be true for `ProjectResource` builders as well.

Given this it might make more sense to pause this PR until we are a bit further along in that thinking. Its a high priority so it shouldn't be too long.

However, this is a good reminder that we need to make whatever new API we come up with work with project.v0 resources as well!

mehrshadshams:
Thanks for the context @mitchdenny I will close this PR then and look for the guidance.

PR #4591, opened 2024-06-19
Use custom names for dashboard auth and antiforgery cookies

### LINKED ISSUE #4263: Use different cookie name for dashboard
## Introduction

Currently the aspire dashboard uses 2 cookies for authentication
* `.AspNetCore.Cookies`
* `.AspNetCore.Antiforgery.<id>`

These cookie names are the defaults for ASP.NET Core. This means that when making an ASP.NET application that uses cookie auth, there is a cookie name conflict with the dashboard since both will be running on `localhost`. This means that when logging into the hosted application, you are logged out of the dashboard and vice versa.

While this can be worked around by changing the cookie name from the application side, it is not a great experience to have the dashboard conflict with the out of the box cookie auth in ASP.NET core.

## Proposal

Use a different cookie name for the dashboard so its login cookie doesn't conflict with other ASP.NET Core applications by default.

## Notes

This may be related to some of the causes of #3899

adamint:
The cookie names here are changed to @DamianEdwards's suggestions.

Fixes #4263 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4591)


DamianEdwards (approved):
Looks good from naming POV.

PR #4609, opened 2024-06-20
Onboard new inter-branch merge flow [Configuration]

f-alizada:
### Context
Onboarding new merge flow based on the github actions

### Changes Made
Added configuration file. The flow was copied from: https://github.com/dotnet/versions/blob/main/Maestro/subscriptions.json. Currently it will not be read or change any behaviour of the merge flow but will be used by the new workflow.

### References
Documentation https://github.com/dotnet/arcade/blob/main/Documentation/Maestro/New-Inter-Branch-Merge-Approach.md
One pager: https://github.com/dotnet/dnceng/blob/main/Documentation/OnePagers/github-action-inter-branch-merge.md
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4609)


github-merge-flow.jsonc
```diff
@@ -0,0 +1,10 @@
+// IMPORTANT: This file is read by the merge flow from main branch only. 
+{
+    "merge-flow-configurations": {
+        // Automate opening PRs to merge aspire release/8.0-preview5 branch back to release/8.0
+        "release/8.0-preview5":{
+            "MergeToBranch": "release/8.0",
+            "ExtraSwitches": "-QuietComments"
+        }
```
eerhardt, line 8: preview5 is no longer active.

cc @joperezr

f-alizada: Thank you for letting us know! 
That was the only usage of: github-dnceng-branch-merge-pr-generator action in 
https://github.com/dotnet/versions/blob/3144a00684544208a7037c79ddbbecca78c4cc59/Maestro/subscriptions.json#L625-L640
by dotnet/aspire.

joperezr:
Thanks for the PR @f-alizada. As @eerhardt pointed out, we don't really need this flow any longer. We could set one up for release/8.0 -> main, but honestly I don't expect that to be used more than once or twice in the future so we can probably just handle that manually. Does this new process support wildcards? If so, then having one for release/* -> main would be interesting.

f-alizada:
> Thanks for the PR @f-alizada. As @eerhardt pointed out, we don't really need this flow any longer. We could set one up for release/8.0 -> main, but honestly I don't expect that to be used more than once or twice in the future so we can probably just handle that manually.

Thank you for the clarification @joperezr! 
In such scenario let me offboard this repository from legacy merge flow, since there is only one item with deprecated branch :) 

>  Does this new process support wildcards? If so, then having one for release/* -> main would be interesting.

Answer is yes and no :) 
The configuration (mapping of the branches flow) should contain the branch names explicitly like in the current PR, 
To trigger the flow there is a possibility to create the wildcards for branches like in the PR: https://github.com/dotnet/aspire/pull/4610/files

Note:  The workflow file should be presented in the triggered branch. 

Anternatively: If there will be no much usage for current setup, there is always a possibility to merge the config and workflow into the main branch and once the release branches are created and the mapping is setup, the triggers will work.

danmoseley:
@f-alizada what is the status here? I am closing, feel free to reopen if this is still needed but it's very old now.

PR #4622, opened 2024-06-21
[CI] Don't trigger the builds when CODEOWNERS file has a change

radical:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4622)

### DIFF
--- a/eng/pipelines/azure-pipelines-public.yml
+++ b/eng/pipelines/azure-pipelines-public.yml
@@ -8,13 +8,10 @@ trigger:
     include:
       - '*'
     exclude:
+      - '**.md'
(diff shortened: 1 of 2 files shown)


danmoseley (approved):
Anything else we might want that our other repos list here?

radical:
> Anything else we might want that our other repos list here?

Thanks for the question. I looked at https://github.com/dotnet/runtime/blob/main/eng/pipelines/runtime.yml#L13-L21, and updated the list here to exclude all `.md` files. And made the same change for the other triggers for public, and internal pipeline.

PR #4629, opened 2024-06-24
Add support for exemplars in metrics UI

### LINKED ISSUE #336: Feature Idea: Connecting Metrics to Traces
Now that we have both Metrics and Traces in the dashboard, we should consider connecting the two, so I can drill in to data I see in my Metrics graphs, and figure out why something is taking so long.

For example, in the sample app, the first time I click the cart button on the top right, it causes a spike in duration of HTTP requests.

image

It would be great if I could drill in to that request to see what took so long.

One idea is to use [exemplars](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exemplars), which is a feature in OpenTelemetry that allows getting a value in a histogram bucket and being able to see the trace ID for that value, so you can jump to a trace.

cc @davidfowl @JamesNK @samsp-msft

### LINKED ISSUE #4639: Display metrics exemplars in metrics table view
PR https://github.com/dotnet/aspire/pull/4629 adds exemplars to plotly graph. But they're not yet available in the table view.

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/336
Fixes https://github.com/dotnet/aspire/issues/4639

**tldr** example tracing data now shows up on metrics graphs as points. Click on a point to navigate to the trace UI. The $5 name for this feature is [exemplars](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/metrics/exemplars/README.md).

Done:

* Captures exemplars from OTLP metrics data
* Adds plotly trace bound to the tracing points
* Mouse over a point shows a tooltip summary of the tracing data if available. It's possible the metrics data arrives before tracing data. If it's not available yet, then the trace ID is shown.
* Clicking on the trace point navigates to the trace UI, displays the linked trace, and opens the span detail view. 
* Updates OTEL dependency to 1.9.0. Required for exemplar data from .NET

Todo:

* ~Someone can click on an exemplar to view the trace detail before the trace or span is received by the dashboard. This primarily happens because OTEL includes exemplars with metrics immediately but doesn't send a span until it is finished. What should happen? One option is a loading indicator + message somewhere.~ Dialog is displayed on metrics page that data isn't ready. Navigates when data is ready, or can be canceled by user.
* ~By default, opentelemetry-dotnet doesn't send exemplars with metrics for performance reasons. We can either enable this feature by updating the service defaults with new C# configuration or set an OTEL env var in Aspire. I think the OTEL env var should always be set in local development, and I'm undecided about defaulting to on in service defaults. It depends whether we think this should be on in production.~ Set env var in development

Demo:

exemplars

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4629)


JamesNK:
cc @samsp-msft @noahfalk @tarekgh @lmolkova @reyang @davidfowl

mitchdenny:
When you have a lot of data points all on top of one another its going to get hard to select the right one. Are there any user interface patterns for solving this? I know in some CAD systems (which also have the problem of clusters of points/planes/etc overlapping) is to click-hold and get a context menu with all the nearby points of interest to then select.

samsp-msft:
The point of exemplars is to provide a sampling of values, and not on every request. The OTel API has a way for specifying the sampler. We should probably setup this in ServiceDefaults to have a high sample rate during debug, and more sensible for production.

@reyang @cijothomas @codeblanch

cijothomas:
> The point of exemplars is to provide a sampling of values, and not on every request.

That is true. The default in OTel .NET SDK is to collect at most 1 Exemplar per collection for Non-Histograms, and at most N Exemplars per collection for Histograms, where N is the number of histogram buckets. 

UI's can further filter the Exemplars to make UI less cluttered, esp. when multiple instances are reporting Exemplars..

JamesNK:
What do you think about defaulting exemplars to on in production? Good or bad?

Docs say opentelemetry-dotnet defaults them to off, against recommended conventions, because of performance. Do you have more details to help us make a decision?

reyang:
> What do you think about defaulting exemplars to on in production? Good or bad?

Having it on-by-default for histograms seems very useful (and the cost seems reasonable). For counters/gauges, I don't know how many users would want to use exemplars.

> Docs say opentelemetry-dotnet defaults them to off, against recommended conventions, because of performance. Do you have more details to help us make a decision?

Check this https://github.com/open-telemetry/opentelemetry-specification/pull/3870#issuecomment-2066865603

CodeBlanch:
> Having it on-by-default for histograms seems very useful (and the cost seems reasonable).

@JamesNK Just FYI in case you haven't noticed in .NET we have a special feature to enable exemplars just for histograms. Look for `OTEL_DOTNET_EXPERIMENTAL_METRICS_EXEMPLAR_FILTER_HISTOGRAMS` on here: https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk#exemplarfilter

samsp-msft:
> > What do you think about defaulting exemplars to on in production? Good or bad?
> 
> Having it on-by-default for histograms seems very useful (and the cost seems reasonable). For counters/gauges, I don't know how many users would want to use exemplars.
> 
> > Docs say opentelemetry-dotnet defaults them to off, against recommended conventions, because of performance. Do you have more details to help us make a decision?
> 
> Check this [open-telemetry/opentelemetry-specification#3870 (comment)](https://github.com/open-telemetry/opentelemetry-specification/pull/3870#issuecomment-2066865603)

I'm wondering if we should be going with Trace-based exemplars, and adding something smarter for the trace sampling, which currently defaults to 100%. The overhead to metrics for the exemplar is large compared to typical aggregated metric data, but should be small in relation to the payload for each trace.

We probably have too high a default trace collection rate for production usage.

kvenkatrajan:
@JamesNK  thanks for putting this together so quickly..looks great! Did some initial testing around this today and from a data perspective things look great. A few things we would need to review with the UX board on : 

- The size/color of the icon used for the examplar 

- @mitchdenny's comment above on discoverability

> When you have a lot of data points all on top of one another its going to get hard to select the right one. Are there any user interface patterns for solving this? 

Additionally from an accessbility perspective, how do we envision the data being displayed on the table view? Would the user need to filter by http.route to display the link to the span element? Please add an issue for tracking the display in table view.

JamesNK:
> > Having it on-by-default for histograms seems very useful (and the cost seems reasonable).
> 
> @JamesNK Just FYI in case you haven't noticed in .NET we have a special feature to enable exemplars just for histograms. Look for `OTEL_DOTNET_EXPERIMENTAL_METRICS_EXEMPLAR_FILTER_HISTOGRAMS` on here: [open-telemetry/opentelemetry-dotnet@`main`/docs/metrics/customizing-the-sdk#exemplarfilter](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/customizing-the-sdk?rgh-link-date=2024-06-24T17%3A10%3A04Z#exemplarfilter)

@CodeBlanch Optimizing for performance isn't a large concern in development. I think we can set `OTEL_METRICS_EXEMPLAR_FILTER` to `trace_based` based by default.

> I'm wondering if we should be going with Trace-based exemplars, and adding something smarter for the trace sampling, which currently defaults to 100%. The overhead to metrics for the exemplar is large compared to typical aggregated metric data, but should be small in relation to the payload for each trace.

@samsp-msft Aspire sets trace sampling to off in development (i.e. collect all traces). After deployment, it uses whatever is the opentelemetry-dotnet default is for trace sampling.

JamesNK:
> UI's can further filter the Exemplars to make UI less cluttered, esp. when multiple instances are reporting Exemplars..

@cijothomas Thanks for the info! I think some UI filtering will be needed. For example, if someone sets the metrics chart timespan to one day, there could be a huge number of exemplars across that time. Thousands of points squeezed onto a graph won't be very useful.

JamesNK:
> @JamesNK thanks for putting this together so quickly..looks great! Did some initial testing around this today and from a data perspective things look great. A few things we would need to review with the UX board on :
> 
> * The size/color of the icon used for the examplar
> * @mitchdenny's comment above on discoverability
> 
> > When you have a lot of data points all on top of one another its going to get hard to select the right one. Are there any user interface patterns for solving this?
> 
> CC: @leslierichardson95
> 
> Additionally from an accessbility perspective, how do we envision the data being displayed on the table view? Would the user need to filter by http.route to display the link to the span element? Please add an issue for tracking the display in table view.

Issue for table view: https://github.com/dotnet/aspire/issues/4639

Can you start UX review process started and CC me? I haven't done it before so I'd like to learn at the same time.

JamesNK:
All remaining work is done. Please review.

noahfalk:
Its really cool to have exemplars included in there 👍

As someone experienced in the area I can make a good guess what the UI is showing me, but I am wondering if less experienced users will also understand it. This is my guess at a novice user thinking process:

- I see dots on the graph, but what do these dots mean? Normally a dot in a line graph visually is placed on the line and it means the graph has a specific y-value at this x-value. If the dots land on the line I'd probably (incorrectly) guess it means that, but then why aren't the dots at each place the line changes slope? If the dots aren't on the line then what are they?
- OK, I asked my coworker and they explained what exemplars are. If I hover over a dot it tells me the value, but is the popup telling me the value of this particular exemplar or the value of the time-series overall at this point?
- The dots are all close together, if I click on one which one did I actually hit? If I fidget the mouse can click the others?

Asking some less experienced users is probably better than my own guesses though. If these guesses are right I'm no UI expert but I can offer one suggestion that might help:
image

Hovering over or clicking on the "Exemplars" link could give more info about exemplars. Hovering over each diamond would give info about all the exemplars available near that point in time. The hover info could be a small summary like "4 example traces" and then clicking on it goes to new page where each exemplar is a table row with the metric value and clickable links to each trace. Alternately if you can do a little table with links inside the popup then you might be able to show all the exemplars inline without needing an intermediate page.

> What do you think about defaulting exemplars to on in production? Good or bad?

I would think its good as long as the sampling rate on the exemplars gives a total data volume that is a small fraction of what is being used for traces. @reyang @cijothomas, do the exemplar sampling controls let people produce that result even if it is just approximate?

JamesNK:
Seeing the exemplar value height on the graph is useful to pick the high value exemplar.

> The dots are all close together, if I click on one which one did I actually hit? If I fidget the mouse can click the others?

You will always go to the trace whose details are displayed in the popup.

> I see dots on the graph, but what do these dots mean?

I updated the chart to show it in the legend:
image

To be honest I don't think most people will know what an exemplar is though.

> If I hover over a dot it tells me the value, but is the popup telling me the value of this particular exemplar or the value of the time-series overall at this point?

The popup includes the span name. I hope people connect the dots and know that it represents a span. If they don't, clicking on a point (indicated by the pointer cursor on mouse over) takes you to the span, so they'll learn pretty quickly.

noahfalk:
> I updated the chart to show it in the legend:

I think that definitely helps. I still hope we can have some grouping or filtering to avoid dots that are spatially too close together.

reyang:
> > I updated the chart to show it in the legend:
> 
> I think that definitely helps. I still hope we can have some grouping or filtering to avoid dots that are spatially too close together.

+1 on grouping - when the dots are too getting too close, collapse them similar to what a power outage map would do, and allow user to drill down:

image

adamint:
@JamesNK just a note that if we merge this without addressing #4639, there will be a sev2 a11y bug filed on us after the next accessibility review.

samsp-msft:
>I'm not a fan of displaying them just on the x axis. A visual representation of exemplar value height on the graph is useful to pick a high value exemplar.

+1 - the value of exemplars is to be able to look at the outliers and see why they are different from the others. Showing them with a vertical position makes it much easier to understand where that exemplar lives in relation to the overall values.

> > I updated the chart to show it in the legend:
> 
> I think that definitely helps. I still hope we can have some grouping or filtering to avoid dots that are spatially too close together.

If there are too many close together, that probably says you have too many values. I am less concerned about being able to select an individual one on that basis, especially if we have the table view, as you then get the list you are after.

If there are a bunch close together, then we could take you to the table view - but how to take you to a specific set of entries and highlight them is a new UX affordance we don't yet have.

noahfalk:
> Showing them with a vertical position makes it much easier to understand where that exemplar lives in relation to the overall values.

If a counter has a sudden spike, the exemplars presumably all still have value 1. It just happens that there were a lot of counter increments in that time period. This strategy works for histograms but not for all metric types.

> I am less concerned about being able to select an individual one on that basis

I agree, I'm also not that concerned about selecting specific exemplars. I'm more concerned that we shouldn't show too many exemplar data points in the UI if they are overly difficult to select or if the large number of them causes too much visual clutter.

JamesNK:
> If a counter has a sudden spike, the exemplars presumably all still have value 1. It just happens that there were a lot of counter increments in that time period. This strategy works for histograms but not for all metric types.

I just played around with exemplars on non-histogram charts and you're right that it doesn't work as well. An exemplar on the `aspnetcore.route_matched` counter with a value of 1 doesn't mean anything when 10,000 routes have been matched.

They either need to be placed on 0 (easy to do) or on the current y-axis value (harder). The current y-axis seems like it makes more sense.

For now, I've disabled showing exemplars on non-histograms.

JamesNK:
@adamint @noahfalk To kill two birds with one stone, I've added exemplars to the metrics table view. It is both accessible and provides a UI that doesn't have the issue of many points at the same location.

exemplars-table

Directory.Packages.props
```diff
@@ -122,13 +122,13 @@
...
-    <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.8.1" />
-    <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.8.1" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.GrpcNetClient" Version="1.8.0-beta.1" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.8.1" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.8.1" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.8.1" />
+    <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.9.0" />
+    <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.9.0" />
+    <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.9.0" />
+    <PackageVersion Include="OpenTelemetry.Instrumentation.GrpcNetClient" Version="1.9.0-beta.1" />
```
mitchdenny, line 128: Just unlucky I guess ;)

JamesNK: Because it's beta? It's because the grpc semantic conventions aren't stable yet.

JamesNK:
Please review 🙏

reyang (approved):
LGTM, @JamesNK @samsp-msft would it make sense for two of you to show a demo to the OpenTelemetry community during the community/maintainers meeting? I think it will help to raise the awareness and also collect feedback from users.

JamesNK:
What time is it? Might not work for me. Ping us details on teams/email.

noahfalk:
I think the table is very nice 👍 Is there anything that will prevent the graph from becoming cluttered with many exemplars if the app is running under sustained load? It seemed like if P50 stayed low but not zero it might get obscured behind lots of exemplar dots all concentrated near the x-axis.

kvenkatrajan:
[heart]         Kay Venkatrajan reacted to your message:
________________________________

mitchdenny (approved):
My approval is scoped to the app model side of things (very minor changes ... no concerns). Only thing I would suggest is regenerating the sample manifests under playground.

samsp-msft:
> LGTM, @JamesNK @samsp-msft would it make sense for two of you to show a demo to the OpenTelemetry community during the community/maintainers meeting? I think it will help to raise the awareness and also collect feedback from users.

Yes, can do, when is it?

timheuer:
UI feedback: any way to padd numbers and/or align the visuals better? when icons are shown it's better they are more uniform aligned in table views rather than appear to shift position.  Similarly (perhaps more nuanced because the number of exemplars changes) if the button for exemplars could be more uniform-width?

image

JamesNK:
I can give the exemplar button a minimum width so an exemplar number between 1-999 has a consistent width. But at some point the button must grow to handle large numbers so it won't always be consistent.

mitchdenny:
You could right align the cells. That would make the icons line up at least.

timheuer:
> The icon placement is more difficult

is the number always a 3 decimal value? e.g. instead of `0.01` make it `0.010` fixing/padding the decimal len? or move the icons to the other side?

JamesNK:
> @JamesNK can you add the changes to the ServiceDefaults project to enable exemplars, and setting the configuration to enable it just for histograms so it matches with the dashboard implementation.

Will do in a follow up PR - https://github.com/dotnet/aspire/issues/4839

JamesNK:
Comparison of marker sizes.

Size 16 (current):

image

Size 14:

image

Size 12:

image

Size 10:

image

JamesNK:
I think leaving it at 16 is good.
14 feels small but would work.
12 and 10 are too small.

JamesNK:
I'm going to merge as is to unblock other work, and we can bike shed marker sizes and make the change in a follow up PR.

kvenkatrajan:
[celebrate]     Kay Venkatrajan reacted to your message:
________________________________

PR #4636, opened 2024-06-24
[ci] Check that dev-certs are installed on windows

radical:
Prompted by tests failing on windows:
`Aspire.Dashboard.Tests.Integration.DashboardClientAuthTests.ConnectsToResourceService_ApiKey(useHttps: True)`
`Aspire.Dashboard.Tests.Integration.DashboardClientAuthTests.ConnectsToResourceService_Unsecured(useHttps: True)`

.. with ..

```
System.InvalidOperationException : Unable to configure HTTPS endpoint. No server certificate was specified, and the default developer certificate could not be found or is out of date.
To generate a developer certificate run 'dotnet dev-certs https'. To trust the certificate (Windows and macOS only) run 'dotnet dev-certs https --trust'.
For more information on configuring HTTPS see https://go.microsoft.com/fwlink/?linkid=848054.
```

This seems to have failed on a particular build machine. It's not clear why. So this PR adds a check to get more information in case the dev-certs are not installed, or not valid.

Co-authored by: @DamianEdwards 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4636)

### DIFF
--- a/eng/pipelines/templates/BuildAndTest.yml
+++ b/eng/pipelines/templates/BuildAndTest.yml
@@ -41,7 +41,7 @@ steps:
         displayName: Install devcerts
 
     - ${{ if eq(parameters.isWindows, 'true') }}:
-      - script: dotnet dev-certs https --check
+      - script: dotnet dev-certs https --verbose
         displayName: Install devcerts
 
     - script: ${{ parameters.dotnetScript }} dotnet-coverage collect
--- a/eng/pipelines/templates/BuildAndTest.yml
+++ b/eng/pipelines/templates/BuildAndTest.yml
@@ -41,7 +41,7 @@ steps:
         displayName: Install devcerts
 
     - ${{ if eq(parameters.isWindows, 'true') }}:
-      - script: dotnet dev-certs https --trust
+      - script: dotnet dev-certs https --check
         displayName: Install devcerts
 
     - script: ${{ parameters.dotnetScript }} dotnet-coverage collect
--- a/eng/pipelines/templates/BuildAndTest.yml
+++ b/eng/pipelines/templates/BuildAndTest.yml
@@ -41,8 +41,8 @@ steps:
         displayName: Install devcerts
 
     - ${{ if eq(parameters.isWindows, 'true') }}:
-      - script: dotnet dev-certs https --verbose
-        displayName: Install devcerts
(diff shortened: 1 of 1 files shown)


radical:
Hit the failures on `main` - https://dev.azure.com/dnceng-public/public/_build/results?buildId=718671&view=ms.vss-test-web.build-test-results-tab&runId=18009230&paneView=debug&resultId=100049

radical:
Why wasn't this required before though?

eerhardt (approved):

DamianEdwards:
We could try adding a step that validates the cert first and fail the build if it's not found. This can be done by running `dotnet dev-certs https --check`. The .NET SDK adds the dev cert as part of the first-run experience. Has anything changed WRT to how the .NET SDK is setup for the agents recently?

radical:
> We could try adding a step that validates the cert first and fail the build if it's not found. This can be done by running `dotnet dev-certs https --check`. The .NET SDK adds the dev cert as part of the first-run experience. Has anything changed WRT to how the .NET SDK is setup for the agents recently?

`Run the command with both --check and --trust options to ensure that the certificate is not only valid but also trusted.`
.. I'll add `--check` to the command line.

DamianEdwards:
TBC, adding `--check` means it won't actually perform the installation or trust, just the check. I meant if we're not sure why it's not working all of a sudden we could first add a step that just does the check to see what it reports. For example it could be that the cert is there but out of date.

radical:
> TBC, adding `--check` means it won't actually perform the installation or trust, just the check. I meant if we're not sure why it's not working all of a sudden we could first add a step that just does the check to see what it reports. For example it could be that the cert is there but out of date.

I don't think it is worth trying to find that right now, because it happened on only one build so might be specific to a particular machine.
Do you think we need a `--check` as a second step like `dotnet dev-certs https --trust & dotnet dev-certs https --trust --check`? Or would the first part ensure that the cert is not out-of-date and is trusted?

radical:
> A confirmation prompt will be displayed if the certificate was not previously trusted. Click yes on the prompt to trust the certificate.

Yikes! :D

radical:
Trying `--check` as suggested by @DamianEdwards now.

DamianEdwards (approved):

PR #4327, opened 2024-05-29
Pass IServiceProvider to configureBuilder of AddKafkaConsumer and AddKafkaProducer

g7ed6e:
Resolves #4323 

I had to deal with RS0027 and RS0028 and finally moved from optional parameters to explicit parameters overloads as parameter types differ.

sample usage:
```c#
var builder = Host.CreateApplicationBuilder(args);

builder.AddServiceDefaults();

builder.Services.AddSingleton<ISchemaRegistryClient>(provider =>
{
    var schemaRegistryConfig = new SchemaRegistryConfig
    {
        Url = "http://localhost:8081"
    };
    return new CachedSchemaRegistryClient(schemaRegistryConfig);
});
builder.AddKafkaProducer<Null, User>("kafka", configureBuilder: (provider, producerBuilder) =>
{
    var schemaRegistryClient = provider.GetRequiredService<ISchemaRegistryClient>();
    producerBuilder.SetValueSerializer(new AvroSerializer<User>(schemaRegistryClient));
});

builder.Services.AddHostedService<ContinuousProducerWorker>();

builder.Build().Run();
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4327)

PR #4642, opened 2024-06-25
Add Password To Redis

### LINKED ISSUE #3838: Revisit Redis password options
The Redis resource doesn't allow setting a password today, and this carries forward to deployment time too, i.e. Redis containers published from an AppHost have no password set. The resources for PostgreSQL, MSSQL, etc. do allow setting a password and **do so by default**.

We should revisit the default status of the Redis resource. As changing defaults has compatibility implications, we might want to consider doing it in stages, e.g.:

- in 8.1:
  - change the Redis resource to allow setting a password parameter (but don't set one by default in run mode)
  - by default, on publish a generated password expression is added in the manifest so that published Redis containers have a password set
- in 9.0:
  - change the Redis resource to behave more like the database resources, in that that a password parameter is used and a password is generated by default and used in local dev too

Alirexaa:
close: #3838
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4642)


src/Aspire.Hosting.Redis/PublicAPI.Shipped.txt
```diff
@@ -7,7 +7,6 @@ Aspire.Hosting.ApplicationModel.RedisResource.RedisResource(string! name) -> voi
 Aspire.Hosting.Redis.RedisCommanderResource
 Aspire.Hosting.Redis.RedisCommanderResource.RedisCommanderResource(string! name) -> void
 Aspire.Hosting.RedisBuilderExtensions
-static Aspire.Hosting.RedisBuilderExtensions.AddRedis(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name, int? port = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.RedisResource!>!
```
eerhardt, line 10: We shouldn't be editing this file. But instead you can add a `*REMOVED*` line to the Unshipped.txt file. See https://github.com/dotnet/aspire/pull/4327/files#diff-6a097377cfc55516be6e941106e3e5575efaa3ab602e773d13d76b803c68c989R25 for an example.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -24,13 +24,72 @@ public static class RedisBuilderExtensions
...
+            }
+            else if (builder.ExecutionContext.IsRunMode)
+            {
+                var redis = new RedisResource(name);
+                return builder.AddResource(redis)
+                              .WithEndpoint(port: port, targetPort: 6379, name: RedisResource.PrimaryEndpointName)
+                              .WithImage(RedisContainerImageTags.Image, RedisContainerImageTags.Tag)
+                              .WithImageRegistry(RedisContainerImageTags.Registry);
+            }
+            throw new InvalidOperationException();
```
eerhardt, line 81: Can we add a message to this exception to explain what went wrong?

mitchdenny: It is a shame that we have to do these parameters in this order (its different to `AddPostgres`).

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -24,13 +24,72 @@ public static class RedisBuilderExtensions
...
-                      .WithEndpoint(port: port, targetPort: 6379, name: RedisResource.PrimaryEndpointName)
-                      .WithImage(RedisContainerImageTags.Image, RedisContainerImageTags.Tag)
-                      .WithImageRegistry(RedisContainerImageTags.Registry);
+        return builder.AddRedis(name, port, null);
+    }
+
+    /// <summary>
+    /// Adds a Redis container to the application model.
+    /// </summary>
+    /// <remarks>
```
mitchdenny, line 35: Could you expand on the remarks here to explain that if a password parameter is not supplied that a password resource will be added to the model a publish time etc. Also as we are trying to improve our docs as we go add in a few `<example>` tags - one for with the password and one for without.

mitchdenny:
I'm comfortable with this change pending @eerhardt 's comments about the API txt files and some incremental improvements to the XML docs since we are in the neighbhorhood.

I'm a little disappointed that about the parameter ordering but I can't think of a way that isn't a binary/source breaking change that isn't also disruptive.

One more iteration and we should be able to stamp an approval on this. Thanks once again for your efforts @Alirexaa.

src/Aspire.Hosting.Redis/RedisResource.cs
```diff
@@ -18,9 +29,22 @@ public class RedisResource(string name) : ContainerResource(name), IResourceWith
...
 
+    /// <summary>
+    /// Gets the parameter that contains the Redis server password.
+    /// </summary>
+    public ParameterResource? PasswordParameter { get; }
+
+    /// <summary>
+    /// Gets the parameter that contains the Redis server username.
+    /// </summary>
+
```
radical, line 40:
```suggestion
```

src/Aspire.Hosting.Redis/RedisResource.cs
```diff
@@ -18,9 +29,22 @@ public class RedisResource(string name) : ContainerResource(name), IResourceWith
...
     private ReferenceExpression ConnectionString =>
+        PasswordParameter is null ?
+        ReferenceExpression.Create(
+            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}") :
         ReferenceExpression.Create(
-            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)},password={PasswordParameter}")
+        ;
```
radical, line 47: nit:
```csharp
        PasswordParameter is null
            ? ReferenceExpression.Create(
                $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}")
            : ReferenceExpression.Create(
                $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)},password={PasswordParameter}");
```

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -69,7 +69,24 @@ public void AddRedisContainerAddsAnnotationMetadata()
...
+        appBuilder.AddRedis("myRedis", password: pass).PublishAsContainer();
+
+        using var app = appBuilder.Build();
+
+        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
+
+        var connectionStringResource = Assert.Single(appModel.Resources.OfType<IResourceWithConnectionString>());
+        Assert.Equal("{myRedis.bindings.tcp.host}:{myRedis.bindings.tcp.port},password={pass.value}", connectionStringResource.ConnectionStringExpression.ValueExpression);
+    }
+    [Fact]
```
radical, line 88:
```suggestion

    [Fact]
```

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -69,7 +69,24 @@ public void AddRedisContainerAddsAnnotationMetadata()
...
+    public async Task RedisCreatesConnectionStringWithPassword()
```
radical, line 89: Should this be `*WithoutPassword`?
```suggestion
    public async Task RedisCreatesConnectionStringWithoutPassword()
```

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -69,7 +69,24 @@ public void AddRedisContainerAddsAnnotationMetadata()
     }
 
     [Fact]
-    public async Task RedisCreatesConnectionString()
+    public void RedisCreatesConnectionString()
```
radical, line 72:
```suggestion
    public void RedisCreatesConnectionStringWithPassword()
```

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -341,4 +394,29 @@ public void WithPersistenceAddsCommandLineArgsAnnotation()
         Assert.True(redis.Resource.TryGetAnnotationsOfType<CommandLineArgsCallbackAnnotation>(out var argsAnnotations));
         Assert.NotNull(argsAnnotations.SingleOrDefault());
     }
+    [Fact]
```
radical, line 397:
```suggestion

    [Fact]
```

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -19,18 +19,129 @@ public static class RedisBuilderExtensions
...
+    {
+        return builder.AddRedis(name, port, null);
+    }
+
+    /// <summary>
+    /// Adds a Redis container to the application model.
+    /// </summary>
+    /// <remarks>
+    /// The default image is "redis" and the tag is "7.2.4".
+    /// If a password parameter is not supplied, a password resource will be added to the model at publish time.
```
radical, line 54: Just curious - what password would be used in this case? And does that conflict with `password">The parameter used to provide the password for the Redis resource. If <see langword="null"/> resource will be not password protected`

radical: Maybe we also need a test to check that if the password parameter is not supplied then a password resource is being added to the model at publish time?

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -111,6 +128,40 @@ public async Task VerifyManifest()
         Assert.Equal(expectedManifest, manifest.ToString());
     }
 
+    [Fact]
+    public async Task VerifyWithPasswordManifest()
```
radical, line 132: Do we also need tests for when `builder.Configuration["Parameters:pass"] = password` is missing?

tests/Aspire.Hosting.Tests/Redis/AddRedisTests.cs
```diff
@@ -69,7 +69,24 @@ public void AddRedisContainerAddsAnnotationMetadata()
...
     [Fact]
-    public async Task RedisCreatesConnectionString()
+    public void RedisCreatesConnectionString()
+    {
+        var appBuilder = DistributedApplication.CreateBuilder();
+        var password = "p@ssw0rd1";
+        appBuilder.Configuration["Parameters:pass"] = password;
+
+        var pass = appBuilder.AddParameter("pass");
+        appBuilder.AddRedis("myRedis", password: pass).PublishAsContainer();
```
radical, line 79: Maybe another test for `AddRedis("myredis", port: .., password: ..)`?

radical:
Some of my comments might be non-sensical, or could be addressed in a follow-up PR. I'll let others decide that, since I'm very new to this code:)

playground/Redis/Redis.AppHost/Program.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+var builder = DistributedApplication.CreateBuilder(args);
+builder.Configuration["Parameters:pass"] = "123456";
+var password = builder.AddParameter("pass");
+
+var redis = builder.AddRedis("redis", password: password);
```
eerhardt, line 8: Can we `WithRedisCommander` here too?

src/Aspire.Hosting.Redis/RedisResource.cs
```diff
@@ -18,9 +29,22 @@ public class RedisResource(string name) : ContainerResource(name), IResourceWith
...
+    /// Gets the parameter that contains the Redis server username.
+    /// </summary>
+
     private ReferenceExpression ConnectionString =>
+        PasswordParameter is null ?
+        ReferenceExpression.Create(
+            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}") :
         ReferenceExpression.Create(
-            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)}");
+            $"{PrimaryEndpoint.Property(EndpointProperty.Host)}:{PrimaryEndpoint.Property(EndpointProperty.Port)},password={PasswordParameter}")
```
DamianEdwards, line 46: Does the password need to be encoded here, e.g. URI encoded?

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -19,18 +19,129 @@ public static class RedisBuilderExtensions
...
+        this IDistributedApplicationBuilder builder,
+        string name,
+        int? port = null,
+        IResourceBuilder<ParameterResource>? password = null)
+    {
+
+        if (password?.Resource is not null)
+        {
+
+            var redis = new RedisResource(name, password.Resource);
```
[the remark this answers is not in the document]
Alirexaa: How?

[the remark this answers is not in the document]
mgravell: We've discussed this for the client a few times but it has never hit the top of the pile. If it is a blocker, we can prioritize. I would suggest looking at ado-net quoted values as the template, but obvs we'd need to think a little about how to do it without impacting existing use of any new token

mgravell:
@eerhardt We've discussed this for the client a few times but it has never hit the top of the pile. If it is a blocker, we can prioritize. I would suggest looking at ado-net quoted values as the template, but obvs we'd need to think a little about how to do it without impacting existing use of any new token

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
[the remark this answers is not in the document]
eerhardt: Are you able to write a test?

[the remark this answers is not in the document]
Alirexaa: Do you mean Writing e2e test?

I think RedisCommander start failed when Redis have password but password was not provided in connectionString (here REDIS_HOSTS environments variable).

I could check RedisCommander start when Redis have password, Right?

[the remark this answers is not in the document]
eerhardt: Yes, I think writing a test that has:

* a Redis resource
* a RedisCommander resource

And running it, ensuring that the RedisCommander resource starts without any errors.

eerhardt:
> @eerhardt We've discussed this for the client a few times but it has never hit the top of the pile. If it is a blocker, we can prioritize. I would suggest looking at ado-net quoted values as the template, but obvs we'd need to think a little about how to do it without impacting existing use of any new token

I don't think it is blocking this work. I just wanted to point out that we can't URL encode the connection string here.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -19,18 +19,124 @@ public static class RedisBuilderExtensions
...
+    /// </item>
+    /// </list>
+    /// </example>
+    public static IResourceBuilder<RedisResource> AddRedis(
+        this IDistributedApplicationBuilder builder,
+        string name,
+        int? port = null,
+        IResourceBuilder<ParameterResource>? password = null)
+    {
+        if (password?.Resource is not null)
```
eerhardt, line 101: @DamianEdwards @mitchdenny @davidfowl - what do you think about not having this logic and just always adding a password if it is null here? Yes, it is technically a behavior breaking change. But it will make Redis consistent with the others that have passwords - mysql, postgres, etc.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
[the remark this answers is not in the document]
Alirexaa: @eerhardt, please take a look at the written test.
Also, it needs to import the RedisCommander image to `netaspireci.azurecr.io` to work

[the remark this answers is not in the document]
eerhardt: Thanks. I've imported the RedisCommander image to netaspireci.azurecr.io.

src/Aspire.Hosting.Redis/RedisCommanderConfigWriterHook.cs
```diff
@@ -31,7 +31,15 @@ public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, C
...
             {
-                var hostString = $"{(hostsVariableBuilder.Length > 0 ? "," : string.Empty)}{redisInstance.Name}:{redisInstance.PrimaryEndpoint.ContainerHost}:{redisInstance.PrimaryEndpoint.Port}:0";
+                var hostString = string.Empty;
+                if (redisInstance.PasswordParameter is not null)
+                {
+                    hostString = $"{(hostsVariableBuilder.Length > 0 ? "," : string.Empty)}{redisInstance.Name}:{redisInstance.PrimaryEndpoint.ContainerHost}:{redisInstance.PrimaryEndpoint.Port}:0:{redisInstance.PasswordParameter.Value}";
+                }
+                else
+                {
+                    hostString = $"{(hostsVariableBuilder.Length > 0 ? "," : string.Empty)}{redisInstance.Name}:{redisInstance.PrimaryEndpoint.ContainerHost}:{redisInstance.PrimaryEndpoint.Port}:0";
```
radical, line 41: Test needed for this case.

Alirexaa: Isn't these enough?

https://github.com/dotnet/aspire/blob/43e1cf4bb6493848c7fc60d15fe565677e69abda/tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs#L159-L180

https://github.com/dotnet/aspire/blob/43e1cf4bb6493848c7fc60d15fe565677e69abda/tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs#L182-L208

radical: Do these cover the `else` block of no password available?

Alirexaa: Could you please explain the exact senario you mean?
For me it seems cover.

radical: Maybe I'm misunderstanding something. This was prompted by looking at the code coverage report for that file. And looking at the code my understanding is that the test has `builder.AddRedis("myredis1")` and does not add a password explicitly, so we generate and add one by default. If that's the case then would this line of code be hit, as it is would run `if (redisInstance.PasswordParameter is not null)` is false.

Please correct me if I misunderstood the code.

Screenshot 2024-07-11 at 19 24 16

Alirexaa: in these two mentioned cases, we use the `Run` mode so the password is null and we do not add the password by default. so else branch be hit.
for just now (8.1), we generate a password by default when the execution context is `Publish` and the password was not added explicitly.

I think we just need a test case when a password was not provided and the execution context is `Publish`.

Alirexaa: > I think we just need a test case when a password was not provided and the execution context is Publish.

@radical, there is no need to test for this situation, because RedisCommander be excluded from the manifest.

https://github.com/dotnet/aspire/blob/cf61a58a739f059469bcbfbff7d5121526f6fa88/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs#L56-L61

Alirexaa:
What is the status of PR?

Currntly i'm working on #5165,
I need playground app there,
Besides that, if redis have password, i need to support password there.

@eerhardt @mitchdenny

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
mitchdenny: I agree let's shoot for consistency here.

mitchdenny:
@Alirexaa see: https://github.com/dotnet/aspire/pull/4642#discussion_r1669073634

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
Alirexaa: so, we don't need to check the execution context, always adding a password to the resource whether the app is in publish mode or run mode. right?

mitchdenny: If it is in publish mode and the password isn't provided we'll create a synthetic parameter like we do for Postgres.

Alirexaa: we already did. look at this

```
        {
            var passwordParameter = password?.Resource ?? ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder, $"{name}-password");
            var redis = new RedisResource(name, passwordParameter);

            return builder.AddResource(redis)
                      .WithEndpoint(port: port, targetPort: 6379, name: RedisResource.PrimaryEndpointName)
                      .WithImage(RedisContainerImageTags.Image, RedisContainerImageTags.Tag)
                      .WithImageRegistry(RedisContainerImageTags.Registry);
                      .WithImageRegistry(RedisContainerImageTags.Registry)
                      .WithPassword(passwordParameter);
        }
        else if (builder.ExecutionContext.IsRunMode)
        {
            var redis = new RedisResource(name);
            return builder.AddResource(redis)
                          .WithEndpoint(port: port, targetPort: 6379, name: RedisResource.PrimaryEndpointName)
                          .WithImage(RedisContainerImageTags.Image, RedisContainerImageTags.Tag)
                          .WithImageRegistry(RedisContainerImageTags.Registry);
        }
```

eerhardt: I've push an update for what I was talking about. We always add a password when calling `.AddRedis("name")` now. Just like other resources like PostgreSQL.

mitchdenny:
I've been talking a closer look at this from a security perspective, and I think we need to make some changes. When I run this locally the args are visble in the dashboard without any obfuscation (unlike environment variables).

I'm looking at the implications for deployment now too.

eerhardt:
> I've been talking a closer look at this from a security perspective, and I think we need to make some changes. When I run this locally the args are visble in the dashboard without any obfuscation (unlike environment variables).
> 
> I'm looking at the implications for deployment now too.
> 
> OK, so deployment doesn't use the args which means that the password isn't required. So we are going to need to fix that in `azd` before this change would have any impact (it has a local impact obviously but we need to think through to deployment for this one).

Good catch.

Reading https://redis.io/docs/latest/operate/oss_and_stack/management/security/#authentication, the other options for enabling a password are:

1. Adding a user and password to the [Access Control List](https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/)
2. Use the `requirepass` setting in the redis.conf file.

Either approach would require being able to write a file that is mounted into the container. Is that possible during deployment? @ellismg @vhvb1989

Alternatively, we could modify the Access Control List by connecting to the redis container after it is running, and issuing commands to it. I don't believe this is possible during deployment.

eerhardt:
>  When I run this locally the args are visble in the dashboard without any obfuscation (unlike environment variables).

@drewnoakes @adamint - is it possible to mark command line arguments as "secret"? So they get obfuscated in the dashboard.

vhvb1989:
> Either approach would require being able to write a file that is mounted into the container. Is that possible during deployment?

That's `bindMounts`, isn't it?
Like the sample from MySql [here](https://github.com/dotnet/aspire/blob/main/playground/mysql/MySqlDb.AppHost/Program.cs#L9)

The only difference would be that the file that is uploaded is generated, instead of committing it to the repo (or it can be a file in the repo with references to secured-args)

We support bind mount with an alpha feature in azd.  During provision, azd creates an storage account and one file-share which is set up as volume in the Azure Container Environment.  After provision, azd uploads the local content to the file-share. At the end, azd deploys the Azure Container App using the volume

drewnoakes:
> @drewnoakes @adamint - is it possible to mark command line arguments as "secret"? So they get obfuscated in the dashboard.

#5065 tracks marking resource properties as secrets, however that has to be specified by the resource service. We could hard-code some known properties as sensitive. If we did that, which would you think should be included? The current list is https://github.com/dotnet/aspire/blob/0f2931a94873d09e5a76f72485000e5388bf08dd/src/Shared/Model/KnownProperties.cs

mitchdenny:
As much as I would like this in for 8.2, I think we need to hold this PR until we get this resolved. I think hiding the args in the dashboard is fine for local given that if you have access to the dashboard you could uncover the secrets in environment variables anyway.

I'm more concerned about deployment where people will have a reasonable expectation that the if you specify the password that it'll be used when deployed to the cloud.

Frustratingly Redis Stack actually makes use of a `REDIS_ARGS` environment variable which can use to specify the `-requirepass [password]` argument.

eerhardt:
> Frustratingly Redis Stack actually makes use of a REDIS_ARGS environment variable which can use to specify the -requirepass [password] argument.

Yeah, I was wondering if it would make sense for us to switch from `library/redis` to `redis/redis-stack-server` instead. And then [Add WithRedisInsight (dotnet/aspire#5227)](https://github.com/dotnet/aspire/pull/5227) would switch the image from `redis/redis-stack-server` to `redis/redis-stack` and add the appropriate endpoints.

drewnoakes:
> > @drewnoakes @adamint - is it possible to mark command line arguments as "secret"? So they get obfuscated in the dashboard.
> 
> #5065 tracks marking resource properties as secrets, however that has to be specified by the resource service. We could hard-code some known properties as sensitive. If we did that, which would you think should be included? The current list is https://github.com/dotnet/aspire/blob/0f2931a94873d09e5a76f72485000e5388bf08dd/src/Shared/Model/KnownProperties.cs

This is in review in https://github.com/dotnet/aspire/pull/5380

mitchdenny:
> Yeah, I was wondering if it would make sense for us to switch from `library/redis` to `redis/redis-stack-server` instead. And then [Add WithRedisInsight (dotnet/aspire#5227)](https://github.com/dotnet/aspire/pull/5227) would switch the image from `redis/redis-stack-server` to `redis/redis-stack` and add the appropriate endpoints.

I think this is an option we should seriously consider. I think we would just need to look through where there are any licensing implications for folks deploying that image into production. There might be extra licensing requirements that we wouldn't want to silently swap people into.

eerhardt:
I've revived this PR with the latest code in `main`. The intent is to get this change into 9.1. The [initial concerns](https://github.com/dotnet/aspire/pull/4642#issuecomment-2292514163) that blocked this change should no longer be a problem. The command line args in the dashboard are hidden/masked by default now.

One complication is that Redis Insight needs to be initialized before we add a database with a password. I'll add a comment to the code to start a conversation about it there.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -305,10 +379,55 @@ await pipeline.ExecuteAsync(async (ctx) =>
...
+        {
+            return;
+        }
+
+        var jsonContent = JsonContent.Create(new
+        {
+            agreements = new
+            {
+                // all 4 are required to be set
+                eula = false,
```
eerhardt, line 402: This is unfortunate, but it is the only way I could get it to work. See https://github.com/RedisInsight/RedisInsight/issues/3452 for more information on why this is needed.

Note that this isn't accepting the EULA (it is sending false). But the consequences of initializing this is the pop up no longer happens on first run of Redis Insight, which is a nice user experience.

Thoughts? @sebastienros @DamianEdwards @Alirexaa

davidfowl: What are the options?

src/Aspire.Hosting.Redis/RedisResource.cs
```diff
@@ -18,9 +28,20 @@ public class RedisResource(string name) : ContainerResource(name), IResourceWith
...
 
+    /// <summary>
+    /// Gets the parameter that contains the Redis server password.
+    /// </summary>
+    public ParameterResource? PasswordParameter { get; }
+
+    /// <summary>
+    /// Gets the parameter that contains the Redis server username.
+    /// </summary>
     private ReferenceExpression ConnectionString =>
```
davidfowl, line 39: Prefer this pattern https://github.com/dotnet/aspire/blob/9e2a2dfbeda1b85501b3f80ad58f50f4609a1a09/src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs#L56

davidfowl (approved):
Do we mark this as a breaking change?

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
DamianEdwards: Seems [they're suggesting we wait](https://github.com/RedisInsight/RedisInsight/issues/3452#issuecomment-2623640956) until they implement support for adding databases via environment variables instead.

eerhardt:
> is it possible to specify the user for Redis? Or is that unnecessary in this scenario, with Aspire just using the default user?

Right now, no. We are just adding password protection for now. If necessary, we can also add support for user in the future.

> Similarly, there is increasing support for azure etc identity tokens for auth.

We support Azure Redis and Entra ID already in 9.0. See https://learn.microsoft.com/dotnet/aspire/caching/stackexchange-redis-integration#add-azure-cache-for-redis-client

eerhardt:
> Do we mark this as a breaking change?

I don't think this is a breaking change. We can log it in the changelog, but I'm not sure exactly what it would break. The only thing I can think of is if someone is explicitly cutting the password out of the connection string getting flown to the app (or passing the connection string a custom way - not using our Hosting ConnectionString).

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
eerhardt: I've opened [[Bug]: Posting eula: false to /api/settings suppresses the first run dialog (RedisInsight/RedisInsight#4340)](https://github.com/RedisInsight/RedisInsight/issues/4340) for the first run dialog issue.

Let's move forward with this PR for now. We can readjust later.

eerhardt:
Thank you, @Alirexaa for your continued contributions!

Issue #4651, opened 2024-06-25
Failing test `Aspire.Hosting.Tests.DistributedApplicationTests.TestServicesWithMultipleReplicas`


radical:
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=719871
Build error leg or test failing: Aspire.Hosting.Tests.DistributedApplicationTests.TestServicesWithMultipleReplicas
Pull request: https://github.com/dotnet/aspire.git/pull/4645

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

```
System.Net.Http.HttpRequestException : An error occurred while sending the request.
---- System.IO.IOException : Unable to read data from the transport connection: Connection reset by peer.
-------- System.Net.Sockets.SocketException : Connection reset by peer

   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Logging.LoggingHttpMessageHandler.<SendCoreAsync>g__Core|5_0(HttpRequestMessage request, Boolean useAsync, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Logging.LoggingScopeHttpMessageHandler.<SendCoreAsync>g__Core|5_0(HttpRequestMessage request, Boolean useAsync, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.GetStringAsyncCore(HttpRequestMessage request, CancellationToken cancellationToken)
   at Aspire.Hosting.Tests.DistributedApplicationTests.TestServicesWithMultipleReplicas() in /_/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs:line 191
[...]

Issue #4650, opened 2024-06-25
Failing test `Aspire.Hosting.Testing.Tests.TestingFactoryTests.HttpClientGetTest`


radical:
[Failing build](https://github.com/dotnet/aspire/pull/4647/checks?check_run_id=26663127382), and [another](https://github.com/dotnet/aspire/pull/4645/checks?check_run_id=26662883345).

```
System.Net.Http.HttpRequestException : Connection refused (localhost:5150)
---- System.Net.Sockets.SocketException : Connection refused

   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.CreateHttp11ConnectionAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.AddHttp11ConnectionAsync(QueueItem queueItem)
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellationAsync(CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Logging.LoggingHttpMessageHandler.<SendCoreAsync>g__Core|5_0(HttpRequestMessage request, Boolean useAsync, CancellationToken cancellationToken)
   at Microsoft.Extensions.Http.Resilience.ResilienceHandler.<>c.<<SendAsync>b__3_0>d.MoveNext()
--- End of stack trace from previous location ---
   at Polly.Outcome`1.ThrowIfException()
   at Microsoft.Extensions.Http.Resilience.ResilienceHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
[...]

PR #4653, opened 2024-06-25
Disable some hosting tests failing on CI

radical:
Issues: https://github.com/dotnet/aspire/issues/4650 and https://github.com/dotnet/aspire/issues/4651

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4653)

### DIFF
--- a/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
+++ b/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
@@ -171,6 +171,7 @@ public class DistributedApplicationTests
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/4651", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnCI))]
     public async Task TestServicesWithMultipleReplicas()
     {
         var replicaCount = 3;
--- a/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
+++ b/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
@@ -139,6 +139,7 @@ public class DistributedApplicationTests
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/4651", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnCI))]
     public async Task AllocatedPortsAssignedAfterHookRuns()
     {
         using var testProgram = CreateTestProgram();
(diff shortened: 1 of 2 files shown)


tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
```diff
@@ -171,6 +171,7 @@ public Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, C
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/4651", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnCI))]
```
eerhardt, line 174: Why isn't this

```suggestion
    [ActiveIssue("https://github.com/dotnet/aspire/issues/4651")]
```

?

If a test is flaky, I don't want it to fail on my machine locally only to realize it might be broken and not due to my changes.

radical: This was working fine earlier on local runs, and CI. But started to fail around the time we enabled docker based tests to run on the build machine. But I think this affected by other tests that run before this.

Instead of disabling this one, I will revert all the tests in `tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs` which had `[LocalOnly*]` back to being local-only. This is what @mitchdenny suggested in https://github.com/dotnet/aspire/pull/4142#issuecomment-2188780945 too.

radical: I retained the attributed as-is on this test because it was also marked `LocalOnly*` earlier.

adamint (approved):

tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
radical: @eerhardt I haven't heard from anyone about this test failing locally. But I can disable it for local too, if that's preferable.

radical: Merging this to get a green CI. But I can make this change in a follow up PR if you want.

eerhardt: > I haven't heard from anyone about this test failing locally.

My guess is that it isn't getting run as often locally as it is in CI.

My thinking is that CI machines are just normal computers. If we have races / flaky failures in our tests on CI, chances are we do locally too.

PR #4655, opened 2024-06-25
Bumping dependency versions and updating Configuration Schema for components

joperezr:
In preparation for our 8.0.2 release, this change is updating all[^footnote] of our package dependencies to latest versions out in Nuget.org

[^footnote]: With the exception of Azure.Open.AI package as that one is still in beta and new version has breaking changes that require more risk than what we want in this servicing.

cc: @eerhardt can you please review and make sure dependencies look correct?

FYI: @radical @danmoseley @DamianEdwards @davidfowl @maddymontaquila  This is in preparation of our 8.0.2 release later this week.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4655)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,16 +9,16 @@
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.308.15" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.301.23" />
-    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.301.58" />
-    <PackageVersion Include="AWSSDK.Core" Version="3.7.304.20" />
+    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.308.9" />
+    <PackageVersion Include="AWSSDK.SQS" Version="3.7.301.17" />
+    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.301.52" />
+    <PackageVersion Include="AWSSDK.Core" Version="3.7.304.14" />
     <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.301" />
     <PackageVersion Include="AWS.Messaging" Version="0.9.1" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.4" />
     <PackageVersion Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
     <!-- Azure SDK for .NET dependencies -->
-    <PackageVersion Include="Azure.AI.OpenAI" Version="1.0.0-beta.16" />
+    <PackageVersion Include="Azure.AI.OpenAI" Version="1.0.0-beta.17" />
     <PackageVersion Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
     <PackageVersion Include="Azure.Identity" Version="1.12.0" />
@@ -90,7 +90,7 @@
     <PackageVersion Include="Microsoft.Extensions.TimeProvider.Testing" Version="$(MicrosoftExtensionsTimeProviderTestingVersion)" />
     <!-- external dependencies -->
     <PackageVersion Include="Confluent.Kafka" Version="2.4.0" />
-    <PackageVersion Include="Dapper" Version="2.1.35" />
+    <PackageVersion Include="Dapper" Version="2.1.37" />
     <PackageVersion Include="DnsClient" Version="1.7.0" />
     <PackageVersion Include="Google.Protobuf" Version="3.27.1" />
     <PackageVersion Include="Grpc.AspNetCore" Version="2.63.0" />
@@ -99,9 +99,9 @@
     <PackageVersion Include="Humanizer.Core" Version="2.14.1" />
     <PackageVersion Include="KubernetesClient" Version="14.0.2" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.2.1" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.8.1" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.8.0" />
-    <PackageVersion Include="MongoDB.Driver" Version="2.26.0" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.7.2" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.7.2" />
+    <PackageVersion Include="MongoDB.Driver" Version="2.25.0" />
     <PackageVersion Include="MongoDB.Driver.Core.Extensions.DiagnosticSources" Version="1.4.0" />
     <PackageVersion Include="MySqlConnector.DependencyInjection" Version="2.3.6" />
     <PackageVersion Include="MySqlConnector.Logging.Microsoft.Extensions.Logging" Version="2.1.0" />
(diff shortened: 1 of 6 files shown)


Directory.Packages.props
```diff
@@ -9,55 +9,55 @@
...
+    <PackageVersion Include="AWS.Messaging" Version="0.9.1" />
+    <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.4" />
     <PackageVersion Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
     <!-- Azure SDK for .NET dependencies -->
     <PackageVersion Include="Azure.AI.OpenAI" Version="1.0.0-beta.16" />
     <PackageVersion Include="Azure.Data.Tables" Version="12.8.3" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
-    <PackageVersion Include="Azure.Identity" Version="1.11.2" />
-    <PackageVersion Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
+    <PackageVersion Include="Azure.Identity" Version="1.12.0" />
```
DamianEdwards, line 24: How comfortable are we with bumping minor versions of dependencies in a servicing release? This will result in new API showing up for folks in their apps after doing a servicing update. I notice there's a few dependencies we're updating to new minor versions in this set (not including `0.x` versioned packages) and we should be clear about what our plans are on this moving forward.

eerhardt: I'm comfortable with it. This has been our plan. See https://github.com/dotnet/aspire/tree/main/src/Components#dependency-versioning

> Aspire component packages will be serviced regularly, capturing the latest available versions of the client libraries and Microsoft extensions they depend on, making it possible for applications built on Aspire to always benefit from the latest features and fixes.

Directory.Packages.props
```diff
@@ -86,78 +86,78 @@
...
     <PackageVersion Include="Testcontainers.MongoDb" Version="$(TestcontainersPackageVersion)" />
-    <PackageVersion Include="Testcontainers.MsSql" Version="$(TestcontainersPackageVersion)" />
-    <PackageVersion Include="Testcontainers.MySql" Version="$(TestcontainersPackageVersion)" />
-    <PackageVersion Include="Testcontainers.PostgreSQL" Version="$(TestcontainersPackageVersion)" />
-    <PackageVersion Include="Testcontainers.RabbitMq" Version="$(TestcontainersPackageVersion)" />
+    <PackageVersion Include="Testcontainers.MsSql" Version="3.9.0" />
+    <PackageVersion Include="Testcontainers.MySql" Version="3.9.0" />
+    <PackageVersion Include="Testcontainers.PostgreSQL" Version="3.9.0" />
+    <PackageVersion Include="Testcontainers.RabbitMq" Version="3.9.0" />
     <PackageVersion Include="Testcontainers.Redis" Version="$(TestcontainersPackageVersion)" />
```
eerhardt, line 149: 1. Why are removing the common MSBuild property in the other lines?
2. Why isn't this one updated like the rest of them?

Note that TestContainers are only used by our tests, so updating it here isn't buying any customer value.

Directory.Packages.props
```diff
@@ -86,78 +86,78 @@
...
+    <PackageVersion Include="Npgsql.DependencyInjection" Version="8.0.3" />
+    <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.4" />
+    <PackageVersion Include="Oracle.EntityFrameworkCore" Version="8.23.40" />
+    <PackageVersion Include="Polly.Core" Version="8.4.0" />
+    <PackageVersion Include="Polly.Extensions" Version="8.4.0" />
     <PackageVersion Include="Pomelo.EntityFrameworkCore.MySql" Version="8.0.2" />
     <PackageVersion Include="Qdrant.Client" Version="1.9.0" />
     <PackageVersion Include="RabbitMQ.Client" Version="[6.8.1,7.0.0)" />
-    <PackageVersion Include="StackExchange.Redis" Version="2.7.33" />
+    <PackageVersion Include="StackExchange.Redis" Version="2.8.0" />
```
eerhardt, line 117: It's kind of surprising to jump out in front of `main` with our dependencies. We haven't taken this update there. Do we plan on making the same change in `main`?

Directory.Packages.props
DamianEdwards: OK. I admit I'm a bit uneasy about it but we can adjust what we do moving forward if we hit any issues.

Directory.Packages.props
DamianEdwards: Yeah it seems unwise to bump deps for the first time in a servicing branch and then ship very quickly from there, rather than updating them in main and allowing some bake time. That said, we should be regularly bumping deps in main (we discussed weekly IIRC) so this becomes less risky.

joperezr: Yes, this commit will get merged into main after releasing as part of the release/8.0 -> main merge that happens after each release

Directory.Packages.props
joperezr: I noticed msbuild properties were removed ( I used a tool for the updates) I will fix that.

Directory.Packages.props
eerhardt: > that happens after each release

That doesn't seem like the right process/steps/flow. I think these should happen in main first to get more coverage.

Directory.Packages.props
joperezr: >Note that TestContainers are only used by our tests, so updating it here isn't buying any customer value.

Good call, reverting that.

Directory.Packages.props
joperezr: Sounds good to me. I can revert this particular one to match what we have in main.

eerhardt: To be honest, instead of reverting, I think we should do it in both now. Getting the updated dependency in our 8.0.2 release sounds good to me. I just want us to also make it in main at the same time (or sooner).

joperezr: Given 8.0.2 is set to ship this week, and we don't have that much time, I think I like your original suggestion of keeping with the versions we have in main now to reduce risk as people have been validating main now for a while. I'll update this PR to use versions that match main, and then I'll put up a PR in main for the newer versions so we can ship those in 8.1. Sound good?

DamianEdwards: That sounds good, with the addition of ensuring we include any updates that lift transitive dependencies with vulnerabilities (e.g. `Azure.Identity`).

radical:
```
❌src\Components\Directory.Build.targets(117,5): error : ConfigurationSchema.json is out of date. Run 'dotnet build --no-incremental /p:UpdateConfigurationSchema=true' to update it.
❌playground\orleans\OrleansServer\Program.cs(17,32): error CS0246: The type or namespace name 'IPersistentState<>' could not be found (are you missing a using directive or an assembly reference?)
❌playground\orleans\OrleansServer\Program.cs(17,6): error CS0246: The type or namespace name 'PersistentStateAttribute' could not be found (are you missing a using directive or an assembly reference?)
```

Directory.Packages.props
danmoseley: Any issues are presumably more likely from major version updates like Newtonsoft Json Schema here.

src/Components/Aspire.Azure.Storage.Blobs/ConfigurationSchema.json
```diff
@@ -77,6 +77,29 @@
                           "format": "uri",
                           "description": "Gets or sets the secondary storage 'System.Uri' that can be read from for the storage account if the\naccount is enabled for RA-GRS.\n\nIf this property is set, the secondary Uri will be used for GET or HEAD requests during retries.\nIf the status of the response from the secondary Uri is a 404, then subsequent retries for\nthe request will not use the secondary Uri again, as this indicates that the resource\nmay not have propagated there yet. Otherwise, subsequent retries will alternate back and forth\nbetween primary and secondary Uri."
                         },
+                        "Request100ContinueOptions": {
```
danmoseley, line 80: Where does this stuff come from?

danmoseley (approved):
Worth describing the process you used here? Eg link to the script.

src/Components/Aspire.Azure.Storage.Blobs/ConfigurationSchema.json
joperezr: This is the tooling that @eerhardt wrote to make sure the configuration schema for the configuration options in components stay up-to-date. Essentially each time that one of our transitive dependencies adds properties to their options type (which we then use in Aspire), our build infra will detect it and fail the build saying that the configuration schema needs an update and it will also show the command you need to run in order to update it.

joperezr:
> Worth describing the process you used here? Eg link to the script.

Sure, I did the following:

```shell
# from the root of the repo
./build.cmd -restore # to ensure that we start off from a good point and dotnet-outdated can traverse dependencies.
dotnet new nugetconfig --force # to override the the NuGet.config with one that only has NuGet.org as the source
dotnet outdated --no-restore -u .\Aspire.sln # to perform the update of packages
git checkout -- NuGet.config # to undo the change to nuget.config file
```

If you don't have dotnet-outdated tool install yet, you can get it by following the steps on their repo: https://github.com/dotnet-outdated/dotnet-outdated

src/Components/Aspire.Azure.Storage.Blobs/ConfigurationSchema.json
joperezr: All of this file is auto-generated. The description and info are grabbed  from the tripple slash comments extracted by our build tooling from those transitive dependencies.

Directory.Packages.props
eerhardt: Newtonsoft Json Schema is only used by our Hosting tests. so there is no concern. (On a side note, we should move off of it and use `JsonSchema.Net` everywhere instead. We are already using `JsonSchema.Net` in the components tests.)

eerhardt (approved):
LGTM. Thanks!

danmoseley:
> Sure, I did the following:

Ah, that's exactly what I was doing. sounds good

danmoseley:
do we want to pull any testing improvements from main into release/8.0 too? if they are separable and worthwhile. we have the playwright stuff, testcontainers stuff.

Directory.Packages.props
danmoseley: maybe this file should segregate references like that that don't bleed to customers, so we know our own build is complete validation.

eerhardt: > maybe this file should segregate references like that

It tries to be. If you look at the different sections:

https://github.com/dotnet/aspire/blob/21e35286025460824be1f1d149a17d60181b2b10/Directory.Packages.props#L132-L133

https://github.com/dotnet/aspire/blob/21e35286025460824be1f1d149a17d60181b2b10/Directory.Packages.props#L140-L141

https://github.com/dotnet/aspire/blob/21e35286025460824be1f1d149a17d60181b2b10/Directory.Packages.props#L158-L159

It's just that the Newtonsoft.Json.Schema dependency didn't follow this segregration and was added by Newtonsoft.Json instead:

https://github.com/dotnet/aspire/blob/21e35286025460824be1f1d149a17d60181b2b10/Directory.Packages.props#L165-L167

joperezr:
I think that most of that actually depends on a lot of infrastructure that is also only on main, and given the fact that we don't really plan on shipping from the release/8.0 branch, I'm not sure if it's worth the effort. For this particular servicing, I've already reached out to @balachir to have his team do a round of validation with 8.0.2 build to ensure things look great.

I do agree that moving forward, we'll want to more heavily rely on automation for coverage and not just our CTI team.

Issue #4658, opened 2024-06-25
Need a way to get to an IServiceProvider from EnvironmentCallbackContext


karolz-ms:
Scenario: I am writing an environment callback handler and want to configure my resources slightly differently depending on what version of DCP I am using. To do this, I would like to get an instance of IDcpDependencyCheckService from a service provider. 

According to @mitchdenny there isn't a good way to do this today.

mitchdenny:
@karolz-ms is this blocking for 8.1? Just thinking about priorities. I think it should be easy enough to do.

karolz-ms:
@mitchdenny it would help, otherwise I would have to do a workaround that then would need to be taken out in subsequent releases.

mitchdenny:
Addressed in #4668

PR #4661, opened 2024-06-26
Use uniform convention for naming and annotating DCP objects with regards to OTEL

### LINKED ISSUE #4302: Trace span names not showing useful titles

karolz-ms:
Fixes https://github.com/dotnet/aspire/issues/4302

With this change all "leaf" DCP objects created by Aspire during local run (Executables and Containers) will follow uniform naming convention: `serviceName-serviceInstanceId`. All leaf objects also get a pair of annotations for the service name and service instance ID, respectively. This allows to reliably convert, in both directions, between DCP object name and OTEL (service name, service instance ID) pair.

A couple of problems remain, hence this is a draft PR for now
- [x] As-is, this REQUIRES latest DCP (0.5.5), which is probably not what we want to enforce for 8.1 release. A "fallback" to old behavior is currently somewhat hard to implement due to https://github.com/dotnet/aspire/issues/4658
- [x] Traces from Executable replicas are not working correctly. They are collected and listed in the context of the whole service (replica set), but not attributed to any individual replica. There is probably some dashboard service/dashboard change missing here. @JamesNK @drewnoakes  any suggestions? Update: @JamesNK Appears to work for me.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4661)

UPDATE 7/1/2024: ready for review, all problems from draft phase fixed


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -72,6 +72,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                                           IDcpDependencyCheckService dcpDependencyCheckService)
 {
     private const string DebugSessionPortVar = "DEBUG_SESSION_PORT";
+    private const int RandomNameSuffixLength = 8;
```
JamesNK, line 75: Add a comment about why this length is used.

src/Aspire.Hosting/Dcp/DcpVersion.cs
```diff
@@ -5,8 +5,8 @@ namespace Aspire.Hosting.Dcp;
 
 internal static class DcpVersion
 {
-    public static Version MinimumVersionInclusive = new Version(0, 1, 55);
-    public static Version MinimumVersionIdeProtocolV1 = new Version(0, 1, 61);
+    public static Version MinimumVersionInclusive = new Version(0, 2, 3); // Aspire GA (8.0) release
+    public static Version MinimumVersionAspire_8_1 = new Version(0,5,6);
```
JamesNK, line 9:
```suggestion
    public static Version MinimumVersionAspire_8_1 = new Version(0, 5, 6);
```

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -56,8 +58,22 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
             }
 
             // Set the service name and instance id to the resource name and UID. Values are injected by DCP.
-            context.EnvironmentVariables["OTEL_RESOURCE_ATTRIBUTES"] = "service.instance.id={{- .Name -}}";
-            context.EnvironmentVariables["OTEL_SERVICE_NAME"] = "{{- index .Annotations \"otel-service-name\" -}}";
+            var dcpDependencyCheckService = context.ExecutionContext?.ServiceProvider?.GetService(typeof(IDcpDependencyCheckService)) as IDcpDependencyCheckService;
```
JamesNK, line 61: Can ExecutionContext and ServiceProvider be null here? Are null forgiving operators needed?

Also, use `GetRequiredService<IDcpDependencyCheckService>()`

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -56,8 +58,22 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
...
+            if (dcpDependencyCheckService is null)
+            {
+                throw new InvalidOperationException("DCP dependency check service is not available"); // Should never happen.
+            }
+            var dcpInfo = await dcpDependencyCheckService.GetDcpInfoAsync(context.CancellationToken).ConfigureAwait(false);
+            if (dcpInfo?.Version?.CompareTo(DcpVersion.MinimumVersionAspire_8_1) >= 0)
```
JamesNK, line 67: I don't think Version on dcp info can be null (at least, I didn't see the extra null operator in other places)
```suggestion
            if (dcpInfo?.Version.CompareTo(DcpVersion.MinimumVersionAspire_8_1) >= 0)
```

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: added

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
karolz-ms: It is defined as nullable

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
karolz-ms: Thanks, that is much better. Fixed.

JamesNK (approved):

PR #4668, opened 2024-06-26
Expose IServiceProvider on DistributedApplicationExecutionContext.

mitchdenny:
Experimenting with exposing `IServiceProvider` via `DistributedApplicationExecutionContext` with minimal API breaking changes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4668)

### DIFF
--- a/src/Aspire.Hosting/DistributedApplicationExecutionContext.cs
+++ b/src/Aspire.Hosting/DistributedApplicationExecutionContext.cs
@@ -14,6 +14,17 @@ public class DistributedApplicationExecutionContext(DistributedApplicationOperat
     /// </summary>
     public DistributedApplicationOperation Operation { get; } = operation;
 
+    private IServiceProvider? _serviceProvider;
+
+    /// <summary>
+    /// The <see cref="IServiceProvider"/> for the AppHost.
+    /// </summary>
+    public IServiceProvider ServiceProvider
+    {
+        get =>_serviceProvider ?? throw new InvalidOperationException("The ServiceProvider has not been set.");
+        internal set => _serviceProvider = value;
+    }
+
     /// <summary>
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting/DistributedApplicationExecutionContext.cs
```diff
@@ -14,6 +14,17 @@ public class DistributedApplicationExecutionContext(DistributedApplicationOperat
...
 
+    private IServiceProvider? _serviceProvider;
+
+    /// <summary>
+    /// The <see cref="IServiceProvider"/> for the AppHost.
+    /// </summary>
+    public IServiceProvider ServiceProvider
+    {
+        get =>_serviceProvider ?? throw new InvalidOperationException("The ServiceProvider has not been set.");
+        internal set => _serviceProvider = value;
```
mitchdenny, line 25: This probably won't be able to stay internal because it makes it impossible to unit test easily.

davidfowl: It can't be internal, then you can't write tests for it.

karolz-ms (approved):
I'll take this

JamesNK:
I think you need another constructor. It could take both arguments: DistributedApplicationOperation and IServiceProvider. That doesn't scale.

Or you could create an options type, e.g. `DistributedApplicationExecutionContextOptions` that has properties to set on it. A new constructor that takes the options copies the values to the new `DistributedApplicationExecutionContext`.

mitchdenny:
@JamesNK pretty much went with your proposal here.

karolz-ms (approved):

PR #4679, opened 2024-06-26
New WithEndpointsInEnvironment method

### LINKED ISSUE #4472: Allow users to define what endpoints are selected as ASPNETCORE_URLS, HTTP/HTTPS_PORTS and Kestrel endpoints
When defining endpoints on project resources, we have some default heuristics for expressing which of these endpoints are included in ASPNETCORE_URLS and which ones are included at publish time for HTTP/HTTPS_PORTS or if kestrel endpoints are defined, a way to use those.

We'd like to add some methods to allow overriding how these get included as an override in case you need to tweak the behavior. The ultimate escape hatch is setting the environment variable yourself as we don't override what users set, but maybe a more explicit way to configure set it would make sense.

This will be turned into an API proposal once we figure out the pattern.

davidebbo:
Fixes #4472
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4679)


src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -557,6 +556,26 @@ public static IResourceBuilder<ProjectResource> DisableForwardedHeaders(this IRe
...
+    public static IResourceBuilder<ProjectResource> WithEndpointsInEnvironment(
+        this IResourceBuilder<ProjectResource> builder, Func<EndpointAnnotation, bool> filter)
+    {
+        foreach (var e in builder.Resource.Annotations.OfType<EndpointAnnotation>())
+        {
+            if (!filter(e))
+            {
+                builder.Resource.SkipEndpointEnvironment(e);
+            }
+        }
```
davidfowl, line 575: I think we want to store the filter and have it run when the environment is being evaluated.

davidebbo: Interesting. Is the reasoning that the state of the `EndpointAnnotation` could change between the call to `WithEndpointsInEnvironment` and the env evaluation, such that the user filter would execute differently before and after?

davidebbo: Well, regardless of whether it makes a semantic difference, your suggesting makes the implementation cleaner, so I changed it accordingly.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -578,8 +591,15 @@ private static IConfiguration GetConfiguration(ProjectResource projectResource)
...
 
-    static bool IsValidAspNetCoreUrl(EndpointAnnotation e) =>
-        e.UriScheme is "http" or "https" && e.TargetPortEnvironmentVariable is null;
+    private static bool ShouldInjectEndpointEnvironment(this IResourceBuilder<ProjectResource> builder, EndpointReference e)
+    {
+        // Check a set of conditions to determine if we should inject environment variables for this endpoint.
+        var endpoint = e.EndpointAnnotation;
+        return
+            endpoint.UriScheme is "http" or "https" && // Only process http and https endpoints
+            endpoint.TargetPortEnvironmentVariable is null && // Skip if target port env variable was set
```
davidfowl, line 600: Do we keep these or are these in the default filter?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -557,6 +556,20 @@ public static IResourceBuilder<ProjectResource> DisableForwardedHeaders(this IRe
...
+    /// Set a filter that determines if environment variables are injected for a given endpoint.
+    /// By default, all endpoints are included (if this method is not called).
+    /// </summary>
+    /// <param name="builder">The project resource builder.</param>
+    /// <param name="filter">The filter callback that returns true if and only if the endpoint should be included.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ProjectResource> WithEndpointsInEnvironment(
+        this IResourceBuilder<ProjectResource> builder, Func<EndpointAnnotation, bool> filter)
+    {
+        builder.Resource.InjectEnvironmentForEndpoint = filter;
```
davidfowl, line 569: Should we care about composition? Being able to get the previous filter?

davidebbo: Pretty advanced scenario that no one will probably use, but I suppose we could. That could mean two things:
1. Automatically call the parent chain, transparently to the user. All of them need to return true for a given endpoint to be injected.
2. Pass the previous filter to the user and have them decide whether to delegate to it.

The second is a little more flexible as it lets you completely override previous filters. However, it makes the object model uglier for the user, e.g.

c#
```
.WithEndpointsInEnvironment(e,prev => e.Name != "FirstHttpEndpoint" && prev(e))
```

So if we do this, I'd vote for the first.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Yes, they could go in the default filter, but that then makes an even stronger case for doing the transparent delegation, per other comment. Otherwise, if a user chooses not to delegate, they'll lose those default filter, which is very likely not what they want.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: Lets do it later.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidfowl: OK I can buy that.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: I ended up implementing the first approach (no change to user model). PTAL.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
davidebbo: Done as part of the composition change.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -557,6 +556,20 @@ public static IResourceBuilder<ProjectResource> DisableForwardedHeaders(this IRe
...
+    /// </summary>
+    /// <param name="builder">The project resource builder.</param>
+    /// <param name="filter">The filter callback that returns true if and only if the endpoint should be included.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<ProjectResource> WithEndpointsInEnvironment(
+        this IResourceBuilder<ProjectResource> builder, Func<EndpointAnnotation, bool> filter)
+    {
+        builder.Resource.AddEnvironmentInjectionFilter(filter);
+        return builder;
+    }
```
davidfowl, line 571: If we had add and no remove, do we need a way to reset?

davidebbo: That feels like a really edge scenario.

The scenario for passing multiple filters is to have two independent pieces of code say "I want to exclude this and that endpoint from env vars". It sort of makes sense, because different pieces of code may know about a different subset of endpoints, so they're not really stepping on each others' feet.

A reset would be saying: "ignore all the exclusions that other callers told you about, and include everything in the env vars". Is this really worth having a separate public method for this? If we did have it, I'd call it `WithAllEndpointsInEnvironment()`, and not take a filter at all. But personally I'd leave it out.

mitchdenny: If people want to make this something that is a bit more dynamic they can always add their own filter which then has its own rules mechanism which they can mutate. But I am with @davidebbo I don't think we need it.

mitchdenny: ... having said that, we could just make this an annotation and loop over it (a well trodden patttern in Aspire).

mitchdenny:
OK, so I'll reiterate my comment above here that I think that the `EnvironmentInjectionFilterEntry` type could just be an annotation (which means it could possibly work across resource types more easily as well).

Also we need a live test that covers the endpoint filtering so that the code path in `ApplicationExecutor` is validated.

davidebbo:
> OK, so I'll reiterate my comment above here that I think that the `EnvironmentInjectionFilterEntry` type could just be an annotation (which means it could possibly work across resource types more easily as well).

To make sure I understand, what you're suggestion is an implementation change, but not a change to the public API, right? i.e. we would add an annotation for each filter, instead of using alternate data structures. And this annotation would be internal, right?

mitchdenny:
@davidebbo that's right. The annotation will probably need to be public also.

davidebbo:
Got it, I'll experiment with that tonight.

davidebbo:
@mitchdenny I changed it to used annotations to see what that feels like.

Misc comments/questions:
- It fits well into the annotation model, and does make the code a bit simpler.
- The annotation index has no bearing on the behavior. i.e. I did not make it such that a given filter only affects annotations that come before it, as that felt quirky.
- It's a bit less efficient as we need to cycle through all annotations each time we want to test a filter.
- I kept the annotation internal for now, as it has kind of an ugly name, and I'm not clear what scenario we're enabling by making it public. If public, users might try to add it to other resource types, and I think for now, @davidfowl wanted to limit this to Projects. This could easily be revisited later.
- By 'live test', do you mean a full functional test as opposed to unit test? Can you point me to a good example test that I can either add to, or clone as a separate test for this? Thanks!

src/Aspire.Hosting/ApplicationModel/EnvironmentInjectionFilterAnnotation.cs
```diff
@@ -0,0 +1,12 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Annotation that holds a filter that determines if environment variables should be injected for a given endpoint.
+/// </summary>
+internal class EnvironmentInjectionFilterAnnotation(Func<EndpointAnnotation, bool> filter) : IResourceAnnotation
```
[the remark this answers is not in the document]
mitchdenny: Pushed an update for this. Hope you don't mind.

mitchdenny (approved):

mitchdenny:
@davidebbo this is looking good. Thanks for this submission. Unless there is anything else you think we should wait for I'd like to get this merged in.

davidebbo:
@mitchdenny I think it's good to go. Thanks for the review and suggestions!

PR #4689, opened 2024-06-26
[tests] Extract playwright bits to Playwright.targets, and separate class

radical:
... to allow re-use outside the workload tests.

- Extracts the `Playwright` specific msbuild bits to `tests/Shared/Playwright/`
- A test project can import `tests/Shared/Playwright/Playwright.targets` to get support for installing the dependencies on build.

Prompted by https://github.com/dotnet/aspire/pull/4386 .

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4689)


tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
```diff
@@ -11,14 +11,13 @@
...
     <TestArchiveTestsDir>$(TestArchiveTestsDirForWorkloadTests)</TestArchiveTestsDir>
 
-    <InstallChromeForPlaywright Condition="'$(InstallChromeForPlaywright)' == '' and '$(CODESPACES)' == 'true'">true</InstallChromeForPlaywright>
-    <InstallChromeForPlaywright Condition="'$(InstallChromeForPlaywright)' == '' and '$(ArchiveTests)' == 'true' and '$(ContinuousIntegrationBuild)' == 'true'">true</InstallChromeForPlaywright>
-    <InstallChromeForPlaywright Condition="'$(InstallChromeForPlaywright)' == '' and '$(OS)' == 'Windows_NT' and '$(ContinuousIntegrationBuild)' != 'true'">true</InstallChromeForPlaywright>
+    <BrowsersForPlaywrightToInstall Condition="'$(BrowsersForPlaywrightToInstall)' == '' and '$(CODESPACES)' == 'true'">true</BrowsersForPlaywrightToInstall>
+    <BrowsersForPlaywrightToInstall Condition="'$(BrowsersForPlaywrightToInstall)' == '' and '$(ArchiveTests)' == 'true' and '$(ContinuousIntegrationBuild)' == 'true'">true</BrowsersForPlaywrightToInstall>
+    <BrowsersForPlaywrightToInstall Condition="'$(BrowsersForPlaywrightToInstall)' == '' and '$(OS)' == 'Windows_NT' and '$(ContinuousIntegrationBuild)' != 'true'">true</BrowsersForPlaywrightToInstall>
```
radical, line 16:
```suggestion
    <InstallBrowsersForPlaywright Condition="'$(InstallBrowsersForPlaywright)' == '' and '$(CODESPACES)' == 'true'">true</InstallBrowsersForPlaywright>
    <InstallBrowsersForPlaywright Condition="'$(InstallBrowsersForPlaywright)' == '' and '$(ArchiveTests)' == 'true' and '$(ContinuousIntegrationBuild)' == 'true'">true</InstallBrowsersForPlaywright>
    <InstallBrowsersForPlaywright Condition="'$(InstallBrowsersForPlaywright)' == '' and '$(OS)' == 'Windows_NT' and '$(ContinuousIntegrationBuild)' != 'true'">true</InstallBrowsersForPlaywright>
```

tests/Shared/Playwright/Playwright.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)*.cs" Link="Playwright" />
+
+    <PackageReference Include="Microsoft.Playwright" />
+  </ItemGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)provisioning.targets" />
```
eerhardt, line 8: Is there a reason this is a separate .targets file? Why not just put this in Playwright.targets itself?

tests/Shared/Playwright/PlaywrightProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+        string? browserPath = Environment.GetEnvironmentVariable(BrowserPathEnvironmentVariableName);
+        if (!string.IsNullOrEmpty(browserPath) && !File.Exists(browserPath))
+        {
+            throw new FileNotFoundException($"Browser path {BrowserPathEnvironmentVariableName}='{browserPath}' does not exist");
+        }
+
+        options ??= new() { Headless = true };
+        options.ExecutablePath ??= browserPath;
+
+        if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) && string.IsNullOrEmpty(browserPath))
```
eerhardt, line 26:
```suggestion
        if (OperatingSystem.IsMacOS() && string.IsNullOrEmpty(browserPath))
```

tests/Shared/Playwright/PlaywrightProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+    public static void DetectAndSetInstalledPlaywrightDependenciesPath(DirectoryInfo? repoRoot = null)
+    {
+        if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(PlaywrightBrowsersPathEnvironmentVariableName)))
+        {
+            // this would be the case for helix where the path is set to a
+            // payload directory
+            return;
+        }
+
+        if (repoRoot is null)
```
eerhardt, line 47: Does this case ever happen? If repoRoot is null its because we are in Helix, right?

eerhardt (approved):

tests/Shared/Playwright/PlaywrightProvider.cs
radical: This one is if the caller didn't pass in a pre-computed `repoRoot` then we do the detection here. In `BuildEnvironment.cs` we already have `repoRoot` computed so this would avoid it being repeated. But when using it outside the workload tests, the caller might not have it, and thus pass `null` here.

But now that you mention it, I'm repeating this code in BuildEnvironment.cs . This should become a separate method .. somewhere!

radical: Extract that to a `TestUtils.FindRepoRoot`.

PR #4693, opened 2024-06-27
Update ResourceNotificationServiceLogging

### LINKED ISSUE #4686: ResourceNotificationService logs that a resource changed its state when it didn't
Found this as part of updating the dotnet/aspire-samples tests to use the new resource log forwarding features in 8.1

The `ResourceNotificationService` logs that a resource has changed its state for every single resource notification published, rather than just when the actual state of the resource has changed:

https://github.com/dotnet/aspire/blob/21e35286025460824be1f1d149a17d60181b2b10/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L165-L176

This results in a lots of repeated logs about resource state changes where there is no state value, repeated state values, or the string "(null)".

```
[2024-06-26 21:21:42Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource weatherapi/weatherapi_6223aa3a-73q794o changed state: 
[2024-06-26 21:21:42Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource weatherapi/weatherapi_6223aa3a-73q794o changed state: 
[2024-06-26 21:21:42Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource weatherapi/weatherapi_6223aa3a-73q794o changed state: 
[2024-06-26 21:21:42Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource weatherapi/weatherapi_6223aa3a-73q794o changed state: Running
[2024-06-26 21:21:42Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource weatherapi/weatherapi_6223aa3a-73q794o changed state: Running -> Running
[2024-06-26 21:10:36Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource frontend/frontend_4cb5bdc8 changed state: 
[2024-06-26 21:10:36Z] dbug: Aspire.Hosting.ApplicationModel.ResourceNotificationService[0]
      Resource cache/cache_4cb5bdc8 changed state: (null)
```

I suggest we change the debug level log to only be written when an actual state change occurs that will result in a meaningful log message, and add a new trace level log that is emitted for every published update with the relevant details.

DamianEdwards:
Fixes #4686
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4693)

### DIFF
--- a/tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
+++ b/tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
@@ -305,6 +305,7 @@ public class ResourceNotificationTests
         var logs = logger.Collector.GetSnapshot();
 
         // Initial state text, log just the new state
+        Assert.Single(logs.Where(l => l.Level == LogLevel.Debug));
         Assert.Contains(logs, l => l.Level == LogLevel.Debug && l.Message.Contains("Resource resource1/resource1 changed state: SomeState"));
 
         logger.Collector.Clear();
@@ -314,6 +315,7 @@ public class ResourceNotificationTests
 
         logs = logger.Collector.GetSnapshot();
 
+        Assert.Empty(logs.Where(l => l.Level == LogLevel.Debug));
         Assert.DoesNotContain(logs, l => l.Level == LogLevel.Debug && l.Message.Contains("Resource resource1/resource1 changed state: SomeState"));
 
         logger.Collector.Clear();
@@ -323,6 +325,7 @@ public class ResourceNotificationTests
 
         logs = logger.Collector.GetSnapshot();
 
+        Assert.Single(logs.Where(l => l.Level == LogLevel.Debug));
         Assert.Contains(logs, l => l.Level == LogLevel.Debug && l.Message.Contains("Resource resource1/resource1 changed state: SomeState -> NewState"));
 
         logger.Collector.Clear();
@@ -332,6 +335,7 @@ public class ResourceNotificationTests
 
         logs = logger.Collector.GetSnapshot();
 
+        Assert.Empty(logs.Where(l => l.Level == LogLevel.Debug));
         Assert.DoesNotContain(logs, l => l.Level == LogLevel.Debug && l.Message.Contains("Resource resource1/resource1 changed state:"));
 
         logger.Collector.Clear();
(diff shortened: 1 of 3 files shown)


mitchdenny:
Is there any chance that we'll get two resource notifications for a resource being processed in parallel. I'm wondering if there is potential for resource states to rapidly oscillate and then we don't fire them.

DamianEdwards:
> Is there any chance that we'll get two resource notifications for a resource being processed in parallel. I'm wondering if there is potential for resource states to rapidly oscillate and then we don't fire them.

The publish method takes a lock on the state, so all updates for a resource get effectively sequentialized anyway, so while races might result in "extra" transitions being logged, I'm not sure any can be missed, e.g.

1. no state -> state1
2. state2 update comes in
3. state1 update comes in
4. state2 and state1 updates race such that state1 gets the lock first
5. no update is logged for state1 update as resource is already in that state
6. state2 update is logged as transition from state1 to state2

If in step 4 state2 update gets the lock first, then state1 to state2 transition is logged, and then state1 update logs transition from state2 to state1 (the "extra" update).

If we wanted to solve that we could use a `Channel` or similar but ultimately calls to `Publish` still race each other until the sync point.

tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
```diff
@@ -291,6 +293,87 @@ public async Task WaitingOnResourceThrowsOperationCanceledExceptionIfResourceDoe
...
+        var resource1 = new CustomResource("resource1");
+        var logger = new FakeLogger<ResourceNotificationService>();
+        var notificationService = new ResourceNotificationService(logger, new TestHostApplicationLifetime());
+
+        await notificationService.PublishUpdateAsync(resource1, snapshot => snapshot with { State = "SomeState" });
+
+        var logs = logger.Collector.GetSnapshot();
+
+        // Initial state text, log just the new state
+        Assert.Contains(logs, l => l.Level == LogLevel.Debug && l.Message.Contains("Resource resource1/resource1 changed state: SomeState"));
```
eerhardt, line 308: Do we want to assert how many things are in the logs?

eerhardt (approved):

tests/Aspire.Hosting.Tests/ResourceNotificationTests.cs
DamianEdwards: Yeah good point. I don't like making log tests perfectly tied to the logs as it makes them super fragile but in this case asserting the number of logs makes sense.

DamianEdwards: Adding assertion for the log counts.

PR #4696, opened 2024-06-27
Fix regression in port allocation for non http/https endpoints

davidfowl:
- We should only consider the http and https schemes for default port values (null)
- Updated the test to account for this

From https://github.com/dotnet/aspire/pull/4529#issuecomment-2193806629, found by a customer deploying the orleans example.

cc @davidebbo 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4696)


src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -378,7 +378,7 @@ public void WriteBindings(IResource resource)
 
                     // Check whether the project view this endpoint as Default (for its scheme).
                     // If so, we don't specify the target port, as it will get one from the deployment tool.
-                    (ProjectResource project, string uriScheme, null, _) when !schemesEncountered.Contains(uriScheme) => null,
+                    (ProjectResource project, string uriScheme, null, _) when uriScheme is "http" or "https" && !schemesEncountered.Contains(uriScheme) => null,
```
davidebbo, line 381: Maybe a bit better to do the http/https test at the time we add to the set? The set is really `schemesEncounteredAmongHttpAndHttps`

mitchdenny (approved):

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -358,7 +358,9 @@ public void WriteBindings(IResource resource)
             // 2. Default endpoints added by the framework
             // 3. Explicitly added endpoints
             // But wherever they come from, we treat the first one as Default, for each scheme.
-            var schemesEncountered = new HashSet<string>();
+            var httpSchemesEncountered = new HashSet<string>();
+
+            static bool IsHttpScheme(string scheme) => scheme is "http" or "https";
```
davidebbo, line 363: Ideally, we'd have a centralized helper for this check, as we make the same check for the env var injection (and possibly other places)

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
```diff
@@ -358,7 +358,9 @@ public void WriteBindings(IResource resource)
             // 2. Default endpoints added by the framework
             // 3. Explicitly added endpoints
             // But wherever they come from, we treat the first one as Default, for each scheme.
```
radical, line 360: Does this comment need to updated to reflect that this behavior is only for http/https?

tests/Aspire.Hosting.Tests/WithEndpointTests.cs
```diff
@@ -459,7 +459,8 @@ public async Task VerifyManifestProjectWithDefaultHttpEndpointsDoesNotAllocatePo
             .WithHttpEndpoint(name: "hp")       // Won't get targetPort since it's the first http
             .WithHttpEndpoint(name: "hp2")      // Will get a targetPort
             .WithHttpsEndpoint(name: "hps")     // Won't get targetPort since it's the first https
-            .WithHttpsEndpoint(name: "hps2");   // Will get a targetPort
+            .WithHttpsEndpoint(name: "hps2")   // Will get a targetPort
+            .WithEndpoint(scheme: "tcp", name: "tcp0");  // Will get a targetPort
```
radical, line 463: super nit: maybe an additional `.WithEndpoint(scheme: "tcp", name: "tcp1")` to ensure that it works for a second tcp endpoint too.

src/Aspire.Hosting/Publishing/ManifestPublishingContext.cs
davidebbo: Yes, and also that it only applies to project resources. e.g.

```c#
// This is used to determine if an http/https endpoint on a Project Resource should be treated as the Default endpoint
```

radical:
Merge `main` to pick up fixes for the unrelated tests failing here.

PR #4698, opened 2024-06-27
Use composite key from OTEL service name and instance id

### LINKED ISSUE #4217: Standalone Mode - Data being aggregated under a single application service name
I am using the Aspire Dashboard via Docker "latest" (Aspire version: 8.0.0-preview.7.24251.11+4952a6e837599afd6bc710bd6c010a2eacd65763) in stand alone mode for local development.

I have two separate applications ("App1" and "App2") configured in Program.cs identically like so, where the only deviation is the `serviceName` argument for the `.AddService` call: 

```
        var otelResourceBuilder = ResourceBuilder
            .CreateEmpty()
            .AddTelemetrySdk()
            .AddEnvironmentVariableDetector()
            .AddService(
                serviceName: "App1", // "App2" on the second app
                serviceInstanceId: Environment.MachineName);

        builder.Logging.AddOpenTelemetry(uu =>
        {
            uu.SetResourceBuilder(otelResourceBuilder);

               // ... options omitted
        });
        
        builder.Services.AddOpenTelemetry()
            .ConfigureResource(options =>
            {
                options = otelResourceBuilder;
            })
            .WithMetrics(uu =>
            {
               // ... registrations omitted
                uu.SetResourceBuilder(otelResourceBuilder);
            })
            .WithTracing(uu =>
            {
               // ... registrations omitted
                uu.SetResourceBuilder(otelResourceBuilder);
            });
```

I was expecting that when i run both applications at the same time, I could use the service drop down, and select each app individually to see their respective tracing and logging and metrics, etc.

However, I believe there is some bug that is aggregating everything under the first application's service name: In this screen shot, App2 was the first to hit the Aspire Dashboard, and you can see that when App1 started later, their logs were also grouped under App2's resource.

image

At first I thought it was an issue on my side, but after double and triple checking the service names, restarting the docker container, waiting 15+ minutes thinking the logs would switch, everything gets stuck under the first resource. 

To sanity check myself, I pulled down the Jaeger all-in-one docker container, and ran the same app, and they appropiately broke out the services as I would have expected to:

image

Is this actually a bug with the Aspire Dashboard or am I just incorrectly registering the service?

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/4217

The dashboard incorrectly assumes that each app instance has a unique instance id. That's true in Aspire land, but it isn't guaranteed when the standalone dashboard receives telemetry from 3rd party apps.

For example, someone might use the machine name as a common instance ID:

* Name = `App1`, Instance ID = `machine-name-1`
* Name = `App2`, Instance ID = `machine-name-1`

This PR changes dashboard to use `service.name` + `service.instance.id` as a composite key to identify apps. Allows for apps with different names to share an instance id.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4698)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
@@ -25,7 +25,7 @@ public partial class ChartContainer : ComponentBase, IAsyncDisposable
     private readonly InstrumentViewModel _instrumentViewModel = new InstrumentViewModel();
 
     [Parameter, EditorRequired]
-    public required string ApplicationId { get; set; }
+    public required ApplicationKey ApplicationKey { get; set; }
 
     [Parameter, EditorRequired]
     public required string MeterName { get; set; }
@@ -177,7 +177,7 @@ public partial class ChartContainer : ComponentBase, IAsyncDisposable
 
         var instrument = TelemetryRepository.GetInstrument(new GetInstrumentRequest
         {
-            ApplicationServiceId = ApplicationId,
+            ApplicationKey = ApplicationKey,
             MeterName = MeterName,
             InstrumentName = InstrumentName,
             StartTime = startDate,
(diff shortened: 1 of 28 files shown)


JamesNK:
Please review 🙏

adamint (approved):

Issue #4177, opened 2024-05-14
Add Ollama Component


luisquintanilla:
TLDR: Add [Ollama](https://ollama.com/) Component to Aspire similar to the [OpenAI](https://learn.microsoft.com/en-us/dotnet/aspire/azureai/azureai-openai-component?tabs=dotnet-cli) component.

## Configuration

You can run Ollama using containers. The general process is as follows:

1. Pull Ollama image

  - GPU

    ```bash
    docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
    ```

  - CPU

    ```bash
    docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
    ```
2. Download the model

```bash
docker exec -it ollama ollama run llama2
```

For more details, see the [blog announcement](https://ollama.com/blog/ollama-is-now-available-as-an-official-docker-image). 

In Aspire, this roughly translates into the following code inside the *AppHost* project:

```bash
var ollama =
    builder.AddContainer("ollama", "ollama/ollama")
        .WithVolumeMount("ollama", "/root/.ollama", VolumeMountType.Bind)
        .WithVolumeMount("./ollamaconfig", "/usr/config", VolumeMountType.Bind)
        .WithHttpEndpoint(hostPort: 11434, name: "ollama")
        .WithEntrypoint("/usr/config/entrypoint.sh");
```

The custom Entrypoint is to account for the second step which downloads the model. In this case, the entrypoint consists of a few scripts.

**entrypoint.sh**

```bash
#!/bin/bash

# Run model script
/usr/config/pullmodel.sh &

#Get model
/bin/ollama serve
```

**pullmodel.sh**

```bash
#!/bin/bash

errcode=1
start_time=$SECONDS
end_by=$((start_time + 60))

echo "Starting check for ollama start-up at $start_time, will end at $end_by"

while [[ $SECONDS -lt $end_by && ($errcode -ne 0)]]; do
	ollama list
	errcode=$?
	sleep 1
done

elapsed_time=$((SECONDS - start_time))

echo "Stopped checking for ollama start-up after $elapsed_time seconds (errcode=$errcode, seconds=$SECONDS)"

# Pull the model
ollama pull phi3
```

### Proposal

Simplify configuration of Ollama with a [custom resource type / component](https://learn.microsoft.com/dotnet/aspire/extensibility/custom-resources?tabs=unix). 

[...]

PR #4707, opened 2024-06-27
DRAFT: Adding Ollama as host resource

timheuer:
Starting WIP for adding Ollama (#4177) as a hosting resource for local LLM.  Incomplete, but draft PR for soliciting pattern of 'initializing' the ollama resource with models, which is required for the resource to be usable.

This uses a client library to perform the model pulls.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4707)


mitchdenny:
What does the deployment model look like here. Is the expectation that this is primarily for local dev scenarios and when deploying to production you would use a hosted AI service (for performance reasons).

timheuer:
> What does the deployment model look like here. Is the expectation that this is primarily for local dev scenarios and when deploying to production you would use a hosted AI service (for performance reasons).

Yes, that's my assumption is that this is a local dev scenario. I think that is a common pattern here, and that you'd have a conditional check for deployment and use a connection string to an OpenAI compatible endpoint (https://github.com/dotnet/aspire/pull/4707/files#diff-6b205cd5a1b204bf166dc8863d7fd0ca24c1bb0dea3d2f14ae73314ffd00e3a5R3-R9)

@luisquintanilla does that sound right? Are you seeing Ollama as a production-deployed LLM?

luisquintanilla:
> > What does the deployment model look like here. Is the expectation that this is primarily for local dev scenarios and when deploying to production you would use a hosted AI service (for performance reasons).
> 
> Yes, that's my assumption is that this is a local dev scenario. I think that is a common pattern here, and that you'd have a conditional check for deployment and use a connection string to an OpenAI compatible endpoint (https://github.com/dotnet/aspire/pull/4707/files#diff-6b205cd5a1b204bf166dc8863d7fd0ca24c1bb0dea3d2f14ae73314ffd00e3a5R3-R9)
> 
> @luisquintanilla does that sound right? Are you seeing Ollama as a production-deployed LLM?

Sounds about right. 

As I see it, there are two potential paths to production:

1. [Azure AI Model Catalog](https://azure.microsoft.com/en-us/products/ai-model-catalog) - You can deploy many of the open models provided by Ollama as a service to Azure. The challenge here is that since it's not 1:1 with Ollama, you'd have to use another component / client when going to production. 
2. [Azure Container Apps Services](https://learn.microsoft.com/en-us/azure/container-apps/services) - Enables you to take the same Ollama configuration to production so you can reuse the Ollama component / client / SK connector. 

Given that the 2nd doesn't exist today, I'd recommend prioritizing the Azure AI Model Catalog. 

One more thing to keep in mind:

> use a connection string to an OpenAI compatible endpoint

While this is somewhat of a standard, it's not "official", and implementations may be subject to change. Not unique to Ollama, but just something to be aware of.

timheuer:
Mostly @mitchdenny curious of opinions on the pattern here on 1) model acquisition and 2) using client lib to do the work

mitchdenny:
I'm happy with the answer that Ollama is primarily focused on local dev. I can see downsides of encouraging folks to download such a large model into a Container App instance. We should probably consider how expensive that would be for different model sizes (same issue for all compute environments that charge for compute / RAM per minute/second).

timheuer:
Closing this draft per discussions with @luisquintanilla on possible interesting deployment scenarios. I'll keep the branch active with these changes.

PR #4708, opened 2024-06-27
Remove DCP and Dashboard from the .NET Aspire SDK Workload

### LINKED ISSUE #5185: Remove DCP and Dashboard from the .NET Aspire Workload
Both DCP and Dashboard are part of your application when running it in a local environment, which means their version should be tied (and controlled) by the application as opposed to be tied to tooling (The workload you happen to have installed). This issue is to track the work of removing DCP and Dashboard from the workload, and instead allowing them to come from NuGet packages referenced by the aspire project.

joperezr:
Fixes https://github.com/dotnet/aspire/issues/5185

cc: @davidfowl @danegsta @mhutch @DamianEdwards @marcpopMSFT @eerhardt @danmoseley @maddymontaquila @radical 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4708)


docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,43 @@
+# Design proposal for the .NET Aspire workload going forward
```
joperezr, line 1: For some reason, in the PR review I'm not getting the images loaded. If you want to visualize the doc, you can do so here: https://github.com/joperezr/aspire/blob/workloads/MSBuildSDKs/docs/specs/future-of-workload.md

docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Proposed solution
+
+Given the above, we propose to move the DCP and Dashboard components out of the workload, and into NuGet packages that are referenced as dependencies of the Aspire.Hosting.AppHost package. This will allow the AppHost project to reference the version of Aspire.Hosting that it wants, and then the DCP and Dashboard will be installed as dependencies of that version of Aspire.Hosting. This will allow the AppHost project to have full control over the version of the DCP and Dashboard that it wants to use, and will also allow customers to have multiple AppHost projects that reference different versions of Aspire.Hosting without any issues. This will also allow us to update the DCP and Dashboard independently of the workload, which will allow us to ship updates to these components more frequently and with less risk, with a far less complex compatibility matrix.
```
DamianEdwards, line 19: What would happen if we were to remove the `Aspire.Hosting.SDK` completely, or update it to not rely on being loaded before restore? I understand that today it does some things based on those assumptions, but I'm interested in what in the experience would change if we removed it from the workload completely. There's also the option of a NuGet-based SDK (which I've resisted so far but I think we should explore all options).

docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,43 @@
...
+- We can update the DCP and Dashboard independently of the workload, which will allow us to ship updates to these components more frequently and with less risk.
+- Customers can stay in the latest version of the workload without worrying about compatibility issues with the AppHost project.
+
+### Cons
+
+- The main drawback of removing DCP and Dashboard from the workload, is the potential impacts it may have with restore times and local package caches. DCP especially is not small in size (~150mb) and it is also a native executable (written in Go), so we currently have separate packages each carrying the executable for each individual platform. One thing we were getting for free with DCP being part of the workload, is that the SDK authoring would automatically only install the DCP package for the platform that the SDK is being installed on. NuGet doesn't have great support today for platform-specific dependencies, which means that with the above proposal the Aspire.Hosting.AppHost package would depend on all of the DCP packages for all platforms, which would increase the size of the things that need to be restored, as well as the time it takes to do it.
```
DamianEdwards, line 38: Could we not utilize the ref-split packages pattern to achieve the same outcome of only downloading the runtime packages required for the current runtime? IIUC this is exactly how the .NET SDK itself avoids downloading runtime packages for shared frameworks until specifically required.

docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,43 @@
...
+- Additionally, workloads also have a way of keeping a reference count on workload packs, so it is able to clean them up when workloads don't reference it any more (for example, if you update your workload, then the old DCP package will be cleaned up automatically). This doesn't happen with regular NuGet packages, where customers are responsible for cleaning up their NuGet package caches themselves.
```
DamianEdwards, line 39: I'm not super concerned about this as managing the package restore location is already a scenario users have to manage.

danegsta:
We should test the DCP nuget packages on Mac/Linux to make sure there won't be any issues with the executable binaries; the SDK/workloads have a special mechanism for applying executable permissions (the UnixFilePermissions.xml file) and I'm not sure if simple nuget files support the same behavior.

docs/specs/future-of-workload.md
joperezr: These are the contents of `Aspire.Hosting.SDK` in question:

https://github.com/dotnet/aspire/blob/88c5c19567a12e2957e06952bbd9ef7a1a25a20e/src/Aspire.Hosting.Sdk/SDK/Sdk.targets#L9-L36

The most important part of this that **_needs_** to be evaluated before NuGet restore is the bit that is updating `ProjectReference` items. That code is what tells MSBuild that these aren't real project references, so it shouldn't treat them as such in restore. Without these, you'd have problems like restore graph of the AppHost also restoring packages coming from those project references (which could lift up dependencies, cause incompatible issues when packages are not applicable to some other project type, etc.), and it would also cause failures when those project references are targeting incompatible frameworks (as MSBuild would prevent that from working). For instance, if you had a net8.0 AppHost,  and a net9.0 frontend, that is allowed if this wasn't an SDK. @eerhardt can likely provide many more examples as he was the original one that added this logic and after different attempts found that the only way forward (if we wanted to use ProjectReference item) would be to use MSBuild SDKs.

There are alternatives though, that we can consider:
- We could decide to remove Aspire.Hosting.SDK from the workload, and instead have it be an SDK that is imported from the AppHost project itself (via `<Sdk Name="Aspire.Hosting.Sdk" Version="version" />`). The drawback of that is that it breaks existing customers as they need to change their projects, as well as MSBuild SDKs don't have a way to notify if there is an update, so people would likely get stuck on an old version of the SDK for a long time.
- We could decide not to use ProjectReference at all, and instead use a different Item for this. This would also break existing customers that would have to redefine how their apphost project references project resources, and would likely also impact some of the gestures that you get (especially in VS) automatically with Project References (e.g. drag and drop projects to automatically reference them, `dotnet add reference ..` support, etc.)

Overall, the reason why I picked this as the proposal is that the code in that SDK is very minimal and unlikely to change  going forward (proof of it being it hasn't changed since it was added), hence in favor of not breaking existing customers and not regress some of the scenarios, it feels like keeping that single SDK in the workload is a good tradeoff.

docs/specs/future-of-workload.md
joperezr: Sorry, I'm not sure I'm following here. Shared frameworks are kind of special here as they are built-in on the SDK itself, and one of the "implicit" references, so the SDK has special logic to deal with that and not have to pull in all different platform specific runtime packs. For regular NuGet packages, AFAIK there isn't a NuGet built-in way of restoring platform-specific packages unless you restore using a RID, which IMO is not something we want to enforce on our customers. That's how the whole runtime.json stuff for RID specific dependencies worked, it basically relied on people doing `dotnet restore -r  win-x64` so the right section of the runtime.json could be picked.

docs/specs/future-of-workload.md
joperezr: Fair enough, I still wanted to mention it because technically it is a drawback, but I agree with you that is not a major concern and it still seems like the right move.

joperezr:
> We should test the DCP nuget packages on Mac/Linux to make sure there won't be any issues with the executable binaries; the SDK/workloads have a special mechanism for applying executable permissions (the UnixFilePermissions.xml file) and I'm not sure if simple nuget files support the same behavior.

Good point. I did test this on Linux and Windows, and all scenarios I tested work as expected, but I didn't test on Mac yet.

docs/specs/future-of-workload.md
DamianEdwards: Thanks for the extra detail. I think it's important this proposal has details of all the possible approaches and the trade-offs with each so we can make an informed decision.

> The drawback of that is that it breaks existing customers as they need to change their projects, as well as MSBuild SDKs don't have a way to notify if there is an update, so people would likely get stuck on an old version of the SDK for a long time.

I've been very resistant to use [NuGet-resolved MSBuild SDKs](https://learn.microsoft.com/visualstudio/msbuild/how-to-use-project-sdk) in the past but the reasons cited here seem like that could be reasonably mitigated with targets in `Aspire.Hosting.AppHost` that check the SDK and version being used, and if necessary emit a warning advising the user what action to take to bring them up to date. For example, assuming we make the changes in `8.2`:

- If a user updates `Aspire.Hosting.AppHost` to `8.2` but doesn't change anything else about the project, the targets within can detect that (`IsAspireHost==true`, `UsingAspireNuGetSDK!=true`) and instruct the user to remove the `IsAspireHost` property (which triggers the workload detection today) and change the project SDK to `<Project SDK="Aspire.Hosting.SDK/8.2.0">`.
- Moving forward, if a new version of `Aspire.Hosting.AppHost` introduces a requirement for a new SDK min version, that can be detected by the targets and emit a warning instructing the user to update the SDK reference to the new version.

The initial project change is one-time only, and after that a change to SDK version in the project is only required when needed (likely rarely) and the user can get clear instructions on what to do.

I'm still keen to keep using `ProjectReference` for the reasons you cited (i.e. existing tooling gestures just work) which means keeping the requirement of an MSBuild SDK so using NuGet-resolved MSBuild SDKs seems like the approach we should consider (and scrutinize) more closely.

docs/specs/future-of-workload.md
DamianEdwards: I just meant using the pattern of splitting different platform implementations into `runtime.[rid].` prefixed packages and then having our targets in `Aspire.Hosting` take care of restoring them and injecting the appropriate paths into the AppHost assembly. I think this is very similar to how things like the ILCompiler toolchain works today, e.g. [runtime.win-x64.Microsoft.DotNet.ILCompiler](https://www.nuget.org/packages/runtime.win-x64.Microsoft.DotNet.ILCompiler/)

docs/specs/future-of-workload.md
joperezr: Just to make sure I'm following, the alternative you are suggesting is different from my proposal in that you want to remove Aspire.Hosting.SDK from the workload, and instead move it to be an SDK reference coming from AppHost project itself, correct?

DamianEdwards: Correct. In short, I want to ensure we explore options that remove the need for the workload to be installed at all.

joperezr: Gotcha. So I did try that out, and it indeed works. I can add this to the doc as alternative solutions, but I saw two minor drawbacks with that approach:

- It makes it so that people will need that one time update on their project in order to add the import of the SDK.
- MSBuild SDKs don't have a good update mechanism, meaning if we ever wanted to ship a new update, consumers wouldn't be notified in any way, so they would have to know to go an update their csproj to use the new versions on their own. Having it on the workload, on the other hand, at least will still make it that running workload update will bring the updates, and it also means consumers will get the notification in the console and VS when their workloads have updates.

Both of these are not terrible, but given the fact that at least for the first iteration we need to still keep a workload (given we need to update users machines so DCP and Dashboard are not included in it) then I decided it was a better approach to keep the Hosting SDK in the workload and then revisit this if we look into removing the workload all together, which this proposal would totally allow.

docs/specs/future-of-workload.md
eerhardt: (Note: I'm not an expert here, so I could be wrong)

The ILCompiler and crossgen are "special" because they are known by the SDK explicitly.

https://github.com/dotnet/sdk/blob/e80b6ebc8b92ce2026f82cbfa34257d05d58cf83/src/Installer/redist-installer/targets/GenerateBundledVersions.targets#L553-L565

And then that `**RID**` gets replaced with the correct RID in a C# task:

https://github.com/dotnet/sdk/blob/e80b6ebc8b92ce2026f82cbfa34257d05d58cf83/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs#L789

I'm not sure we could take this same approach outside of the SDK.

There is/was a feature in NuGet that we used to use called `runtime.json`, but AFAIK it was never officially supported outside of System.* packages (which stopped using the feature).

See:

* https://github.com/NuGet/Home/issues/6083
* https://github.com/NuGet/Home/issues/10571
* https://github.com/NuGet/Home/issues/1660
* https://github.com/NuGet/Home/issues/3114

docs/specs/future-of-workload.md
maddymontaquila: It's definitely a concern we should bring back up with the nuget folks (cough cough @JonDouglas ) because it's an eternal pain point lol. And, in the future, if Aspire is used by people who aren't primarily .NET devs they might not know to clear it. But agree not a blocker, especially for now!

docs/specs/future-of-workload.md
maddymontaquila: We can use a targetednotification (aka gold bar) in VS on Aspire projects using a certain version via whatever Kusto query gives us that info to tell them to migrate, but I am hesitant to break people (gold bars also get such a low interaction rate lol). Also in VS Code there isn't the same type of TargetedNotif infrastructure, so you would have to ask C#DK to hardcode a popup notification in there for us lol.

docs/specs/future-of-workload.md
joperezr: I also believe this is not something that any package can do, but I'll look into it.

docs/specs/future-of-workload.md
danegsta: I know that, at least in the past, there were some severe performance implications related to using nuget based SDKs in Visual Studio. Has that been resolved?

joperezr: If the problem you refer to is that they get restored in the UI thread, then I don't think that's fixed. However, the Hosting SDK being just a single targets file with less than 50 lines of code I'm not too concerned. (I'd be much more concerned if we were restoring Dashboard and DCP as MSBuild SDKs which at some point I tried in my prototype)

DamianEdwards: RE this

>MSBuild SDKs don't have a good update mechanism, meaning if we ever wanted to ship a new update, consumers wouldn't be notified in any way, so they would have to know to go an update their csproj to use the new versions on their own. Having it on the workload, on the other hand, at least will still make it that running workload update will bring the updates, and it also means consumers will get the notification in the console and VS when their workloads have updates.

As I lay out in https://github.com/dotnet/aspire/pull/4708/files#r1657925144 I think we can easily improve this by adding checks in the targets brought in my `Aspire.Hosting.AppHost`.

docs/specs/future-of-workload.md
DamianEdwards: I believe you're right @eerhardt, but why can't we take a similar approach? What is it that the SDK does from its MSBuild targets that's so special that we can't do it from targets in `Aspire.Hosting.AppHost`? We only need to facilitate the download of the required runtime package at the right time. Things like `dotnet new install` do this today, as well as the SDK itself of course for scenarios like self-contained publish. Once the runtime-specific package is downloaded its path can be injected into the AppHost app via its targets.

joperezr: I may be missing something here, but the main difference would be that this SDK logic (because it is in the SDK) is able to pick the right RID and add the right PackageReference _before_ restore happens. If we were to add something like this in our Aspire.Hosting.AppHost targets, by definition those cannot run at restore time (as they need to be restored in the initial phase), meaning the only way to add the PackageReferences for the specific RID packages from there would be at Build time (as opposed to Restore time) which is a big no-no. Restore graph should never be modified post initial-restore as a lot of things break if at build time you are trying to add new references.

DamianEdwards: Sorry, I'm not saying you do the restore as PackageReferences, that's not required if you're just trying to ensure packages are downloaded. These are tools, there's nothing in them that needs to be referenced by the app. We just need them on the machine, so manually restoring at some other point is fine.

joperezr: I see. How would we find the right path to the tools then? The other thing that these packages have is a targets file that sets the path to the tools relative to the targets file itself, and the targets are only loaded because this is a PackageReference. If we don't load these targets and instead assume what the right path to the restored tools are, we may run into issues when customers don't use default restore path, which I suppose is not terribly un-common.

docs/specs/future-of-workload.md
joperezr: Yup, that's a fair point. I think that in the future if we want to not have a workload at all then having this be an MSBuild SDK would be relatively easy and we can use what you suggest to keep it up-to-date when needed. I would just rather do this change in stages, where we first focus on removing DCP and Dashboard, and then later we can move Project Templates to IDE/SDK and Hosting.SDK to be an MSBuild sdk.

docs/specs/future-of-workload.md
eerhardt: > What is it that the SDK does from its MSBuild targets that's so special that we can't do it from targets in Aspire.Hosting.AppHost?

It gets to inject logic that influences restore. That Task I referenced above gets run before restore happens.

When would the download of DCP/Dashboard actually happen?

1. During `dotnet restore` - this is when the `Aspire.Hosting.AppHost` nupkg would get restored. Note that MSBuild logic inside that nupkg can't run during restore because it isn't there yet - it hasn't been restored.
2. During `dotnet build` - here we CAN run logic from inside the `Aspire.Hosting.AppHost` nupkg. So we can download the nupkgs for DCP/Dashboard. But note that the MSBuild logic from inside those nupkgs can't influence the build anymore, because the build is already happening. I'm not sure there is a way to dynamically load more MSBuild logic once the build as started.

So to make it work this way with `Aspire.Hosting.AppHost`, we would need to download the executable packages during build - and we would need to have the logic about where the executables exist inside those nupkgs in `Aspire.Hosting.AppHost`.

To actually do the download during restore, we would need to inject MSBuild logic into the project that influences restore - i.e. either in the workload as we do today, or in an MSBuild Sdk.

joperezr: I'll chat with @rainersigwald about this to get his feedback. Even if they are not package references and  we are just downloading the tools, hitting the network at build time is typically an anti-pattern. I'm also concerned with the assumptions we'd have to make from `Aspire.Hosting.AppHost` package targets to be able to find the restored path to the tools.

DamianEdwards: We'd only perform the download if required and the assumptions about paths could be avoided by controlling where the download is made to (again, just like `dotnet new install` does).

eerhardt: I don't think doing the download during build is going to be a great user experience. I also don't think we want to sign up for handling the download of these packages ourselves. That seems like a bug farm.

If we are going with RID-split packages, the best option (IMO) is to get NuGet to manage the downloading/restoring/caching/etc, which means getting our logic injected so it can influence restore.

docs/specs/future-of-workload.md
JonDouglas: Yes, please discuss your NuGet needs for this work long term with us. Whether that's moving away from workloads or not, we'd love to help.

docs/specs/future-of-workload.md
DamianEdwards: > If we are going with RID-split packages, the best option (IMO) is to get NuGet to manage the downloading/restoring/caching/etc, which means getting our logic injected so it can influence restore.

Yeah I can't argue with that. Is it as simple as having our SDK targets set `<UseCurrentRuntimeIdentifier>` to "true" before restore, and having `Aspire.Hosting.AppHost` depend on `Aspire.Hosting.Orchestration`, which in turn contains a `runtime.json` file that defines the runtime-specific packages to bring in?

radical: There is https://learn.microsoft.com/en-us/nuget/consume-packages/packagedownload-functionality also. This will just download the specific nuget as part of the regular restore.

`dotnet/runtime` uses this for wasm in a couple of places.

eerhardt: >  Is it as simple as having our SDK targets set <UseCurrentRuntimeIdentifier> to "true" before restore, and having Aspire.Hosting.AppHost depend on Aspire.Hosting.Orchestration, which in turn contains a runtime.json file that defines the runtime-specific packages to bring in?

I don't think we want to use `runtime.json`. See https://github.com/dotnet/runtime/issues/49137 where the runtime is trying to get off that feature.

docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,43 @@
...
+
+Also, there is no real need of including the Aspire.Hosting package as part of the workload, as there is no guarantee that this is the version that the AppHost project will want to use, therefore we would also remove this package from the workload. In the end, the workload will only contain the Aspire.Hosting.SDK and Aspire.ProjectTemplates components, which are the ones that can be truly considered as tooling components, and in both cases you would want to always be in the latest version. These two are also not tightly coupled to the version of the Aspire.Hosting package that the AppHost project references, so they can be updated independently of the AppHost.
```
marcpopMSFT, line 21: The tight version coupling that DCP and Dashboard have makes sense in the explanation above. For Aspire.Hosting, I wasn't exactly clear. Is it just that it's not needed to load before restore so it doesn't need to be a workload so just include it in the set of packages?

docs/specs/future-of-workload.md
marcpopMSFT: I can confirm that msbuild SDKs are still loaded on the UI thread. Jeff did a bunch of work to fix deadlocks but depending in network connectivity and size, there can be impact. msbuild sdks are fairly commonly used though and much of the perf/hang complaints come from folks with fairly complex builds and larger sdks.

docs/specs/future-of-workload.md
marcpopMSFT: CC @dsplaisted who's our expert on packagedownload as well as the special processing we do for RIDs for framework packages.

docs/specs/future-of-workload.md
DamianEdwards: > Is it just that it's not needed to load before restore so it doesn't need to be a workload so just include it in the set of packages?

Correct. `Aspire.Hosting` is just a normal package in that sense.

docs/specs/future-of-workload.md
dsplaisted: I don't believe the existing NuGet runtime.json / RID-specific functionality would work for this at all (even if we weren't trying to avoid using it in general).  That feature lets you get different packages based on the RuntimeIdentifier the app is using.  However, what we want here is to get the DCP etc. packages for the RuntimeIdentifier of the host.  If we support cross-targeting (ie building an Aspire app for linux-x64 from a windows-x64 machine), then NuGet's runtime.json functionality doesn't help.

It may be possible to avoid downloading DCP etc. for all platforms with logic in the Aspire SDK targets.  There could be logic that detected a PackageReference to `Aspire.Hosting.AppHost`, and if it was found would add `PackageReference` or `PackageDownload` items to the corresponding platform-specific versions of the DCP packages.  That might have some issues and wouldn't work for transitive references to `Aspire.Hosting.AppHost`, but if I understand correctly there may not be scenarios where that package is transitively referenced.

docs/specs/future-of-workload.md
dsplaisted: NuGet SDKs are used "a lot" (for some value of "a lot"), but I think nowhere near as much as we expect Aspire to be used.  If we move all of the large components out of the workload, then I would consider moving what's left (mainly just props and targets I think) into the .NET SDK itself rather than out into a NuGet MSBuild SDK.  You would lose the ability to version those components separately from the SDK, but I think that's actually probably an advantage, as there wouldn't be an extra, nonstandard version in the project file that had to be managed.

docs/specs/future-of-workload.md
DamianEdwards: > If we support cross-targeting (ie building an Aspire app for linux-x64 from a windows-x64 machine), then NuGet's runtime.json functionality doesn't help.

The AppHost project is never deployed so I don't think this is an issue.

> It may be possible to avoid downloading DCP etc. for all platforms with logic in the Aspire SDK targets. There could be logic that detected a PackageReference to Aspire.Hosting.AppHost, and if it was found would add PackageReference or PackageDownload items to the corresponding platform-specific versions of the DCP packages. That might have some issues and wouldn't work for transitive references to Aspire.Hosting.AppHost, but if I understand correctly there may not be scenarios where that package is transitively referenced.

This is the approach being proposed now I believe.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,105 @@
...
+         to avoid NuGet restore complaining about using CPM but having versioned PackageReferences. -->
+    <ItemGroup Condition=" '$(AppHostVersion)' != '' and '$(_UsingCentralPackageManagement)' == 'true' ">
+      <PackageVersion Include="Aspire.Dashboard.Sdk.$(NETCoreSdkRuntimeIdentifier)"
+  Version="$(AppHostVersion)" />
+      <PackageVersion Include="Aspire.Hosting.Orchestration.$(NETCoreSdkRuntimeIdentifier)"
+  Version="$(AppHostVersion)" />
+
+      <PackageReference Include="Aspire.Dashboard.Sdk.$(NETCoreSdkRuntimeIdentifier)" />
+      <PackageReference Include="Aspire.Hosting.Orchestration.$(NETCoreSdkRuntimeIdentifier)" />
+    </ItemGroup>
```
danegsta, line 116: Would using IsImplicitlyDefined=true be an option for these PackageReferences? I believe implicit references are exempted from the CPM PackageVersion requirement (although it's been a while since I tested). It'd potentially allow for simplifying these references.

joperezr: I can test that, but I also don't know what happens if you set that in target files that are not inside the main .NET SDK.

joperezr: Yup, you are right, this works. I'll do that to simplify

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
   </Target>
 
-  <Target Name="__WarnOnMininumVsVersionMissing" BeforeTargets="PrepareForBuild" Condition="'$(BuildingInsideVisualStudio)' == 'true' and $([MSBuild]::VersionLessThan('$(MSBuildVersion)', '17.10.0'))">
-    <Warning Code="ASPIRE003" Text="$(MSBuildProjectName) is a .NET Aspire AppHost project that requires Visual Studio version 17.10 or above to work correctly. You are using version $(MSBuildVersion)." />
+  <!-- This target extracts the version of Aspire.Hosting.AppHost referenced by the project, and adds
+       a reference to Aspire.Dashboard.Sdk and Aspire.Hosting.Orchestration for the build-time platform using
+       the same version. This is done here dynamically to avoid having to pull in DCP and Dashboard packages
+       for all of the platforms. This mechanism can be disabled by setting `SkipAddReferenceToDashboardAndDCP` to `true` -->
+  <Target Name="AddReferenceToDashboardAndDCP" BeforeTargets="Restore;CollectPackageReferences" Condition="'$(SkipAddReferenceToDashboardAndDCP)' != 'true'">
```
radical, line 63: If `SkipAddReferenceToDashboardAndDCP` is only meant for tests then maybe rename this to be a private property like `_ SkipAddReferenceToDashboardAndDCP`.

tests/Shared/Aspire.Workload.Testing.targets
```diff
@@ -39,30 +39,20 @@
...
-    <!-- FIXME:
-      Explicitly invoke Build on these two projects so the GenerateTextReplacementFiles target gets triggered.
-      This can be removed once the workloadtesting targets support a custom list of targets to invoke
-    -->
-    <MSBuild Projects="$(RepoRoot)eng/dcppack/dcppack.csproj"
-              Properties="$(_CombinedProps)"
-              Targets="Restore;Build" />
-    <MSBuild Projects="$(RepoRoot)eng/dashboardpack/dashboardpack.csproj"
-              Properties="$(_CombinedProps)"
-              Targets="Restore;Build" />
```
radical, line 65: Removing this breaks local command line builds:
1. `./build.sh -pack`
2. `dotnet build tests/workloads.proj`
3. `dotnet test tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj`

`/Users/ankj/dev/a2/artifacts/bin/Aspire.Workload.Tests/Release/net8.0/nuget-cache-Net80/aspire.dashboard.sdk.osx-arm64/8.2.0-dev/buildTransitive/Aspire.Dashboard.Sdk.osx-arm64.props(2,3): error MSB4019: The imported project "/Users/ankj/dev/a2/artifacts/bin/Aspire.Workload.Tests/Release/net8.0/nuget-cache-Net80/aspire.dashboard.sdk.osx-arm64/8.2.0-dev/build/@PackageName@.props" was not found. Confirm that the expression in the Import declaration "../build/@PackageName@.props" is correct, and that the file exists on disk. [/private/var/folders/dj/z2lsb9l9087356sgyzrlklvr0000gn/T/testroot/aspire-aspire-xunit/aspire-aspire-xunit.AppHost/aspire-aspire-xunit.AppHost.csproj]`

tests/Directory.Build.props
```diff
@@ -15,6 +15,12 @@
 
     <!-- This is useful for local runs -->
     <VSTestResultsDirectory>$(ArtifactsTestResultsDir)</VSTestResultsDirectory>
+
+    <!-- 
+      Test applications have their own way of referencing Aspire.Hosting.AppHost, as well as DCP and Dashboard, so we disable
+      the Aspire.Hosting.SDK targets that will automatically add these references to projects. 
+    -->
+    <SkipAddReferenceToDashboardAndDCP Condition="'$(TestsRunningOutsideOfRepo)' != 'true'">true</SkipAddReferenceToDashboardAndDCP>
```
radical, line 23: This should be moved to `tests/Shared/RepoTesting/Aspire.RepoTesting.props`, and can be dropped from `playground/Directory.Build.props`.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+
+    <PropertyGroup Condition="'$(AppHostVersion)' != ''">
+      <__CurrentAppHostVersionMessage> You are using version $(AppHostVersion).</__CurrentAppHostVersionMessage>
+    </PropertyGroup>
+
+    <!-- At this point, we should have the version either by CPM or PackageReference, so we fail if not.  -->
+    <Error Condition="'$(AppHostVersion)' == '' or $([MSBuild]::VersionLessThan('$(AppHostVersion)', '8.2.0'))" 
+    Text="$(MSBuildProjectName) is a .NET Aspire AppHost project that needs a package reference to Aspire.Hosting.AppHost version 8.2.0 or above to work correctly.$(__CurrentAppHostVersionMessage)" />
+
+    <!-- If not using CPM, then directly add the two PackageReferences -->
```
radical, line 97: This comment sounds wrong because irrespective of CPM we are adding the `PackageReferences`, IIUC.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+    <ItemGroup Condition=" '$(AppHostVersion)' != ''">
```
radical, line 98: The condition can be dropped as it will always be `true`.

radical:
Most of my feedback can be addressed in a follow up.

eng/Build.props
```diff
@@ -1,8 +1,10 @@
 <Project>
   <ItemGroup Condition="'$(DotNetBuildFromSource)' != 'true' and '$(DotNetBuild)' != 'true'">
     <ProjectToBuild Include="$(RepoRoot)src\**\*.csproj" Exclude="$(RepoRoot)src\Aspire.ProjectTemplates\templates\**\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)tests\**\*.csproj"  />
+    <ProjectToBuild Include="$(RepoRoot)eng\dcppack\**\*.csproj" Exclude="$(RepoRoot)src\Aspire.ProjectTemplates\templates\**\*.csproj" />
+    <ProjectToBuild Include="$(RepoRoot)eng\dashboardpack\**\*.csproj" Exclude="$(RepoRoot)src\Aspire.ProjectTemplates\templates\**\*.csproj" />
```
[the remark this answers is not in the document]
joperezr: Ooh, good catch, I think it is a copy paste error.

eng/dcppack/Aspire.Hosting.Orchestration.win-x86.csproj
```diff
@@ -0,0 +1,9 @@
+<Project>
```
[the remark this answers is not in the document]
joperezr: Yeah, we probably can now. We originally needed the individual projects because we were having P2P references from the AppHost package, beacause we needed to have a package reference. Now that those are being dynamically added by the SDK, we no longer need this. I'll try to get this done in the next day but if I can get to it I may have to punt this to a follow up PR.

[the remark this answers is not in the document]
radical: Follow up PR sounds perfectly reasonable for this.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+      <AppHostPackageReference Include="@(PackageReference)" Condition="'%(Identity)' ==
+  'Aspire.Hosting.AppHost'" />
+    </ItemGroup>
+
+    <!-- Extract the Version metadata -->
+    <PropertyGroup>
+      <AppHostVersion>%(AppHostPackageReference.Version)</AppHostVersion>
+    </PropertyGroup>
+
+    <!-- If the version is still empty, then we assume the project is using Central Package Management,
```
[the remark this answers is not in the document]
joperezr: I don't know CPM enough to see if this will always be the case (as-in not sure if you could have PackageVersion without setting that). If you know that for sure, I can change it, or we can change in a follow-up.

tests/Aspire.Hosting.Tests/MSBuildTests.cs
```diff
@@ -53,6 +53,13 @@ public class Class1
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <IsAspireHost>true</IsAspireHost>
+
+    <!-- 
+      Test applications have their own way of referencing Aspire.Hosting.AppHost, as well as DCP and Dashboard, so we disable
+      the Aspire.Hosting.SDK targets that will automatically add these references to projects. 
+    -->
+    <SkipAddReferenceToDashboardAndDCP Condition="'$(TestsRunningOutsideOfRepo)' != 'true'">true</SkipAddReferenceToDashboardAndDCP>
+    <AspireHostingSDKVersion>8.2.0</AspireHostingSDKVersion>
```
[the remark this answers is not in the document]
joperezr: Not really, this just needs a minimum version which we check in the AppHost, which won't change across versions until we have a new breaking change.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+  <!-- This target extracts the version of Aspire.Hosting.AppHost referenced by the project, and adds
+       a reference to Aspire.Dashboard.Sdk and Aspire.Hosting.Orchestration for the build-time platform using
+       the same version. This is done here dynamically to avoid having to pull in DCP and Dashboard packages
+       for all of the platforms. This mechanism can be disabled by setting `SkipAddReferenceToDashboardAndDCP` to `true` -->
+  <Target Name="AddReferenceToDashboardAndDCP" BeforeTargets="Restore;CollectPackageReferences" Condition="'$(SkipAddReferenceToDashboardAndDCP)' != 'true'">
+
+    <!-- First, we assume project is not using Central Package Management, and we try to extract the version
+         from PackageReference Items. -->
+    <ItemGroup>
+      <AppHostPackageReference Include="@(PackageReference)" Condition="'%(Identity)' ==
```
radical, line 68: nit: make the item private, since this will get used in user projects too.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+  'Aspire.Hosting.AppHost'" />
+    </ItemGroup>
+
+    <!-- Extract the Version metadata -->
+    <PropertyGroup>
+      <AppHostVersion>%(AppHostPackageReference.Version)</AppHostVersion>
```
radical, line 74: nit: private

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
```diff
@@ -13,34 +15,92 @@
...
+    <PropertyGroup Condition="'$(AppHostVersion)' == ''">
+      <AppHostVersion>%(AppHostPackageReference.Version)</AppHostVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(AppHostVersion)' != ''">
+      <__CurrentAppHostVersionMessage> You are using version $(AppHostVersion).</__CurrentAppHostVersionMessage>
+    </PropertyGroup>
+
+    <!-- At this point, we should have the version either by CPM or PackageReference, so we fail if not.  -->
+    <Error Condition="'$(AppHostVersion)' == '' or $([MSBuild]::VersionLessThan('$(AppHostVersion)', '8.2.0'))"
```
radical, line 94: nit: It might be useful to split this into two separate error messages, to help with debugging when this fails.

1. `AppHostVersion==''`, we could not figure out the version to use, and have a better error message in case of CPM.
2. `AppHostVersion < 8.2.0`

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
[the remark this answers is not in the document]
radical: That makes sense. I was mainly thinking about the case where we might be able to "know" that when CPM is enabled, and a `PackageVersion` must be available, and thus show a good error.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
joperezr: It is not meant only for internal. It is common for packages that are doing significant work to allow switches so that the user can optionally disable pieces of functionality. That was the thinking when this condition was set.

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
joperezr: We would know at that point. If 1., then the warning will not have the version you are using (see line 90 above). If 2 then you'd see which version you are using in the error message.

radical (approved):
This is great work! LGTM! 👍

docs/specs/future-of-workload.md
```diff
@@ -0,0 +1,33 @@
...
+- Aspire.Hosting.Orchestration (MSBuild SDK): This SDK contains two main things: the Developer Control Plane executable (DCP) as well as a targets file which defines properties for the AppHost to be able to find the DCP executable and run it.
+- Aspire.Dashboard.SDK (MSBuild SDK): Similar to the Orchestration SDK, this one provides the .NET Aspire Dashboard application (self-contained app that depends on .NET 8 runtime) and the targets file that defines properties for the AppHost to be able to find the Dashboard executable and launch it.
+- Aspire.Hosting (Regular NuGet Library): This package isn't really required to be in the workload, as this is a PackageReference that all AppHost projects will have. The advantage that we get currently from having it in the workload is that this would install it into the global NuGet cache when the workload is installed, so that the AppHost projects can find it without needing to restore it from the NuGet feed. This is helpful especially for offline scenarios. The package itself provides the Aspire.Hosting library which contains all of the main logic in Aspire for building and running .NET Aspire applications.
```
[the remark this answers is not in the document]
joperezr: I did actually mean Aspire.Hosting as that is the one that we had in the workload, but the wording is wrong here I agree, so I'll fix that.

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
```diff
@@ -246,6 +246,12 @@ namespace Projects%3B
     <MakeDir Directories="$(AspireManifestPublishOutputPath)" />
   </Target>
 
+  <Target Name="__ErrorOnMininumWorkloadVersionMissing" BeforeTargets="PrepareForBuild"
+    Condition="'$(IsAspireHost)' == 'true' and ('$(AspireHostingSDKVersion)' == '' or $([MSBuild]::VersionLessThan('$(AspireHostingSDKVersion)', '8.2.0')))">
+    <Error Code="ASPIRE005"
+      Text="$(MSBuildProjectName) project requires a newer version of the .NET Aspire Workload to work correctly. Please run `dotnet workload update`." />
```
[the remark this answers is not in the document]
joperezr: > Should this message say what version is required?

I'm not so sure about this one, but I am open to suggestions. The thinking is that this minimum version will change over time, so I'm not sure if we want consistent messages, especially when the fix is to run `dotnet workload update` which you can't really use to select which version to move to (as it always goes to latest).

> Do we need a test for this?

I do think this would be good and I thought about it, but the problem is that it won't be trivial as I don't think we have a way of running tests with an SDK without the workload, or with an older workload, but @radical can correct me if I'm wrong.

tests/Shared/RepoTesting/Aspire.RepoTesting.targets
```diff
@@ -143,4 +143,9 @@
     <Import Project="$(RepoRoot)src\Aspire.Hosting.Sdk\SDK\Sdk.targets" Condition="Exists('$(RepoRoot)src\Aspire.Hosting.Sdk\SDK\Sdk.targets')" />
   </ImportGroup>
 
+  <PropertyGroup Condition="'$(RepoRoot)' != 'null' and '$(TestsRunningOutsideOfRepo)' != 'true' and '$(IsAspireHost)' == 'true'">
```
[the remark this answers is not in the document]
joperezr: I used the same condition than the one used by the itemgroup above (as I want it to apply exactly in the same situation). If we are changing one, we should change both.

eng/dashboardpack/Aspire.Dashboard.Sdk.linux-arm64.csproj
```diff
@@ -0,0 +1,9 @@
+<Project>
```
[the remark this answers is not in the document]
joperezr: We could, I believe. We actually no longer need to have all of these separate projects anymore, so in a follow up PR I am going to explore how to reduce and simplify this.

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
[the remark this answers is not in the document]
radical: > > Do we need a test for this?

> I do think this would be good and I thought about it, but the problem is that it won't be trivial as I don't think we have a way of running tests with an SDK without the workload, or with an older workload, but @radical can correct me if I'm wrong.

We do have a way to run tests with a SDK that has the updated manifest from `artifacts`, but does not have the workload installed. I can add that test.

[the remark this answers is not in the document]
radical: > with an older workload
.. this would need some work. Do we want this one?

src/Aspire.Hosting.Sdk/SDK/Sdk.targets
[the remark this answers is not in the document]
radical: `SkipImplicitAspireReferences`?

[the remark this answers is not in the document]
radical: or `AspireSkipImplicitReferences`

Issue #4706, opened 2024-06-27
After 10K traces the count in the dashboard does not update


kvenkatrajan:
Refer to : https://github.com/microsoft/azure-container-apps/issues/1213

https://github.com/microsoft/azure-container-apps/issues/1213#issuecomment-2195497452

JamesNK:
@snehapar9 Is the dashboard not updating with new traces, or has the number stopped at 10k? The number stopping at 10k is expected, as that's the in-memory limit. At that point, old traces are removed as new traces are added.

Continuing to increase the number without the data isn't technically possible.

What about adding an indication somewhere to the UI that the memory limit has been reached and old traces are being removed?

leslierichardson95:
> @snehapar9 Is the dashboard not updating with new traces, or has the number stopped at 10k? The number stopping at 10k is expected, as that's the in-memory limit. At that point, old traces are removed as new traces are added.
> 
> Continuing to increase the number without the data isn't technically possible.
> 
> It seems like the problem is that the user thinks telemetry is no longer updating. What about adding an indication somewhere to the UI that the memory limit has been reached and old traces are being removed?

@JamesNK @kvenkatrajan could we consider adding a disclaimer like "The 10,000 most-recent traces are displayed"?  It could be added as a subtitle that persists under the Traces header.

JamesNK:
There is a PR here: https://github.com/dotnet/aspire/pull/4799. The text is easy to change.

I don't think it should be something that is always displayed. The majority of the time the dashboard won't reach the maximum data limit. The message should be displayed in that situation, and not unnecessarily.

PR #4712, opened 2024-06-28
Add test for exceeding trace limit

JamesNK:
Addresses https://github.com/dotnet/aspire/issues/4706
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4712)

### DIFF
--- a/tests/Aspire.Dashboard.Tests/TelemetryRepositoryTests/TraceTests.cs
+++ b/tests/Aspire.Dashboard.Tests/TelemetryRepositoryTests/TraceTests.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Globalization;
 using Aspire.Dashboard.Otlp.Model;
 using Aspire.Dashboard.Otlp.Storage;
 using Google.Protobuf.Collections;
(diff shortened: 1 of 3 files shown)


tests/Aspire.Dashboard.Tests/TelemetryRepositoryTests/TraceTests.cs
```diff
@@ -488,4 +489,79 @@ public void AddTraces_AttributeAndEventLimits_LimitsApplied()
...
+
+        var testTime = s_testTime.AddDays(1);
+
+        // Act
+        for (var i = 0; i < 2000; i++)
+        {
+            var traceId = (i + 1).ToString(CultureInfo.InvariantCulture);
+
+            // Insert traces out of order to stress the circular buffer type.
+            var startTime = testTime.AddMinutes(i + (i % 2 == 0 ? -5 : 0));
```
radical, line 507: Should we also try to test for a case where the startTime is before the time of the first entry in the traces at any given time, and maybe after the last one?

radical (approved):

tests/Aspire.Dashboard.Tests/TelemetryRepositoryTests/TraceTests.cs
JamesNK: I looked at code coverage at least some tests hit that path.

Issue #4526, opened 2024-06-15
Configure Aspire Dashboard with key-per-file configuration provider


mu88:
Hi 👋🏻

**Disclaimer:** Since my issue is more a feature request than an API change, I decided not to use one of the predefined templates, but I'm happy to provide any further information you need.

I'm running the Aspire Dashboard as a standalone app via Docker which works nicely 👌🏻 this is my current `docker-compose.yml`:

```yaml
services:
  aspire-dashboard:
    image: mcr.microsoft.com/dotnet/aspire-dashboard
    container_name: aspire-dashboard
    ports:
      - 4317:18889
      - 18888:18888
    environment:
      - Dashboard__Frontend__BrowserToken=Tralala
```

You immediately see the flaw: it contains the browser token as cleartext. Especially since I put my `docker-compose.yml` files under version control, that's not really great.

For my other .NET apps, I'm using the [key-per-file configuration provider](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration-providers#key-per-file-configuration-provider) which would allow the following setting:

```yaml
services:
  aspire-dashboard:
    image: mcr.microsoft.com/dotnet/aspire-dashboard
    container_name: aspire-dashboard
    ports:
      - 4317:18889
      - 18888:18888
    secrets:
      - Dashboard__Frontend__BrowserToken
secrets:
  Dashboard__Frontend__BrowserToken:
    file: Dashboard__Frontend__BrowserToken.txt
```

This way, I could easily put my `docker-compose.yml` under version control.

After a quick and non-comprehensive look, I think this might be implemented by adding the following lines [here](https://github.com/dotnet/aspire/blob/main/src/Aspire.Dashboard/DashboardWebApplication.cs#L83):

```c#
if (!string.IsNullOrEmpty(builder.Configuration[DashboardConfigNames.SecretsPath]))
{
    builder.Configuration.AddKeyPerFile(directoryPath: builder.Configuration[DashboardConfigNames.SecretsPath], optional: false);
}
```

With this change, the final `docker-compose.yml` could look like this:

```yaml
services:
  aspire-dashboard:
    image: mcr.microsoft.com/dotnet/aspire-dashboard
    container_name: aspire-dashboard
    ports:
      - 4317:18889
      - 18888:18888
    environment:
      - Dashboard__SecretsPath=/run/secrets
    secrets:
      - Dashboard__Frontend__BrowserToken
secrets:
  Dashboard__Frontend__BrowserToken:
[...]

PR #4728, opened 2024-06-29
Configure Aspire Dashboard with key-per-file configuration provider (#4526)

mu88:
This PR adds the [.NET key-per-file configuration provider](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration-providers#key-per-file-configuration-provider) to the Aspire Dashboard (see issue https://github.com/dotnet/aspire/issues/4526).

Thereby, one can use Docker secrets like this:

```yaml
services:
  aspire-dashboard:
    image: mcr.microsoft.com/dotnet/aspire-dashboard
    container_name: aspire-dashboard
    ports:
      - 4317:18889
      - 18888:18888
    environment:
      - Dashboard__FileConfigDirectory=/run/secrets
    secrets:
      - Dashboard__Frontend__BrowserToken
secrets:
  Dashboard__Frontend__BrowserToken:
    file: Dashboard__Frontend__BrowserToken.txt
```

Another PR for updating [the docs](https://github.com/dotnet/docs-aspire/blob/main/docs/fundamentals/dashboard/configuration.md) will follow.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4728)

### DIFF
--- a/src/Shared/DashboardConfigNames.cs
+++ b/src/Shared/DashboardConfigNames.cs
@@ -12,6 +12,7 @@ internal static class DashboardConfigNames
     public static readonly ConfigName DashboardConfigFilePathName = new("DOTNET_DASHBOARD_CONFIG_FILE_PATH");
     public static readonly ConfigName ResourceServiceUrlName = new("DOTNET_RESOURCE_SERVICE_ENDPOINT_URL");
 
+    public static readonly ConfigName DashboardFileConfigDirectory = new("Dashboard:FileConfigDirectory", "DASHBOARD__FILECONFIGDIRECTORY");
     public static readonly ConfigName DashboardOtlpAuthModeName = new("Dashboard:Otlp:AuthMode", "DASHBOARD__OTLP__AUTHMODE");
     public static readonly ConfigName DashboardOtlpPrimaryApiKeyName = new("Dashboard:Otlp:PrimaryApiKey", "DASHBOARD__OTLP__PRIMARYAPIKEY");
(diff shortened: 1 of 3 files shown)


davidfowl:
@JamesNK @drewnoakes this change enables docker secrets and k8s config maps.

drewnoakes (approved):
LGTM. Thanks very much @mu88.

src/Shared/DashboardConfigNames.cs
```diff
@@ -12,6 +12,7 @@ internal static class DashboardConfigNames
     public static readonly ConfigName DashboardConfigFilePathName = new("DOTNET_DASHBOARD_CONFIG_FILE_PATH");
     public static readonly ConfigName ResourceServiceUrlName = new("DOTNET_RESOURCE_SERVICE_ENDPOINT_URL");
 
+    public static readonly ConfigName DashboardFileConfigDirectory = new("Dashboard:FileConfigDirectory", "DASHBOARD__FILECONFIGDIRECTORY");
```
JamesNK, line 15: This config name scheme (`Dashboard:xxx:yyy`) is for binding onto the dashboard options. That's not happening here so this scheme shouldn't be used. The name should be like `ASPNETCORE_URLS`, `DOTNET_DASHBOARD_CONFIG_FILE_PATH`, etc.

I think it should be `DOTNET_DASHBOARD_CONFIG_FILE_DIRECTORY`. It's close to the file path config config value, but we can't go back in time and add JSON to the existing config name so we'll have to live with it.

JamesNK: Issue: https://github.com/dotnet/aspire/issues/5760

PR #4744, opened 2024-07-02
Disabling key access for storage by default.

### LINKED ISSUE #4713: Azure Storage should disable keys by default
The default deployment of an Azure Storage account uses a ServiceUri connection mode and MSI to connect to the resource.

Aspire should disable key access, since it's not used, and also improves resource security.

aaronpowell:
Fixes #4713
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4744)


mitchdenny:
This is good ... thanks @aaronpowell. I think for some of these kinds of tweaks away from the default I'd like to set a test trap for the `Azure.Provisioning` APIs. Basically, I'd like a test case that uses the callback to set the value back to true.

I want to make sure that if someone for some reason wants to enable keys they are able to and that the Bicep generation doesn't break.

aaronpowell:
> This is good ... thanks @aaronpowell. I think for some of these kinds of tweaks away from the default I'd like to set a test trap for the `Azure.Provisioning` APIs. Basically, I'd like a test case that uses the callback to set the value back to true.
> 
> I want to make sure that if someone for some reason wants to enable keys they are able to and that the Bicep generation doesn't break.

Sure, I could add another test that will verify that if you want to override a default then it's applied as expected.

aaronpowell:
@mitchdenny want to trigger a re-run of the tests?

JamesNK:
The repo changes rapidly and this PR is a few weeks old. Would be a good idea to rebase on latest changes.

aaronpowell:
Good point @JamesNK 

I'll wait for this test run to complete and do a sync with `main`

mitchdenny (approved):
LGTM. Thanks @aaronpowell

PR #4745, opened 2024-07-02
Bump Polly to 8.4.1

martincostello:
- Bump Polly to [8.4.1](https://github.com/App-vNext/Polly/releases/tag/8.4.1) to pick up bug fixes:
  - https://github.com/App-vNext/Polly/pull/2164
  - https://github.com/App-vNext/Polly/pull/2166
- Group Polly dependabot updates to avoid multiple PRs such as:
  - #4167
  - #4200

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4745)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -53,6 +53,9 @@ updates:
       Grpc:
         patterns:
           - "Grpc.*"
+      Polly:
+        patterns:
+          - "Polly.*"
     labels:
       - "area-codeflow"
 
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -112,8 +112,8 @@
     <PackageVersion Include="Npgsql.DependencyInjection" Version="8.0.3" />
     <PackageVersion Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="8.0.4" />
     <PackageVersion Include="Oracle.EntityFrameworkCore" Version="8.23.40" />
-    <PackageVersion Include="Polly.Core" Version="8.4.0" />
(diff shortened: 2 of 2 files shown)


martincostello:
I think the packages need mirroring.

danmoseley:
I thought I did.. I'll try again

danmoseley (approved):

danmoseley:
I had mirrored Polly but not Polly.Extensions.

We really need an the internal mirror process to be easier, it makes me type in every package and version number, each one is a pipeline run unless I just want latest .. maybe @joperezr already has a script...

danmoseley:
for this reason, I'm inclined to mostly ignore random updates eg AWS patch versions and then catch up for our releases. whereas this Polly update, it's clear it brings improvements we need.

joperezr:
> I had mirrored Polly but not Polly.Extensions.
> 
> 
> 
> We really need an the internal mirror process to be easier, it makes me type in every package and version number, each one is a pipeline run unless I just want latest .. maybe @joperezr already has a script...

Technically speaking, you can likely get away with a single run most of the time if you keep version as "latest". Given the whole point of these updates is to mirror latest on NuGet, having "latest" on the pipeline means "latest for each package available on NuGet" which is most of the time what you want with these updates. That said, I think it is great feedback, and we should ask @dotnet/dnceng to see if they can allow passing in key-value pairs instead.

danmoseley:
cc @radical fyi ^^

danmoseley:
Looks like it takes 46 mins to do the whole lot.

PR #4759, opened 2024-07-02
Preserve the type of application builder in chained extension methods

sliekens:
This is a slight improvement to the usability of extensions in the ServiceDefaults project template. Previously, you couldn't chain `WebApplication.CreateBuilder` and the `ServiceDefaults` extensions.

For example, this code won't compile:

``` csharp
var builder = WebApplication.CreateBuilder(args)
    .AddServiceDefaults();

var app = builder.Build();
               // ^^^^^
               // CS1061: 'IHostApplicationBuilder' does not contain a definition for 'Build' and no accessible extension method 'Build' accepting a first argument of type 'IHostApplicationBuilder' could be found
```

After this change, compilation will succeed.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4759)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
+++ b/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Diagnostics.HealthChecks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.ServiceDiscovery;
 using OpenTelemetry;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
@@ -14,7 +15,7 @@ namespace Microsoft.Extensions.Hosting;
 // To learn more about using this project, see https://aka.ms/dotnet/aspire/service-defaults
 public static class Extensions
 {
-    public static IHostApplicationBuilder AddServiceDefaults(this IHostApplicationBuilder builder)
+    public static TBuilder AddServiceDefaults<TBuilder>(this TBuilder builder) where TBuilder : IHostApplicationBuilder
     {
         builder.ConfigureOpenTelemetry();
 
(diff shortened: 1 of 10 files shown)


eerhardt:
@mitchdenny @DamianEdwards @davidfowl - thoughts here?

If we take this change, we need to update ServiceDefaults everywhere - playground apps, all the templates, etc.

mitchdenny:
I don't have a problem with this change, but the PR will need to include the fixes to playground apps etc as @eerhardt has identified.

I would defer to folks who have had to deal with the reality of project template changes over time and what issues that can introduce though.

DamianEdwards:
This seems like something we should consider for 9.0

mitchdenny:
@sliekens did you want to add the playground app updates here too?

sliekens:
Yeah I realized after creating the PR that the ServiceDefaults template was not the only place to make changes, though I'm not sure how many other projects would need to be changed.

mitchdenny:
If you look under /playground you'll see that we generally share a single service defaults project there.

sliekens:
@mitchdenny I'm noticing other kinds of drift between the different copies of the Extensions.cs, is it okay if I re-align them while I'm making this change?

src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
```diff
@@ -31,10 +32,16 @@ public static IHostApplicationBuilder AddServiceDefaults(this IHostApplicationBu
             http.AddServiceDiscovery();
         });
 
+        // Uncomment the following to restrict the allowed schemes for service discovery.
+        // builder.Services.Configure<ServiceDiscoveryOptions>(options =>
+        // {
+        //     options.AllowedSchemes = ["https"];
+        // });
+
```
DamianEdwards, line 40: Hmm where did this come from? I'm not seeing it in the current templates. Did we have this and then remove it at some point?

DamianEdwards (changes requested):
I'm supportive of this change in the templates, but the remaining playground apps need to be updated and the builds errors regarding the unnecessary `using` directive need to be fixed.

src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/Extensions.cs
DamianEdwards: Ah got it, thanks. It's fine to add it to the other templates then, but we don't need it in the playground apps.

sliekens: OK, it's now removed from the playground apps

tests/TestingAppHost1/TestingAppHost1.ServiceDefaults/Extensions.cs
```diff
@@ -1,6 +1,3 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
DamianEdwards, line 1: You need to add this header back, it's causing a build failure.

DamianEdwards (changes requested):
Seems you inadvertently deleted the required header from one of the files.

tests/TestingAppHost1/TestingAppHost1.ServiceDefaults/Extensions.cs
sliekens: Oh man, I put it back and I'm gonna watch for the build to finish successfully this time 🤣.

DamianEdwards (approved):
Thank you!

PR #4772, opened 2024-07-03
Add WithDataVolume for AzureEventHubs emulator

sebastienros:
Context: https://github.com/dotnet/aspire/pull/4209/files#r1619294488

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4772)


mitchdenny:
Does it make sense to support read only data volumes for event hubs emulator?

sebastienros:
> Does it make sense to support read only data volumes for event hubs emulator?

@mitchdenny good point, so this would make sense for an emulated storage used directly, hence why we support it, but in the case of eventhubs there is no scenario for this?

mitchdenny:
I can't think of one. It's not like a writable snapshot

src/Aspire.Hosting.Azure.EventHubs/PublicAPI.Unshipped.txt
```diff
@@ -5,5 +5,8 @@ Aspire.Hosting.Azure.AzureEventHubsResource.IsEmulator.get -> bool
 override Aspire.Hosting.Azure.AzureEventHubsEmulatorResource.Annotations.get -> Aspire.Hosting.ApplicationModel.ResourceAnnotationCollection!
 override Aspire.Hosting.Azure.AzureEventHubsEmulatorResource.Name.get -> string!
 static Aspire.Hosting.AzureEventHubsExtensions.RunAsEmulator(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>! builder, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsEmulatorResource!>!>? configureContainer = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>!
-static Aspire.Hosting.AzureEventHubsExtensions.WithDataBindMount(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsEmulatorResource!>! builder, string? path = null, bool isReadOnly = false) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsEmulatorResource!>!
```
eerhardt, line 8: If we haven't shipped this API yet, I don't think we need to add all these Obsolete overloads.

tests/Aspire.Hosting.Tests/Azure/AzureEventHubsExtensionsTests.cs
```diff
@@ -0,0 +1,122 @@
...
+    public void AzureEventHubsUseEmulatorCallbackWithWithDataBindMountResultsInBindMountAnnotationWithDefaultPath()
+    {
+        using var builder = TestDistributedApplicationBuilder.Create();
+        var eventHubs = builder.AddAzureEventHubs("eh").RunAsEmulator(configureContainer: builder =>
+        {
+            builder.WithDataBindMount();
+        });
+
+        // Ignoring the annotation created for the custom Config.json file
+        var volumeAnnotation = eventHubs.Resource.Annotations.OfType<ContainerMountAnnotation>().Where(a => !a.Target.Contains("Config.json")).Single();
```
eerhardt, line 22: (nit) repeated elsewhere.

```suggestion
        var volumeAnnotation = eventHubs.Resource.Annotations.OfType<ContainerMountAnnotation>().Single(a => !a.Target.Contains("Config.json"));
```

src/Aspire.Hosting.Azure.EventHubs/PublicAPI.Unshipped.txt
sebastienros: Forgot about that. I feel stupid as I reminded it to the EH team this morning.

eerhardt (approved):
LGTM. Thanks!

PR #4774, opened 2024-07-03
DisableLocalAuth for AppConfig.

mitchdenny:
Following up on an issue raised by @mjrousos where we are leaving local auth enabled on the Azure AppConfig instances we create. This is unnecessary since we setup managed identity/service principal auth when we deploy.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4774)


eerhardt:
Are we going to do the same for Service Bus as well?

Are there other Azure services we should do this for?

mitchdenny:
> Are we going to do the same for Service Bus as well?

https://github.com/dotnet/aspire/pull/4938

> Are there other Azure services we should do this for?

Probably yes, fishing them out - but we shouldn't hold this.

eerhardt (approved):

PR #4782, opened 2024-07-03
Bump templates package references

sebastienros:
Was reviewing templates and checked versions. Should be ok to bump once in a while.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4782)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-mstest/Aspire.Tests.1.csproj
+++ b/src/Aspire.ProjectTemplates/templates/aspire-mstest/Aspire.Tests.1.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Aspire.Hosting.Testing" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
-    <PackageReference Include="MSTest" Version="3.1.1" />
+    <PackageReference Include="MSTest" Version="3.4.3" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


sebastienros:
@DamianEdwards do we have automated tests for these or should I check each of them locally?

DamianEdwards:
@sebastienros tests were added by @radical I believe?

DamianEdwards:
There's a reason the test packages aren't kept up to date apparently but adding @timheuer for context

timheuer:
@Evangelink for Aspire-specifically, do you think we should keep the test packages up-to-date? It doesn't seem to fall into the scenarios you noted why we don't update them for all?

Evangelink:
It seems fine for me to be on latest here. You will have to either do update PRs often or to update your infra (dependabot maybe) to have updates.

Evangelink:
> There's a reason the test packages aren't kept up to date apparently but adding @timheuer for context

Long story short, it was hard to get approval to bug fix SDK to bring new versions of NuGet packages. And the release cycle of ask doesn't match updates of framework packages. 

Lastly, it's not allowed to be using floating versions.

We are happy to revise our strategy if you have other recommendations or if we can get help with approvals.

DamianEdwards:
> Long story short, it was hard to get approval to bug fix SDK to bring new versions of NuGet packages. And the release cycle of ask doesn't match updates of framework packages.

Hmm that seems like something that should be revisited as shipping templates with out-of-date (i.e. unsupported) package references seems less than ideal. I thought other templates were getting their dependencies updated as part of SDK feature band releases but checking the [history for ASP.NET Core templates](https://github.com/dotnet/aspnetcore/tree/release/8.0/src/ProjectTemplates/Web.ProjectTemplates) suggests I'm wrong. Adding @mkArtakMSFT for confirmation on what we do for ASP.NET Core templates.

radical:
> @sebastienros tests were added by @radical I believe?

Yes - https://github.com/dotnet/aspire/blob/c76edc2159176d91b683ee63ead8fa641e34ebca/tests/Aspire.Workload.Tests/PerTestFrameworkTemplatesTests.cs#L39-L157 .

radical:
Fixed by https://github.com/dotnet/aspire/pull/4816 . Closing this.

danmoseley:
Yeah, my bad for not seeing this (actually we should probably have an owner of this regular chore). Although, there were more updates than in this PR

PR #4789, opened 2024-07-03
Display more info when local orchestrator check fails with unexpected error

### LINKED ISSUE #4691: Unhelpful error message if DCP dependency check fails
If the dependency check that uses `dcp info` command fails, with `dcp` returning non-zero code, an exception is thrown. The way we process this exception here https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs#L105 cause all output of the `dcp` command to be discarded.

The result is the user only sees a message that says "dcp info returned non-zero exit code", which is not helpful:

image

We should instead include all the output from `dcp` in the log, so that the user is informed about why the dependency check failed.

karolz-ms:
Fixes https://github.com/dotnet/aspire/issues/4691

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4789)

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
+++ b/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
@@ -57,8 +57,6 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
 
             try
             {
-                
-
                 var arguments = "info";
                 if (!string.IsNullOrEmpty(containerRuntime))
                 {
--- a/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
+++ b/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
@@ -126,7 +126,8 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
                     try
                     {
                         await processDisposable.DisposeAsync().ConfigureAwait(false);
-                    } catch { } // Dispose (dcp info process termination) is best effort.
+                    }
+                    catch { } // Dispose (dcp info process termination) is best effort.
                 }
             }
         }
--- a/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
+++ b/src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
@@ -87,10 +87,10 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
                 if (processResult.ExitCode != 0)
                 {
                     throw new DistributedApplicationException(string.Format(
-                    CultureInfo.InvariantCulture,
-                    Resources.DcpDependencyCheckFailedMessage,
-                    $"'dcp info{arguments}' returned exit code {processResult.ExitCode}. {errorStringBuilder.ToString()} {outputStringBuilder.ToString()}"
-                ));
+                        CultureInfo.InvariantCulture,
+                        Resources.DcpDependencyCheckFailedMessage,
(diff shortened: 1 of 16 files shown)


src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -52,34 +52,45 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
...
                 else
                 {
-                    await task.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    processResult = await task.WaitAsync(cancellationToken).ConfigureAwait(false);
+                }
+
+                if (processResult.ExitCode != 0)
+                {
+                    throw new DistributedApplicationException(string.Format(
+                    CultureInfo.InvariantCulture,
```
danegsta, line 90: nit: indentation is off

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -52,34 +52,45 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
...
+                    Resources.DcpDependencyCheckFailedMessage,
+                    $"'dcp info{arguments}' returned exit code {processResult.ExitCode}. {errorStringBuilder.ToString()} {outputStringBuilder.ToString()}"
```
danegsta, line 92: `arguments` already contains `info`, so this will double print.

danegsta: It may be worth separating the stderr and stdout strings with an Environment.NewLine

karolz-ms: Good catch

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
```diff
@@ -52,34 +52,45 @@ internal sealed partial class DcpDependencyCheck : IDcpDependencyCheckService
...
                 }
                 else
                 {
-                    await task.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    processResult = await task.WaitAsync(cancellationToken).ConfigureAwait(false);
+                }
+
+                if (processResult.ExitCode != 0)
+                {
+                    throw new DistributedApplicationException(string.Format(
```
danegsta, line 89: The actual exception from the initial issue was actually throwing from the call to `DisposeAsync` in the finally block. The cancellation token had fired, triggering a cancellation exception, but the finally block rethrew a separate exception due to the non-zero exit code from stopping the process.

I think we need to add a try/catch block around that `DisposeAsync` call and re-throw (or wrap and re-throw) the original exception if there is one.

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
karolz-ms: fixed

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
karolz-ms: fixed

src/Aspire.Hosting/Dcp/DcpDependencyCheck.cs
karolz-ms: Good point. Added a suppression of exceptions from the call to `DisposeAsync()` to ensure that original exception propagates.

danegsta (approved):

PR #4790, opened 2024-07-03
[ci] Enable source-index for the repo

radical:
This is to get enrolled into indexing for `https://source.dot.net`.

### DIFF
--- a/eng/pipelines/azure-pipelines.yml
+++ b/eng/pipelines/azure-pipelines.yml
@@ -122,7 +122,7 @@ extends:
           enablePublishBuildAssets: true
           enableTelemetry: true
           enableSourceBuild: true
-          enableSourceIndex: false
+          enableSourceIndex: true
           publishAssetsImmediately: true
           # Publish build logs
           enablePublishBuildArtifacts: true


eng/pipelines/azure-pipelines.yml
```diff
@@ -122,7 +122,7 @@ extends:
           enablePublishBuildAssets: true
           enableTelemetry: true
           enableSourceBuild: true
-          enableSourceIndex: false
+          enableSourceIndex: true
```
eerhardt, line 125: Does this need any conditions? Like just on windows, or just on official builds, just on `main`, etc?

eerhardt (approved):

eng/pipelines/azure-pipelines.yml
radical: Good question!

1. This file is for the internal pipeline. The public one is https://github.com/dotnet/aspire/blob/eb0bd80e0ef2e0b06820ee8fa953281167928447/eng/pipelines/azure-pipelines-public.yml#L92
2. This pipeline runs only the Windows build right now. There is a [linux build](https://github.com/dotnet/aspire/blob/eb0bd80e0ef2e0b06820ee8fa953281167928447/eng/pipelines/azure-pipelines.yml#L165-L193) in this same file conditioned on `_RunAsPublic`, which should be defaulting to `false` for internal builds - https://github.com/dotnet/aspire/blob/eb0bd80e0ef2e0b06820ee8fa953281167928447/eng/pipelines/common-variables.yml#L13-L15

radical: @RussKie why is there a linux job conditioned on `_RunAsPublic` in https://github.com/dotnet/aspire/blob/eb0bd80e0ef2e0b06820ee8fa953281167928447/eng/pipelines/azure-pipelines.yml#L165-L193 ? When is that run?

RussKie: The Ubuntu leg is only run on the public CI. I'm a little vague as to why it's not run on the internal CI (been many months since I asked that question) - I think it was something to do with lack of docker support... 
Things may have changed since, so it's possible this condition can be removed now.

radical: Ok, so no change needed in this PR.

PR #4795, opened 2024-07-04
ConfigSchemaGenerator: Ignore case and support named options

### LINKED ISSUE #4773: ConfigSchemaGenerator should ignore casing of configuration keys

### Describe the bug

When using ConfigSchemaGenerator, the casing of segments in the config path should be ignored. This is because configuration keys in .NET are case-insensitive.

For example, the following entries should be merged into a single subtree:
```c#
[assembly: ConfigurationSchema("Management:Endpoints:Health", typeof(HealthEndpointOptions))]
[assembly: ConfigurationSchema("management:endpoints:Actuator", typeof(HypermediaEndpointOptions))]
```

I would expect the difference in casing to be ignored, so that the JSON fragment below is produced:
```json
{
  "type": "object",
  "properties": {
    "Management": {
      "type": "object",
      "properties": {
        "Endpoints": {
          "type": "object",
          "properties": {
            "Actuator": {},
            "Health": {}
          }
        }
      }
    }
  }
}
```

The actual behavior is that no unification takes place:
```json
{
  "type": "object",
  "properties": {
    "Management": {
      "type": "object",
      "properties": {
        "Endpoints": {
          "type": "object",
          "properties": {
            "Health": {}
          }
        }
      }
    },
    "management": {
      "type": "object",
      "properties": {
        "endpoints": {
          "type": "object",
          "properties": {
            "Actuator": {}
          }
        }
      }
    }
  }
}
```

The same problem possibly applies to the class properties.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

This should be easy to fix, because `JsonObject` has a constructor that takes a `JsonNodeOptions`, which contains the boolean property `PropertyNameCaseInsensitive`.

### LINKED ISSUE #4775: Support named options in ConfigSchemaGenerator

It would be nice to be able to express named options in ConfigSchemaGenerator.

For example:
```c#
[assembly: ConfigurationSchema("Certificates:*", typeof(CertificateSettings))]

public class CertificateSettings
{
    /// <summary>
    /// The private key of the certificate, in base64 format.
    /// </summary>
    public string? PrivateKey { get; set; }
}
```

So that users get IntelliSense when adding the following named options to `appsettings.json`:
```json
{
  "Certificates": {
    "OrderService": {
      "PrivateKey": "..."
    },
    "BasketService": {
      "PrivateKey": "..."
    }
  }
}
```

This could be accomplished by producing the following JSON schema:
```json
{
  "type": "object",
  "properties": {
    "Certificates": {
      "type": "object",
      "patternProperties": {
        "^.+$": {
          "type": "object",
          "properties": {
            "PrivateKey": {
              "type": "string",
              "description": "The private key of the certificate, in base64 format."
            }
          }
        }
      }
    }
  }
}
```

image

### Additional context

I would recommend to keep it simple and treat the `*` as a marker value in a path segment, not a true wildcard. So while `"Certificates:*:Other:*"` would work, `"Certificates:O*ther"` would not.

The regex in `patternProperties` could be narrowed down to what's allowed for the name of a .NET property, although I don't know the exact rules that apply.

bart-vmware:
This PR updates the Config Schema Generator tool to handle casing differences in log categories, config paths, type names, and property names. There's no way to determine the "right" casing, ie: should "One" win over "one"? Therefore I've ~_only_~ made sure that the nodes get unified (so you'll either get "One" or "one", but never both), and that the last occurrence always wins, and capture this behavior in tests:

- ~`OverwritesLogCategoriesWithDifferentCasing`~ `LastUsedCasingOfLogCategoryWins`: Casing of the last occurrence wins.
- ~`MergesPropertiesAtConfigPathWithDifferentCasing`~ `LastUsedCasingOfConfigPathWins`: Casing of the ~first~ last occurrence wins.
- ~`MergesTypesWithDifferentCasing`~ `LastUsedTypeDocumentationWins`: ~Casing~ Documentation of the last occurrence wins.
- ~`OverwritesPropertyWithDifferentCasing`~ `LastUsedCasingOfPropertyWins`: Casing of the last occurrence wins.

Additionally, this PR implements support for named options, as proposed in #4775.

Fixes #4773.
Closes #4775.

/cc @eerhardt 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4795)

Edit: Renamed tests to clarify their purpose, and ensure the last occurrence always wins.


tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
```diff
@@ -997,7 +999,144 @@ public record TestSettings2
...
+                    }
+                  }
+                }
+              }
+            }
+            """);
+    }
+
+    [Fact]
+    public void MergesTypesWithDifferentCasing()
```
eerhardt, line 1058: I'm a bit confused by this test. The string `testSETTINGS` doesn't show up at all in the JSON. Should it?

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
```diff
@@ -1297,6 +1436,122 @@ public class TestSettings
             """);
     }
 
+    [Fact]
+    public void OverwritesLogCategoriesWithDifferentCasing()
```
eerhardt, line 1440: I think we should be consistent when we have conflicts. Either the first one should win, or the last one should win. I don't think we should have different behavior depending on where the conflict exists.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
bart-vmware: No, it shouldn't. The names of non-primitive types never show up in the schema. This was already the case and is the correct behavior.

What this PR changes is that the schema contains: `"description": "Replaced documentation."`. Without the changes in this PR, it results in: `"description": "Initial documentation."`. The test shows the last occurrence wins, which used to be the first occurrence (but there was no test for it).

As you indicated we should be consistent whether the first or last occurrence wins, I'm going to rename the tests to include this information. The new test names should make it easier to understand their purpose.

tests/ConfigurationSchemaGenerator.Tests/GeneratorTests.cs
bart-vmware: Ok, I've made the last one win in all cases and clarified the test names.
And updated the PR description to reflect these changes.

eerhardt (approved):
LGTM. Thanks again for the great contributions, @bart-vmware!

bart-vmware:
> LGTM. Thanks again for the great contributions, @bart-vmware!

@eerhardt Thanks for timely reviewing and providing good feedback.
Can you please restart the cibuild, so this gets merged?

PR #4803, opened 2024-07-07
Adds overload for AddSqlServerDbContext - Aspire.Microsoft.EntityFrameworkCore.SqlServer

Juulsn:
There was no overload for AddSqlServerDbContext in Aspire.Microsoft.EntityFrameworkCore.SqlServer where configureDbContextOptions has an IServiceProvider as delegate parameter.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4803)

### DIFF
--- a/src/Components/Aspire.Microsoft.EntityFrameworkCore.SqlServer/AspireSqlServerEFCoreSqlClientExtensions.cs
+++ b/src/Components/Aspire.Microsoft.EntityFrameworkCore.SqlServer/AspireSqlServerEFCoreSqlClientExtensions.cs
@@ -15,7 +15,7 @@ using OpenTelemetry.Trace;
 namespace Microsoft.Extensions.Hosting;
 
 /// <summary>
-/// Extension methods for configuring EntityFrameworkCore DbContext to Azure SQL, MS SQL server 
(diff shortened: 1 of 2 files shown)


DerJuulsn:
@dotnet-policy-service agree

eerhardt:
Thanks for the contribution, @DerJuulsn.

1. The build is currently failing. Can you fix it?
2. Check out https://github.com/dotnet/aspire/pull/4327, where we did something similar for Kafka. I think we may need more overloads.
3. If we are doing this for SqlServer, we should do it for the other EF libraries as well.
4. In EntityFramework.Core 9.0, we've fixed composability of DbContextOptions with https://github.com/dotnet/efcore/pull/32518. We may no longer need this as developers can compose the options after the call to `builder.AddSqlServerDbContext<TContext>();`.

cc @AndriySvyryd

AndriySvyryd:
The additional overload isn't present by design. See https://github.com/dotnet/aspire/pull/4115#issuecomment-2155620767.

DerJuulsn:
Just found out about the enrich methods and definitely do agree with the design choice.

Would be great if the documentation could be a bit improve in this subject :)

> In EntityFramework.Core 9.0, we've fixed composability of DbContextOptions

Awesome!

PR #4816, opened 2024-07-08
Update project template package versions to latest

### LINKED ISSUE #3975: Test templates aren't using latest stable test libraries
Out of the box NuGet shows updates available for the following libraries. We should bump to latest stable:

- [ ] Coverlet to 6.0.2
- [ ] xunit to 2.7.1
- [ ] xunit.runner.visualstudio to 2.5.8
- [ ] mstest 3.3.1
- [ ] nunit to 4.1.0
- [ ] nunit.analyzers to 4.2.0

danmoseley:
Fix #3975

Dependabot does not understand or update project templates, perhaps because it sees the directory.packages.props or uses the .sln. Nor does `dotnet outdated` work (at least not easily) because the projects cannot load successfully as they are (they pick up the checked in build props and targets, etc which are not intended for them)

So we need to remember to periodically update them by hand until we write an elaborate script to make `dotnet outdated` work.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4816)

### DIFF
--- a/src/Aspire.ProjectTemplates/templates/aspire-mstest/Aspire.Tests.1.csproj
+++ b/src/Aspire.ProjectTemplates/templates/aspire-mstest/Aspire.Tests.1.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Aspire.Hosting.Testing" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
-    <PackageReference Include="MSTest" Version="3.1.1" />
+    <PackageReference Include="MSTest" Version="3.4.3" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Aspire.ProjectTemplates/templates/aspire-nunit/Aspire.Tests.1.csproj
+++ b/src/Aspire.ProjectTemplates/templates/aspire-nunit/Aspire.Tests.1.csproj
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Aspire.Hosting.Testing" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
-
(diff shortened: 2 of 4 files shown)


radical:
Tests failing with `Unable to find package NUnit.Analyzers with version (>= 4.2.0)`

danmoseley:
Right, I assumed templates resolved from public nuget.org -- since that's what customers will do. But I guess tests restore them using our nuget.config that does not use that? It doesn't matter much, it means I have to go mirror those packages now.

radical:
> Right, I assumed templates resolved from public nuget.org -- since that's what customers will do. But I guess tests restore them using our nuget.config that does not use that? It doesn't matter much, it means I have to go mirror those packages now.

I triggered that, and checked for the other packages.

danmoseley:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2490779&view=results

danmoseley:
oh, whoops

radical:
This overlaps with https://github.com/dotnet/aspire/pull/4816 , but there are more version updates here.

radical (approved):

danmoseley:
@radical I believe Jose had created a "tasks we need to do to release an Aspire version" list? I don't know where it is, but "update template dependencies" should be on there.

radical:
> This overlaps with #4816 , but there are more version updates here.

Oops, I meant https://github.com/dotnet/aspire/pull/4782

danmoseley:
@balachir will your team validate each of these templates before we ship 8.1?

balachir:
@danmoseley yes, these tests are part of our tests for 8.1 and we'll make sure to validate them with these dependencies updated.

cc: @v-bennettyue @v-elenafeng

PR #4841, opened 2024-07-10
Add cgmanifest.json for templates dependencies

sebastienros:
This makes sure every dependency in the templates is recorded and takes part in CG alerting.

@joperezr This should be checked before any release. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4841)


src/Aspire.ProjectTemplates/cgmanifest.json
```diff
@@ -0,0 +1,141 @@
...
+                    "name": "Microsoft.Extensions.Http.Resilience",
+                    "version": "8.6.0"
+                }
+            }
+        },
+        {
+            "component": {
+                "type": "nuget",
+                "nuget": {
+                    "name": "Microsoft.Extensions.ServiceDiscovery",
```
eerhardt, line 63: This is built from dotnet/aspire. Should it be here?

sebastienros: This is used directly from the templates, so I think this should be added just because we know it's a direct dependency. Unless we say we already use it so we don't have to add it. IMO it doesn't hurt to keep it there.

https://github.com/dotnet/aspire/blob/main/src/Aspire.ProjectTemplates/templates/aspire-empty/AspireApplication.1.ServiceDefaults/AspireApplication.1.ServiceDefaults.csproj#L14

danmoseley: Having it here will flag if somehow the templates do not point at the latest version. (Presumably because our script has a problem?)

sebastienros: I am currently trying to update the files such that we don't have static versions in the project templates but use Directory.Packages.props

sebastienros: Couldn't find a good way to centrally manage it so only removing this dependency for now.
It's part of the solution to there is no reason to add it to cgmanifest, we don't set the version manually, it's automated on build so it will never be a different version.

src/Aspire.ProjectTemplates/cgmanifest.json
```diff
@@ -0,0 +1,132 @@
...
+                    "name": "OpenTelemetry.Instrumentation.Runtime",
+                    "version": "1.9.0"
+                }
+            }
+        },
+        {
+            "component": {
+                "type": "nuget",
+                "nuget": {
+                    "name": "Microsoft.Extensions.Http.Resilience",
```
danmoseley, line 54: why this one specifically when templates have these versioned the same way?
```
<PackageReference Include="Aspire.Hosting.AppHost" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
<PackageReference Include="Aspire.Hosting.Redis" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
<PackageReference Include="Aspire.Hosting.Testing" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
<PackageReference Include="Aspire.StackExchange.Redis.OutputCaching" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
<PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="!!REPLACE_WITH_EXTENSIONS_VERSION!!" />
<PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
```

danmoseley (approved):
aside from comment, lgtm.

what will prevent this getting out of date with respect to templates, I assume eyeballs..

danmoseley:
@sebastienros do we have any external dependencies checked in in code form? Similar to how dotnet/runtime has zlib and maybe libunwind checked in and cgmanifest for them: https://github.com/dotnet/runtime/blob/a42b063686e9e4c263e6a773b5836a594a62343b/src/native/external/cgmanifest.json

danmoseley:
> what will prevent this getting out of date with respect to templates, I assume eyeballs..

Oh, discussed above, presumably eyeballs now and perhaps eventually a script.

src/Aspire.ProjectTemplates/cgmanifest.json
danmoseley: oh I guess you just haven't pulled it yet per above

sebastienros: `Microsoft.Extensions.Http.Resilience` is using `REPLACE_WITH_EXTENSIONS_VERSION`, not `REPLACE_WITH_LATEST_VERSION` which represents locally built packages that CG would pick as a standard artifact.

sebastienros:
> eventually a script

The goal is to have it driven by `Directory.Packages.props`, will create an issue when the file is merged.

danmoseley:
@sebastienros would you mind checking whether we need one of these for dotnet/extensions and/or dotnet/aspire-samples and any others that are active (and prod?) in https://github.com/dotnet/aspnetcore-internal/blob/main/docs/engineering/ASP.NET-repos.md like aspnet/samples, signalr/signalr.

src/Aspire.ProjectTemplates/cgmanifest.json
danmoseley: Ah - thanks - my eyeballs missed that.

PR #4846, opened 2024-07-10
Add DotnetWatchAnnotation to configure dotnet watch startup command

Juulsn:
Added a new annotation which can be used to configure the `dotnet watch` command, mainly the -no-hot-reload argument.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4846)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -1046,7 +1046,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                         projectMetadata.ProjectPath
                     ];
 
-                    if (dwa.EnableHotReload)
+                    if (!dwa.EnableHotReload)
                     {
                         exeSpec.Args.Add("--no-hot-reload");
                     }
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1036,23 +1036,29 @@ private void PrepareProjectExecutables()
...
+                        "watch",
+                        "--non-interactive",
                         "--project",
-                        projectMetadata.ProjectPath,
+                        projectMetadata.ProjectPath
                     ];
+
+                    if (dwa.EnableHotReload)
+                    {
+                        exeSpec.Args.Add("--no-hot-reload");
```
eerhardt, line 1051: This logic seems conflicting. "If enable hot reload" => "--no-hot-reload".

Do we need to rename the property?

mitchdenny:
I think it would be good to open an issue for this PR to discuss the API change.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
DerJuulsn: should be if not enabled -> ...

PR #4857, opened 2024-07-11
Load plotly as a module on metrics page

### LINKED ISSUE #4155: See whether plotly JavaScript can be imported on metrics page
Ploty JavaScript file is 3.5MB and is referenced from the root of the dashboard app, and so it must be downloaded before the dashboard page works: https://github.com/dotnet/aspire/blob/68e36e42ec109f7c2aeba48b74b1987547effac0/src/Aspire.Dashboard/Components/App.razor#L32

We might be able to make the plotly file dynamically imported on the metrics page. The goal is to make the file only downloaded if plotly is used.

Related:
* https://github.com/dotnet/aspire/issues/4149
* https://github.com/dotnet/aspire/issues/4150

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/4155
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4857)


mitchdenny (approved):

src/Aspire.Dashboard/Components/App.razor
```diff
@@ -28,7 +28,5 @@
     <Routes @rendermode="@(new InteractiveServerRenderMode(prerender: false))" />
     <script src="_framework/blazor.web.js"></script>
     <script src="js/app.js"></script>
-    <script src="js/theme.js" type="module"></script>
```
tlmii, line 31: Does removing this (presumably leaving it to the module load in OnAfterRenderAsync in MainLayout) cause any delays in theme switching? It feels like there could be a fairly substantial timing delay between when the two loads happen. But maybe it doesn't make a difference.

JamesNK: I didn't think of that. Good point. I'll measure.

If I re-add it I'll put a comment explaining why.

PR #4869, opened 2024-07-12
Escape braces in ReferenceExpression.AppendFormatted()

### LINKED ISSUE #3787: Lliterals with {} break ReferenceExpression

davidebbo:
This is an alternative PR to address #3787 in a much simpler way. I also think that the existing PR has [correctness issues](https://github.com/dotnet/aspire/pull/4308/files#r1672288949), though it's possible that I'm misinterpreting the intended behavior.

So the first step is to **agree on the intended behavior**. Specifically, if you have:

```c#
var s = "{{x}}";
var expr = ReferenceExpression.Create($"{s}").ValueExpression;
```

should you get `"{{x}}"` (which this PR does), or `"{x}"` (other PR). In my view, the `"{{x}}"` input is a plain literal that is not meant to be interpreted and should be preserved verbatim. i.e. it is not a format string.

I could not find any existing tests for `ReferenceExpression`, but if we agree that this is a good direction, I will add some.

Fixes #3787

/cc @davidfowl
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4869)


src/Aspire.Hosting/ApplicationModel/ReferenceExpression.cs
```diff
@@ -112,7 +112,9 @@ public static ReferenceExpression Create(in ExpressionInterpolatedStringHandler
         /// <param name="value">The formatted string to be appended to the interpolated string.</param>
         public readonly void AppendFormatted(string? value)
         {
-            _builder.Append(value);
+            // The value that comes in is a literal string that is not mean to be interpreted.
```
danmoseley, line 115:
```suggestion
            // The value that comes in is a literal string that is not meant to be interpreted.
```

mitchdenny:
Is there a way we can add a test to cover this?

davidebbo:
@mitchdenny yes, my initial PR comment touches on this. In short, I first want to get an agreement on the expected behavior, because there is a disagreement between my understanding, and what the earlier PR (https://github.com/dotnet/aspire/pull/4308) does.

davidebbo:
See my comments in https://github.com/dotnet/aspire/pull/4308/files#diff-0dc92a853b063a53c061dabbef5a2b330f21cd2d374c7270c378580e1a424fc0 for the list of test cases in the PR that I believe to be incorrect.

mitchdenny:
@davidebbo TBH we didn't think too much about some of these corner cases when dealing with this. But my expectation after looking at the APIs just now is that if someone passes in an argument with the string `{{x}}`` I would expect that to be interpretted as the literal string `{x}` - that is the double `{{` escape `{`.

mitchdenny:
The main thing is that if we formalize this, we need test coverage to make sure we don't go back on the decision :)

davidebbo:
@jeffl8n has some nice tests [here](https://github.com/dotnet/aspire/pull/4308/files#diff-0dc92a853b063a53c061dabbef5a2b330f21cd2d374c7270c378580e1a424fc0) in his PR, so with his authorization, we could just use that (with some tweaks). Or potentially have him submit the test a a separate PR so he gets credit for the code?

For the example above, I view

```c#
var s = "{{x}}";
var expr = ReferenceExpression.Create($"{s}").ValueExpression;
```

As very similar to:

```c#
var s = "{{x}}";
Console.WriteLine($"{s}");
```

which prints ``{{x}}``, not `{x}`. Why would this be any different? `s` is just a literal string in both cases, and not a format string.

mitchdenny:
> @jeffl8n has some nice tests [here](https://github.com/dotnet/aspire/pull/4308/files#diff-0dc92a853b063a53c061dabbef5a2b330f21cd2d374c7270c378580e1a424fc0) in his PR, so with his authorization, we could just use that (with some tweaks). Or potentially have him submit the test a a separate PR so he gets credit for the code?
> 
> For the example above, I view
> 
> ```cs
> var s = "{{x}}";
> var expr = ReferenceExpression.Create($"{s}").ValueExpression;
> ```
> 
> As very similar to:
> 
> ```cs
> var s = "{{x}}";
> Console.WriteLine($"{s}");
> ```
> 
> which prints `{{x}}`, not `{x}`. Why would this be any different? `s` is just a literal string in both cases, and not a format string.

Ah I've made a mistake. I didn't note that lack of string interpolation of the `s` variable assignment.

davidebbo:
As a middle ground, I added tests that specially target the code path that I'm modifying, which makes this PR complete.

@jeffl8n has additional ReferenceExpression tests for other scenarios, which he should contribute as a separate PR.

mitchdenny (approved):

davidebbo:
@mitchdenny I think this PR should be mergeable now. Thanks!

PR #4882, opened 2024-07-13
Add en-GB to supported Dashboard cultures

martincostello:
Support `en-GB` in the dashboard so dates and times are formatted appropriately for users in the UK.

I haven't added a test as I wasn't sure where to add such a test - if you'd like one, just point me in the right direction.

### Before

before

```html
<fluent-data-grid-cell cell-id="c12" cell-type="default" grid-column="4" class="col-justify-start " style="height: 46px; align-content: center; grid-column: 4;" title="7/13/2024 11:16:29 AM" aria-label="7/13/2024 11:16:29 AM" b-w6qdxfylwy="" tabindex="-1" role="gridcell">
  <!--!-->
  11:16:29 AM
</fluent-data-grid-cell>
```

### After

after

```html
<fluent-data-grid-cell cell-id="c133" cell-type="default" grid-column="4" class="col-justify-start " style="height: 46px; align-content: center; grid-column: 4;" title="13/07/2024 11:47:05" aria-label="13/07/2024 11:47:05" b-w6qdxfylwy="" tabindex="-1" role="gridcell">
  <!--!-->
  11:47:05
</fluent-data-grid-cell>
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4882)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -188,7 +188,8 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         var supportedLanguages = new[]
         {
             "en", "cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-BR", "ru", "tr", "zh-Hans", "zh-Hant", // Standard cultures for compliance.
-            "zh-CN" // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
+            "zh-CN", // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
+            "en-GB", // Support UK DateTime formatting (24-hour clock, dd/MM/yyyy)
         };
 
         _app.UseRequestLocalization(new RequestLocalizationOptions()


mitchdenny (approved):

mitchdenny:
As an Australian I fully endorse this change :)

danmoseley:
I was curious to compare to the ones eg that the dotnet homepage supports: "en-US","es-ES","fr-FR","id-ID","it-IT","ja-JP","ko-KR","pt-BR","ru-RU","zh-CN","zh-TW". They have Indonesian, eg. I expect that is based on where they get traffic from, so perhaps rather different to the dashboard.

JamesNK:
How far do we go with this? In dates the month/day is around the wrong way for en-AU, en-NZ, en-SG, etc.

I wonder if there is configuration to automatic say use the more specific culture settings for formatting, and then use the approximate translations, e.g. en-GB, en-US, etc all use the en translations.

Who is the expert on request localization middleware that we can ask?

martincostello:
> How far do we go with this?

I opened #4884 to ask that question.

samsp-msft:
AFAIK this can be handled by Blazor based on headers sent by the browser on what their locales are. Then when you use the date formatting functions, it should pick up the locale based on Blazor interpreting the headers and setting the context.

https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-8.0

> Small nit - this is globalization not localization.

Issue #2281, opened 2024-02-17
Row background highlighting on mouse hover


malisancube:
May you add a highlight on mouse hover for the grids. Its a bit dificult to focus on the row. More like the list we have here on github for filling issues.

tlmii:
Status update here: 

Currently there are no color tokens for `-hover` for the `--neutral-layer-*` backgrounds that we use for the layers in the site. The included hover state (`--neutral-fill-hover`) works ok for the main grids in dark mode, barely for the main grids and details grids in light mode and not at all for the details panel grids in dark mode. 

Working with the fluentui folks and @vnbaaij to see how easily we can create new recipes based on our layers and use them. Otherwise I'll find some values that work and hardcode them.

tlmii:
The changes needed to be able to create our own design tokens are working their way through, starting here: https://github.com/microsoft/fluentui/pull/30784. Will need to be brought into the blazor components first then we can start using it.

kvenkatrajan:
Moving to 8.1. @tlmii  had a look might take some new javascript work to make it work and additional testing - probably safer to move it to after GA.

Issue #4834, opened 2024-07-10
Improve CheckDashboardHasResourcesAsync in workload tests


JamesNK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

`CheckDashboardHasResourcesAsync` has some improvements I'd like to see to make it less brittle to changes and improve the error message output.

### Expected Behavior

* Asserts are hardcoded to cell position and number. The test will break if the number or order of columns changes. I'd like the test to get cells using something other than position. Perhaps we can give each cell a class to look up. Or if that isn't possible, then embed content in divs with classes. I think it's ok to make some modest changes to the generated HTML to make testing better.
* If a row's content doesn't match the expected content, then the test doesn't match the row and skips adding it to found rows. But that means the error message just says that the resource wasn't found. It doesn't say exactly what didn't match for the row (e.g. the wrong number of endpoints). Knowing why it didn't match would be useful.

JamesNK:
cc @radical

radical:
@adamint would you be interested in taking this up, since you recently worked on dashboard playwright tests?

adamint:
@radical sure!

JamesNK:
Also, queries such as `//fluent-data-grid-row[@class='resource-row']` require the exact classes to match the query. It would be better if the query returned elements named `fluent-data-grid-row` with the class `resource-row` and any other classes.

Is there a better way to write this kind of query that isn't so brittle to HTML changes?

See https://github.com/dotnet/aspire/pull/4892#discussion_r1680435584

adamint:
> Also, queries such as `//fluent-data-grid-row[@class='resource-row']` require the exact classes to match the query. It would be better if the query returned elements named `fluent-data-grid-row` with the class `resource-row` and any other classes.
> 
> Is there a better way to write this kind of query that isn't so brittle to HTML changes?
> 
> See [#4892 (comment)](https://github.com/dotnet/aspire/pull/4892#discussion_r1680435584)

Yes, it is recommended not to use css classes or html elements where possible

PR #4892, opened 2024-07-15
Create hover tokens for grid row hover effect

tlmii:
Finally got around to taking a stab at #2281 to create hover tokens for the neutral-layer-1/neutral-layer-2 colors we use as our primary background colors from the current styles provided by the designers. Those tokens are not automatically provided by Fluent so we need to create them ourselves.

- Creates hover tokens for neutral-layer-1/neutral-layer-2 that we use as primary background colors for the current styles provided by the designers. Those tokens aren't automatically provided by Fluent like other -hover variants so we need to create them ourselves
- The above tokens are created as generically as possible so that we can tweak them as we want. E.g. if we decide that the layer 2 light color is too light/dark, we can change the `neutralLayer2HoverLightDelta` up/down by 1 to move it to the next value in the color palette.
- Ties the new tokens into the `FluentDataGrid`'s `ShowHover` functionality.
- Tweaks some of the hover effects on buttons in the grids to make them make a little more sense when the row is hovered

grid-hover-colors

I gave a look at all of the pages as I went through this, but I'm sure I missed some things that don't look quite right with this change. Will try to give it another pass tomorrow, but wanted to get the PR out for others eyes as well.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4892)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
+++ b/src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
@@ -59,7 +59,7 @@
                         </div>
                     </DetailsTitleTemplate>
                     <Summary>
-                        <FluentDataGrid Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" Class="trace-view-grid" ResizableColumns="true" ItemsProvider="@GetData" TGridItem="SpanWaterfallViewModel" RowClass="@GetRowClass" GridTemplateColumns="4fr 12fr 85px">
+                        <FluentDataGrid Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" Class="trace-view-grid" ResizableColumns="true" ItemsProvider="@GetData" TGridItem="SpanWaterfallViewModel" RowClass="@GetRowClass" GridTemplateColumns="4fr 12fr 85px" ShowHover="true">
                             <TemplateColumn Title="Name">
                                 <div class="col-long-content" title="@context.GetTooltip(_applications)" @onclick="() => OnShowPropertiesAsync(context, null)">
                                     @{
--- a/src/Aspire.Dashboard/wwwroot/js/app-theme.js
+++ b/src/Aspire.Dashboard/wwwroot/js/app-theme.js
@@ -3,7 +3,10 @@ import {
     baseLayerLuminance,
     SwatchRGB,
     fillColor,
-    neutralLayerL2
+    neutralLayerL2,
+    neutralPalette,
+    DesignToken,
+    neutralFillLayerRestDelta
 } from "/_content/Microsoft.FluentUI.AspNetCore.Components/Microsoft.FluentUI.AspNetCore.Components.lib.module.js";
 
 const currentThemeCookieName = "currentTheme";
@@ -12,6 +15,7 @@ const themeSettingDark = "Dark";
 const themeSettingLight = "Light";
 const darkThemeLuminance = 0.19;
 const lightThemeLuminance = 1.0;
+const darknessLuminanceTarget = (-0.1 + Math.sqrt(0.21)) / 2;
 
 /**
  * Updates the current theme on the site based on the specified theme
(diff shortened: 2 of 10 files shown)


src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
```diff
@@ -53,7 +53,7 @@
                 </div>
             </DetailsTitleTemplate>
             <Summary>
-                <FluentDataGrid Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" Class="trace-view-grid" ResizableColumns="true" ItemsProvider="@GetData" TGridItem="SpanWaterfallViewModel" RowClass="@GetRowClass" GridTemplateColumns="4fr 12fr 85px">
+                <FluentDataGrid Virtualize="true" GenerateHeader="GenerateHeaderOption.Sticky" Class="trace-view-grid" ResizableColumns="true" ItemsProvider="@GetData" TGridItem="SpanWaterfallViewModel" RowClass="@GetRowClass" GridTemplateColumns="4fr 12fr 85px" ShowHover="true">
```
JamesNK, line 56: The trace detail page already had some CSS for highlighting the row. Should it be removed?

src/Aspire.Dashboard/wwwroot/js/app-theme.js
```diff
@@ -160,11 +164,91 @@ function applyTheme(theme) {
...
+/**
+ * 
+ * @param {Swatch} color
+ * @returns {boolean}
+ */
+function isDark(color) {
+    return color.relativeLuminance <= darknessLuminanceTarget;
+}
+
+function createAdditionalDesignTokens() {
```
JamesNK, line 204: Add a comment about why new design tokens are needed

JamesNK (commented):
I played around with it briefly and it seems to work in the places I tested.

Visually it looks good. The main thing I don't like is the cursor has changed to a pointer. I think it's confusing to have that cursor when there is nothing to click.

The simple solution is to revert the cursor change, but I'm still interested in making rows clickable. That would be a nice improvement. But probably not in the scope of this PR.

src/Aspire.Dashboard/wwwroot/js/app-theme.js
tlmii: Added

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor
tlmii: I removed the row highlighting line. There was some code to make the text stand out more when the row was highlighted - I left that part in place.

tlmii:
> The main thing I don't like is the cursor has changed to a pointer. I think it's confusing to have that cursor when there is nothing to click.
> 
> The simple solution is to revert the cursor change, but I'm still interested in making rows clickable. That would be a nice improvement. But probably not in the scope of this PR.

@vnbaaij Have you gotten any other feedback about this? It does seem like the cursor change should be tied to the row being selectable, not just to the hover effect. 

@JamesNK For now I overrode it with a more specific selector to put it back to `auto`.

vnbaaij:
I haven't seen any remarks on that. But think it makes sense to only have a special cursor when needed.

@JamesNK we have the OnRowClick event since a couple of versions ago

JamesNK:
> I haven't seen any remarks on that. But think it makes sense to only have a special cursor when needed.

+1

> @JamesNK we have the OnRowClick event since a couple of versions ago

Great! We should update Aspire to use it.

JamesNK:
@tlmii I tested it out and it looks good except the change to remove the pointer cursor also removed it from the trace detail page where we want to keep it.

JamesNK:
@tlmii I think the template tests need to be updated. They examine HTML so it's likely hover is changing the HTML output.

JamesNK (approved):

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
```diff
@@ -61,7 +61,7 @@ protected static async Task<ResourceRow[]> CheckDashboardHasResourcesAsync(IPage
             await Task.Delay(500);
 
             // _testOutput.WriteLine($"Checking for rows again");
-            ILocator rowsLocator = dashboardPage.Locator("//fluent-data-grid-row[@class='resource-row']");
+            ILocator rowsLocator = dashboardPage.Locator("//fluent-data-grid-row[@class='hover resource-row']");
```
JamesNK, line 64: Hmm, I don't like that this requires the exact right classes be specified. Unrelated to this change though.

JamesNK (approved):

tests/Aspire.Workload.Tests/WorkloadTestsBase.cs
tlmii: Yeah it seems a little brittle. Maybe just switching on the row type instead might be better (to skip header rows)? Even that seems a little brittle. But I didn't even realize this existed until you pointed it out, so I wasn't trying to delve into it too far yet, just trying to unblock.

JamesNK: I was hit by it last week. I vented here: https://github.com/dotnet/aspire/issues/4834

I'll add class name brittleness to the list

PR #3103, opened 2024-03-22
Change how resources expose urls to the dashboard

davidfowl:
- The resource server used to have a services and endpoints split, that is now a single object called a url which has a name, a full url (uri format) and a bool for if that endpoint is internal (or an implementation detail). Internal urls only show up in the details pane.
- Clean up the way we extract urls from the dcp information by mapping the app model resource and endpoint within each service resource in DCP.
- Updated tests

Fixes https://github.com/dotnet/aspire/issues/3078

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/3103)

PR #4908, opened 2024-07-15
Identify replica sets based on owner info provided by DCP, group applications by their replica set in resource dropdowns, add tests for console logs application resource grouping

### LINKED ISSUE #1233: Identify replica sets for dashboard
We don't have proper display for replica set yet (see #567), but as mentioned in #1229, we need to know sometimes how many replicas a resource have. Having a proper mechanism to identify would be helpful.

### LINKED ISSUE #1411: Unit Tests for Resource Name display text
We've got a few pieces of code that modify the resource name display text:

1. `DcpDataSource.ComputeExecutableDisplayName`, which sometimes removes the replica id if present
2. `ResourceFormatter.GetName`, which adds a uid
3. `ResourceViewModel.GetResourceName`, which calls ResourceFormatter.GetName under some circumstances
4. `OtlpApplication.GetResourceName`, which calls ResourceFormatter.GetName under some circumstances

We should consolidate this code into a common location as appropriate, make it testable (currently some of it is private methods) and write unit tests for it to ensure we're covering the various naming nuances. We already hit one with #1408, I'm sure we'll hit more.

### LINKED ISSUE #4103: Different instances of same resource type grouped as replicas in dashboard
In 8.0.0-preview.6.24214.1 and 8.0.0-preview.7.24251.11, different instances of project resources are grouped as replicas, even if they are not replicas.

This happens for:

- Structured logs
- Traces
- Metrics

But not for:

- Console

adamint:
Currently, we are grouping applications based on whether their display names are identical, instead of using the `ExecutableReplicaSet` property from DCP. 

The approach I took was
1. Flow owner information, which includes the above property, from DCP through to the dashboard frontend
2. Modify the 'application grouping' view model creation logic to use the real replica set for a resource
3. Replace `(replica set)` with `(application)` on traces/logs resource dropdowns, because the OTLP applications there can be grouped, but are not replica sets. 
image
4. Add a test around behavior for console logs view model creation (#1411)
 
Fixes #1411 
Fixes #4103
Fixes #1233
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4908)


adamint:
@JamesNK Could you advise on what to do about `ResourceOutgoingPeerResolverTests.NameAndDisplayNameDifferent_MultipleInstances_ReturnName` 
- does the previous behavior of `ResourceViewModel.GetResourceName` need to stay the same for containers?

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -166,34 +166,43 @@ protected override void OnAfterRender(bool firstRender)
...
-            .GroupBy(r => r.Value.DisplayName, r => r.Value))
+            .OrderBy(c => c.Value.Name, StringComparers.ResourceName)
+            .GroupBy(r => r.Value.GetReplicaSetOrDefault()?.Uid ?? r.Value.Uid))
         {
-            if (resourceGroupsByApplicationName.Count() > 1)
+            string applicationName;
+
+            if (grouping.Count() > 1)
             {
+                applicationName = grouping.First().Value.GetReplicaSetOrDefault()!.Name;
```
JamesNK, line 185: What is the situation where the replica set name is different from DisplayName?

src/Aspire.Dashboard/Otlp/Model/OtlpApplicationType.cs
```diff
@@ -5,7 +5,7 @@ namespace Aspire.Dashboard.Otlp.Model;
 
 public enum OtlpApplicationType
 {
-    ReplicaSet,
+    ApplicationGrouping,
     Singleton,
     ReplicaInstance
```
JamesNK, line 10: ReplicaInstance should be named to just Instance?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -666,10 +668,19 @@ private CustomResourceSnapshot ToSnapshot(Executable executable, CustomResourceS
...
         };
     }
 
+    private static ImmutableArray<OwnerReferenceSnapshot> GetOwners(CustomResource resource)
+    {
+        return [
+            ..resource.Metadata.OwnerReferences
+                .Select(ownerReference => new OwnerReferenceSnapshot(ownerReference.Kind, ownerReference.Name, ownerReference.Uid))
+        ];
+    }
```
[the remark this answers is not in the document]
karolz-ms: This is a K8s API concept and probably not the best name: it is a list of objects that depend on the given object. https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#System Most common scenario is when there is some sort of N:M relationship that is modeled as a separate object, the motivation for multiple owners being keeping the model consistent and cascade deletion.

For example (we are not doing this today, but we might in future): an Endpoint is exposed on a Network, implements a Service, and is implemented by an instance of a Container or Executable. The Network, Service, and implementing Container/Executable could all be "owner" objects.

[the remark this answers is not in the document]
adamint: @karolz-ms - understandable, do you have a better name in mind? Maybe Links?

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: For context, I was told (a while ago) that we should not assume that replica sets can be derived from the resource name and that we should at some point use info from dcp. Wondering if that’s changed, would love Karol’s thoughts

src/Aspire.Dashboard/Components/Controls/ResourceSelectOptionTemplate.razor
```diff
@@ -6,12 +6,14 @@
...
 <FluentStack Orientation="Orientation.Horizontal" HorizontalGap="3">
-    @if (ViewModel.Id?.Type is OtlpApplicationType.ReplicaSet)
+    @if (ViewModel.Id is { } typeDetails && typeDetails.Type is OtlpApplicationType.ApplicationGrouping)
     {
         <FluentIcon Icon="Icons.Regular.Size20.AppsList"
                     Color="Color.Accent"
                     Style="vertical-align: text-bottom;"/>
-        <span>@ViewModel.Name (<span style="color: var(--accent-foreground-rest)">@Loc[nameof(ControlsStrings.ReplicaSetLower)]</span>)</span>
+
+        var typeName = typeDetails.ReplicaSetName is not null ? Loc[nameof(ControlsStrings.ReplicaSetLower)] : Loc[nameof(ControlsStrings.ApplicationLower)];
```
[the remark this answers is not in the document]
adamint: changed to just say application

src/Aspire.Dashboard/Components/Controls/ResourceSelectOptionTemplate.razor
```diff
@@ -6,12 +6,14 @@
...
+        <span>@ViewModel.Name (<span style="color: var(--accent-foreground-rest)">@typeName</span>)</span>
```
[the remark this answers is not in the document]
adamint: changed

src/Aspire.Dashboard/Model/KnownOwnerProperties.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Dashboard.Model;
+
+public static class KnownOwnerProperties
```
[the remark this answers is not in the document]
adamint: Changed to internal

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
karolz-ms: We did in fact make a change (recently, https://github.com/microsoft/usvc-apiserver/pull/410) that made it official that names of the Executables belonging to a replica set are derived from the replica set by applying a unique suffix. 

As for `DisplayName`, I could not find any place in Aspire code that sets this property other than https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Dashboard/DashboardServiceData.cs#L39  If I am right, then DisplayName and resource name are the same.

FWIIW, inside DCP we do generate a "replica display name" as an annotation on an Executable. It is built off of replica set name and monotonically increasing counter. At some point we thought that might be something the dashboard could leverage, but it never happened.

adamint: > We did in fact make a change (recently, https://github.com/microsoft/usvc-apiserver/pull/410) that made it official that names of the Executables belonging to a replica set are derived from the replica set by applying a unique suffix.

So, the owner information part can be removed then? 🙂

karolz-ms: You do not have to scan the owner objects (although there is no harm in that), but please do not parse the name of the Executable either. Instead, if you want to know the replica set name, there is now an annotation for that `executable-replica-set.usvc-dev.developer.microsoft.com/replica-set-name` This will be put on every Executable launched by a replica set.

Caveat though: this is added by DCP 0.5.6 or newer; DCP 0.2.3 (Aspire 8.0 version) does not have this. So scanning owners may be your best bet, at least short-term, if backward compat is a concern

adamint: In that case, it might be better to scan owners? And then if there are additional owners we can show in the future on the dashboard, it will be easier?

karolz-ms: That would be my recommendation, yes.

JamesNK: I still don't understand what bug is actually being fixed here. What is broken today that this fixes? How do I repo what is broken today so I can see it for myself?

adamint: There is no broken behavior, but per @karolz-ms's comment above,
> please do not parse the name of the Executable either

is what we're doing already with the `DisplayName` https://github.com/dotnet/aspire/pull/4908/files#diff-e2cbd8471a0c18b0125d89b07d3d720c283699131220c4f573967e138f7ff4cb

The main purpose of this PR is to add unit tests for this behavior (#1411), but I took the opportunity to change `GetResourceName` to use the actual replica set information.

JamesNK: We're not parsing the name. The linked change doesn't show anything that parses the name.

As far as I can tell the DisplayName we have on the resource is already the replica set name. This seems like more complicated way of getting the same result we already have.

JamesNK (changes requested):
I'm fine with the changes in this PR except everything to do with getting and returning Owner info to the dashboard. It doesn't fix anything, it makes things more complicated, and it adds fields to the proto contract that we should think careful about before changing.

Think Owner changes should be removed from this PR.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: @karolz-ms could you weigh in as well with your thoughts?

karolz-ms: Can one of you point me to the current implementation that sets the `DisplayName` property on Executables that belong to the replica set? The one that sets it to replica set name?

JamesNK: I believe it comes from the `resource-name` annotation.

image

karolz-ms: If that is the case, the DisplayName is not necessarily the name of the Executable replica set. The `resource-name` is an annotation that ApplicationExecutor class seems to be applying; it was introduced by https://github.com/dotnet/aspire/pull/3103. Compare line https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L1012 and line https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L1019 and notice that the Executable replica set name can have suffixes that `resouce-name` annotation will not have. 

If the goal here is to group Executables by its owning replica set (including the group for Executables that have no owning replica set, i.e. standalone) then what @adamint has seems sound to me.

JamesNK: If my understanding is correct, the difference is the `resource-name` could be missing `DcpOptions.ResourceNameSuffix`. What is that for, and when it would be set?

I tried configuring in a host JSON config file:

```json
{
  "DcpPublisher": {
    "ResourceNameSuffix": "r"
  }
}
```

And I found that setting that didn't impact grouping values in the UI:

image

The only oddity I see is the grouped name is `catalogservice` and not `catalogservice-r`. Does that matter?

karolz-ms: > If my understanding is correct, the difference is the `resource-name` could be missing `DcpOptions.ResourceNameSuffix`. What is that for, and when it would be set?
> 
The suffix is used by tests to ensure that object have probabilistically unique names, e.g. https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs#L163 

> 
> The only oddity I see is the grouped name is `catalogservice` and not `catalogservice-r`. Does that matter? Is it desirable? The display name is `catalogservice` everywhere else.

That is for you to decide, not me ☺️ I am just trying to help by answering questions about what is guaranteed/safe in terms of resource naming and DCP orchestration.

JamesNK: In that case, resource name and replica set name will always be the same in real world usage.

adamint: Sounds good, I'll remove the owner changes then. @JamesNK

src/Aspire.Dashboard/Model/KnownOwnerProperties.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Dashboard.Model;
+
+internal static class KnownOwnerProperties
+{
+    public const string ExecutableReplicaSetKind = "ExecutableReplicaSet";
+}
```
JamesNK, line 9: This file should be removed.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -23,6 +23,7 @@ public sealed class ResourceViewModel
     public required ImmutableArray<UrlViewModel> Urls { get; init; }
     public required FrozenDictionary<string, Value> Properties { get; init; }
     public required ImmutableArray<CommandViewModel> Commands { get; init; }
+
```
JamesNK, line 26:
```suggestion
```

src/Aspire.Dashboard/Components/Controls/ResourceSelectOptionTemplate.razor
```diff
@@ -6,14 +6,15 @@
...
 
 <FluentStack Orientation="Orientation.Horizontal" HorizontalGap="3">
-    @if (ViewModel.Id?.Type is OtlpApplicationType.ReplicaSet)
+    @if (ViewModel.Id is { } typeDetails && typeDetails.Type is OtlpApplicationType.ResourceGrouping)
     {
         <FluentIcon Icon="Icons.Regular.Size20.AppsList"
                     Color="Color.Accent"
                     Style="vertical-align: text-bottom;"/>
-        <span>@ViewModel.Name (<span style="color: var(--accent-foreground-rest)">@Loc[nameof(ControlsStrings.ReplicaSetLower)]</span>)</span>
+
```
JamesNK, line 14:
```suggestion
```

JamesNK (approved):

src/Aspire.Dashboard/Components/Controls/ResourceSelectOptionTemplate.razor
```diff
@@ -6,14 +6,14 @@
 @inject IStringLocalizer<ControlsStrings> Loc
 
 <FluentStack Orientation="Orientation.Horizontal" HorizontalGap="3">
-    @if (ViewModel.Id?.Type is OtlpApplicationType.ReplicaSet)
+    @if (ViewModel.Id is { } typeDetails && typeDetails.Type is OtlpApplicationType.ResourceGrouping)
```
drewnoakes, line 9: Can be a little more concise by matching with a property expression, in this case on the `Type` property:

```suggestion
    @if (ViewModel.Id is { Type: OtlpApplicationType.ResourceGrouping })
```

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -42,7 +42,7 @@ message ResourceCommand {
 
 // Represents a request to execute a command.
 // Sent by the dashboard to DashboardService.ExecuteResourceCommand.
-// Constructed with data from a corresponding 
+// Constructed with data from a corresponding
```
drewnoakes, line 45: This comment is incomplete. It should read:

```suggestion
// Constructed with data from a corresponding ResourceCommand.
```

src/Aspire.Dashboard/Components/Controls/ResourceSelectOptionTemplate.razor
adamint: done

JamesNK (approved):

PR #4911, opened 2024-07-15
Update runtime, aspnetcore, and extensions versions in a single dependabot PR

eerhardt:
That way we don't get conflicting PRs like

* https://github.com/dotnet/aspire/pull/4897
* https://github.com/dotnet/aspire/pull/4898
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4911)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,18 +29,11 @@ updates:
       AspNetCoreHealthChecks:
         patterns:
           - "AspNetCore.HealthChecks.*"
-      AspNetCore:
-        patterns:
-          - "Microsoft.AspNetCore.*"
-          - "Microsoft.Extensions.Features"
       AWS:
         patterns:
           - "AWS.*"
           - "AWSSDK.*"
           - "OpenTelemetry.*.AWS"
-      MicrosoftExtensions:
-        patterns:
-          - "Microsoft.Extensions.*"
       EntityFrameworkCore:
         patterns:
           - "Microsoft.EntityFrameworkCore.*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -50,6 +43,10 @@ updates:
       OpenTelemetry:
         patterns:
           - "OpenTelemetry.*"
+      NetPlatform:
+        patterns:
+          - "Microsoft.AspNetCore.*"
+          - "Microsoft.Extensions.*"
```
danmoseley, line 49: Add "Microsoft.Orleans*" as well?

also feel free to replace/update https://github.com/dotnet/aspire-samples/pull/369

danmoseley (approved):

danmoseley:
aside, it doesn't matter for what you're doing here, but these are wildcard patterns not regexes. So ".*" matches a literal dot.

.github/dependabot.yml
eerhardt: > Add "Microsoft.Orleans*" as well?

Does it ship at the same time on patch Tuesdays? Or do you mean create a new group for Orleans?

eerhardt:
> Thanks for doing this Eric. If preferred, we can change it so that we only subscribe to extensions which is the highest in the dependency tree, and make the rest of the updates use coherence dependencies so they get all bumped and matching in the same PR.

Dependabot supports this? If there is a better way than what I did here, feel free to update it.

joperezr:
No, sorry should've been more clear. We'd be using dependency-flow for these updates (instead of dependabot) and have dependabot ignore these package ids.

PR #4923, opened 2024-07-16
[tests] Extract support for building tests outside the repo into re-usable props+targets

radical:
## What
Extract bits required for running tests outside of repo into `Aspire.RepoTesting.{props,targets}`. And consolidate the various support bits (tasks/targets) spread around the repo. When running tests outside the repo they need some extra msbuild bits to allow building with aspire nugets instead of direct `ProjectReference`s to source projects.

## Why

There is a need to have some other projects run out-of-repo too, for example, `Aspire.Hosting.Testing.Tests`, and upcoming `Aspire.Playground.Tests`.

## Details

This provides support for running tests outside of the repo, for example on a helix agent.
- For this you need the source of the tests, and any dependencies.
- Instead of direct `ProjectReferences` to the various Aspire hosting, and component projects use
  `@(AspireProjectOrPackageReference)`.
  - These are converted to `ProjectReference` when `$(TestsRunningOutsideOfRepo) != true`.
  - But converted to `PackageReference` when `$(TestsRunningOutsideOfRepo) == true`.

- To allow building such test projects, the build is isolated and patched to build outside the
  repo by adding appropriate `Directory.Build.{props,targets}`, and `Directory.Packages.props`
  - and using a custom `nuget.config` which resolves the Aspire packages from the locally built packages
  - and a `Directory.Packages.Versions.props` is generated with PackageVersions taken from the repo
    - This also adds properties named in `@(PropertyForHelixRun)` from the repo, like `$(NetCurrent)`

Also:
- Update other project to track the new changes

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4923)


Directory.Build.props
```diff
@@ -7,7 +7,8 @@
     <LangVersion>preview</LangVersion>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <SharedDir>$(MSBuildThisFileDirectory)/src/Shared/</SharedDir>
-    <TestsSharedDir>$(MSBuildThisFileDirectory)/tests/Shared/</TestsSharedDir>
+    <TestsSharedDir>$([MSBuild]::NormalizeDirectory($(MSBuildThisFileDirectory), 'tests', 'Shared'))</TestsSharedDir>
+    <TestsSharedRepoTestingDir>$([MSBuild]::NormalizeDirectory($(TestsSharedDir), 'RepoTesting'))</TestsSharedRepoTestingDir>
```
eerhardt, line 11: (nit) can/should these go into tests/Directory.Build.props?

tests/Directory.Build.targets
```diff
@@ -19,12 +21,42 @@
...
     <MakeDir Directories="$(TestArchiveTestsDir)" />
-    <ZipDirectory SourceDirectory="$(OutDir)"
+    <ZipDirectory SourceDirectory="$(TestsArchiveSourceDir)"
                   DestinationFile="$([MSBuild]::NormalizePath('$(TestArchiveTestsDir)', '$(MSBuildProjectName).zip'))"
                   Overwrite="true" />
   </Target>
 
-  <Import Project="Shared\Aspire.Workload.Testing.targets" Condition="'$(IsWorkloadTestProject)' == 'true'" />
+  <!-- Used for running one helix job per test class -->
+  <Target Name="_ExtractTestClassNames"
```
eerhardt, line 35: Why do we want one helix job per test class? That seems extreme, isn't it?

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
```diff
@@ -22,14 +27,14 @@
...
-    <None Include="..\.editorconfig" Link="testassets\%(FileName)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
-    <None Condition="'$(InstallWorkloadForTesting)' == 'true'" Include="$(PatchedNuGetConfigPath)" Link="testassets\testproject\nuget.config" CopyToOutputDirectory="PreserveNewest" />
-    <None Include="$(RepoRoot)Directory.Packages.props" Link="testassets\testproject\Directory.Packages.repo.props" CopyToOutputDirectory="PreserveNewest" />
-
     <PackageReference Include="Microsoft.Extensions.Http.Resilience" />
     <PackageReference Include="Microsoft.DotNet.XUnitExtensions" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(_BuildForTestsRunningOutsideOfRepo)' == 'true'">
+    <None Include="..\testproject\**\*" Link="testassets\testproject\%(RecursiveDir)%(FileName)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
```
eerhardt, line 35: Why do we need both this and line 16? Can it be consolidated to 1 place?

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
```diff
@@ -22,14 +27,14 @@
...
+    <None Include="..\.editorconfig" Link="testassets\%(FileName)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
```
eerhardt, line 36: Should copying the tests\.editorconfig be part of the "infra"?

tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj
```diff
@@ -28,13 +27,14 @@
...
-    <ProjectReference Include="..\..\src\Aspire.Hosting.Dapr\Aspire.Hosting.Dapr.csproj" />
+    <ProjectReference Include="..\..\src\Aspire.Hosting.Dapr\Aspire.Hosting.Dapr.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.Garnet\Aspire.Hosting.Garnet.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.Milvus\Aspire.Hosting.Milvus.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.MongoDB\Aspire.Hosting.MongoDB.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.MySql\Aspire.Hosting.MySql.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.Nats\Aspire.Hosting.Nats.csproj" IsAspireProjectResource="false" />
-    <ProjectReference Include="..\..\src\Aspire.Hosting.Testing\Aspire.Hosting.Testing.csproj" />
+    <ProjectReference Include="..\..\src\Aspire.Hosting.Qdrant\Aspire.Hosting.Qdrant.csproj" IsAspireProjectResource="false" />
```
eerhardt, line 36:
```suggestion
```

Shouldn't be added these back

tests/Aspire.Hosting.Tests/Directory.Build.props
```diff
@@ -1,8 +1,7 @@
 <Project>
+  <PropertyGroup>
+    <IsAspireHost>true</IsAspireHost>
```
eerhardt, line 3: Is there a way to make it so this stays defined in the .csproj?

tests/Shared/RepoTesting/Directory.Packages.Helix.props
```diff
@@ -30,24 +41,53 @@
...
+    <PackageVersion Include="Aspire.StackExchange.Redis.OutputCaching" Version="$(PackageVersion)" />
+
+    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery" Version="$(PackageVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Abstractions" Version="$(PackageVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Dns" Version="$(PackageVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery.Yarp" Version="$(PackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup Label="For tests">
+    <PackageVersion Include="xunit" Version="$(XunitVersion)" />
```
eerhardt, line 82: Why are the xunit packages needed to be defined here, but not any other external packages?

tests/Shared/RepoTesting/Aspire.Testing.Repo.props
```diff
@@ -0,0 +1,15 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <!-- IsAspireHost property might not be available here, so allow infering that from the project name -->
+  <ImportGroup Condition="'$(RepoRoot)' != 'null' and '$(TestsRunningOutsideOfRepo)' != 'true' and ($(MSBuildProjectName.EndsWith('AppHost')) or '$(IsAspireHost)' == 'true')">
+    <Import Project="$(RepoRoot)src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.props" Condition="Exists('$(RepoRoot)src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.props')" />
+  </ImportGroup>
+
+ <PropertyGroup>
+    <SharedDir Condition="'$(SharedDir)' != ''">$([MSBuild]::EnsureTrailingSlash($(SharedDir)))</SharedDir>
```
eerhardt, line 13: Can we make it so that whoever defined SharedDir to ensure the trailing slash? We usually don't do this for directory properties. We just assumed whoever set it did it right.

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
```diff
@@ -8,39 +8,22 @@
...
-    <PackageReference Include="Aspire.Hosting.MySql" />
-    <PackageReference Include="Aspire.Hosting.NodeJs" />
-    <PackageReference Include="Aspire.Hosting.Oracle" />
-    <PackageReference Include="Aspire.Hosting.PostgreSQL" />
-    <PackageReference Include="Aspire.Hosting.RabbitMQ" />
-    <PackageReference Include="Aspire.Hosting.Redis" />
-    <PackageReference Include="Aspire.Hosting.SqlServer" />
-    <PackageReference Include="Aspire.Hosting.Valkey" />
+  <ItemGroup>
+    <NonComponentReferenceForTests Include="Aspire.Hosting.Azure" IsAspireProjectResource="false" />
```
eerhardt, line 12: Why is this named NonComponentReferenceForTests? What does "Component" matter?

eerhardt (commented):
Some questions

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
radical: Lack of a better name! Please suggest one :)

tests/Directory.Build.targets
radical: This is used for workload tests, where the tests are expensive. I'm just making it available for other projects.

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
radical: The property definition gets used by the testing.repo targets to know where to copy the various msbuild files. This block can use that property though. Fixing.

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
radical: This exists in `main`. IIRC, this was become some test seemed to indirectly depend on this file being available. So, we are copying it here as part of preparing the output dir to be usable for out-of-repo builds.

radical: And I didn't hit a need for this other projects that use out-of-repo (like future playground, and hosting.testing tests).

tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj
radical: Oops, bad merge!

tests/Aspire.Hosting.Tests/Directory.Build.props
radical: We can keep it there *also*! It is needed basically for this in `tests/Shared/RepoTesting/Aspire.Testing.Repo.props` which gets imported early - before the project file itself:
```xml
  <!-- IsAspireHost property might not be available here, so allow infering that from the project name -->
  <ImportGroup Condition="'$(RepoRoot)' != 'null' and '$(TestsRunningOutsideOfRepo)' != 'true' and ($(MSBuildProjectName.EndsWith('AppHost')) or '$(IsAspireHost)' == 'true')">
    <Import Project="$(RepoRoot)src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.props" Condition="Exists('$(RepoRoot)src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.props')" />
  </ImportGroup>
```

We are using `($(MSBuildProjectName.EndsWith('AppHost')) or '$(IsAspireHost)' == 'true')` to determine whether it is an aspire host or not. And in this case the name does not help. Open to any suggestions of course.

tests/Shared/RepoTesting/Directory.Packages.Helix.props
radical: When a test project is built in the repo, Arcade adds these references automatically. But when we are building outside the repo, the Arcade targets are not used, thus we need to explicitly add these. 

Any other external packages needed by a project would be added by those projects themselves. We add `PackageVersion` for them in `./tests/Shared/RepoTesting/Directory.Packages.Helix.props`. Hm this should be added in a common `README.md` too.

Directory.Build.props
radical: I had to put these here to be accessible from `playground` apps, which would get similar changes to remove the hosting targets import.

tests/Directory.Build.props
```diff
@@ -1,13 +1,16 @@
 <Project>
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
+  <Import Project="$(TestsSharedRepoTestingDir)Aspire.Testing.Repo.props" />
```
radical, line 4: There is a little disconnect between the naming of this - `Aspire.Testing.Repo.*`, and the directory in `tests/Shared/RepoTesting`.

`tests/Shared/TestingRepo` sounds odd though. I could rename the targets to be `Aspire.Repo.Testing.*`. Any preference?

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
eerhardt: Maybe drop `Component` and just have `ReferenceForTests`. Or `OutsideOfRepoReference`?

tests/Directory.Build.props
eerhardt: Maybe `Aspire.RepoTesting.*`? That would align nicely with the directory name.

radical: Fixed.

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
radical: After some offline discussion with @eerhardt I was able to combine the two into a single item `AspireProjectOrPackageReference`. Feel free to suggest a different name though :)

tests/Shared/nuget-with-package-source-mapping.config
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
```
eerhardt, line 1: This is the 3rd time we have a nuget.config copied in the repo. Is there a way to use the root and just tweak it in places we need it to be special?

eerhardt (approved):
LGTM

tests/Shared/nuget-with-package-source-mapping.config
radical: Yeah, we had a task doing that earlier because we needed to put in the path to the built nugets. But now we use `%BUILT_NUGETS_PATH%` as the source in the nuget.config for tests, so it always resolves it from the environment variable.

I'll bring the task back to get the sources from the root nuget.config, add the source for built-nugets, and add the package source mapping as needed.
Added this as follow up work - https://github.com/dotnet/aspire/issues/4946 .

tests/Shared/WorkloadTesting/Aspire.Shared.WorkloadTesting.targets
```diff
@@ -1,10 +1,16 @@
...
-   <Compile Include="$(MSBuildThisFileDirectory)*.cs" Link="WorkloadTestingCommon" />
+  <Import Project="$(MSBuildThisFileDirectory)..\Playwright\Playwright.targets" />
 
-   <!-- needed for the RequiresDocker attribute, and discoverer -->
-   <ProjectReference Include="$(RepoRoot)tests\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" />
- </ItemGroup>
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)*.cs" Link="WorkloadTestingCommon" />
+
+    <_DataFile Include="$(MSBuildThisFileDirectory)data\**\*" Link="testassets\%(RecursiveDir)%(FileName)%(Extension)" CopyToOutputDirectory="PreserveNewest" />
```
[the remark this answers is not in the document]
radical: Added as follow up in https://github.com/dotnet/aspire/issues/4946

tests/Shared/WorkloadTesting/data/nuget8.config
```diff
@@ -8,6 +8,7 @@
     <clear />
     <add key="built-local" value="%BUILT_NUGETS_PATH%" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
```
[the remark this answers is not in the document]
radical: I don't recall why I had to that one :( I will remove this in the follow up PR, and see what fails.
Also, with using a single nuget.org for sources this will coming from the root nuget.config .

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
```diff
@@ -11,8 +11,13 @@
...
     <InstallWorkloadForTesting>$(TestUsingWorkloads)</InstallWorkloadForTesting>
 
-    <TestsRunningOutsideOfRepo Condition="'$(TestsRunningOutsideOfRepo)' == '' and '$(ContinuousIntegrationBuild)' == 'true'">true</TestsRunningOutsideOfRepo>
-    <DefineConstants Condition="'$(TestsRunningOutsideOfRepo)' == 'true'">TESTS_RUNNING_OUTSIDE_OF_REPO;$(DefineConstants)</DefineConstants>
+    <!-- This isn't needed as the sdk will add them by default -->
+    <IncludeTestPackages>false</IncludeTestPackages>
+    <DeployOutsideOfRepoSupportFilesDir>testassets\testproject\</DeployOutsideOfRepoSupportFilesDir>
+
+    <_BuildForTestsRunningOutsideOfRepo Condition="'$(TestsRunningOutsideOfRepo)' == 'true' or '$(ContinuousIntegrationBuild)' == 'true'">true</_BuildForTestsRunningOutsideOfRepo>
+    <DeployOutsideOfRepoSupportFiles>$(_BuildForTestsRunningOutsideOfRepo)</DeployOutsideOfRepoSupportFiles>
```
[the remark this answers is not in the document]
radical: The condition is needed in multiple places, and it clarifies that this project is being *built* for running tests out-of-repo vs project itself being prepared to run out of repo.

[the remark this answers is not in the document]
radical: Merging the PR, but if we decide for a change here then I can do it in the follow up.

PR #4936, opened 2024-07-17
Limit number of console log messages retained in dashboard process

### LINKED ISSUE #4849: Bound number of console logs retained in the dashboard process
The app host has a hard-coded limit of 10,000 in its circular buffer.

The dashboard client does not have any such limit, allowing memory to grow indefinitely.

The dashboard should have such a limit. Perhaps it should be provided by the server, or via configuration.

drewnoakes:
Fixes #4849

Defaults limit to 10,000, matching the limit used in the AppHost.

Value may be controlled via configuration. For example, given:

```json
{
  "DetailedErrors": true,
  "Dashboard": {
    "Frontend": {
      "MaxConsoleLogCount": 5
    }
  }
}
```

Only five rows are visible (no scrollbar here):

bounded-console-logs

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4936)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
@@ -7,7 +7,6 @@ using System.Diagnostics;
 using Aspire.Dashboard.Components.Controls;
 using Aspire.Dashboard.Components.Layout;
 using Aspire.Dashboard.Components.Resize;
-using Aspire.Dashboard.Configuration;
 using Aspire.Dashboard.Extensions;
 using Aspire.Dashboard.Model;
 using Aspire.Dashboard.Model.Otlp;
@@ -16,7 +15,6 @@ using Aspire.Dashboard.Resources;
 using Aspire.Dashboard.Utils;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Server.ProtectedBrowserStorage;
-using Microsoft.Extensions.Options;
 
 namespace Aspire.Dashboard.Components.Pages;
 
@@ -39,9 +37,6 @@ public sealed partial class ConsoleLogs : ComponentBase, IAsyncDisposable, IPage
 
     [Inject]
     public required DimensionManager DimensionManager { get; init; }
-
-    [Inject]
-    public required IOptionsMonitor<DashboardOptions> Options { get; init; }
 
     [CascadingParameter]
     public required ViewportInformation ViewportInformation { get; init; }
@@ -183,8 +178,6 @@ public sealed partial class ConsoleLogs : ComponentBase, IAsyncDisposable, IPage
     {
         if (firstRender)
         {
-            LogViewerViewModel.LogEntries.MaximumEntryCount = Options.CurrentValue.Frontend.MaxConsoleLogCount;
-
             // Let anyone waiting know that the render is complete, so we have access to the underlying log viewer.
             _whenDomReady.SetResult();
         }
--- a/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using Aspire.Dashboard.Components.Controls;
 using Aspire.Dashboard.Components.Layout;
 using Aspire.Dashboard.Components.Resize;
+using Aspire.Dashboard.Configuration;
 using Aspire.Dashboard.Extensions;
 using Aspire.Dashboard.Model;
 using Aspire.Dashboard.Model.Otlp;
@@ -15,6 +16,7 @@ using Aspire.Dashboard.Resources;
 using Aspire.Dashboard.Utils;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Server.ProtectedBrowserStorage;
+using Microsoft.Extensions.Options;
 
 namespace Aspire.Dashboard.Components.Pages;
 
@@ -37,6 +39,9 @@ public sealed partial class ConsoleLogs : ComponentBase, IAsyncDisposable, IPage
 
     [Inject]
     public required DimensionManager DimensionManager { get; init; }
+
+    [Inject]
+    public required IOptionsMonitor<DashboardOptions> Options { get; init; }
 
     [CascadingParameter]
     public required ViewportInformation ViewportInformation { get; init; }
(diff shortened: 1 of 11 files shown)


src/Aspire.Dashboard/ConsoleLogs/LogEntries.cs
```diff
@@ -65,15 +67,23 @@ void InsertLogEntry(List<LogEntry> logEntries, int index, LogEntry logEntry)
...
             else
             {
-                logEntry.LineNumber = logEntries[index - 1].LineNumber + 1;
+                logEntry.LineNumber = _logEntries[index - 1].LineNumber + 1;
+            }
+
+            // Trim old log messages if we have a maximum and we're over it.
+            if (MaximumEntryCount is not (null or 0) && _logEntries.Count >= MaximumEntryCount && index is not 0)
+            {
+                _logEntries.RemoveAt(0);
```
JamesNK, line 76: Removing at index 0 will copy all the data of the list back one position. At 10,000 items, and possibly hundreds of log entries a second this could consume a lot of machine resources.

I wrote `CircularBuffer<T>` for telemetry that handles this automatically (removing oldest just involves changing start/end indexes).

I think you should see whether `CircularBuffer<T>` makes sense here.

src/Aspire.Dashboard/README.md
```diff
@@ -63,6 +63,10 @@ Set `Dashboard:Frontend:AuthMode` to `OpenIdConnect`, then add the following con
 - `Dashboard:Frontend:OpenIdConnect:RequiredClaimType` specifies the (optional) claim that be present for authorized users. Defaults to empty.
 - `Dashboard:Frontend:OpenIdConnect:RequiredClaimValue` specifies the (optional) value of the required claim. Only used if `Dashboard:Frontend:OpenIdConnect:RequireClaimType` is also specified. Defaults to empty.
 
+#### Memory limits
+
+- `Dashboard:Frontend:ConsoleLogHistoryLimit` specifies the (optional) maximum number of console log messages to keep in memory. Defaults to 10,000. When the limit is reached, the oldest messages are removed.
```
JamesNK, line 68: The telemetry limits configuration have this naming standard: MaxLogCount

I think this should be called MaxConsoleLogCount to follow that standard.

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -55,6 +55,11 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
                 break;
         }
 
+        if (options.Frontend.ConsoleLogHistoryLimit <= 0)
+        {
+            errorMessages.Add("ConsoleLogHistoryLimit must be greater than zero.");
```
JamesNK, line 60: Add this name to DashboardConfigNames and use the ConfigKey property as the name. See other error messages in this type for examples.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -131,6 +131,16 @@ public sealed class FrontendOptions
     public string? EndpointUrls { get; set; }
     public FrontendAuthMode? AuthMode { get; set; }
     public string? BrowserToken { get; set; }
+
+    /// <summary>
+    /// Gets and sets an optional limit on the number of log messages to be retained in the viewer.
+    /// </summary>
+    /// <remarks>
+    /// When non-null, the viewer will retain at most this number of log messages. When the limit is reached, the oldest messages will be removed.
```
JamesNK, line 139: The property isn't nullable

tests/Aspire.Dashboard.Tests/ConsoleLogsTests/LogEntriesTests.cs
```diff
@@ -31,4 +31,44 @@ public void InsertSorted_OutOfOrderWithSameTimestamp_ReturnInOrder()
             l => Assert.Equal("2", l.Content),
             l => Assert.Equal("3", l.Content));
     }
+
+    [Fact]
+    public void InsertSorted_TrimsToMaximumEntryCount_Ordered()
+    {
+        // Arrange
+        var logEntries = new LogEntries { MaximumEntryCount = 2, BaseLineNumber = 1 };
```
JamesNK, line 39: Could you add a test with MaximumEntryCount = int.MaxValue? Just double check that items are still correctly added when there isn't really a limit.

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -55,6 +55,11 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
                 break;
         }
 
+        if (options.Frontend.ConsoleLogHistoryLimit <= 0)
```
JamesNK, line 58: Should zero be allowed? Maybe someone wants to disable showing console logs.

I'm not sure whether OTEL telemetry limits can be zero. I don't think I ever tested.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -178,6 +183,8 @@ protected override void OnAfterRender(bool firstRender)
     {
         if (firstRender)
         {
+            LogViewerViewModel.LogEntries.MaximumEntryCount = Options.CurrentValue.Frontend.ConsoleLogHistoryLimit;
```
JamesNK, line 186: Since this VM is now injected from DI, its ctor could have `IOptionsMonitor<DashboardOptions>`. The max could be set in the VM ctor without requiring console logs to wire them up. Seems a bit more encapsulated to me.

tests/Aspire.Dashboard.Tests/ConsoleLogsTests/LogEntriesTests.cs
drewnoakes: The first test covers this I think.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
drewnoakes: Agree, that is nicer.

src/Aspire.Dashboard/ConsoleLogs/LogEntries.cs
drewnoakes: I'd considered adding a circular buffer here but didn't realise we already had one in the code. Will use it.

JamesNK (approved):

Issue #4878, opened 2024-07-12
Can't WaitForResourceAsync KnownResourceStates.Finished


eerhardt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

It's not possible to get notification when a resource has been stopped/finished when after `await app.StopAsync();`.

I'm trying to write a test that uses a Docker Volume, and when the test is done, it deletes the volume. However, the volume is still in use even after `await app.StopAsync();` returns, which means the container using the volume is still running.

I'd like a way to get notified when the resource/container is finished so I can delete the volume.

One reason this isn't possible today is that WaitForResourceAsync cancels once the app is stopping here:
https://github.com/dotnet/aspire/blob/681f2e754dc849f96abafcd2829783e69155abf1/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L89-L93

Once the app is stopping, it cancels this loop, which means we won't get more notifications about the resource.

### Expected Behavior

I'd like to get a notification when the container is finished.

### Steps To Reproduce

```csharp
await using var app = await appHost.BuildAsync();
var resourceNotificationService = app.Services.GetRequiredService<ResourceNotificationService>();
await app.StartAsync();

await app.StopAsync();

await resourceNotificationService.WaitForResourceAsync("webfrontend", KnownResourceStates.Finished).WaitAsync(TimeSpan.FromSeconds(30));
```

### Anything else?

cc @DamianEdwards @karolz-ms @mitchdenny

karolz-ms:
@eerhardt and I chatted. We think the main issue is that as soon as `app.StopAsync()` is called, the resource service and ApplicationExecutor stops listening to resource changes. So even if `DcpOptions.DeleteResourcesOnShutdown` is set, there is no way for the Aspire model to learn that the container (or any resource for that matter) has changed state, or was deleted.

It would probably make sense to change this behavior and differentiate between stopping the application (which would stop application executables and containers, but preserve the Aspire model and its ability to get state updates) vs disposing of the application (which would dispose of the model, stop DCP orchestrator etc.)

davidfowl:
@DamianEdwards this is the code you added to unhook the notification service on shutdown automatically. We'd need to allow listening to changes past the stopping of the application (but before disposal)

DamianEdwards:
> We'd need to allow listening to changes past the stopping of the application (but before disposal)

Is there an existing signal for that or would we need to do something custom?

davidfowl:
[...]

PR #4947, opened 2024-07-17
Kafka WithDataVolume and WithDataBindMount don't persist messages

### LINKED ISSUE #4909: Kafka WithDataVolume and WithDataBindMount don't persist messages

### Describe the bug

Calling WithDataVolume or WithDataBindMount on a Kafka resource doesn't persist data across runs of the application. We should ensure we have the right settings set up so the data gets persisted across app runs.

### Expected Behavior

When using WithDataVolume or WithDataBindMount, the messages that are produced should be perserved across app runs. So when the app spins up again, consuming those messages should be possible.

### Steps To Reproduce

Run the following test

```csharp
    [Theory]
    [InlineData(true)]
    [InlineData(false)]
    [RequiresDocker]
    public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
    {
        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(3));
        var topic = "test-topic";
        string? volumeName = null;
        string? bindMountPath = null;

        try
        {
            var builder1 = CreateDistributedApplicationBuilder();
            var kafka1 = builder1.AddKafka("kafka");

            if (useVolume)
            {
                // Use a deterministic volume name to prevent them from exhausting the machines if deletion fails
                volumeName = VolumeNameGenerator.CreateVolumeName(kafka1, nameof(WithDataShouldPersistStateBetweenUsages));

                // if the volume already exists (because of a crashing previous run), try to delete it
                DockerUtils.AttemptDeleteDockerVolume(volumeName);
                kafka1.WithDataVolume(volumeName);
            }
            else
            {
                bindMountPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
                kafka1.WithDataBindMount(bindMountPath);
            }

            using (var app = builder1.Build())
            {
                await app.StartAsync();
                try
                {
                    var hb = Host.CreateApplicationBuilder();

                    hb.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
                    {
                        [$"ConnectionStrings:{kafka1.Resource.Name}"] = await kafka1.Resource.ConnectionStringExpression.GetValueAsync(default)
                    });

        hb.AddKafkaProducer<string, string>("kafka");

                    using (var host = hb.Build())
                    {
                        await host.StartAsync();

                        var producer = host.Services.GetRequiredService<IProducer<string, string>>();
                        for (var i = 0; i < 10; i++)
                        {
                            await producer.ProduceAsync(topic, new Message<string, string> { Key = "test-key", Value = $"test-value{i}" });
                        }
                    }
                }
                finally
                {
                    // Stops the container, or the Volume/mount would still be in use
                    await app.StopAsync();
                }
            }

            var builder2 = CreateDistributedApplicationBuilder();
            var kafka2 = builder2.AddKafka("kafka");

            if (useVolume)
            {
                kafka2.WithDataVolume(volumeName);
            }
            else
            {
                kafka2.WithDataBindMount(bindMountPath!);
            }

            using (var app = builder2.Build())
            {
                await app.StartAsync();
                try
                {
                    var hb = Host.CreateApplicationBuilder();

                    hb.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
                    {
                        [$"ConnectionStrings:{kafka2.Resource.Name}"] = await kafka2.Resource.ConnectionStringExpression.GetValueAsync(default)
                    });

                    hb.AddKafkaConsumer<string, string>("kafka", consumerBuilder =>
                    {
                        consumerBuilder.Config.GroupId = "aspire-consumer-group";
                        consumerBuilder.Config.AutoOffsetReset = AutoOffsetReset.Earliest;
                    });

                    using (var host = hb.Build())
                    {
                        await host.StartAsync();

                        var consumer = host.Services.GetRequiredService<IConsumer<string, string>>();
                        consumer.Subscribe(topic);
                        for (var i = 0; i < 10; i++)
                        {
                            var result = consumer.Consume(cts.Token);

                            Assert.Equal($"test-key", result.Message.Key);
                            Assert.Equal($"test-value{i}", result.Message.Value);
                        }
                    }
                }
                finally
                {
                    // Stops the container, or the Volume/mount would still be in use
                    await app.StopAsync();
                }
            }
        }
        finally
        {
            if (volumeName is not null)
            {
                DockerUtils.AttemptDeleteDockerVolume(volumeName);
            }

            if (bindMountPath is not null)
            {
                try
                {
                    File.Delete(bindMountPath);
                }
                catch
                {
                    // Don't fail test if we can't clean the temporary folder
                }
            }
        }
    }
```

### Exceptions (if any)

```
  Message: 
Confluent.Kafka.ConsumeException : Subscribed topic not available: test-topic: Broker: Unknown topic or partition

  Stack Trace: 
Consumer`2.Consume(Int32 millisecondsTimeout)
Consumer`2.Consume(CancellationToken cancellationToken)
KafkaFunctionalTests.WithDataShouldPersistStateBetweenUsages(Boolean useVolume) line 169
```

### .NET Version info

_No response_

### Anything else?

_No response_

stbau04:
Changes Kafka data directory (unit test mentioned in #4909 still fails but with an other exception - seems as the new data directory is working)
Fixes  #4909
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4947)

### DIFF
--- a/src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
@@ -91,7 +91,7 @@ public static class KafkaBuilderExtensions
     /// <param name="isReadOnly">A flag that indicates if this is a read-only volume.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<KafkaServerResource> WithDataVolume(this IResourceBuilder<KafkaServerResource> builder, string? name = null, bool isReadOnly = false)
-        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/var/lib/kafka/data", isReadOnly);
+        => builder.WithVolume(name ?? VolumeNameGenerator.CreateVolumeName(builder, "data"), "/tmp/kraft-combined-logs", isReadOnly);
 
     /// <summary>
     /// Adds a bind mount for the data folder to a Kafka container resource.
@@ -101,7 +101,7 @@ public static class KafkaBuilderExtensions
     /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<KafkaServerResource> WithDataBindMount(this IResourceBuilder<KafkaServerResource> builder, string source, bool isReadOnly = false)
-        => builder.WithBindMount(source, "/var/lib/kafka/data", isReadOnly);
+        => builder.WithBindMount(source, "/tmp/kraft-combined-logs", isReadOnly);
 
     private static void ConfigureKafkaContainer(EnvironmentCallbackContext context, KafkaServerResource resource)
     {
--- a/tests/Aspire.Hosting.Kafka.Tests/KafkaFunctionalTests.cs
+++ b/tests/Aspire.Hosting.Kafka.Tests/KafkaFunctionalTests.cs
@@ -74,7 +74,6 @@ public class KafkaFunctionalTests(ITestOutputHelper testOutputHelper)
     }
 
     [Theory]
-    [ActiveIssue("https://github.com/dotnet/aspire/issues/4909")]
     [InlineData(true)]
     [InlineData(false)]
     [RequiresDocker]


eerhardt:
Can you also enable this test?

https://github.com/dotnet/aspire/blob/8605f40ac701fd711295c4939cd5fa2553bbb593/tests/Aspire.Hosting.Kafka.Tests/KafkaFunctionalTests.cs#L76-L81

stbau04:
Yeah sure, test is updated

stbau04:
The kafka tests are still failing, but with an other exception. @eerhardt should they already work? (There are also exceptions in the Playwright test, but i guess this is not related to the kafka resource?)

eerhardt:
> The kafka tests are still failing, but with an other exception. @eerhardt should they already work? 

If they need to be fixed, can you please fix them? I wasn't able to valdiate them because of this issue. But I followed the same test pattern as used elsewhere.

> (There are also exceptions in the Playwright test, but i guess this is not related to the kafka resource?)

Yeah, this is unrelated flakiness in our tests that we are trying to fix.

stbau04:
@eerhardt There are currently two problems with the resource:
- Volumes/Bind mounts are mounted so that they can only be written by root user - Kafka is run using a different user. The best approach i found for that problem would be to start Kafka as root using (--user root) on the container, but I am not sure whether this has some disadvantages/is a bad practice
- The `await app.StopAsync();` does not actually wait for the container to exit, meaning that the first container still has a lock on the files while the second container is already trying to read - leading to a timeout. The only fix i could currently figure out would be to add a delay before starting the second container (at most with a one minute delay it works). Is there any other way to ensure that the container is shut down?

eerhardt:
>  Is there any other way to ensure that the container is shut down?

We have https://github.com/dotnet/aspire/issues/4878 to try to come up with a solution for that. cc @karolz-ms @mitchdenny 

> Volumes/Bind mounts are mounted so that they can only be written by root user 

You mean the root user inside the container? And Kafka is running as a separate user from the root user, so it can't access the directory?

@karolz-ms @danegsta - thoughts on this scenario? Does DCP support giving access to volumes and mounts to other users?

mitchdenny:
Hrm interesting. Yeah I think we need to solve this. I don't think we should expect the container workloads to change the user they are running as since that seems like a lot of containerized workloads would break if we did that.

How is this solved outside of Aspire when people want to use bindmounts and data volumes. Its not like we are doing anything special here?

karolz-ms:
@stbau04 this is a known issue with bind mounts, in the sense that, depending on what container orchestrator flavor/host OS you use, there might be different steps necessary to ensure that the process running inside the container has permissions to the bind mount https://www.reddit.com/r/docker/comments/1b8dnm6/is_there_an_easy_solution_to_the_permission/

That said there is the host side and the container side, and so you can try to change the user the container is running as (as you tried, running as `root`), or you can try adjust permissions on the folder that you are binding into the container so that the container user has access. Again, steps may differ depending on container orchestrator (Docker and Podman are supported by Aspire) and host OS, but the rule of thumb is that when specifying permissions on the host side, the user ID must match the user ID for the effective user inside the container. Running container as non-root and changing permissions on the host side is considered a safer practice than running as root inside the container. Hope this helps!

eerhardt:
@karolz-ms - do you think we should remove this API until we can make it work? It doesn't sound like there is a straight-forward path to making it work.

danegsta:
One thing to check for Docker volume mounts (both named volumes and bind mounts) is that the behavior can be different depending on whether you're mounting to a path that already exists in a container image or not. If you volume mount to a non-existent folder (even if it's a child of an existing folder), Docker will create the new folder path in the container using root permissions, even if the container itself will run as a non-root user. If the folder already exists, it will preserve the existing permissions inside the container. For bind mounts, it'll also preserve the local file system permissions for any child files/folders that are introduced as part of the mount (named volumes are usually easier to deal with if you don't need to write to/read from files on the host).

eerhardt:
@stbau04 - are you still working on this PR? Can you try the solution above?

stbau04:
@eerhardt Sadly i don't have a lot time at the moment to figure something out. I can give the solution a try but in my opinion it's not a good idea to automatically manipulate the host folder permissions

stbau04:
@eerhardt Seems like changing the directory permission on the host does not work for windows: https://stackoverflow.com/questions/66388158/changing-file-permissions-from-a-mounted-file-from-inside-the-docker-container
So i think there are are basically two possible approaches:
- Run the container as root user (--u root)
- Modify the container image so that chmod is executed (or maybe ensure that the folder already exists with right permissions)

I am soon on holiday for a few weeks, so i am not able to do any more work/research on this topic in the near future

eerhardt:
Thanks for the information and the attempts, @stbau04.

> i am not able to do any more work/research on this topic in the near future

Closing this PR until it can be actively worked on.

PR #4951, opened 2024-07-18
Fix Aspire.Hosting.Tests build on VS

radical:
When building only the `Aspire.Hosting.Tests` in VS, it would fail with
errors about `ActiveIssue` not being found. This is because the project
was being compiled with none of the dependencies from
`Aspire.Components.Common.Tests` getting used in the
`Aspire.Hosting.Tests` build.

Building the solution worked fine though.

The problem seems to be a duplicate project reference to
`Aspire.Components.Common.Tests`:

```
<ProjectReference Include="..\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" IsAspireProjectResource="false" />

<ProjectReference Include="..\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" />
```

Removing the second reference fixes the issue. I'm not sure *why* this
broke VS though.

Note that this works perfectly find on command line, whether building the whole solution or just the tests project on Windows, and macOS.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4951)


eerhardt (approved):
Nice find. I wonder why it doesn't fail when you duplicate the ProjectReferences like this...

Issue #4893, opened 2024-07-15
[AzureTools][Aspire] The sequence number of the log changes when returning "Resource" and then clicking "View" of the project or container under the "Logs" in the Dashboard again


v-mengwe:
**Update:** More is broken here than just the order of logs. See https://github.com/dotnet/aspire/issues/4893#issuecomment-2274700970 for details.

---

**Regression:** Not a regression, this issue also reproduces on VS 17.11.0 preview 3 with Aspire 8.0.0

**Clean machine:** Win11 x64 23h2 ENU
**VS Version:** VS 17.12.0 Preview 1.0 [35113.27.main]
**Aspire Version:** 8.1.0-preview.1.24362.2
Apply NuGet Feeds

**REPRO STEPS:**

1. Create a new .NET Aspire Starter Application with redis checked -> Run `dotnet run` in the AppHost.project folder.
2. Click "View" of the project or container under the "Logs" in the Dashboard.
3. Return to "Resource" -> Click "View" of the project or container under the "Logs" in the Dashboard again.
4. Check whether the sequence number of the log remains unchanged.

**Expect:**
The sequence number of the log remains unchanged.

**Actual:**
The sequence number of the log changes.
![Actual1 gif](https://github.com/user-attachments/assets/50c1ce64-821e-44c5-a0af-94126e02146f)

**More Info:**
This issue also **reproduces** for other projects.

adamint:
@mitchdenny this isn't coming from dashboard, maybe from AppHost? see mention above.

mitchdenny:
I think we discussed tackling this in 8.2.

karolz-ms:
Based on the screenshot, this might be a duplicate of https://github.com/dotnet/aspire/issues/4942, which was fixed yesterday.

JamesNK:
The blank lines are unrelated.

JamesNK:
The issues with logging are caused by how logging was originally designed: the host is always listening to logs from DCP and stores them in memory. That design was then changed to [only listening when someone views the Console logs page in the dashboard](https://github.com/dotnet/aspire/pull/3235).

The change to only listen while the Console logs page is open means that logs are fetched from DCP and "replayed" into memory in the host. That causes issues:

[...]

PR #4955, opened 2024-07-18
Fix line numbers in console

JamesNK:
Addresses https://github.com/dotnet/aspire/issues/4893

Storing line number state in logger doesn't work because they're reused between watches. This PR moves tracking the log number up a level to `ResourceLogSource`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4955)


JamesNK:
This isn't going to work

PR #4957, opened 2024-07-18
Clear log messages when disposed

### LINKED ISSUE #4954: Dashboard log viewer showing duplicate log entries.

### Describe the bug

I have been able to reliably reproduce a log entry duplication issue. This is what it looks like:

image

### Expected Behavior

Log entries shouldn't be duplicated in the log viewer in the dashboard.

### Steps To Reproduce

For this repro I am using the SqlServerEndToEnd in the playgrounds folder in the repo.

1. Set `playground/SqlServerEndToEnd/SqlServerEndToEnd.AppHost` as the startup project.
2. Launch the project and wait for all the resources to finish starting up.
3. Click View (logs column) on the api resource.
4. Observe that the logs are shown once.
5. Click Resources (left hand nav)
6. Click View (logs column) for the api resource.
7. Observe that the logs are shown TWICE!
8. ... repeat.

If you hit refresh, the the correct data is displayed. What I think is happening is that when you jump between the log viewer and back to the make resources page the same LogViewer blazor component instance is being used and so the logentries field is getting the entries inserted multiple times.

To prove this, if you alternate the resources you navigate into and you can see that log entries between resources are bleeding into one another.

### Exceptions (if any)

_No response_

### .NET Version info

main branch

### Anything else?

_No response_

drewnoakes:
Fixes #4954, preventing duplicated log messages from appearing when navigating back to the previously displayed log.

Also encapsulate management of the base line number within the `LogEntries` class, and clear it along with log entries to ensure state doesn't leak between logs.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4957)


src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
```diff
@@ -104,6 +103,7 @@ internal async Task ClearLogsAsync()
     public async ValueTask DisposeAsync()
     {
         await _cancellationSeries.ClearAsync();
+        ViewModel.LogEntries.Clear();
```
JamesNK, line 106: Could you add a TODO here that this can be removed when https://github.com/dotnet/aspire/issues/4961 is done.

I think the solution here will be temporary and long term we can make this dependency transient so it doesn't matter if there is left over state.

adamint (approved):

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
adamint: @kvenkatrajan asked me to merge, so I went ahead and did this

adamint:
Flaky tests failing again...

adamint:
Different failed tests now, still flaky (unrelated to these changes)

PR #4965, opened 2024-07-18
Move ResourceLoggerForwarderService link to correct project

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4965)

### DIFF
--- a/src/Aspire.Hosting.Testing/ResourceLoggerForwarderService.cs
+++ b/src/Aspire.Hosting.Testing/ResourceLoggerForwarderService.cs
@@ -10,7 +10,7 @@ namespace Aspire.Hosting.Testing;
 /// <summary>
 /// A background service that watches resource logs and forwards them to the host's <see cref="ILogger"/> infrastructure.
 /// </summary>
-internal sealed class ResourceLoggerForwarderService(
+public sealed class ResourceLoggerForwarderService(
     ResourceNotificationService resourceNotificationService,
     ResourceLoggerService resourceLoggerService,
     IHostEnvironment hostEnvironment,
--- a/src/Aspire.Hosting.PostgreSQL/Aspire.Hosting.PostgreSQL.csproj
+++ b/src/Aspire.Hosting.PostgreSQL/Aspire.Hosting.PostgreSQL.csproj
@@ -15,7 +15,6 @@
   <ItemGroup>
     <Compile Include="$(SharedDir)StringComparers.cs" Link="Utils\StringComparers.cs" />
     <Compile Include="$(SharedDir)VolumeNameGenerator.cs" Link="Utils\VolumeNameGenerator.cs" />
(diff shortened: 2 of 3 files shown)


sebastienros:
I realized I made a mistake when I replicated it in another test.

tests/Aspire.Hosting.PostgreSQL.Tests/Aspire.Hosting.PostgreSQL.Tests.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
     <Compile Include="$(RepoRoot)src\Aspire.Hosting\ApplicationModel\UserSecretsParameterDefault.cs" Link="Utils\UserSecretsParameterDefault.cs" />
+    <Compile Include="$(RepoRoot)src\Aspire.Hosting.Testing\ResourceLoggerForwarderService.cs" Link="Utils\ResourceLoggerForwarderService.cs" />
```
eerhardt, line 17: Instead of this, can we reference Aspire.Hosting.Testing directly? That way we don't need to compile the source into the test. We can use the product.

sebastienros: It's internal right now, wasn't sure if we wanted to make it public (I assumed this would have been public in the first place).

sebastienros: Made it public

src/Aspire.Hosting.Testing/ResourceLoggerForwarderService.cs
```diff
@@ -10,7 +10,7 @@ namespace Aspire.Hosting.Testing;
 /// <summary>
 /// A background service that watches resource logs and forwards them to the host's <see cref="ILogger"/> infrastructure.
 /// </summary>
-internal sealed class ResourceLoggerForwarderService(
+public sealed class ResourceLoggerForwarderService(
```
eerhardt, line 13: @DamianEdwards @ReubenBond @mitchdenny - thoughts on making this public in `Aspire.Hosting.Testing`?

mitchdenny: To be honest I'd rather not make this API public at this point in time. I feel like if we were going to provide a mechanism to do this we'd probably have some kind of discrete interface that people could resolve to subscribe to log events or even just register a Logger<T> to forward the logs to.

I probably wouldn't expose the actual logging pump (the background service).

eerhardt: Can someone make an API proposal? This seems like a common scenario - we are using it in a bunch of hosting resource tests now.

eerhardt:
Superseeded by https://github.com/dotnet/aspire/pull/5193. Closing.

PR #4989, opened 2024-07-19
[release/8.1] Enable full signing and stable package versions

joperezr:
Getting ready for 8.1 release.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4989)

### DIFF
--- a/src/Aspire.Hosting.Azure.WebPubSub/Aspire.Hosting.Azure.WebPubSub.csproj
+++ b/src/Aspire.Hosting.Azure.WebPubSub/Aspire.Hosting.Azure.WebPubSub.csproj
@@ -6,6 +6,8 @@
     <PackageTags>aspire hosting azure webpubsub pubsub websocket</PackageTags>
     <Description>Azure WebPubSub resource types for .NET Aspire.</Description>
     <PackageIconFullPath>$(SharedDir)AzureWebPubSub_256x.png</PackageIconFullPath>
+    <!-- This library can't ship stable until Azure.Provisioning.WebPubSub ships stable. -->
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
     <!-- Disable package validation as this package hasn't shipped yet. -->
     <EnablePackageValidation>false</EnablePackageValidation>
   </PropertyGroup>
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -5,7 +5,7 @@
   <!-- Used by Aspire.Workload.Testing.targets -->
   <Target Name="GetPackageName" Returns="@(PackageName)">
     <ItemGroup>
-      <PackageName Include="$(MSBuildProjectName)" Condition="'$(IsPackable)' == 'true'" />
+      <PackageName Include="$(MSBuildProjectName)" Condition="'$(IsPackable)' == 'true'" WithPackageVersion="$(PackageVersion)" />
     </ItemGroup>
   </Target>
 </Project>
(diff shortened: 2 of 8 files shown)


src/Aspire.Hosting.Azure.WebPubSub/Aspire.Hosting.Azure.WebPubSub.csproj
```diff
@@ -6,6 +6,8 @@
     <PackageTags>aspire hosting azure webpubsub pubsub websocket</PackageTags>
     <Description>Azure WebPubSub resource types for .NET Aspire.</Description>
     <PackageIconFullPath>$(SharedDir)AzureWebPubSub_256x.png</PackageIconFullPath>
+    <!-- This library can't ship stable until Azure.Provisioning.WebPubSub ships stable. -->
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
eerhardt, line 10: FYI @vicancy

eerhardt (approved):

src/Directory.Build.targets
```diff
@@ -5,7 +5,7 @@
   <!-- Used by Aspire.Workload.Testing.targets -->
   <Target Name="GetPackageName" Returns="@(PackageName)">
     <ItemGroup>
-      <PackageName Include="$(MSBuildProjectName)" Condition="'$(IsPackable)' == 'true'" />
+      <PackageName Include="$(MSBuildProjectName)" Condition="'$(IsPackable)' == 'true'" WithPackageVersion="$(PackageVersion)" />
```
radical, line 8: Just curious, why is this change needed?

radical:
Is this branch meant for `release/8.0` instead of `main`?

src/Directory.Build.targets
joperezr: Because not all packages are going stable, so given this PR is stabilizing package versions, some will be like: `8.1.0` and some other swill be like `8.1.0-ci`. This change is ensuring that each project is returning the actual package version it will produce, as opposed to just assuming that all will have the same one.

joperezr:
Ohh great catch, yes retargeting.

src/Directory.Build.targets
radical: Ah, good idea!!

tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <ProductVersion>8.1.0</ProductVersion>
```
eerhardt, line 5: Why do we need to hard code this? Why can't it be generated from the other MSBuild properties?

joperezr: I tried locally using major and minor version but I think this project didn't import the right props and due to timing just wanted to unblock the build. End to End tests are failing now, so I guess now I have to fix than anyway.

joperezr: Looks like there are several assumptions that the testing infrastructure makes around package versions which may not be correct as they assume that the same package version will be used everywhere, which will certainly not always be the case. I just spent some time trying to fix another package testing issue I was seeing, but couldn't get the versions to flow correctly so due to time I just pushed another change which hardcodes the versions for now to unblock the build and be able to get a build out for validation. We should of course undo these hard coded hacks, but we should also fix these assumptions on versions. cc: @radical

eerhardt: Do we have an issue tracking this to fix it in `main`?

eerhardt:
@joperezr - Are you going to forward port any relevant changes here to `main`?

joperezr:
After release, we always merge back the release branch into main in order to make sure that changes that were done are not missed.

PR #5001, opened 2024-07-21
Add restart policy to resources

urumo:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5001)

This pull request implements restart policy for container based resources with Docker style restart annotations from issue #5003 .

Changes:
* Add `.WithRestart(RestartPolicy)` method to all container based projects..

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/IResourceWithRestartPolicy.cs
+++ b/src/Aspire.Hosting/ApplicationModel/IResourceWithRestartPolicy.cs
@@ -1,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents a resource with a restart policy.
+/// </summary>
+public interface IResourceWithRestartPolicy : IResource
+{
+}
--- a/src/Aspire.Hosting/Dcp/Model/Container.cs
+++ b/src/Aspire.Hosting/Dcp/Model/Container.cs
@@ -3,6 +3,7 @@
 
 using System.Net.Sockets;
 using System.Text.Json.Serialization;
+using Aspire.Hosting.ApplicationModel;
 using k8s.Models;
 
 namespace Aspire.Hosting.Dcp.Model;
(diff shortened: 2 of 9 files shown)


urumo:
@dotnet-policy-service agree

urumo:
I just have one question, in which file can I start implementing the behaviour of the new annotation?

Alirexaa:
Do we have an issue with this PR?

The restart policy should support DCP first, I don't know it is supported right now, but it's better to file an issue and discuss that first.

urumo:
> Do we have an issue with this PR?
Not at the moment of creating the PR per se, but I created #5003 
> The restart policy should support in DCP first, I don't know it is supported right now
As a matter of fact it is, which saved so much time while implementing the feature.

src/Aspire.Hosting/ApplicationModel/IResourceWithRestartPolicy.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents a resource with a restart policy.
+/// </summary>
+public interface IResourceWithRestartPolicy : IResource
```
davidfowl, line 9: I don't think we need this.

src/Aspire.Hosting/Dcp/Model/Container.cs
```diff
@@ -346,5 +348,28 @@ public static Container Create(string name, string image)
...
         || (this.Status?.State == ContainerState.FailedToStart && this.Status?.ContainerId is not null);
-}
 
+    private static string GetRestartPolicy(ResourceAnnotationCollection? containerAnnotations)
+    {
+        if (containerAnnotations is null)
+        {
+            return ContainerRestartPolicy.None;
+        }
+        var policy =
```
davidfowl, line 358: Why is this using the type name?

danmoseley:
@urumo I'm closing due to lack of update. Feel free to reopen to complete it. Thanks!

Issue #3185, opened 2024-03-26
Add more test coverage for the WithDataVolume and WithDataBindMount resource extension methods


DamianEdwards:
The `WithDataVolume` and `WithDataBindMount` resource extension methods added recently could do with more test coverage.

joperezr:
This one is dependent on https://github.com/dotnet/aspire/pull/4468. A big part of the work here is adding cleanup so volumes are cleaned up after test runs.

eerhardt:
I'm going to close this out as "completed". The only integration package that doesn't have these tests is Aspire.Hosting.Oracle, which is coming in https://github.com/dotnet/aspire/pull/4943. And half of them are disabled anyway because it doesn't work today. See https://github.com/dotnet/aspire/issues/5191.

PR #5002, opened 2024-07-21
Extract Aspire.Hosting.Nats.Tests project

Alirexaa:
Contributes to https://github.com/dotnet/aspire/issues/3185
Contributes to https://github.com/dotnet/aspire/issues/4294
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5002)


tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
```diff
@@ -0,0 +1,244 @@
...
+                                {
+                                    break;
+                                }
+                            }
+
+                            foreach (var item in events.Select((appEvent, i) => new { i, appEvent }))
+                            {
+                                Assert.Equal($"test-event-{item.i}", item.appEvent.Name);
+                                Assert.Equal($"test-event-description-{item.i}", item.appEvent.Description);
+                            }
```
eerhardt, line 206: This should verify the expected number came through. We should convert it to `for (var i = 0; i < 10; i++)`. That way if less than 10 come through, the test fails.

tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        using var host = hb.Build();
+
+        await host.StartAsync();
+
+        await pipeline.ExecuteAsync(async token =>
+        {
+            var natsConnection = host.Services.GetRequiredService<INatsConnection>();
+
+            var rtt = await natsConnection.PingAsync(token);
```
eerhardt, line 54: Should we do something a little more than ping it? Can we send and consume a message?

Alirexaa: I will do.

tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
eerhardt: That would help. But using a for loop would elimitate the Select and anonymous type that is happening here.

tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
```diff
@@ -0,0 +1,244 @@
...
+                volumeName = VolumeNameGenerator.CreateVolumeName(nats1, nameof(WithDataShouldPersistStateBetweenUsages));
+
+                // if the volume already exists (because of a crashing previous run), try to delete it
+                DockerUtils.AttemptDeleteDockerVolume(volumeName);
+                nats1.WithDataVolume(volumeName);
+            }
+            else
+            {
+                bindMountPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+                nats1.WithDataBindMount(bindMountPath);
```
radical, line 88: Just curious - `WithDataBindMount` is mounting to `/var/lib/nats` in the container. How is this test hitting that?

tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
eerhardt: I don't think this was resolved.

eerhardt (approved):
LGTM. Thanks again for the contributions!

Alirexaa:
> LGTM. Thanks again for the contributions!

it's pleasure.

Issue #4877, opened 2024-07-12
Allow mock AddProject calls in external Hosting extensions tests


eerhardt:
When creating an external Hosting extension library, and trying to write tests for it, it is useful to be able to add a mock `builder.AddProject<MockProject>("mockProject")` so you can test that the connection string of your Resource gets hooked up to the project correctly. We do this today in a couple of tests, Qdrant and Milvus.

https://github.com/dotnet/aspire/blob/3ddc4cacc5b23f68acd09307b5741e6760bd5fc7/tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs#L176-L184

https://github.com/dotnet/aspire/blob/3ddc4cacc5b23f68acd09307b5741e6760bd5fc7/tests/Aspire.Hosting.Tests/Qdrant/AddQdrantTests.cs#L314-L319

The issue is that while `IProjectMetadata` is public, it contains `internal` DIM'd properties 

https://github.com/dotnet/aspire/blob/3ddc4cacc5b23f68acd09307b5741e6760bd5fc7/src/Aspire.Hosting/IProjectMetadata.cs#L13-L23

This means that unless `Aspire.Hosting` has InternalsVisibleTo the test (which it won't if these extensions are outside this repo), the test can't fill in these "for testing" properties, and an exception occurs because we try to read the ProjectPath, which doesn't exist.

To workaround this, when adding the project the test can set `o => o.ExcludeLaunchProfile = true` which tells the hosting code to not try to load the launch profile. But we should have a better way for extension test authors to enable this scenario.

Options I can think of:
* We could make these properties public, which would mean making LaunchSettings public
* We could add a special API for tests to use.
    * `public interface ITestProjectMetadata : IProjectMetadata` or `IMockProjectMetadata` or similar
    * or `builder.AddMockProject("name")`
    * etc

cc @mitchdenny @davidfowl @sebastienros @radical @ReubenBond

mitchdenny:
For these scenarios couldn't we just use the pathy version of `AddProject` since it doesn't check if the files actually exist at this stage of processing?

captainsafia:
Bumping this issue up since some of the recent Azure Functions work has sparked new interest in this area.

We have a scenario in Azure Functions that requires us to do some special processing on the launch profile associated with the functions app to pull out the user-defined `--port` property and wire it up as an HTTP endpoint on the Aspire resource.

[...]

PR #5013, opened 2024-07-22
Extract Aspire.Hosting.Azure.Tests

eerhardt:
Contributes to #4294

This also allows us to remove the Aspire.Hosting.Tests.SharedShim, now that VolumeNameGenerator is only found in Aspire.Hosting.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5013)


tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
```diff
@@ -732,7 +732,7 @@ public async Task WithReferenceAppInsightsSetsEnvironmentVariable()
 
         appInsights.Resource.Outputs["appInsightsConnectionString"] = "myinstrumentationkey";
 
-        var serviceA = builder.AddProject<ProjectA>("serviceA")
+        var serviceA = builder.AddProject<ProjectA>("serviceA", o => o.ExcludeLaunchProfile = true)
```
radical, line 735: Just curious - why is this needed now?

tests/Aspire.Hosting.Azure.Tests/AzureSchemaTests.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xunit;
+
+namespace Aspire.Hosting.Azure.Tests;
+
+public class AzureSchemaTests
+{
+    [Fact]
+    public void ValidateApplicationSamples()
+    {
+        new SchemaTests().ValidateApplicationSamples("CdkResourceWithChildResource", (IDistributedApplicationBuilder builder) =>
```
radical, line 14: Why does this say `CdkResourceWithChildResource`? Will CDK be used here because of `PublishAsAzurePostgresFlexibleServer`?

And is `CDK` - AWS CDK being used for deploying to Azure here?

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
eerhardt: See https://github.com/dotnet/aspire/issues/4877

tests/Aspire.Hosting.Azure.Tests/AzureSchemaTests.cs
eerhardt: This is just moving the existing test in:

https://github.com/dotnet/aspire/blob/df935f4ee2e76d088235ea0fe91a0b61405cb3fe/tests/Aspire.Hosting.Tests/Schema/SchemaTests.cs#L110-L114

That test needed to be moved because Aspire.Hosting.Tests is no longer referencing the Aspire.Hosting.Azure libraries. So I moved the test to the Azure tests.

radical (approved):

joperezr:
@eerhardt no need to backport as release/8.1 branch is not expected to ship assets from it again. `main` branch is currently the 8.1.1 branch. Essentially `release/*` branches are mainly stabilization branches, as the intent is to "ship from `main`" each month. If the content is worthy enough for a minor version bump, then we will, otherwise we'll keep it as a patch update.

PR #5018, opened 2024-07-22
Add viewer for long strings in GridValue

### LINKED ISSUE #4412: Dashboard: Traces & Log messages - viewer for long fields
There is an initiative to use OTel to provide telemetry for LLM applications - recording the prompt and response text. These are typically long and can be plain text or json. When looking at these in the Aspire dashboard, in the trace and logs view, the details pane has a one line display with a "..." and button to copy to the dashboard. 

image

Visual Studio debugger has the same problem when its showing variables in the watch window. It has an option to pop out the value in a modal dialog, where text can be more easily examined. We should have something similar for the dashboard.

image

image

When the text is opened in the viewer, we should have a quick mechanism to sniff to see if it looks to be json, and then provide an option to format the content as json - as it typically will not have any formatting. The VS one has the ability to do base64, url encoding/decoding. Those are probably not needed here, but JSON and possibly XML formatting would be very useful.
If it fails to parse that would also be really useful to know, so spitting out the error in those cases would be good.

adamint:
Fixes #4412, and converts the copy button to a menu where the user can select whether to open a popup viewer or copy the cell text.

Menu appearance:
image

Dialog appearance:
image

Per spec, adds a plaintext, xml, and json formatting option. The initial formatting option is selected by trying to parse the string using an xml/json parser.

image

demo:
https://github.com/user-attachments/assets/5d0ef395-13f3-4e59-9a3f-9b6842469ca2

Normally, we would trigger popup opening just by adding an `OnClick` listener to the FluentMenuItems, but we are encountering a bug (probably in fluentui-blazor or fluent webcomponents) causing this to be swallowed sometimes when enclosed in `PropertyGrid`. @vnbaaij is aware, I am looking into this. A workaround is invoking a function from JS similar to how the copy button currently works. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5018)


mitchdenny:
This is nice. This might be relevant for DevTunnels support where we want to be able to render a QR code associated with a particular endpoint so people can point their mobile device at it.

tlmii:
Just curious: Do we have any indication how common it is that people just want to copy? Are we trading quick one-click access to that most-common task for two-clicks for all tasks? I'm fine accepting either "the two clicks aren't that bad" or "copy isn't that much more common than what we expect visualize will be" or even "we don't know and we'll see if users complain after we do this", I just want to make sure that's an intentional usability decision.

drewnoakes:
A minor point, but I feel like this icon represents filtering options rather than formatting options:

image

Maybe:
- image `Icons.Regular.Size20.PaintBrush()`
- image `Icons.Regular.Size20.TextEditStyle()`

JamesNK:
This looks good.

I haven't tested it yet, but from my first impression of the screenshots, here are some ideas to make it better:

1. Add a new, default option, called Auto. When the popup opens, we try to guess the value type. For example, we try to parse it as JSON, and if that succeeds then the value is automatically displayed as JSON. If that doesn't work (catch the exception) then we try to parse it as XML, and so on. If nothing works, then display as JSON. The default option would say something like "Auto (JSON)", depending on what we end up displaying. An auto option makes the user do less work to get a better result.
2. There should be a copy button on the popup. I could imagine someone options the popup to see the full value, and then decides they want to copy it to the clip board. We don't want to make them close the popup and then choose the copy value. Follow the same UI as the exception details popup in structured logs?
3. I think line numbers would be useful here. I'm guessing that currently long lines wrap (like they do in the console page) so they would help indicate when content has a new line vs content that is just long. A lot of the CSS from the console page could be reused as all the same display choices apply here.
4. I think there should be some differentiation in background color between the dialog and the scrollable content area. Lower priority.

JamesNK:
s

adamint:
> Add a new, default option, called Auto. When the popup opens, we try to guess the value type. For example, we try to parse it as JSON, and if that succeeds then the value is automatically displayed as JSON. If that doesn't work (catch the exception) then we try to parse it as XML, and so on. If nothing works, then display as JSON. The default option would say something like "Auto (JSON)", depending on what we end up displaying. An auto option makes the user do less work to get a better result.

This is already done by default when you open the dialog, do you feel that a separate option is important to have rather than preselecting?

> There should be a copy button on the popup. I could imagine someone options the popup to see the full value, and then decides they want to copy it to the clip board. We don't want to make them close the popup and then choose the copy value. The exception details popup has a copy button. Reuse that design?
If there is a copy button here, an interesting question is what gets copied? The formatted value (has new lines and indenting), or the original

I would assume the formatted value so as not to confuse the user, or both options?

adamint:
> Just curious: Do we have any indication how common it is that people just want to copy? Are we trading quick one-click access to that most-common task for two-clicks for all tasks? I'm fine accepting either "the two clicks aren't that bad" or "copy isn't that much more common than what we expect visualize will be" or even "we don't know and we'll see if users complain after we do this", I just want to make sure that's an intentional usability decision.

I don’t think we have any information on that. @kvenkatrajan

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,116 @@
...
+    private bool TryFormatJson()
+    {
+        try
+        {
+            using var doc = JsonDocument.Parse(
+                Content.Text,
+                new JsonDocumentOptions
+                {
+                    AllowTrailingCommas = true,
+                    CommentHandling = JsonCommentHandling.Skip
```
drewnoakes, line 80: I don't think we want to be skipping comments!

JamesNK:
> Just curious: Do we have any indication how common it is that people just want to copy? Are we trading quick one-click access to that most-common task for two-clicks for all tasks? I'm fine accepting either "the two clicks aren't that bad" or "copy isn't that much more common than what we expect visualize will be" or even "we don't know and we'll see if users complain after we do this", I just want to make sure that's an intentional usability decision.

I think we can mitigate this issue by only displaying the `...` menu option when we think there will be multiple lines. For example, the `Source` column on the resource grid is always an image name with one line. We should just display the copy button as we did before. But the details view for a resource should display `...` for its values because the value could be anything, e.g. JSON in an OTEL span's attribute value.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+            );
+
+            var memoryStream = new MemoryStream();
+            using var utf8JsonWriter = new Utf8JsonWriter(memoryStream, new JsonWriterOptions { Indented = true });
+            doc.WriteTo(utf8JsonWriter);
+            utf8JsonWriter.Flush();
+
+            _formattedText = Encoding.Default.GetString(memoryStream.ToArray());
```
drewnoakes, line 89: I don't think it's correct to use the default encoding here. The `Utf8JsonWriter` will write UTF8 bytes, which can only be decoded as UTF8.

drewnoakes, line 89: Did you consider using `JsonSerializer` here instead?

Something like:

```suggestion
            _formattedText = JsonSerializer.Serialize(doc.RootElement, new() { WriteIndented = true });
```

JamesNK:
> This is already done by default when you open the dialog, do you feel that a separate option is important to have rather than preselecting?

Nice, I didn't know that. I don't know which is best, I'll try it out and see what it's like. Preselecting like is already happening is probably fine. That's the important part.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    private void OnFormatOptionChanged()
+    {
+        if (_selectedOption == XmlFormat)
+        {
+            TryFormatXml();
+        }
+        else if (_selectedOption == JsonFormat)
+        {
+            TryFormatJson();
```
drewnoakes, line 108: If these "try" operations fail, should we set `_selectedOption` back to the item representing no formatting?

JamesNK:
> I would assume the formatted value so as not to confuse the user, or both options?

Let's go with copying the formatted value. If someone wants to copy without formatting, then they can select plaintext and then copy. We can change if there is feedback.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Aspire.Dashboard.Extensions;
 
-internal static class FluentUIExtensions
+internal static class FluentUiExtensions
```
drewnoakes, line 6: In .NET, two-letter acronyms are capitalized. So this change makes the name non-idiomatic.

Was there a reason for this change? If so, can we come up with a different name? If you really want to keep this name then the file name should be changed to match it.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
```diff
@@ -25,4 +25,20 @@ internal static class FluentUIExtensions
 
         return attributes;
     }
+
+    public static Dictionary<string, object> GetOpenTextVisualizerAdditionalAttributes(string textValue, string textValueDescription, params (string Attribute, object Value)[] additionalAttributes)
+    {
+        var attributes = new Dictionary<string, object>
```
drewnoakes, line 31: Key comparer?

JamesNK:
Here is an idea: color formatting of JSON and XML keywords. We could do it server side, but that would be a fair amount of work. There are JS libraries that do this that we could use:

https://github.com/highlightjs/highlight.js

We know the content type so we can tell the JS library for highlight as JSON for JSON, etc.

drewnoakes (commented):
LGTM though would be good to get a UI expert's input. I share Tim's view about the extra click for copy, so would be interested to know if we have additional options here.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
JamesNK: +1. Add some unit tests with various width unicode characters.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
```diff
@@ -0,0 +1,51 @@
...
+    <FluentStack VerticalAlignment="VerticalAlignment.Center">
+        <FluentIcon Value="@(new Icons.Regular.Size24.SlideSearch())" />
+        <FluentLabel Typo="Typography.PaneHeader">
+            @Content.Description
+        </FluentLabel>
+
+        <FluentButton id="@_openSelectFormatButtonId" slot="end"
+                      Style="margin-left: auto"
+                      Appearance="Appearance.Accent"
+                      IconEnd="@(new Icons.Regular.Size24.Filter())"
```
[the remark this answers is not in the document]
JamesNK: I agree. I think CodeTextEdit or SlideTextEdit would be good.

image

JamesNK:
Large content doesn't overflow with a scrollbar:
image

JamesNK:
Menu is cutoff when located in details view and the details view is narrow:

image

JamesNK:
I wrote a log with a fairly large JSON attribute (64kb minified). There is an error when trying to view it:

text-visualizer-json-error

JamesNK:
Add an example of formatted data to stress app: https://github.com/dotnet/aspire/pull/5021

JamesNK:
format-select

1. The button is currently always highlighted. In other places in the dashboard (e.g. resource type filter) the button is only highlighted if a non-default value is chosen. I think that should be done here as well. e.g. Plaintext = unhighlighted, anything else = highlighted
2. I think that clicking on an option should close the menu. There is a `MenuButton` control in FluentUI. I think it might be the right kind of control to use in this situation - https://www.fluentui-blazor.net/MenuButton
3. Nothing happens when clicking on an invalid option, e.g. selecting XML for non-XML content. I would have expected an error message. Since we're checking the content when the visualizer launches, what about disabling invalid options in the select, e.g. disable XML and JSON options if the content is a regular sentence. Then we don't need to worry about showing an error message.

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.cs
```diff
@@ -248,25 +252,46 @@ private void CloseMobileNavMenu()
...
             {
                 await JS.InvokeVoidAsync("window.unregisterGlobalKeydownListener", h);
             }
-            catch (JSDisconnectedException)
+
+            if (_textVisualizerHandler is not null)
             {
-                // Per https://learn.microsoft.com/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-7.0#javascript-interop-calls-without-a-circuit
-                // this is one of the calls that will fail if the circuit is disconnected, and we just need to catch the exception so it doesn't pollute the logs
+                await JS.InvokeVoidAsync("window.unregisterOpenTextVisualizerOnClick", _textVisualizerHandler);
```
JamesNK, line 287: Also dispose _textVisualizerHandler

leslierichardson95:
This looks pretty good for the most part!  Regarding the copying concern, how much space would we lose if we added the copy icon directly to the table inline with the vertical dots?

src/Aspire.Dashboard/wwwroot/js/app.js
```diff
@@ -299,3 +299,25 @@ window.listenToWindowResize = function(dotnetHelper) {
...
+
+    document.addEventListener('click', onClickListener);
+
+    return {
+        onClickListener: onClickListener,
+    }
+}
+
+window.unregisterOpenTextVisualizerOnClick = function (listener) {
+    document.removeEventListener('click', listener);
```
JamesNK, line 322: `registerOpenTextVisualizerOnClick` is returning an object with the function as the `onClickListener` property value.

`unregisterOpenTextVisualizerOnClick` is passed the object from the server, but the `onClickListener` property value isn't accessed to get the func to pass to removeEventListener.

It doesn't look like this would work to me.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
```diff
@@ -25,4 +25,20 @@ internal static class FluentUIExtensions
...
+        {
+            { "data-textvisualizer-text", textValue },
```
JamesNK, line 33: The copy button and the text visualizer button have separate attributes containing the text value. In case the data is very large, I'd like to combine them to one shared value.

JamesNK:
> This looks pretty good for the most part! Regarding the copying concern, how much space would we lose if we added the copy icon directly to the table inline with the vertical dots?

The size of the button.

I suggest you try the current change out. I found that two clicks to copy a value was fine because the menu is opened right next to the `...` button.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
adamint: Our choices are either `Skip` or `Disallow`, we don't have the option to allow. Calling this method with `Allow` will always result in a JsonException, it is not allowed. It is preferable to `Skip` as opposed to `Disallow`

adamint: `Note that, as of .NET Core 3.1, JsonDocument and JsonSerializer only support skipping or disallowing of comments; they do not support loading them. If you try to set JsonCommentHandling.Allow for either, you will get an exception:`

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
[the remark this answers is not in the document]
adamint: Changed to stealth and `SlideTextEdit`

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
adamint: I don't think so. The 'JSON' view for a non-JSON string is just the same as the unformatted view.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
adamint: I was not aware of the convention. Reverted.

adamint:
> This looks pretty good for the most part! Regarding the copying concern, how much space would we lose if we added the copy icon directly to the table inline with the vertical dots?

It would be the same width as the existing copy button, twice - we would then have a button to open the viewer and remove the menu

adamint:
> 1. The button is currently always highlighted. In other places in the dashboard (e.g. resource type filter) the button is only highlighted if a non-default value is chosen. I think that should be done here as well. e.g. Plaintext = unhighlighted, anything else = highlighted

Done.

> 2. I think that clicking on an option should close the menu. There is a `MenuButton` control in FluentUI. I think it might be the right kind of control to use in this situation - https://www.fluentui-blazor.net/MenuButton

I made it a popup with radio buttons based on @samsp-msft's mockup. @samsp-msft and @leslierichardson95 for design

> 3. Nothing happens when clicking on an invalid option, e.g. selecting XML for non-XML content. I would have expected an error message. Since we're checking the content when the visualizer launches, what about disabling invalid options in the select? For example, disable XML and JSON options if the content is a regular sentence. Then we don't need to worry about showing an error message.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
adamint: Changed to data-text

src/Aspire.Dashboard/wwwroot/js/app.js
adamint: fixed

adamint:
> Here is an idea: color formatting of JSON and XML keywords. We could do it server side, but that would be a fair amount of work. There are JS libraries that do this that we could use:
> 
> https://github.com/highlightjs/highlight.js
> 
> We know the content type so we can tell the JS library for highlight as JSON for JSON, etc.

This is a good idea, caveat is that the way console logs are written now, each line is its own individual span - making it a little more complicated. 

I (think) I've addressed all the feedback so far, except for the circuit disconnection issue

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
drewnoakes: That's a shame. I'd consider using `Disallow` then, so that no information is lost.

Alternatively, add an infobar to the UI that's shown when comments are removed. To detect this we could parse with `Disallow` then again with `Skip` if the first parse throws.

JamesNK:
> I made it a popup with radio buttons based on @samsp-msft's mockup. @samsp-msft and @leslierichardson95 for design

A popup that shows a list of radio buttons is weird. A button menu is the idiomatic way to do this.

JamesNK:
I don't think you've pushed all your changes. I'll take another look when everything is done and pushed to the PR.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -14,44 +16,53 @@ public partial class TextVisualizerDialog : ComponentBase
...
+        if (lines.Count > 0 && lines[0].Length == 0)
         {
-            _selectedOption = JsonFormat;
-            return;
+            lines.RemoveAt(0);
         }
 
-        _selectedOption = PlaintextFormat;
-        _formattedText = Content.Text;
+        return lines.Select((line, index) => new ResourceLogLine(index, line, false)).ToList();
```
danmoseley, line 65: aside -- in the JSON formatting (existing code below) it passes `CommentHandling = JsonCommentHandling.Skip` ... should that be .Allow? Are JSON comments lost when formatting JSON?

adamint: No, it is not allowed to be `Allow` - this will throw an exception in modern versions of .NET Core. JSON comments are unsupported, the only reason `Skip` is chosen over `Disallow` was to allow formatting of otherwise valid JSON.

adamint:
> I don't think you've pushed all your changes. I'll take another look when everything is done and pushed to the PR.

@JamesNK it is all pushed now. I have fixed the disconnection issue - maximum message size was hit - by streaming the string to the server, added highlighting (see below), which ended up being fairly complicated because each line is displayed individually (so I had to apply highlighting to each line), and added a copy to clipboard button in the dialog itself. I also added a few component bunit tests.

image

image

adamint:
hang on I realize I messed up the console log CSS

JamesNK:
Context menu is cut off and adds scroll bars:

image

@vnbaaij How do we make this popup menu appear above the splitter area?

JamesNK:
@adamint I pushed a change to format JSON using only the reader/writer. It allows comments to be preserved.

JamesNK:
This menu should show the currently selected option:

image

For example, in the resource's dropdown:

image

JamesNK:
The copy button always copies the unformatted value. I think it should copy what the user currently sees (formatted text for JSON and XML if that is the selected option).

JamesNK:
There is a gap at the start of content:

image

Is that intentional? I think line 0 should be at the top of the scrollable area.

JamesNK:
The formatted JSON value is wrong. In screen capture below you can see the end JSON isn't valid. Switching to unformatted fixes it. After that, it works.

value-popup-json-wrong-content

I double checked this wasn't a regression from my JSON formatting change to allow comments. It happens before and after my change.

JamesNK:
XML doc type is lost. I remember there was a comment from someone else about parsing content as a document rather than an element to preserve it. Was that change made?

value-popup-xml-doctype

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
danmoseley: If I'm inspecting a value in order to debug my app I might lose trust in the tool if I found it was not displaying all the content.

JamesNK: I fixed it here: [`fd2dc69` (#5018)](https://github.com/dotnet/aspire/pull/5018/commits/fd2dc694a3d073593b539330cb5bbdd37b0afdb2)

`JsonDocument` doesn't support comments. But the lower-level reader and writer does.

JamesNK:
Tests are needed.

* Unit tests for formatting JSON and XML. Double check it is indented, comments are retained, xml doc type is retained, multi-byte characters are handled correctly, very large data, etc.
* Consider adding component tests for TextVisualizerDialog. Checking the formatted value is correct (this issue - https://github.com/dotnet/aspire/pull/5018#issuecomment-2259442446) would be a useful thing to validate.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
```diff
@@ -0,0 +1,80 @@
...
+
+            @ControlsStringsLoc[nameof(ControlsStrings.GridValueCopyToClipboard)]
+        </FluentButton>
+    </div>
+</FluentDialogBody>
+
+<FluentDialogFooter Visible="false" />
+
+<link rel="stylesheet" href="css/highlight-11.10.0-default.min.css" />
+<script src="js/highlight-11.10.0.min.js"></script>
```
JamesNK, line 75: Is this JS file only called from the server? If so, it can be imported. That's the idiomatic way to add JS to a component.

If it is called from client JS then things are more complicated.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
```diff
@@ -0,0 +1,80 @@
...
+        </span>
+
+            @ControlsStringsLoc[nameof(ControlsStrings.GridValueCopyToClipboard)]
+        </FluentButton>
+    </div>
+</FluentDialogBody>
+
+<FluentDialogFooter Visible="false" />
+
+<link rel="stylesheet" href="css/highlight-11.10.0-default.min.css" />
```
JamesNK, line 74: Linking to a stylesheet inside a component is odd.

I think the content should either be in the TextVisualizerDialog.razor.css file (with a comment to say where it is from) or just add this to the main layout. Could improve in the future.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private string GetLogContentClass()
+    {
+        return $"log-content highlight-line language-{FormatKind}";
+    }
+
+    private ICollection<StringLogLine> GetLines()
+    {
+        var lines = Regex.Split(FormattedText, Environment.NewLine, RegexOptions.Compiled).ToList();
```
JamesNK, line 81: `Environment.NewLine` changes based on your environment (unsurprisingly). What about unexpected line endings?

I think formatted data will always use `Environment.NewLine` but unformatted data could have anything.

This is what AI generated when I asked it to split and handle different line endings:

```cs
public static string[] SplitIntoLines(string text)
{
    if (string.IsNullOrEmpty(text))
        return Array.Empty<string>();

    return text.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
}
```

Should have unit tests.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,230 @@
...
+        if (lines.Count > 0 && lines[0].Length == 0)
```
JamesNK, line 82: What is this for? Is there always an empty first line that needs to be removed or is it for certain data?

Add a comment about why it is needed.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.js
```diff
@@ -0,0 +1,30 @@
+let highlightObserver= new MutationObserver((mutations) => {
```
JamesNK, line 1: Import highlightjs file here. See app-metrics.js for an example of doing this.

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
```diff
@@ -25,4 +25,29 @@ internal static class FluentUIExtensions
...
+
+        foreach (var (attribute, value) in additionalAttributes)
+        {
+            attributes.Add(attribute, value);
+        }
+
+        return attributes;
+    }
+
+    private class AttributeKeyComparer : IEqualityComparer<string>
```
JamesNK, line 45: Why is this needed? Could https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparer.ordinal?view=net-8.0 be used?

Also, it could be added as another property to StringComparers and StringComparisons.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
JamesNK: I see it's called from the client. Ignore this comment and import in the client JS file: https://github.com/dotnet/aspire/pull/5018#discussion_r1697802200

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
adamint: Why can’t we only link inside the component? That way, the stylesheet is conditionally imported

JamesNK: CSS link in the page doesn't stop content from being displayed unstyled. There is the danger of a flash of unstyled content when doing this.

It's best practice to place CSS files in the page head so they are loaded before content is displayed.

JamesNK: https://stackoverflow.com/questions/1642212/whats-the-difference-if-i-put-css-file-inside-head-or-body

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
adamint: done

src/Aspire.Dashboard/Extensions/FluentUIExtensions.cs
adamint: Drew suggested it, I like using `StringComparer.Ordinal`. Changed

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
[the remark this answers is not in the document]
adamint: Changed

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
adamint: I've moved the CSS imports into `MainLayout.razor` because we have to conditionally import a dark or light theme in highlight.js depending on the effective theme. I also extended `ThemeManager` to obtain this effective theme on first render and subsequent theme changes.

Let me know if you can think of a better way

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
adamint: removed

adamint:
> The formatted JSON value is wrong. In screen capture below you can see the end JSON isn't valid. Switching to unformatted fixes it. After that, it works.
> 
> value-popup-json-wrong-content [ value-popup-json-wrong-content ](https://private-user-images.githubusercontent.com/303201/353656202-b5880aa9-92d2-464f-9222-8fe4dfadb60f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI1Mjg1OTIsIm5iZiI6MTcyMjUyODI5MiwicGF0aCI6Ii8zMDMyMDEvMzUzNjU2MjAyLWI1ODgwYWE5LTkyZDItNDY0Zi05MjIyLThmZTRkZmFkYjYwZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwODAxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDgwMVQxNjA0NTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYjdhMThlOTdhMTdiNTQyZjRmNTFmZDZlY2JkMjUwMTM5YzZlNWZkYjUzMjAyZjljMjZmZDUzNjdlYjI4ZDRkJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.kV-J5zLiMkuJDVJaMTQxtMMIyZmPOo41uOVlgVjN7UU) [ ](https://private-user-images.githubusercontent.com/303201/353656202-b5880aa9-92d2-464f-9222-8fe4dfadb60f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI1Mjg1OTIsIm5iZiI6MTcyMjUyODI5MiwicGF0aCI6Ii8zMDMyMDEvMzUzNjU2MjAyLWI1ODgwYWE5LTkyZDItNDY0Zi05MjIyLThmZTRkZmFkYjYwZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwODAxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDgwMVQxNjA0NTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYjdhMThlOTdhMTdiNTQyZjRmNTFmZDZlY2JkMjUwMTM5YzZlNWZkYjUzMjAyZjljMjZmZDUzNjdlYjI4ZDRkJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.kV-J5zLiMkuJDVJaMTQxtMMIyZmPOo41uOVlgVjN7UU)
> 
> I double checked this wasn't a regression from my JSON formatting change to allow comments. It happens before and after my change.

Yeah, I've checked and this is a `Virtualize` component bug @vnbaaij. The data passed into that component (from `GetLines()` is correct and matches the expected JSON). I guess the temporary solution is remove virtualization

adamint:
> The copy button always copies the unformatted value. I think it should copy what the user currently sees (formatted text for JSON and XML if that is the selected option).

Changed to formatted text

adamint:
> There is a gap at the start of content:
> 
> image
> 
> Is that intentional? I think line 0 should be at the top of the scrollable area.

It wasn't intentional, it came along with the console log viewer styling. Removed it

adamint:
> XML doc type is lost. I remember there was a comment from someone else about parsing content as a document rather than an element to preserve it. Was that change made? Does the writer also need to change to write the doc type.

Yes, that change was made, but seems like `XDocument.ToString` omits the doctype. I've made a change to write the XDocument to a stringwriter, which preserves the doctype but also adds encoding information, unfortunately.

adamint:
> The formatted JSON value is wrong. In screen capture below you can see the end JSON isn't valid. Switching to unformatted fixes it. After that, it works.
> 
> value-popup-json-wrong-content [ value-popup-json-wrong-content ](https://private-user-images.githubusercontent.com/303201/353656202-b5880aa9-92d2-464f-9222-8fe4dfadb60f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI1Mjg1OTIsIm5iZiI6MTcyMjUyODI5MiwicGF0aCI6Ii8zMDMyMDEvMzUzNjU2MjAyLWI1ODgwYWE5LTkyZDItNDY0Zi05MjIyLThmZTRkZmFkYjYwZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwODAxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDgwMVQxNjA0NTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYjdhMThlOTdhMTdiNTQyZjRmNTFmZDZlY2JkMjUwMTM5YzZlNWZkYjUzMjAyZjljMjZmZDUzNjdlYjI4ZDRkJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.kV-J5zLiMkuJDVJaMTQxtMMIyZmPOo41uOVlgVjN7UU) [ ](https://private-user-images.githubusercontent.com/303201/353656202-b5880aa9-92d2-464f-9222-8fe4dfadb60f.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI1Mjg1OTIsIm5iZiI6MTcyMjUyODI5MiwicGF0aCI6Ii8zMDMyMDEvMzUzNjU2MjAyLWI1ODgwYWE5LTkyZDItNDY0Zi05MjIyLThmZTRkZmFkYjYwZi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwODAxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDgwMVQxNjA0NTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYjdhMThlOTdhMTdiNTQyZjRmNTFmZDZlY2JkMjUwMTM5YzZlNWZkYjUzMjAyZjljMjZmZDUzNjdlYjI4ZDRkJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.kV-J5zLiMkuJDVJaMTQxtMMIyZmPOo41uOVlgVjN7UU)
> 
> I double checked this wasn't a regression from my JSON formatting change to allow comments. It happens before and after my change.

Fixed. I did not account for existing elements being changed instead of new ones being created from the virtualization.

adamint:
Highlighting now respects dark theme.

adamint:
> Tests are needed.
> 
> * Unit tests for formatting JSON and XML. Double check it is indented, comments are retained, xml doc type is retained, multi-byte characters are handled correctly, very large data, etc.
> * Consider adding component tests for TextVisualizerDialog. Checking the formatted value is correct (this issue - [Add viewer for long strings in GridValue #5018 (comment)](https://github.com/dotnet/aspire/pull/5018#issuecomment-2259442446)) would be a useful thing to validate.

The JS would have to be tested to have caught that issue, as the right content was being passed to the Virtualize component. Same with the SignalR issue - added a few more tests 😄

samsp-msft:
> Here is an idea: color formatting of JSON and XML keywords. We could do it server side, but that would be a fair amount of work. There are JS libraries that do this that we could use:
> 
> https://github.com/highlightjs/highlight.js
> 
> We know the content type so we can tell the JS library for highlight as JSON for JSON, etc.

Love it, but lets get this in first.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
samsp-msft: What happens if you have invalid JSON or xml? Is it only going to format if its valid? I almost want the errors if the user explicitly selects json, but it's not valid json.

tests/Aspire.Dashboard.Components.Tests/Controls/TextVisualizerDialogTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+namespace Aspire.Dashboard.Components.Tests.Controls;
+
+public class TextVisualizerDialogTests : TestContext
+{
+    [Fact]
+    public async Task Render_TextVisualizerDialog_WithValidJson_FormatsJsonAsync()
+    {
+        var rawJson = """
+                      // line comment
```
samsp-msft, line 20: Any reason why we are changing the comment type as part of the formatting?

samsp-msft:
nit: The line numbers start at 0, but I think most editors start at line 1.

samsp-msft:
nit: Can we move the copy to clipboard button to the top of the window - which will optimize space. It could go between the format dropdown and the close button.
image

There is also a bunch of blank space between the bottom of the text and that button.

samsp-msft:
This may be the same as the issue James noted above, but the menu can get lost behind other parts of the UI:

image

tests/Aspire.Dashboard.Components.Tests/Controls/TextVisualizerDialogTests.cs
JamesNK: JSON API doesn’t track different comment types. There isn’t anything we can easily do about it.

adamint:
> This may be the same as the issue James noted above, but the menu can get lost behind other parts of the UI:
> 
> image

I cannot think of a fix for this. @vnbaaij?

JamesNK:
> nit: Can we move the copy to clipboard button to the top of the window - which will optimize space. It could go between the format dropdown and the close button. image
> 
> There is also a bunch of blank space between the bottom of the text and that button. Moving the button to the top should be a much more efficient use of space.

I agree the amount of spacing seems to large. The current copy button position is consistent with another popup. I think we should leave it where it is (with less spacing) and consider the hi in both places later.

JamesNK:
Some UI thing I noticed: the dialog vertical size doesn’t shrink to match the content. It should take the whole screen to show one line. Also, there should be a max with on the dialog so it doesn’t become too wide.

adamint:
> Some UI things I noticed:
> 
> * The dialog vertical size doesn’t shrink to match the content. It should take the whole screen to show one line.
> * Also, there should be a max width on the dialog, so it doesn’t become too wide.

image

Max width is set to 600px.

JamesNK:
I think 1000px is the right spot. Enough room for most values, without growing too large on wide screens.

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.cs
```diff
@@ -248,27 +261,54 @@ private void CloseMobileNavMenu()
...
     }
 
+    [JSInvokable]
+    public async Task OpenTextVisualizerAsync(IJSStreamReference valueStream, string valueDescription)
+    {
+        var width = ViewportInformation.IsDesktop ? "75vw" : "100vw";
+        var parameters = new DialogParameters
+        {
+            Title = valueDescription,
+            Width = $"min(600px, {width})",
```
JamesNK, line 271:
```suggestion
            Width = $"min(1000px, {width})",
```

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
```diff
@@ -0,0 +1,77 @@
...
+                    </span>
+        @ControlsStringsLoc[nameof(ControlsStrings.GridValueCopyToClipboard)]
+    </FluentButton>
+</FluentDialogBody>
+
+<FluentDialogFooter Visible="false" />
+
+@code {
+    [Parameter, EditorRequired]
+    public required TextVisualizerDialogViewModel Content { get; set; }
```
JamesNK, line 76: Is there a reason to put the parameter in the razor file and not in the cs file? I've seen them in a number of razor files.

Code in a cs file offers better tooling experience compared to razor which is why I think it's better..

JamesNK:
I made style fixes. It's quicker for me to do it rather than communicate here. Also, there was a regression in the old console viewer that I fixed.

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.cs
```diff
@@ -76,6 +78,10 @@ protected override async Task OnInitializedAsync()
                 var newValue = ThemeManager.Theme!;
 
                 await _jsModule.InvokeVoidAsync("updateTheme", newValue);
+                var effectiveTheme = await _jsModule.InvokeAsync<string>("getCurrentTheme");
+                ThemeManager.EffectiveTheme = effectiveTheme;
+
+                await InvokeAsync(StateHasChanged);
```
JamesNK, line 84: Be careful with calling StateHasChanged and minimize how often it is used. I think in this case, it could be limited to just the new text dialog. Listening to theme change and calling state change in the dialog would be better than doing it globally.

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.cs
```diff
@@ -124,8 +130,15 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
...
             _shortcutManagerReference = DotNetObjectReference.Create(ShortcutManager);
+            _layoutReference = DotNetObjectReference.Create(this);
             _keyboardHandlers = await JS.InvokeAsync<IJSObjectReference>("window.registerGlobalKeydownListener", _shortcutManagerReference);
+            _textVisualizerHandler = await JS.InvokeAsync<IJSObjectReference>("window.registerOpenTextVisualizerOnClick", _layoutReference);
             ShortcutManager.AddGlobalKeydownListener(this);
+
+            // Get initial theme and then rerender the page to apply it.
+            await ThemeManager.RaiseThemeChangedAsync(await _jsModule.InvokeAsync<string>("getThemeCookieValue"));
+
+            await InvokeAsync(StateHasChanged);
```
JamesNK, line 141: What is this for? Performance of the initial page load is important. Re-rendering the page after it was rendered will have a negative impact.

JamesNK: I looked into what is happening here and see why you're doing this. For consistency with the rest of the app, theme switching should be (as much as possible) a client-side operation. The highlight styling was re-rending the page to add or remove CSS links is very different than what we're doing everywhere else.

I updated the code to use one stylesheet and control colors via variables.

Before I did this, I made a bunch of changes around getting the "effective" theme on the server. It turns out that's no longer required. I'm going to leave it in in case it is ever needed in the future.

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
```diff
@@ -0,0 +1,256 @@
...
+    protected override async Task OnAfterRenderAsync(bool firstRender)
+    {
+        if (firstRender)
+        {
+            _jsModule = await JS.InvokeAsync<IJSObjectReference>("import", "/Components/Dialogs/TextVisualizerDialog.razor.js");
+        }
+
+        if (_jsModule is not null && FormatKind != PlaintextFormat)
+        {
+            await _jsModule.InvokeVoidAsync("connectObserver", _logContainerId);
```
JamesNK, line 66: This looks like it can be called multipled times if the control re-renders. Is that ok?
Should the observer still be running if the format switches back to plain text?

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
adamint: No reason. Moved

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor.cs
adamint: > This looks like it can be called multipled times if the control re-renders. Is that ok?

It's necessary to re-highlight and re-observe if any text/the format has changed, so I'd say yes it's okay. However, the previous observer could be disconnected during subsequent calls. I made this change

> Should the observer still be running if the format switches back to plain text?

No, made that change.

JamesNK (approved):
The only major remaining issue is the interaction with splitters. It would be good to know a fix is coming in fluentui.

Issue #4005, opened 2024-04-29
[Bug] dashboard can't filter spanid log


gmij:
example:
![image](https://github.com/dotnet/aspire/assets/22893579/5c401a80-556b-45c3-9831-c35b0e1e13d1)
![image](https://github.com/dotnet/aspire/assets/22893579/9a3ac0aa-5bd4-4d33-abc6-c8653ca908bf)

for example. 
trace => detail => log => (auto filter spanid, but not logs ),

samsp-msft:
We are not 100% sure of what exactly you are asking for - can you please supply some details. Is this that you want to just see the logs for a specific span?

gmij:
Yes。 Only Show specific span Logs or trace log can group by span， Because one trace log too many.

kvenkatrajan:
Are you referring to maybe a log link on the span (like we have on the trace curently)? And clicking on it shows the logs only of that given span?

gmij:
> Are you referring to maybe a log link on the span (like we have on the trace curently)? And clicking on it shows the logs only of that given span?

Yes, specific span logs.

leidegre:
Right now, they way it works, is that when you drill into a specific trace/span you cannot find traces nested under it. I have a huge trace of 100s of activities/spans but I'm looking for specific spans here. There's no way for me to find it. I cannot even use the browser search function because of list view virtualization.

So there needs to be a very basic quick filter feature on this page, so that you can find spans/traces nested under other spans/traces. Ideally it should be possible to search for a trace/span (by activity name) regardless of where and how it has been nested. It makes sense to show the root trace first but if I search for something, I'd like to be able to find it. And also, if I want to look at a sub tree of trace/spans, like a scope to this feature, that would be ideal as well.

A large trace that brings together a lot of telemetry is very difficult to navigate. We're talking 700+ spans nested inside one big span.

kvenkatrajan:
@gmij - thanks makese sense.

@adamint / @JamesNK  - can either of you have a look please?

leslierichardson95:
Closing this issue since #5023 appears to have fixed the problem.

PR #5023, opened 2024-07-23
Update StructuredLogs to use LogFilter fields

aaron-hardin:
This should address #4005

Locally I had a similar issue where the 'View logs' button did not work on either the Trace or Span level. 
Once on the Structured logs page, selecting the correct field from the 'Edit filter' dialog would show 'log.traceid' and 'log.spanid' in the address bar.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5023)


JamesNK (approved):

JamesNK:
Thanks for the fix.

PR #4950, opened 2024-07-18
Update to Azure.AI.OpenAI 2.0.0-beta.2

eerhardt:
This allows users to use the new client library.

Fix #4466
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4950)

PR #5033, opened 2024-07-23
Marked shipped API and set new PackageValidation version

joperezr:
Now that 8.1 has shipped, this is marking all of the relevant APIs as shipped as well as enabling package validation on new packages and setting the right baseline version.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5033)


src/Components/Aspire.Azure.Messaging.EventHubs/Aspire.Azure.Messaging.EventHubs.csproj
```diff
@@ -10,7 +10,7 @@
     <PackageIconFullPath>$(SharedDir)AzureEventHubs_256x.png</PackageIconFullPath>
     <NoWarn>$(NoWarn);SYSLIB1100;SYSLIB1101</NoWarn>
     <!-- This package hasn't shipped stable, so package validation runs against previously shipped version of it. -->
-    <PackageValidationBaselineVersion>8.0.1-preview.8.24267.1</PackageValidationBaselineVersion>
+    <PackageValidationBaselineVersion>8.1.0-preview.1.24373.2</PackageValidationBaselineVersion>
```
eerhardt, line 13: Why does this one need this, but the others that haven't shipped stable don't?

src/Aspire.Hosting.Elasticsearch/PublicAPI.Shipped.txt
```diff
@@ -1 +1,11 @@
 #nullable enable
+Aspire.Hosting.ApplicationModel.ElasticsearchResource
```
eerhardt, line 2: This one hasn't shipped stable yet.

joperezr: Oh, good catch, thought I removed all of the non stable ones.

src/Components/Aspire.Azure.Messaging.EventHubs/Aspire.Azure.Messaging.EventHubs.csproj
joperezr: All of the ones that haven't shipped stable should need it. If not I'm expecting build will fail.

src/Aspire.Hosting.AWS/Aspire.Hosting.AWS.csproj
```diff
@@ -8,7 +8,7 @@
     <PackageTags>aspire hosting aws</PackageTags>
     <Description>Add support for provisioning AWS application resources and configuring the AWS SDK for .NET.</Description>
     <!-- This package hasn't shipped stable, so package validation runs against previously shipped version of it. -->
-    <PackageValidationBaselineVersion>8.0.1-preview.8.24267.1</PackageValidationBaselineVersion>
+    <PackageValidationBaselineVersion>8.1.0-preview.1.24373.2</PackageValidationBaselineVersion>
```
eerhardt, line 11: I don't want to block the rest of this PR, but I think we shouldn't do package validation until we actually ship stable. (Or at least, don't do public API validation in the package validation)

The reasoning is that we are free to change the public APIs until the library ships stable. Doing package validation just gets in the way and confuses people when they try to change the APIs. For example, in https://github.com/dotnet/aspire/pull/4950/, I needed to disable validation because I was updating to the new Azure.AI.OpenAI dependency.

eerhardt (approved):
The change looks good to go.

I do want to start a discussion on whether we do package validation on preview packages.

src/Aspire.Hosting.AWS/Aspire.Hosting.AWS.csproj
joperezr: > I needed to disable validation because I was updating to the new Azure.AI.OpenAI dependency.

Instead of disabling validation, you can alternatively just pass the property `/p:GenerateCompatibilitySuppressionFile=true` as indicated [here](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/diagnostic-ids#suppression-file) in order to generate a suppression file.

joperezr:
This is now stale and has been superseded. closing.

PR #5041, opened 2024-07-24
Disabling AOAI resource key access by default.

### LINKED ISSUE #5040: Azure OpenAI resources deploy with keys enabled but use Managed Identity
Similar to [Azure Storage should disable keys by default](https://github.com/dotnet/aspire/issues/4713), but this time with Azure OpenAI resources.

We can disable this by setting `DisableLocalAuth` to `true` on `CognitiveServicesAccountProperties`.

aaronpowell:
Updated test for bicep changed and added test for property override.

Fixes #5040

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5041)


tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
```diff
@@ -2781,6 +2781,121 @@ public async Task AddAzureOpenAI()
...
+            }
+
+            output connectionString string = 'Endpoint=${cognitiveServicesAccount_wXAGTFUId.properties.endpoint}'
+            
+            """;
+        output.WriteLine(manifest.BicepText);
+        Assert.Equal(expectedBicep, manifest.BicepText);
+    }
+
+    [Fact]
```
eerhardt, line 2840: Would it make sense to turn the above test into a Theory instead? There appears to only be 2 lines different in these 2 tests.

eerhardt (commented):
The product change looks good to me. I just have one question about the test code.

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
aaronpowell: I'm happy to convert it to a Theory over a Fact. I tend to shy away from Theory if they result in conditional logic within the test itself, but that's more personal preference, so I'll defer to whatever is preferred by the aspire team.

eerhardt: > if they result in conditional logic within the test itself

1 or 2 "if" checks vs. a hundred lines of duplication seems like a good tradeoff to me.

aaronpowell: I'll refactor the test and update the PR

aaronpowell: Test has been updated to use a theory.

eerhardt (approved):

PR #4695, opened 2024-06-27
Setting minimum TLS to 1.2 by default

aaronpowell:
Fixes #4692

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4695)

PR #5043, opened 2024-07-24
Setting a default TLS version for Azure SQL to 1.2

### LINKED ISSUE #5042: Azure SQL deploys without setting TLS version
Similar to [Azure Storage should provision with TLS 1.2 by default](https://github.com/dotnet/aspire/issues/4692). It should set a TLS version of at least 1.2 to avoid S360 errors.

aaronpowell:
Updated tests and added tests to validate user-specified override

Fixes #5042

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5043)


mitchdenny (approved):

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
```diff
@@ -28,6 +28,7 @@ internal static IResourceBuilder<SqlServerServerResource> PublishAsAzureSqlDatab
             sqlServer.AssignProperty(x => x.Administrators.Sid, construct.PrincipalIdParameter);
             sqlServer.AssignProperty(x => x.Administrators.Login, construct.PrincipalNameParameter);
             sqlServer.AssignProperty(x => x.Administrators.TenantId, "subscription().tenantId");
+            sqlServer.AssignProperty(x => x.MinimalTlsVersion, "'1.3'");
```
eerhardt, line 31: Why are we using 1.3 here, but https://github.com/dotnet/aspire/pull/4695 set the minimum to 1.2?

aaronpowell: Azure Storage doesn't support TLS 1.3 (see https://learn.microsoft.com/azure/storage/common/transport-layer-security-configure-minimum-version).

The reason I was setting 1.3 here was defaulting to the latest supported version, not minimal consistency.

eerhardt: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.SQL.MinTLS/ and https://learn.microsoft.com/en-us/azure/azure-sql/database/connectivity-settings?view=azuresql&tabs=azure-portal#minimum-tls-version says the recommended minimum is to set it to 1.2.

The learn docs even have a caution:

image

aaronpowell: Cool, wasn't aware of that (I should read the docs properly 🤣).

I'll drop it to 1.2 as the min for Azure SQL

eerhardt: I pushed an update to change it to 1.2. I'd like to get this into the 8.2 release, which is getting snapped from main tomorrow.

aaronpowell: I was just about to update myself, you beat me to it

eerhardt (approved):

Issue #5055, opened 2024-07-24
Sql Server `WithDataBindMount` Failed To Start


Alirexaa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Adding SqlServer using `WithDataBindMount` failed to start.
I use Windows 10 and think SqlServer has some issues with mount volume to the Windows host (see: https://stackoverflow.com/a/66238175).

### Expected Behavior

The container starts without any errors.

### Steps To Reproduce

Just add `SqlServerServerResource` using `WithDataBindMount`

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var sqlserver = builder.AddSqlServer("sqlserver")
.WithDataBindMount("C:\\sqldata");

builder.Build().Run();
```

### .NET Version info

.NET SDK:
 Version:           8.0.303
 Commit:            29ab8e3268
 Workload version:  8.0.300-manifests.280f8661
 MSBuild version:   17.10.4+10fbfbf2e

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19045
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.303\

.NET workloads installed:
 [android]
   Installation Source: SDK 8.0.300, VS 17.10.34916.146
   Manifest Version:    34.0.95/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.android\34.0.95\WorkloadManifest.json
   Install Type:        FileBased

 [aspire]
   Installation Source: SDK 8.0.300, VS 17.10.34916.146
   Manifest Version:    8.0.1/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.aspire\8.0.1\WorkloadManifest.json
   Install Type:        FileBased

 [ios]
   Installation Source: SDK 8.0.300, VS 17.10.34916.146
   Manifest Version:    17.2.8053/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.ios\17.2.8053\WorkloadManifest.json
   Install Type:        FileBased

 [maccatalyst]
   Installation Source: SDK 8.0.300, VS 17.10.34916.146
   Manifest Version:    17.2.8053/8.0.100
[...]

PR #4943, opened 2024-07-17
Extract Aspire.Hosting.Oracle.Tests

sebastienros:
Contributes to https://github.com/dotnet/aspire/issues/4294

Fixes #3161

Tests disabled because they don't work on the CI machines, c.f. https://github.com/dotnet/aspire/issues/5362
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/4943)

PR #5056, opened 2024-07-24
Extract Aspire.Hosting.SqlServer.Tests project

Alirexaa:
Contributes to https://github.com/dotnet/aspire/issues/3185
Contributes to https://github.com/dotnet/aspire/issues/4294
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5056)


Alirexaa:
WithDataShouldPersistStateBetweenUsages(useVolume: true) work on my local but WithDataShouldPersistStateBetweenUsages(useVolume: false) failed because of https://github.com/dotnet/aspire/issues/5055.
For now, I disabled this test to discuss it.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,284 @@
...
+        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
+        var pipeline = new ResiliencePipelineBuilder()
+            .AddRetry(new() { MaxRetryAttempts = 10, Delay = TimeSpan.FromSeconds(10) })
+            .Build();
+
+        try
+        {
+            var builder1 = CreateDistributedApplicationBuilder();
+
+            var password = "P@ssw0rd1";
```
sebastienros, line 86: Don't set passwords like this or this will be blocked on our internal CI. Check the other db functional tests, there is a way to read the default generated value.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,284 @@
...
+                // Use a deterministic volume name to prevent them from exhausting the machines if deletion fails
+                volumeName = VolumeNameGenerator.CreateVolumeName(sqlserver1, nameof(WithDataShouldPersistStateBetweenUsages));
+
+                // If the volume already exists (because of a crashing previous run), try to delete it
+                DockerUtils.AttemptDeleteDockerVolume(volumeName);
+                sqlserver1.WithDataVolume(volumeName);
+            }
+            else
+            {
+                bindMountPath = Directory.CreateTempSubdirectory().FullName;
```
sebastienros, line 106: The permissions issue might be due to this method. Try creating a standard temporary folder for the cases when this one fails. This method explicitly removes some permissions which might be necessary in this case.

Alirexaa: I did, But still not working.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,285 @@
...
+        try
+        {
+            var builder1 = CreateDistributedApplicationBuilder();
+
+            var password = "p@ssw0rd1";
+
+            var passwordParameter = builder1.AddParameter("pwd");
+            builder1.Configuration["Parameters:pwd"] = password;
+
+            var sqlserver1 = builder1.AddSqlServer("sqlserver", passwordParameter);
```
sebastienros, line 91: The default logic creates a random password. Here we just get its value to reuse later.

```suggestion
            var sqlserver1 = builder1.AddSqlServer("sqlserver");
            var password = sqlserver1 .Resource.PasswordParameter.Value;
```

sebastienros:
Something I was asked to do, and maybe you can take [what I did as an example](https://github.com/dotnet/aspire/pull/4943), if to also use the EF component in this functional test to extend the coverage.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -71,5 +71,21 @@ public static IResourceBuilder<SqlServerServerResource> WithDataVolume(this IRes
     /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<SqlServerServerResource> WithDataBindMount(this IResourceBuilder<SqlServerServerResource> builder, string source, bool isReadOnly = false)
-        => builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
+    {
+        // c.f. https://learn.microsoft.com/sql/linux/sql-server-linux-docker-container-configure?view=sql-server-ver15&pivots=cs1-bash#mount-a-host-directory-as-data-volume
```
eerhardt, line 75: Why is this done only for `WithDataBindMount` and not for `WithDataVolume`?

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
```diff
@@ -31,4 +29,8 @@
     <PackageReference Include="Polly.Core" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="SqlServer\" />
+  </ItemGroup>
+
```
eerhardt, line 35:
```suggestion
```

tests/Aspire.Hosting.SqlServer.Tests/TestDbContext.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public DbContextOptions<TestDbContext> Options { get; }
+
+    public DbSet<Car> Cars => Set<Car>();
+
+    public class Car
+    {
+        public int Id { get; set; }
+        public string Brand { get; set; } = default!;
```
eerhardt, line 22:
```suggestion
        public required string Brand { get; set; }
```

Is this better than shutting off the nullable warning?

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,318 @@
...
+            selectCommand.CommandText = $"SELECT 1";
+            var results = await selectCommand.ExecuteReaderAsync(token);
+
+            Assert.True(results.HasRows);
+        }, cts.Token);
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false, Skip = "https://github.com/dotnet/aspire/issues/5055")]
```
eerhardt, line 122: It seems concerning that we are changing the `WithDataBindMount` implementation in this PR, but then also shutting off the test that checks `WithDataBindMount`.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,318 @@
...
+        using var app = builder.Build();
+
+        await app.StartAsync(cts.Token);
+
+        var hb = Host.CreateApplicationBuilder();
+
+        hb.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
+        {
+            [$"ConnectionStrings:{tempDb.Resource.Name}"] = await tempDb.Resource.ConnectionStringExpression.GetValueAsync(default),
+        });
```
eerhardt, line 46:
```suggestion
        hb.Configuration[$"ConnectionStrings:{tempDb.Resource.Name}"] = await tempDb.Resource.ConnectionStringExpression.GetValueAsync(default);
```

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,318 @@
...
+        await dbContext.SaveChangesAsync(cts.Token);
+
+        var cars = await dbContext.Cars.ToListAsync(cts.Token);
+        Assert.Single(cars);
+        Assert.Equal("BatMobile", cars[0].Brand);
+    }
+
+    [Fact]
+    [RequiresDocker]
+    public async Task VerifySqlServerResource()
```
eerhardt, line 74: Is it possible to test both EF and non-EF in the same test? That way we don't have to spin up the docker container twice.

tests/Aspire.Hosting.SqlServer.Tests/Aspire.Hosting.SqlServer.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" />
```
eerhardt, line 8:
```suggestion
```

I don't think this is necessary, since `Aspire.Hosting.Tests` brings it in.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
sebastienros: Binding specific folders only applies to folder mounts, not volumes. Volumes just work fine locally without changes, and the documentation only uses it for folders mounts too.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
sebastienros: It is concerning but it has never been enabled on the CI, and this would have to be the same for other containers: Oracle, MongoDB. I have tried many things without success, setting Unix permissions, creating the folder differently, ... Nothing works on the CI even though it works locally, both with a Windows or WSL2 test runner.

I will try on one of the perf machines which has native Linux (same 22.04 as the CI) in case I can repro the problem.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
eerhardt: @rwestMSFT - I see you wrote the documentation being referenced here in https://github.com/MicrosoftDocs/sql-docs/commit/51dddc6d9c9fd8fdeb9f78fec3172089c28bd5d7. Is there a reason we need to split all 3 directories when binding to a host directory, but when using a docker volume you just need a single volume? Why doesn't it work to map the single top-level `/var/opt/mssql` directory to the host machine?

cc @JerryNixon

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,291 @@
...
+            {
+                bindMountPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+                Directory.CreateDirectory(bindMountPath);
+
+                sqlserver1.WithDataBindMount(bindMountPath);
+
+                if (!OperatingSystem.IsWindows())
+                {
+                    // Change permissions for non-root accounts (container user account)
+                    System.Diagnostics.Process.Start("/usr/bin/env", $"sudo chmod -R a+rwx {bindMountPath}").WaitForExit();
```
eerhardt, line 130: Should be able to use `File.SetUnixFileMode` here. Much easier than starting a new process.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
rwestMSFT: @eerhardt My name appears on thousands of articles. Please refer to the `author` and metadata when reviewing SQL Docs content.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,297 @@
...
+                if (!OperatingSystem.IsWindows())
+                {
+                    // Change permissions for non-root accounts (container user account)
+                    const UnixFileMode OwnershipPermissions =
+                       UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.UserExecute |
+                       UnixFileMode.GroupRead | UnixFileMode.GroupWrite | UnixFileMode.GroupExecute |
+                       UnixFileMode.OtherRead | UnixFileMode.OtherWrite | UnixFileMode.OtherExecute;
+
+                    File.SetUnixFileMode(bindMountPath, OwnershipPermissions);
+                    //System.Diagnostics.Process.Start("/usr/bin/env", $"sudo chmod -R a+rwx {bindMountPath}").WaitForExit();
```
eerhardt, line 136:
```suggestion
```

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        await pipeline.ExecuteAsync(async token =>
+        {
+            var dbContext = host.Services.GetRequiredService<TestDbContext>();
+            var databaseCreator = (RelationalDatabaseCreator)dbContext.Database.GetService<IDatabaseCreator>();
+            await databaseCreator.CreateTablesAsync(token);
+        }, cts.Token);
+
+        var dbContext = host.Services.GetRequiredService<TestDbContext>();
+        var databaseCreator = (RelationalDatabaseCreator)dbContext.Database.GetService<IDatabaseCreator>();
```
eerhardt, line 77: Can we do what the existing test was doing? That seems simpler:

```
await dbContext.Database.EnsureCreatedAsync();
```

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,296 @@
...
+
+public class SqlServerFunctionalTests(ITestOutputHelper testOutputHelper)
+{
+    [Fact]
+    [RequiresDocker]
+    public async Task VerifySqlServerResource()
+    {
+        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+        var pipeline = new ResiliencePipelineBuilder()
+            .AddRetry(new() { MaxRetryAttempts = 10, Delay = TimeSpan.FromSeconds(10) })
```
eerhardt, line 29: Is 10 seconds too long to wait in between retries? How long does it typically take to spin up?

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
eerhardt: @amitkh-msft - Looks like you are the "author" of the article referenced above. Any chance you can answer the above question?

amitkh-msft: Ack. Sure, please give me some time. I will look at this and try to answer the question.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -0,0 +1,281 @@
...
+                    {
+                        await conn.OpenAsync(token);
+                        Assert.Fail("Waiting for database to be down");
+                    }
+                    catch
+                    {
+                        // Failing means the database is correctly down
+                        return;
+                    }
+                }, cts.Token);
```
eerhardt, line 190: I'm not sure I totally get what this chunk of code is trying to do.

Either the connection open will throw, or the next line `Assert.Fail` will throw. So either way an exception will occur, and that exception will be caught and eaten, and the code returns sucessfully.

Why do we have a "pipeline.ExecuteAsync" around this code? It will never retry.

eerhardt (approved):
I think my only real concern is the product change to `WithDataBindMount`. I don't understand why the 3 directories are needed, as opposed to just 1 root level directory.

But we can follow up later around this, if needed.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
sebastienros: Right, the Fail is in the wrong place, I fixed the pattern somewhere else and missed this one.

We want the exception to be thrown if OpenAsync succeeds, so outside of the try/finally. The goal is to wait for the server to not respond to requests anymore.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
eerhardt: Any thoughts @amitkh-msft? Why can't we just bind mount to the root directory instead of the 3 separate ones?

PR #5058, opened 2024-07-24
[tests] Work around sqlserver 2022-latest latest image breaking with testcontainers

radical:
[tests] Workaround sqlserver 2022-latest latest image failing

.. with testcontainers because the path `/opt/mssql-tools/bin/sqlcmd`
changed to `/opt/mssql-tools18/bin/sqlcmd`. But this hardcoded path is
being used in testcontainers, causing it to fail.

https://github.com/testcontainers/testcontainers-dotnet/issues/1220
https://github.com/microsoft/mssql-docker/issues/892

Issue: https://github.com/dotnet/aspire/issues/5057

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5058)

Issue #1997, opened 2024-01-31
Container-like resource image tag management


mitchdenny:
We have a number of resources that use container images for the local inner loop (and also for production deployments via manifest). One of the challenges we have is how to approach which specific image is being used. In most cases this is captured in the `ContainerImageAnnotation`. To change the container image used you need to mutate/replace this annotation.

It would be good to provide a helper extension method on container-like resources to make this easier. For things that directly derive from `ContainerResource` (most non-cloud specific resources) we could just have a `WithImageTag(...)` method. This would take an enumeration such as `ImageTag.Latest` which when interpretted by the extension would set the tag to a `latest` tag.

Some containers might have more nuanced tagging strategies and so the extension method could be make more explicit for specific resources to make this mechanism easy for developers (with the escape hatch mechanism staying in place).

Related to all this is how we decide to update the default version. It might be that we always want to float to the latest, but tell customers that best practice is to pin to a particular tag once they have got things running to minimize disruption caused by accidental major version upgrades.

There are security implications of taking the latest and there are security implications of not picking up updates. Additionally we don't have a mechanism at the moment to force a pull of a container image from an upstream source. Once it is in the local Docker daemon cache - perhaps this is something that DCP should manage (cache refreshes).

/cc @davidfowl @karolz-ms @blowdart @joperezr

karolz-ms:
In my opinion, for development time we should allow users to use ANY image they want for a given component, and not just restrict the customization ability to tag changes. There are many scenarios where using a custom image is beneficial or required. For example, it might be a company policy to only used "blessed" images from company image registry. Or the developer might want to add some additional debugging tools to the image.

Granted, this is somewhat more advanced scenario and there might be issues with custom images, but that is not a reason strong enough to not allow completely custom images.

I agree that using specific image version/tag, instead of "latest", generally leads to less trouble.

As for stale image checks, we have prior art for that in our VS/VS Code container extensions. We could have the AppHost or DCP could do a similar staleness check in the background and emit warnings during the application run if stale images are detected.

eerhardt:
I think we should make 3 changes here:

1. Allow the developer to pick the tag.
    - As discussed above with an API, or using the AppHost configuration to map container resource names to tags.
2. Allow the developer to pick the registry.
    - Some enterprises don't allow devs to use images from docker hub and want their own registry for security reasons. Just like how we don't use nuget.org in our repo.
    - I imagine us having this at 2 levels:
[...]

PR #5059, opened 2024-07-24
Use digest for sqlserver and fix tests

joperezr:
An update to mssql/server image tag `2022-latest` broke our tests as we can no longer communicate with the container. Chatting offline with our experts from the dotnet-docker repo, they suggested we should instead depend on digests as opposed to tags, to be able to guarantee determinism, and then just make sure we update these digests in a frequent basis to ensure we get fixes for them. This PR is switching just the sqlserver hosting component to use digest instead and it also does that for the tests so that they pass again, but we will likely want to make an overhaul pass to all of our hosting components to similarly depend on digests as opposed to tags.

cc: @mitchdenny @radical @eerhardt 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5059)

### DIFF
--- a/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
+++ b/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
@@ -27,7 +27,8 @@ public static class SqlServerBuilderExtensions
         var sqlServer = new SqlServerServerResource(name, passwordParameter);
         return builder.AddResource(sqlServer)
                       .WithEndpoint(port: port, targetPort: 1433, name: SqlServerServerResource.PrimaryEndpointName)
-                      .WithImage(SqlServerContainerImageTags.Image, SqlServerContainerImageTags.Tag)
+                      .WithImage(SqlServerContainerImageTags.Image)
+                      .WithImageSHA256(SqlServerContainerImageTags.Digest)
                       .WithImageRegistry(SqlServerContainerImageTags.Registry)
                       .WithEnvironment("ACCEPT_EULA", "Y")
                       .WithEnvironment(context =>
--- a/src/Aspire.Hosting.SqlServer/SqlServerContainerImageTags.cs
+++ b/src/Aspire.Hosting.SqlServer/SqlServerContainerImageTags.cs
@@ -8,5 +8,5 @@ internal static class SqlServerContainerImageTags
     public const string Registry = "mcr.microsoft.com";
     public const string Image = "mssql/server";
     // Tracking tag: 2022-latest. NOTE: latest  digest fails the tests. We need to investigate.
-    public const string Digest = "sha256:c4369c38385eba011c10906dc8892425831275bb035d5ce69656da8e29de50d8";
+    public const string Digest = "c4369c38385eba011c10906dc8892425831275bb035d5ce69656da8e29de50d8";
 }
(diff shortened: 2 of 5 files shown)


src/Aspire.Hosting.SqlServer/SqlServerContainerImageTags.cs
```diff
@@ -7,5 +7,6 @@ internal static class SqlServerContainerImageTags
 {
     public const string Registry = "mcr.microsoft.com";
     public const string Image = "mssql/server";
-    public const string Tag = "2022-latest";
+    // Tracking tag: 2022-latest. NOTE: latest  digest fails the tests. We need to investigate.
+    public const string Digest = "sha256:c4369c38385eba011c10906dc8892425831275bb035d5ce69656da8e29de50d8";
```
radical, line 11: Just curious - how did you get this?

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -27,7 +27,8 @@ public static IResourceBuilder<SqlServerServerResource> AddSqlServer(this IDistr
         var sqlServer = new SqlServerServerResource(name, passwordParameter);
         return builder.AddResource(sqlServer)
                       .WithEndpoint(port: port, targetPort: 1433, name: SqlServerServerResource.PrimaryEndpointName)
-                      .WithImage(SqlServerContainerImageTags.Image, SqlServerContainerImageTags.Tag)
+                      .WithImage(SqlServerContainerImageTags.Image)
+                      .WithImageSHA256(SqlServerContainerImageTags.Digest)
```
eerhardt, line 31: > they suggested we should instead depend on digests as opposed to tags, to be able to guarantee determinism

Note that this is OK to unblock our tests, but this doesn't comply with our long term plan.

https://github.com/dotnet/aspire/issues/3933
https://github.com/dotnet/aspire/issues/1997#issuecomment-1924592452

If

joperezr: Let's have a follow up discussion just to finalize on our plan going forward. We may very well end up following the same plan we have, I just want to make sure we take into consideration the concept of determinism and the fact that we have now been broken by this, so we should check what our customers would expect us to do.

For this particular case, it's worth noting that the underlying image doesn't follow semver  which seems to be one of the keys of our plan.

src/Aspire.Hosting.SqlServer/SqlServerContainerImageTags.cs
joperezr: It wasn't straight forward as this is a past tag, and docker doesn't have a way of tracking that. It's easy to check the current digest of an image (tag) using `docker inspect`, but not the history. In this case, I checked the image supported tags at https://mcr.microsoft.com/en-us/product/mssql/server/tags, and saw that `2022-latest` points to the same one as `2022-CU14-ubuntu-22.04` and noticed there is also a `2022-CU13-ubuntu-22.04` which I assumed was the previous one and when running the tests it worked.

joperezr:
Closing in favor of https://github.com/dotnet/aspire/pull/5058 where @radical was able to keep the tests enabled without the need to use digests. We should still figure out how to fix this not only for our tests, but also for our hosting component.

PR #5080, opened 2024-07-25
Prepare playground apps to allow testing

radical:
This PR has changes to prepare the `playground/` apps to be run as part of future tests in `Aspire.Playground.Tests`. The changes follow these broad patterns:

- remove all the per-project `Directory.Build.{props,targets}`, and instead of these ..
    - .. add root `Directory.Build.{props,targets}` and use `Aspire.RepoTesting.*`
    - And update all the playground projects to use `AspireProjectOrPackageReference` instead of `ProjectReference`
- `Aspire.Hosting.Testing` expects to identify AppHost assemblies by a name ending in `.AppHost`.
    - rename some projects that don't follow that convention.

- Update the apps to use `Aspire.RepoTesting*` which allows building outside of repo.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5080)


radical:
TODO
- [ ] check that the playground apps are still runnable from VS

playground/AWS/AWS.AppHost/AWS.AppHost.csproj
```diff
@@ -13,10 +13,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\..\src\Aspire.Dashboard\Aspire.Dashboard.csproj" />
-    <ProjectReference Include="..\..\..\src\Aspire.Hosting.AWS\Aspire.Hosting.AWS.csproj" IsAspireProjectResource="false" />    
-    <ProjectReference Include="..\..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" IsAspireProjectResource="false" />
-    
+    <ProjectReference Include="..\..\..\src\Aspire.Dashboard\Aspire.Dashboard.csproj" Condition="'$(BuildForTest)' != 'true'" />
```
eerhardt, line 16: Do you think we could automatically inject this ProjectReference in a common .targets file? That way each playground project doesn't have to.

playground/AWS/AWS.AppHost/AWS.AppHost.csproj
```diff
@@ -13,10 +13,10 @@
...
+    <AspireProjectOrPackageReference Include="Aspire.Hosting.AWS" IsAspireProjectResource="false" />
```
eerhardt, line 17: Similar comment here for `IsAspireProjectResource="false"`. We should only need this when it is a `ProjectReference`. So when we create the `AspireProjectOrPackageReference` => `ProjectReference`, we should be able to inject this metadata.

playground/mongo/Mongo.AppHost/Mongo.AppHost.csproj
```diff
@@ -13,12 +13,15 @@
...
 
   <ItemGroup>
-    <ProjectReference Include="..\..\..\src\Aspire.Dashboard\Aspire.Dashboard.csproj" />
-    <ProjectReference Include="..\..\..\src\Aspire.Hosting.Azure\Aspire.Hosting.Azure.csproj" IsAspireProjectResource="false" />
-    <ProjectReference Include="..\..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" IsAspireProjectResource="false" />
-    <ProjectReference Include="..\..\..\src\Aspire.Hosting.MongoDB\Aspire.Hosting.MongoDB.csproj" IsAspireProjectResource="false" />
-        
+    <ProjectReference Include="..\..\..\src\Aspire.Dashboard\Aspire.Dashboard.csproj" Condition="'$(BuildForTest)' != 'true'" />
+
+    <!-- <AspireProjectOrPackageReference Include="Aspire.Hosting" IsAspireProjectResource="false" /> -->
```
eerhardt, line 18: Do we need this?

playground/Directory.Build.props
```diff
@@ -9,6 +13,8 @@
       CS1712: Type parameter 'type_parameter' has no matching typeparam tag in the XML comment on 'type_or_member' (but other type parameters do)
     -->
     <NoWarn>$(NoWarn),1573,1591,1712</NoWarn>
+
+    <SharedDir Condition="'$(SharedDir)' == ''">$(MSBuildThisFileDirectory)..\src\Shared\</SharedDir>
```
eerhardt, line 17: Why isn't this already set from the root Directory.Build.props?

eerhardt (commented):
This looks awesome. I had a couple suggestions to make it even cleaner. Then I think we can merge it.

playground/Directory.Build.props
radical: This is to cover the case when `playground` is being built from outside the repo, and thus doesn't have that root Directory.Build.props .
I'm not super happy with this, and some others we'll need for playground, but I don't overdo that at least just yet!

eerhardt: Is it needed for more than KnownResourceNames.cs?

Looks like that only contains a single const string

https://github.com/dotnet/aspire/blob/11e90df77649e3abeab69b37912b6fd7f449f50a/src/Shared/KnownResourceNames.cs#L8

Maybe we just copy that string into a `playground/KnownResourceNames.cs` file. It's just a simple string. Easier to copy it once than to try to copy around files at build time.

radical: btw, there will be a bunch of copying around done on CI when archiving the playground tests. But this one is clearly a one-off, so I pushed the change as suggested.

eerhardt (approved):
Nice!

PR #5083, opened 2024-07-26
Preserve selected instrument when switching resources

JamesNK:
Change metrics page to preserve the selected instrument when switching resources if the new resource also has that instrument.

For example, `http.client.request.duration` is selected in App1. Switching to App2 keeps `http.client.request.duration` selected if App2 also has values. If App2 doesn't have `http.client.request.duration` then the instrument is cleared (current behavior).

The goal is to make switching between resources more seamless.

---

Also, I wrote a bunit test for the metrics page that tests this change.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5083)


src/Aspire.Dashboard/Components/Controls/ResourceSelect.razor.cs
```diff
@@ -32,6 +32,11 @@ public partial class ResourceSelect
 
     private FluentSelect<SelectViewModel<ResourceTypeDetails>>? _resourceSelectComponent;
 
+    private static void ValuedChanged(string value)
+    {
+        // Do nothing. Required for bunit change to trigger SelectedOptionChanged.
+    }
```
JamesNK, line 38: @vnbaaij I'm confused why this is needed. It's required for the other change event to trigger via:

```cs
        var resourceSelect = cut.FindComponent<ResourceSelect>();
        var innerSelect = resourceSelect.Find("fluent-select");
        innerSelect.Change("TestApp2");
```

adamint (approved):

Issue #5047, opened 2024-07-24
Check List: Validate arguments of public methods


Zombach:
### Is your feature request related to a problem? Please describe the problem.

Check List to [#2649](https://github.com/dotnet/aspire/issues/2649)
Source [CA1062](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062)

Cause
An externally visible method dereferences one of its reference arguments without verifying whether that argument is null (Nothing in Visual Basic).

You can [configure](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#configure-code-to-analyze) this rule to exclude certain types and parameters from analysis. You can also [indicate null-check validation methods](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#null-check-validation-methods).

Rule description
All reference arguments that are passed to externally visible methods should be checked against null. If appropriate, throw an [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) when the argument is null.

If a method can be called from an unknown assembly because it is declared public or protected, you should validate all parameters of the method. If the method is designed to be called only by known assemblies, mark the method internal and apply the [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute) attribute to the assembly that contains the method.

### Describe the solution you'd like

I suggest covering this with additional tests.
Check every reference argument in the public API  `ArgumentNullException.ThrowIfNull(source)`.

Using the example of Aspire.Hosting.Redis
RedisBuilderExtensions.cs
```cs
public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicationBuilder builder, string name, int? port = null)
{
    ArgumentNullException.ThrowIfNull(builder);

    var redis = new RedisResource(name);
    return builder.AddResource(redis)
                  .WithEndpoint(port: port, targetPort: 6379, name: RedisResource.PrimaryEndpointName)
                  .WithImage(RedisContainerImageTags.Image, RedisContainerImageTags.Tag)
                  .WithImageRegistry(RedisContainerImageTags.Registry);
}

[Fact]
public void AddRedisContainerShouldThrowsWhenBuilderIsNull()
{
    IDistributedApplicationBuilder builder = null!;
    const string name = "Redis";

    var action = () => builder.AddRedis(name);

[...]

Issue #2649, opened 2024-03-05
Review use of ArgumentException/ArgumentNullException in all public APIs


DamianEdwards:
I noticed we seem to have lots of public APIs that accept reference type parameters not marked as nullable but then don't check that they're actually non-null and throw an `ArgumentNullException` when necessary. We should decide what we're doing and review all public APIs for appropriate argument validation.

e.g. https://github.com/dotnet/aspire/blob/240cde74a83cc5378a90c2b6eeb8d65f43658d80/src/Aspire.Hosting/Extensions/ResourceBuilderExtensions.cs#L92-L98

mitchdenny:
@JamesNK your PR mentioned this issue, do you think we've got 100% coverage here now from your #2809 PR?

JamesNK:
Probably not. And there have been new APIs since then.

PR #5099, opened 2024-07-29
Adding public API test coverage for Aspire.Hosting.Qdrant

Zombach:
Initial issues [#2649](https://github.com/dotnet/aspire/issues/2649)
Issues [Check List: Validate arguments of public methods](https://github.com/dotnet/aspire/issues/5047)

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5099)


tests/Aspire.Hosting.Qdrant.Tests/QdrantPublicApiTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        Assert.Multiple(() =>
+        {
+            var exception = Assert.Throws<ArgumentNullException>(action);
+            Assert.Equal(nameof(builder), exception.ParamName);
+        });
+    }
+
+    [Fact]
+    public void WithDataBindMountShouldThrowsWhenNameIsNull()
```
radical, line 73: This is testing the `source` (source directory) parameter, and not `name`.

tests/Aspire.Hosting.Qdrant.Tests/QdrantPublicApiTests.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        var distributedApplicationBuilder = DistributedApplication.CreateBuilder([]);
+        const string name = "Qdrant";
+        var apiKeyParameter = ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(distributedApplicationBuilder, $"{name}-Key", special: false);
+        var resource = new QdrantServerResource(name, apiKeyParameter);
+        var builder = distributedApplicationBuilder.AddResource(resource);
+        string source = null!;
+
+        var action = () => builder.WithDataBindMount(source);
```
radical, line 82: Do we need all this or would something like this be enough?:
```csharp
        IResourceBuilder<QdrantServerResource> builder = null!;
        string source = null!;
        var action = () => builder.WithDataBindMount(source);
```

@eerhardt @sebastienros thoughts?

sebastienros: I agree, it doesn't need to be that complicated. Even the `Multiple` usage in all these tests, since the second assertion depends on the first, we don't care about the second result if the first one failed.

Zombach: > Multiple

Indeed, I will remove Multiple.

Zombach: > Do we need all this or would something like this be enough?:
> 
> ```cs
>         IResourceBuilder<QdrantServerResource> builder = null!;
>         string source = null!;
>         var action = () => builder.WithDataBindMount(source);
> ```
> 
> @eerhardt @sebastienros thoughts?

But in this case. We will not handle the error when builder 'is not null' and source 'is null'

sebastienros: Full test

```c#
var builder = TestDistributedApplicationBuilder.Create();
var qdrant = builder.AddQdrant("qdrant");
string source = null!;

var action = () => qdrant.WithDataBindMount(source);
Assert.Equal(nameof(source), exception.ParamName);
```

sebastienros (approved):

tests/Aspire.Hosting.Qdrant.Tests/QdrantPublicApiTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Utils;
+using Xunit;
+
+namespace Aspire.Hosting.Qdrant.Tests;
+
+public class QdrantPublicApiTests
+{
+    #region QdrantBuilderExtensions
```
eerhardt, line 12:
```suggestion
```

We don't use regions in this repo.

eerhardt (approved):

PR #5104, opened 2024-07-29
DRAFT: Pubsub eventing model in app model.

mitchdenny:
This is a spike on introducing a pub/sub eventing model in the app model. At the moment I've built some infrastructure and used it to bridge the existing lifecycle hook subscriber infrastructure (so lifecycle hooks are just implemented on top of this new model).

Some explanatory notes:

1. Events all implement `ILifecycleEvent`. This allows for some generic constraints in various places to support resolution of event subscribers based on the event type they subscribe to.
2. There is a single "publisher" API ... `ILifecycleEventPublisher`. The purpose of the implementor of this is to take an event and make sure it finds its way to all the subscribers.
3. Dispatchers (e.g. `AfterEndpointsAllocatedEventDispatcher` which implements `ILifechcleEventDispatcher<TLifecycleEvent>`. This is provided by the code that defines an event type. The purpose of this type is to allow some processing of subscribers before sending the event to them. For example if you want to route an event to a specific subscriber because that subscriber is explicitly wired to dispatch events for a particular resource then this optimization can happen here. I'm not 100% on this abstraction yet other than I think we'll need something like it.
4. ILifecycleEventSubscriber<TLifecycleEvent>; implemented by subscribers to events and injected into DI.

Here is the way someone would start publishing a new event:

1. Define a class that implements `ILifecycleEvent`.
2. Define a "handler" that implements `ILifecycleEventDispatcher<T>` and register it to DI - add any custom processing logic to `DispatchAsync(...)`.
3. Define a custom subscriber interface that derives from `ILifecycleEventSubscriber<T>` and adds any additional properties. For example if you wanted a subscriber model that worked per-resource you would expose a resource property on this subscriber which is then implemented by the subscribers concrete implementation. The dispatcher logic could then query that and make sure that the event is routed to the right place without having to call every single subscriber.

We could probably optimize this a bit by having some base implementations of these boilerplate bits.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5104)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -62,7 +62,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                                           ILogger<DistributedApplication> distributedApplicationLogger,
                                           DistributedApplicationModel model,
                                           IKubernetesService kubernetesService,
-                                          IEnumerable<IDistributedApplicationLifecycleHook> lifecycleHooks,
+                                          ILifecycleEventPublisher lifecycleEventPublisher,
                                           IConfiguration configuration,
                                           DistributedApplicationOptions distributedApplicationOptions,
                                           IOptions<DcpOptions> options,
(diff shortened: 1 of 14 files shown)


mitchdenny:
I'm thinking it might actually be better for resource-specific subscriptions to be an annotation with a callback and the callback would have a context object (the event) which provides access to the resource and if applicable the `IServiceProvider`.

mitchdenny:
Closing this PR off as we probably won't end up using this approach (better approach is #5223

PR #5110, opened 2024-07-29
Adding public API test coverage for Aspire.Hosting.Python

Zombach:
Initial issues [#2649](https://github.com/dotnet/aspire/issues/2649)
Issues [Check List: Validate arguments of public methods](https://github.com/dotnet/aspire/issues/5047)

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5110)


src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
```diff
@@ -60,6 +60,8 @@ public static class PythonProjectResourceBuilderExtensions
     public static IResourceBuilder<PythonProjectResource> AddPythonProject(
         this IDistributedApplicationBuilder builder, string name, string projectDirectory, string scriptPath, params string[] scriptArgs)
     {
+        ArgumentNullException.ThrowIfNull(builder);
```
radical, line 63: Maybe add for `scriptArgs` also.

tests/Aspire.Hosting.Python.Tests/PythonPublicApiTests.cs
```diff
@@ -0,0 +1,319 @@
...
+using Xunit;
+
+namespace Aspire.Hosting.Python.Tests;
+
+public class PythonPublicApiTests
+{
+    #region PythonProjectResource
+
+    [Fact]
+    public void CtorPythonProjectResourceShouldThrowsWhenNameIsNull()
```
radical, line 13: nit: Instead of `ShouldThrows` using `ShouldThrow` would be better.

src/Aspire.Hosting.Python/PythonProjectResourceBuilderExtensions.cs
Zombach: > Maybe add for `scriptArgs` also.

We do this check on line 118.
And since we don’t use it directly in line 64, the check is not needed in this case.
But I can add a check for the others if you think it's necessary!

tests/Aspire.Hosting.Python.Tests/PythonPublicApiTests.cs
```diff
@@ -0,0 +1,277 @@
...
+    public void CtorPythonProjectResourceShouldThrowWhenProjectDirectoryIsNull()
+    {
+        const string name = "Python";
+        const string executablePath = "/src/python";
+        string projectDirectory = null!;
+
+        var action = () => new PythonProjectResource(name, executablePath, projectDirectory);
+
+        var exception = Assert.Throws<ArgumentNullException>(action);
+        Assert.Equal("workingDirectory", exception.ParamName);
```
radical, line 48: `PythonProjectResource`'s parameter is named `projectDirectory` which is then passed to the `ExecutableResource..ctor` to the parameter named `workingDirectory`.

And because the exception is being thrown from `ExecutableResource` we get `workingDirectory` instead of `projectDirectory`. That sounds confusing for the user. Maybe we should have the null check on the .ctor arguments for `PythonProjectResource` directly.

@mitchdenny @eerhardt @sebastienros Is there a pattern that we follow in such cases?

Zombach: To be honest, I would add a check, it would be more transparent.
But due to the fact that in PythonProjectResource these arguments are not used at the moment. according to CA1062, we can skip this check, since it is performed in depth.
But if someone adds interaction with arguments to PythonProjectResource, then they will have to add a check for 'null'.
That's why I'm in favor of adding it right away.

eerhardt: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types

> ✔️ DO set the ParamName property when throwing one of the subclasses of ArgumentException.

> This property represents the name of the parameter that caused the exception to be thrown.

I read this as the ParamName should be set to the name of the parameter that was invalid. In this case `projectDirectory`.

mitchdenny: I think we need to do the null check in the PythonResource ctor just for clarity. I'm not a fan of the ThrowIfNull method, seems a bit odd to do it that way.

mitchdenny: On second thought it probably doesn't matter to the end result.

mitchdenny (approved):

mitchdenny:
@Zombach thanks once again for all your work on some of these PRs. Appreciate your efforts.

PR #5119, opened 2024-07-30
Adding public API test coverage for Aspire.Hosting.Elasticsearch

Alirexaa:
Contributes to #2649 and #5047
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5119)


src/Aspire.Hosting.Elasticsearch/ElasticsearchBuilderExtensions.cs
```diff
@@ -44,6 +44,9 @@ public static class ElasticsearchBuilderExtensions
         IResourceBuilder<ParameterResource>? password = null,
         int? port = null)
     {
+        ArgumentNullException.ThrowIfNull(name);
+        ArgumentNullException.ThrowIfNull(builder);
+
```
Zombach, line 49: This is purely my opinion, but I would change the order so that there is a sequence similar to the incoming parameters. builder, name

src/Aspire.Hosting.Elasticsearch/ElasticsearchResource.cs
```diff
@@ -26,6 +26,7 @@ public class ElasticsearchResource : ContainerResource, IResourceWithConnectionS
     /// <param name="password">A parameter that contains the Elasticsearch superuser password.</param>
     public ElasticsearchResource(string name, ParameterResource password) : base(name)
     {
+        ArgumentNullException.ThrowIfNull(name);
```
Zombach, line 29: ContainerResource has name checking. 
And since we do not use the name in the bottom context, checking here for 'null' is not needed at the moment. If in the future, in this constructor, the use of a name appears, then the need for 'ThrowIsNull' will appear

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchPublicApiTests.cs
```diff
@@ -0,0 +1,116 @@
...
+{
+    [Fact]
+    public void AddElasticsearchContainerShouldThrowsWhenBuilderIsNull()
+    {
+        IDistributedApplicationBuilder builder = null!;
+        const string name = "Elasticsearch";
+
+        var action = () => builder.AddElasticsearch(name);
+
+        Assert.Multiple(() =>
```
Zombach, line 19: Based on this [comment](https://github.com/dotnet/aspire/pull/5099#discussion_r1696261977). Assert.Multyple is not necessary.

You should just leave it here:
Assert.Equal(nameof(builder), exception.ParamName);

Same for the code below

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchPublicApiTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            var exception = Assert.Throws<ArgumentNullException>(action);
+            Assert.Equal(nameof(name), exception.ParamName);
+        });
+    }
+    [Fact]
+    public void CtorElasticsearchResourceShouldThrowsWhenPasswordIsNull()
+    {
+        var builder = new DistributedApplicationBuilder([]);
```
Zombach, line 104: Why is builder here?

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchPublicApiTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+using Xunit;
+
+namespace Aspire.Hosting.Elasticsearch.Tests;
+
+public class ElasticsearchPublicApiTests
+{
+    [Fact]
+    public void AddElasticsearchContainerShouldThrowsWhenBuilderIsNull()
```
Zombach, line 12: Sorry my typo. It needs to be fixed
ShouldThrow'S' => ShouldThrow

Same for the code below

sebastienros:
@Zombach Small request or for whoever merges them, can you name them with a more meaningful title, something like "Adding public API test coverage for Aspire.Hosting.Elasticsearch". CA1062 is quite cryptic when you are not familiar with these PRs.

Zombach (approved):

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchPublicApiTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        var action = () => builder.AddElasticsearch(name);
+
+        var exception = Assert.Throws<ArgumentNullException>(action);
+        Assert.Equal(nameof(builder), exception.ParamName);
+    }
+
+    [Fact]
+    public void AddElasticsearchContainerShouldThrowWhenNameIsNull()
+    {
+        IDistributedApplicationBuilder builder = new DistributedApplicationBuilder([]);
```
Zombach, line 26: As I found out. Better to use TestDistributedApplicationBuilder.Create();

Alirexaa: AFIK `TestDistributedApplicationBuilder` is suitable when starting the app. Here, it seems unnecessary.

Zombach: > AFIK `TestDistributedApplicationBuilder` is suitable when starting the app. Here, it seems unnecessary.

I wanted this more from the point of view that there would be a single code base for checking the “public zero”.
Although you may be right. View this at your own discretion. Well, or ask the guys what they think about this

Alirexaa:
@sebastienros, this PR is ready to merge.

mitchdenny (approved):

PR #5123, opened 2024-07-30
Redirect to homepage if login page is requested when token auth is not enabled

### LINKED ISSUE #3775: Token login page should redirect to root if token auth is not configured
If for some reason a user tries to access the `/login` page when token auth is not configured (either because of mismatch between IDE/workload/Apphost packageref or just because they enter it manually) the page will display but nothing they do will work.

We should consider redirecting to `/` from the login page if auth is not configured for token auth.

adamint:
Looping in a wider range of reviewers because this is modifying a sensitive code path. Fixes #3775 by redirecting if the requested path equals `/login` and auth mode is not `token`, including when the token query parameter (`t`) is not included in the URL.

demo, token auth NOT enabled:

https://github.com/user-attachments/assets/5ff07fa4-85d8-43d6-a7fb-9099a22bfe1a

demo, token auth enabled (login page still shows and works as normal)

https://github.com/user-attachments/assets/48c1e978-70d5-48e1-bb39-8ad447d70e3c

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5123)

### DIFF
--- a/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
+++ b/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
@@ -65,6 +65,7 @@ internal sealed class ValidateTokenMiddleware
 
         await _next(context).ConfigureAwait(false);
     }
+
     private static void RedirectAfterValidation(HttpContext context)
     {
         if (context.Request.Query.TryGetValue("returnUrl", out var returnUrl))
--- a/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
+++ b/src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
@@ -26,9 +26,15 @@ internal sealed class ValidateTokenMiddleware
 
     public async Task InvokeAsync(HttpContext context)
     {
-        if (context.Request.Path.Equals("/login", StringComparisons.UrlPath) && context.Request.Query.TryGetValue("t", out var value))
+        if (context.Request.Path.Equals("/login", StringComparisons.UrlPath))
         {
-            if (_options.CurrentValue.Frontend.AuthMode == FrontendAuthMode.BrowserToken)
+            if (_options.CurrentValue.Frontend.AuthMode != FrontendAuthMode.BrowserToken)
+            {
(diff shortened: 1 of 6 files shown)


JamesNK (changes requested):
Test please. Can either be an integration test or you could test the middleware on its own.

adamint:
> Test please. Can either be an integration test or you could test the middleware on its own.

Tested middleware in isolation

https://github.com/dotnet/aspire/pull/5123/files#diff-7a2aaaa04e1d12692092b7e6f57a966eb5450d85638bce7c0158ed615c93bad1

tests/Aspire.Dashboard.Tests/Middleware/ValidateTokenMiddlewareTests.cs
```diff
@@ -0,0 +1,107 @@
...
+                    })
+                    .Configure(app =>
+                    {
+                        app.UseMiddleware<ValidateTokenMiddleware>();
+                    });
+            })
+            .StartAsync();
+    }
+
+    private static IOptionsMonitor<DashboardOptions> GetOptionsMonitorMock(FrontendAuthMode authMode, string expectedToken)
```
JamesNK, line 88: Change this to set options normally and remove NSubstitute dependency.

Should be able to do something like this:

```cs
services.AddOptions();
services.Configure<DashboardOptions>(o =>
{
    // configure dashboard options
});
```

src/Aspire.Dashboard/Model/ValidateTokenMiddleware.cs
```diff
@@ -62,14 +61,21 @@ public async Task InvokeAsync(HttpContext context)
...
             }
-            else
-            {
-                _logger.LogDebug($"Request to validate token URL but auth mode isn't set to {FrontendAuthMode.BrowserToken}.");
-            }
         }
 
         await _next(context).ConfigureAwait(false);
     }
+    private static void RedirectAfterValidation(HttpContext context)
```
JamesNK, line 68:
```suggestion

    private static void RedirectAfterValidation(HttpContext context)
```

JamesNK (approved):

Issue #4884, opened 2024-07-13
Add more supported UI cultures for Aspire Dashboard


martincostello:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

After I opened #4882 it occurred to me that there are probably more cultures where the preferred formatting for dates is different to the neutral culture, as is the case with `en` and `en-GB`.

While supporting _every_ case is probably overkill, there might be more values that are worth adding to the supported languages for the dashboard to respect the formatting defaults of more users of the Aspire dashboard in locations outside of the US, France, Spain etc.

Using the console application source below, on Windows 11 Version 22H2 (OS build 22621.3880) with in the UK with en-GB configured, I get the following list of cultures whose formatting for `DateTime.ToString()` differs from their parent culture (e.g. comparing `en` against `en-GB`):

```console
en-AE: 13/07/2024 1:13:01 PM
en-AG: 13/07/2024 1:13:01 pm
en-AI: 13/07/2024 13:13:01
en-AT: 13/07/2024 13:13:01
en-AU: 13/7/2024 1:13:01 pm
en-BB: 13/07/2024 1:13:01 pm
en-BE: 13/07/2024 13:13:01
en-BI: 7/13/2024 13:13:01
en-BM: 13/07/2024 1:13:01 pm
en-BS: 13/07/2024 1:13:01 pm
en-BW: 13/07/2024 13:13:01
en-BZ: 13/07/2024 13:13:01
en-CA: 2024-07-13 1:13:01 p.m.
en-CC: 13/07/2024 13:13:01
en-CH: 13/07/2024 13:13:01
en-CK: 13/07/2024 13:13:01
en-CM: 13/07/2024 13:13:01
en-CX: 13/07/2024 13:13:01
en-CY: 13/07/2024 1:13:01 pm
en-DE: 13/07/2024 13:13:01
en-DK: 13/07/2024 13.13.01
en-DM: 13/07/2024 1:13:01 pm
en-ER: 13/07/2024 1:13:01 pm
en-FI: 13/07/2024 13.13.01
en-FJ: 13/07/2024 1:13:01 pm
en-FK: 13/07/2024 13:13:01
en-FM: 13/07/2024 1:13:01 pm
en-GB: 13/07/2024 13:13:01
en-GD: 13/07/2024 1:13:01 pm
en-GG: 13/07/2024 13:13:01
en-GH: 13/07/2024 1:13:01 pm
[...]

PR #5143, opened 2024-08-01
Add support for all cultures corresponding to localizations we have

adamint:
I looked into this and it appears that, since we must explicitly enumerate the cultures we support, we can simply enable support for all cultures that we have language translations for. 

For some cultures, we localize into a specific locale - ie, pt-BR, but some are a language code - ie, fr. We can enumerate through all cultures in `CultureInfo` and filter out cultures whose language code we support.

Example (showing en-GB and en):
image
(en)

image
(en-GB)

Hopefully, this solves #4884 once and for all
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5143)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -67,6 +67,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
 
     public IReadOnlyList<string> ValidationFailures => _validationFailures;
 
+    // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
     internal static HashSet<string> LocalizedCultures { get; } = new(StringComparer.OrdinalIgnoreCase)
     {
         "en", "cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-BR", "ru", "tr", "zh-Hans", "zh-Hant", // Standard cultures for compliance.
@@ -202,8 +203,8 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         var supportedCultures = GetSupportedCultures();
 
         _app.UseRequestLocalization(new RequestLocalizationOptions()
-            .AddSupportedCultures([.. supportedCultures])
-            .AddSupportedUICultures([.. supportedCultures]));
+            .AddSupportedCultures(supportedCultures)
+            .AddSupportedUICultures(supportedCultures));
 
         WriteVersion(_logger);
 
@@ -334,17 +335,16 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         }
     }
 
-    internal static List<string> GetSupportedCultures()
+    internal static string[] GetSupportedCultures()
     {
-        // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
         var supportedCultures = CultureInfo.GetCultures(CultureTypes.AllCultures)
-            .Where(culture => LocalizedCultures.Contains(culture.Name) || LocalizedCultures.Contains(culture.TwoLetterISOLanguageName))
+            .Where(culture => LocalizedCultures.Contains(culture.TwoLetterISOLanguageName) || LocalizedCultures.Contains(culture.Name))
             .Select(culture => culture.Name)
             .ToList();
 
         // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
         supportedCultures.Add("zh-CN");
-        return supportedCultures;
+        return supportedCultures.ToArray();
     }
 
     private ILogger<DashboardWebApplication> GetLogger()
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -67,6 +67,11 @@ public sealed class DashboardWebApplication : IAsyncDisposable
 
     public IReadOnlyList<string> ValidationFailures => _validationFailures;
 
+    internal static HashSet<string> LocalizedCultures { get; } = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "en", "cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-BR", "ru", "tr", "zh-Hans", "zh-Hant", // Standard cultures for compliance.
+    };
+
     /// <summary>
     /// Create a new instance of the <see cref="DashboardWebApplication"/> class.
     /// </summary>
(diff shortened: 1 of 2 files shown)


adamint:
cc @samsp-msft @kvenkatrajan

martincostello:
Nice - out of curiosity, what does the full list end up being when it runs on your machine?

adamint:
> Nice - out of curiosity, what does the full list end up being when it runs on your machine?

It's several hundred cultures 😄

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -194,18 +194,23 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
...
-        // our language list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
-        var supportedLanguages = new[]
+        // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
+        var localizedLanguages = new[]
         {
             "en", "cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-BR", "ru", "tr", "zh-Hans", "zh-Hant", // Standard cultures for compliance.
-            "zh-CN", // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
-            "en-GB", // Support UK DateTime formatting (24-hour clock, dd/MM/yyyy)
-        };
+        }.ToHashSet();
```
JamesNK, line 201: Is the default comparer correct? Should it ignore case?

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -194,18 +194,23 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
...
+
+        var supportedCultures = CultureInfo.GetCultures(CultureTypes.AllCultures)
+            .Where(culture => localizedLanguages.Contains(culture.Name) || localizedLanguages.Contains(culture.TwoLetterISOLanguageName))
+            .Select(culture => culture.Name)
+            .ToList();
```
JamesNK, line 206: Split out into a method and write a unit test to check this works correctly. Double check a few, popular, expected cultures are in the supported cultures.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -194,18 +194,23 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
 
         _logger = GetLogger();
 
-        // this needs to be explicitly enumerated for each supported language
-        // our language list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
-        var supportedLanguages = new[]
+        // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
+        var localizedLanguages = new[]
```
JamesNK, line 198: nit: use collection initializer to create localizedLanguages collection. Potentially a little more efficient than an extension method on array.

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: e.g. all the values in localizedLanguages, en-GB, en-US, zh-CN, de-DE, fr-FR, etc.

JamesNK:
This works on my Windows computer but is there a chance it breaks in the docker container? Maybe create an issue to test it there once the chances reach the nightly docker container.

src/Aspire.Dashboard/DashboardWebApplication.cs
adamint: It doesn't hurt to ignore case here.

src/Aspire.Dashboard/DashboardWebApplication.cs
adamint: done

adamint:
> This works on my Windows computer but is there a chance it breaks in the docker container? Maybe create an issue to test it there once the chances reach the nightly docker container.

We can create an issue, but I would not expect any break in the docker container, especially considering this works on my Mac computer as well.

tests/Aspire.Dashboard.Tests/GlobalizationTests.cs
```diff
@@ -0,0 +1,23 @@
...
+    public void GetSupportedCultures_IncludesPopularCultures()
+    {
+        var supportedCultures = DashboardWebApplication.GetSupportedCultures();
+
+        foreach (var localizedCulture in DashboardWebApplication.LocalizedCultures)
+        {
+            Assert.Contains(localizedCulture, supportedCultures);
+        }
+
+        Assert.Contains("en-GB", supportedCultures);
```
JamesNK, line 20: Add a comment that these are a couple of examples of derived cultures we expect to be available.

tests/Aspire.Dashboard.Tests/GlobalizationTests.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Xunit;
+
+namespace Aspire.Dashboard.Tests;
+
+public class GlobalizationTests
+{
+    [Fact]
+    public void GetSupportedCultures_IncludesPopularCultures()
```
JamesNK, line 11: nit: Add arrange, act, assert comments to tests. See other tests for examples. They're a simple way to add a bit of structure to tests. Not that valuable in short tests like this, but very useful in longer tests.

JamesNK (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -194,18 +199,11 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
...
-            "zh-CN", // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
-            "en-GB", // Support UK DateTime formatting (24-hour clock, dd/MM/yyyy)
-        };
+        var supportedCultures = GetSupportedCultures();
 
         _app.UseRequestLocalization(new RequestLocalizationOptions()
-            .AddSupportedCultures(supportedLanguages)
-            .AddSupportedUICultures(supportedLanguages));
+            .AddSupportedCultures([.. supportedCultures])
+            .AddSupportedUICultures([.. supportedCultures]));
```
drewnoakes, line 206: This creates three collections for the supported cultures, as the `[.. list]` operator here is creating a new array for each.

Perhaps instead `GetSupportedCultures()` should return `string[]` to avoid these two collection copies. It'd simplify the code a little, and allocate slightly less.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -336,6 +334,19 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
         }
     }
 
+    internal static List<string> GetSupportedCultures()
+    {
+        // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
+        var supportedCultures = CultureInfo.GetCultures(CultureTypes.AllCultures)
+            .Where(culture => LocalizedCultures.Contains(culture.Name) || LocalizedCultures.Contains(culture.TwoLetterISOLanguageName))
```
drewnoakes, line 341: Most of the supported cultures are two-letter, so inverting this might result in less overall work:

```suggestion
            .Where(culture => LocalizedCultures.Contains(culture.TwoLetterISOLanguageName) || LocalizedCultures.Contains(culture.Name))
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -336,6 +334,19 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
         }
     }
 
+    internal static List<string> GetSupportedCultures()
+    {
+        // our localization list comes from https://github.com/dotnet/arcade/blob/89008f339a79931cc49c739e9dbc1a27c608b379/src/Microsoft.DotNet.XliffTasks/build/Microsoft.DotNet.XliffTasks.props#L22
```
drewnoakes, line 339: Feels to me like this comment would be better alongside the `LocalizedCultures` property above.

drewnoakes (approved):

tests/Aspire.Dashboard.Tests/GlobalizationTests.cs
adamint: I don't see any arrange part, but I added act and assert comments

PR #5145, opened 2024-08-02
Better malformed launchsettings error

### LINKED ISSUE #4307: aspire to an existing project error startup
I added aspire to an existing project... when I try to start it, the print error occurs.
If I remove the Web_api project and leave only RabbitMq, it works normally. Any tips?

image

mitchdenny:
Fixes #4307
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5145)

### DIFF
--- a/src/Aspire.Hosting/LaunchProfileExtensions.cs
+++ b/src/Aspire.Hosting/LaunchProfileExtensions.cs
@@ -30,37 +30,26 @@ internal static class LaunchProfileExtensions
 
     internal static NamedLaunchProfile? GetEffectiveLaunchProfile(this ProjectResource projectResource, bool throwIfNotFound = false)
     {
-        try
+        string? launchProfileName = projectResource.SelectLaunchProfileName();
+        if (string.IsNullOrEmpty(launchProfileName))
         {
-            string? launchProfileName = projectResource.SelectLaunchProfileName();
-            if (string.IsNullOrEmpty(launchProfileName))
-            {
-                return null;
-            }
-
-            var profiles = projectResource.GetLaunchSettings()?.Profiles;
-            if (profiles is null)
-            {
-                return null;
-            }
-
-            var found = profiles.TryGetValue(launchProfileName, out var launchProfile);
-            if (!found && throwIfNotFound)
-            {
-                var message = string.Format(CultureInfo.InvariantCulture, Resources.LaunchSettingsFileDoesNotContainProfileExceptionMessage, launchProfileName);
-                throw new DistributedApplicationException(message);
-            }
-
-            return launchProfile is not null ? new (launchProfileName, launchProfile) : default;
+            return null;
         }
-        catch (JsonException ex)
+
+        var profiles = projectResource.GetLaunchSettings()?.Profiles;
+        if (profiles is null)
         {
-            var metadata = projectResource.GetProjectMetadata();
-            var projectFileInfo = new FileInfo(metadata.ProjectPath);
-            var launchSettingsFilePath = Path.Combine(projectFileInfo.DirectoryName!, "Properties", "launchSettings.json");
-            var message = $"Failed to get effective launch profile for project resource '{projectResource.Name}'. There is malformed JSON in the project's launch settings file at '{launchSettingsFilePath}'.";
-            throw new DistributedApplicationException(message, ex);
+            return null;
         }
+
+        var found = profiles.TryGetValue(launchProfileName, out var launchProfile);
+        if (!found && throwIfNotFound)
+        {
+            var message = string.Format(CultureInfo.InvariantCulture, Resources.LaunchSettingsFileDoesNotContainProfileExceptionMessage, launchProfileName);
+            throw new DistributedApplicationException(message);
+        }
+
+        return launchProfile is not null ? new (launchProfileName, launchProfile) : default;
     }
 
     private static LaunchSettings? GetLaunchSettings(this IProjectMetadata projectMetadata, string resourceName)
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/LaunchProfileExtensions.cs
```diff
@@ -30,26 +30,37 @@ internal static class LaunchProfileExtensions
...
+            return launchProfile is not null ? new (launchProfileName, launchProfile) : default;
+        }
+        catch (JsonException ex)
         {
-            var message = string.Format(CultureInfo.InvariantCulture, Resources.LaunchSettingsFileDoesNotContainProfileExceptionMessage, launchProfileName);
-            throw new DistributedApplicationException(message);
+            var metadata = projectResource.GetProjectMetadata();
+            var projectFileInfo = new FileInfo(metadata.ProjectPath);
+            var launchSettingsFilePath = Path.Combine(projectFileInfo.DirectoryName!, "Properties", "launchSettings.json");
+            var message = $"Failed to get effective launch profile because of malformed JSON in '{launchSettingsFilePath}' associated with project resource '{projectResource.Name}'. See inner exception for details.'";
```
JamesNK, line 61:
```suggestion
            var message = $"Failed to get effective launch profile for project resource '{projectResource.Name}'. There is malformed JSON in the project's launch settings file at '{launchSettingsFilePath}'.";
```

JamesNK: IMO this message is a bit better. "See inner exception for details" is implicit with any exception.

src/Aspire.Hosting/LaunchProfileExtensions.cs
```diff
@@ -30,26 +30,37 @@ internal static class LaunchProfileExtensions
...
+            {
+                var message = string.Format(CultureInfo.InvariantCulture, Resources.LaunchSettingsFileDoesNotContainProfileExceptionMessage, launchProfileName);
+                throw new DistributedApplicationException(message);
+            }
 
-        var found = profiles.TryGetValue(launchProfileName, out var launchProfile);
-        if (!found && throwIfNotFound)
+            return launchProfile is not null ? new (launchProfileName, launchProfile) : default;
+        }
+        catch (JsonException ex)
```
eerhardt, line 56: Would it be better if this `try-catch` was moved closer to what was actually doing the Json parsing?

Specifically, if we moved it around lines 93-95 below:

```C#
        using var stream = File.OpenRead(launchSettingsFilePath);
        var settings = JsonSerializer.Deserialize(stream, LaunchSettingsSerializerContext.Default.LaunchSettings);
        return settings;
```

Then we wouldn't need to recreate the file path here, we would have it from the rest of the method.

src/Aspire.Hosting/LaunchProfileExtensions.cs
```diff
@@ -25,34 +25,45 @@ internal static class LaunchProfileExtensions
...
+            var found = profiles.TryGetValue(launchProfileName, out var launchProfile);
+            if (!found && throwIfNotFound)
+            {
+                var message = string.Format(CultureInfo.InvariantCulture, Resources.LaunchSettingsFileDoesNotContainProfileExceptionMessage, launchProfileName);
+                throw new DistributedApplicationException(message);
+            }
+
+            return launchProfile is not null ? new (launchProfileName, launchProfile) : default;
+        }
+        catch (JsonException ex)
```
eerhardt, line 56: Do we still need this try-catch here now that we are catching the JsonException around the call to `JsonSerializer.Deserialize`?

mitchdenny: Good catch ... I forgot to trim that code out!

tests/Aspire.Hosting.Tests/ProjectResourceTests.cs
```diff
@@ -14,6 +14,47 @@ namespace Aspire.Hosting.Tests;
...
 public class ProjectResourceTests
 {
+    [Fact]
+    public async Task AddProjectWithInvalidLaunchSettingsShouldThrowSpecificError()
+    {
+        var projectDetails = await PrepareProjectWithMalformedLaunchSettingsAsync();
+
+        var ex = Assert.Throws<DistributedApplicationException>(() =>
+        {
+            var appBuilder = CreateBuilder();
```
eerhardt, line 24:
```suggestion
        var appBuilder = CreateBuilder();
        var ex = Assert.Throws<DistributedApplicationException>(() =>
        {
```

(nit) I like to keep the action inside the Assert.Throws as minimal as possible, so it is obvious what line is throwing the exception.

tests/Aspire.Hosting.Tests/ProjectResourceTests.cs
```diff
@@ -14,6 +14,47 @@ namespace Aspire.Hosting.Tests;
...
+            var launchSettingsContent = """
+                                        this { is } { mal formed! >
+                                        """;
+
+            var projectDirectoryPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            var projectFilePath = Path.Combine(projectDirectoryPath, "Project.csproj");
+            var propertiesDirectoryPath = Path.Combine(projectDirectoryPath, "Properties");
+            var launchSettingsFilePath = Path.Combine(propertiesDirectoryPath, "launchSettings.json");
+
+            Directory.CreateDirectory(projectDirectoryPath);
```
eerhardt, line 48:
```suggestion
            var projectDirectoryPath = Directory.CreateTempSubdirectory().FullName;
            var projectFilePath = Path.Combine(projectDirectoryPath, "Project.csproj");
            var propertiesDirectoryPath = Path.Combine(projectDirectoryPath, "Properties");
            var launchSettingsFilePath = Path.Combine(propertiesDirectoryPath, "launchSettings.json");

```

Using CreateTempSubdirectory is easier.

eerhardt, line 48: Should we be deleting this directory when the test is done so we don't clutter up the temp directory?

eerhardt (approved):
LGTM

PR #5150, opened 2024-08-02
Support filtering telemetry to all instances of a resource

### LINKED ISSUE #5137: View logs/traces from all instances of an app with resource filter

If there are multiple instances for a named app then they are grouped together in the resource selector:
image

The instances for the app can be selected to filter results to that instance.

However, someone might want to see results for all instances. The top-level option (it says `replica set` in this screenshot but it's being renamed to `application`) is disabled. There is no way to select that you want to view logs/traces from all 3 instances of the lineconfiguration app at once.

User request for this feature: https://github.com/dotnet/aspire/issues/4103#issuecomment-2262710428

The top-level option should be enabled. Selecting the top-level `lineconfiguration (application)` option will cause the logs/traces page to show results for all instances of the app.

We could _maybe_ also do this for metrics. However, it's more technically challenging.

### Additional context

_No response_

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/5137

Aspire supports collecting data from multiple instances of an app, e.g. `CatalogService-123` and `CatalogService-456`. When this happens, the resource selector shows instances grouped together. Selecting an instance shows telemetry for only that instance.

This PR adds support for showing telemetry across all instances of an app. In the resource selector on telemetry pages it enables the group option, e.g. `CatalogService (application)`. Now you can filter telemetry to an instance of an app or view all instances at once.

The new behavior is straight forward on structured logs and tracing pages. They show items for all instances. The metrics page is more complicated. Selecting `CatalogService (application)` will show all the meters/instruments across instances. Selecting an instrument then shows data from all instances in the graph. In histogram graphs, the percentiles are calculated using data from all instances. In counter graphs, counts are added together from all instances. For example, if `consoto.login.count` is 10 in one instance, and 20 in another, the graph will display 30.

Demo:
resource-filter-all-logs

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5150)

### DIFF
--- a/src/Aspire.Dashboard/Otlp/Storage/ApplicationKey.cs
+++ b/src/Aspire.Dashboard/Otlp/Storage/ApplicationKey.cs
@@ -3,7 +3,7 @@
 
 namespace Aspire.Dashboard.Otlp.Storage;
 
-public readonly record struct ApplicationKey(string Name, string InstanceId) : IComparable<ApplicationKey>
+public readonly record struct ApplicationKey(string Name, string? InstanceId) : IComparable<ApplicationKey>
 {
     public int CompareTo(ApplicationKey other)
     {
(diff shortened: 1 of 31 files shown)


samsp-msft:
For non-histogram metrics, can we have an option to graph each resource as its own line on the same graph? I have a similar request for dimensioned metrics - to be able to draw a different line for each dimension value for a specific dimension.

src/Aspire.Dashboard/Otlp/Storage/ApplicationKey.cs
```diff
@@ -23,23 +23,30 @@ public bool EqualsCompositeName(string name)
...
         }
-        if (name[Name.Length] != '-')
+        else
         {
-            return false;
-        }
-        if (!name.AsSpan(Name.Length + 1, InstanceId.Length).Equals(InstanceId, StringComparisons.ResourceName))
-        {
-            return false;
+            return string.Equals(Name, name, StringComparisons.ResourceName);
```
drewnoakes, line 49: Does this need to consider when `other.InstanceId` is non-null?

JamesNK: No. It's supposed to match other apps with the same name and multiple instances.

JamesNK: Added a comment.

JamesNK:
> For non-histogram metrics, can we have an option to graph each resource as its own line on the same graph? I have a similar request for dimensioned metrics - to be able to draw a different line for each dimension value for a specific dimension.

Sure, but that's a different feature and it's not happening in this PR. Create an issue for it.

adamint (approved):

PR #5151, opened 2024-08-02
Move tooltip position in login to top for a11y

### LINKED ISSUE #4171: [AzureTools][Aspire] The displayed information by hovering over “Where do I find the token?” is not adaptive when reducing the size of the browser page
**VS Version:** VS 17.11.0 Preview 2.0 [34913.145.main]
**Aspire Version:** 8.1.0-preview.1.24259.4
Apply NuGet Feeds

**REPRO STEPS:**
1. Make sure there is no localhost in your browser.
2. Open Terminal, run the following commands to create and run the project using https:
```
dotnet new aspire -o aspire
cd aspire/aspire.AppHost
dotnet run
```
3. Reduce the size of your browser page by dragging and dropping with your mouse -> Click the link without token in the command prompt, move mouse to 'Where do I find the token?'.

**Expect:**
The displayed information is adaptive.

**Actual:**
The displayed information is not adaptive. (Only part of the information is displayed, the browser must be enlarged to display the full message.)
Actu

**More info:**
This issue also **reproduces** on VS 17.11.0 Preview 2.0 [34913.145.main] with Aspire 8.0.0.

adamint:
We are not able to have a 'preferred' positioning for the login help tooltip, so we should put at top so it is not cut off, even if it looks slightly worse.

Longer-term @vnbaaij we should add a preferred positioning option for tooltips.

Fixes #4171 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5151)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Login.razor
+++ b/src/Aspire.Dashboard/Components/Pages/Login.razor
@@ -32,7 +32,7 @@
                 <div class="token-entry-footer">
                     <a href="@NavigationManager.Uri" id="helpLink">@Loc[nameof(Dashboard.Resources.Login.WhereIsMyTokenLinkText)]</a>
                     <FluentButton Appearance="Appearance.Accent" Type="ButtonType.Submit">@Loc[nameof(Dashboard.Resources.Login.LogInButtonText)]</FluentButton>
-                    <FluentTooltip Anchor="helpLink" Position="TooltipPosition.Bottom" MaxWidth="650px">
+                    <FluentTooltip Anchor="helpLink" MaxWidth="650px">
                         <div class="token-help-container">
                             <div>
                                 <a href="/img/TokenExample.png" target="_blank">


JamesNK:
This is fine but if you want FluentUI to be improved then you should create an issue in their repo with details.

Also, if there is more work to be done here then the issue should stay open.

adamint:
> This is fine but if you want FluentUI to be improved then you should create an issue in their repo with details.
> 
> Also, if there is more work to be done here then the issue should stay open.

The issue must be closed because the accessibility concern will have ameliorated. Since we have an aesthetic concern, we can open a different issue.

JamesNK:
Opening another one is fine. The important thing is the work is tracked (an issue for fluentui for them to improve popup, an issue for us to track whether fluentui is done and to bring into Aspire)

JamesNK (approved):

PR #5155, opened 2024-08-02
Correct the path of the TestShop.AppHost

IEvangelist:
I was working on https://github.com/dotnet/docs-aspire/issues/1407, so I pulled latest from `main` and spun up the TestShop solution. Visual Studio was complaining that it couldn't find the AppHost, and it was right - it's the wrong path. This PR corrects the path of the TestShop.AppHost.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5155)


eerhardt (approved):
Thanks!

PR #5161, opened 2024-08-02
Adding public API test coverage Aspire.Hosting.NodeJs

Alirexaa:
Contributes to #2649 and #5047
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5161)


Zombach (commented):
Missing check in constructor

https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.NodeJs/NodeAppResource.cs

Alirexaa:
> Missing check in constructor
> 
> https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.NodeJs/NodeAppResource.cs

`ExecutableResource` has check for that.

Zombach:
> > Missing check in constructor
> > https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting.NodeJs/NodeAppResource.cs
> 
> `ExecutableResource` has check for that.
Yes, I know, there was just a case [message](https://github.com/dotnet/aspire/pull/5110#discussion_r1700402328) when the name of a parameter differed from the name of the parameter being checked in the ExecutableResource.
And it turns out that if someone makes changes to the code and changes the name, or starts using parameters in the NodeAppResource class, an error will occur.
I'm certainly not sure. But I would cover all public APIs with this check,
when it's not entirely transparent. Not everyone knows that the check for 'null' is found internally in the ExecutableResource.
But this is just my opinion, I don’t know what others will say about this

tests/Aspire.Hosting.NodeJs.Tests/NodeJsPublicApiTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        var workingDirectory = ".\\app";
+
+        var action = () => builder.AddNpmApp(name: name, workingDirectory: workingDirectory);
+
+        var exception = Assert.Throws<ArgumentNullException>(action);
+        Assert.Equal(nameof(name), exception.ParamName);
+    }
+
+    [Fact]
+    public void AddNpmAppShouldThrowWhenScriptNameIsNull()
```
sebastienros, line 89: @Alirexaa you missed this one. It has a default value but it doesn't mean we still can't call it with a `null` value.

sebastienros: Also I had to force push to revert the changes in the .sln file.

sebastienros (approved):

PR #5172, opened 2024-08-05
Add certificate allow list configuration

### LINKED ISSUE #2995: Dashboard OTLP auth follow up work
* ~OTLP API should support primary and secondary keys. Help make rotation possible.~
* Cert auth currently allows any valid cert. Should provide some way to limit client cert. @drewnoakes I think the resource service will do something similar. Is cert auth being added locally, or is the work just in the dashboard to provide a client cert for cloud hosting?

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/2995

Adds an optional allow list to certification authentication.

The allow list is optional for two reasons:
1. Backwards compatibility
2. I investigated Azure API management to see what it does for certificate validation. Configuration to require a specific certificate - https://learn.microsoft.com/en-us/azure/api-management/api-management-howto-mutual-certificates-for-clients#policy-to-validate-client-certificates - is supported as an optional step that people can choose to add.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5172)

### DIFF
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -61,6 +61,11 @@ public sealed class ResourceServiceClientCertificateOptions
     public StoreLocation? Location { get; set; }
 }
 
+public sealed class AllowedCertificateRule
+{
+    public string? Thumbprint { get; set; }
+}
+
 // Don't set values after validating/parsing options.
 public sealed class OtlpOptions
 {
@@ -75,6 +80,8 @@ public sealed class OtlpOptions
     public string? GrpcEndpointUrl { get; set; }
 
     public string? HttpEndpointUrl { get; set; }
+
+    public List<AllowedCertificateRule> AllowedCertificates { get; set; } = new();
 
     public Uri? GetGrpcEndpointUri()
     {
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -176,7 +176,7 @@ public sealed class FrontendOptions
     {
         if (string.IsNullOrEmpty(EndpointUrls))
         {
-            errorMessage = "One or more frontend endpoint URLs are not configured. Specify a Dashboard:Frontend:EndpointUrls value.";
+            errorMessage = $"One or more frontend endpoint URLs are not configured. Specify a {DashboardConfigNames.DashboardFrontendUrlName.ConfigKey} value.";
             return false;
         }
         else
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -176,7 +176,7 @@ public sealed class FrontendOptions
     {
         if (string.IsNullOrEmpty(EndpointUrls))
         {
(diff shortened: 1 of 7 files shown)


src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -61,6 +61,11 @@ public sealed class ResourceServiceClientCertificateOptions
     public StoreLocation? Location { get; set; }
 }
 
+public sealed class AllowedCertificateRule
+{
+    public string? Thumbprint { get; set; }
```
drewnoakes, line 66: Wondering why this is nullable. Why would we want an instance of this with a null thumbprint?

If the system was configured with a single rule having a null thumbprint, I don't think any connection would ever be accepted.

X509Certificate2.Thumbprint doens't seem to be nullable either.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -560,6 +560,27 @@ private static void ConfigureAuthentication(WebApplicationBuilder builder, Dashb
                 {
                     OnCertificateValidated = context =>
                     {
+                        var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<DashboardOptions>>().Value;
+                        if (options.Otlp.AllowedCertificates is { Count: > 0 } allowList)
```
eerhardt, line 564: For my knowledge, why do we need our own custom option? Doesn't Kestrel have an existing setting for which certs are valid?

JamesNK: ASP.NET Core cert auth provides settings for allowing or rejecting invalid certs, but there is just an event for determining whether a valid cert is allowed or not.

https://learn.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-8.0 - see `OnCertificateValidated`

> Conceptually, the validation of the certificate is an authorization concern. Adding a check on, for example, an issuer or thumbprint in an authorization policy, rather than inside OnCertificateValidated, is perfectly acceptable.

This PR adds configuration and code for that check.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
JamesNK: The property is set by configuration so it could be left out and be null.

But I will add some validation logic for letting the person launching the dashboard that they're made a mistake.

tests/Aspire.Dashboard.Tests/Integration/StartupTests.cs
```diff
@@ -43,8 +43,24 @@ public async Task Configuration_NoExtraConfig_Error()
 
         // Assert
         Assert.Collection(app.ValidationFailures,
-            s => s.Contains("Dashboard:Frontend:EndpointUrls"),
-            s => s.Contains("Dashboard:Otlp:EndpointUrl"));
+            s => Assert.Contains("ASPNETCORE_URLS", s),
+            s => Assert.Contains("DOTNET_DASHBOARD_OTLP_ENDPOINT_URL", s));
```
JamesNK, line 47: Unrelated, but I noticed this test is wrong when adding the allow certs test.

drewnoakes (approved):

Issue #4626, opened 2024-06-23
Resource service client certificate testing


drewnoakes:
(Split out from #3739)

Either set up automated testing for this scenario or add steps to manual test runs to validate the use of certificates over this channel.

kvenkatrajan:
@drewnoakes please confirm if Bala has the steps for this.

drewnoakes:
I've coordinated with Bala and will share info and test steps here.

drewnoakes:
# Manual Certificate Testing

The dashboard supports use of client certificates for authorization when connecting to a resource service. This support exists for custom resource services.

The following instructions explain the rationale for these tests, as well as the required steps and outcomes along the way.

## Generate keys

Some keys and certificates are required for testing. These can be created once and reused across test runs.

This script places these files in `C:\certs\`. You can use a different path, but must update the path wherever it exists throughout the following example.

```bash
# create root CA details
openssl genrsa -aes256 -out rootCA.key 4096
openssl req -x509 -new -nodes -key rootCA.key -sha256 -days 3650 -out rootCA.crt -subj "//CN=MyRootCA"

# server
openssl genrsa -aes256 -out server.key 4096
openssl req -new -key server.key -out server.csr -subj "//CN=localhost"
openssl x509 -req -in server.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out server.crt -days 365 -sha256
openssl pkcs12 -export -out server.pfx -inkey server.key -in server.crt

# client
openssl genrsa -aes256 -out client.key 4096
openssl req -new -key client.key -out client.csr -subj "//CN=localhost"
openssl x509 -req -in client.csr -CA rootCA.crt -CAkey rootCA.key -CAcreateserial -out client.crt -days 365 -sha256
openssl pkcs12 -export -out client.pfx -inkey client.key -in client.crt
```

## Modify the app host

Certificate testing requires a resource service that actually supports certificates, and none exist today (that we have access to). However we can modify the Aspire AppHost model and the `TestShop` playground to demonstrate this working end-to-end.

1. Clone the `dotnet/aspire` repo down.
1. Apply the following changes (correct as of ceb5d5a0af5b350198da4468401b59d7c7bca3aa):

```diff
diff --git a/playground/TestShop/TestShop.AppHost/Properties/launchSettings.json b/playground/TestShop/TestShop.AppHost/Properties/launchSettings.json
index ea78f2933..8ec303999 100644
[...]

PR #5173, opened 2024-08-05
Fix configuration binding

drewnoakes:
The dashboard supports the use of client certificates for authorization to the resource service. This requires a custom resource server, and we're not aware of any that use this feature.

Throughout the Aspire documentation, we advertise values like `Dashboard:ResourceServiceClient:ClientCertificate:Source` to configure certificates, however these values will not bind to the documented configuration values. The CLR property was `ClientCertificates` (plural), not the documented `ClientCertificate` (singular).

This change makes the dashboard respect the documented configuration names.

We don't have any automated tests here yet. We have manual test instructions in https://github.com/dotnet/aspire/issues/4626 for vendors to use, which assume this fix has been applied.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5173)

### DIFF
--- a/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
+++ b/src/Aspire.Dashboard/Configuration/DashboardOptions.cs
@@ -26,7 +26,7 @@ public sealed class ResourceServiceClientOptions
 
     public string? Url { get; set; }
     public ResourceClientAuthMode? AuthMode { get; set; }
-    public ResourceServiceClientCertificateOptions ClientCertificates { get; set; } = new();
+    public ResourceServiceClientCertificateOptions ClientCertificate { get; set; } = new();
     public string? ApiKey { get; set; }
 
     public Uri? GetUri() => _parsedUrl;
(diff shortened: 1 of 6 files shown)


JamesNK:
> Throughout the Aspire documentation, we advertise values like `Dashboard:ResourceServiceClient:ClientCertificates:Source` to configure certificates, however these values will not bind to the documented configuration values.

nit: Typo in the issue description. `Dashboard:ResourceServiceClient:ClientCertificate:Source` is what is in the docs.

JamesNK (approved):

JamesNK:
I assume fixing https://github.com/dotnet/aspire/issues/4626 involves creating integration tests?

JamesNK:
An end-to-end test with client cert auth would be useful.

However, to simply test the configuration works as expected you could write an integration test like here - https://github.com/dotnet/aspire/blob/02a0bc6fef3f2353b306f3ada63445f3bf133b83/tests/Aspire.Dashboard.Tests/Integration/FrontendBrowserTokenAuthTests.cs#L31-L36 - which passes the config valid when the dashboard starts up, then get the dashboard options from the running app and assert that the expected property is set. Consider doing that in this PR or at least in an immediate follow up PR.

drewnoakes:
> I assume fixing #4626 involves creating integration tests?

Ultimately, but for now Bala's team is going to do manual testing, following the steps outlined in https://github.com/dotnet/aspire/issues/4626#issuecomment-2268171540.

> you could write an integration test like here

I added one, though it wouldn't have caught the bug this PR fixes.

JamesNK:
I don't think you can say a test wouldn't have helped here. If the config key in the test didn't set data onto the options because of a bad property name, the test would have caught the bug. And having all the configs as constants in one place is useful when writing docs because you can double check config names in the docs against the config names in the list.

PR #5177, opened 2024-08-05
Add CORS to OTLP HTTP endpoint

JamesNK:
Addresses https://github.com/dotnet/aspire/issues/4917

Implements 1 & 2 from https://github.com/dotnet/aspire/issues/4917#issuecomment-2257348510
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5177)

PR #5212, opened 2024-08-07
Enable CORS for with OTLP HTTP endpoint and add playground app

### LINKED ISSUE #4917: Configuration to allow CORS with OTLP HTTP

Someone writing a rich browser app, e.g. React, might want to send telemetry via OTLP HTTP to the dashboard. The dashboard supports OTLP HTTP, but standard browser security blocks post requests from the browser app to the dashboard because they're hosted on different ports.

Configuration to enable CORS for the dashboard's OTLP HTTP endpoints.

### Additional context

_No response_

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/4917

This PR:

* Builds on top of https://github.com/dotnet/aspire/pull/5177. The previous PR allowed the dashboard to be configured to support CORS. This PR improves the host to automatically configure the dashboard with the allowed origins of all HTTP endpoints in the Aspire app.
* Adds a playground app that uses the `opentelemetry-js` SDK to send OTLP data to the dashboard.

Demo:
browser-open-telemetry

The eagle eyed might have noticed that the browser span started before the server span. It's caused by the browser and server clocks not being perfectly synced. They are off by about 10ms. As far as I can tell there isn't anything that we can do about it.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5212)


src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -160,6 +161,32 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                        if (endpoint.Port != null)
+                        {
+                            corsOrigins.Add($"{endpoint.UriScheme}://localhost:{endpoint.Port}");
+                        }
+                        if (endpoint.TargetPort != null)
+                        {
+                            corsOrigins.Add($"{endpoint.UriScheme}://localhost:{endpoint.TargetPort}");
+                        }
+                    }
+                }
```
JamesNK, line 186: @mitchdenny Is this the right way to get all of the resource endpoints? I noticed that sometimes the `EndpointAnnotation.Port` value was missing or wrong.

For example, one endpoint in TestShop says it has a port of 80, which is wrong. But it has an `EndpointAnnotation.AllocatedEndpoint` which looks like it reports the correct port.

JamesNK:
cc @samsp-msft @aaronpowell @DamianEdwards

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
mitchdenny: Yeah allocated endpoint is what you want.

mitchdenny:
Just thinking whether we need a threat model for this if only just to verify that we've though that we aren't making anything less secure by setting CORS up for all the resource endpoints. I don't think we do because someone who has compromised a web UI on one of the resource endpoints probably has control of the container the UI was launched from, at which point CORS isn't relevant.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
mitchdenny: Yeah, if you want the actual ports as it was determined by DCP then you want the allocated endpoint. Docs can definitely be improved.

If the Port on the EndpointAnnotation is null you are basically saying that you want DCP to allocate a random one.

tests/Aspire.Hosting.Tests/Dashboard/DashboardResourceTests.cs
```diff
@@ -263,6 +263,68 @@ public async Task DashboardResourceServiceUriIsSet()
...
+        var dashboard = Assert.Single(model.Resources.Where(r => r.Name == "aspire-dashboard"));
+
+        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(dashboard, DistributedApplicationOperation.Run, app.Services);
+
+        Assert.Equal("http://localhost:8080,http://localhost:58080", config.Single(e => e.Key == DashboardConfigNames.DashboardOtlpCorsAllowedOriginsKeyName.EnvVarName).Value);
+        Assert.Equal("*", config.Single(e => e.Key == DashboardConfigNames.DashboardOtlpCorsAllowedHeadersKeyName.EnvVarName).Value);
+    }
+
+    [Fact]
+    public async Task DashboardResource_OtlpGrpcEndpoint_CorsEnvVarSet()
```
JamesNK, line 298:
```suggestion
    public async Task DashboardResource_OtlpGrpcEndpoint_CorsEnvVarNotSet()
```

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -160,6 +161,42 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                    .ToList();
+
+                var corsOrigins = new HashSet<string>();
+                foreach (var endpoint in allResourceEndpoints)
+                {
+                    if (endpoint.UriScheme is "http" or "https")
+                    {
+                        // Prefer allocated endpoint over EndpointAnnotation.Port.
+                        var origin = endpoint.AllocatedEndpoint?.UriString;
+                        if (origin == null && endpoint.Port != null)
```
mitchdenny, line 179: Not sure if this fallback logic is necessary. We should always have an allocated endpoint at this stage.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -160,6 +161,42 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                        {
+                            origin = $"{endpoint.UriScheme}://localhost:{endpoint.Port}";
```
mitchdenny, line 181: `localhost` could have an extreme corner case here. Lets say someone spins up a headless browser in a container (perhaps for test automation). Then localhost will likely be the wrong hostname (platform dependent).

mitchdenny (approved):
LGTM. Added a few comments but nothing that should block.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -148,7 +148,39 @@ public Func<EndpointInfo> OtlpServiceHttpEndPointAccessor
...
+                // Default policy allows the dashboard's origins.
+                // This is added so CORS middleware doesn't report failure for dashboard browser requests that include an origin header.
+                options.AddDefaultPolicy(builder =>
+                {
+                    builder.WithOrigins(dashboardOptions.Frontend.GetEndpointUris().Select(uri => uri.OriginalString).ToArray());
+                    builder.AllowAnyHeader();
+                    builder.AllowAnyMethod();
+                });
+
+                options.AddPolicy("OtlpHttp", builder =>
```
drewnoakes, line 164: Seems like `OtlpHttp` is a constant that's referenced again below in `OtlpHttpEndpointsBuilder.cs`. Extract to a shared location?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -160,6 +161,42 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
                 context.EnvironmentVariables[DashboardConfigNames.DashboardOtlpHttpUrlName.EnvVarName] = otlpHttpEndpointUrl;
+
+                var model = context.ExecutionContext.ServiceProvider.GetRequiredService<DistributedApplicationModel>();
+                var allResourceEndpoints = model.Resources
+                    .Where(r => !string.Equals(r.Name, KnownResourceNames.AspireDashboard, StringComparisons.ResourceName))
+                    .SelectMany(r => r.Annotations)
+                    .OfType<EndpointAnnotation>()
+                    .ToList();
+
+                var corsOrigins = new HashSet<string>();
```
drewnoakes, line 172: Explicit comparator?

drewnoakes (approved):

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
JamesNK: Let's wait and see whether there is feedback about the automatic CORS policy not working. Can figure out improvements then.

Issue #2036, opened 2024-02-01
Don't hardcode project, and dcp paths in the assemblies


radical:
Currently we add various resolved `dcp` paths, and some projects in `AppHost.dll`. This makes the build output tied to the same machine or at least one with the same layout.

Instead, we should consider adding all this information to a new json (or a custom section in `runtimeconfig.json`) file which gets deployed with the build output. The project paths can be keyed by their names. `launchsettings` needed from the various projects can also be directly referenced or deployed with the AppHost output.

cc @eerhardt @joperezr

eerhardt:
FYI @davidfowl @mitchdenny @danegsta 

I think this would be a good idea. We hit this trying to run tests on a machine that is different than the build machine. If we instead wrote this information out into a json file, the tests could patch up the hard-coded paths and the AppHost would run on the new machine.

We could still do the same `IProjectMetadata` implementation we have today. But instead of returning the hard-coded path:

https://github.com/dotnet/aspire/blob/0d7749b3c34755cc1006ccfee9aa24f7343e4aea/src/Aspire.Hosting/build/Aspire.Hosting.targets#L40-L42

this code instead would say:

```C#
public class ]]>%(ClassName)<![CDATA[ : global::Aspire.Hosting.IProjectMetadata
{
  public string ProjectPath => 
    // load Aspire.AppHost.json file
    // return json.Projects["%(ClassName)"];
```

Similarly we can write the other paths we are hard-coding today into this json file:

```C#
[assembly: AssemblyMetadata("dcpclipath", "C:\\Users\\eerhardt\\.nuget\\packages\\microsoft.developercontrolplane.windows-amd64\\0.1.49\\tools\\dcp.exe")]
[assembly: AssemblyMetadata("dcpextensionpaths", "C:\\Users\\eerhardt\\.nuget\\packages\\microsoft.developercontrolplane.windows-amd64\\0.1.49\\tools\\ext\\")]
[assembly: AssemblyMetadata("dcpbinpath", "C:\\Users\\eerhardt\\.nuget\\packages\\microsoft.developercontrolplane.windows-amd64\\0.1.49\\tools\\ext\\bin\\")]
[assembly: AssemblyMetadata("apphostprojectpath", "C:\\git\\aspire\\playground\\mongo\\Mongo.AppHost")]
[...]

PR #5215, opened 2024-08-07
[tests] Move more tests to helix by overriding dcp path from assembly metadata

radical:
## Description

`AppHost` build embeds metadata with paths for `DcpCliPath`, `DcpExtensionsPath`, and `DcpBinPath`. This ties the apphost to the machine that it was built on.
This PR:
- When `DcpPublisher:CliPath` is available from the configuration, set the other two paths based on that. 
- Makes `$(RunTestsOnHelix)=true` the default.

This allows running most of the tests on helix. The remaining ones depend on being able to run the test projects which doesn't work on helix because the path to the project files is embedded in the assembly metadata.

Also, download helix test result `.trx` files to have all the `.trx` files available in `artifacts`.

Contributes to https://github.com/dotnet/aspire/issues/2036 .

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5215)

### DIFF
--- a/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
+++ b/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
@@ -186,10 +186,28 @@ namespace Projects%3B
 
   <!-- This target registers the location of the Aspire orchestration dependencies -->
   <Target Name="SetOrchestrationDiscoveryAttributes" BeforeTargets="GetAssemblyAttributes" Condition=" '$(DcpDir)' != '' ">
+    <PropertyGroup>
+      <DcpDir>$([MSBuild]::EnsureTrailingSlash('$(DcpDir)'))</DcpDir>
+      <DcpExtensionsDir Condition=" '$(DcpExtensionsDir)' == '' ">$([MSBuild]::NormalizePath($(DcpDir), 'ext'))</DcpExtensionsDir>
+      <DcpExtensionsDir>$([MSBuild]::EnsureTrailingSlash('$(DcpExtensionsDir)'))</DcpExtensionsDir>
+      <DcpBinDir Condition=" '$(DcpBinDir)' == '' ">$([MSBuild]::NormalizePath($(DcpExtensionsDir), 'bin'))</DcpBinDir>
+      <DcpBinDir>$([MSBuild]::EnsureTrailingSlash('$(DcpBinDir)'))</DcpBinDir>
+      <DcpCliPath Condition=" '$(DcpCliPath)'  == '' ">$([MSBuild]::NormalizePath($(DcpDir), 'dcp'))</DcpCliPath>
+      <DcpCliPath Condition=" '$(OS)' == 'Windows_NT' and !$(DcpCliPath.EndsWith('.exe')) ">$(DcpCliPath).exe</DcpCliPath>
+    </PropertyGroup>
+
     <ItemGroup>
       <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
-        <_Parameter1>dcpdir</_Parameter1>
-        <_Parameter2>$(DcpDir)</_Parameter2>
+        <_Parameter1>dcpclipath</_Parameter1>
+        <_Parameter2>$(DcpCliPath)</_Parameter2>
+      </AssemblyAttribute>
+      <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
+        <_Parameter1>dcpextensionpaths</_Parameter1>
+        <_Parameter2>$(DcpExtensionsDir)</_Parameter2>
+      </AssemblyAttribute>
+      <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
+        <_Parameter1>dcpbinpath</_Parameter1>
+        <_Parameter2>$(DcpBinDir)</_Parameter2>
       </AssemblyAttribute>
       <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
         <_Parameter1>apphostprojectpath</_Parameter1>
--- a/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
+++ b/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
@@ -198,16 +198,8 @@ namespace Projects%3B
 
     <ItemGroup>
       <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
-        <_Parameter1>dcpclipath</_Parameter1>
-        <_Parameter2>$(DcpCliPath)</_Parameter2>
-      </AssemblyAttribute>
-      <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
-        <_Parameter1>dcpextensionpaths</_Parameter1>
-        <_Parameter2>$(DcpExtensionsDir)</_Parameter2>
-      </AssemblyAttribute>
-      <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
-        <_Parameter1>dcpbinpath</_Parameter1>
-        <_Parameter2>$(DcpBinDir)</_Parameter2>
+        <_Parameter1>dcpdir</_Parameter1>
+        <_Parameter2>$(DcpDir)</_Parameter2>
       </AssemblyAttribute>
       <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
         <_Parameter1>apphostprojectpath</_Parameter1>
--- a/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
+++ b/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
@@ -186,16 +186,6 @@ namespace Projects%3B
 
   <!-- This target registers the location of the Aspire orchestration dependencies -->
   <Target Name="SetOrchestrationDiscoveryAttributes" BeforeTargets="GetAssemblyAttributes" Condition=" '$(DcpDir)' != '' ">
-    <PropertyGroup>
-      <DcpDir>$([MSBuild]::EnsureTrailingSlash('$(DcpDir)'))</DcpDir>
-      <DcpExtensionsDir Condition=" '$(DcpExtensionsDir)' == '' ">$([MSBuild]::NormalizePath($(DcpDir), 'ext'))</DcpExtensionsDir>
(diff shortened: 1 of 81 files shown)


src/Aspire.Hosting/Dcp/DcpOptions.cs
```diff
@@ -113,16 +111,22 @@ public void Configure(DcpOptions options)
         var dcpPublisherConfiguration = configuration.GetSection(DcpPublisher);
         var assemblyMetadata = appOptions.Assembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
 
-        if (!string.IsNullOrEmpty(dcpPublisherConfiguration[nameof(options.CliPath)]))
+        // Find dcp tools path in the following order:
+        // 1. Environment variable ASPIRE_DCP_DIR
+        // 2. Configuration value
+        // 3. Assembly metadata
+        if (Environment.GetEnvironmentVariable("ASPIRE_DCP_DIR") is var dcpDirEnvVar && !string.IsNullOrEmpty(dcpDirEnvVar))
```
radical, line 118: Should this be `DOTNET_ASPIRE_DCP_DIR`?

karolz-ms: I'd think so, this is the prefix we have been using for environment variables in other places, including this file

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
```diff
@@ -198,16 +198,8 @@ namespace Projects%3B
```
eerhardt, line 198: Can we remove lines 191-196 above?

```
      <DcpExtensionsDir Condition=" '$(DcpExtensionsDir)' == '' ">$([MSBuild]::NormalizePath($(DcpDir), 'ext'))</DcpExtensionsDir>
      <DcpExtensionsDir>$([MSBuild]::EnsureTrailingSlash('$(DcpExtensionsDir)'))</DcpExtensionsDir>
      <DcpBinDir Condition=" '$(DcpBinDir)' == '' ">$([MSBuild]::NormalizePath($(DcpExtensionsDir), 'bin'))</DcpBinDir>
      <DcpBinDir>$([MSBuild]::EnsureTrailingSlash('$(DcpBinDir)'))</DcpBinDir>
      <DcpCliPath Condition=" '$(DcpCliPath)'  == '' ">$([MSBuild]::NormalizePath($(DcpDir), 'dcp'))</DcpCliPath>
      <DcpCliPath Condition=" '$(OS)' == 'Windows_NT' and !$(DcpCliPath.EndsWith('.exe')) ">$(DcpCliPath).exe</DcpCliPath>
```

Those properties don't seem to be used anymore.

src/Aspire.Hosting/Dcp/DcpOptions.cs
```diff
@@ -113,16 +111,22 @@ public void Configure(DcpOptions options)
         var dcpPublisherConfiguration = configuration.GetSection(DcpPublisher);
         var assemblyMetadata = appOptions.Assembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
 
-        if (!string.IsNullOrEmpty(dcpPublisherConfiguration[nameof(options.CliPath)]))
```
eerhardt, line 116: Are there any backwards compatibility concerns we need to maintain here? For example, if someone used an `8.1` Aspire.Hosting.AppHost package and an `8.2` Aspire.Hosting package, I think this would break. Are we concerned about that level of mismatch?

src/Aspire.Hosting/Dcp/DcpOptions.cs
```diff
@@ -113,16 +111,22 @@ public void Configure(DcpOptions options)
...
+        // Find dcp tools path in the following order:
+        // 1. Environment variable DOTNET_ASPIRE_DCP_DIR
+        // 2. Configuration value
```
eerhardt, line 116: Environment variables already factor into the Configuration. I don't think we need a separate/new environment variable. Instead, the tests can just set the `DcpPublisher__CliPath` environment variable.

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
radical: done.

src/Aspire.Hosting/Dcp/DcpOptions.cs
radical: We could also compute all 3 paths based on this `CliPath` then. I'm not sure why that is not being done currently.

tests/Aspire.Azure.AI.OpenAI.Tests/Aspire.Azure.AI.OpenAI.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <RunTestsOnHelix>true</RunTestsOnHelix>
```
eerhardt, line 5: At this point (or some near future point), would it make sense to default RunTestsOnHelix=true, and have the other tests opt out?

radical: why not now!

radical:
@karolz-ms @danegsta This PR changes the behavior when you specify `clipath` config by computing the other 2 paths baed on that. Does that break any assumptions?

src/Aspire.Hosting/Dcp/DcpOptions.cs
davidfowl: Yep.

radical: One option would be to retain the 3 paths in the metadata as-is, but when `DcpPublisher__CliPath` is set *then* override the 3 paths based on the `cliPath`.

radical: Made the change.

tests/Aspire.Hosting.Milvus.Tests/Aspire.Hosting.Milvus.Tests.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
     <NoWarn>$(NoWarn);CS8002</NoWarn> <!-- Milvus.Client packages are not signed -->
+    <!-- Cannot be moved to helix as this uses TestProject -->
```
eerhardt, line 6: What does "this uses TestProject" mean?

radical: It's because of this test:
https://github.com/dotnet/aspire/blob/357a0b2e35b9188ba7c4437be3287b9b86388176/tests/Aspire.Hosting.Qdrant.Tests/AddQdrantTests.cs#L167-L204
and
https://github.com/dotnet/aspire/blob/357a0b2e35b9188ba7c4437be3287b9b86388176/tests/Aspire.Hosting.Milvus.Tests/AddMilvusTests.cs#L95-L128

Maybe we can move these tests to `Aspire.Hosting.Tests` which runs on the build machine, which would allow Qdrant+Milvus ones to run on helix?

radical: Also, if this kind of a test is useful then we should add it for other `Aspire.Hosting.*` also.

eerhardt: Yeah, i think we should rewrite this a bit. I'll put up a separate PR.

eerhardt: This ended up being simple. TestProgram isn't being used. I pushed a change to this PR for it.

radical: nice! thanks 👍

danegsta:
> @karolz-ms @danegsta This PR changes the behavior when you specify `clipath` config by computing the other 2 paths based on that. Does that break any assumptions?

That should be just fine.

tests/helix/send-to-helix-basictests.targets
```diff
@@ -3,9 +3,15 @@
   <PropertyGroup>
     <WorkItemArchiveWildCard>$(TestArchiveTestsDir)**/*.zip</WorkItemArchiveWildCard>
     <BuildHelixWorkItemsDependsOn>$(BuildHelixWorkItemsDependsOn);BuildHelixWorkItemsForDefaultTests</BuildHelixWorkItemsDependsOn>
+    <NeedsDcpPathOverride>true</NeedsDcpPathOverride>
   </PropertyGroup>
 
   <Target Name="BuildHelixWorkItemsForDefaultTests">
+    <ItemGroup>
+      <!-- needed for Aspire.Hosting.Container.Tests -->
+      <HelixPreCommand Include="$(_EnvVarSetKeyword) DOCKER_BUILDKIT=1" />
```
eerhardt, line 12: Is it possible to put this in the Hosting.Container.Tests so it doesn't need to be set everywhere?

eerhardt (approved):
Awesome! Nice work, let's get this in.

tests/helix/send-to-helix-basictests.targets
radical: Would that be useful? Right now it would have to be by conditionally adding it on the work item itself.

eerhardt: My thinking is that we wouldn't need a comment like `<!-- needed for Aspire.Hosting.Container.Tests -->` if it was in the Aspire.Hosting.Container.Tests project.

radical: Agreed. In the follow up I'm generating run scripts which would allow individual test projects to control their own setups directly.

PR #5216, opened 2024-08-07
Fix TestBuilderTests flakyness

sebastienros:
Fixes

```
Aspire.Hosting.Testing.Tests.TestingBuilderTests.HttpClientGetTest(genericEntryPoint: True) [FAIL]
Aspire.Hosting.Testing.Tests.TestingBuilderTests.HttpClientGetTest(genericEntryPoint: False) [FAIL]
Aspire.Hosting.Testing.Tests.TestingBuilderTests.SelectsFirstLaunchProfile(genericEntryPoint: True) [FAIL]
Aspire.Hosting.Testing.Tests.TestingBuilderTests.SelectsFirstLaunchProfile(genericEntryPoint: False) [FAIL]
Aspire.Hosting.Testing.Tests.TestingFactoryTests.HttpClientGetTest [FAIL]
```

Fixed the tests on local Linux box. Was failing consistently before.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5216)

### DIFF
--- a/tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
+++ b/tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
@@ -3,6 +3,7 @@
 
 using System.Net.Http.Json;
 using Aspire.Components.Common.Tests;
+using Aspire.Hosting.Tests.Utils;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -46,6 +47,9 @@ public class TestingFactoryTests(DistributedApplicationFixture<Projects.TestingA
         // Wait for resource to start.
         var rns = _app.Services.GetRequiredService<ResourceNotificationService>();
         await rns.WaitForResourceAsync("mywebapp1").WaitAsync(TimeSpan.FromSeconds(60));
+
+        // Wait for the application to be ready
+        await _app.WaitForText("Application started.").WaitAsync(TimeSpan.FromMinutes(1));
 
         var httpClient = _app.CreateHttpClientWithResilience("mywebapp1");
         var result1 = await httpClient.GetFromJsonAsync<WeatherForecast[]>("/weatherforecast");
--- a/tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
+++ b/tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
@@ -49,7 +49,7 @@ public class TestingFactoryTests(DistributedApplicationFixture<Projects.TestingA
         await rns.WaitForResourceAsync("mywebapp1").WaitAsync(TimeSpan.FromSeconds(60));
 
         // Wait for the application to be ready
-        await _app.WaitForText("Application started.").WaitAsync(TimeSpan.FromMinutes(1));
+        await _app.WaitForTextAsync("Application started.").WaitAsync(TimeSpan.FromMinutes(1));
 
         var httpClient = _app.CreateHttpClientWithResilience("mywebapp1");
         var result1 = await httpClient.GetFromJsonAsync<WeatherForecast[]>("/weatherforecast");
--- a/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
+++ b/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
@@ -1,0 +1,100 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using IResource = Aspire.Hosting.ApplicationModel.IResource;
+using ResourceLoggerService = Aspire.Hosting.ApplicationModel.ResourceLoggerService;
+using ResourceNotificationService = Aspire.Hosting.ApplicationModel.ResourceNotificationService;
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Hosting.Tests.Utils;
+
+public static class LoggerNotificationExtensions
+{
+    /// <summary>
+    /// Waits for the specified text to be logged.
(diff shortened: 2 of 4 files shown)


tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    /// <summary>
+    /// Waits for the specified text to be logged.
+    /// </summary>
+    /// <param name="app">The <see cref="DistributedApplication" /> instance to watch.</param>
+    /// <param name="logText">The text to wait for.</param>
+    /// <param name="resource">An optional <see cref="IResource"/> instance to filter the logs for.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    public static Task WaitForText(this DistributedApplication app, string logText, IResource? resource = null, CancellationToken cancellationToken = default)
```
radical, line 23: nit: maybe use `Async` suffix for the name - `WaitForTextAsync`

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+        ArgumentNullException.ThrowIfNull(logTexts);
+
+        var hostApplicationLifetime = app.Services.GetRequiredService<IHostApplicationLifetime>();
+
+        using var watchCts = CancellationTokenSource.CreateLinkedTokenSource(hostApplicationLifetime.ApplicationStopping, cancellationToken);
+        var watchToken = watchCts.Token;
+
+        var tcs = new TaskCompletionSource();
+
+        _ = Task.Run(() => WatchNotifications(app, resource, logTexts, tcs, watchToken), watchToken);
```
radical, line 51: Since nothing will be watching this task's errors, maybe we should have a `try/catch` around `WatchNotifications` and print the exception?

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    public static Task WaitForText(this DistributedApplication app, IEnumerable<string> logTexts, IResource? resource = null, CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(app);
+        ArgumentNullException.ThrowIfNull(logTexts);
+
+        var hostApplicationLifetime = app.Services.GetRequiredService<IHostApplicationLifetime>();
+
+        using var watchCts = CancellationTokenSource.CreateLinkedTokenSource(hostApplicationLifetime.ApplicationStopping, cancellationToken);
```
radical, line 46: When would this get disposed? At the end of this method, or end of `Task.Run` on line 51?

sebastienros:
NB: this is very similar to what `ResourceLoggerForwarderService` is doing but ti wait for specific events instead of just forwarding it to logs.

radical:
This should fix https://github.com/dotnet/aspire/issues/4650 .

radical:
> This should fix #4650 .

.. which should allow re-enabling the corresponding tests in `TestingFactoryTests`.

radical (approved):
LGTM 👍

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                var resourceId = resourceEvent.ResourceId;
+
+                if (loggingResourceIds.Add(resourceId))
+                {
+                    // Start watching the logs for this resource ID
+                    logWatchTasks.Add(WatchResourceLogs(tcs, resourceId, logTexts, resourceLoggerService, cancellationToken));
+                }
+            }
```
eerhardt, line 82: How does this loop ever stop? Only once the application stops?

tests/Aspire.Hosting.Testing.Tests/TestingFactoryTests.cs
```diff
@@ -47,6 +48,9 @@ public async Task HttpClientGetTest()
         var rns = _app.Services.GetRequiredService<ResourceNotificationService>();
         await rns.WaitForResourceAsync("mywebapp1").WaitAsync(TimeSpan.FromSeconds(60));
 
+        // Wait for the application to be ready
+        await _app.WaitForTextAsync("Application started.").WaitAsync(TimeSpan.FromMinutes(1));
```
eerhardt, line 52: We are already `await rns.WaitForResourceAsync("mywebapp1")` above. Why is this wait needed?

If the existing code isn't good enough, then i think we might need to change our test template code as well, because the previous code is basically what the template spits out:

https://github.com/dotnet/aspire/blob/024b48bf581f6790b0878128c6acff7863d3d5f7/src/Aspire.ProjectTemplates/templates/aspire-xunit/IntegrationTest1.cs#L25-L32

cc @DamianEdwards @ReubenBond

sebastienros: I assume `WaitForResourceAsync` doesn't account for when the service can actually serve requests. I dumped the logs for this test and added markers before/after each call:

```
2024-08-08T18:39:12.3897554Z WaitForResourceAsync - Start
2024-08-08T18:39:13.5023642Z WaitForResourceAsync - Stopped
2024-08-08T18:39:13.5024903Z WaitForTextAsync - Start
2024-08-08T18:39:14.3305581Z WaitForTextAsync - Stopped
```

So quite some delay between WaitForResourceAsync and the application is ready.

eerhardt: My concern is how does the tests in our template work today without this? Do we hve this flakiness issue in them?

PR #5220, opened 2024-08-07
Hosting.Keycloak: Fix `importDirectory` param validation in `WithRealmImport`

radical:
## Description

The import path is expected to be relative to the AppHost path, but here
it was being used as-is - relative to the current working directory.
Instead, expand the path relative to the AppHost path and use that for `Directory.Exists` check.

This manifested as a failure to import the file when running the app
with `Aspire.Hosting.Testing` which would start the apphost with
cwd!=apphost .

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5220)


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -243,6 +243,13 @@ private void EnsureEntryPointStarted()
                 {
                     EnsureDepsFile(_entryPoint);
 
+                    _appHostProjectPath = ResolveProjectPath(_entryPoint.Assembly);
+                    if (!string.IsNullOrEmpty(_appHostProjectPath))
+                    {
+                        // FIXME: reset to old one at some point?
+                        Environment.CurrentDirectory = _appHostProjectPath;
```
eerhardt, line 250: I don't think we should be changing the global CWD during these tests. That would cause problems if you are running multiple tests in parallel.

I think the original problem is in WithRealmImport.

https://github.com/dotnet/aspire/blob/024b48bf581f6790b0878128c6acff7863d3d5f7/src/Aspire.Hosting.Keycloak/KeycloakResourceBuilderExtensions.cs#L141-L146

That Directory.Exists check isn't right. The `importDirectory` is relative to the apphost directory. See the next level down in `.WithBindMount`:

https://github.com/dotnet/aspire/blob/024b48bf581f6790b0878128c6acff7863d3d5f7/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L78-L80

I think we should just remove that Directory.Exists call to fix the original issue.

radical: Great feedback, thanks!

radical: Done.

radical:
cc @julioct

tests/TestingAppHost1/TestingAppHost1.AppHost/Program.cs
```diff
@@ -6,10 +6,14 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
 builder.AddRedis("redis1");
+var keycloak = builder.AddKeycloak("keycloak", 8080)
```
julioct, line 9: You probably don't need to specify a port in this case.

src/Aspire.Hosting.Keycloak/KeycloakResourceBuilderExtensions.cs
```diff
@@ -138,13 +138,7 @@ public static IResourceBuilder<KeycloakResource> WithDataBindMount(this IResourc
     {
         ArgumentNullException.ThrowIfNull(importDirectory);
 
-        if (!Directory.Exists(importDirectory))
```
joperezr, line 141: Should we add a test here now that specifies a non existing directory and make sure we throw the right exception?

tests/TestingAppHost1/realms/weathershop-realm.json
```diff
@@ -0,0 +1,1933 @@
+{
```
eerhardt, line 1: Do we need this duplicate 2,000 line json file?

tests/TestingAppHost1/TestingAppHost1.AppHost/TestingAppHost1.AppHost.csproj
```diff
@@ -12,6 +12,7 @@
     <ProjectReference Include="..\..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\..\src\Aspire.Hosting.PostgreSQL\Aspire.Hosting.PostgreSQL.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\..\src\Aspire.Hosting.Redis\Aspire.Hosting.Redis.csproj" IsAspireProjectResource="false" />
+    <ProjectReference Include="..\..\..\src\Aspire.Hosting.Keycloak\Aspire.Hosting.Keycloak.csproj" IsAspireProjectResource="false" />
```
eerhardt, line 15: I don't think we should make this test depend on Keycloak.

radical:
We can drop the changes in `TestingAppHost1`, but then we need to add a test somewhere else. I did try adding `Keycloak` playground app to the tests for which `/alive`, and `/health` work fine but `/weatherforecast` fails with ` Endpoint 'http://localhost:52119/weatherforecast' for resource 'apiservice' in app 'Keycloak.AppHost' returned status code Unauthorized`.

https://github.com/dotnet/aspire/blob/972f9a9128e949964249fe33b627de9ce67be964/playground/keycloak/Keycloak.ApiService/Program.cs#L28-L40

radical:
Looks like `WithBindMount` does not expect the path to exist. And `ContainerMountAnnotation` even allows a null source path.
https://github.com/dotnet/aspire/blob/ac3dbc4e0d6982ebf69b980789f9b06a90ed63c1/src/Aspire.Hosting/ApplicationModel/ContainerMountAnnotation.cs#L18-L19

I'll restore the check, and expand the path using the apphost path.

src/Aspire.Hosting.Keycloak/KeycloakResourceBuilderExtensions.cs
radical: Added.

tests/TestingAppHost1/TestingAppHost1.AppHost/Program.cs
radical: Dropped that change completely.

julioct (approved):
LGTM

src/Aspire.Hosting.Keycloak/KeycloakResourceBuilderExtensions.cs
```diff
@@ -152,9 +152,10 @@ public static IResourceBuilder<KeycloakResource> WithDataBindMount(this IResourc
...
+        var importDirectoryFullPath = Path.GetFullPath(importDirectory, builder.ApplicationBuilder.AppHostDirectory);
+        if (!Directory.Exists(importDirectoryFullPath))
         {
-            throw new DirectoryNotFoundException($"The realm import directory {importDirectory} does not exist.");
+            throw new DirectoryNotFoundException($"The realm import directory {importDirectory} does not exist under {builder.ApplicationBuilder.AppHostDirectory} .");
         }
 
         builder.WithBindMount(importDirectory, RealmImportDirectory, isReadOnly);
```
eerhardt, line 161: Would it make sense to change this:

```suggestion
        builder.WithBindMount(importDirectoryFullPath, RealmImportDirectory, isReadOnly);
```

That way we ensure the path we are checking above is the path being used.

eerhardt (approved):

eerhardt (approved):

radical:
Merge after https://github.com/dotnet/aspire/pull/5276 .

PR #5223, opened 2024-08-08
.NET Aspire Eventing API

mitchdenny:
## Description

This PR explores the idea of introducing an eventing mechanism into .NET Aspire. It adds an `Eventing` property to the builder and injects an `IDistributedApplicationEventing` service into DI which can be used throughout the lifecycle of the application to fire off events.

Any code that can access the application builder or resolve the service from DI can publish an event and an event is extensible. Events are global so it is up to the subscriber to define the callback to filter out specific events they are not interested in (although event of a specific type only go to subscribers for that type in the current implementation).

Related #5234 (API Design issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No (not yet)
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [ ] No (this will be done - this is just a spie)
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [x] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1478
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5223)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1567,6 +1569,9 @@ private async Task CreateContainerAsync(AppResource cr, ILogger resourceLogger,
         }
 
         await kubernetesService.CreateAsync(dcpContainerResource, cancellationToken).ConfigureAwait(false);
+
+        var containerStartedEvent = new ContainerResourceStartedEvent(modelContainerResource);
+        await eventing.PublishAsync(containerStartedEvent, cancellationToken).ConfigureAwait(false);
```
eerhardt, line 1574: My understanding is that kubernetes follows a "desired state" and "actual state" model. Just because the `CreateAsync` call completed doesn't necessarily mean the container has been started.

So are we going to have timing issues that this "container resource started" event is fired, but the actual container hasn't been started yet?

cc @karolz-ms

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    private readonly Dictionary<Type, List<DistributedApplicationEventSubscription>> _subscriptions = new();
+
+    public async Task PublishAsync<T>(T @event, CancellationToken cancellationToken) where T : IDistributedApplicationEvent
+    {
+        if (_subscriptions.TryGetValue(typeof(T), out var subscriptions))
+        {
+            foreach (var subscription in subscriptions)
+            {
+                await subscription.Callback(@event, cancellationToken).ConfigureAwait(false);
```
eerhardt, line 13: Do we want these subscriptions to run in serial order like this? Each one needing to wait for the previous to finish.

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
```diff
@@ -0,0 +1,41 @@
...
+
+internal class DistributedApplicationEventing : IDistributedApplicationEventing
+{
+    private readonly Dictionary<Type, List<DistributedApplicationEventSubscription>> _subscriptions = new();
+
+    public async Task PublishAsync<T>(T @event, CancellationToken cancellationToken) where T : IDistributedApplicationEvent
+    {
+        if (_subscriptions.TryGetValue(typeof(T), out var subscriptions))
+        {
+            foreach (var subscription in subscriptions)
```
eerhardt, line 11: We probably need to worry about concurrency here. What happens if an event is subscribed/unsubscribed while the event is being fired? This is going to throw because the list is modified while enumerating it.

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
```diff
@@ -0,0 +1,30 @@
+namespace Aspire.Hosting.Eventing;
+
+/// <summary>
+/// Supports publishing and subscribing to events which are executed during the AppHost lifecycle.
+/// </summary>
+public interface IDistributedApplicationEventing
```
karolz-ms, line 6: `IDistributedApplicationEventHub`? ☺️

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
```diff
@@ -0,0 +1,30 @@
...
+    void Unsubscribe(DistributedApplicationEventSubscription subscription);
+
+    /// <summary>
+    /// Publishes an event to all subscribes of the specific event type.
+    /// </summary>
+    /// <typeparam name="T">The type of the event</typeparam>
+    /// <param name="event">The event.</param>
+    /// <param name="cancellationToken">A cancellation token.</param>
+    /// <returns>A task that can be awaited.</returns>
+    Task PublishAsync<T>(T @event, CancellationToken cancellationToken) where T : IDistributedApplicationEvent;
```
karolz-ms, line 29: What is the semantics associated with the returned Task? E.g. is it that the Task ends when all current subscribers have processed the event?

I would expect some documentation about it.

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                await subscription.Callback(@event, cancellationToken).ConfigureAwait(false);
+            }
+        }
+    }
+
+    public DistributedApplicationEventSubscription Subscribe<T>(Func<T, CancellationToken, Task> callback) where T : IDistributedApplicationEvent
+    {
+        var subscription = new DistributedApplicationEventSubscription(async (@event, ct) =>
```
karolz-ms, line 20: Should we synchronize access to subscription list in case this is called concurrently?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: @eerhardt is right. If by "container started" we want to convey "it is running and there is good chance that it will reply to requests", then one needs to wait for `Container.Status.State` property https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Dcp/Model/Container.cs#L259 to become "Running". And in future, an even better event to have would be when the Container becomes healthy.

src/Aspire.Hosting/Eventing/IDistributedApplicationEvent.cs
```diff
@@ -0,0 +1,8 @@
+namespace Aspire.Hosting.Eventing;
+
+/// <summary>
+/// Represents an event that is published during the lifecycle of the AppHost.
+/// </summary>
+public interface IDistributedApplicationEvent
```
davidfowl, line 6: Should this have the resource or should there be a derived event that has the resource? (IDistributedApplicationResourceEvent)

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <typeparam name="T">The type of the event.</typeparam>
+    /// <param name="callback">A callback to handle the event.</param>
+    /// <returns>A subscription instance which can be used to unsubscribe </returns>
+    DistributedApplicationEventSubscription Subscribe<T>(Func<T, CancellationToken, Task> callback) where T : IDistributedApplicationEvent;
+
+    /// <summary>
+    /// Unsubscribe from an event.
+    /// </summary>
+    /// <param name="subscription">The specific subscription to unsubscribe.</param>
+    void Unsubscribe(DistributedApplicationEventSubscription subscription);
```
davidfowl, line 20: Make the DistributedApplicationEventSubscription, IDisposable instead?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: This is just an example event the idea of the pub sub model is that we can fire specific events at lots of different places. In the scenario you've listed above we'd probably listen to the resource notifications coming back.

The reality is that even if we fire the event a bit later there is no guarantee it'll be live for servicing requests.

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
mitchdenny: It is a good question. I think whether it runs serially or not is something that the publisher of the event gets to decide and not something that we enforce in eventing itself.

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
mitchdenny: Yep ... we will need to copy the list.

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
mitchdenny: I think the Event Hubs team will send us hate mail if we do that ;)

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
mitchdenny: Yeah good question. I think the assumption would be that all subscribers have returned a completed task.

src/Aspire.Hosting/Eventing/IDistributedApplicationEvent.cs
mitchdenny: See latest push.

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
mitchdenny: Even if we make it disposable we need the unsubscribe :)

src/Aspire.Hosting/Eventing/IDistributedApplicationEventing.cs
karolz-ms: That's too funny 😆

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
eerhardt: I think I'm confused at why I would use this new `ContainerResourceStartedEvent` capability over listening to the resource notifications. Getting an event before something is ready to use isn't great.

We are hitting these situations all over in our tests - needing to know when something is ready, or stopped. See

https://github.com/dotnet/aspire/blob/0d8b294cf7d3423e9b4a6473b09166f4c46e9c7a/tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs#L67-L70

https://github.com/dotnet/aspire/issues/4878

Are you saying these new, higher-level events would be built on top of the ResourceNotificationService? Like that "WaitForTextAsync" is today?

https://github.com/dotnet/aspire/blob/0d8b294cf7d3423e9b4a6473b09166f4c46e9c7a/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs#L59-L80

So we'd have 2 levels of "events"? Lower level `ResourceNotificationService`. And then higher-level `IDistributedApplicationEventing`.

mitchdenny: The way to think of this new API is a more flexible version of lifecycle events which adds the ability to publish and subscribe to arbitrary events without having to implement an interface and inject it into DI.

I would expect that some extensions would use the feed from the resource notification service to fire more specific events related to a particular resource.

davidfowl: One big difference with RNS is that the events are synchronous, as in, they run I’ll inline. RNS publishes events and they can’t be waited on inline and that means the publisher can’t react to mutations during the event nor can it wait for the event to complete before progressing the workflow.

mitchdenny:
@radical this (build failures) looks like it is related to building and testing the playground projects?

https://dev.azure.com/dnceng-public/public/_build/results?buildId=773478&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=5138e247-88cb-5ba0-8eee-e2fdd127e7e8&l=641

mitchdenny:
Blocked on: https://github.com/dotnet/aspire/pull/5273

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -226,20 +233,62 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
...
         {
-            _innerBuilder.Services.AddLifecycleHook<ContainerRegistryHook>();
+            Eventing.Subscribe<BeforeStartEvent>((e, ct) => UpdateContainerRegistryAsync(options));
         }
 
         _innerBuilder.Services.AddSingleton(ExecutionContext);
         LogBuilderConstructed(this);
     }
 
+    private Task MutateHttp2TransportAsync(BeforeStartEvent beforeStartEvent, CancellationToken cancellationToken)
```
davidfowl, line 251: nit: I'm not loving these being methods inside of this class.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -226,20 +233,62 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
...
+            resourceWithContainerImage.Annotation.Registry = options.ContainerRegistryOverride;
+        }
+
+        return Task.CompletedTask;
+    }
+
+    private Task ExcludeDashboardFromManifestAsync(BeforeStartEvent beforeStartEvent, CancellationToken cancellationToken)
+    {
+        // When developing locally, exclude the dashboard from the manifest. This only affects our playground projects in practice.
+        if (Resources.SingleOrDefault(r => StringComparers.ResourceName.Equals(r.Name, KnownResourceNames.AspireDashboard)) is { } dashboardResource)
```
davidfowl, line 284: Seeing this it makes we wonder if we should pass the model as part of the event data for parity.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: Yeah some of these global scenarios feel better as a lifecycle hook mostly for code structuring reasons. We can put these things into a separate class possibly with an external method but that seems like overkill in the same way that a lifecycle hook does.

tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
```diff
@@ -15,7 +15,11 @@
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)src\Aspire.Hosting.Redis\RedisContainerImageTags.cs" />
-    <Compile Include="$(RepoRoot)src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
+    <Compile Remove="C:\Code\aspire\src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
+    <Compile Remove="C:\Code\aspire\src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
+    <Compile Remove="C:\Code\aspire\src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
+    <Compile Remove="C:\Code\aspire\src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
+    <Compile Remove="C:\Code\aspire\src\Aspire.Hosting.Redis\RedisCommanderConfigWriterHook.cs" />
```
davidfowl, line 22: Remove
```suggestion
```

mitchdenny: what the heck ... how did they get in there?

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: updated events.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: I've moved them out into a separate internal class (grouped together just for some structuring). Also made them static since they just rely on the event itself now for everything they need.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -174,6 +172,63 @@ public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResou
             configureContainer(surrogateBuilder);
         }
 
+#pragma warning disable ASPIREEVENTING001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
```
eerhardt, line 175: Should we just disable this warning globally? It adds a lot of noise, and we are aware that it is an experimental API.

src/Aspire.Hosting/Eventing/DistributedApplicationEventSubscription.cs
```diff
@@ -0,0 +1,30 @@
+using System.Diagnostics.CodeAnalysis;
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting.Eventing;
+
+/// <summary>
+/// Represents a subscription to an event that is published during the lifecycle of the AppHost.
+/// </summary>
+/// <param name="callback">Callback to invoke when the event is published.</param>
+[Experimental("ASPIREEVENTING001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
```
eerhardt, line 10: Do we really need it to be `ASPIREEVENTING001`? How many "eventing" warnings do we image there will be?

Instead, should it just be `ASPIRE00x`?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
mitchdenny: Done.

src/Aspire.Hosting/Eventing/DistributedApplicationEventSubscription.cs
mitchdenny: I like them to be a bit more verbose because it means you don't need to lookup what it means :)

davidfowl (approved):

PR #5230, opened 2024-08-08
Allow NoWarn on ProjectReference to supress ASPIRE004

CZEMacLeod:
## Description

Allow the suppression of the `ASPIRE004` error message per `ProjectReference` using the `NoWarn` property.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5230)

### DIFF
--- a/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
+++ b/src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
@@ -171,7 +171,7 @@ namespace Projects%3B
     </GetNonExecutableReferences>
 
     <Warning Code="ASPIRE004"
-             Condition="'@(_AspireNonExecutableProjectResource)' != ''"
+             Condition="'@(_AspireNonExecutableProjectResource)' != '' AND (%(_AspireNonExecutableProjectResource.NoWarn)!='ASPIRE004')"
              Text="'%(_AspireNonExecutableProjectResource.OriginalItemSpec)' is referenced by an Aspire Host project, but it is not an executable. Did you mean to set IsAspireProjectResource=&quot;false&quot;?" />
   </Target>
 


eerhardt:
Can you describe the scenario where you would want to suppress this warning?

CZEMacLeod:
@eerhardt Yes - a project reference to a class library (or in my case an IIS Express SDK project) where you want the project information to be able to use it in a resource but it is not executable.
```xml
  <ItemGroup>
    <ProjectReference Include="..\..\..\src\C3D\Extensions\Aspire\IISExpress\C3D.Extensions.Aspire.IISExpress.csproj" IsAspireProjectResource="false" />
    <ProjectReference Include="..\..\..\src\C3D\Extensions\Aspire\WaitFor\C3D.Extensions.Aspire.WaitFor.csproj" IsAspireProjectResource="false" />
    <ProjectReference Include="..\EF6\EF6WebApp.csproj" NoWarn="ASPIRE004"/>
  </ItemGroup>
```

This is similar to suppressing nuget package warning etc.
You could do a global `<NoWarn>ASPIRE004</NoWarn>` (although I haven't tested that) - but then you might not notice when you add a library that should be `IsAspireProjectResource="false"`.
Maybe an alternative could be if you explicitly do `IsAspireProjectResource="true"` then it wouldn't warn, but this matched the existing mechanism (which I tried first and found it didn't work here).

eerhardt:
> Maybe an alternative could be if you explicitly do IsAspireProjectResource="true" then it wouldn't warn 

I think that approach would be better. I'm not sure the "NoWarn" on the ProjectReference is a gesture used anywhere else. So I wouldn't want to invent it here. Also the way the current PR is, it only supports a single "NoWarn", not a list.

Also, can you add a test for this scenario? Here is where the existing tests are:

https://github.com/dotnet/aspire/blob/583014c2884952abbaeaf5052aa64155b72176bd/tests/Aspire.Hosting.Tests/MSBuildTests.cs#L12-L16

CZEMacLeod:
@eerhardt NoWarn is metadata used on `PackageReference` and `ProjectReference` for nuget to suppress NU1701 etc. type errors. My issue is similar in style to https://github.com/NuGet/Home/issues/5999

I'll have a look at the tests and see what I can do.
Yes - I did implement it only for a single error, rather than a list - but I felt a simple way of doing it was enough as there is only the one error you need to deal with just now. If additional warnings were added in the future, the condition could be updated.

I'm struggling with the running a local version of aspire just now, so it may be a while before I can get to dealing with the tests.
(And I've broken the aspire workload for all development!)

CZEMacLeod:
Oh and another area where this would be useful is https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/585
They specifically mention the `ASPIRE004` error there.

danmoseley:
@CZEMacLeod I'm going to close this as its stale, feel free to reopen if you're able to add the test.

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
```diff
@@ -171,7 +171,7 @@ namespace Projects%3B
     </GetNonExecutableReferences>
 
     <Warning Code="ASPIRE004"
-             Condition="'@(_AspireNonExecutableProjectResource)' != ''"
+             Condition="'@(_AspireNonExecutableProjectResource)' != '' AND (%(_AspireNonExecutableProjectResource.NoWarn)!='ASPIRE004')"
```
danmoseley, line 174: nit, spacing around !=
do we normally have AND in upper case

PR #5233, opened 2024-08-08
Improve notification loop exit

sebastienros:
- Cancel `WaitAsync` loop when text is found.
- Ignore `OperationCanceledException` as they are expected.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5233)

### DIFF
--- a/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
+++ b/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
@@ -1,13 +1,11 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using IResource = Aspire.Hosting.ApplicationModel.IResource;
-using ResourceLoggerService = Aspire.Hosting.ApplicationModel.ResourceLoggerService;
-using ResourceNotificationService = Aspire.Hosting.ApplicationModel.ResourceNotificationService;
-
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using ResourceLoggerService = Aspire.Hosting.ApplicationModel.ResourceLoggerService;
+using ResourceNotificationService = Aspire.Hosting.ApplicationModel.ResourceNotificationService;
 
 namespace Aspire.Hosting.Tests.Utils;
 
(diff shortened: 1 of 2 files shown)


tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -77,21 +74,23 @@ await foreach (var resourceEvent in resourceNotificationService.WatchAsync(cance
                 if (loggingResourceIds.Add(resourceId))
                 {
                     // Start watching the logs for this resource ID
-                    logWatchTasks.Add(WatchResourceLogs(tcs, resourceId, logTexts, resourceLoggerService, cancellationToken));
+                    logWatchTasks.Add(WatchResourceLogs(tcs, resourceId, logTexts, resourceLoggerService, cancellationTokenSource));
```
radical, line 77: Are we using `logWatchTasks` here to root them to the method so they don't get collected?

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -77,21 +74,23 @@ await foreach (var resourceEvent in resourceNotificationService.WatchAsync(cance
...
                 }
             }
-
-            await Task.WhenAny(logWatchTasks).ConfigureAwait(false);
+        }
+        catch (OperationCanceledException)
```
radical, line 81: Is this restricted to the case where `cancellationTokenSource.IsCancellationRequested==true`?

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -65,9 +62,9 @@ private static async Task WatchNotifications(DistributedApplication app, IResour
 
         try
         {
-            await foreach (var resourceEvent in resourceNotificationService.WatchAsync(cancellationToken).ConfigureAwait(false))
+            await foreach (var resourceEvent in resourceNotificationService.WatchAsync(cancellationTokenSource.Token).ConfigureAwait(false))
```
radical, line 65: Just curious - why do we continue watching for resource notifications once we have found one matching `resourceName`?

Would these resource events be like - starting up, running, stopped etc? What happens if a resource starts, we call `WatchResourceLogs`, then it crashes, and dcp starts it up again? Do we need to call `WatchResourceLogs` again?
Does the `resourceEvent.ResourceId` stay the same over multiple notification messages, and even with restarts?

sebastienros: There could actually be multiple "resourceId" for a single resource. Example when we look for "oracle" there is another one like "oracle-xxxx-yyyy" for the docker container. And they will both match the resource name.

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
sebastienros: I believe these would just be on the ThreadPool so they wouldn't be collected, but not sure. Feels safer this way, unless someone says it's fine without.

radical (approved):

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
```diff
@@ -1,13 +1,11 @@
...
 
-using IResource = Aspire.Hosting.ApplicationModel.IResource;
-using ResourceLoggerService = Aspire.Hosting.ApplicationModel.ResourceLoggerService;
-using ResourceNotificationService = Aspire.Hosting.ApplicationModel.ResourceNotificationService;
-
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using ResourceLoggerService = Aspire.Hosting.ApplicationModel.ResourceLoggerService;
+using ResourceNotificationService = Aspire.Hosting.ApplicationModel.ResourceNotificationService;
```
eerhardt, line 8: Why are these `using X = ...` necessary?

eerhardt (approved):
LGTM

tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs
sebastienros: I think it's because at some point there was a conflict between an included file and a reference one. Not necessary anymore so removing.

PR #5235, opened 2024-08-09
Adds improved default auth to Milvus

### LINKED ISSUE #4966: Modify Milvus hosting resource to use generated password by default

### Describe the bug

Miluvs changed to support config-based password default (instead of global default) (see: https://github.com/milvus-io/milvus/issues/33058). This bug tracks modifying reacting to that change in the Aspire.Hosting.Milvus package to enable a generated default password if none provided, similar to other resources.  This will only work on a later tagged image of the milvus container image. /cc @eerhardt

timheuer:
- Updates Milvus hosting to leverage dynamically generated default password if none provided (similar to other resources) and resulting cascading changes + tests.
- Updates tags for contianer image that support setting auth

## Description

Milvus added the ability to set the root password. This change adapts to that and follows similar pattern to other config-based auth (e.g., MySql is a good analog). This requires the support of a minimum new Milvus version (2.4.7). 

Fixes #4966

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] No. Follow-up changes expected. - haven't seen a pattern where we've made a change that requires a specific new resource version
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No - follows same pattern as existing pattern/resources
  - [ ] No
  - [x] Yes
    - Link to aspire-docs issue: [dotnet/docs-aspire#1475](https://github.com/dotnet/docs-aspire/issues/1475)
  - [ ] No
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5235)


timheuer:
@eerhardt had to redo the PR (my branch was way stale).  Addressed your prior feedback.

Review: this does change resource and enables a null param were there wasn't before. Not sure if we've made a similiar change there -- seeking feedback on desired correct way here.

eerhardt:
> Review: this does change resource and enables a null param were there wasn't before. Not sure if we've made a similiar change there -- seeking feedback on desired correct way here.

I believe you did it correctly. 👍

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
```diff
@@ -13,6 +13,7 @@
     <AspireProjectOrPackageReference Include="Aspire.Hosting.Azure.CosmosDB" />
     <AspireProjectOrPackageReference Include="Aspire.Hosting.Azure.EventHubs" />
     <AspireProjectOrPackageReference Include="Aspire.Hosting.AppHost" />
+    <AspireProjectOrPackageReference Include="Aspire.Hosting.Milvus" />
```
eerhardt, line 16: As part of the work for https://github.com/dotnet/aspire/issues/4294, we are now doing these functional, end-to-end tests in the Aspire.Hosting.X.Tests functional tests. So you can remove these changes to the `testproject` and Aspire.Hosting.Tests. We don't need to test all the integrations in our core tests anymore.

tests/Aspire.Hosting.Milvus.Tests/MilvusPublicApiTests.cs
```diff
@@ -35,20 +35,6 @@ public void AddMilvusContainerShouldThrowWhenNameIsNull()
         Assert.Equal(nameof(name), exception.ParamName);
     }
 
-    [Fact]
-    public void AddMilvusContainerShouldThrowWhenApiKeyIsNull()
```
eerhardt, line 39: Can you update 

https://github.com/dotnet/aspire/blob/a7f23ec23cb810adff8c44969be7fcb051b6b871/tests/Aspire.Hosting.Milvus.Tests/MilvusFunctionalTests.cs#L35-L37

To no longer pass in an "apikey" parameter? As that is the default case now.

Then we should have coverage for both passing a parameter (all the existing tests) and for not passing a parameter (in that functional test).

src/Aspire.Hosting.Milvus/MilvusServerResource.cs
```diff
@@ -40,7 +40,7 @@ public MilvusServerResource(string name, ParameterResource apiKey) : base(name)
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
        ReferenceExpression.Create(
-            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Url)};Key={ApiKeyParameter}");
+            $"Endpoint={PrimaryEndpoint.Property(EndpointProperty.Url)};Key=root:{ApiKeyParameter}");
```
eerhardt, line 43: This is a small breaking change to existing apps. But I think it makes sense going forward.

eerhardt, line 43: When someone goes to deploy their app, are there Milvus endpoints they can point to? I'm wondering how the `root` gets replaced to some other user.

timheuer: We have a similar pattern for MySql https://github.com/dotnet/aspire/blob/ba949b9ca1fa2379e83da648a3e5c822fdce1437/src/Aspire.Hosting.MySql/MySqlServerResource.cs#L41 -- I assume the answer here would be they would use the Client config and set the Key param there more specifically. Similarly like MySql, Milvus could be configured to add more users with roles, etc. and then specify the desired key for actions. However, this is only configurable post startup in Milvus. root is always defined initially.

tests/testproject/TestProject.AppHost/TestProject.AppHost.csproj
timheuer: Will do, I saw redis and a postgres still in there so wasn't sure.

eerhardt: Yeah - my plan is to only have a core set of extensions (< 5) in the end-to-end test. The test is still good, we just don't need it to grow indefinitely.

radical:
Would this allow adding Milvus playground app to the tests like in https://github.com/dotnet/aspire/pull/5244 ? When I had tried it earlier on `main` it was failing with `Parameter resource could not be used because configuration key 'Parameters:milvusauth' is missing and the Parameter has no default value.`.

timheuer:
Any other feedback @eerhardt?

eerhardt (approved):
This looks good to me. Nice work! Much easier to get started with Milvus.

Issue #5247, opened 2024-08-09
Some tests timing out on helix


radical:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=804618&view=results
Build error leg or test failing: Aspire.Hosting.Elasticsearch.Tests.WorkItemExecution
Pull request: https://github.com/dotnet/aspire/pull/5243

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "Aborting test run: test run timeout of [0-9]+ milliseconds exceeded",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=804618
**Error message validated:** `[Aborting test run: test run timeout of [0-9]+ milliseconds exceeded`]
**Result validation:** :white_check_mark: Known issue matched with the provided build.
**Validation performed at:** 9/11/2024 7:28:44 PM UTC

#### Summary
|24-Hour Hit Count|7-Day Hit Count|1-Month Count|
|---|---|---|
|0|0|0|

radical:
cc @eerhardt @sebastienros 

One of the tests timing out is `Aspire.Hosting.Elasticsearch.Tests` - [log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5243-merge-b8dff7d94dac4dbb8a/Aspire.Hosting.Elasticsearch.Tests/1/console.613f6e41.log?helixlogtype=result).

Why is this timing out after 10minutes on helix when all the tests combined didn't take that long on the build machine?
- One reason could be that the helix agents are under-powered compared to the build machines. And these tests start up a new app per xunit test in `ElasticsearchFunctionalTests`.

radical:
I chose a broader error message to match against, so we can get any tests that are timing out on helix now.

mitchdenny:
I am still hitting this today on this PR: https://github.com/dotnet/aspire/pull/5223

radical:
> I am still hitting this today on this PR: #5223

I'll bump the timeouts. We are hitting new ones as we moved new tests to helix last week.

radical:
[...]

PR #5270, opened 2024-08-12
[tests] Bump timeout for Aspire.Hosting.Elasticsearch.Tests to 20 mins

radical:
## Description

Contributes to https://github.com/dotnet/aspire/issues/5247 .

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5270)

### DIFF
--- a/tests/helix/send-to-helix-basictests.targets
+++ b/tests/helix/send-to-helix-basictests.targets
@@ -10,6 +10,9 @@
     <ItemGroup>
       <!-- needed for Aspire.Hosting.Container.Tests -->
       <HelixPreCommand Include="$(_EnvVarSetKeyword) DOCKER_BUILDKIT=1" />
+
+      <_TestRunCommandArguments Condition="'$(OS)' != 'Windows_NT'" Include="-- RunConfiguration.TestSessionTimeout=$TEST_TIMEOUT" />
+      <_TestRunCommandArguments Condition="'$(OS)' == 'Windows_NT'" Include="-- RunConfiguration.TestSessionTimeout=%TEST_TIMEOUT%" />
     </ItemGroup>
 
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


tests/helix/send-to-helix-basictests.targets
```diff
@@ -21,10 +24,14 @@
 
     <ItemGroup>
       <_DefaultWorkItems Include="$(WorkItemArchiveWildCard)" />
+      <!-- runsettings timeout in ms, default to 10 mins -->
+      <_DefaultWorkItems TimeoutMs="600000" />
+
+      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Hosting.Elasticsearch.Tests'" TimeoutMs="1200000" />
```
sebastienros, line 30: This can't be done with a runsettings file? If this is done centrally here this will be missed out.

radical: It can, but you would need a copy of the `.runsettings` file for any project that is overriding it. I have some changes planned that would give the projects more control over their own helix run. But this should get us closer to green CI for now.

sebastienros: Ok, as long as you intend to de-centralize it later. Something like <HelixTimeout> in projects would be great.

sebastienros (approved):

tests/helix/send-to-helix-basictests.targets
eerhardt, line 30: This is a really long time. Do we know why it is taking that long?

radical: I think because elasticsearch takes a few minutes to start up, and in the functional tests we are starting that up per test.

radical: That's what it looked like from the log. The helix agents is underpowered compared to the build machine, so it would affect it too.

PR #5294, opened 2024-08-14
[tests] Add tests for more Playground apps

radical:
- Add `milvus` playground app to the tests.
- Skip the redis commander in `TestShop` for tests, as it requires connecting to other containers which is not supported on our CI yet. And we don't really connect, and check the web app.
- Reduce log level to `debug` from `trace`.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5294)

### DIFF
--- a/playground/TestShop/TestShop.AppHost/Program.cs
+++ b/playground/TestShop/TestShop.AppHost/Program.cs
@@ -6,11 +6,14 @@ var catalogDb = builder.AddPostgres("postgres")
                        .AddDatabase("catalogdb");
 
 var basketCache = builder.AddRedis("basketcache")
-                         .WithDataVolume()
-                         .WithRedisCommander(c =>
-                         {
-                             c.WithHostPort(33801);
-                         });
+                         .WithDataVolume();
+
+#if !SKIP_DASHBOARD_REFERENCE
+basketCache.WithRedisCommander(c =>
+                     {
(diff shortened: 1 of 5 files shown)


playground/TestShop/TestShop.AppHost/Program.cs
```diff
@@ -6,11 +6,14 @@
...
 var basketCache = builder.AddRedis("basketcache")
-                         .WithDataVolume()
-                         .WithRedisCommander(c =>
-                         {
-                             c.WithHostPort(33801);
-                         });
+                         .WithDataVolume();
+
+#if !SKIP_DASHBOARD_REFERENCE
+basketCache.WithRedisCommander(c =>
```
sebastienros, line 12: indent

sebastienros (approved):

playground/TestShop/TestShop.AppHost/Program.cs
radical: The file has top-level statements, so zero indent. Or did you mean for the `{..}` block? Could you share the suggested indent in that case, and I can commit that?

radical: I can change in a follow up.

PR #5296, opened 2024-08-15
Improve support for non-primitive OTEL values

### LINKED ISSUE #4506: Dashboard: OTLP AnyValue is mapped to a string value which is insufficient for object graphs
The OTLP proto uses the type `AnyValue` throughout the type definitions when it is expressing a value. `AnyValue` is a variant equivalent, used to map strings, numbers booleans etc. From .NET telemetry production, today, we only use those types. However the proto spec for `AnyValue` can also contain arrays and maps. 

From [common.proto](https://github.com/dotnet/aspire/blob/bda316874feb6087ffee04eb6b0529574d40cf79/src/Aspire.Dashboard/Otlp/opentelemetry/proto/common/v1/common.proto#L28C1-L67C2): 
``` proto
// AnyValue is used to represent any type of attribute value. AnyValue may contain a
// primitive value such as a string or integer or it may contain an arbitrary nested
// object containing arrays, key-value lists and primitives.
message AnyValue {
  // The value is one of the listed fields. It is valid for all values to be unspecified
  // in which case this AnyValue is considered to be "empty".
  oneof value {
    string string_value = 1;
    bool bool_value = 2;
    int64 int_value = 3;
    double double_value = 4;
    ArrayValue array_value = 5;
    KeyValueList kvlist_value = 6;
    bytes bytes_value = 7;
  }
}

// ArrayValue is a list of AnyValue messages. We need ArrayValue as a message
// since oneof in AnyValue does not allow repeated fields.
message ArrayValue {
  // Array of values. The array may be empty (contain 0 elements).
  repeated AnyValue values = 1;
}

// KeyValueList is a list of KeyValue messages. We need KeyValueList as a message
// since `oneof` in AnyValue does not allow repeated fields. Everywhere else where we need
// a list of KeyValue messages (e.g. in Span) we use `repeated KeyValue` directly to
// avoid unnecessary extra wrapping (which slows down the protocol). The 2 approaches
// are semantically equivalent.
message KeyValueList {
  // A collection of key/value pairs of key-value pairs. The list may be empty (may
  // contain 0 elements).
  // The keys MUST be unique (it is not allowed to have more than one
  // value with the same key).
  repeated KeyValue values = 1;
}

// KeyValue is a key-value pair that is used to store Span attributes, Link
// attributes, etc.
message KeyValue {
  string key = 1;
  AnyValue value = 2;
}
```

The dashboard typically is converting instances of AnyValue to string [using](https://github.com/dotnet/aspire/blob/bda316874feb6087ffee04eb6b0529574d40cf79/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs#L74C13-L83C11) 
``` c#
    public static string GetString(this AnyValue value) =>
        value.ValueCase switch
        {
            AnyValue.ValueOneofCase.StringValue => value.StringValue,
            AnyValue.ValueOneofCase.IntValue => value.IntValue.ToString(CultureInfo.InvariantCulture),
            AnyValue.ValueOneofCase.DoubleValue => value.DoubleValue.ToString(CultureInfo.InvariantCulture),
            AnyValue.ValueOneofCase.BoolValue => value.BoolValue ? "true" : "false",
            AnyValue.ValueOneofCase.BytesValue => value.BytesValue.ToHexString(),
            _ => value.ToString(),
        };
```
This doesn't account for these other types. 

The simple/hack fix would be to also convert these to strings, but we probably need to consider storing attributes with an equivalent to `AnyValue`, and have appropriate display functionality that can display a graph of values.

I am expecting this to become an issue with Event definitions in OTLP - which overlay over the logs specification. https://github.com/open-telemetry/oteps/blob/main/text/0202-events-and-logs-api.md. Events body will be an AnyValue which is typically going to be a dictionary/map.

JamesNK:
## Description

OTLP data such as attributes on telemetry can be complex data. Realistically, almost everyone uses strings or other primitive types for attributes, but we should handle complex data.

Internally the dashboard stores attributes as strings, so we need a good conversion process when converting incoming values to strings.

This PR:

* Improves how complex data, such as objects and arrays, are serialized to JSON
* Adds tests around creating strings from OTLP values

Example of a simple array with a boolean:

* Before: `{ "arrayValue": { "values": [ { "boolValue": true } ] } }`
* After: `[true]`

Fixes https://github.com/dotnet/aspire/issues/4506

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5296)

### DIFF
--- a/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs
+++ b/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs
@@ -108,6 +108,8 @@ public static class OtlpHelpers
 
     private static JsonNode? ConvertAnyValue(AnyValue value)
     {
+        // Recursively convert AnyValue types to JsonNode types to produce more idiomatic JSON.
+        // Recursing over incoming values is safe because Protobuf serializer imposes a safe limit on recursive messages.
         return value.ValueCase switch
         {
             AnyValue.ValueOneofCase.StringValue => JsonValue.Create(value.StringValue),
--- a/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs
+++ b/src/Aspire.Dashboard/Otlp/Model/OtlpHelpers.cs
@@ -5,6 +5,8 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Nodes;
 using Aspire.Dashboard.Configuration;
 using Aspire.Dashboard.Otlp.Storage;
 using Google.Protobuf;
@@ -16,6 +18,11 @@ namespace Aspire.Dashboard.Otlp.Model;
 
 public static class OtlpHelpers
 {
+    private static readonly JsonSerializerOptions s_jsonSerializerOptions = new JsonSerializerOptions
+    {
+        WriteIndented = false
+    };
+
     public static ApplicationKey GetApplicationKey(this Resource resource)
     {
         string? serviceName = null;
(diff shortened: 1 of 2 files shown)


samsp-msft (approved):
Solves a problem in a neat way.

samsp-msft:
fixes #4506, especially when combined with the long string visualizer. Eg converting an AnyValue map to a json string gives a good visualization without having to invent a nested tabular display etc. If the dev want's to copy or compare the values, having json is pretty good.

PR #5302, opened 2024-08-15
Populate dashpages from configuration

### LINKED ISSUE #5289: Populate dashpages from configuration
Initially, the set of dashpages will come from the dashboard's configuration. We should populate the relevant view models from that configuration when needed.

drewnoakes:
Fixes #5289

This change causes the dashboard to look for a file named `dashpages.json` in the current directory. If found, it's parsed to provide the set of dashpages shown on the metrics page.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] No. Follow-up changes expected. (this is a PR to a feature branch)
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No (not yet, we will doc dashpages entirely before the feature branch merges)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5302)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -29,6 +29,7 @@ using Microsoft.AspNetCore.Server.Kestrel;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Options;
 using Microsoft.FluentUI.AspNetCore.Components;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
@@ -218,6 +219,9 @@ public sealed class DashboardWebApplication : IAsyncDisposable
         builder.Services.AddScoped<ISessionStorage, SessionBrowserStorage>();
 
         builder.Services.AddScoped<DimensionManager>();
+
+        builder.Services.AddSingleton<IFileProvider>(new PhysicalFileProvider(Directory.GetCurrentDirectory()));
+        builder.Services.AddSingleton<IDashpagePersistence, DashpageJsonFilePersistence>();
 
         builder.Services.AddLocalization();
 
--- a/src/Aspire.Dashboard/Model/DashpageModels.cs
+++ b/src/Aspire.Dashboard/Model/DashpageModels.cs
@@ -1,0 +1,107 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Immutable;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.FileProviders;
+using static Aspire.Dashboard.Components.Pages.Metrics;
+
+namespace Aspire.Dashboard.Model;
+
+/// <summary>
+/// Data about a dashpage.
+/// </summary>
+public sealed class DashpageDefinition
+{
+    /// <summary>
+    /// Gets the unique name of this dashpage.
+    /// </summary>
+    /// <remarks>
+    /// Also used as a display name.
+    /// </remarks>
+    [JsonPropertyName("name")]
+    public required string Name { get; init; }
+
+    /// <summary>
+    /// Gets the relative priority of this dashpage, controlling ordering of dashpages in UI.
+    /// </summary>
+    /// <remarks>
+    /// Higher priorities appear higher in lists.
+    /// When priorities are equal, dashpages are ordered by name.
+    /// </remarks>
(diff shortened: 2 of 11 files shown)


tests/Aspire.Dashboard.Tests/Model/DashpageJsonFilePersistenceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                            "instrument": "envoy.connection.count",
+                            "required": true,   // non-default
+                            "resource": "envoy" // non-default
+                        }
+                    ]
+                }
+            ]
+            """;
+
+        var dashpages = DashpageJsonFilePersistence.Deserialize(json);
```
radical, line 57: Can the test use ` DashpageJsonFilePersistence.GetDashpagesAsync` instead since that is the public API?

drewnoakes: That requires writing a file or mocking the file provider which I wanted to avoid.

radical: Writing a json file to be loaded by the public api? It can have the same contents as in this existing test, right?

src/Aspire.Dashboard/Model/DashpageModels.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            if (fileInfo.Exists == false)
+            {
+                _dashpages = [];
+            }
+            else
+            {
+                using Stream stream = fileInfo.CreateReadStream();
+
+                using StreamReader reader = new(stream, Encoding.UTF8, leaveOpen: true);
```
radical, line 92: Just curious - why is `leaveOpen: true` needed here?

drewnoakes: Either way would be fine. I prefer to have the `using Stream` close the stream, and the reader not actually touch the stream when disposed, but closing the stream twice would also be safe.

tests/Aspire.Dashboard.Tests/Model/DashpageJsonFilePersistenceTests.cs
drewnoakes: The main thing I want to test is the transformation between string and object model. The other API tests more layers than needed here.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -219,6 +220,9 @@ public DashboardWebApplication(
 
         builder.Services.AddScoped<DimensionManager>();
 
+        builder.Services.AddSingleton<IFileProvider>(new PhysicalFileProvider(Directory.GetCurrentDirectory()));
```
adamint, line 223: why is this necessary?

src/Aspire.Dashboard/dashpages.json
```diff
@@ -0,0 +1,38 @@
+[
+  /* Comments are ignored */
```
adamint, line 2: What do you think about documenting this schema elsewhere, and removing the comments? It is a good practice to avoid commenting JSON, even though our deserializer currently handles it.

adamint (approved):

src/Aspire.Dashboard/DashboardWebApplication.cs
drewnoakes: We're the first consumer of the `IFileProvider` abstraction. It allows the other service `DashpageJsonFilePersistence` access to the file system.

src/Aspire.Dashboard/dashpages.json
drewnoakes: Comments were just because I copied this from a unit test where I wanted to validate that comments are tolerated.

Adding a schema makes sense.

drewnoakes:
`Aspire.EndToEnd.Tests.IntegrationServicesTests.Aspire.EndToEnd.Tests.IntegrationServicesTests.VerifyCosmosComponentWorks(resourceName: efcosmos)` timed out after 10 minutes.

src/Aspire.Dashboard/dashpages.json
adamint: I do not think we should officially support JSON with comments. It is not valid JSON.

drewnoakes: It's very common to want to document your configuration. It's not very friendly to disallow that when it's so easy to permit it. I see no downside.

drewnoakes: For example, the .NET Project System ignores comments in `launchSettings.json` files.

PR #5311, opened 2024-08-15
Add resource relationship to app model

### LINKED ISSUE #2595: Graph view of resource relationships
I think it would be pretty useful to see what resources a resource is referencing in the dashboard.

```cs
       .WithReference(basketService)
       .WithReference(catalogService.GetEndpoint("http"));
```

In the example above, the `frontend` resource would have two references. A reference would have:

* Resource name, e.g. `catalogservice`
* Endpoint name, e.g. `http`
* Value - URI string?, e.g. `http://localhost:5653/`

The resource model today doesn't have references, just the env vars used by the app. You could try to reverse engineer references by looking at environment variables, but it would be a pain in the butt. It is much easier to model it explicitly and display values from the model.

Edit: We could also display the reverse. What resources reference the current resource, e.g. `basketservice` is referenced by `frontend` and `apigateway`.

In the UI, call it dependencies and used by? That's what NuGet does:

* Dependencies - https://www.nuget.org/packages/Grpc.Net.Client#dependencies-body-tab
* Used by - https://www.nuget.org/packages/Grpc.Net.Client#usedby-body-tab

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/2595

This PR adds:

* Relationship annotation and `WithRelationship` extension method to app model
* `WithRelationship` calls are in various methods that add or reference resources.
* Relationship field to resource service proto contract
* Relationship field to `ResourceViewModel`.
* UI to resource details to display `References` and `Back references`
* Color to resource name on resources page and resource names in detail's references UI

image

image

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5311)

### DIFF
--- a/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
@@ -206,6 +206,8 @@ public static class PostgresBuilderExtensions
 
             configureContainer?.Invoke(pgAdminContainerBuilder);
 
+            pgAdminContainerBuilder.WithRelationship(builder.Resource, "PgAdmin");
+
             return builder;
         }
     }
@@ -267,7 +269,6 @@ public static class PostgresBuilderExtensions
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
     public static IResourceBuilder<PostgresServerResource> WithPgWeb(this IResourceBuilder<PostgresServerResource> builder, Action<IResourceBuilder<PgWebContainerResource>>? configureContainer = null, string? containerName = null)
     {
-
         if (builder.ApplicationBuilder.Resources.OfType<PgWebContainerResource>().SingleOrDefault() is { } existingPgWebResource)
         {
             var builderForExistingResource = builder.ApplicationBuilder.CreateResourceBuilder(existingPgWebResource);
@@ -289,6 +290,8 @@ public static class PostgresBuilderExtensions
                                                .ExcludeFromManifest();
 
             configureContainer?.Invoke(pgwebContainerBuilder);
+
+            pgwebContainerBuilder.WithRelationship(builder.Resource, "PgWeb");
 
             builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>(async (e, ct) =>
             {
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Immutable;
+using Aspire.Dashboard.Model;
 using Aspire.Hosting.Dcp.Model;
 using HealthStatus = Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus;
 
@@ -105,6 +106,11 @@ public sealed record CustomResourceSnapshot
     /// </summary>
     public ImmutableArray<ResourceCommandSnapshot> Commands { get; init; } = [];
 
+    /// <summary>
+    /// The relationships to other resources.
+    /// </summary>
+    public ImmutableArray<RelationshipSnapshot> Relationships { get; init; } = [];
+
     internal static HealthStatus? ComputeHealthStatus(ImmutableArray<HealthReportSnapshot> healthReports, string? state)
     {
         if (state != KnownResourceStates.Running)
@@ -161,6 +167,13 @@ public sealed record UrlSnapshot(string Name, string Url, bool IsInternal);
 /// <param name="MountType">Gets the mount type, such as <see cref="VolumeMountType.Bind"/> or <see cref="VolumeMountType.Volume"/></param>
 /// <param name="IsReadOnly">Whether the volume mount is read-only or not.</param>
 public sealed record VolumeSnapshot(string? Source, string Target, string MountType, bool IsReadOnly);
+
+/// <summary>
+/// A snapshot of a relationship.
+/// </summary>
+/// <param name="ResourceName">The name of the resource the relationship is to.</param>
+/// <param name="Type">The relationship type.</param>
+public sealed record RelationshipSnapshot(string ResourceName, string Type);
 
 /// <summary>
 /// A snapshot of the resource property.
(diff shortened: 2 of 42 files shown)


src/Aspire.Hosting/ApplicationModel/ResourceRelationshipAnnotation.cs
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+public class ResourceRelationshipAnnotation(IResource resource, string type, string? description) : IResourceAnnotation
+{
+    public IResource Resource { get; } = resource;
+
+    public string Type { get; } = type;
+
+    public string? Description { get; } = description;
```
mitchdenny, line 14: I'm wondering whether this is going to present a localization issue?

src/Aspire.Hosting/ApplicationModel/ResourceRelationshipAnnotation.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+public class ResourceRelationshipAnnotation(IResource resource, string type, string? description) : IResourceAnnotation
```
mitchdenny, line 8: I'm not sure how I feel about the stringy `type`. I'm wondering whether it would be better to have a `ResourceRelationshipType` which can be inherited. It could then expose a way of getting a simple string value, but also provide a way of providing extra details around the relationship.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +555,16 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
     {
         return builder.WithAnnotation(ManifestPublishingCallbackAnnotation.Ignore);
     }
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+#pragma warning disable RS0016 // Add public types and members to the declared API
+    public static IResourceBuilder<T> WithRelationship<T>(
```
mitchdenny, line 561: I think we should consider having an overload of this which takes an Action<RelationshipAnnotation> which allows developers to stuff some extra data into the relationship property bag. There isn't an easy way to get to it otherwise.

JamesNK: IMO we could remove properties. Nothing uses it. And we could add it later once we identify a use for it.

mitchdenny: Thats fair. Keep the API lean until we need it. Agreed!

eerhardt: One place where something like this could be used is when we add references to Azure Resources, we could add some information about how the application is going to use the Azure Resource. For example, when adding a reference from a .NET app to an Azure Storage account, we could add that this app will only ever read from the Storage account. So when we deploy the app, we only give it "read" permissions. Similar for other resources as well - i.e. ServiceBus (which app is the reader and which is the writer).

eerhardt: I would imagine it would be modelled as information on the "relationship". Can you have an annotation on an annotation?

mitchdenny:
From an app model perspective I am happy with where this is at except for the XML doc comments ;)

JamesNK:
I think this is ready for folks to review.

UI feedback is welcome on adding colors to resources. This is the same thing the dashboard does in telemetry pages, i.e. `frontend` could be assigned a blue color that shows up in various UI including the span waterfall on trace detail.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -349,6 +349,8 @@ public static IResourceBuilder<TDestination> WithReference<TDestination>(this IR
         var resource = source.Resource;
         connectionName ??= resource.Name;
 
+        builder.WithRelationship(resource, "ConnectionString");
```
mitchdenny, line 352: WellKnownRelationshipTypes constants class?

mitchdenny:
I think this is looking good. Just a few minor things and a bit of a UX nit:

image

Why not just hyperlink the resource name?

JamesNK:
Yeah, it doesn't look that great. However, it's the UI we're using everywell else. Buttons looking a bit ugly is why I've been switching to the action column with `...` button.

I'll play around with options later. I don't want to hold this PR up on a new UI design for row links.

davidfowl:
I am not a fan on the colors, I think it is too noisy indeed.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
JamesNK: Done

JamesNK:
All feedback applied.

leslierichardson95 (approved):
Looks good to me (including the color tags)!

JamesNK:
I've already removed the colors. To avoid colors and UI discussion being a blocker, this PR can be merged without them, and then we can discuss later whether they're wanted or not.

Adding (or removing) them isn't difficult.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -998,4 +1007,30 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="resource">The resource that the relationship is to.</param>
+    /// <param name="type">The relationship type.</param>
+    /// <returns>A resource builder.</returns>
+    /// <remarks>
+    /// <para>
+    /// The <c>WithRelationship</c> method is used to add relationships to the resource. Relationships are used to link
+    /// resources together in UI. The <paramref name="type"/> indicates information about the relationship type.
+    /// </para>
+    /// </remarks>
```
mitchdenny, line 1024: nit: It would be good to have an `<example>` here as well.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -118,7 +118,9 @@ public static IResourceBuilder<PostgresDatabaseResource> AddDatabase(this IResou
 
         builder.Resource.AddDatabase(name, databaseName);
         var postgresDatabase = new PostgresDatabaseResource(name, databaseName, builder.Resource);
-        return builder.ApplicationBuilder.AddResource(postgresDatabase);
+        var databaseBuilder = builder.ApplicationBuilder.AddResource(postgresDatabase);
+
+        return databaseBuilder;
```
mitchdenny, line 123: nit: Redundant change.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -105,6 +106,12 @@ internal init
     /// </summary>
     public ImmutableArray<ResourceCommandSnapshot> Commands { get; init; } = [];
 
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+#pragma warning disable RS0016 // Add public types and members to the declared API
+    public ImmutableArray<RelationshipSnapshot> Relationships { get; init; } = [];
+#pragma warning restore RS0016 // Add public types and members to the declared API
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
```
mitchdenny, line 113: Missing docs?

mitchdenny (approved):
Few small things but this looks good. Either do those minor tweaks now or as a follow-up.

PR #5318, opened 2024-08-16
[tests] Misc fixes for playground tests

radical:
- Run AppHostRunsCleanly only for apphosts that don't have TestEndpoints defined
- Run AppHostRunsCleanly only for apphosts that don't have TestEndpoints defined
- cleanup TestEndpoints - serialization not required
- add Mongo which can get sanity checked by AppHostRunsCleanly, even though the TestEndpoints for it are blocked

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5318)

### DIFF
--- a/tests/Aspire.Playground.Tests/AppHostTests.cs
+++ b/tests/Aspire.Playground.Tests/AppHostTests.cs
@@ -3,7 +3,6 @@
 
 using System.Net;
 using System.Text;
-using System.Text.Json;
 using System.Text.RegularExpressions;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Tests.Utils;
@@ -312,11 +311,8 @@ public class AppHostTests
     }
 }
 
-public class TestEndpoints : IXunitSerializable
+public class TestEndpoints
 {
-    // Required for deserialization
-    public TestEndpoints() { }
-
     public TestEndpoints(string appHost, Dictionary<string, List<string>> resourceEndpoints, List<ReadyStateText>? waitForTexts = null)
     {
         AppHost = appHost;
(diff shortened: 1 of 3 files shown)


tests/helix/send-to-helix-buildonhelixtests.targets
```diff
@@ -27,6 +27,10 @@
     <ItemGroup>
       <_TestRunCommandArguments Remove="@(_TestRunCommandArguments)" />
 
+      <_TestBlameArguments Remove="@(_TestBlameArguments)" />
```
eerhardt, line 30: What's already in this item that need to be cleared?

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -145,9 +145,22 @@ public async Task TestEndpointsReturnOk(TestEndpoints testEndpoints)
...
+        foreach (var testEndpoint in GetAllTestEndpoints())
+        {
+            appHostsWithTestEndpoints.Add(testEndpoint.AppHost);
+        }
+
         var theoryData = new TheoryData<string>();
         foreach (var asm in appHostAssemblies)
         {
+            var appHostName = Path.GetFileNameWithoutExtension(asm);
+            if (appHostsWithTestEndpoints.Contains(appHostName))
```
eerhardt, line 159: Can you add a comment why we are filtering these?

eerhardt (approved):

tests/helix/send-to-helix-buildonhelixtests.targets
radical: https://github.com/dotnet/aspire/blob/79cdd0688ada0a5d0b0c4e3b91cc7f06353648ff/tests/helix/send-to-helix-inner.proj#L66-L70

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: I added a comment, and renamed the method so it is clearer.

tests/helix/send-to-helix-buildonhelixtests.targets
eerhardt: Why don't we want the defaults? Maybe adding a comment here could help.

radical: I had this on the commit message, which is not useful enough as it would get lost in the squashed commit: `Remove the blame-hang-timeout for playground tests, as it is pretty much one main test running`.

I'll add a comment with that expanded to:
`Remove the blame-hang-timeout for playground tests, as it is pretty much one main test running - 'TestEndpointsReturnOk'. IOW, the blame-hang timeout would have to be at least as long as the full TestSession timeout.`

Is that clearer? I'm in two minds now whether to instead use `20min` blame-hang-timeout, which is the test session timeout, so it won't really serve any purpose. But if we add more tests and the session timeout goes up, then this hang timeout would become relevant.

radical: Adding in a follow up.

PR #5319, opened 2024-08-16
[tests] Add AspireEventHub to playground tests

radical:
- And Skip playground apps that have specific tests from `AppHostRunsCleanlyForAppsWithNoOtherTests`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5319)


radical:
Depends on https://github.com/dotnet/aspire/pull/5318 .

radical:
Blocked on https://github.com/dotnet/aspire/issues/5326 .

radical:
Blocked on https://github.com/dotnet/aspire/issues/5371 .

davidfowl:
This should be unblocked

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -108,23 +114,11 @@ public async Task TestEndpointsReturnOk(TestEndpoints testEndpoints)
...
-                        resilience.Retry.MaxRetryAttempts = 30;
-                        resilience.CircuitBreaker.SamplingDuration = resilience.AttemptTimeout.Timeout * 2;
-                    });
-            });
+            using var client = CreateHttpClientWithResilience(app, resource);
 
             foreach (var path in endpoints)
             {
-                _testOutput.WriteLine($"Calling endpoint '{client.BaseAddress}{path.TrimStart('/')} for resource '{resource}' in app '{Path.GetFileNameWithoutExtension(appHostPath)}'");
+                Console.WriteLine($"Calling endpoint '{client.BaseAddress}{path.TrimStart('/')} for resource '{resource}' in app '{Path.GetFileNameWithoutExtension(appHostPath)}'");
```
eerhardt, line 121: Is this right? I would think it should stay _testOutput.

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -285,16 +318,51 @@ public static IList<TestEndpoints> GetAllTestEndpoints()
...
+            throw new XunitException($"Timed out waiting for the consumer message to be logged: '{consumerMessage}'", te);
+        }
+    }
+
     public static TheoryData<TestEndpoints> TestEndpoints()
     {
         TheoryData<TestEndpoints> theoryData = new();
         foreach (var candidateTestEndpoint in GetAllTestEndpoints())
         {
+            if (candidateTestEndpoint.Skip)
```
eerhardt, line 361: Why do we have test endpoints declared that should be skipped?

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -137,22 +131,47 @@ public async Task TestEndpointsReturnOk(TestEndpoints testEndpoints)
                 Assert.True(HttpStatusCode.OK == response.StatusCode, $"Endpoint '{client.BaseAddress}{path.TrimStart('/')}' for resource '{resource}' in app '{Path.GetFileNameWithoutExtension(appHostPath)}' returned status code {response.StatusCode}");
             }
         }
+        if (testEndpoints.WhenReady != null)
+        {
+            foreach (var resource in resourceEndpoints.Keys)
+            {
+                await testEndpoints.WhenReady(app, appHostPath, _testOutput);
+            }
```
eerhardt, line 139: Why are we firing the `WhenReady` for each `resource`, but not passing the `resource` into the method?

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -285,16 +318,51 @@ public static IList<TestEndpoints> GetAllTestEndpoints()
...
 
+    private static async Task TestEventHubsAppHost(DistributedApplication app, string appHostPath, ITestOutputHelper testOutput)
+    {
+        using var client = CreateHttpClientWithResilience(app, "api");
+
+        var consumerMessage = "Hello, from /test sent via producerClient";
+        var consumerMessageLoggedTask =
+                app.WaitForTextAsync(log => log.Contains(consumerMessage), resourceName: "consumer")
+                    .WaitAsync(TimeSpan.FromMinutes(2))
+                    .ConfigureAwait(false);
```
eerhardt, line 338: Why is this declared up here? Why not just inline it in the `await` inside the `try catch` below?

tests/Aspire.Playground.Tests/AppHostTests.cs
```diff
@@ -285,16 +318,51 @@ public static IList<TestEndpoints> GetAllTestEndpoints()
                     new ("basketservice", "Application started"),
                     new ("postgres", "database system is ready to accept connections"),
                 ]),
+            // Test endpoints disabled for AppHostRunsCleanlyForAppsWithNoOtherTests as they have specific tests
+            new TestEndpoints("WithDockerfile.AppHost", []),
+            new TestEndpoints("KafkaBasic.AppHost", []),
+            new TestEndpoints("AzureFunctionsEndToEnd.AppHost", [])
```
eerhardt, line 324: It feels like this should be a separate "exclusion" list or something. Putting these empty ones in here feels wrong.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: Correct. I had changed it for debugging. It is useful when the whole thing times out, and so you don't get a .trx with the log. I'll drop it.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: These have specific tests in `ProjectSpecificTests.cs`, so don't need this simple "does the app at least run cleanly".

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: Or maybe then the `AppHostRunsCleanly*` test should get an explicit list, which would be `[]` right now. And if we add more apps which we want to at least build-run check, then we can add those explicitly.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: Actually, yeah, this shouldn't be invoke once-per-endpoint, it should be invoked just once.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: I'm trying to avoid missing this message.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: I dropped the test entirely as it has no real purpose now. Any apps that we are testing are either tested with `TestEndpointsReturnOk`, or with more specific tests.
The only use left for the `AppHostRunsCleanly` was for apps that we just want to build+run, but even there we don't really check for errors, so it isn't useful.

tests/Aspire.Playground.Tests/AppHostTests.cs
eerhardt: Doesn't it read through the whole log from the start? If it already happened, I think it would still get seen.

eerhardt (approved):
My last comment is pretty minor. Mostly a question for my understanding.

This change LGTM.

tests/Aspire.Playground.Tests/AppHostTests.cs
radical: You are correct, it does play the whole log from the start! Maybe it should be renamed to like `FindOrWaitForTextAsync`.

davidfowl:
@radical I think the test is passing but there are some exceptions in the logs that look by design (retries?)

PR #5322, opened 2024-08-16
DRAFT: WaitFor with healthcheck awareness

mitchdenny:
## Description

This PR introduces a new `WaitFor` extension method that waits for a resource to start before allowing another resource to start. I'll attempt to summarize the approach here.

1. We introduce a new per-resource event that uses the eventing system called `ResourceCreatedEvent`: https://github.com/dotnet/aspire/pull/5322/files#diff-16f56cefe5f85117e6b36030d12f71ddfc60230018b25de447f352e3d3031185R18
2. We introduce a `WaitAnnotation` that is used by the various orchestrators to wait for a resource to be properly initialized. I've only tested this out with local containers so far but I believe I can adapt this for the AzureProvisioner as well. The annotation just represents a callback which we await ... the implementation will often just be a `TaskCompletionSource.Task` that we await.
3. Where a resource has health checks registered, we wait for those healthchecks to become healthy. We introduce a `HealthCheckAnnotation` which captures the name of the health check we should wait for, and inside the resource created event handler that we register we use the health check service to poll this health check. This decouples the waitfor implementation from the underlying health check logic.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: TODO
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5322)

### DIFF
--- a/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
@@ -6,6 +6,8 @@ using System.Text;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Redis;
 using Aspire.Hosting.Utils;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
 
 namespace Aspire.Hosting;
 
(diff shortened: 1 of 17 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1161,8 +1161,19 @@ await notificationService.PublishUpdateAsync(cr.ModelResource, s => s with
         }
     }
 
+    private async Task WaitForDependencyResourcesAsync(IResource resource, CancellationToken cancellationToken)
+    {
+        var resourceLogger = loggerService.GetLogger(resource);
+        var waitContext = new WaitContext(resourceLogger, serviceProvider);
+        var waitAnnotations = resource.Annotations.OfType<WaitAnnotation>().Select(w => w.WaitAsync(waitContext, cancellationToken)).ToArray();
+        Task.WaitAll(waitAnnotations, cancellationToken);
```
davidfowl, line 1169: Is this a mistake? This is blocking

mitchdenny: Not a mistake (at this point). This is the spot which enforces the wait to ensure other resources are healthy - but it highlights a problem with `ApplicationExecutor` that we have today which is that all the resources are started sequentially.

I think it would be good to discuss strategy here.

davidfowl: Resources are not started sequentially.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -28,11 +30,26 @@ public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicat
...
+            .WithImageRegistry(RedisContainerImageTags.Registry)
+            .WithAnnotation(new HealthCheckAnnotation($"StackExchange.Redis_{name}"));
+
+        builder.Eventing.Subscribe<ResourceCreatedEvent>(resource, async (e, ct) =>
+        {
+            var connectionString = await resource.GetConnectionStringAsync(ct).ConfigureAwait(false);
+            builder.Configuration[$"ConnectionStrings:{name}"] = connectionString;
+        });
+
+        builder.AddKeyedRedisClient(name, configureOptions: o =>
```
eerhardt, line 46: I think this is kind of heavy-weight, isn't it? We just want a HealthCheck registered, not the full client + telemetry, etc.

Why not just add a HealthCheck?

Also, should this be "opt-in" via a new API? So we aren't always doing this.

mitchdenny: I think we could try to make it lighter but we'd probably end up with pretty similar code to what we have in the Aspire integration packages - so we may as well just reuse them.

The main wrinkle is that we generally assume that the connection string is available at the point that we register the dependencies with DI - an assumption that does not hold in this scenario because we don't learn the connection string until after the container is built and endpoints are allocated.

So far I've managed to work around it but I'm not sure how long my luck will hold :)

mitchdenny:
We've got everything from this in main now.

PR #5350, opened 2024-08-20
Aspire.Hosting.Testing: Fix loading apphost assembly from a directory outside the entry assembly's location

radical:
Loading an AppHost assembly with `DistributedApplicationTestingBuilder.CreateAsync` from a directory other than the test assembly's location fails with:

```
  Error Message:
   System.InvalidOperationException : Missing deps file 'TestProject.AppHost.deps.json'. Make sure the project has been built.
  Stack Trace:
     at Aspire.Hosting.Testing.DistributedApplicationFactory.EnsureDepsFile(Type entryPoint) in /Users/ankj/dev/a3/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs:line 342
   at Aspire.Hosting.Testing.DistributedApplicationFactory.EnsureEntryPointStarted() in /Users/ankj/dev/a3/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs:line 252
   at Aspire.Hosting.Testing.DistributedApplicationFactory.ResolveBuilderAsync(CancellationToken cancellationToken) in /Users/ankj/dev/a3/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs:line 44
   at Aspire.Hosting.Testing.DistributedApplicationTestingBuilder.SuspendingDistributedApplicationFactory.CreateBuilderAsync(CancellationToken cancellationToken) in /Users/ankj/dev/a3/src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs:line 96
   at Aspire.Hosting.Testing.DistributedApplicationTestingBuilder.CreateAsync(Type entryPoint, String[] args, Action`2 configureBuilder, CancellationToken cancellationToken) in /Users/ankj/dev/a3/src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs:line 86
   at Aspire.Hosting.Testing.Tests.TestingBuilderTests.CanLoadFromDirectoryOutsideOfAppContextBaseDirectory() in /Users/ankj/dev/a3/tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs:line 47
```

.. as it tries to look for the `deps.json` file in `AppContext.BaseDirectory` instead of next to the AppHost assembly.

Feedback from @ eerhardt: This code path is checking for a `.deps.json` file, which is not useful for Aspire. AFAICT, this was necessary way back in the original MVC days when we would explicitly read the .deps.json file. I've traced the code all the way back to [aspnet/Mvc#6708 (files)](https://github.com/aspnet/Mvc/pull/6708/files#diff-dfb443073875510f427cf8ae0eaccf7dac74bda2a3454892fbe0b3ad76e67b04L56), which DefaultAssemblyPartDiscoveryProvider would read the .deps.json file.

Based on that we drop the `EnsureDepsFile` method avoiding the problem completely.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5350)


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -336,7 +336,8 @@ private static void EnsureDepsFile(Type entryPoint)
         }
 
         var depsFileName = $"{entryPoint.Assembly.GetName().Name}.deps.json";
-        var depsFile = new FileInfo(Path.Combine(AppContext.BaseDirectory, depsFileName));
+        // entryPoint.Assembly.Location is always an absolute path
+        var depsFile = new FileInfo(Path.Combine(Path.GetDirectoryName(entryPoint.Assembly.Location)!, depsFileName));
         if (!depsFile.Exists)
         {
             throw new InvalidOperationException($"Missing deps file '{Path.GetFileName(depsFile.FullName)}'. Make sure the project has been built.");
```
eerhardt, line 343:
```suggestion
```

I don't think we need this code that checks if there is a deps.json file and we should delete it.

AFAICT, this was necessary way back in the original MVC days when we would explicitly read the .deps.json file. I've traced the code all the way back to https://github.com/aspnet/Mvc/pull/6708/files#diff-dfb443073875510f427cf8ae0eaccf7dac74bda2a3454892fbe0b3ad76e67b04L56, which DefaultAssemblyPartDiscoveryProvider would read the .deps.json file.

For .NET Aspire, I don't think this check adds value. And as can be seen in this PR, causes more issues.

radical: Dropped it.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -328,21 +326,6 @@ private void ThrowIfNotInitialized()
         }
     }
 
-    private static void EnsureDepsFile(Type entryPoint)
-    {
-        if (entryPoint.Assembly.EntryPoint == null)
-        {
-            throw new InvalidOperationException($"Assembly of specified type {entryPoint.Name} does not have an entry point.");
-        }
```
eerhardt, line 336: This check may still be useful though...

tests/Aspire.Hosting.Testing.Tests/Aspire.Hosting.Testing.Tests.csproj
```diff
@@ -15,13 +15,22 @@
...
   <ItemGroup>
+    <ProjectReference Include="..\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.AppHost\Aspire.Hosting.AppHost.csproj" IsAspireProjectResource="false" />
     <ProjectReference Include="..\..\src\Aspire.Hosting.Testing\Aspire.Hosting.Testing.csproj" IsAspireProjectResource="false" />
-    <ProjectReference Include="..\Aspire.Hosting.Tests\Aspire.Hosting.Tests.csproj" />
     <ProjectReference Include="..\TestingAppHost1\TestingAppHost1.AppHost\TestingAppHost1.AppHost.csproj" />
+
+    <!-- This ProjectReference is *only* to get this project built, but not referenced directly.
+         It will be loaded dynamically by individual tests. -->
+    <ProjectReference Include="..\testproject\TestProject.AppHost\TestProject.AppHost.csproj" ReferenceOutputAssembly="false" IsAspireProjectResource="false" Private="false" />
```
eerhardt, line 33: I don't get why this is needed. How does it allow the project to build?

radical: This adds a project dependency, so msbuild will ensure that the project is built with the test project.
- The `ReferenceOutputAssembly=false` causes the output assembly from `TestProject` to not participate in the `Compile` for `Aspire.Hosting.Testing.Tests`.
- `IsAspireProjectResource=false` - we don't want to have code generated for `TestProject`, so that the test can load the assembly from disk instead.
- `Private=false` - ensures that none of the files (the `.deps.json` and some other files) are copied to `Aspire.Hosting.Testing.Tests` `bindir`.

radical: continued: And this allows `Aspire.Hosting.Testing.Tests` assembly to have no "ties" to `TestProject`, and the test can cleanly load the assembly via reflection.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
radical: Added that back, and a corresponding test.

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -13,6 +16,58 @@ namespace Aspire.Hosting.Testing.Tests;
...
+    [Fact]
+    public async Task ThrowsForAssemblyWithoutAnEntrypoint()
+    {
+        var classLibraryAssemblyPath = Directory.GetFiles(
+            AppContext.BaseDirectory,
+            "ClassLibrary.dll",
+            SearchOption.AllDirectories).Single();
+
+        var appHostAssembly = Assembly.LoadFrom(Path.Combine(AppContext.BaseDirectory, classLibraryAssemblyPath));
+        var appHostType = appHostAssembly.GetTypes().Single(t => t.FullName == "ClassLibrary.Class1");
```
eerhardt, line 65: Instead of a whole new assembly, could we just use an existing assembly? Like `Aspire.Components.Common.Tests`?

eerhardt (approved):
Just one small suggestion to simplify. Other than that, LGTM.

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
radical: `Aspire.Components.Common.Tests` gets an auto-generated `.entrypoint`:
```
.custom instance void [Microsoft.VisualStudio.TestPlatform.ObjectModel]Microsoft.VisualStudio.TestPlatform.TestSDKAutoGeneratedCode::.ctor()
```
I randomly used `Microsoft.Extensions.Logging.ConsoleLoggerExtensions` instead.

PR #5354, opened 2024-08-20
Persistent container support

danegsta:
## Description

Adds a new annotation and API to support marking container resources with a persistent lifetime. This will allow Aspire to use container resources with a longer lifetime than the AppHost (in addition to the current behavior that creates and tears down all container resources with the AppHost). Opening this as a draft PR as there is additional work to finalize the API and add tests required. 

This PR adds a new `ContainerLifetimeAnnotation` annotation and `WithContainerLifetime` resource builder API to allow configuring the expected lifetime of container resources. With this initial implementation, a container would be able to have one of two lifetime settings, `Default` or `Persistent`.

#### Default Lifetime

This lifetime mode is the same as the existing behavior for container resources. Default lifetime resources will be created when the AppHost starts based on the resource builder configuration and shut down when the AppHost terminates.

#### Persistent Lifetime

This is a new optional lifetime mode for container resources. A persistent container resource is identified by the assigned container/resource name. When the AppHost is started, if no running containers are found that match the specified service name, a new container resource will be created, just like for AppHost lifetime resources. However, if a container is found with the specified service name, Aspire will use that resource instead of creating a new one. When the AppHost shuts down, this container resource will not be terminated. This will allow a container to be re-used across multiple runs of the AppHost.

### Example

```C#
var catalogDb = builder.AddPostgres("postgres")
                       .WithContainerLifetime(ContainerLifetimeType.Persistent)
                       .WithDataVolume()
                       .WithPgAdmin()
                       .AddDatabase("catalogdb");
```

### Caveats

With this initial design, a persistent container resource will never be recreated if a matching container is found, even if the resource builder definition has been modified in such a way that the existing container is no longer compatible with the new configuration. Users will need to manually terminate their existing container via the Docker/Podman CLI or management GUI to trigger a recreation of the resource matching the new configuration. 

Granularity of container lifetime is limited to matching that of the AppHost or running as a globally persistent resource; we don't yet have a mechanism to spin up or tear down longer running sets of container services as part of a development session, this means that users that want longer running resources, but not constantly running resources will have to implement their own scripts or manually cleanup individual resources when they're not longer needed.

Both of these scenarios are cases that we'd like to improve on in future PRs, but we still feel this initial design should help reduce pain points related to container startup that we've heard from users of Aspire.

Finally, this initial API is marked as `Experimental` as we may need to revisit the design as we look to improve some of the identified limitations of the experience in future PRs.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5354)


src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -233,4 +233,20 @@ public static int GetReplicaCount(this IResource resource)
...
         }
     }
+
+    /// <summary>
+    /// Gets the lifetime type of the container for the specified resoruce. Defaults to <see cref="ContainerLifetimeType.AppHost"/> if
+    /// no <see cref="ContainerLifetimeAnnotation"/> is found.
+    /// </summary>
+    /// <param name="resource">The resource to the get the ContainerLifetimeType for.</param>
+    /// <returns></returns>
+    public static ContainerLifetimeType GetContainerLifetimeType(this IResource resource)
```
mitchdenny, line 243: This is the only thing that gives me pause right now. I'm wondering whether it might be best to make this internal for now. We probably want to mark the WithContainerLifetime API as experimental.

danegsta: Yeah, I felt a bit odd about the method since it's only really relevant to a resource that also has a container image annotation, but it also simplifies the ApplicationExecutor logic quite a bit.

src/Aspire.Hosting/ApplicationModel/ContainerLifetimeAnnotation.cs
```diff
@@ -0,0 +1,33 @@
...
+
+/// <summary>
+/// Lifetime modes for container resources
+/// </summary>
+public enum ContainerLifetimeType
+{
+    /// <summary>
+    /// The resource is tied to the lifetime of the AppHost.
+    /// </summary>
+    AppHost,
```
davidfowl, line 16:
```suggestion
    Default
```

danegsta: Does it make sense to have both `Default` and `AppHost`? Obviously `Default` would currently be the same as `AppHost`, but if we foresee a potential future where the meaning of `Default` could change, users may want to explicitly opt-out by keeping a resource in `AppHost` lifetime mode.

davidfowl: I think default without apphost

mitchdenny: Should this be a [Flags]?

danegsta: Seems like the values would be mutually exclusive; is there any particular benefit to adding `[Flags]` on an Enum with exclusive values?

davidfowl: I don't think so.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1327,11 +1327,23 @@ private void PrepareContainers()
                 throw new InvalidOperationException();
             }
 
-            var nameSuffix = GetRandomNameSuffix();
+            var nameSuffix = string.Empty;
+
+            if (container.GetContainerLifetimeType() == ContainerLifetimeType.Default)
```
davidfowl, line 1332: @ReubenBond we need to make sure this doesn't regress testing scenarios right?

danegsta: I've been hesitant to add tests to this PR for persistent resources as I don't want to introduce any test instability by leaving resources behind after a test run.

davidfowl:
@mitchdenny how do  you feel about merging this so we can start kicking the tires?

mitchdenny (approved):

mitchdenny:
Let's get it in. Would be good to socialise this with some of the other changes for 9.0

PR #5357, opened 2024-08-20
Fix console logs content changing order and the line number increasing

### LINKED ISSUE #4893: [AzureTools][Aspire] The sequence number of the log changes when returning "Resource" and then clicking "View" of the project or container under the "Logs" in the Dashboard again

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/4893

This PR fixes log order changing and line numbers increasing. Note that it doesn't address anything related to logs from `ILogger` inside the app host. That can be split out into its own issue and fixed later.

Changes:

* Turns on timestamp in executable logs
* Host parses date from incoming log lines
* Inserts log lines in backlog in date sorted order
* Adds some logging unit tests

console-logs-line-number-fix

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5357)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Threading.Channels;
 using Aspire.Dashboard.Otlp.Storage;
@@ -56,6 +58,17 @@ public class ResourceLoggerService
         ArgumentNullException.ThrowIfNull(resourceName);
 
         return GetResourceLoggerState(resourceName).Logger;
+    }
+
+    /// <summary>
+    /// The internal logger is used when adding logs from resource's stream logs.
+    /// It allows the parsed date from text to be used as the log line date.
+    /// </summary>
+    internal Action<DateTime, string, bool> GetInternalLogger(string resourceName)
+    {
+        ArgumentNullException.ThrowIfNull(resourceName);
+
+        return GetResourceLoggerState(resourceName).AddLog;
     }
 
     /// <summary>
@@ -159,7 +172,8 @@ public class ResourceLoggerService
         }
     }
 
-    private ResourceLoggerState GetResourceLoggerState(string resourceName) =>
+    // Internal for testing.
+    internal ResourceLoggerState GetResourceLoggerState(string resourceName) =>
         _loggers.GetOrAdd(resourceName, (name, context) =>
         {
             var state = new ResourceLoggerState();
@@ -168,15 +182,19 @@ public class ResourceLoggerService
         },
         this);
 
+    internal sealed record InternalLogLine(DateTime DateTimeUtc, string Message, bool IsError);
+
     /// <summary>
     /// A logger for the resource to write to.
     /// </summary>
-    private sealed class ResourceLoggerState
+    internal sealed class ResourceLoggerState
     {
         private readonly ResourceLogger _logger;
         private readonly CancellationTokenSource _logStreamCts = new();
 
-        private readonly CircularBuffer<LogLine> _backlog = new(10000);
+        private Task? _backlogReplayCompleteTask;
+        private long _lastLogReceivedTimestamp;
+        private readonly CircularBuffer<InternalLogLine> _backlog = new(10000);
 
         /// <summary>
         /// Creates a new <see cref="ResourceLoggerState"/>.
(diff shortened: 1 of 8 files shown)


mitchdenny (approved):

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -220,45 +238,46 @@ public event Action<bool> OnSubscribersChanged
...
             OnNewLog += Log;
 
-            backlogSnapshot = GetBacklogSnapshot();
-            flushBacklogSync.Set();
+            // Add a small delay to ensure the backlog is replayed from DCP and ordered correctly.
+            await EnsureBacklogReplayAsync(cancellationToken).ConfigureAwait(false);
+
+            lock (_backlog)
+            {
+                backlogSnapshot = GetBacklogSnapshot();
```
eerhardt, line 268: `GetBacklogSnapshot()` already locks inside of it. There isn't a reason to lock here (in an async method) as well.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -267,11 +286,51 @@ void Log(LogLine log)
...
+
+            async Task StartBacklogReplayAsync(CancellationToken cancellationToken)
+            {
+                var delay = TimeSpan.FromMilliseconds(100);
+
+                // There could be an initial burst of logs as they're replayed. Give them the opportunity to be loaded
+                // into the backlog in the correct order and returned before streaming logs as they arrive.
+                for (var i = 0; i < 3; i++)
+                {
+                    await Task.Delay(delay, cancellationToken).ConfigureAwait(false);
```
eerhardt, line 319: This seems fragile. Is there a way to use concurrency controls (reset events, semaphores, etc) to ensure the ordering happens correctly instead of delaying based on time?

JamesNK: Yeah, I agree that's not the best.

The problem is that ordering values in the backlog doesn't mean much once you start to stream log items as they're received. Sending items as they're received is fine after they logs have been replayed from DCP.

I think what is needed from DCP is an indication of how many lines are being replayed. Then we can tell the logger service that the backlog is full (and all the log items in it are sorted correctly), and that it can send the backlog and send new log lines in real time.

eerhardt: I don't understand why the existing `flushBacklogSync` doesn't work. We delay writing to the channel until we've grabbed the backlog snapshot.

JamesNK: That’s not the problem this change is fixing

see https://github.com/dotnet/aspire/issues/4893#issuecomment-2274700970

eerhardt: Why do we need to "Add a small delay to ensure the backlog is replayed from DCP and ordered correctly."? Why isn't the order of operations:

1. Grab the backlog snapshot
2. The following can happen concurrently
a. After the snapshot has been made, as new logs get written, write them to the channel
b. Write the snapshot to the listener
3. After the snapshot has been written, write the messages from the channel

davidfowl:
I don't fully understand this change tbh. Why is there a random delay? Does that happen for custom resources or is this only for DCP based logs? Why isn't the date parsing in that logic and why is it in the core of the resource logger?

src/Aspire.Hosting/Dcp/ResourceLogSource.cs
```diff
@@ -34,12 +34,10 @@ public async IAsyncEnumerator<LogEntryList> GetAsyncEnumerator(CancellationToken
 
         var streamTasks = new List<Task>();
 
-        var timestamps = resource is Container; // Timestamps are available only for Containers as of Aspire P5.
-
         if (resource is Container && dcpVersion?.CompareTo(DcpVersion.MinimumVersionAspire_8_1) >= 0)
         {
-            var startupStderrStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdErr, follow: true, timestamps: timestamps, cancellationToken).ConfigureAwait(false);
-            var startupStdoutStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdOut, follow: true, timestamps: timestamps, cancellationToken).ConfigureAwait(false);
+            var startupStderrStream = await kubernetesService.GetLogStreamAsync(resource, Logs.StreamTypeStartupStdErr, follow: true, timestamps: true, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
karolz-ms: I'd prefer to keep the `timestamps` parameter on the `KubernetesService` class in case we find a use case for logs without timestamps in future.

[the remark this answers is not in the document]
eerhardt: We can always add it back, once we find a need. It is internal.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
JamesNK: > Grab the backlog snapshot

This doesn't work because it's empty. When you view the console logs page, and the request goes to the host to display logs, the snapshot has nothing in it. We can't immediately start streaming all the initial logs that DCP returns to us because we receive them out of order.

The delay is to give the opportunity for DCP to send us the initial logs, them to be put in order in the backlog, then the backlog returned to the UI, then new logs are streamed as they arrive.

PR #5359, opened 2024-08-20
Introduce ResourceCreatingEvent

mitchdenny:
## Description

This PR introduces a new eventing event `ResourceCreatingEvent. This work is related to enabling `WaitFor` support.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No 
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1478
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5359)


src/Aspire.Hosting/ApplicationModel/BeforeResourceStartedEvent.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+/// This event is raised by orchestrators before they have started a new resource.
+/// </summary>
+/// <param name="resource">The resource that is being created.</param>
+/// <param name="services">The <see cref="IServiceProvider"/> for the app host.</param>
+/// <remarks>
+/// Resources that are created by orchestrators may not yet be ready to handle requests.
+/// </remarks>
+public class BeforeResourceStartedEvent(IResource resource, IServiceProvider services) : IDistributedApplicationResourceEvent
```
eerhardt, line 16: We typically use the words "-ing" and "-ed" for before and after events. Examples:

https://github.com/dotnet/runtime/blob/3622bfa4329faa73551b2508aade7777a04a3ca8/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/IHostApplicationLifetime.cs#L22-L28

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events

✔️ DO give events names with a concept of before and after, using the present and past tenses.

For example, a close event that is raised before a window is closed would be called Closing, and one that is raised after the window is closed would be called Closed.

❌ DO NOT use "Before" or "After" prefixes or postfixes to indicate pre- and post-events. Use present and past tenses as just described.

tests/Aspire.Hosting.Tests/Eventing/DistributedApplicationBuilderEventingTests.cs
```diff
@@ -19,6 +19,79 @@ public void CanResolveIDistributedApplicationEventingFromDI()
         Assert.Equal(builder.Eventing, eventing);
     }
 
+    [Fact]
```
radical, line 22:
```suggestion
    [Fact]
    [RequiresDocker]
```

tests/Aspire.Hosting.Tests/Eventing/DistributedApplicationBuilderEventingTests.cs
```diff
@@ -19,6 +19,79 @@ public void CanResolveIDistributedApplicationEventingFromDI()
...
+        var allFired = ManualResetEvent.WaitAll(
+            [beforeResourceStartedEvent.WaitHandle, afterResourceCreatedEvent.WaitHandle],
+            TimeSpan.FromSeconds(10)
+            );
+
+        Assert.True(allFired);
+        await app.StopAsync();
+    }
+
+    [Fact]
```
radical, line 59:
```suggestion
    [Fact]
    [RequiresDocker]
```

src/Aspire.Hosting/ApplicationModel/BeforeResourceStartedEvent.cs
mitchdenny: @davidfowl should I just revert my commit :)

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -272,6 +278,9 @@ await provisioner.GetOrCreateResourceAsync(
                     cancellationToken).ConfigureAwait(false);
 
                 resource.ProvisioningTaskCompletionSource?.TrySetResult();
+
+                var afterResourceCreatedEvent = new AfterResourceStartedEvent(resource, serviceProvider);
+                await eventing.PublishAsync(afterResourceCreatedEvent, cancellationToken).ConfigureAwait(false);
```
davidfowl, line 283: This makes me wonder if we need another event that fires on failure since these only fire on success.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -217,6 +220,9 @@ private async Task ProvisionAzureResources(
 
     private async Task ProcessResourceAsync(IConfiguration configuration, Lazy<Task<ProvisioningContext>> provisioningContextLazy, IAzureResource resource, CancellationToken cancellationToken)
     {
+        var beforeResourceStartedEvent = new BeforeResourceStartedEvent(resource, serviceProvider);
+        await eventing.PublishAsync(beforeResourceStartedEvent, cancellationToken).ConfigureAwait(false);
```
davidfowl, line 224: Is this too early to do anything useful with the resource? Do we need to fire this later? I wonder if we're better off firing these events in the BicepProvisioner?

mitchdenny: I think if there are things that happen in the Bicep lifecycle that are interesting then we should define events for those separately and keep these events consistent across the various things doing orchestrations.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
mitchdenny: I think that would be a Bicep specific event.

src/Aspire.Hosting/ApplicationModel/BeforeResourceStartedEvent.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <summary>
+/// This event is raised by orchestrators before they have started a new resource.
+/// </summary>
+/// <param name="resource">The resource that is being created.</param>
+/// <param name="services">The <see cref="IServiceProvider"/> for the app host.</param>
+/// <remarks>
+/// Resources that are created by orchestrators may not yet be ready to handle requests.
+/// </remarks>
+[Experimental("ASPIREEVENTING001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
+public class BeforeResourceStartedEvent(IResource resource, IServiceProvider services) : IDistributedApplicationResourceEvent
```
eerhardt, line 18: I think this name still violates the naming guidelines. 

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events

✔️ DO give events names with a concept of before and after, using the present and past tenses.

For example, a close event that is raised before a window is closed would be called Closing, and one that is raised after the window is closed would be called Closed.

❌ DO NOT use "Before" or "After" prefixes or postfixes to indicate pre- and post-events. Use present and past tenses as just described.

```suggestion
public class ResourceStartingEvent(IResource resource, IServiceProvider services) : IDistributedApplicationResourceEvent
```

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
eerhardt, line 224: Are there tests that can be written for the Azure provisioning event?

davidfowl: That component isn't currently testable

eerhardt: 😢

davidfowl (approved):
@eerhardt we're going to rename in the follow up

src/Aspire.Hosting/ApplicationModel/BeforeResourceStartedEvent.cs
mitchdenny: Agree these names aren't 100% right. But we will be doing a follow up.

PR #5366, opened 2024-08-20
ElasticSearchInstrumentationEndToEnd: Increase timeout for RemoteExecutor

radical:
.. from 60s to 120s. This has been failing frequently on CI with:

```
Aspire.Elastic.Clients.Elasticsearch.Tests.AspireElasticClientExtensionsTest.ElasticsearchInstrumentationEndToEnd [FAIL]
  Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Half-way through waiting for remote process.
  Timed out at 8/20/2024 6:22:40 PM after 60000ms waiting for remote process.
  	Process ID: 29680
  	Handle: 11984
  	Name: dotnet
  	MainModule: /datadisks/disk1/work/B7250A1D/p/dotnet-cli/dotnet
  	StartTime: 8/20/2024 6:21:39 PM
  	TotalProcessorTime: 00:00:00.7300000

  Stack Trace:
    /_/src/Microsoft.DotNet.RemoteExecutor/src/RemoteInvokeHandle.cs(225,0): at Microsoft.DotNet.RemoteExecutor.RemoteInvokeHandle.Dispose(Boolean disposing)
    /_/src/Microsoft.DotNet.RemoteExecutor/src/RemoteInvokeHandle.cs(55,0): at Microsoft.DotNet.RemoteExecutor.RemoteInvokeHandle.Dispose()
    /_/tests/Aspire.Elastic.Clients.Elasticsearch.Tests/AspireElasticClientExtensionsTest.cs(145,0): at Aspire.Elastic.Clients.Elasticsearch.Tests.AspireElasticClientExtensionsTest.ElasticsearchInstrumentationEndToEnd()
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
       at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```
- **ElasticSearchInstrumentationEndToEnd: Increase timeout for RemoteExectuor**
- **Add some helix retries for Aspire.Elastic.Clients.Elasticsearch.Tests**
- Also, drop old helix retry conditions which were for https://github.com/dotnet/aspire/issues/2850, but the issue has since been fixed.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5366)

### DIFF
--- a/eng/test-configuration.json
+++ b/eng/test-configuration.json
@@ -5,6 +5,8 @@
   "retryOnRules": [
     { "testAssembly": { "regex": "Aspire.EndToEnd.*" }, "failureMessage": { "regex": "App run failed" } },
     { "testAssembly": { "regex": "Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*" } },
-    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } }
+    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } },
+    { "testName": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Tests.ConformanceTests.HealthCheckReportsExpectedStatus" } },
+    { "testAssembly": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Test" }, "failureMessage": { "contains": "RemoteExecutionException : Half-way through waiting for remote process." } }
   ]
 }
--- a/eng/test-configuration.json
+++ b/eng/test-configuration.json
@@ -3,10 +3,7 @@
   "defaultOnFailure": "fail",
   "localRerunCount": 1,
   "retryOnRules": [
-    { "testAssembly": { "regex": "Aspire.EndToEnd.*" }, "failureMessage": { "regex": "App run failed" } },
-    { "testAssembly": { "regex": "Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*" } },
-    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } },
     { "testName": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Tests.ConformanceTests.HealthCheckReportsExpectedStatus" } },
-    { "testAssembly": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Test" }, "failureMessage": { "contains": "RemoteExecutionException : Half-way through waiting for remote process." } }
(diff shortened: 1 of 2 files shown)


eng/test-configuration.json
```diff
@@ -5,6 +5,8 @@
   "retryOnRules": [
     { "testAssembly": { "regex": "Aspire.EndToEnd.*" }, "failureMessage": { "regex": "App run failed" } },
     { "testAssembly": { "regex": "Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*" } },
-    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } }
+    { "testAssembly": { "regex": "Aspire.Microsoft.EntityFrameworkCore.SqlServer.*" } },
```
sebastienros, line 8: Are we sure this assembly needs retry? Isn't it unit tests only?

eng/test-configuration.json
```diff
@@ -5,6 +5,8 @@
...
+    { "testName": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Tests.ConformanceTests.HealthCheckReportsExpectedStatus" } },
```
sebastienros, line 9: What I see is that the "testName/contains" rule might be more appropriate than 
"testAssembly/regex" in the preceeding cases, since the regex is not really a regex, and `contains` could be used with the assembly name as the prefix instead.

eng/test-configuration.json
```diff
@@ -5,6 +5,8 @@
...
+    { "testAssembly": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Test" }, "failureMessage": { "contains": "RemoteExecutionException : Half-way through waiting for remote process." } }
```
sebastienros, line 10: Funny, after my two previous comments, I see a mix, "contains" on "testAssembly". Not wrong though.

sebastienros (approved):
Just made comments unrelated to the goal of this PR.

eng/test-configuration.json
radical: These two can be dropped. We had added retries for old issues which have since been fixed.
https://github.com/dotnet/aspire/commit/4976ee1366bb8119703be6d0a15f997975ea732b

sebastienros: My point is more about:
- When to use `testAssembly` vs `testName`
- That `"Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.*"` is not a regex, or at least it's doing what we assume it tries to do by chance (the dot can be anything and the `*` appears right after a dot, double points for luck here).

PR #5374, opened 2024-08-21
Bump default AzureOpenAIDeployment sku

### LINKED ISSUE #4970: Reconsider default Azure OpenAI rate limits
Today, when using Azure OpenAI with .NET Aspire, you get the following defaults on your deployment:

* SkuName: "Standard"
* SkuCapacity: 1

https://github.com/dotnet/aspire/blob/604f62f5c917f9a855b5188b1be78deac6234ad9/src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIDeployment.cs#L14

`capacity | integer | This represents the amount of  quota you are assigning to this deployment. A value of 1 equals 1,000 Tokens per Minute (TPM). A value of 10 equals 10k Tokens per Minute (TPM).`

(One problem is we are using default parameter values here, which hard-codes these values into the caller. Basically meaning we can't update these values in the future without people recompiling their code).

We should re-consider the default value of 1,000 tokens per minute. With this limit not even the simple playground app in the repo works out of the box:

https://github.com/dotnet/aspire/blob/604f62f5c917f9a855b5188b1be78deac6234ad9/playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/Home.razor#L31

Just running this prompt twice in a row (because of how Blazor Server works with prerendering) causes `429` rate limit errors.

One reason to use such a low limit is because deployment will succeed even if using a Free Trial subscripton: https://learn.microsoft.com/azure/ai-services/openai/quotas-limits#other-offer-types. However, suffering through rate limit errors for everyone may be worse than erroring for the low end subscriptions (free, student).

When I try a value too big, I get a pretty decent error:

```
Azure.RequestFailedException: The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '476657d0-1ec0-4746-9125-6b0c12cddabc'. See inner errors for details.
Status: 400 (Bad Request)
ErrorCode: InvalidTemplateDeployment

Content:
{"error":{"code":"InvalidTemplateDeployment","message":"The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '476657d0-1ec0-4746-9125-6b0c12cddabc'. See inner errors for details.","details":[{"code":"InsufficientQuota","message":"This operation require 10000 new capacity in quota Tokens Per Minute (thousands) - gpt-4o, which is bigger than the current available capacity 150. The current quota usage is 0 and the quota limit is 150 for quota Tokens Per Minute (thousands) - gpt-4o."}]}}
```

Giving a higher limit will make more scenarios work out of the box for more customers, while it may mean some customers experience an error during deployment - with a good error they can search for how to fix.

sebastienros:
Fixes #4970

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5374)

### DIFF
--- a/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
+++ b/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
@@ -5,8 +5,6 @@ var builder = DistributedApplication.CreateBuilder(args);
 
 var deploymentAndModelName = "gpt-4o";
 var openai = builder.AddAzureOpenAI("openai").AddDeployment(
-    // the default SKU capacity of 1,000 TPM is not high enough for the OpenAIEndToEnd_WebStory project
-    // see https://github.com/dotnet/aspire/issues/4970
     new(deploymentAndModelName, deploymentAndModelName, "2024-05-13")
     );
 
--- a/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
+++ b/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
@@ -7,7 +7,7 @@ var deploymentAndModelName = "gpt-4o";
 var openai = builder.AddAzureOpenAI("openai").AddDeployment(
     // the default SKU capacity of 1,000 TPM is not high enough for the OpenAIEndToEnd_WebStory project
     // see https://github.com/dotnet/aspire/issues/4970
-    new(deploymentAndModelName, deploymentAndModelName, "2024-05-13", skuCapacity: 8)
+    new(deploymentAndModelName, deploymentAndModelName, "2024-05-13")
     );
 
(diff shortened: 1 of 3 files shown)


playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/Program.cs
```diff
@@ -7,7 +7,7 @@
 var openai = builder.AddAzureOpenAI("openai").AddDeployment(
     // the default SKU capacity of 1,000 TPM is not high enough for the OpenAIEndToEnd_WebStory project
     // see https://github.com/dotnet/aspire/issues/4970
```
eerhardt, line 9:
```suggestion
```

eerhardt (approved):
LGTM. Thanks!

Issue #4974, opened 2024-07-18
Show useful connection information in details pane of the resources


JerryNixon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

![image](https://github.com/user-attachments/assets/eca781a1-0274-4c96-9546-594d30e696e5)

When a SQL Server reports something along the lines of `tcp://localhost:1234` it is accurately indicating information about the sever, but it is not indicating valuable information to the developer because connection to the server requires the IP address and a comma, like `127.0.0.1,1234` which they would use in SSMS, or ADS, or VSCode, or Visual Studio SQL Server Object Explorer. Should they instead use the provided string, to novice developers this will appear correct, but will fail on them every single time. 

When this is a web site or web API, the text becomes a link which can be clicked. How amazing would it be if this became a link here too - but for SQL server launching any of the tools or a simple web interface to interrogate the schema. Or even set this value programmatically when I create a SqlServer in AppHost. Sorry. I realize I just leaned into a second or even third issue. Please disregard this paragraph.

## Solution

1. Show the IP address.
2. Replace the colon with a comma.

## Goal

Make it so developers have it easier.

samsp-msft:
AFAIK this is the data that we get from DCP for the container, and don't know whether its SQL or some other component. It will probably need coordination between:
- App model to know the specifics about how an endpoint should be presented to the user for the specific service
- DCP to supply that information to the dashboard via the gRPC interface
- Dashboard to display it.

@mitchdenny @karolz-ms @drewnoakes  for their respective parts of the problem.

drewnoakes:
The resource service exposes a set of `Url`s with the following properties:

https://github.com/dotnet/aspire/blob/df935f4ee2e76d088235ea0fe91a0b61405cb3fe/src/Aspire.Hosting/Dashboard/proto/resource_service.proto#L113-L121

The protocol, as written, assumes that the `full_url` value is a valid RFC 1738 URL.

We may be able to provide a "display value" for these URLs, where some component close to DCP or the app model produces that optional display value when a different, non-URL form is required.

karolz-ms:
DCP does not really "know" it is dealing with a SQL Server. It only understands fairly low-level objects: containers, executables, IP addresses, ports, and the like.

[...]

Issue #1644, opened 2024-01-11
Consider getting known properties for resource details panel from resource server


JamesNK:
A resource has a list of properties. The resource details panel has UI to display all properties for a resource. To improve the user experience, the details UI has the concept of known properties. These are properties that are commonly useful, and we want to show people, e.g. resource state. Other less useful properties are hidden by default, e.g. resource.uid. Known properties also provide friendly names (including translation support) and ordering.

There are hardcoded known properties for containers, projects, and executables. Properties for custom resource types are still available, but custom resources don't have the opportunity to customize how properties are displayed.

Consider adding support for defining known properties on a resource, the resource server having an API for the dashboard to get known properties, and the UI then using fetched known properties.

More detail: https://github.com/dotnet/aspire/pull/1542#discussion_r1446972725

cc @drewnoakes

maddymontaquila:
@JamesNK thoughts on this?

joperezr:
As part of triage cleanup ahead of the 13.3 release, this issue is being moved to the Backlog milestone. If you believe it should remain in 13.3, feel free to re-add to the milestone and leave a comment here why it should not be punted.

davidfowl:
This is addressed by #18132.

The dashboard no longer needs to hard-code every important property it should show in the resource details panel. Resource snapshots can now carry display metadata from the resource/server side:

- `ResourcePropertySnapshot.DisplayName` lets the producer provide a friendly label.
- `ResourcePropertySnapshot.IsHighlighted` lets the producer mark a property as important enough to show by default.
- The resource details panel now includes highlighted properties by default, even when they are not one of the dashboard's built-in known properties.

Azure provisioning now uses this for Azure context properties such as subscription, resource group, location, tenant ID, and tenant domain. That means these properties are selected by the resource/provisioning layer and surfaced by the dashboard without adding each one to the dashboard's known-property list.

PR #5380, opened 2024-08-22
Allow resource properties to be marked as secret

### LINKED ISSUE #5065: Allow resource properties to be marked as secret
Today, environment variables are shown in the UI as a series of dots (like a password mask) and the user must flip a toggle to show them. This prevents accidental disclosure to onlookers, such as when screen sharing.

We should add the same facility to resource _properties_, and allow each property to define whether it is a secret or not. Such secret properties would be displayed as masked in the UI by default.

### LINKED ISSUE #5377: Inconsistent sort behavior in Resource Details view
### Describe the bug

On the Resources page, click a resource. The view that appears shows properties, endpoints and environment variables.

image

The environment variable grid supports sorting by clicking on the name or value column headers.

The other two tables do not support sorting.

### Expected Behavior

The tables have uniform appearance and functionality.

### Steps To Reproduce

1. Open dashboard
2. Click on a resource

drewnoakes:
Fixes #5065
Fixes #5377

This change allows selected resource properties in the details view to support masking, and show as masked by default. This behaviour exists today for all environment variables, and has so-far not been available for resource properties.

The resource service specifies which properties may be "sensitive", and the UI shows them masked.

Currently, the app host hides container/executable command line arguments, but it can be amended to include other properties too. Other resource services (like ACA's) can customise this logic to their needs as well.

As part of this change, we create interface `IPropertyGridItem` that view models must implement to be bound to the `PropertyGrid` component. This allows the component to work directly with the items, rather than calling back to the component's user via parameters for callback logic that largely do the same thing.

We also avoid having the `ResourceDetails` code create copies of `ResourcePropertyViewModel` objects for binding the UI. This copying won't work with the new `IPropertyGridItem` as the `PropertyGrid` requires these to be the same mutable object over time. Removing this fixed the inability to unmask resource property values.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5380)


src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -117,111 +113,101 @@ private IEnumerable<DisplayedEndpoint> GetEndpoints()
...
-        // Return the resource properties, with an optional known property.
-        // Order properties by the known property order. Unmatched properties are last.
-        var values = Resource.Properties
-            .Where(p => !p.Value.HasNullValue && !(p.Value.KindCase == Value.KindOneofCase.ListValue && p.Value.ListValue.Values.Count == 0))
-            .GroupJoin(
-                resolvedKnownProperties,
-                p => p.Key,
-                k => k.Key,
-                (p, k) => new SummaryValue { Key = p.Key, Value = p.Value, KnownProperty = k.SingleOrDefault(), Tooltip = GetTooltip(p.Value) })
-            .OrderBy(v => v.KnownProperty != null ? resolvedKnownProperties.IndexOf(v.KnownProperty) : int.MaxValue);
```
drewnoakes, line 140: `ResourcePropertyViewModel` is now a stateful, mutable object. We cannot create copies here. Instead, we reuse the objects for each render, so the UI works correctly without reverting state.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -117,111 +113,101 @@ private IEnumerable<DisplayedEndpoint> GetEndpoints()
...
-        }
-        else
+        void PopulateMissingKnownProperties()
         {
-            // Dates are returned as ISO 8601 text.
-            // Use try parse to check if a value matches ISO 8601 format. If there is a match then convert to a friendly format.
-            if (DateTime.TryParseExact(value, "o", CultureInfo.InvariantCulture, DateTimeStyles.None, out var date))
+            List<KnownProperty>? knownProperties = null;
+
+            // Lazily populate some data on the view models, that wasn't available when they were constructed.
```
drewnoakes, line 131: Longer term, I hope we can avoid patching up these objects by setting these values at construction. For now, this most closely mirrors the code that existed before.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -117,111 +113,101 @@ private IEnumerable<DisplayedEndpoint> GetEndpoints()
...
+
+                    for (var i = 0; i < knownProperties.Count; i++)
+                    {
+                        var kp = knownProperties[i];
+
+                        if (kp.Key == key)
+                        {
+                            found = true;
+                            vm.KnownProperty = kp;
+                            vm.Priority = i;
```
drewnoakes, line 149: We store a priority directly on the object to avoid the _O(N)_ `IndexOf` within an _O(N log N)_ `OrderBy`, leading to _O(N^2 log N)_ which is quasi-quadratic rather than log-linear time complexity.

We also only do this lookup once, rather than on every render of the UI. So after this, it's constant time.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -97,6 +102,68 @@ public EnvironmentVariableViewModel(string name, string? value, bool fromSpec)
...
+
+        Name = name;
+        Value = value;
+        IsValueSensitive = isValueSensitive;
+        KnownProperty = knownProperty;
+        IsValueMasked = isValueSensitive;
+
+        _tooltip = new(() => value.HasStringValue ? value.StringValue : value.ToString());
+
+        _displayValue = new(() =>
```
drewnoakes, line 135: These values can be lengthy strings, so cache them once and reuse.

src/Aspire.Dashboard/ResourceService/Partials.cs
```diff
@@ -24,7 +24,15 @@ public ResourceViewModel ToViewModel()
...
                 CreationTimeStamp = ValidateNotNull(CreatedAt).ToDateTime(),
-                Properties = Properties.ToFrozenDictionary(property => ValidateNotNull(property.Name), property => ValidateNotNull(property.Value), StringComparers.ResourcePropertyName),
+                Properties = Properties.ToFrozenDictionary(
+                    comparer: StringComparers.ResourcePropertyName,
+                    keySelector: property => ValidateNotNull(property.Name),
+                    elementSelector: property => new ResourcePropertyViewModel(
+                        name: ValidateNotNull(property.Name),
+                        value: ValidateNotNull(property.Value),
+                        isValueSensitive: property.IsSensitive,
+                        knownProperty: null, // TODO can we do better here? yes, if the server sends known properties
```
drewnoakes, line 34: In future I hope to set this and the priority value directly here, simplifying the code a little by removing the need for `PopulateMissingKnownProperties`.

src/Aspire.Hosting/Dashboard/proto/Partials.cs
```diff
@@ -1,13 +1,24 @@
...
 partial class Resource
 {
+    /// <summary>
+    /// Names of resource properties known to contain potentially sensitive values that should be masked in the UI by default.
+    /// </summary>
+    private static readonly ImmutableHashSet<string> s_knownSensitiveResourcePropertyNames =
+    [
+        KnownProperties.Container.Args,
+        KnownProperties.Executable.Args,
+    ];
```
drewnoakes, line 20: These are the hard-coded properties that the app host considers to be sensitive for now. We can easily add others if needed.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -117,6 +117,8 @@ message ResourceProperty {
     optional string display_name = 2;
     // The data value. May be null, a number, a string, a boolean, a dictionary of values (Struct), or a list of values (ValueList).
     google.protobuf.Value value = 3;
+    // Whether the value is sensitive and should be masked in the UI by default. The user may explicitly unmask this value to view it.
+    optional bool is_sensitive = 4;
```
drewnoakes, line 121: This is the backwards-compatible proto change that allows a server to specify a property as sensitive.

src/Aspire.Dashboard/Components/Controls/SpanDetails.razor
```diff
@@ -84,25 +72,17 @@
...
                         @if (context.Attributes.Length > 0)
                         {
-                            // There's a weird bug where trying to render a nested FluentDataGrid here with a non-primitive
-                            // TItem leads to the click event not being raised *only in the value (not header) rows*. A
-                            // workaround is to use the attribute index as the item and query the attributes list for its
-                            // name and value in the name/value producers.
-                            <PropertyGrid TItem="int"
-                                          Items="@context.Attributes.Select((_, i) => i).AsQueryable()"
+                            <PropertyGrid TItem="OtlpSpanAttributeItem"
+                                          Items="@context.Attributes.Select(pair => new OtlpSpanAttributeItem(pair.Key, pair.Value)).AsQueryable()"
```
drewnoakes, line 84: @JamesNK can you help me find a way to validate that the bug described here isn't occurring? I stuck a breakpoint in the lambda here but couldn't get it to hit in TestShop. Is there a better playground to use to test this?

JamesNK:
Non-URL endpoints have disappeared.

Before:

image

After:

image

src/Aspire.Dashboard/Components/Controls/PropertyGrid.razor.cs
```diff
@@ -6,10 +6,62 @@
...
+/// </para>
+/// <para>
+/// The property grid has two columns, bound to display strings <see cref="Name"/> and <see cref="Value"/>.
+/// </para>
+/// <para>
+/// The <see cref="IsValueSensitive"/> and <see cref="IsValueMasked"/> properties control masking behavior,
+/// which prevents sensitive data from being displayed in the UI without user interaction.
+/// </para>
+/// </remarks>
+public interface IPropertyGridItem
```
JamesNK, line 24: nit: Call this IPropertyGridViewModel

The standard throughout the UI is to suffix view model to types that are designed to be databound.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -117,111 +113,101 @@ private IEnumerable<DisplayedEndpoint> GetEndpoints()
...
+                    KnownResourceTypes.Executable => _executableProperties,
+                    KnownResourceTypes.Container => _containerProperties,
+                    _ => _resourceProperties
+                };
             }
         }
     }
 
-    private void CheckAllMaskStates()
+    private static RenderFragment GetContentAfterValue(DisplayedEndpoint vm)
```
JamesNK, line 174: I'm guessing this method is where the bug with non-URL addresses is.

Creating render fragments in code behind is laborious and error prone. We should do this in the razor file where HTML can be written inline.

JamesNK: Example of this: https://github.com/dotnet/aspire/blob/16953c4e71f6af49a6af671899a4eef754510580/src/Aspire.Dashboard/Components/Controls/SpanDetails.razor#L161-L194

mitchdenny:
> Non-URL endpoints have disappeared.
> 
> Before:
> 
> image
> 
> After:
> 
> image

Should this have a reveal icon like we do for environment variables, or is this for "hard secrets".

drewnoakes:
> Should this have a reveal icon like we do for environment variables, or is this for "hard secrets".

These aren't masked, they're just straight-up missing. I will fix this.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -117,111 +113,101 @@ private IEnumerable<DisplayedEndpoint> GetEndpoints()
         return ResourceEndpointHelpers.GetEndpoints(Resource, includeInternalUrls: true);
     }
 
-    private IEnumerable<SummaryValue> GetResourceValues()
+    private IEnumerable<ResourcePropertyViewModel> GetResourceProperties(bool ordered)
```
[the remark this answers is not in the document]
drewnoakes: All the complexity here will go away as part of https://github.com/dotnet/aspire/pull/5380#discussion_r1726498633

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -41,86 +41,59 @@
...
+@code {
+    private static RenderFragment GetContentAfterValue(DisplayedEndpoint vm)
+    {
+        if (vm.Url is null)
+        {
+            return @<span>@vm.Text</span>;
+        }
+        else
+        {
+            return @<a href="@vm.Url">@vm.Text</a>;
```
JamesNK, line 96:
```suggestion
            return @<a href="@vm.Url" target="_blank">@vm.Text</a>;
```

JamesNK: Previously the text in the `<a>` tag was the Url property. Is Text always the same value as Url?

https://github.com/dotnet/aspire/blob/3439c433770255f06ef286c1663255f066687324/src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor#L94

src/Aspire.Dashboard/ResourceService/Partials.cs
drewnoakes: This was done in this PR.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
drewnoakes: This was done in this PR.

src/Aspire.Dashboard/Components/Controls/SpanDetails.razor
drewnoakes: @JamesNK do you recall how you determined that using a value type worked? Was there some error message you saw or analysis technique you employed?

davidfowl:
@drewnoakes This is close to being done right?

drewnoakes:
It's close, but blocked on the (minor) bug mentioned above. The cause seems to be a UI update loop that prevents the popup from opening. I haven't made time to investigate yet, given other priorities, but will come back to it soon.

src/Aspire.Dashboard/Components/Controls/SpanDetails.razor
drewnoakes: Pushed a fix for this.

drewnoakes:
> It's close, but blocked on the (minor) bug mentioned above.

This has been fixed.

src/Aspire.Dashboard/Components/Controls/PropertyGrid.razor.cs
drewnoakes: Thanks but I'll leave it as is. It's an interface which describes a role rather than a concrete type. `PropertyGrid` refers to its items as "item", so this keeps things consistent.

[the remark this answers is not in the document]
drewnoakes: Thanks but I prefer this here, alongside its partner. We colocate types like this in many places in the project/repo.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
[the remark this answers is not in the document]
drewnoakes: This method was removed.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
drewnoakes: The new code displays `Text` only when it differs from `Url`, and includes the `Url` in the `title`, so that it appears in a tool tip.

src/Aspire.Hosting/Dashboard/proto/Partials.cs
```diff
@@ -42,7 +53,9 @@ public static Resource FromSnapshot(ResourceSnapshot snapshot)
 
         foreach (var property in snapshot.Properties)
         {
-            resource.Properties.Add(new ResourceProperty { Name = property.Name, Value = property.Value });
+            var isSensitive = s_knownSensitiveResourcePropertyNames.Contains(property.Name);
+
+            resource.Properties.Add(new ResourceProperty { Name = property.Name, Value = property.Value, IsSensitive = isSensitive });
```
davidfowl, line 58: One thing missing is the ability to set this on the CustomResourceSnapshot https://github.com/dotnet/aspire/blob/0490bcbc4318c22774d162b3f49382f148cfa6b2/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs#L106. That would flow here instead of hard coding any known properties.

JamesNK: It would be good to have a way to control what properties are sensitive from the app model, but that can happen in a follow up PR.

drewnoakes: Updated.

I originally wanted to create a separate schema that's sent by the resource service on connection, that defines things like known properties and how to handle them. Then we wouldn't end up merging and sending static reference data along with dynamic property data. But that's beyond the scope of this PR.

davidfowl: We should make it a property (it's breaking change to make it required).

davidfowl (approved):

src/Aspire.Hosting/Dashboard/proto/Partials.cs
drewnoakes: Rolled back the API change.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -146,6 +146,8 @@ message ResourceProperty {
...
```
JamesNK, line 150: This shouldn't be optional. A new bool field will just default to false.
```suggestion
    bool is_sensitive = 4;
```

JamesNK (changes requested):

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
drewnoakes: https://protobuf.dev/programming-guides/proto3/#field-presence suggests that in proto3, the `optional` modifier on message type fields does nothing. We have many usages in this file. Should they all be removed?

I feel like I'm missing something giving you're blocking the PR on this.

JamesNK: That doc isn't very clear. Presence is a bit of a mess in Protobuf. proto2 had it, proto3 removed it, but lots of people still want it so they kind of added it back.

`optional` adds some prescence methods - `HasXXX`, `ClearXXX`, etc - to generated types in C#.

There is no reason we should have those for a bool that defaults to the right value - false - so they're not needed here.

I think removing `optional` might be a breaking change. I'm not sure. I think it's fine to leave in the proto file as a guide for what fields are set or not.

src/Aspire.Dashboard/Components/Controls/PropertyGrid.razor.cs
```diff
@@ -6,13 +6,81 @@
...
     private static readonly RenderFragment<TItem> s_emptyChildContent = _ => builder => { };
 
+    private static readonly GridSort<TItem> s_defaultNameSort = GridSort<TItem>.ByAscending(vm => vm.Name);
+    private static readonly GridSort<TItem> s_defaultValueSort = GridSort<TItem>.ByAscending(vm => vm.Value);
+
     [Parameter, EditorRequired]
     public IQueryable<TItem>? Items { get; set; }
 
+    [Parameter]
+    public Func<TItem, object?> ItemKey { get; init; } = static item => item.Name;
```
JamesNK, line 82: What happens if the name isn't unique? There are a number of situations where we're kind of abusing the property grid to show things other than name/values. I wonder what will happen if someone accidently or indentionally adds items where the name isn't unique.

If there an exception thrown by the control? Or does it render but there are later display errors when updating.

I'm just wondering if ItemKey hasn't been changed, should validate the items names?

cc @vnbaaij

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
drewnoakes: Well in this case the field is actually optional, so what's the issue with leaving the modifier on the field here? A sender doesn't have to specify this value, so can skip it. This documents that.

src/Aspire.Dashboard/Components/Controls/PropertyGrid.razor.cs
drewnoakes: Yes, this requires items to have unique names, which is a change in the case of duplicate names. There isn't much of a way around this as far as I can tell, shy of adding some other kind of identifier to the interface.

Duplicate keys causes exception:

> InvalidOperationException: More than one sibling of component 'Microsoft.FluentUI.AspNetCore.Components.FluentDataGridRow`1[TItem]' has the same key value, 'Property'. Key values must be unique.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
JamesNK: optional means you there is a tri-state boolean. true/false/unset. It's simpler to understand if optional isn't present. In that case true and false are the only options and there is no ambiguity about what unset means.

The comment could be expanded to say "Defaults to false if unset." to remove the ambigutiy. It it's simpler to just remove optional.

JamesNK: > I think removing optional from existing fields might be a breaking change. I'm not sure. I think it's fine to leave in the proto file as a guide for what fields are set or not. Even if it isn't used at runtime, it's still some good documentation for humans.

If you think optional is more confusing than it's worth, I'm coming around to removing it altogether. It could be replaced with a comment on "required" reference fields that a value is required.

drewnoakes: I did some reading here. Let me check what I understand.

- At the wire level, `optional` makes no difference to whether a message is considered well-formed or not. Any difference is purely at the application level where `optional` communicates that receiving applications are designed to handle messages that are missing this field.

- Removing `optional` from the field still doesn't require it to be present in order to deserialize the message. IIUC, it's not a breaking change to add or remove `optional` from a field.

- On the receiver side, slightly different code is generated, with the `optional` variant adding some presence state and accessors. The only difference is being able to differentiate between `false` being explicitly set and `false` as a default value when omitted.

Per the TODO comment a few lines above here, and as mentioned in https://github.com/dotnet/aspire/pull/5380#discussion_r1768336765, I want `display_name` and `is_sensitive` to move to reference data and be optional on this message. That reference data would provide defaults, and any value here could override that. But for that to work these would be optional, and the `Has*` properties needed. That work would address #1644.

I've pushed an updated comment, copied the TODO and linked to the issue.

drewnoakes:
@JamesNK please re-review and both @davidfowl and I are waiting to build on this for other work (e.g. https://github.com/dotnet/aspire/issues/4974).

JamesNK (approved):

PR #5387, opened 2024-08-22
Fix blob items clashing with each other during stable builds

joperezr:
## Description

Once we stabilize builds, we typically hit issues whenever we need to re-spin a new build given the previous build would have already published blobs using stable versioning as the new build will try to push again and these builds would clash. This change is making it so that the version we use for the blobs we publish won't go stable even when the rest of the repo is producing stable packages. This version is only used for these blobs, so no risk this affecting other things. cc: @radical 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5387)

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -26,9 +26,15 @@
   </ItemGroup>
 
   <Target Name="_PublishBlobItems">
+    <!-- 
+      For blob items for the Dashboard, we want to make sure that the version we get back is not stable, even when the repo is producing stable versions.
+      This is because we want to be able to re-spin the build if necessary without hitting issues of blob items clashing with each other. For this reason,
+      We will pass SuppressFinalPackageVersion as true when fetching the package version so that we get back a version with a prerelease suffix.
+    -->
     <MSBuild Projects="$(RepoRoot)src\Microsoft.NET.Sdk.Aspire\Microsoft.NET.Sdk.Aspire.csproj"
         Targets="ReturnPackageVersion"
-        SkipNonexistentProjects="false">
+        SkipNonexistentProjects="false"
+        AdditionalProperties="SuppressFinalPackageVersion=true">
       <Output TaskParameter="TargetOutputs" PropertyName="_PackageVersion" />
     </MSBuild>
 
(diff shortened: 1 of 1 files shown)


radical:
How will Dashboard packages be produced with stable versions? Do we want an internal pipeline build for this branch?

joperezr:
>How will Dashboard packages be produced with stable versions? Do we want an internal pipeline build for this branch?

Everything will continue to build as is, and have the same versions as previously. This small piece of code is only in charge of determining what the blob path should be when these blobs get pushed to blob storage. For example, the relative blob path that gets calculated by `$(_UploadPathRoot)/$(_PackageVersion)/%(Filename)%(Extension)` will evaluate to something like: `aspire/8.1.0/aspire-dashboard-linux-arm64.zip` when building in a branch producing `8.1.0` stable versions. The problem with this, is that this will clash if we have to rebuild from that same branch, as the path will still be the same. We are getting that `8.1.0` version by looking at the package version of one of the projects in the repo. All these changes are doing, is to pass an extra property so that the version that is returned contains prerelease labels, which make it such that the path would instead look like `aspire/8.1.0-rtm.12345.01/aspire-dashboard-linux-arm64.zip`, and then in the given case where we have to re-spin there won't be a potential of clashing on paths as the version would change. So in summary, the only thing we are changing here is the version used to determine the blob path for those blobs, but we are not changing at all the actual build produced for the product. Let me know if that doesn't make sense.

radical:
> > How will Dashboard packages be produced with stable versions? Do we want an internal pipeline build for this branch?
> 
> Everything will continue to build as is, and have the same versions as previously. This small piece of code ...

That makes sense. Thank you for the detailed explanation :)

radical (approved):

PR #5394, opened 2024-08-23
WaitFor/WaitForCompletion implementation.

mitchdenny:
## Description

This PR introduces two new extension APIs to the `IResourceBuilder<T>` interface:

```diff
namespace Aspire.Hosting;

public class ResourceBuilderExtensions
{
+    public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency) where T : IResource
+    public static IResourceBuilder<T> WaitForCompletion<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency) where T : IResource
}
```

Using these APIs it is possible to block the start of one resource depending on the start or completion of another resource.

```csharp
var catalogDbApp = builder.AddProject<Projects.CatalogDb>("catalogdbapp")
                          .WithReference(catalogDb);

var catalogService = builder.AddProject<Projects.CatalogService>("catalogservice")
                            .WithReference(catalogDb)
                            .WithReplicas(2)
                            .WaitForCompletion(catalogDbApp);
```

The `WaitFor(...)` extension will wait until the resource enters the `Running` state, and the `WaitForCompletion` will wait until it is in the `Finished` state.

The `WaitFor`/`WaitForCompletion` methods build upon the eventing mechanism that was recently added. It waits for the resource started event and then uses the `ResourceNotification` service to wait until the resource reaches the desired state. In a subsequent PR we will build on this to support waiting until the resource is healthy in addition to just being in a running state.

Resources in a waiting state will have their state changed to `Waiting` with the same iconography as Starting (hence the small changes to dashboard/known state code).

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1556
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5394)


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,82 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    ///        .WaitFor(messaging);
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency) where T : IResource
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, async (e, ct) =>
+        {
+            var rls = e.Services.GetRequiredService<ResourceLoggerService>();
+            var resourceLogger = rls.GetLogger(builder.Resource);
+            resourceLogger.LogInformation($"Waiting for resource '{dependency.Resource.Name}' to enter the '{KnownResourceStates.Running}' state.");
```
davidfowl, line 582:
```suggestion
            resourceLogger.LogInformation("Waiting for resource '{Name}' to enter the '{State}' state.", dependency.Resource.Name, KnownResourceStates.Running);
```

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,82 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, async (e, ct) =>
+        {
+            var rls = e.Services.GetRequiredService<ResourceLoggerService>();
+            var resourceLogger = rls.GetLogger(builder.Resource);
+            resourceLogger.LogInformation($"Waiting for resource '{dependency.Resource.Name}' to enter the 'Finished' state.");
+
+            var rns = e.Services.GetRequiredService<ResourceNotificationService>();
+            await rns.PublishUpdateAsync(builder.Resource, s => s with { State = KnownResourceStates.Waiting }).ConfigureAwait(false);
+            await rns.WaitForResourceAsync(dependency.Resource.Name, targetState: "Finished", cancellationToken: ct).ConfigureAwait(false);
```
davidfowl, line 627: Not sure this is good enough.

https://github.com/davidfowl/WaitForDependenciesAspire/blob/389f02daf2a3903924875c601dc300371aa99e27/WaitForDependencies.Aspire.Hosting/WaitForDependenciesExtensions.cs#L140

mitchdenny: I think we should wait for `Exited` - but I am not sure about the best way of handling `FailedToStart` especially in the context of container restart scenarios.

davidfowl: It should fail the wait.

mitchdenny: OK this is fixed up. In the end I added an overload to the ResourceNotificationService which allows you to get the snapshot back and also provide a predicate (just makes the API easier to use rather than having to do your own wait loop).

Once I've got the snapshot I check for the exit code and if it is not the target (default 0) then it blows up which blocks the dependent resource starting. Notice that the `WaitForCompletion` only sets the state to Waiting it doesn't set its subsequent state.

Its up for the code firing off the events to interpret what an exception in any of the event subscribers means. For this event it stops the startup of the resource.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,113 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    ///        .WaitForCompletion(dbprep);
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> WaitForCompletion<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency, int exitCode = 0) where T : IResource
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, async (e, ct) =>
+        {
+            // TODO: Decide how we want to interpret inconsistent results from replicas of projects. For now
+            //       if we detect that the project is configured for replicas we will throw an exception.
+            if (dependency.Resource.Annotations.Any(a => a is ReplicaAnnotation ra && ra.Replicas > 1))
```
davidfowl, line 624: TryGetLastAnnotation?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,113 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+            {
+                throw new DistributedApplicationException("WaitForCompletion cannot be used with resources that have replicas.");
```
davidfowl, line 626: This shows up properly as "failed to start" right?

mitchdenny: Yes.

image

And has details:

image

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,111 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, async (e, ct) =>
+        {
+            var rls = e.Services.GetRequiredService<ResourceLoggerService>();
+            var resourceLogger = rls.GetLogger(builder.Resource);
+            resourceLogger.LogInformation("Waiting for resource '{Name}' to enter the '{State}' state.", dependency.Resource.Name, KnownResourceStates.Running);
+
+            var rns = e.Services.GetRequiredService<ResourceNotificationService>();
+            await rns.PublishUpdateAsync(builder.Resource, s => s with { State = KnownResourceStates.Waiting }).ConfigureAwait(false);
+            await rns.WaitForResourceAsync(dependency.Resource.Name, cancellationToken: ct).ConfigureAwait(false);
```
davidfowl, line 586: Does this also need to handle terminal states no?

mitchdenny: So here is my conundrum with this one. `WaitFor` blocks execution until something is starting right now. But will soon block until something is healthy.

It's conceivable that with restart features arriving that a failed start/unhealthy state will be transient - so do we want to push a dependent service into a terminal failed to start state if someone on the dashboard could just "give it a kick" to get it up and running.

davidfowl: Then you could restart this service itself right? I think until we get restart, we should also end waiting for known terminal states. Then we can revisit that decision.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -549,4 +551,111 @@ public static IResourceBuilder<T> ExcludeFromManifest<T>(this IResourceBuilder<T
...
+    /// var builder = DistributedApplication.CreateBuilder(args);
+    /// var messaging = builder.AddRabbitMQ("messaging");
+    /// builder.AddProject&lt;Projects.MyApp&gt;("myapp")
+    ///        .WithReference(messaging)
+    ///        .WaitFor(messaging);
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency) where T : IResource
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, async (e, ct) =>
```
davidfowl, line 578: Another downside to doing it this way instead of a single wait for callback with lots of data is that first one to throw will stop other errors from coming through since the wait is sequential.

mitchdenny: Maybe for this event we should invoke them all in parallel and then await them all together.

davidfowl (approved):

PR #5398, opened 2024-08-23
Logging ordering fixes and feedback

### LINKED ISSUE #4893: [AzureTools][Aspire] The sequence number of the log changes when returning "Resource" and then clicking "View" of the project or container under the "Logs" in the Dashboard again

JamesNK:
## Description

Following feedback from [the previous PR](https://github.com/dotnet/aspire/pull/5357) and discussion with Karol about what options we have for getting logs from DCP, it's clear that there isn't a robust, fast, simple way to stream correctly ordered console logs from the host. The only way to produce the correct order is to do a final sort on the client.

This PR:

* Removes flaky time based delay from the server. It isn't possible to perfectly order log lines before they are returned from the service. Don't attempt to delay sending while ordering.

* Shares code for how log entries are internally represented on the client (dashboard) and server (host). Parsing and sorting logic is shared to ensure the backlog and client results match in content and order.

* Improves how sorting is implemented. Log lines with dates are placed in sorted order, while those without dates are placed in the order they were received. This mades ordering somewhat complicated, but there are tests, and I think it's important for the dashboard to be resilient to receiving whatever data it gets given as logs. In saying that, I want to make sure DCP always includes log lines with a prefixed date.

* Changes ResourceLoggerService to use one lock. I didn't see the benefit of multiple. It makes locking easier to reason with and reduces the chance of accidential deadlock.

* Resource logger state backlog is cleared immediately when there are no longer any subscribers. Previously they were cleared via a notificated sent through a number of channels. A race condition could cause logs to be watched again before they were cleared. That would cause logs to be double returned: the backlog + all logs from the start again.

Fixes https://github.com/dotnet/aspire/issues/4893

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5398)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -494,13 +487,16 @@ private void StartLogStream<T>(T resource) where T : CustomResource
                     {
                         foreach (var (content, isError) in batch)
                         {
-                            if (!ResourceLoggerService.TryParseContentLineDate(content, out var dateTimeUtc))
+                            DateTime? timestamp = null;
+                            var resolvedContent = content;
+
+                            if (TimestampParser.TryParseConsoleTimestamp(resolvedContent, out var result))
```
eerhardt, line 493: Why do we parse the timestamp out here, just to turn around and put them back together in the `ResourceLoggerService`? Why not just leave the timestamp in the content, and pass it all the way up to the dashboard. And only need to parse it out once in the UI?

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -359,6 +331,11 @@ private event Action<InternalLogLine> OnNewLog
 
                 if (raiseSubscribersChanged)
                 {
+                    // Clear backlog immediately.
+                    // Avoids a race between message being subscription changed notification eventually clearing the
+                    // logs and someone else watching logs and getting the backlog + complete replay off all logs.
+                    ClearBacklog();
```
eerhardt, line 337: Isn't there still a race here? This happens outside of the lock. So we could release the lock above, and immediately a new subscriber could get added.

src/Shared/ConsoleLogs/LogEntries.cs
```diff
@@ -0,0 +1,151 @@
...
+                if (missingTimestampIndex == null)
+                {
+                    missingTimestampIndex = rowIndex;
+                }
+
+                continue;
+            }
+
+            // Add log entry if it is later than current entry.
+            if (logEntry.Timestamp.Value >= current.Timestamp.Value)
```
karolz-ms, line 89: What if you receive a bunch of records with the same timestamp value? Presumably you want to insert them in the order received, but I do not think the current implementation will do that? A test for that would probably be handy.

JamesNK: Yes it inserts in order.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
JamesNK: I did this so the content in the host LogEntry was consistent with the client LogEntry.

src/Shared/ConsoleLogs/LogEntries.cs
JamesNK: Test added.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
JamesNK: Added lock around unsubscribe.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
eerhardt: Why do we need them to be consistent? The host can just deal with the logs as lines of content.

it’s only the UI that wants them to be sorted by date time, so why would we force the host to parse the time out?

src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -36,6 +34,7 @@ public static bool TryParseConsoleTimestamp(string text, [NotNullWhen(true)] out
                 }
             }
 
+            var timestamp = match.ValueSpan;
             result = new(content.ToString(), DateTimeOffset.Parse(timestamp.ToString(), CultureInfo.InvariantCulture));
```
eerhardt, line 38:
```suggestion
            result = new(content.ToString(), DateTimeOffset.Parse(timestamp, CultureInfo.InvariantCulture));
```

No reason to create an intermediate string here.

karolz-ms (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
JamesNK: They don't need to be consistent; I just think it's simpler to reason about everything when it's consistent and it allows more code to be shared between the host and dashboard.

In the host, either the `ILogger` logs need to add a timestamp to log line content or DCP logs need to remove a timestamp from log line content. Either way, it's slightly weird.

I chose to make the DCP logs remove the time stamp over the alternative for the reasons above. The downside there are some extra string allocations, the upside is storing the backlog uses less memory. Date a long `DateTime` is smaller than vs 15ish UTF16 chars in the log message.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -282,65 +280,44 @@ void Log(InternalLogLine log)
...
             {
-                bool raiseSubscribersChanged;
-                lock (this)
-                {
-                    raiseSubscribersChanged = _onNewLog is null; // is this the first subscriber?
+                Debug.Assert(Monitor.IsEntered(_lock));
 
-                    _onNewLog += value;
-                }
+                var raiseSubscribersChanged = _onNewLog is null; // is this the first subscriber?
```
drewnoakes, line 318: It's unclear why we only raise a subscriber-change notification for the first subscriber. A comment would be good.

drewnoakes (approved):

PR #5403, opened 2024-08-23
Update package tags for "integration" and "library" terms

DamianEdwards:
Updates the package tags to use the "integration" term. The previously termed "component" packages have "library" added also, so we have "integration hosting" packages and "integration library" packages. The "component" tag isn't being removed as yet as the Visual Studio "Add Aspire package" experience still relies on it.

### DIFF
--- a/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
+++ b/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
     <IsPackable>true</IsPackable>
-    <PackageTags>aspire integration hosting azure</PackageTags>
+    <PackageTags>aspire resource integration hosting azure</PackageTags>
     <Description>Azure resource types for .NET Aspire.</Description>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconFullPath>$(SharedDir)Azure_256x.png</PackageIconFullPath>
--- a/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
+++ b/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
     <IsPackable>true</IsPackable>
-    <PackageTags>aspire resource integration hosting azure</PackageTags>
+    <PackageTags>aspire hosting azure</PackageTags>
     <Description>Azure resource types for .NET Aspire.</Description>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconFullPath>$(SharedDir)Azure_256x.png</PackageIconFullPath>
--- a/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
+++ b/src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
     <IsPackable>true</IsPackable>
-    <PackageTags>aspire hosting azure</PackageTags>
+    <PackageTags>aspire integration hosting azure</PackageTags>
     <Description>Azure resource types for .NET Aspire.</Description>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconFullPath>$(SharedDir)Azure_256x.png</PackageIconFullPath>
--- a/src/Aspire.Hosting.Sdk/Aspire.Hosting.Sdk.csproj
+++ b/src/Aspire.Hosting.Sdk/Aspire.Hosting.Sdk.csproj
@@ -3,7 +3,7 @@
   <Import Project="$(SharedDir)Workload.targets" />
 
   <PropertyGroup>
-    <PackageTags>aspire integration hosting sdk</PackageTags>
+    <PackageTags>aspire hosting sdk</PackageTags>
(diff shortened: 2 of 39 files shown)


src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
     <IsPackable>true</IsPackable>
-    <PackageTags>aspire hosting azure</PackageTags>
+    <PackageTags>aspire integration hosting azure</PackageTags>
```
eerhardt, line 6: This one doesn't really fall into the "integration" category, does it? It is more the "base Azure" library, but doesn't have any Azure resources in it.

src/Aspire.Hosting.Sdk/Aspire.Hosting.Sdk.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$(SharedDir)Workload.targets" />
 
   <PropertyGroup>
-    <PackageTags>aspire hosting sdk</PackageTags>
+    <PackageTags>aspire integration hosting sdk</PackageTags>
```
eerhardt, line 6: This isn't an integration.

src/Components/Directory.Build.props
```diff
@@ -6,7 +6,7 @@
     <IsAotCompatible>true</IsAotCompatible>
     <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
 
-    <ComponentCommonPackageTags>aspire component cloud</ComponentCommonPackageTags>
+    <ComponentCommonPackageTags>aspire integration library component cloud</ComponentCommonPackageTags>
```
eerhardt, line 9: I wonder if "library" is too generic?

We have: "hosting" and "library"? In my mind I've been calling them "client library", or Aspire Client Library Integrations.

eerhardt: But I'm not sure if `client-library` or `clientlibrary` tags make sense...

src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
DamianEdwards: Ah yep, i think I thought it was a meta-package. So folks will never reference this directly?

src/Components/Directory.Build.props
DamianEdwards: We could say "client" instead? Ping @IEvangelist and @jamesmontemagno for their thoughts.

A "client integration" and a "hosting integration" sounds pretty good.

src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
DamianEdwards: It has methods like `AddBicepTemplate` so this can be used standalone, which in my mind makes it an Azure integration package. Thoughts?

src/Components/Directory.Build.props
DamianEdwards: Ping @maddymontaquila too

IEvangelist: I agree that `library` is too generic. `client integration` makes the most sense to me. I'd also advocate for hosting packages have the `resource` tag, as they literally model resources.

1. If you're a hosting integration package:

  - `aspire`, `resource`, `integration`, `hosting`

1. If you're a client integration package:

  - `aspire`, `client`, `integration`

I also believe it would be advantageous of us to include tags from the libraries we're configuring, for example, `redis` for all Redis related packages.

DamianEdwards: Updated. Including tags from the client libraries we're integrating is a bigger more complex change that I'll leave alone for now. Feel free to create a new issue for that if you like.

eerhardt: I don't think we need both `resource` and `hosting`. We have `hosting` today, and the packages have `Hosting` in the name. Adding `resource` is redundant IMO.

src/Aspire.Hosting.Azure/Aspire.Hosting.Azure.csproj
eerhardt: The issue is that all Azure hosting packages reference this one. So it will always be the highest Azure package in the list (sorted by downloads). And today it is the highest package outside of the core Aspire Hosting packages that are already referenced in the AppHost.

image

So I think we should take away `integration` here so it doesn't show up in the list. We want new users to pick the concrete Azure types, not the base class.

DamianEdwards: OK fair enough, will update.

eerhardt (approved):
LGTM

DamianEdwards:
Failure looks like a Helix job timeout. Is this common @radical?

jamesmontemagno (approved):

radical:
> Failure looks like a Helix job timeout. Is this common @radical?

Only the `Cosmos` failure in `Aspire.Playground.Tests` is known to fail sometimes.

The other two:

1. [Mysql](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5403-merge-8a20c59a7b6e4829bc/Aspire.Hosting.MySql.Tests/1/console.cfdb6473.log?helixlogtype=result) - times out waiting for `s_mySqlReadyText = log => log.Contains("ready for connections") && log.Contains("port: 3306");`.

2. [RabbitMQ](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5403-merge-8a20c59a7b6e4829bc/Aspire.Hosting.RabbitMQ.Tests/1/console.853f4443.log?helixlogtype=result) is timing out after 1 minute because the docker image was being pulled before the container could be started.

For a separate PR: We should add `WaitForResources` in general before `WaitForTextAsync` in the tests.

radical:
> We should add `WaitForResources` in general before `WaitForTextAsync` in the tests.

Opened https://github.com/dotnet/aspire/issues/5414

Issue #5404, opened 2024-08-23
Internal Build should not rebuild packages after testing


joperezr:
For our internal builds, our build process is roughly the following:

1. build the product and pack (this builds packages)
2. install dev certs
3. run non-helix tests
4. install sdk+workload for testing (this builds a subset of packages, again)
5. run helix tests
6. run Pack, sign and publish (this builds all packages, again)
7. build workloads

This is not good given we are re-building packages in step 6 after tests have already run against the previously built packages (from steps 1 and 4) which potentially invalidates the test run if something is slightly different in that step 6. I found this when investigating #5396 as it seems like this is what is making all of our tests pass, but the actual final packages we push have a bug. We should fix this by:

- Ensuring that packages in our build are only built once (to avoid duplication)
- Ensure that after running tests, nothing is ever re-built to ensure the integrity of the packages and not invalidate testing.

cc: @radical

Issue #5396, opened 2024-08-23
[WebToolsE2E][Aspire] Unable to build and run aspire project after installing aspire 8.2.0


vmykagapuz:
**REGRESSION**
1. Worked in Aspire 8.1.0 and Aspire 8.2.0 previous build

**INSTALL STEPS**

1. Clean machine: Win11 x64 23H2
2. Install SDK 8.0.401
3. Install Aspire 8.2.0 (8.2.0-preview.1.24422.7)

**REPRO STEPS**

1. Open CMD, create an Aspire Starter App and run the project using following commands:
```
dotnet new aspire-starter -o AspireStarterApp
cd AspireStarterApp/AspireStarterApp.AppHost
dotnet run
```

**ACTUAL**
Build and running the project fails with an error:

> C:\Users\aapt\.nuget\packages\aspire.hosting.orchestration.win-x64\8.2.0-preview.1.24422.7\buildTransitive\Aspire.Hosti
> ng.Orchestration.win-x64.targets(2,3): error MSB4019: The imported project "C:\Users\aapt\.nuget\packages\aspire.hostin
> g.orchestration.win-x64\8.2.0-preview.1.24422.7\build\@PackageName@.targets" was not found. Confirm that the expression
> in the Import declaration "..\build\@PackageName@.targets" is correct, and that the file exists on disk. [C:\Users\aap
> t\aspire1\aspire1.AppHost\aspire1.AppHost.csproj]
>   aspire1.ServiceDefaults -> C:\Users\aapt\aspire1\aspire1.ServiceDefaults\bin\Debug\net8.0\aspire1.ServiceDefaults.dll
>   aspire1.ApiService -> C:\Users\aapt\aspire1\aspire1.ApiService\bin\Debug\net8.0\aspire1.ApiService.dll
>   aspire1.Web -> C:\Users\aapt\aspire1\aspire1.Web\bin\Debug\net8.0\aspire1.Web.dll
>  
> Build FAILED.
>  
> C:\Users\aapt\.nuget\packages\aspire.hosting.orchestration.win-x64\8.2.0-preview.1.24422.7\buildTransitive\Aspire.Hosti
> ng.Orchestration.win-x64.targets(2,3): error MSB4019: The imported project "C:\Users\aapt\.nuget\packages\aspire.hostin
> g.orchestration.win-x64\8.2.0-preview.1.24422.7\build\@PackageName@.targets" was not found. Confirm that the expression
[...]

PR #5411, opened 2024-08-23
Fix issue where replacement text is not getting evaluated when pack

joperezr:
## Description

During our build, we currently are running Pack again after running tests (which is wrong and fixing it is tracked by #5404). Because of that, and due to a recent change that caused a new Restore operation to be called when sending our tests to Helix, the build is now trying to re-run pack one the package that was restored during test runs, but it is doing so by passing `/p:Restore=false /p:Build=false`. Currently, our ReplacementText targets only run if you are running the Build target as well, but if you run Pack and skip the Build, then the replacement targets won't run, meaning the new generated package won't have the placeholders replaced, which is what is causing #5396 to happen. This change is making it so that our Replacement targets also are hooked into the Pack target even when Build gets skipped which will workaround the issue. The ultimate fix will be to not try to re-pack the projects after testing, which will be tracked by #5404.

Fixes # https://github.com/dotnet/aspire/issues/5396

cc: @eerhardt @radical 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5411)

### DIFF
--- a/eng/ReplaceText.targets
+++ b/eng/ReplaceText.targets
@@ -34,7 +34,7 @@
   that use the processed file but preserve all other metadata.
   -->
   <Target Name="GenerateTextReplacementFiles"
-    BeforeTargets="Build;AssignTargetPaths"
+    BeforeTargets="Build;AssignTargetPaths;_GetPackageFiles"
     DependsOnTargets="_GatherTextReplacementFiles;_ProcessTextReplacementFiles"
     Condition="@(None->AnyHaveMetadataValue('PerformTextReplacement', 'True'))"
     >
--- a/eng/ReplaceText.targets
+++ b/eng/ReplaceText.targets
@@ -34,7 +34,7 @@
   that use the processed file but preserve all other metadata.
   -->
   <Target Name="GenerateTextReplacementFiles"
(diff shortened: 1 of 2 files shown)


radical (approved):
Waiting on the internal build for validation.

joperezr:
Found an issue with the initial approach as that was causing even more packages to get re-built post testing, which was causing other packages to be incorrectly built. This new approach is rolling that back and instead just avoiding causing a rebuild at all in that last pack call by no longer depending on restore when sending the tests to helix

joperezr:
I have validated with the internal build that this is working as expected. This is ready to go.

PR #5417, opened 2024-08-24
[ci] Internal build: build packages, and workload nugets only once

### LINKED ISSUE #5404: Internal Build should not rebuild packages after testing

radical:
This ensures that for the internal pipeline the tests are run against the final packages produced, instead of building packages twice - once for testing, and then for publishing.

- And track change in final packages being built before the workload tests are run

Fixes https://github.com/dotnet/aspire/issues/5404

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5417)

### DIFF
--- a/eng/pipelines/templates/BuildAndTest.yml
+++ b/eng/pipelines/templates/BuildAndTest.yml
@@ -21,14 +21,41 @@ parameters:
     default: false
 
 steps:
-  - script: ${{ parameters.buildScript }}
-            -restore -build
-            -pack
-            -configuration ${{ parameters.buildConfig }}
-            /p:ArchiveTests=true
-            /bl:${{ parameters.repoLogPath }}/build.binlog
-            $(_OfficialBuildIdArgs)
-    displayName: Build
+  # Internal pipeline: Build with pack+sign+archive
+  - ${{ if ne(parameters.runAsPublic, 'true') }}:
+    - script: ${{ parameters.buildScript }}
+              -restore -build
+              -pack
+              -sign $(_SignArgs)
+              -publish $(_PublishArgs)
+              -configuration ${{ parameters.buildConfig }}
+              /p:ArchiveTests=true
+              /bl:${{ parameters.repoLogPath }}/build.binlog
+              $(_OfficialBuildIdArgs)
+      displayName: Build
+
+  # Public pipeline - helix tests
+  - ${{ if eq(parameters.runAsPublic, 'true') }}:
+    - script: ${{ parameters.buildScript }}
+              -restore -build
+              -configuration ${{ parameters.buildConfig }}
+              -pack
+              /p:ArchiveTests=${{ lower(eq(parameters.runHelixTests, 'true')) }}
+              /bl:${{ parameters.repoLogPath }}/build.binlog
+              $(_OfficialBuildIdArgs)
+      displayName: Build
+
+  # Run on windows, either for internal pipeline, or for public+non-helix-tests job
+  - ${{ if and(eq(parameters.isWindows, 'true'), or(ne(parameters.runAsPublic, 'true'), ne(parameters.runHelixTests, 'true'))) }}:
+    - script: ${{ parameters.buildScript }}
+        -restore -build
+        -sign $(_SignArgs)
+        -publish $(_PublishArgs)
+        -configuration $(_BuildConfig)
+        /bl:${{ parameters.repoLogPath }}/build-workloads.binlog
+        -projects eng\workloads\workloads.csproj
+        $(_InternalBuildArgs)
+      displayName: Build Workloads
 
   - ${{ if or(ne(parameters.runAsPublic, 'true'), ne(parameters.runHelixTests, 'true')) }}:
     - ${{ if ne(parameters.isWindows, 'true') }}:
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -23,13 +23,45 @@ parameters:
 steps:
   - script: ${{ parameters.buildScript }}
             -restore -build
-            -pack
```
joperezr, line 26: Won't this mean that public builds won't run pack any longer?

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -23,13 +23,45 @@ parameters:
...
             -configuration ${{ parameters.buildConfig }}
             /p:ArchiveTests=true
             /bl:${{ parameters.repoLogPath }}/build.binlog
             $(_OfficialBuildIdArgs)
     displayName: Build
 
+  - ${{ if and(eq(parameters.isWindows, 'true'), ne(parameters.runAsPublic, 'true')) }}:
+    # On windows for internal builds prepare the final signed packages
+    - script: ${{ parameters.buildScript }}
```
joperezr, line 34: Do we need this to be a separate step? If not, one alternative would be to combine this and the above, and then use a variable to control the args for whether or not Sign happens

eng/pipelines/templates/BuildAndTest.yml
radical: They will run them on the `helix tests` job.
```diff
+  - ${{ if and(eq(parameters.runAsPublic, 'true'), eq(parameters.runHelixTests, 'true')) }}:
+    # Build packages for testing
+    - script: ${{ parameters.buildScript }}
+              -restore
+              -pack
+              -configuration ${{ parameters.buildConfig }}
+              /bl:${{ parameters.repoLogPath }}/pack.binlog
+              $(_OfficialBuildIdArgs)
+      displayName: Pack for testing
```

eng/pipelines/templates/BuildAndTest.yml
radical: Separating it has an additional benefit that we skip `pack` for `non-helix tests` job.

> If not, one alternative would be to combine this and the above, and then use a variable to control the args for whether or not Sign happens

Yeah, we can do this. It will end up a little less clear to read. I will see what it looks like.

eng/pipelines/templates/BuildAndTest.yml
joperezr: similar to the other question, but do we really need to do it in two steps? I think this complicates the build and also makes it such that a lot of 1ES steps have to run multiple times. Could we not condense everything into two build calls? (one for the product and one for the workloads)

eng/pipelines/templates/BuildAndTest.yml
joperezr: > Separating it has an additional benefit that we skip pack for non-helix tests job.

Does that really move the needle? From what I've seen (at least locally) pack won't take more than an additional minute or two once build is done.

>Yeah, we can do this. It will end up a little less clear to read.

That's a good point, but I also see other advantages with simplifying it:
- Reduces potential mistakes when variables/parameters are added to t he build (since potentially they won't get set on all the build calls)
- Facilitates local reproing of build issues. Today you have to run 4 build commands in order to mimic what happens in CI, it'd be nice if we can simplify that
- Avoids re-running a bunch of 1ES tasks that get injected each time we call build.cmd as they run each time artifacts are changed.

radical: > Avoids re-running a bunch of 1ES tasks that get injected each time we call build.cmd as they run each time artifacts are changed.

oh!

Agreed, working on it now.

radical:
Validation build - https://dev.azure.com/dnceng/internal/_build/results?buildId=2527148&view=results .

joperezr (commented):
Changes look good to me, thanks for cleaning this up @radical! One thing that I still feel a bit nervous about is the step for installing the workload to the separate dotnet-install in the artifacts dir for testing. Main worry is that: A) we run that after the non-helix tests have passed, so potentially rebuilding something that was already tested, and B) we also build this after signing packages, which might potentially re-build things unintentionally and not sign them.

joperezr (approved):
Meant to approve.

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -21,14 +21,41 @@ parameters:
...
-            /p:ArchiveTests=true
-            /bl:${{ parameters.repoLogPath }}/build.binlog
-            $(_OfficialBuildIdArgs)
-    displayName: Build
+  # Internal pipeline: Build with pack+sign+archive
+  - ${{ if ne(parameters.runAsPublic, 'true') }}:
+    - script: ${{ parameters.buildScript }}
+              -restore -build
+              -pack
+              -sign $(_SignArgs)
```
joperezr, line 29: NIT: Do we need to sign here? Can't we just keep the -sign param on the workloads build and have that sign everything?

radical: TBD: follow up

Issue #5371, opened 2024-08-21
`eventhubs-emulator` crashes on linux/helix


radical:
Starting `mcr.microsoft.com/azure-messaging/eventhubs-emulator:latest` on linux/helix fails. It fails on codespaces too.

```
fail: a.a.aaW[0]
      Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check Emulator Health Check with status Unhealthy completed after 26546.109ms with message 'Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy'
Retry 1 encountered an exception: Emulator Health Check failed.. Waiting 00:00:00 before next retry.
fail: a.a.aaW[0]
      Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check Emulator Health Check with status Unhealthy completed after 27109.5539ms with message 'Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy'
Retry 2 encountered an exception: Emulator Health Check failed.. Waiting 00:00:00 before next retry.
fail: a.a.aaW[0]
      Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy
fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
      Health check Emulator Health Check with status Unhealthy completed after 26307.1877ms with message 'Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy'
An error occurred: Emulator Health Check failed.
```

This needs to be investigated to see if it something in our setup or a bug with the docker image.
Example of a [failing build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=782955&view=ms.vss-test-web.build-test-results-tab).

davidfowl:
This is fixed now.

radical:
> This is fixed now.

Confirmed. Closing the issue.

Issue #5325, opened 2024-08-16
Tracking issue for testing tasks blocked due to no `docker desktop` available on CI


radical:
`Docker desktop` is not yet available on our CI. It is needed to have inter-container communication.

- [ ] EventHubs emulator - https://github.com/dotnet/aspire/issues/5371
- [ ] Testing any docker dependent tests on CI/Windows
- [x] https://github.com/microsoft/usvc/issues/164
- [ ] `RedisInsight` - https://github.com/dotnet/aspire/pull/5227#issuecomment-2293772081
- [x] `WithPgWeb` - https://github.com/dotnet/aspire/pull/5098
- [x] `Mongo` playground app - for `mongo-express` (https://github.com/dotnet/aspire/issues/5274)
- [ ] `TestShop` playground app - for `WithRedisCommander`

cc @eerhardt @sebastienros @joperezr

Alirexaa:
@radical 
[Add WithPgWeb ](https://github.com/dotnet/aspire/pull/5098)
https://github.com/dotnet/aspire/pull/5098#issuecomment-2293762429

davidfowl:
Yep

radical:
Stale.

Issue #5097, opened 2024-07-28
Container for Azure EventHubs emulator fails to startup successfully in Podman


jbparker:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The [local EventHubs emulator](https://learn.microsoft.com/en-us/dotnet/aspire/messaging/azure-event-hubs-component?tabs=dotnet-cli#app-host-usage) does not startup successfully when using the **Aspire.Hosting.Azure.EventHubs** extension method [RunAsEmulator()](https://github.com/dotnet/aspire/blob/8b603eab5bb8934c0b9a3a1a7d9bb9edd5c2828b/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs#L128).

Two containers (**eventhub** and **eventhub-storage**) are started, but the **eventhub** container fails with the message `Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy`.

After failing the health checks, the event hub container eventually exits:

![podman-containers](https://github.com/user-attachments/assets/0091b11a-e9b0-40cb-b6e6-6a19a819ae26)

### Expected Behavior

EventHubs containers created by Aspire hosting should start, pass health checks, and be accessible for Aspire component usage.

### Steps To Reproduce

https://github.com/jbparker/AspireEventHub

Forked from https://github.com/davidfowl/AspireEventHub and updated to use [Event Hubs emulator](https://learn.microsoft.com/en-us/dotnet/aspire/messaging/azure-event-hubs-component?tabs=dotnet-cli#app-host-usage) from https://github.com/dotnet/aspire/issues/3412

### Exceptions (if any)

The event hub container (`eventhubns`) logs health check errors:

> fail: a.a.aaW[0]
>       Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy
> fail: Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService[103]
>       Health check Emulator Health Check with status Unhealthy completed after 27907.2647ms with message 'Emulator Start up probe Unsuccessful. MetadataStore Health status: Unhealthy BlobStore Health status: Unhealthy'
> Retry 1 encountered an exception: Emulator Health Check failed.. Waiting 00:00:00 before next retry.
> ... (meanwhile, after 2nd retry)
> An error occurred: Emulator Health Check failed.

[...]

PR #5418, opened 2024-08-24
Add AzureFunctionsEndToEnd with prototype implementation

### LINKED ISSUE #5408: Add support for `IResourceWithAzureFunctionsConfig` interface
## Background and Motivation

To support Azure Functions worker projects, the Azure Functions host needs to be able to construct its own clients for Azure-based resources to facilitate acting as the middleman between the resource and triggers defined on the function.

Aspire doesn't currently provide any APIs for accessing the configuration properties associated with a resource (including the specific resource type) to support the construction of non-Aspired injected clients.

To mitigate this, we propose adding a new IResource interface to support emitting the configuration used by a given resource to consumers.

## Proposed API

The proposed API adds a new IResourceWithAzureFunctionsConfig to the Azure hosting package and implementations of the interface to the following resources: `AzureQueueStorageResource`, `AzureBlobStorageResource`, `AzureTableStorageResource`, `AzureServiceBusResource`, and `AzureEventHubsResource`.

```diff
// Package: Aspire.Hosting.Azure
namespace Aspire.Hosting.ApplicationModel;

+ public interface IResourceWithAzureFunctionsConfig : IResource
+ {
+   void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
+ }

// Package: Aspire.Hosting.Azure.Storage
namespace Aspire.Hosting.Azure;

public class AzureQueueStorageResource(string name, AzureStorageResource storage) : Resource(name),
    IResourceWithConnectionString,
    IResourceWithParent<AzureStorageResource>,
+   IResourceWithAzureFunctionsConfig
{
+    public void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
} 

public class AzureBlobStorageResource(string name, AzureStorageResource storage) : Resource(name),
    IResourceWithConnectionString,
    IResourceWithParent<AzureStorageResource>,
+   IResourceWithAzureFunctionsConfig
{
+    public void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
} 

public class AzureTableStorageResource(string name, AzureStorageResource storage) : Resource(name),
    IResourceWithConnectionString,
    IResourceWithParent<AzureStorageResource>,
+   IResourceWithAzureFunctionsConfig
{
+    public void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
} 
// Package: Aspire.Hosting.Azure.ServiceBus
namespace Aspire.Hosting.Azure;

public class AzureServiceBusResource(string name, Action<ResourceModuleConstruct> configureConstruct)
    : AzureConstructResource(name, configureConstruct),
      IResourceWithConnectionString,
+     IResourceWithAzureFunctionsConfig
{
+    public void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
} 

// Package: Aspire.Hosting.Azure.EventHubs
namespace Aspire.Hosting.Azure;

public class AzureEventHubsResource(string name, Action<ResourceModuleConstruct> configureConstruct) :
    AzureConstructResource(name, configureConstruct),
    IResourceWithConnectionString,
    IResourceWithEndpoints,
+   IResourceWithAzureFunctionsConfig  
{
+    public void ApplyAzureFunctionsConfiguration(Dictionary<string, object> target, string? connectionName = null)
}  
```  

## Usage Examples

```csharp
// User Code

var storage = builder.AddAzureStorage("storage");
var queue = storage.AddQueues("queue");
var blob = storage.AddBlobs("blob");

var funcApp = builder.AddAzureFunctionsProject<Projects.AspirePlusFunctions_Functions>("myfuncapp")
  .WithReference(queue)
  .WithReference(blob);

builder.Build().Run();

// Framework Code
public class AzureFunctionsProjectResource(string name) 
    : ProjectResource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithServiceDiscovery { }

public static class AzureFunctionsProjectResourceExtensions
{
    public static IResourceBuilder<AzureFunctionsProjectResource> WithReference(this IResourceBuilder<AzureFunctionsProjectResource> builder, IResourceBuilder<IResourceWithAzureFunctionConfig> source, string? name = null)
{    
  return builder.WithEnvironment(delegate (EnvironmentCallbackContext context)
    => source.Resource.ApplyAzureFunctionsConfiguration(context.EnvironmentVariables, name));
}
```

## Alternative Designs

- To future proof this interface, the type returned by the GetAzureFunctionsConfiguration is an opaque dictionary. The contract about what properties resources emit is informal between the target resources and the Functions app resource.

## Risks

- Because the Azure Functions host is using the original configuration information to construct its own clients, it's not able to take advantage of the features that come with Aspire-injected resources. We think this is fine given the requirements of the Functions host.
- This interface is designed specifically for consumption by Azure Functions (because of the implied contract above) but is accessible by anyone. There's no inherent risks to this other than confusion and mismatched expectations.

cc: @fabiocav @eerhardt @davidfowl @mattchenderson

### LINKED ISSUE #5412: Add `AzureFunctionsProjectResource` and extension methods
## Background and Motivation

In order to support modeling Azure Functions application in an Aspire application, we need a custom Azure Functions resource type that encapsulates:

- Azure Functions-specific resource configuration behavior
- Azure Functions-specific project mappings

The Azure Functions resource type will extend `ProjectResource` to support modeling functions workers as .NET projects. 

## Proposed API

All APIs proposed below are net-new and will reside in a new `Aspire.Hosting.Azure.Functions` package. This proposed API is designed to be used in conjunction with https://github.com/dotnet/aspire/issues/5408.

```csharp
// Package: Aspire.Hosting.Azure.Functions

public class AzureFunctionsProjectResource(string name) 
    : ProjectResource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithServiceDiscovery { }

public static class AzureFunctionsProjectResourceExtensions
{
    public static IResourceBuilder<AzureFunctionsProjectResource> AddAzureFunctionsProject<TProject>(this IDistributedApplicationBuilder builder, string name); where TProject : IProjectMetadata, new()

    public static IResourceBuilder<AzureFunctionsProjectResource> WithReference(this IResourceBuilder<AzureFunctionsProjectResource> builder, IResourceBuilder<IResourceWithAzureFunctionConfig> source, string? connectionName = null);
```
## Usage Examples

```csharp

var storage = builder.AddAzureStorage("storage");
var queue = storage.AddQueues("queue");
var blob = storage.AddBlobs("blob");

var eventHubs = builder.AddAzureEventHubs("eventhubs").AddEventHub("myhub");

var serviceBus = builder.AddAzureServiceBus("messaging").AddQueue("myqueue");

var funcApp = builder. AddAzureFunctionsProject <Projects.AspirePlusFunctions_Functions>("myfuncapp")
    .WithReference(eventHubs)
    .WithReference(queue)
    .WithReference(blob);

var apiService = builder.AddProject<Projects.AspirePlusFunctions_ApiService>("apiservice")
    .WithReference(queue)
    .WithReference(blob)
    .WithReference(eventHubs)
    .WithReference(serviceBus);

builder.AddProject<Projects.AspirePlusFunctions_Web>("webfrontend")
    .WithExternalHttpEndpoints()
    .WithReference(funcApp)
    .WithReference(apiService);

builder.Build().Run();
```

captainsafia:
This PR introduces an AzureFunctionsEndToEnd sample application with prototype implementations of certain APIs:

- APIs that currently target `IHostApplicationBuilder` have been reimplemented to target `HostBuilder`. This should be addressed on Azure Functions supports `IHostApplicationBuilder`.
- A custom `AzureFunctionsProjectResource` has been included that launches the `func host` as an executable resource. This is a temporary implementation until we support modeling a Functions application as a project resource. These APIs are formally proposed in https://github.com/dotnet/aspire/issues/5412. Support for mocking the Functions app as a project resource can be added by updating the mock API definitions as follows:

```csharp
public class AzureFunctionsProjectResource(string name) 
    : ProjectResource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithServiceDiscovery { }

public static class AzureFunctionsProjectResourceExtensions
  public static IResourceBuilder<FunctionAppResource> AddAzureFunctionsProject<TProject>(this IDistributedApplicationBuilder builder, string name) where TProject : IProjectMetadata, new()
    {
        var projectDirectory = Path.GetDirectoryName(new TProject().ProjectPath)!;
        var resource = new AzureFunctionsProjectResource(name);
        return builder.AddResource(resource)
            .WithAnnotation(new TProject())
            .WithHttpEndpoint(name: "http", env: "AZFUNCHOSTPORT")
            .WithEnvironment("FUNCTIONS_WORKER_RUNTIME", "dotnet-isolated")
            .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES", "true")
            .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES", "true")
            .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY", "in_memory")
            .WithEnvironment("ASPNETCORE_FORWARDEDHEADERS_ENABLED", "true")
            .WithEnvironment("FUNCTIONS_WORKER_RUNTIME", "dotnet-isolated")
            .WithOtlpExporter()
            .WithManifestPublishingCallback(async (context) => {
                context.Writer.WriteString("type", "function.v0");
                context.Writer.WriteString("path", context.GetManifestRelativePath(new TProject().ProjectPath));
                await context.WriteEnvironmentVariablesAsync(resource);
                context.Writer.WriteStartObject("bindings");
                foreach (var s in new string[]{ "http", "https"}) {
                    context.Writer.WriteStartObject(s);
                    context.Writer.WriteString("scheme", s);
                    context.Writer.WriteString("protocol", "tcp");
                    context.Writer.WriteString("transport", "http");
                    context.Writer.WriteBoolean("external", true);
                    context.Writer.WriteEndObject();
                }
                context.Writer.WriteEndObject();
            });
    }
}
```

And making the following changes to `AzureFunctionsEndToEnd.Functions.csproj`:

```xml
  <PropertyGroup>
    <RunCommand>func</RunCommand>
    <RunArguments>start --no-build --csharp --verbose --worker-runtime dotnetIsolated --port $(AZFUNCHOSTPORT)</RunArguments>
     <RunWorkingDirectory>bin/output</RunWorkingDirectory>
  </PropertyGroup>
```

- A custom `WithReference` implementation is implemented to support mapping resource strings as needed until https://github.com/dotnet/aspire/issues/5408 is implemented.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5418)

Fixes #5412 
Fixes #5408


radical:
A test should be added for this too, see https://github.com/dotnet/aspire/pull/5316/ for an example.

davidfowl:
I'm not sure we want to test this sample yet. This is part of bootstrapping the azure functions work.

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.ApiService/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+
+// Add service defaults & Aspire components.
+builder.AddServiceDefaults();
+builder.AddAzureQueueClient("queue");
+builder.AddAzureBlobClient("blob");
+builder.AddAzureEventHubProducerClient("eventhubs", static settings => settings.EventHubName = "myhub");
+
+var app = builder.Build();
+
+app.MapGet("/publish/asq", async (QueueServiceClient client, CancellationToken cancellationToken) =>
```
eerhardt, line 17: What do you think about simplifying this example, and just putting these as buttons on the web front end?

That way someone only needs to reason about the web app, and the functions app.

For example, right now in the AppHost Program.cs, the webfrontend has a reference to apiService, but it doesn't appear to use/call it.

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+                        .AddRuntimeInstrumentation();
+                })
+                .WithTracing(tracing =>
+                {
+                    tracing.AddAspNetCoreInstrumentation()
+                    // Uncomment the following line to enable gRPC instrumentation (requires the OpenTelemetry.Instrumentation.GrpcNetClient package)
+                    //.AddGrpcClientInstrumentation()
+                        .AddHttpClientInstrumentation();
+                });
+            // services.AddOpenTelemetryExporters(context.Configuration);
```
eerhardt, line 42: Where's this from?

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+                .WithMetrics(metrics =>
+                {
+                    metrics.AddAspNetCoreInstrumentation()
+                        .AddHttpClientInstrumentation()
+                        .AddRuntimeInstrumentation();
+                })
+                .WithTracing(tracing =>
+                {
+                    tracing.AddAspNetCoreInstrumentation()
+                    // Uncomment the following line to enable gRPC instrumentation (requires the OpenTelemetry.Instrumentation.GrpcNetClient package)
```
eerhardt, line 38: I think we can remove this commented out code. No need to put it in here if we aren't using gRPC.

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+var host = new HostBuilder()
+    .AddServiceDefaults()
+    .ConfigureFunctionsWorkerDefaults()
+    .ConfigureServices(services =>
+    {
+        services.AddOpenTelemetry()
+            .UseFunctionsWorkerDefaults()
+            .UseOtlpExporter();
```
eerhardt, line 13: How come this is here and not in `.AddServiceDefaults`?

davidfowl: IHostBuilder vs IHostApplicationBuilder

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/AzureFunctionsEndToEnd.Functions.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" />
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Update="host.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="local.settings.json">
```
eerhardt, line 26: This file doesn't appear to be in your change. Is that intentional?

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/Program.cs
eerhardt: AddServiceDefaults (called above) is also implemented in terms of IHostBuilder. It is in the `HostBuilderExtensions.cs` file.

src/Aspire.Hosting.Azure.Storage/AzureStorageResource.cs
```diff
@@ -38,6 +38,13 @@ public class AzureStorageResource(string name, Action<ResourceModuleConstruct> c
     /// </summary>
     public bool IsEmulator => this.IsContainer();
 
+    /// <summary>
+    /// Gets the connection string for the Azure Storage emulator.
+    /// </summary>
+    /// <returns></returns>
+    public ReferenceExpression GetEmulatorConnectionString() =>
+        ReferenceExpression.Create($"{AzureStorageEmulatorConnectionString.Create(blobPort: EmulatorBlobEndpoint.Port, queuePort: EmulatorQueueEndpoint.Port, tablePort: EmulatorTableEndpoint.Port)}");
```
eerhardt, line 46: @davidfowl - should this throw if `!IsEmulator`?

davidfowl: Yep

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.ApiService/Program.cs
eerhardt: > I think if we wanted the sample to be realistic

These "playground apps" aren't necessarily realistic and aren't intended to be samples. They are just intended for us to be able to try things out ourselves. Keeping them simple is a virtue so we don't have to maintain more than we need.

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/AzureFunctionsEndToEnd.Functions.csproj
captainsafia: Yeah, we can actually nuke this.

I believe this is autogenerated by the Azure Functions templates. The `local.settings.json` file is used to drive configuration options to the Core Tools (like whether or not this is a dotnet worker for example). Since we configure these ourselves when call the Core Tools CLI directly from the Aspire executable resource we don't really need it.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+    /// <param name="builder"></param>
+    /// <param name="name"></param>
+    /// <returns></returns>
+    public static IResourceBuilder<AzureFunctionsProjectResource> AddAzureFunctionsProject<TProject>(this IDistributedApplicationBuilder builder, string name) where TProject : IProjectMetadata, new()
+    {
+        var projectDirectory = Path.GetDirectoryName(new TProject().ProjectPath)!;
+        var resource = new AzureFunctionsProjectResource(name, "func", projectDirectory);
+
+        // Add the default storage resource if it doesn't already exist.
+        var storage = builder.Resources.OfType<AzureStorageResource>().FirstOrDefault(r => r.Name == "azure-functions-default-storage");
```
captainsafia, line 24: Are we assuming that if a user wants to configure their own `AzureWebJobsStorage` they must configure it with this resource name for us to pick it up?

That seems like a sensible constraint to me but wanted to confirm that's what's goin on here.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+                foreach (var item in data.Model.Resources.OfType<AzureFunctionsProjectResource>())
+                {
+                    if (item.HostStorage == storage)
+                    {
+                        removeStorage = false;
+                    }
+
+                    // Before the resource starts, we want to apply the azure functions specific environment variables.
+                    // we look at all of the environment variables and apply the configuration for any resources that implement IResourceWithAzureFunctionsConfig.
+                    item.Annotations.Add(new EnvironmentCallbackAnnotation(static context =>
```
captainsafia, line 44: Nice!

With this model, we could update more integrations packages to implement the `IResourceWithAzureFunctionsConfig` interface and they would light up in the Azure Functions resource without any additional updates on the `Aspire.Hosting.Azure.Functions` package.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+                            };
+
+                            if (config is not null)
+                            {
+                                valuesToRemove.Add(envName);
+                                functionsConfigMapping[name] = config;
+                            }
+                        }
+
+                        // REVIEW: We need to remove the existing values before adding the new ones as there's a conflict with the connection strings.
```
captainsafia, line 65: Is the conflict with the connection strings just the fact that the Funtions host expects connection strings in one format and the Aspire-injected resources don't but they both share the same resource names?

davidfowl: Needs to be investigated. Something in the functions or extensions codebase attempts to look at the `ConnectionStrings__{connectionName}` and throws because it's looking for a specific template format (Key1=Value1;Key2=Value2)

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
davidfowl: Nope, they just need to use `WithHostStorage` to do so. This default is only effective if they don't call that method (we keep this auto injected default)

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/AzureFunctionsEndToEnd.Functions.csproj
davidfowl: Is that representative of a real functions project though? Or should we add this file?

captainsafia: > Is that representative of a real functions project though? Or should we add this file?

The default Functions templates put this file in the `.gitignore` as well ([ref](https://github.com/Azure/azure-functions-templates/blob/f27baf3eb1aeb284739721d63c9c68059e3f27a6/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/_gitignore#L4-L5)) so I think it is fine to omit it. We'll want to make sure all the settings in it work correctly once we move to the `dotnet run` model though.

src/Aspire.Hosting.Azure.Functions/Aspire.Hosting.Azure.Functions.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <PackageTags>aspire hosting azure functions</PackageTags>
+    <Description>Azure fuinctions resource types for .NET Aspire.</Description>
```
eerhardt, line 7:
```suggestion
    <Description>Azure functions resource types for .NET Aspire.</Description>
```

radical:
`fail: AzureFunctionsEndToEnd.AppHost.Resources.funcapp[0] 1: 2024-09-05T07:09:43.6820000Z Can't determine Project to build. Expected 1 .csproj or .fsproj but found 2`

radical:
Also, the eventhubs emulator is failing in the playground test with https://github.com/dotnet/aspire/issues/5371 .

davidfowl:
> Also, the eventhubs emulator is failing in the playground test with https://github.com/dotnet/aspire/issues/5371 .

Do you see this locally or only on the CI?

radical:
> > > Also, the eventhubs emulator is failing in the playground test with #5371 .
> > 
> > 
> > Do you see this locally or only on the CI?
> 
> Also on codespaces . But not locally on macOS.

Could it be because of some need for Docker desktop? More logs would help.

davidfowl:
> Could it be because of some need for Docker desktop? More logs would help.

Yep:

- https://github.com/dotnet/aspire/pull/4192

davidfowl:
Same as https://github.com/dotnet/aspire/issues/5097

radical:
Another to add to https://github.com/dotnet/aspire/issues/5325 then :|

davidfowl:
This ups the urgency of this feature 😄. We'll see if we can work around the need for docker desktop in aspire. Do you recommend we disable the event hubs in the application to unblock this PR?

eerhardt:
> This ups the urgency of this feature 😄. We'll see if we can work around the need for docker desktop in aspire. Do you recommend we disable the event hubs in the application to unblock this PR?

Another option is to get docker desktop installed on our test machines.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
davidfowl: We can't avoid needing a name and a storage resource.

davidfowl:
> Another option is to get docker desktop installed on our test machines.

Docker desktop shouldn't be required, we have a real product gap.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
eerhardt: I'm saying - instead of hard-coding a name `"azure-functions-default-storage"`. Would it be better if we created an internal Annotation (or similar) that only we could add to "mark" the AzureStorageResource as the default one? That way we don't rely on the Name, which is something that a user can manipulate.

eerhardt:
> Docker desktop shouldn't be required, we have a real product gap.

Agreed. But that doesn't mean we shouldn't fully test this feature until the product gap is addressed.

We've known about this gap for a few months now. I guess I don't understand why "This ups the urgency of this feature". Any time a container needs to talk to another container, and Docker desktop isn't installed, it doesn't work today. This is just another "+1".

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
davidfowl: I don't understand how that would work. This resource needs to exist at some point and at that point it needs to have a name.

eerhardt: > and at that point it needs to have a name.

That's fine, we can use some name like `"azure-functions-default-storage"` to give it a name.

I'm talking about how we recognize which AzureStorageResource is the default one. Basing that check on the name doesn't seem great. Imagine if a user wrote code like:

```C#
builder.AddAzureStorage("azure-functions-default-storage");

builder.AddAzureFunctionsProject<Projects.MyFuncs>("MyFuncs");
```

if they did that, this code would pick up the storage account they added on line 1. But instead we really want them to say:

```C#
var defaultStorage = builder.AddAzureStorage("azure-functions-default-storage");

builder.AddAzureFunctionsProject<Projects.MyFuncs>("MyFuncs")
    .WithHostStorage(defaultStorage);
```

davidfowl:
If installing docker desktop is quick then I agree 😄

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
davidfowl: Ah, you want us to mark the default one. That's fine. If they did the above it would explode saying that they are multiple resources with the same name. Our best bet is to pick a name something that has a low likelyhood of collision

captainsafia:
> `fail: AzureFunctionsEndToEnd.AppHost.Resources.funcapp[0] 1: 2024-09-05T07:09:43.6820000Z Can't determine Project to build. Expected 1 .csproj or .fsproj but found 2`

There's ongoing discussion about this in https://github.com/Azure/azure-functions-core-tools/issues/3594 and a potential bug fix in https://github.com/Azure/azure-functions-core-tools/pull/3773 although it hasn't been released yet... :/

radical:
Would you mind opening an issue to track the issue on our side, so we can enable the test when it is resolved, and use `[ActiveIssue("https://github.com/dotnet/aspire/issues/NNNN", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnCI))]` on the test?

captainsafia:
> Would you mind opening an issue to track the issue on our side, so we can enable the test when it is resolved, and use `[ActiveIssue("https://github.com/dotnet/aspire/issues/NNNN", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnCI))]` on the test?

Thanks for the pointer -- I'll update the PR with this.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+/// Extension methods for <see cref="AzureFunctionsProjectResource"/>.
+/// </summary>
+public static class AzureFunctionsProjectResourceExtensions
+{
+    /// <summary>
+    /// Adds an Azure Functions project to the distributed application.
+    /// </summary>
+    /// <typeparam name="TProject">The type of the project metadata, which must implement <see cref="IProjectMetadata"/> and have a parameterless constructor.</typeparam>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/> to which the Azure Functions project will be added.</param>
+    /// <param name="name">The name to associated with the Azure Functions project. This name will be used for service discovery when referenced in a dependency.</param>
```
radical, line 18:
```suggestion
    /// <param name="name">The name to be associated with the Azure Functions project. This name will be used for service discovery when referenced in a dependency.</param>
```

tests/helix/send-to-helix-inner.proj
```diff
@@ -108,6 +111,17 @@
...
 
+  <ItemGroup Condition="'$(NeedsAzureFunctionsCli)' == 'true'">
+    <HelixCorrelationPayload Include="azure-func-cli" Destination="azure-func-cli" Uri="$(_AzureFunctionsCliUrl)" />
+
+    <HelixPreCommand Condition="'$(OS)' != 'Windows_NT'" Include="(cd ${HELIX_CORRELATION_PAYLOAD}/azure-func-cli; chmod +x func gozip)" />
+
+    <HelixPreCommand Condition="'$(OS)' == 'Windows_NT'" Include="set PATH=%HELIX_CORRELATION_PAYLOAD%\azure-func-cli%3B%PATH%" />
+    <HelixPreCommand Condition="'$(OS)' != 'Windows_NT'" Include="export PATH=$HELIX_CORRELATION_PAYLOAD/azure-func-cli:$PATH" />
+
+    <HelixPreCommand Condition="'$(OS)' != 'Windows_NT'" Include="find $HELIX_CORRELATION_PAYLOAD/azure-func-cli -type d" />
```
radical, line 122: This can be dropped now.

```suggestion
```

captainsafia: I know it's effectively inert but I still think it's worthwhile to keep until the Core Tools update is unblocked...unless unnecessarily uploading this payloads to the Helix machine will cause unforeseen issues?

playground/Directory.Build.targets
```diff
@@ -17,4 +20,8 @@
   <PropertyGroup Condition="'$(SkipDashboardProjectReference)' == 'true'">
     <DefineConstants>SKIP_DASHBOARD_REFERENCE;$(DefineConstants)</DefineConstants>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(SkipEventHubsEmulation)' == 'true'">
+    <DefineConstants>SKIP_EVENTHUBS_EMULATION;$(DefineConstants)</DefineConstants>
+  </PropertyGroup>
```
radical, line 26:
```suggestion
  <PropertyGroup>
    <DefineConstants Condition="'$(SkipDashboardProjectReference)' == 'true'">SKIP_DASHBOARD_REFERENCE;$(DefineConstants)</DefineConstants>
    <DefineConstants Condition="'$(SkipEventHubsEmulation)' == 'true'">SKIP_EVENTHUBS_EMULATION;$(DefineConstants)</DefineConstants>
  </PropertyGroup>
```

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
captainsafia: Tracked as a follow-up in https://github.com/dotnet/aspire/issues/5563.

playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.ApiService/Program.cs
captainsafia: We ended up simplifying it the other way around by removing the web UI from this sample app altogether. I think it's important to keep the API -> queue -> Azure Functions flow so this feels like a good landing place.

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
captainsafia: I don't think the collisions are super likely to happen but I added this to https://github.com/dotnet/aspire/issues/5563 to track in a follow-up.

tests/helix/send-to-helix-inner.proj
radical: This will not block sending the payload to helix. I had added this line to check that the payload was being unpacked in the expected directory, for my own debugging.

radical:
Also, should this be fixed? [log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5418-merge-c26ea5321b53455694/Aspire.Playground.Tests/1/console.9dcca55a.log?helixlogtype=result):
`/datadisks/disk1/work/A18408BE/w/B2030943/e/playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/AzureFunctionsEndToEnd.Functions.csproj : warning NU1603: Microsoft.Azure.Functions.Worker.Grpc 1.17.0 depends on Google.Protobuf (>= 3.25.2) but Google.Protobuf 3.25.2 was not found. An approximate best match of Google.Protobuf 3.25.3 was resolved. [/datadisks/disk1/work/A18408BE/w/B2030943/e/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj]`

captainsafia:
> Also, should this be fixed? [log](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5418-merge-c26ea5321b53455694/Aspire.Playground.Tests/1/console.9dcca55a.log?helixlogtype=result): `/datadisks/disk1/work/A18408BE/w/B2030943/e/playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.Functions/AzureFunctionsEndToEnd.Functions.csproj : warning NU1603: Microsoft.Azure.Functions.Worker.Grpc 1.17.0 depends on Google.Protobuf (>= 3.25.2) but Google.Protobuf 3.25.2 was not found. An approximate best match of Google.Protobuf 3.25.3 was resolved. [/datadisks/disk1/work/A18408BE/w/B2030943/e/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj]`

I just set a request to mirror Google.Protobuf@3.25.2.

radical:
We'll need to resolve this issue before merging ([build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=799520&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=5138e247-88cb-5ba0-8eee-e2fdd127e7e8&l=638)):
`.dotnet\sdk\8.0.303\Microsoft.Common.CurrentVersion.targets(5270,5): error MSB3030: Could not copy the file "D:\a\_work\1\s\artifacts\obj\AzureFunctionsEndToEnd.Functions\Release\net8.0\extensions.json" because it was not found.`

davidfowl (approved):

dciborow:
standing by ready for to do some testing on our internal project once this is released. Please feel free to reach out - dciborow.

captainsafia:
> standing by ready for to do some testing on our internal project once this is released. Please feel free to reach out - dciborow.

Great to hear! I'll give the README another run through today to make sure we've captured all the limitations/setup instructions correctly. This is just the first part of the integration and there will be a few kinks to iron out. 😅

PR #5424, opened 2024-08-24
[release/8.2] Enable signing and use stable packages

joperezr:
## Description

Enabling signing and stabilizing package versions for the release.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5424)

### DIFF
--- a/tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
+++ b/tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <ProductVersion>8.2.0</ProductVersion>
 
     <!-- no docker support on helix/windows yet -->
     <RunTestsOnHelix Condition="'$(OS)' == 'Windows_NT'">false</RunTestsOnHelix>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -14,7 +14,7 @@
     <BaselineVersionForPackageValidation>8.0.1</BaselineVersionForPackageValidation>
     <UseVSTestRunner>true</UseVSTestRunner>
     <!-- Enable to remove prerelease label. -->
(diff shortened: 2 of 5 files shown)


tests/Aspire.EndToEnd.Tests/Aspire.EndToEnd.Tests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <ProductVersion>8.2.0</ProductVersion>
```
eerhardt, line 5: Why do we have to hard-code this here and below? Why doesn't `PackageVersion` just work?

joperezr: This one can theoretically be $major.$minor, but below (meaning in the helix packaging props) we can't use PackageVersion because PackageVersion is evaluated for the project being built, which for our test projects won't be a stable prerelease version (meaning it would look like `8.2.0-ci` as opposed to `8.2.0`) The real fix here, is that to get these versions they should be built on-the-fly by making a call onto each one of these packages projects, and asking which version are those producing in order to get a full accurate set. The file below is instead doing the "lazy" approach of "guessing" what the right versions are, which is not accurate when producing stable versions as we have some packages we are stabilizing and some we are keeping in preview.

eerhardt: Oh, it is assuming that the current project's (i.e. the test project's) Version is the same as the built package's versions. Yeah, this seems like something to fix.

joperezr:
Validated in an internal build that this is working as expected.

PR #5425, opened 2024-08-25
Fix dashboard log parsing in host

JamesNK:
Fixes regression from https://github.com/dotnet/aspire/pull/5357

There is code in the Aspire host that listens to dashboard logs. It expects JSON logs from the dashboard, which are parsed and possibly written to the host. That allows for dashboard errors to be visible in the host console.

The previous PR changes Aspire to add a timestamp to executable log content, including the dashboard. Parsing timestamp prefixed messages as JSON fails.

Unfortunatly there were no unit tests here and there is no indication when running the host that there are errors when parsing the logs.

## Customer Impact

Customer won't see errors from the dashboard in the Aspire host console.

## Testing

Unit tests and manual testing.

Before (no error):

image

After (error in console):

image

## Risk

Low. Product change is to trim a timestamp (if present) from content before deserializing it.

## Regression?

Yes

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5425)

### DIFF
--- a/src/Aspire.Hosting/Aspire.Hosting.csproj
+++ b/src/Aspire.Hosting/Aspire.Hosting.csproj
@@ -37,6 +37,7 @@
     <Compile Include="$(SharedDir)LaunchProfile.cs" Link="LaunchProfile.cs" />
     <Compile Include="$(SharedDir)LaunchSettingsSerializerContext.cs" Link="LaunchSettingsSerializerContext.cs" />
     <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
+    <Compile Include="$(SharedDir)ConsoleLogs\*.cs" LinkBase="Utils\ConsoleLogs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Aspire.Hosting/Aspire.Hosting.csproj
+++ b/src/Aspire.Hosting/Aspire.Hosting.csproj
@@ -37,7 +37,7 @@
     <Compile Include="$(SharedDir)LaunchProfile.cs" Link="LaunchProfile.cs" />
     <Compile Include="$(SharedDir)LaunchSettingsSerializerContext.cs" Link="LaunchSettingsSerializerContext.cs" />
     <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
-    <Compile Include="$(SharedDir)ConsoleLogs\*.cs" LinkBase="Utils\ConsoleLogs" />
+    <Compile Include="$(SharedDir)ConsoleLogs\TimestampParser.cs" Link="Utils\ConsoleLogs\TimestampParser.cs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Shared/ConsoleLogs/TimestampParser.cs
+++ b/src/Shared/ConsoleLogs/TimestampParser.cs
@@ -1,0 +1,79 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Dashboard.ConsoleLogs;
+
+internal static partial class TimestampParser
+{
+    private static readonly Regex s_rfc3339RegEx = GenerateRfc3339RegEx();
+
+    public static bool TryParseConsoleTimestamp(string text, [NotNullWhen(true)] out TimestampParserResult? result)
+    {
+        var match = s_rfc3339RegEx.Match(text);
+
+        if (match.Success)
+        {
+            var span = text.AsSpan();
+            var timestamp = span[match.Index..(match.Index + match.Length)];
+
+            ReadOnlySpan<char> content;
+            if (match.Index + match.Length >= span.Length)
+            {
+                content = "";
+            }
+            else
+            {
+                content = span[(match.Index + match.Length)..];
+
+                // Trim whitespace added by logging between timestamp and content.
+                if (char.IsWhiteSpace(content[0]))
+                {
+                    content = content.Slice(1);
+                }
+            }
+
+            result = new(content.ToString(), DateTimeOffset.Parse(timestamp.ToString(), CultureInfo.InvariantCulture));
+            return true;
+        }
+
+        result = default;
+        return false;
+    }
+
+    // Regular Expression for an RFC3339 timestamp, including RFC3339Nano
+    //
+    // Example timestamps:
+    // 2023-10-02T12:56:35.123456789Z
+    // 2023-10-02T13:56:35.123456789+10:00
+    // 2023-10-02T13:56:35.123456789-10:00
+    // 2023-10-02T13:56:35.123456789Z10:00
+    // 2023-10-02T13:56:35.123456Z
+    // 2023-10-02T13:56:35Z
+    //
+    // Explanation:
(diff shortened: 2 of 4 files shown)


src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -0,0 +1,79 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Dashboard.ConsoleLogs;
+
+internal static partial class TimestampParser
```
JamesNK, line 10: Note that this is mostly a copy and paste from what is in the dashboard. There will be duplicate copies after this PR but this will be cleaned up by https://github.com/dotnet/aspire/pull/5398. I did this to minimize changes.

davidfowl:
Do we need to handle 8.2 apphost with 8.1 dashboard?

JamesNK:
The timestamps in dashboard logs are added by DCP. And DCP adding the timestamp is configured by the apphost. And reading the logs is in the apphost.

all the changes are in the apphost so there is no concern about mixing versions of apphost, DCP or dashboard.

src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -0,0 +1,79 @@
...
+    private static readonly Regex s_rfc3339RegEx = GenerateRfc3339RegEx();
+
+    public static bool TryParseConsoleTimestamp(string text, [NotNullWhen(true)] out TimestampParserResult? result)
+    {
+        var match = s_rfc3339RegEx.Match(text);
+
+        if (match.Success)
+        {
+            var span = text.AsSpan();
+            var timestamp = span[match.Index..(match.Index + match.Length)];
```
drewnoakes, line 21: I think this can just be:

```suggestion
            var timestamp = match.ValueSpan;
```

src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -0,0 +1,79 @@
...
+    // (?:[01][0-9]|2[0-3])                                - Two digits for the hour, restricted to 00-23
+    // :                                                   - Separator for the time
+    // (?:[0-5][0-9])                                      - Two digits for the minutes, restricted to 00-59
+    // :                                                   - Separator for the time
+    // (?:[0-5][0-9])                                      - Two digits for the seconds, restricted to 00-59
+    // (?:\\.\\d{1,9})                                     - A period and up to nine digits for the partial seconds
+    // Z                                                   - Literal, same as +00:00
+    // (?:[Z+-](?:[01][0-9]|2[0-3]):(?:[0-5][0-9]))        - Time Zone offset, in the form ZHH:MM or +HH:MM or -HH:MM
+    //
+    // Note: (?:) is a non-capturing group, since we don't care about the values, we are just interested in whether or not there is a match
```
drewnoakes, line 74: Using `RegexOptions.ExplicitCapture` instead of `?:` simplifies the pattern, and ensures that no groups are accidentally captured.

Consider also adding `RegexOptions.CultureInvariant`.

src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -0,0 +1,79 @@
...
+    [GeneratedRegex("^(?:\\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[12][0-9]|3[01])T(?:[01][0-9]|2[0-3]):(?:[0-5][0-9]):(?:[0-5][0-9])(?:\\.\\d{1,9})?(?:Z|(?:[Z+-](?:[01][0-9]|2[0-3]):(?:[0-5][0-9])))?")]
```
drewnoakes, line 75: The duplication here can be avoided by using `RegexOptions.IgnorePatternWhitespace` and including inline comments in the regex pattern. This means they can never get out of sync.

Something like:

```c#
    [GeneratedRegex("""
        ^                                          # Starts the string
        (\d{4})                                    # Four digits for the year
        -                                          # Separator for the date
        (0[1-9]|1[0-2])                            # Two digits for the month, restricted to 01-12
        -                                          # Separator for the date
        (0[1-9]|[12][0-9]|3[01])                   # Two digits for the day, restricted to 01-31
        [T ]                                       # Literal, separator between date and time, either a T or a space
        ([01][0-9]|2[0-3])                         # Two digits for the hour, restricted to 00-23
        :                                          # Separator for the time
        ([0-5][0-9])                               # Two digits for the minutes, restricted to 00-59
        :                                          # Separator for the time
        ([0-5][0-9])                               # Two digits for the seconds, restricted to 00-59
        (\.\d{1,9})                                # A period and up to nine digits for the partial seconds
        Z                                          # Literal, same as +00:00
        ([Z+-](?:[01][0-9]|2[0-3]):(?:[0-5][0-9])) # Time Zone offset, in the form ZHH:MM or +HH:MM or -HH:MM
        """,
        RegexOptions.ExplicitCapture | RegexOptions.InvariantCulture | RegexOptions.IgnorePatternWhitespace)]
```

src/Aspire.Hosting/Aspire.Hosting.csproj
```diff
@@ -37,6 +37,7 @@
     <Compile Include="$(SharedDir)LaunchProfile.cs" Link="LaunchProfile.cs" />
     <Compile Include="$(SharedDir)LaunchSettingsSerializerContext.cs" Link="LaunchSettingsSerializerContext.cs" />
     <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
+    <Compile Include="$(SharedDir)ConsoleLogs\*.cs" LinkBase="Utils\ConsoleLogs" />
```
drewnoakes, line 40: Adding with a wildcard like this could, in future, bring in bits we don't intend to be used nor need. Consider being explicit.

src/Shared/ConsoleLogs/TimestampParser.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Dashboard.ConsoleLogs;
+
+internal static partial class TimestampParser
+{
+    private static readonly Regex s_rfc3339RegEx = GenerateRfc3339RegEx();
```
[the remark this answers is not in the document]
JamesNK: I want to minimize changes here and for 8.2 backport. I'll make regex changes in https://github.com/dotnet/aspire/pull/5398

mitchdenny (approved):

src/Aspire.Hosting/Aspire.Hosting.csproj
JamesNK:
```suggestion
    <Compile Include="$(SharedDir)ConsoleLogs\TimestampParser.cs" Link="Utils\ConsoleLogs\TimestampParser.cs" />
```

drewnoakes (approved):

PR #5428, opened 2024-08-25
Test that StringComparers and StringComparisons are kept in sync

drewnoakes:
These two utility classes should have the same members, with equivalent values. This change adds missing properties to both classes, and adds a test to ensure that they're kept in sync in future.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5428)

### DIFF
--- a/src/Shared/StringComparers.cs
+++ b/src/Shared/StringComparers.cs
@@ -3,6 +3,8 @@
 
 namespace Aspire;
 
+// NOTE unit tests enforce that these two classes are kept in sync
+
 internal static class StringComparers
 {
     public static StringComparer ResourceName => StringComparer.OrdinalIgnoreCase;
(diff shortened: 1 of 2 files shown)


JamesNK:
What about using a type structure that forces them to be declared together? e.g.

```cs
public record KnownComparison(StringComparison Comparison, StringComparer Comparer);

public static class KnownStringFormats
{
    public static readonly ResourceName = new KnownComparison(StringComparison.OriginalIgnoreCase, StringComparer.OriginalIgnoreCase);
   // ... etc
}

// Usage
var resourcesByName = new Dictionary<string, ResourceViewModel>(KnownStringFormats.ResourceName.Comparer);
```

drewnoakes:
I think the test failure occurs when the current culture is also the invariant culture. Investigating.

> What about using a type structure that forces them to be declared together?

This adds some runtime overhead and makes the call site's code more verbose, so I'm not a fan personally.

mitchdenny (approved):
This is a good idea.

PR #5434, opened 2024-08-26
Store non-sensitive UI state without protection

### LINKED ISSUE #4949: Protected storage configuration is lost between dashboard runs from a container

### Describe the bug

The dashboard uses protected storage to save orientation/size of details views. This data is protected using data protection. Data protection keys are recreated each time the dashboard is run from the container in typical usage.

The end result is orientation/size of details views is lost between runs of the standalone dashboard. Theme selection doesn't have this problem because it is saved to a cookie.

Possible solutions

This data isn't sensitive and doesn't need to be encrypted. We can use a solution that doesn't do any encryption to fix this issue.

* Switch to using a cookie.
* Continue to use localStorage but don't protect saved data. I think it's important we **don't** completely replace `ProtectedLocalStorage` in DI with an instance that doesn't encrypt. It might be used internally by Blazor or other libraries to protect data that is actually sensitive. Perhaps have a keyed registration of `ProtectedLocalStorage` that isn't protected by specifying `IDataProtectionProvider` option that doesn't do anything.

### Expected Behavior

Orientation/size of details views is remembered.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

JamesNK:
## Description

Some UI state is stored in protected browser storage: details orientation and size.

Unfortunately protected storage becomes unusable if the dashboard's data protection keys change. This can happen if the dashboard is launched from a docker container (each run has new keys). I've also seen it when updating Aspire version - perhaps the dashboard being located in a different place or a different file name causes new keys? IDK.

The fix is to store long term (localStorage) non-sensitive state without protected. Values will be remembered between runs, and there won't be errors silently thrown in the background when attempting to unencrpyt and failing.

Short term storage (sessionStorage) that is used to store potentially sensitive page data still only allows storage with protection.

I'll double check this is safe with Blazor and security folks.

Fixes https://github.com/dotnet/aspire/issues/4949

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5434)


src/Aspire.Dashboard/Components/Controls/SummaryDetailsView.razor.cs
```diff
@@ -293,13 +292,13 @@ static void GetPanelSizes(
     private string GetSizeStorageKey()
     {
         var viewKey = ViewKey ?? NavigationManager.ToBaseRelativePath(NavigationManager.Uri);
-        return $"SplitterSize_{Orientation}_{viewKey}";
+        return $"Aspire_SplitterSize_{Orientation}_{viewKey}";
```
JamesNK, line 295: I changed the keys to prevent the dashboard from using existing encrypted values. This means existing state is lost when updating to the next Aspire version, but it appears to be lost today anyway when protection keys change.

Going forward we can rely on these values being available long term.

drewnoakes: While we're at it, can we give this a better name? There might be multiple splitters in the dashboard. Let's make this more specific.

src/Aspire.Dashboard/Model/BrowserStorage/LocalBrowserStorage.cs
```diff
@@ -1,13 +1,46 @@
...
+        var json = JsonSerializer.Serialize(value, s_options);
+
+        await SetJsonAsync(key, json).ConfigureAwait(false);
+    }
+
+    private ValueTask SetJsonAsync(string key, string json)
+        => _jsRuntime.InvokeVoidAsync("localStorage.setItem", key, json);
+
+    private ValueTask<string?> GetJsonAsync(string key)
+        => _jsRuntime.InvokeAsync<string?>("localStorage.getItem", key);
```
drewnoakes, line 45: These two methods return `ValueTask`. I'm not sure how often these would return completed tasks, but perhaps we also use value tasks on the new methods on `ILocalStorage`.

src/Aspire.Dashboard/Components/Controls/SummaryDetailsView.razor.cs
JamesNK: viewKey is there to support multiple splitters. It is what lets resources/structured logs/trace detail to have splitters with different settings.

drewnoakes (approved):
Approved, assuming the security review is completed before merge.

src/Aspire.Dashboard/Components/Controls/SummaryDetailsView.razor.cs
drewnoakes: Ok, makes sense.

JamesNK:
> Approved, assuming the security review is completed before merge.

Done over email

PR #5465, opened 2024-08-28
Persist dismissing the OTLP unsecured message bar

JamesNK:
## Description

Running the dashboard without OTLP auth will display a warning message bar. The message bar can be dismissed, but it reappears if the browser is refreshed.

It is annoying to repeatedly show the message to a user who has already seen the warning. They either have to close the message every time, or accept it is always visible.

This PR:

* Persists a flag to local storage when the message bar is dismissed
* Doesn't display message bar if previously dismissed
* Centralizes browser storage keys
* Adds tests

Note that the warning displayed in console logs on startup isn't impacted. It's always displayed.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
+++ b/src/Aspire.Dashboard/Components/Controls/UserProfile.razor
@@ -1,7 +1,7 @@
-﻿<AuthorizeView>
-    <Authorized>
-        @if (_showUserProfileMenu)
-        {
+﻿@if (_showUserProfileMenu)
+{
+    <AuthorizeView>
+        <Authorized>
             <div class="profile-menu-container">
                 <FluentProfileMenu Initials="@_initials"
                                    EMail="@_username"
@@ -28,6 +28,6 @@
                     </ChildContent>
                 </FluentProfileMenu>
             </div>
-        }
-    </Authorized>
-</AuthorizeView>
+        </Authorized>
+    </AuthorizeView>
+}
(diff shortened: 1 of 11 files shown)


src/Aspire.Dashboard/Components/Controls/UserProfile.razor
```diff
@@ -1,7 +1,7 @@
-﻿<AuthorizeView>
-    <Authorized>
-        @if (_showUserProfileMenu)
-        {
+﻿@if (_showUserProfileMenu)
```
JamesNK, line 1: Only include AuthorizeView on the page if the user profile should be displayed (which means the auth mode needs to be a certain value).

This avoids the need to include various authz services in tests when not needed.

drewnoakes (approved):

src/Aspire.Dashboard/Utils/BrowserStorageKeys.cs
```diff
@@ -0,0 +1,26 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.FluentUI.AspNetCore.Components;
+
+namespace Aspire.Dashboard.Utils;
+
+internal static class BrowserStorageKeys
+{
+    public const string UnsecuredTelemetryMessageDismissedKey = "Aspire_UnsecuredTelemetryMessage_Dismissed";
```
JamesNK, line 10:
```suggestion
    public const string UnsecuredTelemetryMessageDismissedKey = "Aspire_Telemetry_UnsecuredMessageDismissed";
```

PR #5478, opened 2024-08-29
Adding hub settings for Web PubSub

vicancy:
## Description

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5478)


davidfowl:
I think I'd prefer to see this look more like the other resources with child resources:

```C#
var wps = builder.AddAzureWebPubSub("wps1");
var hub = builder.AddHub("chat");

builder.AddProject<Projects.WebPubSubWeb>("webfrontend")
var web = builder.AddProject<Projects.WebPubSubWeb>("webfrontend")
    .WithExternalHttpEndpoints()
    .WithReference(wps);

// Now only works in production since localhost is not accessible from Azure
{
    hub.AddEventHandler($"{web.GetEndpoint("https")}/eventhandler/", systemEvents: ["connected"]);
}
```

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
```diff
@@ -0,0 +1,24 @@
...
+using Azure.Provisioning.WebPubSub;
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents an Azure Web PubSub Hub setting resource.
+/// </summary>
+/// <param name="name">The name of the resource.</param>
+/// <param name="webpubsub">The <see cref="AzureWebPubSubResource"/> that the resource belongs to.</param>
+public class AzureWebPubSubHubResource(string name, AzureWebPubSubResource webpubsub) : AzureConstructResource(name, s_empty),
```
davidfowl, line 13: This is a resource but it shouldn't be an `AzureConstructResource`.

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -64,4 +94,78 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
...
+    {
+        var handler = new WebPubSubEventHandler
+        {
+            UrlTemplate = urlValue,
+            UserEventPattern = userEventPattern,
+        };
+
+        if (systemEvents != null)
+        {
+            handler.SystemEvents = new BicepList<string>(systemEvents.Select(i => new BicepValue<string>(i)).ToArray());
```
davidfowl, line 162: I think you can do this:

```suggestion
            handler.SystemEvents = [..systemEvents];
```

playground/webpubsub/WebPubSub.AppHost/Program.cs
```diff
@@ -4,9 +4,14 @@
 builder.AddAzureProvisioning();
```
davidfowl, line 4: You can delete this now, it should be called by AddAzureWebPubSub

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
```diff
@@ -0,0 +1,24 @@
...
+    IResourceWithParent<AzureWebPubSubResource>
+{
+    /// <summary>
+    /// Gets the parent AzureWebPubSubResource of this AzureWebPubSubHubSettingResource.
+    /// </summary>
+    public AzureWebPubSubResource Parent => webpubsub;
+
+    internal List<Action<IResourceBuilder<AzureWebPubSubResource>, ResourceModuleConstruct, WebPubSubHub>> EventHandlers { get; } = new();
```
davidfowl, line 21: Remove the builder

```suggestion
    internal List<Action<ResourceModuleConstruct, WebPubSubHub>> EventHandlers { get; } = new();
```

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubResource.cs
```diff
@@ -14,6 +14,8 @@ public class AzureWebPubSubResource(string name, Action<ResourceModuleConstruct>
     AzureConstructResource(name, configureConstruct),
     IResourceWithConnectionString
 {
+    internal Dictionary<string, IResourceBuilder<AzureWebPubSubHubResource>> Hubs { get; } = new(StringComparer.OrdinalIgnoreCase);
```
davidfowl, line 17:
```suggestion
    internal Dictionary<string, AzureWebPubSubHubResource> Hubs { get; } = new(StringComparer.OrdinalIgnoreCase);
```

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -64,4 +94,78 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
...
+            }
+            else
+            {
+                var count = hubResource.EventHandlers.Count;
+                var urlParameter = new ProvisioningParameter($"{hubName}_url_{count}", typeof(string));
+                var eventHandler = GetWebPubSubEventHandler(urlParameter, userEventPattern, systemEvents, authSettings);
+                hubResource.EventHandlers.Add(eventHandler);
+                construct.Add(urlParameter);
+
+                builder.WithParameter(urlParameter.IdentifierName, urlExpression);
```
davidfowl, line 146:
```suggestion
                hubResource.Parent.Parameters[urlParameter.IdentifierName] = urlExpression;
```

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
vicancy: If it is not an AzureConstructResource, it is not allowed to use "ConfigureConstruct" to customize the hub properties as shown in the test case : 
```cs

        wps.AddHub("abc").ConfigureConstruct(construct =>
        {
            var hub = construct.GetResources().OfType<WebPubSubHub>().First(i => i.Name.Value == "abc");
            hub.Properties.Value!.AnonymousConnectPolicy = "allow";
        });
        ```

Comment (davidfowl) [reply src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs:13]: I see. Why make it empty by default the?

Comment (vicancy) [reply src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs:13]: because the WebPubSubHub has a parent service "WebPubSubService", which is only known in the WebPubSubResource's construct.....

Comment (vicancy) [reply src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs:13]: Ah I see, let me change it to get the parent service from construct

Comment (vicancy) [reply src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs:13]: And `GetBicepTemplateFile` is not expected to be supported by `AzureWebPubSubHubResource` as it is a subresource

Comment (davidfowl) [reply src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs:13]: This is a very nice pattern. I'll think about where we can apply this elsewhere (probably event hubs as well).

File: src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs [hunk 7]
```diff
@@ -64,4 +93,78 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
...
+            if (urlExpression.ManifestExpressions.Count == 0)
+            {
+                var eventHandler = GetWebPubSubEventHandler(urlExpression.Format, userEventPattern, systemEvents, authSettings);
+
+                // when urlExpression is literal string, simply add
+                hubResource.EventHandlers.Add(eventHandler);
+            }
+            else
+            {
+                var count = hubResource.EventHandlers.Count;
```
davidfowl, line 139: Is this right? Should the count be captured outside of this and read from the closure? These run after all of event handlers are added right? Won't they have the same value?

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
davidfowl: Spoke to @eerhardt, the user can customize the construct using the AddAzureWebPubSub and this resource doesn't need to be an AzureConstructResource.

```C#
wps.ConfigureConstruct(construct =>
{
    var hub = construct.GetResources().OfType<WebPubSubHub>().First(i => i.Name.Value == "abc");
    hub.Properties.Value!.AnonymousConnectPolicy = "allow";
});
```

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -56,6 +56,35 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
...
+
+                // create the hub settings with default values
+                if (hub.Properties.Value == null)
+                {
+                    hub.Properties = new WebPubSubHubProperties();
+                }
+                // add to construct
+                construct.Add(hub);
+                // invoke the configure if any
+                hubResource.ConfigureConstruct(construct);
```
eerhardt, line 81: With https://github.com/dotnet/aspire/pull/6258, we've gone away with these inline callbacks. Instead, the developer can `.ConfigureConstruct` on the parent object. That configure method gets the whole `Infrastructure` object, which contains all the bicep resources. The developer can tweak the bicep from there, even on child resources. For example, see 

https://github.com/dotnet/aspire/blob/4e7e0de000bb83f824cca2773892a22b5da89245/playground/cdk/CdkSample.AppHost/Program.cs#L64-L80

eerhardt: I guess this is the same as what @davidfowl says above - https://github.com/dotnet/aspire/pull/5478#discussion_r1799759275

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
eerhardt: > construct.GetResources().OfType<WebPubSubHub>().First(i => i.Name.Value == "abc");

FYI - in general this needs to be `IdentifierName` as sometimes `Name` gets autogenerated with a unique string.

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
```diff
@@ -1749,6 +1750,405 @@ param principalType string
         Assert.Equal(expectedBicep, manifest.BicepText);
     }
 
+    [Fact]
```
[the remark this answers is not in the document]
eerhardt: We can have a follow up issue to fix the existing tests, but new tests should be added in separate files. We've already started this, for example:

* https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Azure.Tests/AzureRedisExtensionsTests.cs
* https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Azure.Tests/AzureServiceBusExtensionsTests.cs
* https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
vicancy: But IdentifierName is get normalized with `Infrastructure.NormalizeIdentifierName` though, it is not necessarily the name user passed in?

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
vicancy: let me rename hubResource, it is actually the CDK object..

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -64,4 +92,82 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
...
+    /// <summary>
+    /// Add event handler setting with expression
+    /// </summary>
+    /// <param name="builder">The builder for a Web PubSub hub.</param>
+    /// <param name="urlTemplateExpression">The expression to evaluate the URL template configured for the event handler.</param>
+    /// <param name="userEventPattern">The user event pattern for the event handler.</param>
+    /// <param name="systemEvents">The system events for the event handler.</param>
+    /// <param name="authSettings">The auth settings configured for the event handler.</param>
+    /// <returns></returns>
+    public static IResourceBuilder<AzureWebPubSubHubResource> AddEventHandler(this IResourceBuilder<AzureWebPubSubHubResource> builder,
```
davidfowl, line 123: you'll want an overload that takes ReferenceExpression as well, that can be follow up

davidfowl (approved):

PR #5479, opened 2024-08-29
Add RedisStack Hosting

### LINKED ISSUE #1663: Support For Redis Stack Using Existing Redis Component
Use Existing Redis Component to add support for Redis stack. 

```cs
var redisStack = builder.AddRedis("redis", stack: true)
                         .WithRedisCommander();
```

and 

```cs 
var redisStack = builder.AddRedisContainer("redis", stack: true);
```

For more information on what redis stack is 
https://redis.io/docs/about/about-stack/

image

Alirexaa:
## Description

We add [RedisStack ](https://redis.io/about/about-stack/) as a new hosting resource.
redis stack

There are two container images for using RedisStack:
There's one intended for development time and another for deployment (`redis/redis-stack` vs. `redis/redis-stack-server`).

according to https://redis.io/docs/latest/operate/oss_and_stack/install/install-stack/docker :

>redis/redis-stack contains both Redis Stack server and Redis Insight. This container is best for local development because you can use the embedded Redis Insight to visualize your data.
>redis/redis-stack-server provides Redis Stack server only. This container is best for production deployment.

The difference between `redis/redis-stack` vs. `redis/redis-stack-server` is Redis Insight, We have a hosting resource for Redis Insight at https://github.com/dotnet/aspire/pull/5227.So we use `redis/redis-stack-server` for both development and deployment.

RedisStack provides an environment variable `REDIS_ARGS`. We can add a password to the container via the environment variable `REDIS_ARGS="--requirepass password"` (password support is not included in this PR yet.)

Fixes #1663

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [X] Yes
      - [X] Yes
      - [ ] No
      - [X] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [X] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5479)


Alirexaa:
For functional tests to pass, we need to import `docker.io/redis/redis-stack-server:7.4.0-v0` to `netaspireci.azurecr.io`

Alirexaa:
@eerhardt @mitchdenny, please take a look.

davidfowl:
Is there no way to make this part of the existing redis resource? This is a ton of duplication.

Alirexaa:
> Is there no way to make this part of the existing redis resource? This is a ton of duplication.

Yes, there is but isn't better to keep resources separate?
What is your proposal API?

eerhardt:
> Is there no way to make this part of the existing redis resource? This is a ton of duplication.

I was almost wondering if we should just replace the Image from `library/redis` to `redis/redis-stack-server` all up. Do we really need both?

Alirexaa:
> > Is there no way to make this part of the existing redis resource? This is a ton of duplication.
> 
> I was almost wondering if we should just replace the Image from `library/redis` to `redis/redis-stack-server` all up. Do we really need both?

we could do this but I should mention that in redis stack we should provide redis args in `REDIS_ARGS` environment variable.

for example
```csharp
return builder.WithAnnotation(new CommandLineArgsCallbackAnnotation(context =>
{
    context.Args.Add("--save");
    context.Args.Add(
        (interval ?? TimeSpan.FromSeconds(60)).TotalSeconds.ToString(CultureInfo.InvariantCulture));
    context.Args.Add(keysChangedThreshold.ToString(CultureInfo.InvariantCulture));
    return Task.CompletedTask;
}), ResourceAnnotationMutationBehavior.Replace);
```
should change to 

```csharp

builder.WithEnvironment("REDIS_ARGS", $"--save {(interval ?? TimeSpan.FromSeconds(60)).TotalSeconds.ToString(CultureInfo.InvariantCulture)} 

```
Also, we could provide a password by `REDIS_ARGS="--requirepass password"`, so the password should work in both development and deployment. referencing to https://github.com/dotnet/aspire/pull/4642

mitchdenny:
> Also, we could provide a password by `REDIS_ARGS="--requirepass password"`, so the password should work in both development and deployment. referencing to #4642

I tried getting this to work in a secure way but I think we'll need parameter support in AZD for run args.

eerhardt:
Thanks for the contribution here @Alirexaa. However, at this point I don't think we should be doing this work. It looks like the functionality of Redis Stack is moving into the common layer.

https://redis.io/blog/the-future-of-redis/

> Simplifying the Redis Experience

> First, we’re focusing on simplicity and the developer experience – both in the core and in our developer experience. Starting with Redis 8, we’re integrating Redis Stack capabilities including query and search, vector search, JSON document support, time series and probabilistic structures into the Community Edition, providing a single, comprehensive Redis that offers all the latest features without the need for separate downloads or packages.

Given this, I'd rather we don't add new API now when this landscape is changing. Instead wait and get guidance from Redis on what images should be supported going forward.

Alirexaa:
> Thanks for the contribution here @Alirexaa. However, at this point I don't think we should be doing this work. It looks like the functionality of Redis Stack is moving into the common layer.
> 
> https://redis.io/blog/the-future-of-redis/
> 
> > Simplifying the Redis Experience
> 
> > First, we’re focusing on simplicity and the developer experience – both in the core and in our developer experience. Starting with Redis 8, we’re integrating Redis Stack capabilities including query and search, vector search, JSON document support, time series and probabilistic structures into the Community Edition, providing a single, comprehensive Redis that offers all the latest features without the need for separate downloads or packages.
> 
> Given this, I'd rather we don't add new API now when this landscape is changing. Instead wait and get guidance from Redis on what images should be supported going forward.


Seems reasonable.

PR #5483, opened 2024-08-29
Upgrade tooling for 9.0x

### LINKED ISSUE #5446: Update Arcade and Tooling on the repo to use 9.0
In order to be able to test and target net9.0, our repository needs to be using the 9.0 version of Arcade as well as the 9.0 SDK. This issue is to track the work of doing that.

radical:
- Update dependencies from `.NET 9 Eng` channel
- Update nuget.config to use `dotnet9` nuget feed instead of `nuget8`
- Update tests to run with 8.0 runtime
- Resolve new xunit analyzer warnings

All the projects still target `net8.0`.

Future TODO:
- add 9.0 templates, and tests

Fixes https://github.com/dotnet/aspire/issues/5446 .

TODO before merge:
- [x] check that code coverage report is unaffected
- [x] Number of tests is the same as a `main` run
- [ ] Diff package build output to ensure contents of packages are unaffected (added by @joperezr)
- [x] Validate with an internal build

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5483)


src/Aspire.Dashboard/Otlp/Storage/ApplicationKey.cs
```diff
@@ -16,7 +16,7 @@ public int CompareTo(ApplicationKey other)
         return string.Compare(InstanceId, other.InstanceId, StringComparisons.ResourceName);
     }
 
-    public bool EqualsCompositeName(string name)
+    public bool EqualsCompositeName(string? name)
```
joperezr, line 19: This is technically a breaking change, although because it is in the Dashboard we don't really care. For other src projects with Nullability issues (if any) I think it'd be best to just baseline the warnings and/or turn off nullability than trying to fix it on this PR.

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -230,7 +230,7 @@ public async Task CrashTests(bool genericEntryPoint, string crashArg)
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(() =>
             genericEntryPoint ? DistributedApplicationTestingBuilder.CreateAsync<Projects.TestingAppHost1_AppHost>([$"--crash-{crashArg}"], cts.Token).WaitAsync(cts.Token)
                                 : DistributedApplicationTestingBuilder.CreateAsync(typeof(Projects.TestingAppHost1_AppHost), [$"--crash-{crashArg}"], cts.Token).WaitAsync(cts.Token))
-                .ConfigureAwait(false);
+;
```
joperezr, line 233: NIT: Fix formatting by removing new-line.

radical: I'll do a cleanup pass once all the tests are passing.

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
```diff
@@ -103,7 +103,7 @@ static async Task<List<string[]>> GetAndValidateCellTexts(ILocator tableLoc)
                     r => Assert.True(DateTime.TryParse(r, out _)),
                     r => Assert.True(int.TryParse(r, out var actualTempC) && actualTempC >= -20 && actualTempC <= 55),
                     r => Assert.True(int.TryParse(r, out var actualTempF) && actualTempF >= -5 && actualTempF <= 133),
-                    r => Assert.Contains(r, ["Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"]));
+                    r => Assert.Contains(r, new HashSet<string>(){"Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"}));
```
joperezr, line 106: Interesting, I would have expected this to work with the new collection literals.

tests/Shared/WorkloadTesting/data/nuget8.config
```diff
@@ -9,7 +9,7 @@
     <add key="built-local" value="%BUILT_NUGETS_PATH%" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
-    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
+    <add key="dotnet9" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json" />
```
joperezr, line 12: So we don't need dotnet8 feed anymore? I think we still push our packages there so we have to update our default channels in darc.

tests/Shared/Aspire.Workload.Testing.targets
```diff
@@ -19,6 +19,9 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Runtime.WorkloadTesting.Internal" />
+
+    <AdditionalSharedFrameworkToInstallArguments Include="-Runtime dotnet -Version 8.0.8 -Channel 8.0" />
```
joperezr, line 23: Is there a way to reuse the dotnet installed by the original build.cmd/sh call so we don't have to add this? I want to avoid needing to update a number in several places (runtime framework in this case 8.0.8) as those change each month.

tests/helix/send-to-helix-inner.proj
```diff
@@ -8,13 +8,15 @@
 
     <IncludeDotNetCli>true</IncludeDotNetCli>
     <DotNetCliPackageType>sdk</DotNetCliPackageType>
+    <DotNetCliVersion>8.0.401</DotNetCliVersion>
```
joperezr, line 11: How will this work for template tests for 9.0 templates? won't they require a 9.0 SDK?

tests/Shared/Aspire.Workload.Testing.targets
radical: This is specifically for workload testing. It will be much cleaner to use the existing support for this. But we could try to use the restored `.dotnet` also, but also need to account for `.dotnet` not being installed when the system `dotnet` is sufficient.

tests/helix/send-to-helix-inner.proj
radical: Still going through a first pass to get tests passing :D

Directory.Build.props
```diff
@@ -3,6 +3,7 @@
   <Import Project="$(RepositoryEngineeringDir)/apicompat/PublicApiAnalyzer.props" />
 
   <PropertyGroup>
+    <NetCurrent>net8.0</NetCurrent>
```
joperezr, line 6: Is this property used anywhere else in arcade or is it solely used by consuming repos? my concern is that changing this in order to avoid having to modify individual projects might unintentionally affect how the infrastructure work.

tests/helix/send-to-helix-inner.proj
radical: This will require us to bundle the sdk ourselves, because arcade doesn't seem to support additional runtimes.
To confirm that this works I think this PR should include a couple of tests where the test project targets net9.0 (best in workload tests).

Directory.Build.props
radical: Good point. I can drop this in the cleanup.

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
radical: IIRC it was not being able to resolve the overload between `HashSet<>` and `SortedSet<>`.

global.json
```diff
@@ -1,16 +1,30 @@
...
   },
   "tools": {
-    "dotnet": "8.0.303"
+    "dotnet": "9.0.100-preview.7.24407.12",
+    "runtimes": {
+      "dotnet/x64": [
+        "8.0.8"
+      ],
+      "dotnet/arm64": [
+        "8.0.8"
```
eerhardt, line 14: Do we have to have this duped by architecture? Can we do what efcore does?

https://github.com/dotnet/efcore/blob/ecfee78eb1fa2b2eaa0dbf945f1d4f8fa571be74/global.json#L9-L12

Directory.Build.props
eerhardt: Looks like efcore calls this `DefaultNetCoreTargetFramework`.

https://github.com/dotnet/efcore/blob/ecfee78eb1fa2b2eaa0dbf945f1d4f8fa571be74/src/EFCore.Abstractions/EFCore.Abstractions.csproj#L5C24-L5C53

tests/Shared/WorkloadTesting/data/nuget8.config
radical: But we shouldn't be fetching any of our packages from these feeds (except `BUILT_NUGETS_PATH`) for tests. Or am I missing something?

radical:
TODO:
- [ ] check that code coverage report is unaffected
- [ ] Number of tests is the same as a `main` run

eng/Version.Details.xml
```diff
@@ -114,56 +114,50 @@
...
-      <Sha>80264e60280e2815e7d65871081ccac04a32445c</Sha>
+      <Sha>dd332f2d4e21daa8b79f84251ab156af9a0b11b2</Sha>
       <SourceBuild RepoName="arcade" ManagedOnly="true" />
     </Dependency>
-    <!-- Intermediate is necessary for source build. -->
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.xliff-tasks" Version="1.0.0-beta.23475.1" CoherentParentDependency="Microsoft.DotNet.Arcade.Sdk">
-      <Uri>https://github.com/dotnet/xliff-tasks</Uri>
-      <Sha>73f0850939d96131c28cf6ea6ee5aacb4da0083a</Sha>
-      <SourceBuild RepoName="xliff-tasks" ManagedOnly="true" />
-    </Dependency>
```
radical, line 167: From the first commit - does dropping this seem correct? The assembly has moved to `dotnet/arcade`.

radical:
Read commit by commit to review this. The first commit is the biggest and is the update of the dependencies, mostly from arcade.

eng/Version.Details.xml
joperezr: I think that is fine, and I think you can also drop the dependency over from lines 152-155 above for that matter (not sure why we had both), but shouldn't we have now a new dependency coming from arcade for it?

eng/Versions.props
```diff
@@ -12,6 +12,12 @@
       This should get updated each time packages are shipped.
     -->
     <BaselineVersionForPackageValidation>8.0.1</BaselineVersionForPackageValidation>
+    <DefaultTargetFramework>net8.0</DefaultTargetFramework>
+    <!-- dotnet 8.0 versions for running tests -->
+    <DotNetRuntimeVersionForTesting>8.0.8</DotNetRuntimeVersionForTesting>
```
joperezr, line 17: I'm wondering how we can keep this one up-to-date. One option would be if we use the `Microsoft.NetCore.App.Ref` package and dependency flow for the version as we know that one will push the update to us. Anyway, no need to fix it now but we should consider how to update this (and the sdk below)

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
```diff
@@ -103,7 +103,7 @@ static async Task<List<string[]>> GetAndValidateCellTexts(ILocator tableLoc)
                     r => Assert.True(DateTime.TryParse(r, out _)),
                     r => Assert.True(int.TryParse(r, out var actualTempC) && actualTempC >= -20 && actualTempC <= 55),
                     r => Assert.True(int.TryParse(r, out var actualTempF) && actualTempF >= -5 && actualTempF <= 133),
-                    r => Assert.Contains(r, ["Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"]));
+                    r => Assert.Contains(r, new HashSet<string>{"Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"}));
```
joperezr, line 106: Still curious why this didn't work with the collection initializer, but oh well, no need to change it.

eng/Versions.props
radical: `dotnet/runtime` is using `Microsoft.DotNet.ApiCompat.Task` for that. Either should work well.

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
radical: `error CS0121: The call is ambiguous between the following methods or
properties: 'Assert.Contains<T>(T, HashSet<T>)' and
'Assert.Contains<T>(T, SortedSet<T>)'`

tests/helix/send-to-helix-inner.proj
```diff
@@ -13,8 +13,7 @@
     <!-- Disable playwright tests on helix/linux. Issue: https://github.com/dotnet/aspire/issues/4623 -->
     <DisablePlaywrightTests Condition="'$(OS)' != 'Windows_NT'">true</DisablePlaywrightTests>
 
-    <GlobalJsonContent>$([System.IO.File]::ReadAllText('$(RepoRoot)global.json'))</GlobalJsonContent>
-    <DotNetCliVersion>$([System.Text.RegularExpressions.Regex]::Match($(GlobalJsonContent), '(%3F&lt;="dotnet": ").*(%3F=")'))</DotNetCliVersion>
+    <DotNetCliVersion>$(DotNetSdkVersionForTesting)</DotNetCliVersion>
```
joperezr, line 16: We'll want to make sure now that we update that property so that it at least equals the one in global.json. I wonder if we can add a target somewhere in our build that detects if we've fallen behind and if so to fail the build...

No need to add that in this change though, don't block this PR for that.

global.json
```diff
@@ -1,16 +1,30 @@
 {
   "sdk": {
-    "version": "8.0.303",
+    "version": "9.0.100-preview.7.24407.12",
```
joperezr, line 3: This is fine for now. We'll want to bump to rc1 that got shipped today soon.

joperezr (approved):
Changes look great @radical thanks for working on this! Let's prioritize getting this merged given the breath as otherwise it is likely to get conflicts with other PRs.

eng/Version.Details.xml
radical: We'll need to add a dependency like `dotnet/runtime` has:
```xml
    <Dependency Name="Microsoft.DotNet.XliffTasks" Version="9.0.0-beta.24423.2">
      <Uri>https://github.com/dotnet/arcade</Uri>
      <Sha>91599268652b51969b8d8088d4f2f2ba7b3ebb19</Sha>
    </Dependency>
```

tests/helix/send-to-helix-inner.proj
radical: Once we get the version from a `dotnet/sdk` dependency, it should always(?) be newer than `global.json`.

eng/Version.Details.xml
radical: added that here.

radical:
Issue to track follow up work - https://github.com/dotnet/aspire/issues/5676

radical:
Internal build for validation - https://dev.azure.com/dnceng/internal/_build/results?buildId=2535499

PR #5506, opened 2024-08-31
Fix issues around metrics with unit rates

### LINKED ISSUE #5496: Rate metrics are not displayed properly

### Describe the bug

I opened a [discussion](https://github.com/dotnet/aspire/discussions/5495#discussion-7112311) about this but I think this might be a bug.

I have a instrument that is using the unit `By/s` which is later exported via OTLP to Aspire dashboard (standalone) but this is showing up in the dashboard as "Bytes Perseconds" which seems wrong.

image

This might be specific to OTLP so I tried to change the unit to `bytes_per_second` this shows up with a nicer label but it isn't really presenting as a throughput or rate metric, i.e. one over.

### Expected Behavior

I would expect this to be recognized as a rate metric of correct units. I would also expect the Aspire dashboard to present this data using MB/s or something suitable and not like now, where it's not even recognizing that these are bytes.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

JamesNK:
## Description

Unit rate display in metrics has some issues. This PR:

* Adds space between per and unit, e.g. `per second`
* Doesn't pluralize rate unit
* Adds tests

After:
image

Fixes https://github.com/dotnet/aspire/issues/5496

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5506)

### DIFF
--- a/playground/Stress/Stress.ApiService/Program.cs
+++ b/playground/Stress/Stress.ApiService/Program.cs
@@ -20,6 +20,11 @@ var app = builder.Build();
 
 app.Lifetime.ApplicationStarted.Register(ConsoleStresser.Stress);
 
+app.Lifetime.ApplicationStarted.Register(() =>
+{
+    _ = app.Services.GetRequiredService<TestMetrics>();
+});
+
 app.MapGet("/", () => "Hello world");
 
 app.MapGet("/write-console", () =>
--- a/playground/Stress/Stress.ApiService/TestMetrics.cs
+++ b/playground/Stress/Stress.ApiService/TestMetrics.cs
@@ -15,15 +15,42 @@ public class TestMetrics : IDisposable
 
     public TestMetrics()
     {
-        _meter = new Meter(MeterName, "1.0.0", new[]
-        {
+        _meter = new Meter(MeterName, "1.0.0",
+        [
             new KeyValuePair<string, object?>("meter-tag", Guid.NewGuid().ToString())
+        ]);
+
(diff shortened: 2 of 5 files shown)


leidegre:
Will this also address the Y axis, will bytes per second show up as, for example 500MB/s? If that's what's suitable for the range/domain.

JamesNK:
> Will this also address the Y axis, will bytes per second show up as, for example 500MB/s? If that's what's suitable for the range/domain.

Do you mean the tooltip? The unit in the tooltip is the same as the graph legend:

image

leidegre:
@JamesNK I was hoping for something like this

image

JamesNK:
We wouldn't do that because the unit can be long. If we changed the graph to have the unit on the axis it would be displayed vertically.

For example, https://plotly.com/javascript/line-charts/#graph-and-axes-titles

adamint (approved):

PR #5515, opened 2024-09-02
WaitFor integration with Health Check Service

mitchdenny:
## Description

This PR integrates `WaitFor` with health checks from the .NET hosting model. Extension methods like `AddRedis(...)` register a health check in DI which uses the connection string which is produced when the resource is provisioned.

A new annotation `HealthCheckAnnotation` keeps track of the health check names associated with each resource and a health check publisher implementation `ResourceNotificationHealthCheckPublisher` transcribes health check status from the health check report to a new property on the ResourceEvent (`HealthStatus`).

In `WaitFor` we determine whether a resources's dependencies has a `HealthCheckAnnotation` and if it does it waits for the health status to return healthy.

This PR also introduces a new event .... `ConnectionStringAvailableEvent` which is used by `AddRedis(...)` and `AddPostgres(...)` to detect when the connection string is available so we can store it in a captured string which is used by the health check. This is significant because it allows Azure hosted resources to participate in the health-check enabled `WaitFor(...)` experience. 

In the case of DCP hosted resource sthe `ConnectionStringAvailableEvent` is fired at the same time as `BeforeResourceStartedEvent` - but in the case of the AzureProvisioner it is fired after the resource is online in Azure (so the order is roughly opposite) - hence why we needed another event (Azure hosted resources don't get endpoint allocated events).

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [x] No
  - [ ] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5515)


src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -38,6 +39,20 @@ public static IResourceBuilder<PostgresServerResource> AddPostgres(this IDistrib
...
+            var connectionString = await postgresServer.GetConnectionStringAsync(ct).ConfigureAwait(false);
+            builder.Configuration[$"{postgresServer.Name}_builtin_connectionstring"] = connectionString;
+        });
+
+        builder.Services.AddHealthChecks().AddNpgSql(sp =>
+            {
+                return builder.Configuration[$"{name}_builtin_connectionstring"]!;
+            },
+            name: $"{name}_check"
+        );
```
davidfowl, line 54: What does using configuration do here over using a variable in the closure?

src/Aspire.Hosting/Health/ResourceHealthService.cs
```diff
@@ -0,0 +1,33 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting.Health;
+
+internal class ResourceHealthService(ILogger<ResourceHealthService> logger, HealthCheckService healthService) : IResourceHealthService
```
davidfowl, line 10: We should be pushing heath status not pulling it. This should look at all resources and publish health updates via RNS

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -607,6 +609,9 @@ public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, I
                     $"Resource '{dependency.Resource.Name}' has entered the '{snapshot.State.Text}' state prematurely."
                     );
             }
+
+            // If we get to here the resource is running so we just need to check on whether it is healthy or not.
+            await rhs.WaitUntilResourceHealthyAsync(dependency.Resource, ct).ConfigureAwait(false);
```
davidfowl, line 614: This shouldn't exist, instead we should tweak the predicate to include the health state of the resource.

src/Aspire.Hosting/Health/ResourceNotificationHealthCheckPublisher.cs
```diff
@@ -0,0 +1,37 @@
...
+                }).ConfigureAwait(false);
+            }
+            else
+            {
+                await resourceNotificationService.PublishUpdateAsync(resource, s => s with
+                {
+                    HealthStatus = HealthStatus.Healthy
+                }).ConfigureAwait(false);
+            }
+        }
```
davidfowl, line 35: I don't think we want this.

```suggestion

```

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: Yeah I cut over to capturing the variable from the enclosing scope.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
mitchdenny: Yeah I removed it.

src/Aspire.Hosting/Health/ResourceNotificationHealthCheckPublisher.cs
mitchdenny: Agreed. I removed this and I'll let the default state be null, and then actively decide whether to wait for healthy state based on the presence of health annotations.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -94,6 +94,14 @@ async Task UpdateStateAsync(IAzureResource resource, Func<CustomResourceSnapshot
             {
                 await notificationService.PublishUpdateAsync(child, stateFactory).ConfigureAwait(false);
             }
+
+            // Some IAzureResource instances are a surrogate for for another resource in the app model
+            // to ensure that resource events are published for the resource that the user expects
+            // we lookup the resource in the app model here and publish the update to it as well.
+            if (appModel.Resources.Where(r => r.Name == resource.Name).SingleOrDefault() is { } nonAzureResource)
+            {
+                await notificationService.PublishUpdateAsync(nonAzureResource, stateFactory).ConfigureAwait(false);
```
mitchdenny, line 103: Need to keep track of the resource and its surrogate from the beginning so we don't need to do this search each time we fire the event.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -38,6 +39,25 @@ public static IResourceBuilder<PostgresServerResource> AddPostgres(this IDistrib
...
+        string? connectionString = null;
+
+        builder.Eventing.Subscribe<ConnectionStringAvailableEvent>(postgresServer, async (@event, ct) =>
+        {
+            connectionString = await postgresServer.GetConnectionStringAsync(ct).ConfigureAwait(false);
+        });
+
+        builder.Services.AddHealthChecks().AddNpgSql(sp => connectionString!, name: $"{name}_check", configure: (connection) =>
+        {
+            // HACK: The Npgsql client defaults to using the username in the connection string if the database is not specified. Here
```
mitchdenny, line 52: Need to figure out if/what we can do better here.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -48,7 +68,8 @@ public static IResourceBuilder<PostgresServerResource> AddPostgres(this IDistrib
                       {
                           context.EnvironmentVariables[UserEnvVarName] = postgresServer.UserNameReference;
                           context.EnvironmentVariables[PasswordEnvVarName] = postgresServer.PasswordParameter;
-                      });
+                      })
+                      .WithAnnotation(new HealthCheckAnnotation($"{postgresServer.Name}_check"));
```
mitchdenny, line 72: Naming of healthchecks coudl be better here. It might actually be better to have a WithHealthcheck method (perhaps internal initially) which takes a callback to configure the health check.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -29,10 +30,21 @@ public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicat
         ArgumentNullException.ThrowIfNull(builder);
 
         var redis = new RedisResource(name);
+
+        string? connectionString = null;
+
+        builder.Eventing.Subscribe<ConnectionStringAvailableEvent>(redis, async (@event, ct) =>
+        {
+            connectionString = await redis.GetConnectionStringAsync(ct).ConfigureAwait(false);
```
mitchdenny, line 38: Need to figure out if there is a way to suppress log noise for the health check until its actually useful (or potentially redirect the log check logger noise).

src/Aspire.Hosting/ApplicationModel/HealthCheckAnnotation.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// An annotation which tracks the name of the health check used to detect to health of a resource.
+/// </summary>
+/// <param name="key">TODO     </param>
+public class HealthCheckAnnotation(string key) : IResourceAnnotation
+{
+    /// <summary>
+    /// TODO
```
mitchdenny, line 13: Fix this up.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -301,6 +312,14 @@ await provisioner.GetOrCreateResourceAsync(
                 resource.ProvisioningTaskCompletionSource?.TrySetException(new InvalidOperationException($"Unable to resolve references from {resource.Name}"));
             }
         }
+
+        async Task PublishConnectionStringAvailableEventAsync()
+        {
+            var resolvedResource = appModel.Resources.Single(r => r.Name == resource.Name);
+
+            var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(resolvedResource, serviceProvider);
+            await eventing.PublishAsync(connectionStringAvailableEvent, cancellationToken).ConfigureAwait(false);
```
davidfowl, line 321: Same here, we shouldn't be searching for anything.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: Decided we can't really do better here (or rather there is no benefit). At the end of the day you still need to compute the name of the health check, so this may as well be it.

mitchdenny:
This is functionally complete for Redis and Postgres (Postgres being an example of a resource with child resources). Moving on to working on tests and polish.

My plan is to merge this and then do other resources as subsequent commits to keep them relatively small and focused.

playground/TestShop/TestShop.AppHost/Program.cs
```diff
@@ -1,9 +1,10 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
-var catalogDb = builder.AddPostgres("postgres")
-                       .WithDataVolume()
-                       .WithPgAdmin()
-                       .AddDatabase("catalogdb");
+var postgres = builder.AddPostgres("postgres")
```
davidfowl, line 3: Why this change?

playground/waitfor/WaitForSandbox.AppHost/Program.cs
```diff
@@ -0,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var pg = builder.AddPostgres("pg")
+                .AsAzurePostgresFlexibleServer()
```
davidfowl, line 7: Are you planning to commit like this?

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
davidfowl: I feel like we're missing a primitive here. Should this instead be health checking the database instead of the server? Shouldn't we also add health checks for each database resource?

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -48,7 +69,8 @@ public static IResourceBuilder<PostgresServerResource> AddPostgres(this IDistrib
                       {
                           context.EnvironmentVariables[UserEnvVarName] = postgresServer.UserNameReference;
                           context.EnvironmentVariables[PasswordEnvVarName] = postgresServer.PasswordParameter;
-                      });
+                      })
+                      .WithAnnotation(new HealthCheckAnnotation(healthCheckName));
```
davidfowl, line 73: We should have a WithHealthCheck method right?

davidfowl: We should also be able to add multiple health checks. The 1:1 mapping between the health check name and the resource name seems like something we need to fix.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
davidfowl: What does the log noise look like? This is a downside (and upside?) of using the same DI container for these health checks. We have to figure out where health check logs should show up in the dashboard.

src/Aspire.Hosting/ApplicationModel/HealthCheckAnnotation.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// An annotation which tracks the name of the health check used to detect to health of a resource.
+/// </summary>
+/// <param name="key">The key for the health check in the app host which associated with this resource.</param>
+public class HealthCheckAnnotation(string key) : IResourceAnnotation
```
davidfowl, line 10: Is it called key or name?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1164,6 +1164,9 @@ await notificationService.PublishUpdateAsync(cr.ModelResource, s => s with
 
     private async Task CreateExecutableAsync(AppResource er, ILogger resourceLogger, CancellationToken cancellationToken)
```
davidfowl, line 1165: these events should only fire for IResourceWithConnectionString

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1164,6 +1164,9 @@ await notificationService.PublishUpdateAsync(cr.ModelResource, s => s with
...
     {
+        var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(er.ModelResource, serviceProvider);
```
davidfowl, line 1167: these events should only fire for IResourceWithConnectionString

src/Aspire.Hosting/Health/ResourceNotificationHealthCheckPublisher.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Aspire.Hosting.Health;
+
+internal class ResourceNotificationHealthCheckPublisher(DistributedApplicationModel model, ResourceNotificationService resourceNotificationService) : IHealthCheckPublisher
+{
+    public async Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
+    {
+        _ = model;
+        _ = resourceNotificationService;
```
davidfowl, line 14: delete
```suggestion
```

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -607,6 +608,14 @@ public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, I
                     $"Resource '{dependency.Resource.Name}' has entered the '{snapshot.State.Text}' state prematurely."
                     );
             }
+
+            // If our dependency resource has health check annotations we want to wait until they turn healthy
+            // otherwise we don't care about their health status.
+            if (dependency.Resource.TryGetAnnotationsOfType<HealthCheckAnnotation>(out var _))
```
davidfowl, line 614: This should be in the initial call, not a call after.  This does not handle terminal states.

davidfowl:
This needs tests

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
mitchdenny: This is what it looks like right now:

image

(not in the dashboard).

The only way I can think to reduce this noise is to suppress the default health service logging completely and then have the publisher write errors into the per-resource error logs.

Thoughts?

playground/TestShop/TestShop.AppHost/Program.cs
mitchdenny: cleaned up. this was left over from my initial explorations when I was using TestShop as a playground.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
mitchdenny: I'll leave it there but commented out.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: I was thinking about adding `WithHealthCheck(...)`. I actually wanted to make it something like this:

```csharp
builder.WithHealthCheck(healthCheckBuilder => {
   healthCheckBuilder.AddBlahCheck(...);
});
```

... and avoid having to set a stringly name altogether (or at least make it auto-generated). But the health checks API surface isn't conducive to that. I'll add a `WithHealthChecks(...)` just so we have the API but its not doing that much for us.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: There is a corner case here that we'll need to explore around project to project references where we might have a health check in the apphost which checks for the liveness of the projects endpoints.

What I am thinking is that we do something like this:

```csharp
var api = builder.AddProject<Projects.BackendApi>("backendapi")
                 .WithEndpoint("https", (ep) => ep.HealthCheck = true);
var frontend = builder.AddProject<Projects.Frontend>("frontend")
                      .WithReference(api).WaitFor(api);
```

When you enable health checks on an endpoint we would automatically inject a HTTP health check for /healthz (or something else) and the frontend project in this case wouldn't start until the backend API was healthy.

mitchdenny: ... I don't see this as critical right now but its what I've been thinking about to approach this if it comes up as a future requirement.

src/Aspire.Hosting/Health/ResourceNotificationHealthCheckPublisher.cs
mitchdenny: I didn't want these two variables getting too big for their boots so I decided to slap them down up front in this class .... fixed.

mitchdenny:
> This needs tests

Yes it does! I didn't think you'd like me #YOLO merge this one.

src/Aspire.Hosting/ApplicationModel/HealthCheckAnnotation.cs
mitchdenny: Key ... inline with what is in the health checks API. I've updated the variable name.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: Are we bubbling the state down to the child resource by default? If I wait for the db what happens right now?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -242,65 +260,71 @@ private async Task ProcessResourceAsync(IConfiguration configuration, Lazy<Task<
...
-
-                resource.ProvisioningTaskCompletionSource?.TrySetException(new InvalidOperationException($"Unable to resolve references from {resource.Name}"));
+                resourceLogger.LogError(ex, "Error provisioning {ResourceName}.", resource.AzureResource.Name);
+                resource.AzureResource.ProvisioningTaskCompletionSource?.TrySetException(new InvalidOperationException($"Unable to resolve references from {resource.AzureResource.Name}"));
             }
         }
+
+        async Task PublishConnectionStringAvailableEventAsync()
+        {
+            var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(resource.Resource, serviceProvider);
```
davidfowl, line 325: Do we need to publish for child resources?

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
davidfowl: I think we do need to do that.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
davidfowl: > The only way I can think to reduce this noise is to suppress the default health service logging completely and then have the publisher write errors into the per-resource error logs.

Yes! But pushing logs through the resource will cause it to clobber the logs come from the resource.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
mitchdenny: I now remember why I did it this way :)

One of the quirks of the resource notification service is that the `Snapshot.Status` conflates resource visibility in the dashboard with status. So when the publisher calls `PublishAsync` for the resource state its going to a resource which is in a hidden state - so simultaneously looking for a running resource and a healthy resource is only possible if in the publisher we look up the replica IDs and publish directly to them.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: This comment might have been in relation to another comment. Regardless ...

Right now the child resource picks up the state of the parent (e.g. it'll move into the Running state when the parent does). But it doesn't have any health checks that might block something that depends on the database from launching.

If we do add a health check to the database resource we'll need to fire before resource started events for child resources artificially. Just need to try it out and find the potential places where it could result in a hang.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
davidfowl: This is because of how WaitForResource is implemented right? Maybe skip hidden resources as part of those checks?

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: The initializer? Lets queue it up for after WaitFor is in. It intersects with it but it should be  a discrete mechanism like health check integration.

davidfowl: No, if I wait for db then it should wait for the parent resource (the db server)'s health check to be 👍🏾

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
mitchdenny: So what I've done here is I've made it so that when the server gets its connection string event fired, it then fires its child resource connection string events. This is a decision we would need to make on a per resource basis but I think it should hold for most resource types.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
mitchdenny: Got child resource waits working.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
mitchdenny: Got child resource waits working.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
mitchdenny: We could push a copy of the error through to each resource which is depending on it, that way the error details are prepended to a resource which isn't running yet. Also its probably the place someone is going to look when they see that they have a resource in a waiting state.

mitchdenny: Will leave this as is until we've done the dashboard work so we can at least debug why something isn't working. Once we've piped the health checks through into the dashboard as part of https://github.com/dotnet/aspire/issues/5569 then we can suppress this.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -29,10 +30,22 @@ public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicat
...
+        });
+
+        var healthCheckKey = $"{name}_check";
+        builder.Services.AddHealthChecks().AddRedis(sp => connectionString!, name: healthCheckKey);
```
davidfowl, line 42:
```suggestion
        builder.Services.AddHealthChecks().AddRedis(sp => connectionString ?? throw new InvalidOperationException("Connection string is unavailable"), name: healthCheckKey);
```

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: OK so what I've done is I've left the health check in for the database. My thinking is by doing this we give folks flexibility. If folks truly want to wait for the database to be healthy they can do that with no effort.

I think what people will likely do is something like we have in the waitfor playground where they have some kind of worker that waits for the server to be ready and then initializes the database, and other services that depend on the database will WaitFor that.

davidfowl (approved):

Issue #295, opened 2023-10-15
Add support for restarting services from the dashboard


davidfowl:
In building apps with the aspire, I'm often using the dashboard as a debugging aid, when a process dies it pretty disruptive to have to restart the entire orchestration project in order to reproduce an issue. It would be great to have the ability to restart a process from the dashboard.

HClausing:
> In building apps with the aspire, I'm often using the dashboard as a debugging aid, when a process dies it pretty disruptive to have to restart the entire orchestration project in order to reproduce an issue. It would be great to have the ability to restart a process from the dashboard.

Just an idea for the future: On some scenarios this restart action can require an authorized user, and optionally with roles or policies requirements.

aL3891:
Was looking to add a request for this as well, both in the case where a service has died (obv not due to an error on my part) and when ive made a change in the running app and need to restart it. 
i guess optionally it could also monitor the source for a service and restart it? (assuming hot reload doesnt just work as is)

davidfowl:
Using dotnet watch handles the latter case on command line and VS does the same when you use ctrl+f5. You can also hot reload when you hit F5

aL3891:
So dotnet watch is transitive? that's nice! i tried it now though and got errors like this in the logs

```

dotnet watch 🚀 Started
C:\Program Files\dotnet\sdk\8.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.GenerateAssemblyInfo.targets(199,5): error MSB3713: The file "obj\Debug\net8.0\LoginserverDb.AssemblyInfo.cs" could not be created. The process cannot access the file 'C:\Funrock\mma_backend2\LoginserverDb\obj\Debug\net8.0\LoginserverDb.AssemblyInfo.cs' because it is being used by another process. [C:\Funrock\mma_backend2\LoginserverDb\LoginserverDb.csproj]

and 


CSC : error CS2012: Cannot open 'C:\Funrock\mma_backend2\MMA.ServiceDefaults\obj\Debug\net8.0\MMA.ServiceDefaults.dll' for writing -- 'The process cannot access the file 'C:\Funrock\mma_backend2\MMA.ServiceDefaults\obj\Debug\net8.0\MMA.ServiceDefaults.dll' because it is being used by another process.' [C:\Funrock\mma_backend2\MMA.ServiceDefaults\MMA.ServiceDefaults.csproj]
[...]

PR #5516, opened 2024-09-02
Commands glow up

JamesNK:
## Description

This PR:

* Renames commands column to actions
* Adds logs and view details to actions menu
* Removes logs and view details columns
* Adds icons to commands
* Adds highlighted commands

Addresses https://github.com/dotnet/aspire/issues/295

Demo:

resource-actions-menu2

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5516)


src/Aspire.Hosting/Dashboard/DashboardService.cs
```diff
@@ -144,4 +142,10 @@ async Task WatchResourceConsoleLogsInternal()
             }
         }
     }
+
+    public override Task<ResourceCommandResponse> ExecuteResourceCommand(ResourceCommandRequest request, ServerCallContext context)
+    {
+        //request.ResourceName
```
davidfowl, line 148:
```suggestion
```

leslierichardson95:
These changes look great!  Suggestion: can we consider changing the tooltip text from just "stop" or "play" to "stop resource" and "play resource" just for some added clarity?

JamesNK:
> These changes look great! Suggestion: can we consider changing the tooltip text from just "stop" or "play" to "stop resource" and "play resource" just for some added clarity?

The buttons here are examples. I'll keep this mind for when they're added in the future.

timheuer:
we're elevating stop/start as the PRIMARY action here, hiding others behind a ... -- are these truly the primary actions we think folks want on a resource? Details has been my primary (viewing config values) with logs being second. Yes because the others aren't there yet, but feels like stop/start/restart aren't going to be primary actions.  Reserve the default space for primary and/or have fallaway logic (show more and responsively hide based on priorirty) /cc @leslierichardson95

DamianEdwards:
> Details has been my primary (viewing config values) with logs being second. Yes because the others aren't there yet, but feels like stop/start/restart aren't going to be primary actions.

Personally I'd like to see Stop/Start and Details (aka info) be the primary actions, but given the new UX it seems we could easily fit all three in icon form (Stop/Start, Details, Logs).

JamesNK:
"View details" doesn't need to be primary. 8.2 introduces clicking the row to show/hide details. That was introduced in 8.2. The main reason that "View details" still exists is for accessibility.

There may be UI space when viewing the resource page on a high-res display, but keep in mind that this column will be displayed in mobile view, and we don't want the `...` button to be hidden.

e.g.
image

Promoting and demoting buttons isn't difficult. I want to stick with the current setup, and we can change based on feedback.

leslierichardson95:
> we're elevating stop/start as the PRIMARY action here, hiding others behind a ... -- are these truly the primary actions we think folks want on a resource? Details has been my primary (viewing config values) with logs being second. Yes because the others aren't there yet, but feels like stop/start/restart aren't going to be primary actions. Reserve the default space for primary and/or have fallaway logic (show more and responsively hide based on priorirty) /cc @leslierichardson95

That's a good point.  For now, it might make more sense to stay consistent with the ACA stop/start resource experience where those actions take less priority over logs / details and are part of the "...".

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -38,6 +38,10 @@ message ResourceCommand {
     // Clients must return any value provided by the server when invoking
     // the command.
     optional google.protobuf.Value parameter = 4;
+    // A flag that indicates whether the command is highlighted in the UI.
+    bool is_highlighted = 5;
```
drewnoakes, line 42: I think this should be optional, so we don't break the contract. It can safely default to false.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -38,6 +38,10 @@ message ResourceCommand {
...
+    // Optional icon content. SVG path is displayed in the UI.
+    optional string icon_content = 6;
```
drewnoakes, line 44: An example of icon content here is:

```xml
<path d="M10.5 8.91a.5.5 0 0 0-1 .09v4.6a.5.5 0 0 0 1-.1V8.91Zm.3-2.16a.75.75 0 1 0-1.5 0 .75.75 0 0 0 1.5 0ZM18 10a8 8 0 1 0-16 0 8 8 0 0 0 16 0ZM3 10a7 7 0 1 1 14 0 7 7 0 0 1-14 0Z"/>
```

That is, it's an SVG path without any size or bounds information. There's an implicit connection between that string and the `IconSize` value used here.

The consuming code adds that dimension information via `IconSize.Size20`.

If we use SVG to allow icons to be defined in this way, we should make sure we at least document this expectation.

drewnoakes, line 44: Defining the icon as SVG means we cannot easily show icons in all clients.

For example, if we want to add a tool window to VS to support interaction with an AppHost (essentially, creating a new kind of client for the resource service) then I don't think we'll be able to display these icons. All solutions I've seen for this involve bringing in other heavy libraries (like SkiaSharp or Xamarin.Forms) which we won't be able to do.

Can we use a richer type here that provides more opportunity for clients to understand the semantics of the icon and provide alternative images?

If we used the name of the FluentUI icon's class, we could instantiate it using reflection to get SVG content. That would have the benefit of modernising automatically when the design language changes, which happens every few years. We don't want the UI to look dated.

Passing an icon name also avoids my sibling comment about implicit size information.

drewnoakes (commented):
Main question here is around the use of SVG in the protocol. See comments below. I'm facing the same issue in extensibility, for top-level navigation items provided by extensions.

Also, we now use the terms "command" and "actions" interchangeably. Can we consolidate? `Command` is already used in the `.proto`, so is hard to change (I think). I like "actions" for column header. Having two terms for the same thing is confusing though.

src/Aspire.Dashboard/Components/Controls/ResourceActions.razor
```diff
@@ -0,0 +1,23 @@
+﻿@namespace Aspire.Dashboard.Components
+@using Aspire.Dashboard.Model
+@using Microsoft.FluentUI.AspNetCore.Components
+
+@foreach (var highlightedCommand in Commands.Where(c => c.IsHighlighted))
+{
+    <FluentButton Appearance="Appearance.Lightweight" Title="@highlightedCommand.DisplayName" OnClick="@(() => CommandSelected.InvokeAsync(highlightedCommand))">
+        @if (!string.IsNullOrEmpty(highlightedCommand.IconContent))
+        {
+            <FluentIcon Value="@(new Icon(highlightedCommand.CommandType, IconVariant.Regular, IconSize.Size20, highlightedCommand.IconContent))" />
```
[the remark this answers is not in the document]
JamesNK: I don't think it's used anywhere.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
JamesNK: In proto, all fields are optional. If it isn't specify then it defaults to the default value (false).

I believe the only impact `optional` keyword has is you can check whether a value was explicitly set. We don't care about that here.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
JamesNK: We could have the fluentui icon name, but those other clients couldn't use it either. All they could do is look at the icon name and substitute in their own icon.

However, if they want to add icons then nothing is stopping them from ignoring this icon content and basing the icon on the command type.

e.g. command type = "Start", UI displays its play icon. command type = "Stop", UI displays its stop icon.

src/Aspire.Dashboard/Components/Controls/ResourceActions.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+    public EventCallback<CommandViewModel> CommandSelected { get; set; }
+
+    [Parameter]
+    public EventCallback<string> OnViewDetails { get; set; }
+
+    [Parameter]
+    public EventCallback OnConsoleLogs { get; set; }
+
+    [Parameter]
+    public string? MenuButtonId { get; set; }
```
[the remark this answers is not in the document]
JamesNK: It's bound to the UI and passed to view details invocation.

davidfowl:
I agree I don’t like usage of svg in the protocol. While more flexible it feels really bad coupling to the html that way. Can we do what we do for status icons or maybe a little more closed down and allowing names from an icon set

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
davidfowl: I prefer the icon name to sending svg directlY. I think orchard does the same thing

JamesNK:
I can make it a name from the FluentUI icons, but if someone wants a custom icon, e.g. an Azure product icon, then they're stuck.

davidfowl:
That’s ok for now

JamesNK:
> Also, we now use the terms "command" and "actions" interchangeably. Can we consolidate? `Command` is already used in the `.proto`, so is hard to change (I think). I like "actions" for column header. Having two terms for the same thing is confusing though.

It's too late to change the proto. But I don't think it will be confusing. Here is how I think of it:

* Grids have action menus. Right now, it is just the resources page, but I think we'll have an actions column on structured logs and traces eventually.
* The resources action menu also contains actions to execute resource commands.

IMO having a different name to disambiguate them is good even.

JamesNK:
All feedback applied.

davidfowl (approved):

PR #5526, opened 2024-09-03
Support displaying different resource properties

### LINKED ISSUE #5126: aspire-dashboard Structured logs -> Log entry details -> Resource attributes not always correct/seem cached

### Describe the bug

Hi, 

when exporting otel logging data from a .NET app to both Seq and Aspire Dashboard (8.0.2 and 8.1.0 show same behavior), aspire dashboard doesn't seem to render attribute values under Structured logs -> Log entry details -> Resource -> attributes correctly at all times (sometimes it seems parts of the attributes get cached). 

First of all, displayed values for a specific log entry differ between seq (correct) and aspire dashboard (not correct).

1.) Run the app, attribute foo=value1
2.) Seq and Aspire dashboard display the entry with foo=value1
3.) stop app, change value to foo=newValue
4.) re run app
5.) Seq=>foo=newValue, aspire dashboard (no restart of container) foo=value1

Yesterday, with 8.0.2, I also had issues that some attribute pairs did not show up at all at times in aspire dashboard (not a rendering issue).

 

### Expected Behavior

_No response_

### Steps To Reproduce

- asp.net 8 app
- builder.Services.AddLogging(logging => logging.AddOpenTelemetry(openTelemetryLoggerOptions => {
            openTelemetryLoggerOptions.SetResourceBuilder(
[...]
                    .AddAttributes(new Dictionary<string, object> {
[...]

### Exceptions (if any)

_No response_

### .NET Version info

SDK 8.0.303
Runtime: 8.0.7
AOT publish

### Anything else?

Affects aspire-dashboard 8.1.0 and 8.0.2

JamesNK:
## Description

Telemetry is sent with a resource and a resource has attributes. The first time the dashboard sees a resource it caches it using the name and instance id as the key.

Usually, the attributes on a resource are static. However, it is possible for telemetry to be sent with a resource with the same key but different attributes. When the telemetry is viewed in the UI we want those resource attributes to be displayed, not the initial cached attributes.

This PR:

* Adds OtlpApplicationView. It has a collection of properties
* Caches OtlpApplicationView on OtlpApplication. The key is the properties. This allows views to be shared by telemetry.
* Changes telemetry to refer to the app view instead of app, allowing for the correct properties to be displayed in the UI.
* Tests

Fixes https://github.com/dotnet/aspire/issues/5126

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5526)

### DIFF
--- a/src/Aspire.Dashboard/Otlp/Model/OtlpApplicationView.cs
+++ b/src/Aspire.Dashboard/Otlp/Model/OtlpApplicationView.cs
@@ -1,0 +1,66 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+using Aspire.Dashboard.Otlp.Storage;
+using Google.Protobuf.Collections;
+using OpenTelemetry.Proto.Common.V1;
+
+namespace Aspire.Dashboard.Otlp.Model;
+
+[DebuggerDisplay("Application = {Application}, Properties = {Properties.Count}")]
+public class OtlpApplicationView
+{
+    public ApplicationKey ApplicationKey => Application.ApplicationKey;
+    public OtlpApplication Application { get; }
+    public KeyValuePair<string, string>[] Properties { get; }
+
+    public OtlpApplicationView(OtlpApplication application, RepeatedField<KeyValue> attributes)
+    {
+        Application = application;
+
+        List<KeyValuePair<string, string>>? properties = null;
+        foreach (var attribute in attributes)
+        {
+            switch (attribute.Key)
+            {
+                case OtlpApplication.SERVICE_NAME:
+                case OtlpApplication.SERVICE_INSTANCE_ID:
+                    // Values passed in via ctor and set to members. Don't add to properties collection.
+                    break;
+                default:
+                    properties ??= [];
+                    properties.Add(new KeyValuePair<string, string>(attribute.Key, attribute.Value.GetString()));
+                    break;
+
+            }
+        }
+
+        if (properties != null)
+        {
+            // Sort so keys are in a consistent order for equality check.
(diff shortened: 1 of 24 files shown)


JamesNK:
cc @samsp-msft

src/Aspire.Dashboard/Otlp/Model/OtlpApplicationView.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public ApplicationKey ApplicationKey => Application.ApplicationKey;
+    public OtlpApplication Application { get; }
+    public KeyValuePair<string, string>[] Properties { get; }
+
+    public OtlpApplicationView(OtlpApplication application, RepeatedField<KeyValue> attributes)
+    {
+        Application = application;
+
+        var properties = new List<KeyValuePair<string, string>>();
```
drewnoakes, line 22: How often would an instance have no properties? If that weren't rare, we could save this allocation in such cases.

drewnoakes (approved):

src/Aspire.Dashboard/Otlp/Model/OtlpApplicationView.cs
JamesNK: There will always be properties, but it might just be the two known properties. I added optimization for that case.

Issue #4429, opened 2024-06-08
AddParameter overloads that supports providing a value 


davidfowl:
## Background and Motivation

There are scenarios 2 scenarios that aren't supported well:
1. Adding a parameter with a default value only used at runtime that doesn't come from configuration, or uses a different configuration schema that we expect by default (not `Parameters:name`).
2. Adding a parameter that uses a default value at publish time (https://github.com/dotnet/aspire/pull/4102/files#r1594886611)

## Proposed API

```diff
namespace Aspire.Hosting;

public static class ParameterResourceBuilderExtensions
{
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistribuedApplicationBuilder builder, string name, string value, bool secret = false);
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistribuedApplicationBuilder builder, string name, Func<string> valueGetter, bool secret = false);
+    public static IResourceBuilder<ParameterResource> AddParameterFromConfiguration(this IDistribuedApplicationBuilder builder, string name, string configurationKey, bool secret = false);
+    public static IResourceBuilder<ParameterResource> PublishDefaultValue(this IResourceBuilder<ParameterResource> parameter);
}
```

## Usage Examples

```csharp
var builder = DistributedApplication.CreateBuilder(args);

if (builder.ExecutionContext.IsRunMode)
{
    // Load key vault secrets, read the keyvault connection information from configuration
    builder.Configuration.AddAzureKeyVaultSecrets("sharedKv");
}

var accessToken = builder.AddParameterFromConfiguration("accessToken", "Auth:AccessToken", secret: true);

builder.AddProject<Projects.WebApplication1>("api")
       .WithEnvironment("AccessToken", accessToken)
       .WithExternalHttpEndpoints();

builder.Build().Run();
```

Default user name

```C#
var username = builder.AddParameter("pg-username", "admin").PublishDefaultValue();

builder.AddPostgres("pg", username: username);
```

## Manifest changes

`PublishDefaultValue` requires a manifest schema change:

```diff
{
    "pg-username": {
      "type": "parameter.v0",
      "value": "{pg-username.inputs.value}",
      "inputs": {
        "value": {
          "type": "string",
+          "default": {
+              "value": "admin"
+          }
        }
      }
    }
]
```

## Risks

Low

eerhardt:
@vhvb1989 @ellismg - does `azd` support default values in the manifest as above? Or do we need to log an issue in azd to get this support?

[...]

PR #5529, opened 2024-09-03
Add AddParameter overloads that take a constant and a ParameterDefault

### LINKED ISSUE #5410: Add API to support configuring parameter default values
## Background and Motivation

Adding parameters to the application model via `builder.AddParameter("name")` is straightforward enough, but sometimes it's desirable to create parameters that have a default value generated via the `GenerateParameterDefault` type, or use a hard-coded default value string, or custom `ParameterDefault` implementation. This is difficult to do today as `ParameterResource` default behaviors are largely controlled by the body of the `builder.AddParameter()` method itself and that method accepts no overloads for controlling parameter default values.

There should be API that makes it simple to set a parameter's default value in the same way that the implicit password parameters for many resources are.

Related:
- #4429
- #3679

### Suggestion 1: Overloads of `AddParameter` that accept `string,` `GenerateParameterDefault`, and `ParameterDefault`.

```diff
namespace Aspire.Hosting;

public static class ParameterResourceBuilderExtensions
{
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, string defaultValue, bool secret = false)
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, GenerateParameterDefault defaultValue, bool secret = true)
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, ParameterDefault defaultValue, bool secret = false)
}
```

The second overload above (the one that accepts `GenerateParameterDefault` would create the `ParameterResource` internally using the `CreateGeneratedParameter` method which ensures the generated value is stored in user secrets automatically.

### Suggestion 2: Extension method to `IResourceBuilder<ParameterResource>` that allows setting the parameter default value

An alternative approach is to add new extension methods that enable setting the default value after the parameter is created and added to the application model with `AddParameter`:

```diff
public static class ParameterResourceBuilderExtensions
{
+    public static IResourceBuilder<ParameterResource> WithDefault(this IResourceBuilder<ParameterResource> builder, string defaultValue)
+    public static IResourceBuilder<ParameterResource> WithDefault(this IResourceBuilder<ParameterResource> builder, GenerateParameterDefault defaultValue)
+    public static IResourceBuilder<ParameterResource> WithDefault(this IResourceBuilder<ParameterResource> builder, ParameterDefault defaultValue)
}
```

The second overload above (the one that accepts `GenerateParameterDefault` would wrap it in an instance of the `UserSecretsParameterDefault` internal class which ensures the generated value is stored in user secrets automatically.

Note that this approach would likely require some reworking of how `AddParameter` is currently implemented as it currently defaults the parameter resource initial state to an error if a value isn't found in configuration.

### Suggestion 2: `AddParameter` overloads
```csharp
var keycloakRealmName = builder.AddParameter("keycloak-realm", "SampleRealm");
var keycloakRealmDisplayName = builder.AddParameter("keycloak-realm-display", "Sample Keycloak Realm");
var webBlazorSsrClientId = builder.AddParameter("web-blazorssr-client-id", "web.blazorssr");
var webBlazorSsrClientName = builder.AddParameter("web-blazorssr-client-name", "Blazor SSR Web App");
var apiWeatherClientId = builder.AddParameter("api-weather-client-id", "api.weather");
var apiWeatherClientName = builder.AddParameter("api-weather-client-name", "Weather API App");
var webBlazorSSRClientSecret = builder.AddParameter("web-blazorssr-client-secret", new GenerateParameterDefault { MinLength = 32, Special = false }, secret: true);
var apiWeatherClientSecret = builder.AddParameter("api-weather-client-secret", new GenerateParameterDefault { MinLength = 32, Special = false }, secret: true);
```

### Suggestion 2: `WithDefault`
```csharp
var keycloakRealmName = builder.AddParameter("keycloak-realm").WithDefault("SampleRealm");
var keycloakRealmDisplayName = builder.AddParameter("keycloak-realm-display").WithDefault("Sample Keycloak Realm");
var webBlazorSsrClientId = builder.AddParameter("web-blazorssr-client-id").WithDefault("web.blazorssr");
var webBlazorSsrClientName = builder.AddParameter("web-blazorssr-client-name").WithDefault("Blazor SSR Web App");
var apiWeatherClientId = builder.AddParameter("api-weather-client-id").WithDefault("api.weather");
var apiWeatherClientName = builder.AddParameter("api-weather-client-name").WithDefault("Weather API App");
var webBlazorSSRClientSecret = builder.AddParameter("web-blazorssr-client-secret", secret: true)
    .WithDefault(new GenerateParameterDefault { MinLength = 32, Special = false });
var apiWeatherClientSecret = builder.AddParameter("api-weather-client-secret", secret: true)
    .WithDefault(new GenerateParameterDefault { MinLength = 32, Special = false });
```

davidebbo:
## Description

See #5410 for details.

Fixes #5410

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [X] Yes
      - [X] Yes
      - [ ] No
      - [ ] Yes
      - [X] No, may not be necessary here.
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [X] Yes
    - Link to aspire-docs issue: None opened yet.
  - [ ] No
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5529)


src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -27,6 +27,43 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// <param name="defaultValue">A ParameterDefault that is used if the name is not found in the configuration</param>
+    /// <param name="secret">Optional flag indicating whether the parameter should be regarded as secret.</param>
+    /// <returns>Resource builder for the parameter.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters",
+                                                     Justification = "third parameters are mutually exclusive.")]
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, ParameterDefault defaultValue, bool secret = false)
+    {
+        return builder.AddParameter(
+            name,
+            parameterDefault => GetParameterValue(builder.Configuration, name, parameterDefault),
```
eerhardt, line 62: What should be the "interaction" with UserSecretsParameterDefault here?

When we generate a default password for things like PostgreSQL databases, we generate the password once, and then store it in User Secrets of the AppHost so it doesn't get generated on the next run. Should this do something similar?

https://github.com/dotnet/aspire/blob/6db273bbef306b1c62ad27f91816ea095a9df972/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs#L190-L201

davidebbo: I see your point. Two options:
1. Have the user call ParameterResourceBuilderExtensions.CreateGeneratedParameter() themselves if they want that behavior. And then pass that to the new `AddParameter` overload.
2. Do it under the cover for them

And yes, 2. is probably better. And now I better understand why @DamianEdwards wanted different overloads for plain `ParameterDefault` vs `GenerateParameterDefault`, which makes this slightly cleaner (but is mostly indistinguishable for the user).

If that sounds good, I'll go ahead and do that.

eerhardt: > Have the user call ParameterResourceBuilderExtensions.CreateGeneratedParameter() themselves if they want that behavior. And then pass that to the new AddParameter overload.

`ParameterResourceBuilderExtensions.CreateGeneratedParameter()` returns a `ParameterResource` (one that isn't directly attached to the App Model). So it can't be passed to the new AddParameter overload - since that takes a `ParameterDefault`.

I think the distinction between plain `ParameterDefault` vs `GenerateParameterDefault` isn't that obvious - especially if one overload does something different.

Do you (or @DamianEdwards or @davidfowl or @mitchdenny) have any thoughts on adding a `bool persistDefaultValue = false` parameter? Or potentially using a different method name for the method that persists the value in user secrets.

davidebbo: Sorry, yes, my `ParameterResourceBuilderExtensions.CreateGeneratedParameter` suggestion made no sense. I meant something like:

```c#
public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, GenerateParameterDefault defaultValue, bool secret = false)
{
    ParameterDefault paramDefault = defaultValue;
    if (builder.ExecutionContext.IsRunMode && builder.AppHostAssembly is not null)
    {
        paramDefault = new UserSecretsParameterDefault(builder.AppHostAssembly, builder.Environment.ApplicationName, name, paramDefault);
    }

    return builder.AddParameter(
        name,
        paramDefault,
        secret: secret);
}
```

Your `persistDefaultValue` suggestion is potentially interesting.
- Pros: avoids special casing GenerateParameterDefault. And there could be other ParameterDefault implementations that need persistence.
- Cons: a bit harder to use, if we assume that GenerateParameterDefault should almost always be persisted

Would like to hear what others thing as well.

davidebbo: @eerhardt I ended up going with a variation:
- Single method that takes in a base `ParameterDefault`. No special case for `GenerateParameterDefault`.
- Instead, persistent is driven by a new `ParameterDefault.NeedsPersistence` property, which is false by default, and that `GenerateParameterDefault` sets to true. I kept `NeedsPersistence` internal, but we could expose (now or later)

PTAL

eerhardt: I think this works. I like that we don't expose it until we need to. But it gives us a nice way of opening it up in the future.

Can you write a test for it?

davidebbo: Thanks for the review. I added a simple test that checks that the `UserSecretsParameterDefault` wrapping only happens when it should.

eerhardt (approved):
This LGTM. Thanks for the contribution, @davidebbo!

@DamianEdwards @davidfowl @mitchdenny - any other thoughts?

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -27,6 +27,54 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// <param name="secret">Optional flag indicating whether the parameter should be regarded as secret.</param>
+    /// <returns>Resource builder for the parameter.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters",
+                                                     Justification = "third parameters are mutually exclusive.")]
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, string defaultValue, bool secret = false)
+    {
+        // An alternate implementation is to use some ConstantParameterDefault implementation that always returns the default value.
+        // However, doing this causes a "default": {} to be written to the manifest, which is not valid.
+        // And note that we ignore parameterDefault in the callback, because it can never be non-null, and we want our own default.
+        return builder.AddParameter(name, parameterDefault => builder.Configuration[$"Parameters:{name}"] ?? defaultValue, secret: secret);
```
davidfowl, line 45: I'm struggling with this one. Why is defaultValue a fallback instead of an override?

eerhardt: It's the "default value". I.e. "if no one else supplied a value - use this one".

We would call it `string value`, if we wanted to always win.

davidfowl: fallback value?

davidebbo: Sounds like a continuation of your earlier conversation: https://github.com/dotnet/aspire/issues/4429#issuecomment-2158838761. And frankly, I'm a bit confused about the comment there that says that "code wins". To me, "config wins", and the code is a fall back, which is what my code is doing.

eerhardt: In C#, when you have "optional arguments", the value that is used if you don't supply a value to the argument is called a "default vaule":

https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments

> Each optional parameter has a default value as part of its definition. If no argument is sent for that parameter, the default value is used.

davidebbo: > In C#, when you have "optional arguments", the value that is used if you don't supply a value to the argument is called a "default value"

Right, though this is quite different from our situation here, where the value can come from two completely different systems (config vs code). Here, I could see `fallbackValue` as a reasonable name, though I don't think `defaultValue` is so bad either.

davidebbo: Started Discord thread to make it easier to discuss: https://discord.com/channels/732297728826277939/1281287950491193436

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -52,18 +52,17 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
     /// <param name="name">Name of parameter resource</param>
     /// <param name="value">A <see cref="ParameterDefault"/> that is used to provide the parameter value</param>
     /// <param name="secret">Optional flag indicating whether the parameter should be regarded as secret.</param>
+    /// <param name="persist">Persist the value to the app host project's user secrets store. This is typically
+    /// sone when the value is generated, so that it stays stable across runs. This is only relevant when
```
eerhardt, line 56:
```suggestion
    /// done when the value is generated, so that it stays stable across runs. This is only relevant when
```

davidfowl (approved):

PR #5538, opened 2024-09-04
Add command annotations and wire up

### LINKED ISSUE #295: Add support for restarting services from the dashboard

### LINKED ISSUE #1410: Flaky test: Aspire.Hosting.Tests.Dashboard.ResourcePublisherTests.ProducesExpectedSnapshotAndUpdates
https://dev.azure.com/dnceng-public/public/_build/results?buildId=500597&view=results

Error message
```
Assert.Throws() Failure
Expected: typeof(System.OperationCanceledException)
Actual:   (No exception was thrown)
```

Stack trace
```
   at Aspire.Hosting.Tests.Dashboard.ResourcePublisherTests.ProducesExpectedSnapshotAndUpdates() in /_/tests/Aspire.Hosting.Tests/Dashboard/ResourcePublisherTests.cs:line 53
```

cc @drewnoakes

JamesNK:
Fixes https://github.com/dotnet/aspire/issues/1410 (cleaned up test while I was here)
Fixes https://github.com/dotnet/aspire/issues/295

Changes in PR:

* Add resource command annotations API
* Resource commands talking to host and invoking annotation callbacks
* Start/stop/restart added to containers/projects/executable
* Start/stop/restart unit and functional tests 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [x] Yes - will schedule
      - [ ] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1661, https://github.com/dotnet/docs-aspire/issues/1662
  - [ ] No 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5538)


src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>
+    /// Gets the target of the mount.
+    /// </summary>
+    public string DisplayName { get; }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public Func<CustomResourceSnapshot, Task<bool>> Visible { get; }
```
JamesNK, line 44: This probably shouldn't be async.

mitchdenny: I think you'll probably want this to take a `ResourceCommandContext` which exposes `IServiceProvider`. The `invokeCommand` could take the same context.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(type);
+        ArgumentNullException.ThrowIfNull(displayName);
+
+        Type = type;
+        DisplayName = displayName;
+        Visible = visible;
+        InvokeCommand = invokeCommand;
+        IconContent = iconContent;
+        IsHighlighted = isHighlighted;
```
mitchdenny, line 28: Not critical for V1, but we'll probably want to have some notion of sort order/priority on the commands and possibly some kind of category.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
mitchdenny: I think the visibility should be an enumeration with `Enabled/Disabled/Hidden`. Unless you want to break enabled/disabled into different values.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        builder.WithCommand(
+            "start",
+            "Start",
+            resourceSnapshot => Task.FromResult(IsStopped(resourceSnapshot.State?.Text)),
+            async (resourceName, services) =>
+            {
+                await services.GetRequiredService<ApplicationExecutor>().StartResourceAsync(resourceName).ConfigureAwait(false);
+            },
+            "<path d=\"M17.22 8.69a1.5 1.5 0 0 1 0 2.62l-10 5.5A1.5 1.5 0 0 1 5 15.5v-11A1.5 1.5 0 0 1 7.22 3.2l10 5.5Zm-.48 1.75a.5.5 0 0 0 0-.88l-10-5.5A.5.5 0 0 0 6 4.5v11c0 .38.4.62.74.44l10-5.5Z\"/>",
```
mitchdenny, line 22: I think we need to allow for custom icons but what safety measures do we need here.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+using Aspire.Hosting.Dcp;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Aspire.Hosting.Dashboard;
+
+internal static class CommandsConfigurationExtensions
+{
+    internal static IResourceBuilder<T> WithLifeCycleCommands<T>(this IResourceBuilder<T> builder) where T : IResource
+    {
+        builder.WithCommand(
```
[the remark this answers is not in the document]
mitchdenny: Yes I think command identifiers should be stringy.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
mitchdenny: Mind you ... there is way more you could do if you have code that is executing here. No need to use a HTML injection attack.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Dcp;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Aspire.Hosting.Dashboard;
+
+internal static class CommandsConfigurationExtensions
+{
+    internal static IResourceBuilder<T> WithLifeCycleCommands<T>(this IResourceBuilder<T> builder) where T : IResource
```
mitchdenny, line 12: This name is probably noo generic.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -33,7 +34,8 @@ public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistribut
                           {
                               context.Args.AddRange(args);
                           }
-                      });
+                      })
+                      .WithLifeCycleCommands();
```
mitchdenny, line 38: Rather than doing this I think it might be better to subscribe to an event and then loop across all the resources and add the commands. Something like this:

```csharp
builder.Eventing.Subscribe<BeforeStartEvent>((@event, ct) => {
 ... loop over the app model and inject commands into the resources.
});
```

The reason is that you'll need to add these common commands to each custom resource (e.g. JS and Python resource types). By using an event you can add this feature to all resource types that are either containers, projects or executable based without needing to leak the command annotation types out of the `Aspire.Hosting` assembly.

mitchdenny: ... although type visibility is probably not an issue since we will want to expose `ResourceCommandAnnotation` in .NET 9.0 to allow for extensibility for folks.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /// <summary>
+    /// Gets the source of the bind mount or name if a volume. Can be <c>null</c> if the mount is an anonymous volume.
```
karolz-ms, line 32: This (and DisplayName) needs a better summary ☺️

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public string Type { get; }
```
karolz-ms, line 34: Consider a different name for this property, perhaps simply `Name`. As I understand, this is used as a key/ID in the set of commands available for a particular resource. `Type` suggests there can be more than one command with the same value of this property.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
karolz-ms, line 28: What does it look like when the command is "highlighted" vs "not highlighted"? How does it affect command operation?

The "highlighted" name suggests "UI styling" which makes me wonder if this information should be part of the view model...

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        builder.WithCommand(
+            "start",
+            "Start",
+            resourceSnapshot => Task.FromResult(IsStopped(resourceSnapshot.State?.Text)),
```
karolz-ms, line 17: This should probably be delegated to ApplicationExecutor as well, because the logic that determines whether something can be started/stopped/restarted may be more complicated than f: resourceState -> bool.

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
```diff
@@ -77,6 +81,18 @@ public async ValueTask DisposeAsync()
         _cts.Dispose();
     }
 
+    internal async Task ExecuteCommandAsync(string resourceId, string type)
+    {
+        if (_resourcePublisher.TryGetResource(resourceId, out _, out var resource))
```
karolz-ms, line 86: If resource does not exist, or the command does not exist, perhaps we should log that as a warning/error.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -179,14 +180,38 @@ void WriteToChannel(ResourceEvent resourceEvent) =>
...
+                }
+                if (visible)
+                {
+                    commands[commandAnnotation.Type] = new ResourceCommandSnapshot(commandAnnotation.Type, commandAnnotation.DisplayName, commandAnnotation.IconContent, commandAnnotation.IsHighlighted);
+                }
+                else
+                {
+                    commands.Remove(commandAnnotation.Type);
+                }
+            }
```
davidfowl, line 211: This doesn't look very efficient...

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
JamesNK: I like it. Change the API to return a state.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
JamesNK: It's internal so it doesn't matter. Suggestions welcome.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
JamesNK: The resource snapshot data comes from ApplicationExecutor. If we need more information on it to make this decision I think it would be better to have it available on the model, and then use it here.

karolz-ms: OK that makes sense

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
JamesNK: See https://github.com/dotnet/aspire/pull/5516. The difference is whether the command is directly displayed on the resources page or is nested in the actions menu.

> The "highlighted" name suggests "UI styling" which makes me wonder if this information should be part of the view model...

Eventually this flag is used in the UI, but the info for whether a command is highlighted or not needs to come from somewhere. That's why it's part of the annotation (along with other display info such as display name and icon).

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
JamesNK: It's run when a resource changes so it shouldn't run often. But I'm sure there are ways to be smarter here about changes.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
karolz-ms: Got it.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
JamesNK: I agree that name would be better than type. Type is already used in the proto API between the host and dashboard.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
JamesNK: Order of annotations can be the order of commands. If there is feedback about it, then we can add a property in the future for additional ordering control.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
JamesNK: If the annotations were added in an event just before the resource was started, there isn't an opportunity for the built-in commands to be removed. I don't know if that's valuable, but I think it's one of the things you brought up in the design meeting.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
JamesNK: I'll get to that later. Just proof of concept right now.

JamesNK:
Made a bunch of changes and applied PR feedback.

I think this is close to being good. (Ignore lack of API xml comments and the fact that start/stop/restart don't actually do anything yet 😄)

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
karolz-ms: OK

karolz-ms (approved):

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            "stop",
+            "Stop",
+            context => !IsStopped(context.ResourceSnapshot.State?.Text) ? ResourceCommandState.Enabled : ResourceCommandState.Hidden,
+            async context =>
+            {
+                var executor = context.ServiceProvider.GetRequiredService<ApplicationExecutor>();
+
+                await executor.StopResourceAsync(context.ResourceName).ConfigureAwait(false);
+            },
+            "<path d=\"M15.5 4c.28 0 .5.22.5.5v11a.5.5 0 0 1-.5.5h-11a.5.5 0 0 1-.5-.5v-11c0-.28.22-.5.5-.5h11Zm-11-1C3.67 3 3 3.67 3 4.5v11c0 .83.67 1.5 1.5 1.5h11c.83 0 1.5-.67 1.5-1.5v-11c0-.83-.67-1.5-1.5-1.5h-11Z\"/>",
```
davidfowl, line 37: I thought we got rid of the svg?

JamesNK: The PR hasn't been rebased on latest command changes.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -219,9 +224,57 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
                     newState.ExitCode, string.Join(", ", newState.EnvironmentVariables.Select(e => $"{e.Name} = {e.Value}")), string.Join(", ", newState.Urls.Select(u => $"{u.Name} = {u.Url}")),
                     string.Join(", ", newState.Properties.Select(p => $"{p.Name} = {p.Value}")));
             }
+        }
+
+        return Task.CompletedTask;
+    }
+
+    private static ImmutableArray<ResourceCommandSnapshot> BuildCommands(IResource resource, CustomResourceSnapshot previousState)
```
davidfowl, line 232: This is too expensive to do every publish. The goal is to figure out which commands are visible given the state correct?

JamesNK: I'll look at making it faster, but how often are you expecting resources to be published? It's not like we need to do thousands of updates per second.

davidfowl: every 5 seconds after the health checks PR is made more aggressive.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -150,7 +149,8 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
             var browserToken = options.DashboardToken;
             var otlpApiKey = options.OtlpApiKey;
 
-            var resourceServiceUrl = await dashboardEndpointProvider.GetResourceServiceUriAsync(context.CancellationToken).ConfigureAwait(false);
+            // Get service from provider to avoid circular reference.
```
davidfowl, line 152: 😢

src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
```diff
@@ -49,7 +49,7 @@ internal sealed class DashboardServiceHost : IHostedService
     public DashboardServiceHost(
         DistributedApplicationOptions options,
         DistributedApplicationModel applicationModel,
-        IKubernetesService kubernetesService,
+        ApplicationExecutor applicationExecutor,
```
davidfowl, line 52: Why?

JamesNK: So we can call start/stop on the executor from the commands.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
JamesNK:
```
System.AggregateException: 'Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Aspire.Hosting.Dashboard.DashboardServiceHost Lifetime: Singleton ImplementationType: Aspire.Hosting.Dashboard.DashboardServiceHost': A circular dependency was detected for the service of type 'Aspire.Hosting.Dashboard.DashboardServiceHost'.
Aspire.Hosting.Dashboard.DashboardServiceHost -> Aspire.Hosting.Dcp.ApplicationExecutor -> System.Collections.Generic.IEnumerable<Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook> -> Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook(Aspire.Hosting.Dashboard.DashboardLifecycleHook) -> Aspire.Hosting.Dcp.IDashboardEndpointProvider(Aspire.Hosting.Dcp.HostDashboardEndpointProvider) -> Aspire.Hosting.Dashboard.DashboardServiceHost) (Error while validating the service descriptor 'ServiceType: Aspire.Hosting.Dcp.IDashboardEndpointProvider Lifetime: Singleton ImplementationType: Aspire.Hosting.Dcp.HostDashboardEndpointProvider': A circular dependency was detected for the service of type 'Aspire.Hosting.Dcp.IDashboardEndpointProvider'.
Aspire.Hosting.Dcp.IDashboardEndpointProvider(Aspire.Hosting.Dcp.HostDashboardEndpointProvider) -> Aspire.Hosting.Dashboard.DashboardServiceHost -> Aspire.Hosting.Dcp.ApplicationExecutor -> System.Collections.Generic.IEnumerable<Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook> -> Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook(Aspire.Hosting.Dashboard.DashboardLifecycleHook) -> Aspire.Hosting.Dcp.IDashboardEndpointProvider) (Error while validating the service descriptor 'ServiceType: Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook Lifetime: Singleton ImplementationType: Aspire.Hosting.Dashboard.DashboardLifecycleHook': A circular dependency was detected for the service of type 'Aspire.Hosting.Dcp.IDashboardEndpointProvider'.
Aspire.Hosting.Dcp.IDashboardEndpointProvider(Aspire.Hosting.Dcp.HostDashboardEndpointProvider) -> Aspire.Hosting.Dashboard.DashboardServiceHost -> Aspire.Hosting.Dcp.ApplicationExecutor -> System.Collections.Generic.IEnumerable<Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook> -> Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook(Aspire.Hosting.Dashboard.DashboardLifecycleHook) -> Aspire.Hosting.Dcp.IDashboardEndpointProvider) (Error while validating the service descriptor 'ServiceType: Aspire.Hosting.Dcp.ApplicationExecutor Lifetime: Singleton ImplementationType: Aspire.Hosting.Dcp.ApplicationExecutor': A circular dependency was detected for the service of type 'Aspire.Hosting.Dcp.ApplicationExecutor'.
Aspire.Hosting.Dcp.ApplicationExecutor -> System.Collections.Generic.IEnumerable<Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook> -> Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook(Aspire.Hosting.Dashboard.DashboardLifecycleHook) -> Aspire.Hosting.Dcp.IDashboardEndpointProvider(Aspire.Hosting.Dcp.HostDashboardEndpointProvider) -> Aspire.Hosting.Dashboard.DashboardServiceHost -> Aspire.Hosting.Dcp.ApplicationExecutor) (Error while validating the service descriptor 'ServiceType: Microsoft.Extensions.Hosting.IHostedService Lifetime: Singleton ImplementationType: Aspire.Hosting.Dcp.DcpHostService': A circular dependency was detected for the service of type 'Aspire.Hosting.Dcp.ApplicationExecutor'.
Microsoft.Extensions.Hosting.IHostedService(Aspire.Hosting.Dcp.DcpHostService) -> Aspire.Hosting.Dcp.ApplicationExecutor -> System.Collections.Generic.IEnumerable<Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook> -> Aspire.Hosting.Lifecycle.IDistributedApplicationLifecycleHook(Aspire.Hosting.Dashboard.DashboardLifecycleHook) -> Aspire.Hosting.Dcp.IDashboardEndpointProvider(Aspire.Hosting.Dcp.HostDashboardEndpointProvider) -> Aspire.Hosting.Dashboard.DashboardServiceHost -> Aspire.Hosting.Dcp.ApplicationExecutor)'
```

src/Aspire.Hosting/Dashboard/DashboardServiceHost.cs
JamesNK: I've thought about this more, and the issue is the ServiceProvider that is passed into the command callback when it executes. It has been the dashboard host's service provider. I think that's wrong, and it should be the app host's service provider.

Using the app host service locator means services registered in the app host by user code are available in registered commands. It also makes ApplicationExecutor available without having to pass it into the dashboard service host.

JamesNK: Fixed.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
JamesNK: No longer a problem.

Directory.Packages.props
```diff
@@ -103,6 +103,7 @@
     <PackageVersion Include="Grpc.Tools" Version="2.65.0" />
     <PackageVersion Include="Humanizer.Core" Version="2.14.1" />
     <PackageVersion Include="KubernetesClient" Version="14.0.9" />
+    <PackageVersion Include="JsonPatch.Net" Version="3.1.1" />
```
davidfowl, line 106: ???????????????????????

JamesNK: Look at where it is used in ApplicationExecutor

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -219,8 +222,82 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
...
             }
+        }
+
+        return Task.CompletedTask;
+    }
+
+    /// <summary>
+    /// Use command annotations to update resource snapshot.
+    /// </summary>
+    private static CustomResourceSnapshot UpdateCommands(IResource resource, CustomResourceSnapshot previousState)
```
davidfowl, line 233: Much better!

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1878,4 +1881,60 @@ private string ReplaceLocalhostWithContainerHost(string value)
...
+            await kubernetesService.PatchAsync(e, patch, cancellationToken).ConfigureAwait(false);
+        }
+        else if (matchingResource.DcpResource is ExecutableReplicaSet rs)
+        {
+            var patch = CreatePatch(rs, obj => obj.Spec.Stop = true);
+            await kubernetesService.PatchAsync(rs, patch, cancellationToken).ConfigureAwait(false);
+        }
+        else
+        {
+            throw new InvalidOperationException($"Unexpected resource type: {matchingResource.DcpResource.GetType().FullName}");
```
davidfowl, line 1929: What is the experience when this throws?

JamesNK: I don't think there is any. Just being defensive.

src/Aspire.Hosting/Dashboard/CommandsConfigurationExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        builder.WithCommand(
+            "restart",
+            "Restart",
+            context => !IsStopped(context.ResourceSnapshot.State?.Text) ? ResourceCommandState.Enabled : ResourceCommandState.Hidden,
+            async context =>
+            {
+                var executor = context.ServiceProvider.GetRequiredService<ApplicationExecutor>();
+
+                await executor.StopResourceAsync(context.ResourceName, context.CancellationToken).ConfigureAwait(false);
+                await executor.StartResourceAsync(context.ResourceName, context.CancellationToken).ConfigureAwait(false);
```
davidfowl, line 49: Does this work? Start/Stop in quick succession?

JamesNK: I talked with @karolz-ms about restarting containers, and it sounds like that work might not get done for 9. Therefor to restart both containers and executables, we'll delete the resource and then recreate it.

That means the `StopResourceAsync` here probably isn't needed. Stopping a resource between deleting doesn't seem necessary.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1039,86 +1039,90 @@ private void PrepareProjectExecutables()
                 throw new InvalidOperationException("A project resource is missing required metadata"); // Should never happen.
             }
 
-            int replicas = project.GetReplicaCount();
+            var replicas = project.GetReplicaCount();
 
-            var ers = ExecutableReplicaSet.Create(GetObjectNameForResource(project), replicas, "dotnet");
-            var exeSpec = ers.Spec.Template.Spec;
-            exeSpec.WorkingDirectory = Path.GetDirectoryName(projectMetadata.ProjectPath);
+            for (var i = 0; i < replicas; i++)
```
davidfowl, line 1044: Lets not implement this in the PR. We can block replicas for now. I don't think we should aim to fix it in this PR.

JamesNK: Could you tell what the specific problems are with switching from replica sets to executables? It's something you and @mitchdenny raised during a meeting. I think you said it would be too difficult to do in .NET 9. I didn't run into any troubles when changing code, and TestShop appeared to still work when I tested it.

The reason I'm trying out executables is I found a fair number of DCP bugs when scaling replica sets during start/stop.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Dashboard;
```
davidfowl, line 5: Lets fix this namespace. This shouldn't be dashboard related.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: I would love us to do this as a follow up. This PR is already doing a lot and changing how replicas work as part of it just increases the complexity of the review and feature (this affect more than start/stop). I would start by not supporting start/stop if there are any replicas and then we can make this change in isolation to see if there are any issues with it.

JamesNK: I'm finding a fair number of work specific to scaling replica sets.
* The DCP needs work to make replica sets have stable names between starting and stopping.
* Scaling up replica sets doesn't appear to work in DCP right now (bug?)
* The dashboard or host needs extra code to associate the resource command from the replica instance to the replica set object.

The point I see here is we will spend a lot of time getting replica sets with start/stop/restart, when we could switch now and focus on getting executables working instead.

davidfowl: OK if this is easier then go for it. If it starts to snowball then we back it out. I just want to avoid blocking this feature on replicas when it's not important. That's phase 2. Phase 1 is all of the other bootstrapping code. Just disable commands on replicas and do a follow up PR with these changes. 

I'll make it easier to review as well.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
JamesNK: I followed the prior art from `ConsoleLogsConfigurationExtensions` which is in this namespace.

davidfowl: Is this API internal? If so I am fine and we can change it later.

JamesNK: I changed it to Aspire.Hosting.ApplicationModel. Let me know if there is somewhere else you want it.

mitchdenny:
@JamesNK I think the updateState callback should be `Func<UpdateCommandStateContext, Task<ResourceCommandState>>` (or ValueTask ... dunno). Also ... is there is a reason it doesn't provide access to the service provider?

JamesNK:
I haven't heard a scenario where service provider or async is needed. If those are needed then the API can change.

If there is a scenario, I would like to know whether it makes sense for the update state methods to be heavy and async, or lightweight:
* A heavy func (resolving services and making async calls) is called every time a snapshot updates - which could be many times a second.
* Or, is better to use some other API hook to add more data to the snapshot via ResourceNotificationService about the state of the resource, and then use the cached properties on the snapshot in the func to control whether a command is enabled or disabled.

I'm sure we've had this conversation before.

mitchdenny:
The purpose of the update state methods is to determine whether the command should be enabled or not. In the implementation of the start/stop commands you are providing the latest resource snapshot which enables you to make a decision as to whether command should be enabled or not - that is generally sufficient for your scenario.

However, when applied more broadly the resource state snapshot is not the only input into deciding whether a command might be enabled or not. For example you might want to query the state of the resource directly to determine whether a command should be available. This would require resolving the connection string to the resource which is an async operation.

We could force people to have to implement a service in DI to do these kinds of async operations in the background - but this would require as to provide the service provider at the very least (so you could resolve the service).

If we are concerned about the overhead of running the state updates for commands then we can always poll at a lower interval.

I'd prefer to give people options (service resolution and async) to allow them to write the code that is right for their particular scenario. Often times you don't get the choose the APIs that you need to interact with and if you just want to quickly mockup a command to add to the app model to optimize something then let's give them the capability.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -0,0 +1,99 @@
...
+public sealed class ResourceCommandAnnotation : IResourceAnnotation
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ResourceCommandAnnotation"/> class.
+    /// </summary>
+    public ResourceCommandAnnotation(
+        string type,
+        string displayName,
+        Func<UpdateCommandStateContext, ResourceCommandState> updateState,
+        Func<ExecuteCommandContext, Task> executeCommand,
```
mitchdenny, line 22: I was thinking about whether we should return something like the equivalent of `Task<IActionResult>` (not actual `IActionResult` just trying to conjure the pattern). The idea is that that no only does the task being awaited some indication of success or failure, but it also allows us to provide the dashboard with additional information that can be used to control what happens.

For example we might want to provide some kind of call to action button that the user can click to navigate away to a page - or in some cases a redirect action which would immediately navigate the page away to a specific URL. For example if someone adds the `WithPgAdmin(...)` command then all the databases on that server would get a command _View in PgAdmin_ that if they clicked on it would be deep linked to that database in the PgAdmin instance that the database is mapped to.

Other options for projects might be doing things like scanning the project directory for .http files and adding a command for each .http file so that when it is clicked the .http request is executed and the JSON result is displayed in a JSON viewer in the screen.

We don't necessarily need all these options straightaway but given that there are some useful scenarios here making this return a more fully fledged object might be an idea. Then we could extend it overtime with more scenarios.

JamesNK:
> However, when applied more broadly the resource state snapshot is not the only input into deciding whether a command might be enabled or not. For example you might want to query the state of the resource directly to determine whether a command should be available. This would require resolving the connection string to the resource which is an async operation.

Here is the problem with what you're saying: the update method is only triggered when the resource snapshot updates. If you need to talk to the database to figure out the state, you can only do that if something updates the resource snapshot. The resource might be in a happy Running state since it starts up and the command state methods are never called.

For example, there is a command to perform an EF migration, and you query the DB to see if the action is enabled or disabled. The command could be enabled and some other tool could perform the migration, but the command stays enabled because the resource snapshot doesn't update.

The way this could work instead is when the resource starts there is a background task that queries the database (or file, or web service, etc) every 10 seconds to figure out whether the command is valid or not. That task uses `ResourceNotificationService` to set a property on the resource snapshot. The command update method then looks at that property to figure out whether it is enabled or not.

> We probably shouldn't be polling this multiple times a second anyway (its UX behind a context menu). If we are concerned about the overhead of running the state updates for commands then we can always poll at a lower interval.

This isn't from polling. The methods are re-evaluated when the resource state updates, which is pushed from the notification service. A resource starting up or shutting down has half a dozen rapid resource state updates as the state and endpoints change.

src/Aspire.Dashboard/Components/Controls/ResourceActions.razor.cs
```diff
@@ -49,7 +49,7 @@ protected override void OnParametersSet()
             OnClick = OnConsoleLogs.InvokeAsync
         });
 
-        var menuCommands = Commands.Where(c => !c.IsHighlighted).ToList();
+        var menuCommands = Commands.Where(c => !c.IsHighlighted && c.State != CommandViewModelState.Hidden).ToList();
```
mitchdenny, line 52: Should the context menu show all commands regardless of whether they are highlighted or not?

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
```diff
@@ -33,7 +33,7 @@
                     Class="severity-icon" />
     }
 }
-else if (Resource.IsStartingOrBuildingOrWaiting())
+else if (Resource.IsUnusableTransitoryState())
```
mitchdenny, line 36: It's a race with #5770 :)

mitchdenny (approved):
LGTM. Some minor points of feedback that can be follow up PRs if necessary.

src/Aspire.Dashboard/Components/Controls/ResourceActions.razor.cs
JamesNK: Dunno. It's easy to change.

src/Aspire.Dashboard/Components/Controls/ResourceActions.razor
```diff
@@ -2,12 +2,12 @@
 @using Aspire.Dashboard.Model
 @using Microsoft.FluentUI.AspNetCore.Components
 
-@foreach (var highlightedCommand in Commands.Where(c => c.IsHighlighted))
+@foreach (var highlightedCommand in Commands.Where(c => c.IsHighlighted && c.State != CommandViewModelState.Hidden))
```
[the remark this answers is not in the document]
JamesNK: Maybe. It's easy to change if necessary.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1952,4 +1960,127 @@ private string ReplaceLocalhostWithContainerHost(string value)
...
+                // This could happen in a race condition, e.g. double clicking start button.
+                resourceNotFound = true;
+            }
+
+            // Ensure resource is deleted. DeleteAsync returns before the resource is completely deleted so we must poll
+            // to discover when it is safe to recreate the resource. This is required because the resources share the same name.
+            if (!resourceNotFound)
+            {
+                // Limit polling to 5 attempts to avoid hanging with an infinite loop.
+                for (var i = 0; i < 5; i++)
```
davidfowl, line 2061: Why don't you use polly 😄

JamesNK: I've used polly zero times and I've used for loops a million times 😋

Next time I make start/stop/restart changes in this area I'll try it out.

davidfowl: It's in this same file

JamesNK: Since the build is failing anyway, changed to polly.

Observation: I haven't used polly before, but should the polly options and pipelines be cached? They're created every run right now. Might not be necessary.

PR #5540, opened 2024-09-04
Allow Dashboard application to roll-forward

### LINKED ISSUE #5445: Allow Dashboard to roll-forward
The Aspire Dashboard currently ships as a NuGet package, which contains it as a framework-dependent application. This application targets net8.0 and doesn't allow roll-forward, which means that if 9.0 SDK users don't have 8.0 runtime installed they will fail to launch the dashboard. To fix this, we need to change the dashboard so that we allow roll-forward, which would then help 9.0 users be able to launch the dashboard.

joperezr:
## Description

In order to better support customers that only have the 9.0 SDK/Runtime installed, we will allow the Dashboard application to roll-forward to the next major, which would then not require the 8.0 runtime to also be installed in order to run the Aspire Dashboard. I have manually validated that with these changes I am able to run an Aspire application in a machine that only has .NET 9 installed. Automated tests will require having a setup where the 9.0 SDK is used, which will be added by @radical with the planned template work for 9.0. Once that is in place, we can go ahead and add automation for this.

Fixes https://github.com/dotnet/aspire/issues/5445

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No. Tests will come in a subsequent PR once 9.0 tests infra is ready.
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5540)

### DIFF
--- a/src/Aspire.Dashboard/Aspire.Dashboard.csproj
+++ b/src/Aspire.Dashboard/Aspire.Dashboard.csproj
@@ -11,6 +11,12 @@
       CS8002: Referenced assembly does not have a strong name
     -->
     <NoWarn>$(NoWarn);CS1591;CS8002</NoWarn>
+
+    <!-- This Application is a framework-dependent application that targets .NET 8.0. This means that the
+    application is not able to run if the place where it is deployed does not have the .NET 8.0 runtime installed.
+    Given we want to be able to support people running on environments where they only have the 9.0 SDK/runtime installed,
+    we allow roll-forward to the next major in order to support these customers.-->
+    <RollForward>Major</RollForward>
   </PropertyGroup>
 
   <PropertyGroup>


src/Aspire.Dashboard/Aspire.Dashboard.csproj
```diff
@@ -11,6 +11,12 @@
       CS8002: Referenced assembly does not have a strong name
     -->
     <NoWarn>$(NoWarn);CS1591;CS8002</NoWarn>
+
+    <!-- This Application is a framework-dependent application that targets .NET 8.0. This means that the
+    application is not able to run if the place where it is deployed does not have the .NET 8.0 runtime installed.
+    Given we want to be able to support people running on environments where they only have the 9.0 SDK/runtime installed,
+    we allow roll-forward to the next major in order to support these customers.-->
+    <RollForward>Major</RollForward>
```
eerhardt, line 19: How are we going to test this?

joperezr: I wrote in the description, the plan is that once we add 9.0 templates (@radical  is working on it) we'll be adding tests specifically for environments with only 9.0 SDK installed (to validate that 9.0 templates can be created and built), so at that time we'll be able to add a test for this by using the 9.0 SDK to create a template, run it, and then validate that the dashboard is loading correctly.

eerhardt: Do we also want to run the Dashboard tests on net9? "validate that the dashboard is loading correctly" is a good first step, but it won't cover all the use cases of the dashboard.

joperezr: We can certainly consider it, but not sure if we would get a lot from it so perhaps we can consider doing it in a format similar to runtime's outerloop runs (run once a day maybe?). Point being those tests would mostly be trying to catch breaking changes between 8.0 and 9.0, which while they are very possible, I'm guessing it is not something that may add a lot of value on testing on each CI run.

joperezr: My above suggestion for "is it loading correctly" was simply to validate that the property is doing it's job and the 9.0 runtime is effectively being able to run the dashboard to begin with.

joperezr:
@eerhardt any other comments or is this good to go? It would be good to include in 9.0 preview 4, which is why I ask

radical:
How can this be tested locally?

joperezr:
The way I did it, was:

1. Build the changes on this PR to generate a new set of packages and workload.
1. Install 9.0 SDK locally to a folder, and add it to your path at the beginning.
2. set `DOTNET_MULTILEVEL_LOOKUP=0` so that only this SDK/runtime is used
3. Install the produced workload on this SDK and create a new aspire project (aspire-starter or anything that has an AppHost)
4. Because we don't yet have 9.0 targeting TFMs, manually open the project and replace all `net8.0` TFMs in the projects with `net9.0` TFMs for the `TargetFramework` properties.
5. Build and Run and make sure the dashboard loads correctly

In the above case, the dashboard would still be targeting 8.0 but the SDK you are using doesn't have that runtime installed, but the dashboard is still able to run due to the rollForward policy.

eerhardt (approved):
Let's do it! :shipit: 

But let's be sure to get some tests when possible.

Issue #2689, opened 2024-03-06
[tests] Workload tests


radical:
These would be tests that exercise the workload, like:

# Template tests
- [ ] #9576
- [x] with no workload installed, a project with `$(IsAspireHost)=true` would fail to build due to missing `aspire` workload
- [x] basic templates can build and run
- [x] Starter template can run (WIP)
- [x] Starter template's included test project works (WIP)
- [x] create+build+run starter template, and run some basic checks including checking the dashboard (WIP)

# Dashboard

- [x] verify all the resources show up (WIP)

# Others
- [x] This is being handled indirectly with playground, and EndToEnd tests. ~~Test per published nuget, to use the nuget and perform some basic validation~~
- [ ] manifest publishing works with `dotnet run`
- [x] default https works on linux
- [x] `ASPIRE_ALLOW_UNSECURED_TRANSPORT`

Future:

radical:
cc @joperezr

eerhardt:
If I was dreaming, we would also have tests that publish an app to Azure. But that is going to take a lot of infrastructure.

> with no workload installed, a project with $(IsAspireHost)=true would fail to build due to missing aspire workload

I'd put this tests pretty low on the priority list.

eerhardt:
What is the medium-to-long-term vision for where the Azure Hosting code lives? If we intend to move it to https://github.com/Azure/azure-sdk-for-net/, it may make sense to wait until then to hook it up to automated tests.

On the Components side, my goal is to eventually move them to the Azure repo, so they can ship and service along with the Azure SDK packages.

radical:
Combined the remaining tasks in #2689 and https://github.com/dotnet/aspire/issues/6076 into https://github.com/dotnet/aspire/issues/9576 . Closing.

PR #5556, opened 2024-09-05
Allow referencing older version of AppHost package for backward compatibility

joperezr:
## Description

This would walk back on a decision we took in 8.2.0 where we forced people to reference the 8.2.0 AppHost package when moving to the new workload, and instead allow for backward compatibility preserving the old behavior where, if that is the case, then the version of dashboard and DCP people would get would match the version of the workload they have installed.

The intention would be to backport this to release/8.2 branch so it can be serviced in 8.2.1

cc: @DamianEdwards @davidfowl @eerhardt @maddymontaquila @radical 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5556)

### DIFF
--- a/src/Aspire.Hosting.Sdk/SDK/Sdk.in.targets
+++ b/src/Aspire.Hosting.Sdk/SDK/Sdk.in.targets
@@ -71,13 +71,17 @@
       <_AppHostVersion>%(_AppHostPackageReference.Version)</_AppHostVersion>
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(_AppHostVersion)' != ''">
-      <__CurrentAppHostVersionMessage> You are using version $(_AppHostVersion).</__CurrentAppHostVersionMessage>
+    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
+      <!-- If we find the version to Aspire.Hosting.AppHost package but it is lower than 8.2.0, then we fall back
+      to use the Dashboard and DCP packages that match the version of the installed workload for backwards compatibility.
+      This results in the same behavior that we had before moving Dashboard and DCP out of the workload, since the version
+      is again just matching to the one the workload has. -->
+      <_AppHostVersion>@VERSION@</_AppHostVersion>
     </PropertyGroup>
 
     <!-- At this point, we should have the version either by CPM or PackageReference, so we fail if not.  -->
-    <Error Condition="'$(_AppHostVersion)' == '' or $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))" 
-       Text="$(MSBuildProjectName) is a .NET Aspire AppHost project that needs a package reference to Aspire.Hosting.AppHost version 8.2.0 or above to work correctly.$(__CurrentAppHostVersionMessage)" />
+    <Error Condition="'$(_AppHostVersion)' == ''" 
+       Text="$(MSBuildProjectName) is a .NET Aspire AppHost project that needs a package reference to Aspire.Hosting.AppHost version 8.2.0 or above to work correctly." />
 
     <!-- Now that we have the version, we add the package references -->
     <ItemGroup>
--- a/tests/Aspire.Workload.Tests/AppHostTemplateTests.cs
+++ b/tests/Aspire.Workload.Tests/AppHostTemplateTests.cs
@@ -1,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Xunit;
+using Xunit.Abstractions;
+using System.Diagnostics;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Workload.Tests;
+
+public partial class AppHostTemplateTests : WorkloadTestsBase
+{
+    public AppHostTemplateTests(ITestOutputHelper testOutput)
+        : base(testOutput)
+    {
+    }
+
+    [Fact]
+    public async Task EnsureProjectsReferencing8_1_0WorkloadCanBuild()
+    {
+        string projectId = "aspire-can-reference-8.1.0";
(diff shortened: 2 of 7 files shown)


maddymontaquila:
do ittttt. there's no way this is going to break CI systems again, right? 😄

joperezr:
> do ittttt. there's no way this is going to break CI systems again, right? 😄

Well now you've jinxed it 🤭....jk.

The plan is that this change would undo the regression, and allow people to remove the manual workarounds they had to do to pin an older version of the workload.

KennethHoff:
This is a temporary solution until 9.0, right? iiuc 9.0 will no longer require installing the workload, as the only thing it will do is keep the templates up-to-date. Is that the right understanding?

DamianEdwards:
@KennethHoff correct

mitchdenny:
Wouldn't this need to go only into the 8.2 branch instead of main?

joperezr:
The plan is to backport to 8.2 once this is merged. We still have to do workload authoring changes for 9 which will modify this code again in main branch, so for now we should have it in both places.

joperezr:
@DamianEdwards @eerhardt @radical any feedback here? Would be good to get this one closed so we can open the backport PR

tests/Aspire.Workload.Tests/AppHostTemplateTests.cs
```diff
@@ -0,0 +1,42 @@
...
+
+public partial class AppHostTemplateTests : WorkloadTestsBase
+{
+    public AppHostTemplateTests(ITestOutputHelper testOutput)
+        : base(testOutput)
+    {
+    }
+
+    [Fact]
+    public async Task EnsureProjectsReferencing8_1_0WorkloadCanBuild()
```
radical, line 18: Should this be named `EnsureProjectsReferencing8_1_0AppHostWithNewerWorkloadCanBuild`?

DamianEdwards:
> @DamianEdwards @eerhardt @radical any feedback here? Would be good to get this one closed so we can open the backport PR

Conceptually good from my side.

tests/Aspire.Workload.Tests/AppHostTemplateTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        string projectId = "aspire-can-reference-8.1.0";
+        await using var project = await AspireProject.CreateNewTemplateProjectAsync(
+            projectId,
+            "aspire-apphost",
```
radical, line 23: Would it be useful to test with the starter template instead, and check that it runs as expected? We have existing code for that.

src/Aspire.Hosting.Sdk/SDK/Sdk.in.targets
```diff
@@ -71,13 +71,17 @@
...
 
-    <PropertyGroup Condition="'$(_AppHostVersion)' != ''">
-      <__CurrentAppHostVersionMessage> You are using version $(_AppHostVersion).</__CurrentAppHostVersionMessage>
+    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
+      <!-- If we find the version to Aspire.Hosting.AppHost package but it is lower than 8.2.0, then we fall back
+      to use the Dashboard and DCP packages that match the version of the installed workload for backwards compatibility.
+      This results in the same behavior that we had before moving Dashboard and DCP out of the workload, since the version
+      is again just matching to the one the workload has. -->
+      <_AppHostVersion>@VERSION@</_AppHostVersion>
     </PropertyGroup>
```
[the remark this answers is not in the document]
joperezr: Exactly correct, which would match the old behavior we had in the workload where dcp and dashboard version always matched the workload. This is for backwards compat only.

tests/Aspire.Workload.Tests/AppHostTemplateTests.cs
joperezr: I saw we weren't testing this template so thought about using it too to get coverage, plus this particular test really only cares about building, as if this scenario is regressed the main issue would be that the user would not be able to build.

radical: Yeah, we need to add tests for `apphost`, and the `servicedefaults` project (https://github.com/dotnet/aspire/issues/2689).

src/Aspire.Hosting.Sdk/SDK/Sdk.in.targets
```diff
@@ -71,13 +71,17 @@
...
 
     <!-- At this point, we should have the version either by CPM or PackageReference, so we fail if not.  -->
-    <Error Condition="'$(_AppHostVersion)' == '' or $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))" 
-       Text="$(MSBuildProjectName) is a .NET Aspire AppHost project that needs a package reference to Aspire.Hosting.AppHost version 8.2.0 or above to work correctly.$(__CurrentAppHostVersionMessage)" />
+    <Error Condition="'$(_AppHostVersion)' == ''"
```
radical, line 83: A test for this would be good too. It doesn't have to be in this PR.
Also, the same for both cases but with CPM?

radical (approved):
Other than the feedback, LGTM 👍

src/Aspire.Hosting.Sdk/SDK/Sdk.in.targets
joperezr: I'll add more tests to this whole scenario with the changes on the Aspire Workload for 9.0

PR #5572, opened 2024-09-06
Show container volumes in the dashboard

### LINKED ISSUE #3969: Make container volumes visible from the dashboard
Volume information doesn't show up anywhere in the dashboard. It should.

drewnoakes:
## Description

Fixes #3969

This plumbs through information about volumes mounted to containers, for display in the dashboard alongside other container details.

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5572)


src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -77,6 +82,14 @@ public sealed record EnvironmentVariableSnapshot(string Name, string? Value, boo
 /// <param name="IsInternal">Determines if this url is internal.</param>
 public sealed record UrlSnapshot(string Name, string Url, bool IsInternal);
 
+/// <summary>
+/// A snapshot of a volume, mounted to a container.
+/// </summary>
+/// <param name="Source">The name of the volume. Can be <c>null</c> if the mount is an anonymous volume.</param>
+/// <param name="Target">The target of the mount.</param>
+/// <param name="IsReadOnly">Whether the volume mount is read-only or not.</param>
+public sealed record VolumeSnapshot(string? Source, string Target, bool IsReadOnly);
```
mitchdenny, line 91: I'm wondering whether we should capture the volume type here. You should be able to tell intuitively from the source but might be worth making it obvious whether it is a bind mount or a volume mount.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -121,6 +121,22 @@
                               HighlightText="@_filter"
                               GridTemplateColumns="1fr 1.5fr" />
             </FluentAccordionItem>
+            <FluentAccordionItem Heading="@ControlStringsLoc[nameof(ControlsStrings.ResourceDetailsVolumesHeader)]" Expanded="true">
```
JamesNK, line 124: This is only valid for containers, right? I think there should be an if around this section to only show for containers.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
JamesNK, line 91: I know the pattern here is to use records, but I have concerns about using records in public API. Let's say we add new fields in the future, how do we do that with a record?

I think a normal C# type with init properties would be easier to expand in the future.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -117,6 +117,12 @@ message Url {
     bool is_internal = 3;
 }
 
+message Volume {
+    string source = 1;
```
JamesNK, line 121: Add comments to new proto fields.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
drewnoakes: I'll add it to the model. In the UI we currently use name/value pairs in the details view, so won't show it (or `IsReadOnly`) without further work.

drewnoakes: That's fair, but let's do that separately and keep the code consistent here.

davidfowl: Agreed

davidfowl:
@drewnoakes Seems like there's a conflict with @mitchdenny's PR. 

Can you show an example of a bind mount?

This change looks good! Thanks!

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
drewnoakes: I found a way to show it: 

image

The visualizer/copy button works with the raw value only (`/var/lib/rabbitmq`), which I think makes the most sense.

This shows both the mount type, and whether the mount is readonly or not. In this case I hard-coded the value to true, as I didn't see an example in the playground.

drewnoakes:
Conflicts fixed.

Found a way to visualize mount type and read only state:

image

The visualizer/copy button works with the raw value only (`/var/lib/rabbitmq`), which I think makes the most sense.

This shows both the mount type, and whether the mount is read only or not. In this case I hard-coded the value to true, as I didn't see an example in the playground.

davidfowl:
Bind mounts have a physical path for the source

davidfowl (approved):

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -121,6 +121,35 @@
...
+                                  IsNameSortable="false"
+                                  IsValueSortable="false"
+                                  EnableValueMasking="false"
+                                  HighlightText="@_filter"
+                                  GridTemplateColumns="1fr 1.5fr">
+                        <ExtraValueContent>
+                            <div>
+                                @if (context.IsReadOnly)
+                                {
+                                    <FluentIcon Value="@(new Icons.Regular.Size12.LockClosed())" />
```
JamesNK, line 145: The icon is a little small. We use 16x16 icons with inline text in other places. For example, the state icon on the resources screen:

https://github.com/dotnet/aspire/blob/b6e3e16393835fd67ca482da440ac7efb158f8c1/src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor#L14-L17

Can you make this 16x16 and make the icon and text vertically middle aligned with each other?

PR #5586, opened 2024-09-07
Support callback, custom config key and publishing for parameters.

### LINKED ISSUE #4429: AddParameter overloads that supports providing a value

davidebbo:
## Description

See #4429 for details.

Fixes #4429

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [X] No. Follow-up changes expected to support new manifest value
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [X] Yes
      - [X] Yes
      - [ ] No
      - [ ] Yes
      - [X] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [X] Yes
    - Link to aspire-docs issue: none yet
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5586)


mitchdenny:
Any thoughts on supporting a `Func<IServiceProvder, Task>` variant?

mitchdenny (approved):

mitchdenny:
Thanks once again for the contribution @davidebbo

davidebbo:
@mitchdenny per discussion with @davidfowl, we decided not to tackle async at this time, though this could be a future thing (it does look hard!).

Note that per https://github.com/dotnet/aspire/issues/4429#issuecomment-2335197079 and [discord chat](https://discord.com/channels/732297728826277939/1281287950491193436/1282247086397587576), we'll be making some design changes for publishing the value, so I'll be evolving the PR.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -33,16 +33,52 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// <param name="name">Name of parameter resource</param>
+    /// <param name="valueGetter">A callback function that returns the value of the parameter</param>
+    /// <param name="publishValue">Indicates whether the value should be published to the manifest.</param>
+    /// <param name="secret">Optional flag indicating whether the parameter should be regarded as secret.</param>
+    /// <returns>Resource builder for the parameter.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters",
+                                                     Justification = "third parameters are mutually exclusive.")]
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, Func<string> valueGetter, bool publishValue = false, bool secret = false)
+    {
+        var val = valueGetter();
```
davidebbo, line 59: @davidfowl @mitchdenny is this method really meaningful if we are instantly evaluating the callback? How is that any different from users calling the previous overload and evaluating themselves?

And we can't really delay the evaluation, because would happen on the next line anyway.

davidfowl: This should be evaluated in side of the lambda, not immediately.

davidebbo: Right, but my point is that this doesn't actually delay the evaluation, because it'll happen anyway before this method call returns. So what does that actually buy?

See the stack that comes right back to the lambda:

```
Aspire.Hosting.ParameterResourceBuilderExtensions.AddParameter.AnonymousMethod__0(Aspire.Hosting.ApplicationModel.ParameterDefault parameterDefault) Line 62	C#
Aspire.Hosting.ApplicationModel.ParameterResource.Value.get() Line 39	C#
Aspire.Hosting.ParameterResourceBuilderExtensions.AddParameter(Aspire.Hosting.IDistributedApplicationBuilder builder, string name, System.Func<Aspire.Hosting.ApplicationModel.ParameterDefault, string> callback, bool secret, bool connectionString, Aspire.Hosting.ApplicationModel.ParameterDefault parameterDefault, string configurationKey) Line 148	C#
Aspire.Hosting.ParameterResourceBuilderExtensions.AddParameter(Aspire.Hosting.IDistributedApplicationBuilder builder, string name, System.Func<string> valueGetter, bool publishValue, bool secret) Line 61	C#
```

davidfowl: We can change it in the future

davidebbo: Ok, I changed it. It added a little bit of complexity to avoid having the lambda evaluated twice.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -33,16 +33,50 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// </summary>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">Name of parameter resource</param>
+    /// <param name="valueGetter">A callback function that returns the value of the parameter</param>
+    /// <param name="publishValue">Indicates whether the value should be published to the manifest.</param>
+    /// <param name="secret">Optional flag indicating whether the parameter should be regarded as secret.</param>
+    /// <returns>Resource builder for the parameter.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters",
+                                                     Justification = "third parameters are mutually exclusive.")]
+    public static IResourceBuilder<ParameterResource> AddParameter(this IDistributedApplicationBuilder builder, string name, Func<string> valueGetter, bool publishValue = false, bool secret = false)
```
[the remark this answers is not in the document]
davidebbo: Yeah, sounds good. Done.

tests/Aspire.Hosting.Tests/AddParameterTests.cs
```diff
@@ -112,82 +112,142 @@ public void ParametersWithConfigurationValueDoNotGetDefaultValue()
...
+        // The manifest should include the default value, since we passed publishValueAsDefault: true
+        var paramManifest = await ManifestUtils.GetManifest(appModel.Resources.OfType<ParameterResource>().Single(r => r.Name == "pass"));
+        var expectedManifest = $$"""
+            {
+              "type": "parameter.v0",
+              "value": "{pass.inputs.value}",
+              "inputs": {
+                "value": {
+                  "type": "string",
+                  "default": {
```
davidfowl, line 208: @mitchdenny This is new schema for parameters to specify a default value. We should update the schema and file an issue on azd to implement this.

@vhvb1989 Thoughts on implementing this idea of no prompting for parameters with defaults? Or being able to suppress them somehow?

davidebbo: > @vhvb1989 Thoughts on implementing this idea of no prompting for parameters with defaults? Or being able to suppress them somehow?

If it's a 'default', there needs to be a way to override them. Maybe there are two modes (e.g. controlled by switches):
- Use all manifest defaults without prompting
- Prompt, but prefill the field with the manifest default

vhvb1989: FYI, azd suports a `generate` field in `default`, like

image

Azd uses the gereate config to produce the value instead of prompt.  If you need a static default value, we could add a rule to the generate config to just use the provided value

@eerhardt 

When we added support for this, we considered a case where azd could ask forlk if they wanted to auto-generate the value (based on the rules) or enter a value manually.  This is supported by azd already (just not mapped to any configuration from the Aspire Manifest).

davidebbo: To me, it feels a little strange to have this in `generate` (because there is no generation), which is why I went with an alternative `value` property, e.g.

```json
              "inputs": {
                "value": {
                  "type": "string",
                  "default": {
                    "value": "DefaultValue"
                  }
                }
              }
```

Either way, whoever owns the aspire manifest schema should decide where to put that. If we want it under `generate` that's easy to do.

davidfowl: I agree that generate is strange. I like the above proposal of having a "default" section.

mitchdenny: The idea behind `generate` was that you could have either `value` or `generate`. If generate was present, then the fields in that sub-object would control the generation logic. So your proposal makes sense.

davidebbo: Great, so then we are all good with the proposed schema change as it is in this PR? If so, what is the next step to get this in?

davidfowl: I'll do a final pass

davidebbo: Also, should I be modifying the schema file accordingly? I've never touched one of those, but I may as well learn how to do it!

davidfowl: Yep

davidebbo: Ok, I pushed the schema change. I also added test coverage for "generate", as it was missing. @mitchdenny PTAL

davidebbo: @mitchdenny Also, how does the schema make its way from this repo to https://github.com/dotnet/dotnet/blob/main/src/aspire/src/Schema/aspire-8.0.json? Automatic or we need to push it?

davidfowl: Manually I think

mitchdenny: Yeah all manual.

davidfowl (approved):

davidfowl:
cc @vhvb1989 @ellismg

davidfowl:
Filed https://github.com/Azure/azure-dev/issues/4359

vhvb1989:
@davidebbo @davidfowl @mitchdenny is it safe to asume that a `default.value` will never be set as `secure` ?, like:

```json
"param-with-value": {
      "type": "parameter.v0",
      "value": "{param-with-value.inputs.value}",
      "inputs": {
        "value": {
          "type": "string",
          "secret": true,
          "default":{
            "value": "default value for param"
          }
        }
      }
    },
```

See the `"secret": true,`  there.

I am working on making `azd` to fail on this case as per: https://learn.microsoft.com/azure/azure-resource-manager/bicep/linter-rule-secure-parameter-default

image

But I am wondering if the AppHost is even allowing this or not?  (maybe is blocked in AppHost so I don't even need to worry)?

davidebbo:
@vhvb1989 we do not block it today, but I agree that we should. So I think we should allow:
```c#
var foo = builder.AddParameter("foo", value: "somesecret", secret: true);
```

as this only uses the value at runtime, but not:
```c#
var foo = builder.AddParameter("foo", value: "somesecret", publishValueAsDefault: true, secret: true);
```

which results in having the secret in the manifest.

@davidfowl @mitchdenny do you agree? If so, can make a follow-up change to prevent this. e.g. an error like `the 'publishValueAsDefault' and 'secret' flags are mutually exclusive, to prevent saving a secret to the manifest`

davidebbo:
I sent #5891 to make this change.

davidebbo:
@vhvb1989 This is now blocked in AppHost, so you should not encounter this unless people mess with their manifest by hand,

PR #5588, opened 2024-09-08
Handle parsing wildcard urls in launch profiles

### LINKED ISSUE #5587: AppHost project crashes on startup: System.UriFormatException: Invalid URI: The hostname could not be parsed.

### Describe the bug

Hey, just trying to integrate Aspire into my existing solution. The first time I tried it, it threw this on startup:
```
System.UriFormatException: Invalid URI: The hostname could not be parsed.
```
I eventually narrowed this down to using an `applicationUrl` of `"http://*:80"` in my web API's `launchSettings.json`. I had this set to listen on all IP addresses as I need to connect to my web API from my smartphone.

### Expected Behavior

This is my first time using Aspire so I don't have the context to know why it reads `launchSettings`, but perhaps these URLs could be skipped. I think in any case the error message could be improved, as I only narrowed it to `launchSettings` by checking the Aspire source code.

### Steps To Reproduce

1. Create new Web API project
2. Set `applicationUrl` to `http://*:80` in the web API's `launchSettings.json`:

```json
{
    "$schema": "https://json.schemastore.org/launchsettings.json",
    "profiles": {
        "DragaliaAPI": {
            "commandName": "Project",
            "environmentVariables": {
            },
            "applicationUrl": "http://*:80"
        }
    }
}
```

3. Create new Aspire app host with this configuration:

```cs

builder.AddProject<Projects.DragaliaAPI>("dragaliaapi");

builder.Build().Run();
```

4. Crash

### Exceptions (if any)

```
Unhandled exception. System.UriFormatException: Invalid URI: The hostname could not be parsed.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind, UriCreationOptions& creationOptions)
   at System.Uri..ctor(String uriString)
   at Aspire.Hosting.ProjectResourceBuilderExtensions.WithProjectDefaults(IResourceBuilder`1 builder, ProjectResourceOptions options) in /_/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs:line 407
   at Aspire.Hosting.ProjectResourceBuilderExtensions.AddProject[TProject](IDistributedApplicationBuilder builder, String name, Action`1 configure) in /_/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs:line 216
   at Aspire.Hosting.ProjectResourceBuilderExtensions.AddProject[TProject](IDistributedApplicationBuilder builder, String name) in /_/src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs:line 63
   at Program.<Main>$(String[] args) in C:\Users\jay0\Projects\Dawnshard\AppHost\Program.cs:line 3
```

### .NET Version info

```
.NET SDK:
 Version:           8.0.401
 Commit:            811edcc344
 Workload version:  8.0.400-manifests.57f7c351
 MSBuild version:   17.11.4+37eb419ad
```

### Anything else?

`Aspire.Hosting.AppHost` v8.2.0.

davidfowl:
## Description

- As part of this change expose TargetHostAddress on EndpointAnnotation which describes the original address that was bound to in both the kestrel endpoint list and the launch profile's application urls.
- Added a test

Fixes #5587

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5588)


davidfowl:
cc @davidebbo

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -104,6 +105,11 @@ public int? TargetPort
     /// </summary>
     public string UriScheme { get; set; }
 
+    /// <summary>
+    /// This is the address the resource is listening on. By default it is localhost.
+    /// </summary>
+    public string TargetHostAddress { get; set; } = "localhost";
```
davidfowl, line 111: New public API. I could keep internal, but it felt better to expose this in WithEndpoint callbacks. This is to address as targetport is to port.

mitchdenny:
What impact does this have around requests that are proxied via DCP? I would have thought that for this to be useful the proxy would need to pass through a host header?

tests/Aspire.Hosting.Tests/ProjectResourceTests.cs
```diff
@@ -590,6 +590,34 @@ public async Task AddProjectWithArgs()
...
+            {
+                e.AllocatedEndpoint = new(e, "localhost", e.Port!.Value, targetPortExpression: "p1");
+            });
+
+        using var app = appBuilder.Build();
+
+        var appModel = app.Services.GetRequiredService<DistributedApplicationModel>();
+        var projectResources = appModel.GetProjectResources();
+
+        var resource = Assert.Single(projectResources);
```
davidebbo, line 613: Should exercise new prop, e.g.

```c#
        foreach (var annotation in resource.Annotations.OfType<EndpointAnnotation>())
        {
            Assert.Equal("*", annotation.TargetHostAddress);
        }
```

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
davidebbo: Would `TargetAddressHost` be a more correct name? Or just `TargetHost`? If we go with the `BindingAddress` terminology, 'address' represents everything, and the host is just the one part that's the host name or IP address.

davidfowl:
> What impact does this have around requests that are proxied via DCP? I would have thought that for this to be useful the proxy would need to pass through a host header?

Doesn't affect it. All we are doing is preserving the input. When you bind to * or 0.0.0.0, requests from localhost still work.

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
davidfowl: I like TargetHost.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -931,7 +931,7 @@ private void AddAllocatedEndpointInfo(IEnumerable<AppResource> resources)
 
                 sp.EndpointAnnotation.AllocatedEndpoint = new AllocatedEndpoint(
                     sp.EndpointAnnotation,
-                    sp.EndpointAnnotation.IsProxied ? svc.AllocatedAddress! : "localhost",
+                    "localhost",
```
karolz-ms, line 934: The reason why `AllocatedAddress` is used here is because if one requests `localhost:1234`, but 1234 is already taken on one of the addresses behind `localhost` name (dual-stack machine), DCP will bind to specific IP address (IPv4 or IPv6) to satisfy the requested port. It will then return the specific IP address that was used.

If the client uses `localhost` to connect in that case, this will still work, but may be quite slow as the client tries the "wrong" address several times before switching to the other one.

I am not sure if you need to change this line as part of this change

davidfowl: Will do

davidfowl: This doesn't work because we use this allocated endpoint address to communicate with these endpoints. It seems like dcp returns 0.0.0.0 instead of localhost, 127.0.0.1 or ::1.

karolz-ms: Returns 0.0.0.0 for what kind of data in EndpointAnnotation?

davidfowl: It's less about what it returns and more about how what is returned gets used. This is the "client" side of the service and we hand this out for other APIs to connect to. Much like port = client port and target port = listen port. It seems like we need that split for address as well.

karolz-ms: Do we need changes or new tests in https://github.com/dotnet/aspire/blob/main/tests/Aspire.Hosting.Tests/Dcp/ApplicationExecutorTests.cs?

karolz-ms: > It's less about what it returns and more about how what is returned gets used. This is the "client" side of the service and we hand this out for other APIs to connect to. Much like port = client port and target port = listen port. It seems like we need that split for address as well.

Not sure I follow. The localhost gets special treatment, but nothing else does. Meaning, if you tell DCP something is listening on address other than localhost (including 0.0.0.0 pseudo-address), it will use that address for the proxy, and this is the address the clients should be using too. So I am not sure why we need to force localhost here.

karolz-ms (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: Chatted with @davidfowl and understand now, 0.0.0.0, as-is, is not useable by clients. Will need to get that fixed on DCP side.

Issue #3101, opened 2024-03-22
Subsequent provisioning of Azure Sql Server fails


miceiken:
Seemingly out of the blue, our pipeline provisioning started failing:
```
  (✓) Done: Resource group: rg-snip
  (✓) Done: Log Analytics workspace: law-snip
  (✓) Done: Container Apps Environment: cae-snip
  (✓) Done: Container Registry: acrsnip
  (✓) Done: Container App: snip
  (✓) Done: Key Vault: secretssnip
  (✓) Done: Application Insights: appinsights-snip
  (x) Failed: Azure SQL Server: sqlserver-snip

ERROR: deployment failed: failing invoking action 'provision', error deploying infrastructure: deploying to subscription:

Deployment Error Details:
InvalidParameterValue: Invalid value given for parameter ExternalAdministratorLoginSid. Specify a valid parameter value.
```

We did try changing the AD admin for the SQL server to a group (with the managed identity as a member) in Azure Portal, but even when changing it back to the managed identity it still fails.

I think it is related to:
https://github.com/Azure/bicep-types-az/issues/2320

https://learn.microsoft.com/en-us/azure/templates/microsoft.sql/2022-05-01-preview/servers?pivots=deployment-language-bicep
> The Azure Active Directory administrator of the server. This can only be used at server create time. If used for server update, it will be ignored or it will result in an error. For updates individual APIs will need to be used

davidfowl:
What version of aspire is this?

miceiken:
The Aspire Preview 4 release

davidfowl:
Can you file the issue here https://github.com/Azure/azure-dev/issues

miceiken:
Done. It seems to me from my rudimentary understanding that the Bicep file for Azure Sql Resource should be amended to use a `Microsoft.Sql/servers/administrators` (and AD) resource, similiar to this comment: https://github.com/Azure/bicep/issues/4988#issuecomment-1833841902

Looking at the Aspire codebase now however, it seems like provisioning in P5 and on will use CDK instead of Bicep templates?

davidfowl:
cc @mitchdenny

mitchdenny:
Hrm ... looks like we will need to get the child resource added to the CDK in order to use this.

mitchdenny:
cc @JoshLove-msft

joperezr:
This is waiting on a change from CDK.

miceiken:
[...]

PR #5592, opened 2024-09-09
Update Azure.Provisioning to latest version

### LINKED ISSUE #5583: Integrate new version of Azure.Provisioning
There are large breaking changes we need to adapt. See https://github.com/Azure/azure-sdk-for-net/pull/45786 for details

eerhardt:
Update to the new Azure.Provisioning libraries.

Fix #5583

cc @mitchdenny @davidfowl @tg-msft @JoshLove-msft 

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5592)


src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
```diff
@@ -32,23 +34,42 @@ public static IResourceBuilder<AzureKeyVaultResource> AddAzureKeyVault(this IDis
...
+
+            var role = KeyVaultBuiltInRole.KeyVaultAdministrator;
+            var roleAssignment = new RoleAssignment($"{KeyVaultBuiltInRole.GetBuiltInRoleName(role)}_{keyVault.ResourceName}")
+            {
+                Scope = new IdentifierExpression(keyVault.ResourceName),
+                PrincipalType = construct.PrincipalTypeParameter,
+                RoleDefinitionId = BicepFunction.GetSubscriptionResourceId("Microsoft.Authorization/roleDefinitions", role.ToString()),
+                PrincipalId = construct.PrincipalIdParameter
+            };
+            construct.Add(roleAssignment);
```
davidfowl, line 72: Is this really the new API? These are fine low level building blocks bug oh man...

playground/cdk/CdkSample.AppHost/mykv.module.bicep
```diff
@@ -1,51 +1,32 @@
...
-  name: 'mysecret'
-  location: location
-  properties: {
-    value: signaturesecret
-  }
-}
+param principalId string
 
-output vaultUri string = keyVault_aMZbuK3Sy.properties.vaultUri
+param principalType string
```
davidfowl, line 32: I guess the order rendered is the order added to the construct. I think we should ask to have that fixed in the CDK itself. Parameters should be first.

playground/cdk/CdkSample.AppHost/mykv.module.bicep
```diff
@@ -1,51 +1,32 @@
...
 
-resource keyVaultSecret_mW5tlkNij 'Microsoft.KeyVault/vaults/secrets@2022-07-01' = {
-  parent: keyVault_aMZbuK3Sy
-  name: 'mysecret'
-  location: location
-  properties: {
-    value: signaturesecret
-  }
-}
+param principalId string
```
davidfowl, line 30: The location parameter is missing, we need to add that everywhere.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
eerhardt: There is an `AssignRole` API, but it assumes a user assigned identity.

https://github.com/Azure/azure-sdk-for-net/blob/ade2cee32561d57a3049d584b17ab749a0bb64c7/sdk/provisioning/Azure.Provisioning.KeyVault/src/Generated/KeyVaultService.cs#L157-L171

```C#
    /// <summary>
    /// Assign a role to a user-assigned identity that grants access to this
    /// KeyVaultService.
    /// </summary>
    /// <param name="role">The role to grant.</param>
    /// <param name="identity">The <see cref="UserAssignedIdentity"/>.</param>
    /// <returns>The <see cref="RoleAssignment"/>.</returns>
    public RoleAssignment AssignRole(KeyVaultBuiltInRole role, UserAssignedIdentity identity) =>
        new($"{identity.ResourceName}_{KeyVaultBuiltInRole.GetBuiltInRoleName(role)}_{ResourceName}")
        {
            Scope = new IdentifierExpression(ResourceName),
            PrincipalType = RoleManagementPrincipalType.ServicePrincipal,
            RoleDefinitionId = BicepFunction.GetSubscriptionResourceId("Microsoft.Authorization/roleDefinitions", role.ToString()),
            PrincipalId = identity.PrincipalId
        };
```

Here we take in a `PrincipleType` as a parameter, so this higher level API doesn't seem to meet our needs.

tg-msft: Yes - this is a known point in time thing.  We're starting with the lowest level layer to make sure you're able to do anything.  We want to add more conveniences that provide reasonable starting points over time kind of like https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/provisioning/Azure.Provisioning.Storage/src/StorageResources.cs#L35

playground/cdk/CdkSample.AppHost/mykv.module.bicep
tg-msft: This is tied up with auto module splitting, but we eventually want it to be customizable.

playground/cdk/CdkSample.AppHost/mykv.module.bicep
tg-msft: It's a known issue.  You're getting your values from https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/provisioning/Azure.Provisioning/src/Primitives/LocationPropertyResolver.cs and we need to update that to inject/reuse a parameter if available.  You can explicitly work around this yourself by setting `Location` or wait for it to be fixed in Azure.Provisioning.

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
```diff
@@ -43,44 +44,49 @@ public static IResourceBuilder<AzureStorageResource> AddAzureStorage(this IDistr
...
-            var queueRole = storageAccount.AssignRole(RoleDefinition.StorageQueueDataContributor);
-            queueRole.AssignProperty(p => p.PrincipalId, construct.PrincipalIdParameter);
-            queueRole.AssignProperty(p => p.PrincipalType, construct.PrincipalTypeParameter);
-
-            storageAccount.AddOutput("blobEndpoint", sa => sa.PrimaryEndpoints.BlobUri);
-            storageAccount.AddOutput("queueEndpoint", sa => sa.PrimaryEndpoints.QueueUri);
-            storageAccount.AddOutput("tableEndpoint", sa => sa.PrimaryEndpoints.TableUri);
+            var storageAccount = new StorageAccount(name)
+            {
+                Location = construct.AddLocationParameter(),
```
davidfowl, line 49: Why `AddLocationParameter`? There should always be a location parameter. Make it a property?

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
```diff
@@ -43,44 +44,49 @@ public static IResourceBuilder<AzureStorageResource> AddAzureStorage(this IDistr
...
+                    RoleDefinitionId = BicepFunction.GetSubscriptionResourceId("Microsoft.Authorization/roleDefinitions", role.ToString()),
+                    PrincipalId = construct.PrincipalIdParameter
+                });
+            }
+
+            AssignRole(StorageBuiltInRole.StorageBlobDataContributor);
+            AssignRole(StorageBuiltInRole.StorageTableDataContributor);
+            AssignRole(StorageBuiltInRole.StorageQueueDataContributor);
+
+            construct.Add(new BicepOutput("blobEndpoint", typeof(string)) { Value = storageAccount.PrimaryEndpoints.Value!.BlobUri });
```
davidfowl, line 87: @tg-msft why is .Value nullable?

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
```diff
@@ -32,23 +34,52 @@ public static IResourceBuilder<AzureKeyVaultResource> AddAzureKeyVault(this IDis
...
         var configureConstruct = (ResourceModuleConstruct construct) =>
         {
-            var keyVault = construct.AddKeyVault(name: construct.Resource.Name);
-            keyVault.AddOutput("vaultUri", x => x.Properties.VaultUri);
+            var keyVault = new KeyVaultService(construct.Resource.Name)
+            {
+                Location = construct.AddLocationParameter(),
+                Properties = new KeyVaultProperties()
+                {
+                    TenantId = new MemberExpression(new FunctionCallExpression(new IdentifierExpression("tenant")), "tenantId"),
```
davidfowl, line 51: Shouldn't this be a built-in bicep function?

@tg-msft What does this turn into?

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
```diff
@@ -32,23 +34,52 @@ public static IResourceBuilder<AzureKeyVaultResource> AddAzureKeyVault(this IDis
...
 
-            keyVault.Properties.Tags["aspire-resource-name"] = construct.Resource.Name;
+            construct.Add(new BicepOutput("vaultUri", typeof(string))
+            {
+                Value =
+                    new MemberExpression(
+                        new MemberExpression(
+                            new IdentifierExpression(keyVault.ResourceName),
+                            "properties"),
+                        "vaultUri")
```
davidfowl, line 69: Why doesn't storage look like this?

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
```diff
@@ -32,23 +34,52 @@ public static IResourceBuilder<AzureKeyVaultResource> AddAzureKeyVault(this IDis
...
+            });
 
-            var keyVaultAdministratorRoleAssignment = keyVault.AssignRole(RoleDefinition.KeyVaultAdministrator);
-            keyVaultAdministratorRoleAssignment.AssignProperty(x => x.PrincipalId, construct.PrincipalIdParameter);
-            keyVaultAdministratorRoleAssignment.AssignProperty(x => x.PrincipalType, construct.PrincipalTypeParameter);
+            keyVault.Tags["aspire-resource-name"] = construct.Resource.Name;
+
+            var role = KeyVaultBuiltInRole.KeyVaultAdministrator;
+            var roleAssignment = new RoleAssignment($"{KeyVaultBuiltInRole.GetBuiltInRoleName(role)}_{keyVault.ResourceName}")
```
davidfowl, line 75: Does GetBuiltInRole have an interface (static interface method). We should be able to make this method generic.

playground/cdk/CdkSample.AppHost/storage.module.bicep
```diff
@@ -1,77 +1,69 @@
...
+
+output tableEndpoint string = storage.properties.primaryEndpoints.table
+
+param storagesku string
+
+param locationOverride string
+
+param principalId string
+
+param principalType string
```
davidfowl, line 69: Can you fix it so these are first? This is so hard to look at 😄

playground/cdk/CdkSample.AppHost/storage.module.bicep
```diff
@@ -1,77 +1,69 @@
...
-resource roleAssignment_HRj6MDafS 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
-  scope: storageAccount_1XR3Um8QY
-  name: guid(storageAccount_1XR3Um8QY.id, principalId, subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '0a9a7e1f-b9d0-4cc4-a60d-0319b160aaa3'))
-  properties: {
-    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '0a9a7e1f-b9d0-4cc4-a60d-0319b160aaa3')
-    principalId: principalId
-    principalType: principalType
-  }
+resource StorageTableDataContributor_storage 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+    name: guid(resourceGroup().id, 'StorageTableDataContributor_storage')
```
davidfowl, line 38: This doesn't look unique enough. Can we do what we had before?

```
  name: guid(keyVault_aMZbuK3Sy.id, principalId, subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '00482a5a-887f-4fb3-b363-3b7fe8e74483'))
```

playground/cdk/CdkSample.AppHost/storage.module.bicep
eerhardt: yeah, it's a pain to reorder everything manually every time I want to diff the before/after.

@tg-msft - is the order in the file based on the order in the collection in memory? I can try to reorder things to make all parameters first.

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
eerhardt: I have this comment on the method:

https://github.com/dotnet/aspire/blob/1e2333561f2a387592d72c2d5896c7465cef9f1f/src/Aspire.Hosting.Azure/AzureConstructResource.cs#L150-L156

@tg-msft and I talked about this should be built into the CDK - add a location parameter if the Location isn't set.

playground/cdk/CdkSample.AppHost/storage.module.bicep
davidfowl: We want to.

eerhardt: This is now done. The parameters aren't always exactly in the same order, but all params are first and all outputs are last. It makes the diff a lot easier.

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -17,6 +18,10 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class AzurePostgresExtensions
 {
+    private const string ServerResourceVersion = "2023-03-01-preview";
```
eerhardt, line 21: @davidfowl @mitchdenny @tg-msft - how should we think about resource versions? Should we always hard-code the version we expect, and then periodically update to new versions on some sort of cadence? Similar to our other dependency versions.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
tg-msft: While not solving the larger problem, I added an `AssignRole(role, principalType, principalId)` overload in https://github.com/Azure/azure-sdk-for-net/pull/45911

playground/cdk/CdkSample.AppHost/mykv.module.bicep
tg-msft: Moved params to the top and outputs to the bottom in https://github.com/Azure/azure-sdk-for-net/pull/45911

playground/cdk/CdkSample.AppHost/mykv.module.bicep
tg-msft: Switched to using a param in https://github.com/Azure/azure-sdk-for-net/pull/45911

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
tg-msft: I think this can be removed once the Dev Feed gets updated.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
tg-msft: Yes, this turns into `tenant().tenantid`.  We just haven't added it to `BicepFunction` yet.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
tg-msft: You should be able to punt on this with the new generated `AssignRole` overloads.

playground/cdk/CdkSample.AppHost/storage.module.bicep
tg-msft: I added back part of module splitting to sort these in https://github.com/Azure/azure-sdk-for-net/pull/45911.

playground/cdk/CdkSample.AppHost/storage.module.bicep
tg-msft: The default value comes from https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/provisioning/Azure.Provisioning/src/Authorization/RoleAssignment.cs#L15 where we don't have enough information to create that value.  You'd need to explicitly set `Name` in your code.  `BicepFunction.CreateGuid` takes `params` so you should be be able to do it without dropping down to expressions.

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
tg-msft: At the moment I have everything default to "latest stable" in `Azure.Provisioning.*` (though I'm pulling the info from ARM so it's not perfect yet).  The long term plan is to have `Azure.Provisioning.*` automatically release whenever the management plane libraries are updated so the "latest stable" will change [seemingly at random.](https://www.nuget.org/packages?q=Azure.ResourceManager.)

As a general rule, moving to newer service versions is goodness for customers but there are occasionally breaking changes.  We could float on latest by default and allow conservative customers to pin, or we could pin and have to manually test/bump the world a couple times a year.  We might want to talk offline with Jeff Richter and Mike Kistler from Azure's Breaking Change Review Board for their advice on how dangerous it would be for customers to float on latest.

src/Aspire.Hosting.Azure.AppConfiguration/AzureAppConfigurationExtensions.cs
```diff
@@ -14,6 +14,8 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class AzureAppConfigurationExtensions
 {
+    private const string StoreResourceVersion = "2023-03-01";
```
eerhardt, line 17: TODO: We should be defining our versions everywhere.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
eerhardt: Fixed.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
eerhardt: > You should be able to punt on this with the new generated AssignRole overloads.

Yep, for all except WebPubSub, which doesn't have a BuiltInRole.

src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs
eerhardt: Yep this is now gone.

We always generate a Location parameter up front in the Infrastructure. And don't explicitly set the `.Location` property on any resource. The CDK's location resolver then wires up the parameter to the parameter for any unset properties.

eerhardt:
This should now be ready for a review from folks. Tests are updated. I verified the bicep changes as much as I could. The `cdk` playground app can be deployed to Azure.

src/Aspire.Hosting.Azure/AzureConstructResource.cs
```diff
@@ -119,78 +120,129 @@ public static IResourceBuilder<T> ConfigureConstruct<T>(this IResourceBuilder<T>
     }
 
     /// <summary>
-    /// Assigns an Aspire parameter resource to an Azure construct resource.
+    /// TODO
```
eerhardt, line 123: The XML docs in this file need to be filled out.

src/Aspire.Hosting.Azure/AzureConstructResource.cs
```diff
@@ -119,78 +120,129 @@ public static IResourceBuilder<T> ConfigureConstruct<T>(this IResourceBuilder<T>
...
     /// <typeparam name="T">Type of the CDK resource.</typeparam>
     /// <param name="resource">The CDK resource.</param>
     /// <param name="propertySelector">Property selection expression.</param>
+    /// <param name="parameterResourceBuilder">Aspire parameter resource builder.</param>
     /// <param name="parameterName">The name of the parameter to be assigned.</param>
-    /// <param name="outputReference">Aspire parameter resource builder.</param>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
-    public static void AssignProperty<T>(this Resource<T> resource, Expression<Func<T, object?>> propertySelector, BicepOutputReference outputReference, string? parameterName = null) where T : notnull
+    [Obsolete("This API is no longer supported and always throws.")]
+    public static void AssignProperty<T>(this global::Azure.Provisioning.Primitives.Resource resource, System.Linq.Expressions.Expression<Func<T, object?>> propertySelector, IResourceBuilder<ParameterResource> parameterResourceBuilder, string? parameterName = null) where T : notnull
```
eerhardt, line 191: @mitchdenny @davidfowl - what do you think about this? It is trying to keep the `AssignProperty` method with an obsoletion so any callers of it get a warning that it is no longer supported.

It may not be worth it, and maybe we should just delete it?

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
eerhardt: I've decided to define these for all resources and put them in a shared file. Some things, like KeyVault, need to be used across many libraries. Thoughts on this approach?

Note I still need to ensure all resources use this pattern. Right now EventHubs doesn't.

playground/cdk/CdkSample.AppHost/storage.module.bicep
eerhardt: This is now fixed with the latest build.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
davidfowl: Thanks @tg-msft !

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
davidfowl: I prefer updating the library and letting that determine the version of the default azure resource. If we don't do that then this will be just like container image tags except it'll be managed in both the Azure.Provisioning libraries and in aspire.

playground/cdk/CdkSample.AppHost/Program.cs
```diff
@@ -45,20 +51,24 @@
     .AddQueue("queue1",
         (_, construct, queue) =>
         {
-            queue.Properties.MaxDeliveryCount = 5;
-            queue.Properties.LockDuration = TimeSpan.FromMinutes(5);
+            queue.MaxDeliveryCount = 5;
+            queue.LockDuration = new StringLiteral("PT5M");
+            // TODO: this should be
+            // queue.LockDuration = TimeSpan.FromMinutes(5);
```
davidfowl, line 57: @tg-msft

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
eerhardt: I would prefer that too, but there are problems with that approach.

1. Not all resources have a default version in the CDK right now. Especially "child" resources. For those we need to provide a version, or else we get errors that it is unset.
2. The default version for some things don't actually work. For example, the CDK doesn't emit an empty "properties" objects, but the default version for ServiceBus topics (I think it was this one) says that you must have a `properties` object. I can look up the exact error, if someone cares.
3. The concern that updating to a new version will break something. @tg-msft assures me that there are breaking changes that happen from time to time. So if we just take the version that the CDK uses, and someone updates their CDK to a new version, with new default resource versions, there is a potential for our bicep to be broken.

eerhardt:
Tests are broken becuase I need to update the test nuget.config file for the new feed.

@radical - is it possible to not have the duplicate test nuget.config?

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
```diff
@@ -21,41 +23,61 @@ internal static IResourceBuilder<SqlServerServerResource> PublishAsAzureSqlDatab
...
+                Tags = { { "aspire-resource-name", construct.Resource.Name } }
+            };
+            construct.Add(sqlServer);
+
+            construct.Add(new SqlFirewallRule("sqlFirewallRule_AllowAllAzureIps", AzureResourceVersions.SqlFirewallRuleResourceVersion)
+            {
+                Parent = sqlServer,
+                Name = "AllowAllAzureIps",
+                StartIPAddress = new StringLiteral("0.0.0.0"),
+                EndIPAddress = new StringLiteral("0.0.0.0")
```
davidfowl, line 48: @tg-msft is there a benefit to StringLiteral? Should there be an implicit conversion from strin?

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
tg-msft: I just need to add some metadata to the spec (i.e., like [this example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/provisioning/Generator/src/Specifications/SignalRSpecification.cs#L29-L35)) but didn't see anything documented at https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles other than `SignalR/Web PubSub Contributor`.  I was going to follow-up with Liangying, but David beat me to it here.  😆  Are https://github.com/microsoft/azure-roles/blob/main/azure_roles.json#L585-L586 sufficient to add?

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
tg-msft: 1 is a bug that will be fixed by either generating the same default as parent resources or using a resolver to copy those.  There's still a bit to design around the versioning story.  I'd like to to have one list of versions per library, but there are some challenges there with ARM.

Let's follow up on 2.  I currently try to omit as many empty objects as possible (i.e., no reason to serialize the entire object graph with empty braces everywhere) but we can design a way to include specific empty objects where needed if using a `new ObjectExpression()` isn't already a sufficient workaround.

playground/cdk/CdkSample.AppHost/Program.cs
tg-msft: It's a known issue - SB uses ISO8601 durations _(that are no longer allowed in Azure, save for recurrences or human-centric time intervals, because no languages support them well)_ and we'll need to carry along a format to serialize TimeSpans.

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
tg-msft: There is an implicit conversion from `T` - you should just be able to say `StartIPAddress = "0.0.0.0"`.

src/Aspire.Hosting.Azure.KeyVault/AzureKeyVaultResourceExtensions.cs
vicancy: How about https://github.com/Azure/azure-sdk-for-net/pull/45977 ? For some reason that the team is now investigating, the builtin roles for Web PubSub are listed in Preview category somehow, so that they are not seen in the documentation.

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
```diff
@@ -642,36 +613,29 @@ public async Task AddApplicationInsightsWithExplicitLawArgumentDoesntGetDefaultP
...
-              tags: {
-                'aspire-resource-name': 'appInsights'
-              }
-              kind: kind
-              properties: {
-                Application_Type: applicationType
-                WorkspaceResourceId: logAnalyticsWorkspaceId
-              }
+            resource appInsights 'Microsoft.Insights/components@2020-02-02' = {
+                name: take('appInsights-${uniqueString(resourceGroup().id)}', 260)
```
davidfowl, line 626: triggered, 2 tabs vs 4. What's the idiomatic bicep way?

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
eerhardt: For Sql, I can change this to the simple `StartIPAddress = "0.0.0.0"`. But that doesn't work for PostgreSQL because Start/End IPAddress are typed as `public BicepValue<IPAddress> StartIPAddress`. 

image

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
eerhardt: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/file uses 2 spaces.

src/Aspire.Hosting.Azure/AzureConstructResource.cs
eerhardt: I'm gong to remove this because it isn't the exact same API. The exisitng API can't be built because it is based on a Type that no longer exists in the new CDK. So I think we should remove these APIs altogether.

eerhardt:
OK, I think this PR is ready for final review and can be merged to unblock other work.

There are follow up issues on the CDK we can make later.

1. Default ResourceVersions for all resources.
2. TimeSpan formatting for ServiceBus
3. WebPubSub roles
4. `.Value!` fix ups and using Expressions as workarounds.

Directory.Packages.props
```diff
@@ -36,23 +37,24 @@
...
-    <PackageVersion Include="Azure.Provisioning.Resources" Version="0.2.0" />
-    <PackageVersion Include="Azure.Provisioning.Search" Version="0.1.0" />
-    <PackageVersion Include="Azure.Provisioning.ServiceBus" Version="0.1.0" />
-    <PackageVersion Include="Azure.Provisioning.SignalR" Version="0.2.0" />
-    <PackageVersion Include="Azure.Provisioning.Sql" Version="0.1.0" />
-    <PackageVersion Include="Azure.Provisioning.Storage" Version="0.2.0" />
-    <PackageVersion Include="Azure.Provisioning.WebPubSub" Version="0.1.0-beta.1" />
+    <PackageVersion Include="Azure.Provisioning" Version="$(AzureProvisiongVersion)" />
+    <PackageVersion Include="Azure.Provisioning.AppConfiguration" Version="$(AzureProvisiongVersion)" />
+    <PackageVersion Include="Azure.Provisioning.ApplicationInsights" Version="1.0.0-alpha.20240912.1" />
```
mitchdenny, line 42: What is happening with this and operational insights?

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
```diff
@@ -1,70 +1,66 @@
...
-  sku: {
-    name: 'Standard_GRS'
-  }
-  kind: 'StorageV2'
-  properties: {
-    accessTier: 'Hot'
-    networkAcls: {
-      defaultAction: 'Allow'
+resource ehstorage 'Microsoft.Storage/storageAccounts@2023-01-01' = {
+    name: take('ehstorage${uniqueString(resourceGroup().id)}', 24)
```
mitchdenny, line 9: You will need `toLower()` here.

mitchdenny: I'm wondering if this is the delta from main or whether this file wasn't regenerated previously so its showing more differences than I'm expecting (e.g. accessTier, allowSharedKeyAccess).

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
```diff
@@ -1,47 +1,36 @@
...
-  tags: {
-    'aspire-resource-name': 'eventhubns'
-  }
-  sku: {
-    name: sku
-  }
-  properties: {
-  }
+resource eventhubns 'Microsoft.EventHub/namespaces@2017-04-01' = {
+    name: take('eventhubns-${uniqueString(resourceGroup().id)}', 256)
```
mitchdenny, line 11: Needs `toLower()`

Directory.Packages.props
eerhardt: The pipeline was split into 2 becuase they couldn't publish all the packages at once. So they get different versions (for now).

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
eerhardt: For `allowSharedKeyAccess`, the bicep is out of date in main. See

https://github.com/dotnet/aspire/pull/4744


I'm not sure about `accessTier`. I'll look to see if the old CDK defaulted that and if we should be setting it by default.

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
eerhardt: @tg-msft told me it wasn't necessary because `eventhubns-` is all lower, and `uniqueString` produces a lowered string.

playground/SqlServerEndToEnd/SqlServerEndToEnd.AppHost/sql1.module.bicep
```diff
@@ -1,49 +1,43 @@
...
-      login: principalName
-      sid: principalId
-      tenantId: subscription().tenantId
-      azureADOnlyAuthentication: true
+resource sql1 'Microsoft.Sql/servers@2021-11-01' = {
+    name: take('sql1-${uniqueString(resourceGroup().id)}', 63)
+    location: location
+    properties: {
+        administrators: {
+            administratorType: 'ActiveDirectory'
```
mitchdenny, line 13: I think we should switch over to using AAD administrators, this is the old way of specifying AAD access. Its caused some problems but we couldn't do anything about it until now.

mitchdenny: https://learn.microsoft.com/en-us/azure/templates/microsoft.sql/servers/administrators?pivots=deployment-language-bicep#resource-format

mitchdenny: Related: https://github.com/dotnet/aspire/issues/3101

playground/bicep/BicepSample.AppHost/redis.module.bicep
```diff
@@ -1,38 +1,33 @@
...
+@description('The location for the resource(s) to be deployed.')
 param location string = resourceGroup().location
 
-@description('')
 param keyVaultName string
 
-
-resource keyVault_IeF8jZvXV 'Microsoft.KeyVault/vaults@2022-07-01' existing = {
-  name: keyVaultName
+resource keyVault 'Microsoft.KeyVault/vaults@2019-09-01' existing = {
```
mitchdenny, line 6: Why the downgrade in API version (surprising) /cc @tg-msft

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
mitchdenny: The resource name in this instance is going to be different. You are going from this:

```
toLower(take('eventhubns${uniqueString(resourceGroup().id)}', 24))
```

to this:

```
take('eventhubns-${uniqueString(resourceGroup().id)}', 256)
```

If you take the same code from Aspire 8.0 to Aspire 9.0 you are going to end up with an extra resource being provisioned. We need to be 100% sure the generated resource names align between Aspire 8.0 and Aspire 9.0, otherwise we are going to break people who have deployed to production.

In this particular case its the `-` between `eventhubns` and the generated unique string which will be a problem.

In the case of storage, the resource name MUST be lower case so there would be a deployment failure if someone named their resource in Aspire with an uppercase (this is why we ended up with `toLower(...)` in the first place in order to normalize names into the most restrictive subset of what Azure requires.

mitchdenny:
OK I've done a full pass over this now. Its looking really good. I've left a few comments but the themes are:

1. Resource name consistency between Aspire 8.0 and Aspire 9.0 (really important unless we've got a good reason to break it.
2. Resource API version regressions in some areas (just need to understand why, and whether its going to cause problems. Would probably pay to deploy playground/cdk sample using 8.0 and 9.0 to the same target resource group and make sure it deploys cleanly (if the resource names line up and the config settings that change are mutable it should work fine.

Not specifically called out in comments above, but some of this has got a lot more verbose. That isn't a problem provided we provide really good documentation.

playground/SqlServerEndToEnd/SqlServerEndToEnd.AppHost/sql1.module.bicep
eerhardt: That's not this change. This change is just updating the versions. Its goal is to not change behavior, if possible.

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
eerhardt: > I'm not sure about accessTier. I'll look to see if the old CDK defaulted that and if we should be setting it by default.

The old CDK was [defaulting this](https://github.com/Azure/azure-sdk-for-net/blob/5283dd0c9ac7713767e1ad7f81018e1ab51f53be/sdk/provisioning/Azure.Provisioning.Storage/src/StorageAccount.cs#L42).

```C#
accessTier: kind == StorageKind.BlobStorage || kind == StorageKind.StorageV2 ? StorageAccountAccessTier.Hot : null))
```

I believe this is the default. When using the new CDK to make a Storage Account, I see:

image


cc @tg-msft

mitchdenny: OK thats good.

Directory.Packages.props
mitchdenny: Let me guess ... the size of the generated YAML is too big and ADO is rejecting it :P Means the provisioning packages need to get moved into the sdk/[service] sub-folders instead of sdk/provisioning. At which point all versions will probably end up being different.

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
tg-msft: Set it to `new IPAddress([0, 0, 0, 0])` and it should turn into `'0.0.0.0'` in Bicep.

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
tg-msft: I'd guess https://github.com/Azure/bicep/blob/main/src/Bicep.Core/PrettyPrintV2/PrettyPrinterV2Options.cs#L9 is as definitive as we'll get for an answer on this.  I'd argue that formatter should be ignored because they allow tabs and therefore can't be trusted, but it's a cheap change in Azure.Provisioning.* so I'll update it later.

Directory.Packages.props
tg-msft: Yep.  That's the plan - and long term, the versions will always be randomly released whenever different management plane libraries are updated.

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
tg-msft: If you care about it, I'd specify it.  Those defaults do occasionally change with new service versions.

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
tg-msft: The old implementation was being overly aggressive.  Now we only sanitize out characters that aren't allowed and will change case where possible (i.e., there's no `toLower` on Storage accounts either because we'll just generate a lowercase prefix).  I plan to remove the `take`s as well when the prefix + 13 is less than max so you get nicer looking bicep. 

If you all want the exact behavior from v8 though, you can implement a property resolver like https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/provisioning/Azure.Provisioning/src/Primitives/ResourceNamePropertyResolver.cs#L186 that toLowers dashless names.  Add it to your provisioning context when calling `.Build` and it'll change how everything is named for you.

playground/bicep/BicepSample.AppHost/redis.module.bicep
tg-msft: Let me look into it.

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
eerhardt: > If you care about it

Do we care about it?

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
eerhardt: That's better than `new StringLiteral("0.0.0.0")`?

playground/AspireEventHub/EventHubs.AppHost/eventhubns.module.bicep
eerhardt: I've updated this PR to follow the existing naming scheme - https://github.com/dotnet/aspire/pull/5592/commits/16ea988c6b7b709fed3e3d1969ae36b2785254ee.

And I've logged https://github.com/dotnet/aspire/issues/5756 to decide on how to address the naming scheme going forward.

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
eerhardt: Also - why is it different between Sql and PostgreSQL? Shouldn't that be addressed in the CDK?

playground/AspireEventHub/EventHubs.AppHost/ehstorage.module.bicep
eerhardt: Will revert to existing setting for this PR.

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
eerhardt: https://github.com/Azure/azure-sdk-for-net/pull/46016 will address this.

src/Aspire.Hosting.Azure.Sql/AzureSqlExtensions.cs
eerhardt: Changed to use `new IPAddress([0, 0, 0, 0])` for PostgreSQL

davidfowl (approved):

PR #5618, opened 2024-09-09
Follow-up accessibility fixes for text visualizer, copy button

adamint:
## Description

This PR fixes two separate issues, and adds one UI improvement.
1. We only captured the onclick element for copy/text visualizer buttons, so pressing enter on these menu items previously would not do anything.
2. The text visualizer title was being truncated on mobile - since this is redundant information, we can hide it on mobile.

There is also a change to display the text visualizer in full screen at very low width/height, 85% vw and auto height otherwise.

Fixes an issue raised by vendors after #3392 was closed.

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No, not relevant
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5618)


src/Aspire.Dashboard/wwwroot/js/app.js
```diff
@@ -37,9 +37,18 @@ function getFluentMenuItemForTarget(element) {
     return null;
 }
 
+function addClickAndKeyboardListener(element, callback) {
+    element.addEventListener("click", callback);
+    element.addEventListener("keydown", function (e) {
+        if (e.keyCode === 13 || e.keyCode === 32) {
```
drewnoakes, line 43: Can we document what 13 and 32 are please? I assume Enter and Space.

drewnoakes (approved):

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
```diff
@@ -9,9 +9,15 @@
...
-            @Content.Description
-        </FluentLabel>
+
+        @if (ViewportInformation.IsDesktop)
+        {
+            // Since it's redundant information, we don't need to show the description on mobile
+            // since it may be cut off
+            <FluentLabel Typo="Typography.PaneHeader" Class="col-long-content" Style="grid-area: dialog-title;">
+                @Content.Description
+            </FluentLabel>
```
JamesNK, line 19: I don't think we should hide this in mobile. Instead, the title should be truncated with an ellipsis when needed like we do for long content elsewhere.

This has the benefit of correctly truncating very long titles in desktop view as well.

JamesNK (changes requested):

src/Aspire.Dashboard/Components/Dialogs/TextVisualizerDialog.razor
adamint: It’s unfortunately an accessibility violation if truncated. All content needs to be able to be displayed in full

JamesNK: We have truncated content all over the dashboard to handle long incoming content. We have to.

What exactly is the rule against this and why is it a problem here?

adamint: The requirement is that all content be accessible to low-vision users - we almost entirely meet this with the text visualizer. However, we have no way to meet it in this specific case unless we want to add another text visualizer button

adamint: @JamesNK

adamint: @JamesNK

JamesNK: No, we're not hiding every piece of content that could end with a `...`
Another solution must be found.

I looked at AzDo and I can immediately see content that is cut off with an ellipsis when it is too large:

image

adamint: We are not AzDo and fall under different accessibility structures than them. There’s also no guarantee that *has* passed accessibility requirements.

As mentioned, we do technically meet the requirement with the text visualizer, but our vendors are saying that is not good enough. I don’t see the problem with hiding the string in this case. We are not 

> hiding every piece of content that could end with a ...

but that information DOES always need to be available in its *entirety* somehow

JamesNK: > I don’t see the problem with hiding the string in this case.

Because it's not a complete fix. The header is still truncated in non-mobile views when viewing some data:

image

Every place we truncate content, we have no idea how long it could be. Names, keys, values are coming from external systems. There will always be situations that apps send data where it is too long to display.

There needs to be a better solution that just hiding it.

JamesNK: How about this: The header text is clickable. Clicking on it toggles truncation on and off. When truncation is off then the text wraps. It will probably look ugly, but usability is preserved in the majority of cases when the content isn't too long, and it's now possible to view all the content.

Does that meet requirements? If it doesn't meet requirements, please put some time into thinking of different alternatives.

adamint: No, that doesn’t meet requirements since it requires clicking.

adamint: I will go back to the vendors again and see if they can close the related bug.

note that the larger problem *will* remain an issue, and will require

> There will always be situations that apps send data where it is too long to display.

linking to the text visualizer allows us to sidestep that requirement

JamesNK: > No, that doesn’t meet requirements since it requires clicking.

> linking to the text visualizer, which allows us to sidestep that requirement

Using the visualizer requires clicking. Why is clicking the header different? If clicking the header is a problem, what about adding a button next to it that does the toggling? I don't understand these rules.

adamint: > Using the visualizer requires clicking

it does not require clicking. It is keyboard-accessible!

JamesNK: Ok, so the rule is a keyboard accessible way to view the complete header is required.

Which of these would work:
1. Keyboard shortcut. It could be generalized to some other page headers (visualizer header, trace detail name header, instrument name header)
2. Header as a link (is it keyboard accessible? I'm sure you must be able to tab onto a link and hit enter to trigger it)
3. If link isn't allowed, then a FluentButton

adamint: I removed the change from this PR, I'm going to follow up with the vendors to see if we can get the issue closed. If not, I'll come back to this comment

JamesNK:
What should I test for after the changes?

adamint:
> > What should I test for after the changes?
> 
> That you can use the keyboard (enter) to initiate copy button and text visualizer actions

Never mind, a bug in fluentui-blazor is causing a blocking issue. Will reopen in the future

PR #5621, opened 2024-09-09
Add OpenAI component

### LINKED ISSUE #4567: Add OpenAI component
Today we have the [Azure.AI.OpenAI](https://github.com/dotnet/aspire/tree/main/src/Components/Aspire.Azure.AI.OpenAI) component when using Azure OpenAI.

Last week an official OpenAI library was introduced with https://devblogs.microsoft.com/dotnet/openai-dotnet-library/. We should consider having a plain `Aspire.OpenAI` component that wraps this library for the cases where an Aspire app wants to use OpenAI, but doesn't want to bring in Azure dependencies.

cc @sebastienros @tg-msft @davidfowl @mitchdenny

sebastienros:
Contributes to #4567 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5621)


src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
```
radical, line 4: Use this property now which is set to `net8.0`, because `NetCurrent` is now `net9.0`.

```suggestion
    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
```

eerhardt:
Is there also work to enable `Aspire.Azure.AI.OpenAI` to work with just an OpenAIClient when configured to talk to OpenAI or a local emulator like Ollama? Or do you see that as a separate PR?

src/Components/Telemetry.md
```diff
@@ -257,6 +257,13 @@ Aspire.Npgsql.EntityFrameworkCore.PostgreSQL:
     - "db.client.connections.timeouts"
     - "db.client.connections.usage"
 
+Aspire.OpenAI:
+- Log categories:
```
eerhardt, line 261:
```suggestion
- Log categories:
  - none
```

eerhardt: FYI - @annelo-msft who is working on the ILogger support for System.ClientModel.

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <!-- This library can't ship stable until OpenAI ships stable. -->
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    <PackageTags>ai openai</PackageTags>
+    <Description>A client for OpenAI that integrates with Aspire, including metrics, logging and telemetry.</Description>
+    <PackageIconFullPath>$(SharedDir)OpenAI_256x.png</PackageIconFullPath>
```
eerhardt, line 10: Do we have/need permission to use this icon?

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <!-- This library can't ship stable until OpenAI ships stable. -->
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    <PackageTags>ai openai</PackageTags>
```
eerhardt, line 8:
```suggestion
    <PackageTags>$(ComponentCommonPackageTags) ai openai</PackageTags>
```

src/Components/Aspire.OpenAI/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.OpenAI;
+using Aspire;
+using Microsoft.Extensions.Hosting;
+
+[assembly: ConfigurationSchema("Aspire:OpenAI", typeof(OpenAISettings))]
+[assembly: ConfigurationSchema("Aspire:OpenAI:ClientOptions", typeof(AspireOpenAIExtensions))]
```
eerhardt, line 9:
```suggestion
[assembly: ConfigurationSchema("Aspire:OpenAI:ClientOptions", typeof(OpenAIClientOptions ))]
```

src/Components/Aspire.OpenAI/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Use configuration providers
+
+The .NET Aspire OpenAI library supports [Microsoft.Extensions.Configuration](https://learn.microsoft.com/dotnet/api/microsoft.extensions.configuration). It loads the `OpenAISettings` and `OpenAIClientOptions` from configuration by using the `Aspire:OpenAI` key. Example `appsettings.json` that configures some of the options:
+
+```json
+{
+  "Aspire": {
+    "OpenAI": {
+      "DisableTracing": false,
```
eerhardt, line 77: We should also show how to configure the ClientOptions.

src/Components/Aspire.OpenAI/AspireOpenAIExtensions.cs
```diff
@@ -0,0 +1,120 @@
...
+            settings.ParseConnectionString(connectionString);
+        }
+
+        configureSettings?.Invoke(settings);
+
+        var options = settings.Endpoint is null
+            ? new OpenAIClientOptions()
+            : new OpenAIClientOptions { Endpoint = settings.Endpoint };
+
+        configureOptions?.Invoke(options);
```
eerhardt, line 83: The `options` should be bound to config as well.

src/Components/Aspire.OpenAI/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
...
+
+[assembly: LoggingCategories()]
```
eerhardt, line 11: Is this necessary? Or does it throw if it isn't specfiied?

tests/Aspire.OpenAI.Tests/ConformanceTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    protected override string ValidJsonConfig => """
+        {
+          "Aspire": {
+            "OpenAI": {
+                "Endpoint": "http://YOUR_URI",
+                "DisableTracing": false,
+                "DisableMetrics": false,
+                "ClientOptions": {
+                "NetworkTimeout": "00:00:02"
```
eerhardt, line 35:
```suggestion
                    "NetworkTimeout": "00:00:02"
```

src/Components/Telemetry.md
annelo-msft: @m-redding is driving this work for SCM, we don't have a confirmed ship date for it yet, but I don't expect it to be a lot longer from here.

src/Components/Aspire.OpenAI/OpenAISettings.cs
```diff
@@ -0,0 +1,60 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Data.Common;
+
+namespace Aspire.OpenAI;
+
+/// <summary>
+/// The settings relevant to accessing OpenAI.
+/// </summary>
+public sealed class OpenAISettings
```
annelo-msft, line 11: @KrzysztofCwalina, any concerns about adding this as a public type in this namespace?  I know you and I had discussed potentially putting such types in the SCM-based client namespaces themselves.

Or, @eerhardt, could you help us understand established precedents around these settings types for Aspire more broadly?  If we e.g. wanted to enable injection of an OpenAI client in a DI container outside the context of Aspire, would we want this type to live in a non-Aspire namespace?

src/Components/Aspire.OpenAI/README.md
```diff
@@ -0,0 +1,123 @@
+# Aspire.OpenAI library
+
+Registers [OpenAIClient](https://github.com/openai/openai-dotnet?tab=readme-ov-file#using-the-openaiclient-class) as a singleton in the DI container for using the OpenAI REST API. Enables corresponding metrics, logging and telemetry.
+
+## Getting started
+
+### Prerequisites
+
+- An OpenAI REST API compatible service like OpenAI.com, Azure AI OpenAI, ollama.com and others.
```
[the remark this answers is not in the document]
annelo-msft: There are differences between the REST APIs of the OpenAI and Azure OpenAI services that require using the different .NET client libraries to communicate with them.

src/Components/Aspire.OpenAI/README.md
```diff
@@ -0,0 +1,123 @@
...
+```csharp
+private readonly OpenAIClient _client;
+
+public ChatController(OpenAIClient client)
+{
+    _client = client;
+}
+```
+
+To learn how to use the `OpenAIClient` class refer to [Using the OpenAIClient class](https://github.com/openai/openai-dotnet?tab=readme-ov-file#using-the-openaiclient-class).
```
annelo-msft, line 39: In Azure SDK terminology, we don't typically refer to clients as classes -- it is not technically incorrect to do so, but this sounds strange to my ears.  It might make sense to reword this to something like "learn how to use the `OpenAI` client library" ?

@AngelosP, can you recommend an official document the Aspire documents could link to for helping Aspire users onboard to the client?  @scottaddie, FYI that this may be an opportunity to build out docs about SCM-based clients that the Aspire team could leverage going forward.

src/Components/Aspire.OpenAI/README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+#### Connection string
+
+Alternatively, a custom connection string can be used.
+
+```json
+{
+  "ConnectionStrings": {
+    "openaiConnectionName": "Endpoint=https://{openai_rest_api_url}Key={account_key};"
```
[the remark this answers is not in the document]
annelo-msft: Does this work for the OpenAI client?  I was not aware that it supported connection strings, but if it does, I will update my mental model.

src/Components/Aspire.OpenAI/README.md
annelo-msft: If you develop a general-purpose solution for SCM-based clients (which use `ClientPipelineOptions` instead of the Azure.Core `ClientOptions` type), let me know!  We are interested in this piece on our team.

@m-redding, @KrzysztofCwalina  FYI

src/Components/Aspire.OpenAI/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+You can also setup the `OpenAIClientOptions` using the optional `Action<OpenAIClientOptions>? configureOptions` parameter of the `AddOpenAIClient` method. For example, to set a custom `NetworkTimeout` value for this client:
+
+```csharp
+builder.AddOpenAIClient("openaiConnectionName", configureOptions: options => options.NetworkTimeout = TimeSpan.FromSeconds(2));
+```
+
+## AppHost extensions
+
+There is no specific AppHost extension corresponding to the OpenAI integration. Instead a connection string resource can be registered:
```
annelo-msft, line 99: For my learning -- what is an AppHost extension and is this something we would want to have a story for in the .NET ecosystem to support better ease of use for OpenAI?  Mostly just curious on this piece.

src/Components/Aspire.OpenAI/OpenAISettings.cs
sebastienros: These settings types are used in Aspire to drive what and how services are registered in DI. This is standard in Aspire, even components have it.

src/Components/Aspire.OpenAI/README.md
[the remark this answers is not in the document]
sebastienros: Just found a new page in docs, updating.

src/Components/Aspire.OpenAI/README.md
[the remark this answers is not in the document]
sebastienros: @annelo-msft this is specific to Aspire, there is no better way right now to pass settings across app boundaries than connection strings. So Aspire components sometimed make their own.

src/Components/Aspire.OpenAI/README.md
sebastienros: An AppHost extension is a package that helps provision or describe the server part of the integration. For instance with Azure.AI.OpenAI there is a server resource that can be registered, will AI models too, such that we can provision the resource when the app is deployed. With OpenAI there is no such thing we can deploy, so we only declare a connection string, like a logical representation of this resource. Ultimately we could have a component that provides Ollama services for example.

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
sebastienros: This is the file from the open GitHub repos and package but the repos is (c) OpenAI, so we can't use it as if Microsoft was the owner of the repos.

https://openai.com/brand/

> Use of the OpenAI logo & logo mark is limited to authorized uses only and the following conditions: ...

Which means we need an authorization. Do you suggest I use the Aspire logo instead?

src/Components/Aspire.OpenAI/AssemblyInfo.cs
sebastienros: Removed. Works without it. I just wanted a placeholder to remember there are no logging categories.

src/Components/Aspire.OpenAI/AspireOpenAIExtensions.cs
sebastienros: Options are now bound, test added.

src/Components/Aspire.OpenAI/OpenAISettings.cs
annelo-msft: If we wanted to also add a type called `OpenAISettings` in the `OpenAI` client namespace, would that be frowned upon?

src/Components/Aspire.OpenAI/README.md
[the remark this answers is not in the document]
annelo-msft: I believe @AngelosP would be the best person to share what are considered official docs for this library.

src/Components/Aspire.OpenAI/README.md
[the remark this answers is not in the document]
annelo-msft: Got it.  Thank you for the explanation!

src/Components/Aspire.OpenAI/OpenAISettings.cs
eerhardt: It wouldn't be great, but it would be possible. The namespace can disambiguate.

For Aspire, the pattern is `{NameOfLibrary}Settings` in the `Aspire.{NameOfLibrary}` namespace.

See

https://github.com/dotnet/aspire/blob/0490bcbc4318c22774d162b3f49382f148cfa6b2/src/Components/Aspire.RabbitMQ.Client/RabbitMQClientSettings.cs#L6-L9

https://github.com/dotnet/aspire/blob/0490bcbc4318c22774d162b3f49382f148cfa6b2/src/Components/Aspire.StackExchange.Redis/StackExchangeRedisSettings.cs#L6-L9

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
eerhardt: @DamianEdwards - you've handled this for other integrations. Can you help here?

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <IsPackable>true</IsPackable>
+    <!-- This library can't ship stable until OpenAI ships stable. -->
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    <PackageTags>$(ComponentCommonPackageTags) ai openai</PackageTags>
+    <Description>A client for OpenAI that integrates with Aspire, including metrics, logging and telemetry.</Description>
```
eerhardt, line 9:
```suggestion
    <Description>A client for OpenAI that integrates with Aspire, including metrics and telemetry.</Description>
```

No logging (yet).

src/Components/Aspire.OpenAI/AspireOpenAIExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        if (builder.Configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            settings.ParseConnectionString(connectionString);
+        }
+
+        configureSettings?.Invoke(settings);
+
+        var options = new OpenAIClientOptions();
+        builder.Configuration.GetSection($"{configurationSectionName}:ClientOptions").Bind(options);
```
eerhardt, line 80: I think we should use the "IOptions" pattern here, similar to how we do it in Redis:

https://github.com/dotnet/aspire/blob/0490bcbc4318c22774d162b3f49382f148cfa6b2/src/Components/Aspire.StackExchange.Redis/AspireRedisExtensions.cs#L89-L96

We are probably inconsistent in the other client libraries, but Redis and the Azure ones follow IOptions pattern.

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
DamianEdwards: I'll defer to @maddymontaquila now :smile:

src/Components/Aspire.OpenAI/OpenAISettings.cs
annelo-msft: Cool, as long as we can keep that an an option in our back pocket, I have no concerns.

src/Components/Aspire.OpenAI/README.md
annelo-msft: Thanks!  My comment here doesn't block this PR in any way, but I'm interested in understanding why we have one for Azure.AI.OpenAI and not OpenAI?  If there are pieces needed uniquely for SCM-based clients I'd be interested to understand the gaps there and if pieces are needed from SCM to address those gaps?

eerhardt: > why we have one for Azure.AI.OpenAI and not OpenAI?

Because you can provision an Azure OpenAI service (using bicep). That's what the "AppHost" extension will do.

You can't provision an OpenAI service. You need to sign up on their website.

src/Components/Aspire.OpenAI/Aspire.OpenAI.csproj
sebastienros: Removing the logo for now until we have an answer.

sebastienros:
Failing on unrelated cosmos tests

eerhardt (approved):
LGTM.

tests/Aspire.OpenAI.Tests/AspireOpenAIExtensionsTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            builder.AddOpenAIClient("openai");
+        }
+
+        using var host = builder.Build();
+        var client = useKeyed ?
+            host.Services.GetRequiredKeyedService<OpenAIClient>("openai") :
+            host.Services.GetRequiredService<OpenAIClient>();
+
+        Assert.NotNull(client);
```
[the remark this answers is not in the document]
eerhardt: I meant, can we assert something like the connection string or URI is correct?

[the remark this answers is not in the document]
sebastienros: That was my initial thought too but I couldn't find it in the model, do you know if it's somewhere? I didn't dig super deep, maybe I missed it. Will check again.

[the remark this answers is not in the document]
sebastienros: Are you fine with private reflection in this case?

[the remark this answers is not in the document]
eerhardt: Nah, let's just live with what we have. If we mis-configured it, we will know because things won't work.

PR #5628, opened 2024-09-09
Enable container-to-container service communication

### LINKED ISSUE #5584: Container to container networking issues with podman and vanilla docker engine
We have a slew of issues across the repo detailing broken networking behavior from both container to host and container to container calls when docker desktop isn't installed. 

- https://github.com/dotnet/aspire/issues/5097
- https://github.com/dotnet/aspire/issues/5379
- https://github.com/dotnet/aspire/issues/4136
- https://github.com/dotnet/aspire/issues/3910
- https://github.com/dotnet/aspire/issues/5510
- https://github.com/dotnet/aspire/issues/5509
- https://github.com/dotnet/aspire/issues/5274
- https://github.com/dotnet/aspire/issues/5391

This is because container to container communication goes through the host network either using (host.docker.internal or host.containers.internal). Docker desktop does extra work to make this possible and we lean on it in aspire for container to container communication. We should solve container to container comms by using docker networks instead.

danegsta:
## Description

This PR implements a custom container network for an Aspire AppHost run and uses it to support direct container-to-container service communication without going through service proxies on the host. This should allow scenarios such as sidecar management containers to work with more container runtime environments. For now we're hardcoding container-to-container connections in individual sidecar resources until we design and implement an updated API for connection service discovery that can better support these scenarios.

Fixes https://github.com/dotnet/aspire/issues/5584

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5628)


src/Aspire.Hosting/ApplicationModel/ContainerNetworkResource.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
```
davidfowl, line 4: Lets remove this.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -914,6 +929,14 @@ private void AddAllocatedEndpointInfo(IEnumerable<AppResource> resources)
 
         foreach (var appResource in resources)
         {
+            if (appResource.DcpResource is Container container)
+            {
+                if (container.Spec.Networks?.Any() == true)
+                {
+                    containerHost = appResource.ModelResource.Name;
```
davidfowl, line 936: This is the thing I'm not sure about. Container host is used by code to talk to host resources from a container. We need a different property for this.

danegsta: For executable resources, the endpoint container host should still be set to `host.docker.internal` (or equivalent), I only override to the container alias if the endpoint provider is a container (and the container is attached to a custom network, which probably isn't necessary with the latest changes).

src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
```diff
@@ -81,6 +82,11 @@ public EndpointReferenceExpression Property(EndpointProperty property)
     /// </summary>
     public int? TargetPort => EndpointAnnotation.TargetPort;
 
+    /// <summary>
+    /// Returns the port to use when referencing this endpoint from a container.
+    /// </summary>
+    public int? ContainerPort => Resource.IsContainer() ? EndpointAnnotation.TargetPort : AllocatedEndpoint.Port;
```
karolz-ms, line 88: `ContainerClientPort` maybe? (I am really struggling to find a good name for this)

There is a big assumption behind this implementation, namely, that all containers are running on a single network.  In future we will want to relax this, but this is probably the best we can do for Aspire 9. We should discuss the deprecation / plan for Endpoint.v2 API before we finalize this.

danegsta: I'm torn; I thought about a more explicit name that implies this is specifically a port to be used to access the endpoint from within a container, but we have the existing `ContainerHost` property that effectively serves the same purpose, so I erred on the side of keeping the new property name consistent.

I'll admit I have a suspicion that we'll end up wanting to always have a default common shared network for our containers, with any future support for custom networks being additive, but I think we should be able to get a better understanding of the full pros and cons and where we want to take container based service discovery in the long term with some discussion in the next few days

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: I'm not sure about this, if you try to make a "self" address that another resource can use to talk to you, it'll be incorrect. The resolution for host depends on the types of both resources.

danegsta: The logic I'm aiming for is that `ContainerHost`/`ContainerPort` are used if the requesting component is a container resource, but the value for each property will depend on the resource that's providing the endpoint.

| Endpoint Provider | Host Value | ContainerHost Value | ContainerPort Value |
| ------------------ | ---------- | ---------------- | --------------------- |
| Executable             | 127.0.0.1    |  "host.docker.internal" |  `Port`               |
| Container               | 127.0.0.1    | Network alias name | `TargetPort` |

The main caveat of the design is that executable resources are expected to use `Host` and `Port` to access services and containers to use `ContainerHost` and `ContainerPort`. I'll admit it's a bit of a kludge to force separate endpoint properties for container and non-container resources.

It might be interesting (long term) to think of the resources and endpoints more like a graph than just the endpoints being metadata on individual resources. The resources are nodes and the endpoints are directional edges between the nodes. Instead of querying a given resources for its endpoint metadata/annotations, we'd start the query from the consuming resource with the destination endpoint as the target. For example, `redisCommanderResource.FindServiceConnection(redisInstance, "tcp");` to get the endpoint/connection info to connect to a redis container from a redis commander sidecar.

src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
```diff
@@ -55,6 +55,7 @@ internal string GetExpression(EndpointProperty property = EndpointProperty.Url)
             EndpointProperty.Port => "port",
             EndpointProperty.Scheme => "scheme",
             EndpointProperty.TargetPort => "targetPort",
+            EndpointProperty.ContainerPort => "containerPort",
```
davidfowl, line 58: We don't want a new expression here, azd will have to change to understand this.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -202,7 +202,7 @@ public static IResourceBuilder<T> WithPgAdmin<T>(this IResourceBuilder<T> builde
                         writer.WriteString("Name", postgresInstance.Name);
                         writer.WriteString("Group", "Servers");
                         writer.WriteString("Host", endpoint.ContainerHost);
-                        writer.WriteNumber("Port", endpoint.Port);
+                        writer.WriteNumber("Port", (decimal)endpoint.ContainerPort!);
```
davidfowl, line 205: Why is this casting to decial?

danegsta: The linter got me confused; updated to cast to int

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: The problem is a component today that is using container host + port. That will start exploding right? Right now we assume a single network (the host network). We also want to handle this logic that replaces localhost with the container host https://github.com/dotnet/aspire/blob/89fdd219ac472f6151f9f4037b41dccbb2ea1063/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L1337

danegsta: Based on our conversation, I switched back to the temporary workaround approach for our sidecar resources. We'll hardcode the target resource container host name/target port and spend some time discussing what new service discovery API for these scenarios should look like before we make any public API changes.

davidfowl:
Errors here:

```
 Expected: myredis1:host.containers.internal:5001:0
[xUnit.net 00:00:05.16]       Actual:   myredis1:myredis1:6379:0
```

davidfowl:
Now the port is wrong 😄

tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs
```diff
@@ -182,21 +180,19 @@ public async Task SingleRedisInstanceProducesCorrectRedisHostsVariable(string co
             DistributedApplicationOperation.Run,
             TestServiceProvider.Instance);
 
-        Assert.Equal($"myredis1:{containerHost}:5001:0", config["REDIS_HOSTS"]);
+        Assert.Equal($"myredis1:{redis.Resource.Name}:5001:0", config["REDIS_HOSTS"]);
```
davidfowl, line 183:
```suggestion
        Assert.Equal($"myredis1:{redis.Resource.Name}:6379:0", config["REDIS_HOSTS"]);
```

tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs
```diff
@@ -208,7 +204,7 @@ public async Task MultipleRedisInstanceProducesCorrectRedisHostsVariable(string
             DistributedApplicationOperation.Run,
             TestServiceProvider.Instance);
 
-        Assert.Equal($"myredis1:{containerHost}:5001:0,myredis2:host2:5002:0", config["REDIS_HOSTS"]);
+        Assert.Equal($"myredis1:{redis1.Resource.Name}:5001:0,myredis2:host2:5002:0", config["REDIS_HOSTS"]);
```
davidfowl, line 207:
```suggestion
        Assert.Equal($"myredis1:{redis1.Resource.Name}:6379:0,myredis2:host2:5002:0", config["REDIS_HOSTS"]);
```

tests/Aspire.Hosting.MongoDB.Tests/AddMongoDBTests.cs
```diff
@@ -154,7 +152,7 @@ public async Task WithMongoExpressUsesContainerHost(string containerHost)
             e =>
             {
                 Assert.Equal("ME_CONFIG_MONGODB_URL", e.Key);
-                Assert.Equal($"mongodb://{containerHost}:3000/?directConnection=true", e.Value);
+                Assert.Equal($"mongodb://mongo:3000/?directConnection=true", e.Value);
```
davidfowl, line 155:
```suggestion
                Assert.Equal($"mongodb://mongo:27017/?directConnection=true", e.Value);
```

tests/Aspire.Hosting.PostgreSQL.Tests/AddPostgresTests.cs
```diff
@@ -476,7 +474,7 @@ public async Task WithPostgresProducesValidServersJsonFile(string containerHost)
         // Make sure the first server is correct.
         Assert.Equal(pg1.Resource.Name, servers.GetProperty("1").GetProperty("Name").GetString());
         Assert.Equal("Servers", servers.GetProperty("1").GetProperty("Group").GetString());
-        Assert.Equal(containerHost, servers.GetProperty("1").GetProperty("Host").GetString());
+        Assert.Equal("mypostgres1", servers.GetProperty("1").GetProperty("Host").GetString());
         Assert.Equal(5001, servers.GetProperty("1").GetProperty("Port").GetInt32());
```
davidfowl, line 478:
```suggestion
        Assert.Equal(5432, servers.GetProperty("1").GetProperty("Port").GetInt32());
```

tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs
```diff
@@ -208,7 +204,7 @@ public async Task MultipleRedisInstanceProducesCorrectRedisHostsVariable(string
             DistributedApplicationOperation.Run,
             TestServiceProvider.Instance);
 
-        Assert.Equal($"myredis1:{containerHost}:5001:0,myredis2:host2:5002:0", config["REDIS_HOSTS"]);
+        Assert.Equal($"myredis1:{redis1.Resource.Name}:6379:0,myredis2:host2:6379:0", config["REDIS_HOSTS"]);
```
davidfowl, line 207:
```suggestion
        Assert.Equal($"myredis1:{redis1.Resource.Name}:6379:0,myredis2:myredis2:6379:0", config["REDIS_HOSTS"]);
```

davidfowl:
Beyond the small changes to ports and host names, it seems like using the docker network slows down the startup of these containers or breaks logs in some way that causes tests to fail.

tests/Aspire.Hosting.Redis.Tests/AddRedisTests.cs
```diff
@@ -208,7 +204,7 @@ public async Task MultipleRedisInstanceProducesCorrectRedisHostsVariable(string
             DistributedApplicationOperation.Run,
             TestServiceProvider.Instance);
 
-        Assert.Equal($"myredis1:{containerHost}:5001:0,myredis2:host2:5002:0", config["REDIS_HOSTS"]);
+        Assert.Equal($"myredis1:{redis1.Resource.Name}:6379:0,myredis2:myredis2:6379:0", config["REDIS_HOSTS"]);
```
mitchdenny, line 207: I don't quite understand why these ports are changing for these tests?

danegsta: Container to container requires using the in-container target port instead of the public port for connections

mitchdenny:
With the container network creation, is this something that might benefit from being persistent by default? What I'm thinking is that if we made it persistent it would allow developers to spin up things manually and attach it to that network if they want to.

mitchdenny:
Is now a good time to discuss Podman testing? At the moment all of our testing is done against Docker/Docker Desktop, but we say we support Podman but don't do any validation in our CI.

danegsta:
> With the container network creation, is this something that might benefit from being persistent by default? What I'm thinking is that if we made it persistent it would allow developers to spin up containers manually and attach it to that network if they want to.

Long term I think it may be cleaner to work out how we want to expose container networks as a primitive in the resource builder API and allow users to configure their networks (including persistent networks to support this container attach scenario) that way.

danegsta:
My main concern is that the V1 approach in this PR is taking an extremely naive approach that depends on a lot of assumptions that could fall apart if we allow non-Aspire managed containers to attach to the network. Depending on tightly managed AppHost scoped networks makes the assumptions a bit more reasonable.

davidfowl:
Other failures are gone! Thanks for the bug fix in dcp @danegsta!

@radical the failure is:

docker command 'CreateContainer' returned with non-zero exit code 1: command output: Stdout: '' Stderr: 'Unable to find image 'netaspireci.azurecr.io/library/mongo-express:1.0' locally\nError response from daemon: manifest for netaspireci.azurecr.io/library/mongo-express:1.0 not found: manifest unknown: manifest tagged by \"1.0\" is not found

How do we get container images on our test feed?

radical:
> @radical the failure is:
> 
> docker command 'CreateContainer' returned with non-zero exit code 1: command output: Stdout: '' Stderr: 'Unable to find image 'netaspireci.azurecr.io/library/mongo-express:1.0' locally\nError response from daemon: manifest for netaspireci.azurecr.io/library/mongo-express:1.0 not found: manifest unknown: manifest tagged by "1.0" is not found

I can add this one.

> How do we get container images on our test feed?

https://github.com/radical/aspire/blob/tests-run-scripts/docs/updating-container-registry.md

davidfowl (approved):

PR #5630, opened 2024-09-09
Update container lifetime API and add schema for lifecycleKey property

danegsta:
## Description

Renames the `WithContainerLifetime` experimental API to `WithLifetime`, changes the `ContainerLifetimeType` enum to `ContainerLifetime` and renames `ContainerLifetimeType.Persistent` to `ContainerLifetime.CreateIfNotExistsPersistOnExit` as a more explicit description of the lifetime behavior.

Adds DCP schema for `lifecycleKey` which can be used to configure the lifetime of persistent containers (if not set, DCP will calculate a default lifecycle key value). If set, DCP will check existing persistent containers for to see if they have a matching lifecycle key. If not, the container will be destroyed and re-created based on the new AppHost config.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No (modified newly added experimental API)
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: TBD
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5630)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ContainerLifetimeAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ContainerLifetimeAnnotation.cs
@@ -15,9 +15,23 @@ public enum ContainerLifetime
     /// </summary>
     Default,
     /// <summary>
-    /// The resource is persistent and will not be disposed of when the AppHost shuts down.
+    /// Attempt to re-use a previously created resource (based on the container name) if one exists. Do not destroy the container on AppHost shutdown.
     /// </summary>
-    CreateIfNotExistsPersistOnExit,
+    /// <remarks>
+    /// In the event that a container with the given name does not exist, a new container will always be created based on the
+    /// current <see cref="ContainerResource"/> configuration.
+    /// <para>When an existing container IS found, Aspire MAY re-use it based on the following criteria:</para>
+    /// <list type="bullet">
+    /// <item>If the container WAS NOT originally created by Aspire, the existing container will be re-used.</item>
+    /// <item>If the container WAS originally created by Aspire:
+    /// <list type="bullet">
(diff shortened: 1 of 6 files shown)


mitchdenny (approved):

davidfowl:
Now I think we should spend some time bikeshedding about that name of this enum. If it's this explicit (and long) then we might need a helper method.

danegsta:
> Now I think we should spend some time bikeshedding about that name of this enum. If it's this explicit (and long) then we might need a helper method.

`Reusable` just popped into my head as a possible option.

danegsta:
> > Now I think we should spend some time bikeshedding about that name of this enum. If it's this explicit (and long) then we might need a helper method.
> 
> `Reusable` just popped into my head as a possible option.

@mitchdenny @davidfowl @DamianEdwards do we want to go with the verbose `CreateIfNotExistsPersistOnExit` or a shorter option like `Persistent`/`Reusable`?

davidfowl:
I think I want to keep the name persistent and add a giant doc comment for what that means instead of encapsulating it in the name with a really long enum.

danegsta:
I reverted the name and made a pass at a more comprehensive doc comment, but I feel like the description of the behavior could still use a bit more work.

davidfowl (approved):

danegsta:
I'll merge this since we're happy with the API; any future improvements to the doc comment can be made in a follow up PR.

Issue #4741, opened 2024-07-01
Create functional tests for additional integrations


radical:
Split from #1566 

We are missing the EntityFramework components in our functional integration tests added in #1217. We should add them, plus any other missing components that would be valuable.

TODO:

- [ ] orleans
- [x] cosmos
- [x] dapr (@Alirexaa: moving to CommunityToolkit)
- [x] nats
- [x] seq
- [x] qdrant

Tests needed that reference these nugets too:
- [ ] Microsoft.Extensions.ServiceDiscovery
- [ ] Microsoft.Extensions.ServiceDiscovery.Abstractions
- [ ] Microsoft.Extensions.ServiceDiscovery.Dns
- [ ] Microsoft.Extensions.ServiceDiscovery.Yarp

cc @eerhardt

eerhardt:
qdrant is done:

https://github.com/dotnet/aspire/blob/9a68b552f847caa62ebc612e4dc23160c29e36d5/tests/Aspire.Hosting.Qdrant.Tests/QdrantFunctionalTests.cs#L23-L63

radical:
cc @sebastienros

Alirexaa:
nats is done too:

https://github.com/dotnet/aspire/blob/9a68b552f847caa62ebc612e4dc23160c29e36d5/tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs#L21-L61

eerhardt:
CosmosDB has both EF and non-EF functional tests in [Extract EventHubs and Cosmos Tests from EndToEndTests and into Aspire.Hosting.Azure.Tests (dotnet/aspire#6538)](https://github.com/dotnet/aspire/pull/6538).

Alirexaa:
Seq is done.

Dapr moved to CommunityToolkit.

davidfowl:
I think this is done

PR #5639, opened 2024-09-10
[tests] Improve Cosmos EndToEnd test

radical:
Wait for the texts indicating that the db, and service are ready for requests before sending the http request. This helps with unnecessary attempts, and wait times in the retry pipeline.

- **[tests] TestProject.WorkerA - reduce pings logged to every 30s instead of 1s to reduce noise in the logs**
- **[tests] TestProject.AppHost - use single line logging to help with matching log texts in the tests**
- **[tests] EndToEnd tests - add some text to wait for before sending the http request**

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5639)

### DIFF
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
@@ -5,7 +5,7 @@ var builder = DistributedApplication.CreateBuilder(args);
 
 var db = builder.AddAzureCosmosDB("cosmos")
                 .AddDatabase("db")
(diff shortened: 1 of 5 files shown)


radical:
Related: https://github.com/dotnet/aspire/pull/5641

radical:
> What do you think about removing this test from the EndToEnd and putting it in its own FunctionalTests? Like we did for the other integrations. I think this is just left over from that work.
> 
> Then we wouldn't need to clutter up the EndToEnd tests with this change.

Makes sense. cc @sebastienros as he has been adding the functional tests.

radical:
Added `cosmos` to https://github.com/dotnet/aspire/issues/4741 . Closing this in favor of that issue.

PR #5688, opened 2024-09-12
Replace use of obsolete X509Certificate2 API

### LINKED ISSUE #5461: Dashboard should use X509CertificateLoader instead of obsolete X509Certificate2 constructor

### Describe the bug

https://github.com/dotnet/aspire/blob/f5d9f80c3c667f5c84debfd427e7b0ae83493f7d/src/Aspire.Dashboard/ResourceService/DashboardClient.cs#L162-L172

`new X509Certificate2` is obsolete in net9.0. Trying to target the dashboard to net9.0 gives a warning here.

See:
* https://aka.ms/dotnet-warnings/SYSLIB0057
* https://github.com/dotnet/runtime/pull/104165
* https://github.com/dotnet/runtime/issues/91763

### Expected Behavior

We should be able to target net9.0 without obsolete warnings.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

cc @drewnoakes @JamesNK

drewnoakes:
Fixes #5461

In .NET 9 the `X509Certificate2` constructor we were using is being made obsolete. Instead, we must use the new `X509CertificateLoader` class.

This new class was added in .NET 9, but is available for downlevel use via the `Microsoft.Bcl.Cryptography` package, which is added in this change.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5688)


Directory.Packages.props
```diff
@@ -103,6 +103,7 @@
     <PackageVersion Include="Grpc.Tools" Version="2.65.0" />
     <PackageVersion Include="Humanizer.Core" Version="2.14.1" />
     <PackageVersion Include="KubernetesClient" Version="14.0.9" />
+    <PackageVersion Include="Microsoft.Bcl.Cryptography" Version="9.0.0-rc.1.24431.7" /> <!-- Can remove Microsoft.Bcl.Cryptography when the repo moves to .NET 9 or later -->
```
JamesNK, line 106: Can we can't get to the final version of the package for the .NET 9 release? Will timings work out?

If not, is it ok to ship a preview package?

JamesNK (approved):

JamesNK:
Are there tests of the client loading certificates? If not, some should be added.

src/Aspire.Dashboard/ResourceService/DashboardClient.cs
```diff
@@ -168,7 +168,23 @@ X509CertificateCollection GetFileCertificate()
                 var filePath = _dashboardOptions.ResourceServiceClient.ClientCertificate.FilePath;
                 var password = _dashboardOptions.ResourceServiceClient.ClientCertificate.Password;
 
-                return [new X509Certificate2(filePath, password)];
+                var certBytes = File.ReadAllBytes(filePath);
+
+                var certContentType = X509Certificate2.GetCertContentType(filePath);
```
CZEMacLeod, line 173: Why are you reading the file twice? You could use the overload taking `ReadOnlySpan<Byte>` and pass certBytes instead of filePath...

drewnoakes: Thanks yes, that was the intention. Good spot. Will fix before merging.

Directory.Packages.props
drewnoakes: @joperezr @davidfowl any concerns here?

drewnoakes:
> Are there tests of the client loading certificates? If not, some should be added.

We have a manual test script for these scenarios for now.

JamesNK (changes requested):
Add a test please.

It doesn't look difficult. There are test certs in the repo + make GetFileCertificate internal + create client with options pointing to the test file, then verify the result of GetFileCertificate.

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        Assert.NotNull(certificates);
+
+        var cert = Assert.Single(certificates.Cast<X509Certificate>());
+
+        Assert.NotNull(cert);
+        Assert.Equal("OU=Development, O=Contoso, L=Alexandria, S=Virginia, C=US", cert.Subject);
+    }
+
+    private sealed class MockLogger : ILogger, IDisposable, IEnumerable
```
JamesNK, line 92: You can use TestSink for this: https://github.com/dotnet/aspire/blob/9b5051080379d62b2066ab334be0e35064f4fdf9/tests/Aspire.Dashboard.Tests/Model/ApplicationsSelectHelpersTests.cs#L94-L103

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
```diff
@@ -0,0 +1,132 @@
...
+namespace Aspire.Dashboard;
+
+public sealed class CertificateUtilTests
+{
+    [Theory]
+    [InlineData("testCert.pfx", X509ContentType.Pkcs12)]
+    [InlineData("https-dsa.pem", X509ContentType.Cert)]
+    public void TestCertificateTypes(string fileName, X509ContentType contentType)
+    {
+        var filePath = TestCertificateLoader.GetCertPath(fileName);
```
JamesNK, line 21: nit: add arrange, assert, act comments to tests

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
drewnoakes: That code would be simpler if it used this `MockLogger` class.

`MockLogger` also throws as soon as an unexpected write occurs, so you get the stack involved.

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
drewnoakes: I don't love those comments. Most tests in this repo don't use them.

.config/CredScanSuppressions.json
```diff
@@ -32,6 +32,10 @@
     {
       "file": "\\tests\\Shared\\TestCertificates\\testCert.pfx",
       "_justification": "Legitimate UT certificate file with private key, from dotnet/aspnetcore"
+    },
+    {
+      "file": "\\tests\\Shared\\TestCertificates\\https-dsa.pem",
```
joperezr, line 37: Did you run an internal build and something was flagging? This is just a public certificate, so I wouldn't expect this to trigger CredScan

Directory.Packages.props
joperezr: There are some concerns, but perhaps not really ship-blockers. Starting from RC2, all of the packages from dotnet/runtime will only flow internally, and so far we are mostly just building externally. Our two options would be either ship depending on RC2 build (which while not ideal wouldn't have a big impact since the dashboard ships as a published app with no package references, and hence it can be stable still), or building internally so that we get the final 9.0 version that will get shipped. cc: @bartonjs

.config/CredScanSuppressions.json
drewnoakes: I was following the directions in https://github.com/dotnet/aspire/blob/main/tests/Shared/TestCertificates/README.md

Are you saying that because there's no private key here, the suppression is redundant? I'm happy to remove it.

drewnoakes: Oh, I see what happened. Will remove and add a little more clarity to those docs.

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
drewnoakes: That's still a lot of extra code to complicate test methods when compared with:

```c#
using var logger = new MockLogger() { "Expected line" };
...
```

src/Aspire.Dashboard/ResourceService/DashboardClient.cs
```diff
@@ -158,42 +164,6 @@ GrpcChannel CreateChannel()
...
-            X509CertificateCollection GetFileCertificate()
-            {
-                Debug.Assert(
-                    _dashboardOptions.ResourceServiceClient.ClientCertificate.FilePath != null,
-                    "FilePath is validated as not null when configuration is loaded.");
-
-                var filePath = _dashboardOptions.ResourceServiceClient.ClientCertificate.FilePath;
-                var password = _dashboardOptions.ResourceServiceClient.ClientCertificate.Password;
-
-                return [new X509Certificate2(filePath, password)];
```
bartonjs, line 171: "99%" of the reason to use the new X509CertificateLoader over the old constructors is to not load a PKCS#12/PFX when you weren't expecting to.  Since you were expecting to load a PFX, and you're loading from a local file, it's not the end of the world if you continue calling the legacy constructor.  (And since you're only loading it once you don't care about "we made parallel-loading the same PFX better" feature, etc.)

So, while it is forward looking to get off of this ctor and on to the new loader model, it's not "important" right now.

src/Aspire.Dashboard/Utils/CertificateUtil.cs
```diff
@@ -0,0 +1,48 @@
...
+            return [X509CertificateLoader.LoadPkcs12(certBytes, password)];
+        }
+        else
+        {
+            if (password is not null)
+            {
+                logger.LogDebug("Resource service certificate {FilePath} has type {Type} which does not support passwords, yet a password was configured. The certificate password will be ignored.", filePath, certContentType);
+            }
+
+            return [X509CertificateLoader.LoadCertificate(certBytes)];
```
bartonjs, line 27: _Technically_ speaking, you have a functionality regression here.  The old code would load any of

* A single certificate
  * Which is 99% useless, but SslStream has a feature where it looks in the cert stores to see if maybe you happened to have specified a certificate that you already know about and have a private key for... (which might only be on Windows?)
* A PKCS#7 SignedCms file.
  * Where it loads the certificate used by the first signer, or fails if it can't find one (or there is no first signer).
  * And then is the same as the "plain certificate" case.
* A Windows SerializedCert blob (Windows only)
  * This "might" have an associated private key, just by virtue of it might have written down "and my private key is named {whatever}" AND that key happens to be available on this machine/user-context.
  * Otherwise it's just a "plain certificate"
  * But basically no one uses this format.
* An Authenticode-signed file (exe, dll, msu, cab, etc). (Windows-only)
  * It extracts the certificate of the first signer.
  * Sounds a bit like SignedCms, eh? Oh, wait, Authenticode uses SignedCms.  Less coincidence, more "same code".
* A PKCS#12/PFX file
  * _Probably_ has a certificate in it with a private key, but, might not.

The new code will throw if it encounters any of those three middle formats, because the new loader is "one method (group), one file format".  The middle ones don't _really_ make sense in context for you, so they're not really important... but this is, technically, marginally, a breaking change.

https://github.com/dotnet/core/blob/main/release-notes/9.0/preview/preview7/libraries.md#changes-to-x509-certificate-loading shows an equivalent to what `new X509Certificate2(bytesOrPath, password, flags)` does.

src/Aspire.Dashboard/Utils/CertificateUtil.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace Aspire.Dashboard;
+
+internal static class CertificateUtil
+{
+    public static X509CertificateCollection GetFileCertificate(string filePath, string? password, ILogger logger)
+    {
+        var certBytes = File.ReadAllBytes(filePath);
```
bartonjs, line 12: You should be able to keep working in terms of the path, and not need to use ReadAllBytes.

`GetCertContentType` accepts path-or-bytes, but the cert load functions you need the "FromFile" variants.

The FromFile variants do things smarter than File.ReadAllBytes; but since it's a one-time/startup cost it probably doesn't matter for you.

tests/Aspire.Dashboard.Tests/Utils/CertificateUtilTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        Assert.Throws<CryptographicException>(() => CertificateUtil.GetFileCertificate(
+            filePath: TestCertificateLoader.GetCertPath("testCert.pfx"),
+            password: "wrongPassword",
+            logger: NullLogger.Instance));
+    }
+
+    [Fact]
+    public void GetFileCertificate_Cert_WithRedundantPassword()
+    {
+        var filePath = TestCertificateLoader.GetCertPath("https-dsa.pem");
```
bartonjs, line 54: It's a little weird to me that you're testing with DSA, since that algorithm itself is now entirely obsolete.  (FIPS 186-5 withdrew it, so it's not just "the least popular", it's "RIP")

drewnoakes:
@bartonjs thanks for the super detailed notes. Given the above, would you suggest just continuing to use the obsolete API and suppressing the warning once we move to `net9.0`? Seems like we're not gaining much from this change, other than an additional package dependency.

bartonjs:
> Would you suggest just continuing to use the obsolete API and suppressing the warning once we move to net9.0?

Once you move to 9, X509CertificateLoader will be part of the platform.  At that time, I recommend you use it rather than the obsolete methods.  I'd probably take the break to only support PKCS#12/PFX at that time (as well as the alternative of loading the cert from a cert store), as none of the other formats make sense in context.

drewnoakes:
This is on hold until the repo moves to .NET 9.

joperezr:
> This is on hold until the repo moves to .NET 9.

That won't happen any time soon though, or where you referring once .NET 9 ships and we can add a dependency to the OOB package? If not, then we can probably just close this one based on @bartonjs feedback.

drewnoakes:
Let's close and revisit as needed in future.

PR #5693, opened 2024-09-12
dashpage ui work

### LINKED ISSUE #5287: Create DashpageChart component to render a single chart in a dashpage
When showing a single metric, we make generous use of the screen real estate.

For dashpages, we'll want a more constrained view, as we want to show multiple metrics in the same area.

We need a component that displays:

- A chart
- A table
- A means to toggle between chart/table
- A means to show the filter

The toggle and filters should be accessed via a discrete icon in the corner. They must be accessible.

---

## Single metric view

image

## DashpageChart view (proposed)

Similar to Azure Data Explorer dashboard components:

image

adamint:
Fixes #5287

- saves current dashpage state
- adds filters to dashpage metrics
- fixes initial render bug

dashpages: 
image
image
image

single metric view:
image
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5693)


src/Aspire.Dashboard/Utils/DashboardUrls.cs
```diff
@@ -45,6 +45,11 @@ public static string MetricsUrl(string? resource = null, string? meter = null, s
                 url = QueryHelpers.AddQueryString(url, "instrument", instrument);
             }
         }
+        if (dashpage is not null)
+        {
+            // dashpage must be querystring parameters because it's valid for the name to contain forward slashes.
```
drewnoakes, line 50: We could change this requirement. None of these exist, so we can't break any existing config.

Query parameters are currently used for single chart views, and it would be nice if we were consistent across these.

adamint: I'm not sure what you're proposing, since this addition allows dashpages to include special query string characters, consistent with instrument

src/Aspire.Dashboard/Components/Controls/Chart/MetricTable.razor
```diff
@@ -18,7 +18,7 @@
     }
 }
 
-<div id="metric-table-container" style="height: 40vh; overflow-y: auto; width:1200px;">
+<div id="@_metricTableDivId" class="metric-table-container">
```
adamint, line 21: we now can have multiple tables, we should use unique ids

src/Aspire.Dashboard/Components/Pages/Metrics.razor.cs
```diff
@@ -95,7 +99,7 @@ protected override async Task OnInitializedAsync()
             SelectedApplication = _selectApplication,
             SelectedDuration = _durations.Single(d => d.Id == s_defaultDuration),
             SelectedViewKind = null,
-            Dashpages = await DashpagePersistence.GetDashpagesAsync(CancellationToken.None)
+            Dashpages = []
```
adamint, line 102: note: previous code fails on first invocation! as ViewModel will be null on first render

drewnoakes (approved):
Thanks for the helpful comments on your own PR.

I pulled the branch down and tried it out. It looks much nicer with this styling, thanks!

---

The only issue I found is that filter popups can appear like this:

image

However, this is fixed in a newer version of the Fluent UI library, and we will upgrade for 9.0, so I think this is fine to leave as is.

Left a few comments, though have pushed changes to address them. The comments are mostly FYI. I will merge this if/when the helix leg passes.

PR #5694, opened 2024-09-12
[tests] Fixes to improve helix test runs hitting docker soft limits for network subnets

radical:
- Disable parallel test runs on helix
	- This would help with too many docker network subnets being allocated in
parallel, and thus hitting soft limits on the helix agents.
- Prune docker networks on helix agents before/after the runs

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5694)

### DIFF
--- a/tests/Directory.Build.targets
+++ b/tests/Directory.Build.targets
@@ -8,6 +8,7 @@
     <VSTestRunSettingsFile Condition="'$(VSTestRunSettingsFile)' == ''">$(RunSettingsFilePath)</VSTestRunSettingsFile>
 
     <DeployRunSettingsFile Condition="'$(DeployRunSettingsFile)' == ''">true</DeployRunSettingsFile>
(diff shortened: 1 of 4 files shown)


tests/Directory.Build.targets
```diff
@@ -8,6 +8,7 @@
     <VSTestRunSettingsFile Condition="'$(VSTestRunSettingsFile)' == ''">$(RunSettingsFilePath)</VSTestRunSettingsFile>
 
     <DeployRunSettingsFile Condition="'$(DeployRunSettingsFile)' == ''">true</DeployRunSettingsFile>
+    <XunitRunnerJson Condition="'$(XunitRunnerJson)' == '' and '$(ArchiveTests)' == 'true'">$(RepoRoot)tests\helix\xunit.runner.json</XunitRunnerJson>
```
joperezr, line 11: Could you add a comment above here to explain why we need a special one for Helix?

joperezr: Meaning adding something like "In Helix we want to disable test parallelization in order to not hit docker network limit issues bla bla.."

joperezr (approved):
Small nit, but looks good otherwise.

PR #5695, opened 2024-09-12
Fix RID regression by adding a task that calculates the best matching RID for platform

### LINKED ISSUE #5486: [release/8.2] Unable to find package Aspire.Dashboard.Sdk.rhel.8-x64 & Aspire.Hosting.Orchestration.rhel.8-x64

### Describe the bug

`dotnet build` fails to restore Nuget packages after running `dotnet workload update` & `dotnet workload install aspire` on RHEL 8.9

### Expected Behavior

_No response_

### Steps To Reproduce

`dotnet workload update`
`dotnet workload install aspire`
`mkdir foo`
`dotnet new aspire-apphost`

### Exceptions (if any)

The template ".NET Aspire App Host" was created successfully.
This template contains technologies from parties other than Microsoft, see https://aka.ms/dotnet/aspire/8.0-third-party-notices for details.

Processing post-creation actions...
Restoring /app/foo/foo.csproj:
  Determining projects to restore...
/app/foo/foo.csproj : error NU1101: Unable to find package Aspire.Dashboard.Sdk.rhel.8-x64. No packages exist with this id in source(s): nuget.org
/app/foo/foo.csproj : error NU1101: Unable to find package Aspire.Hosting.Orchestration.rhel.8-x64. No packages exist with this id in source(s): nuget.org
  Failed to restore /app/foo/foo.csproj (in 3.64 sec).
Restore failed.
Post action failed.
Manual instructions: Run 'dotnet restore'

### .NET Version info

dotnet --info
.NET SDK:
 Version:           8.0.108
 Commit:            665a05cea7
 Workload version:  8.0.100-manifests.3df3d38d

Runtime Environment:
 OS Name:     rhel
 OS Version:  8
 OS Platform: Linux
 RID:         rhel.8-x64
 Base Path:   /usr/lib64/dotnet/sdk/8.0.108/

.NET workloads installed:
 Workload version: 8.0.100-manifests.3df3d38d
 [aspire]
   Installation Source: SDK 8.0.100
   Manifest Version:    8.2.0/8.0.100
   Manifest Path:       /home/acme/.dotnet/sdk-manifests/8.0.100/microsoft.net.sdk.aspire/8.2.0/WorkloadManifest.json
   Install Type:        FileBased

Host:
  Version:      8.0.8
  Architecture: x64
  Commit:       08338fcaa5

.NET SDKs installed:
  8.0.108 [/usr/lib64/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.8 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.8 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]

### Anything else?

Installed Workload Id      Manifest Version      Installation Source
--------------------------------------------------------------------
aspire                     8.2.0/8.0.100         SDK 8.0.100

Red Hat Enterprise Linux release 8.9 (Ootpa)

### LINKED ISSUE #5674: Property CliPath: the path to the dcp executable used for Aaspire orchestration is required; Property DashboardPath: The path to the Aspire Dashboard binaries is missing

### Describe the bug

accrording to https://learn.microsoft.com/en-us/dotnet/aspire/get-started/build-your-first-aspire-app?pivots=dotnet-cli
1st   dotnet new aspire-starter --use-redis-cache --output AspireSample
2nd  modify the csproj file
3rd  dotnet  restore  and succeeded
4th  donet build  and succeeded
5th  dotnet run --project AspireSample/AspireSample.AppHost

when i run 5th step,the error occured:
Property CliPath: the path to the dcp executable used for Aaspire orchestration is required; Property DashboardPath: The path to the Aspire Dashboard binaries is missing

### Expected Behavior

how can i fix the problem and run normally

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Anything else?

the log content:

when 2nd step i write the aspire1.AppHost.csproj file content:

joperezr:
## Description

In Aspire, we have always just cross-compiled our Dashbarod and DCP  packages for a limited set of RIDs (win-x64, win-x86, win-arm64, linux-x64, linux-arm64, osx-x64, and osx-arm64). When running on platforms that had special RIDs that we weren't cross-compiling for (for example, rhel.8) then we were relying on the SDK's logic to automatically pick the best closest Dashboard and DCP pack for that RID. 

However, in Aspire 8.2.0 we removed Dashboard and DCP from the workload and instead just added some sdk targets that were adding these references on the fly, but we were not doing this special matching logic the the SDK used to do and instead we were simply trying to use the current RID, which caused a regression in all of those platforms that have special RIDs which we don't cross-compile for.

In this PR, we are copying the logic that the SDK does for doing that matching, and we are calling it inside of our targets in order to ensure that we are going to get a valid package reference, which will fix the regression in those systems. This has been manually validated in RedHat 8, which was the original report for this regression. This fix, will get backported into release/8.2 branch, and serviced in 8.2.1.

Fixes https://github.com/dotnet/aspire/issues/5486
Fixes https://github.com/dotnet/aspire/issues/5674

cc: @DamianEdwards @radical @eerhardt @danegsta @dsplaisted 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5695)


danegsta:
Did we consider the naive approach we use in the Directory.Build.props for the Aspire repo to platform and arch values? We basically interrogate `[MSBuild]::IsOsPlatform()` and `[System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture` to build our own RID value.

joperezr:
> Did we consider the naive approach we use in the Directory.Build.props for the Aspire repo to platform and arch values? We basically interrogate `[MSBuild]::IsOsPlatform()` and `[System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture` to build our own RID value.

That is another approach, but feels more like a Band-Aid than an actual fix to me. This will fall back to use the same logic that workloads used to do, which will always be more complete. It also means that it is future-proof as it is using the bundled runtime json file included in the SDK. It does add a bit of complexity, but feels like a more complete fix IMO. Open to feedback as always though 😃

danegsta:
> > Did we consider the naive approach we use in the Directory.Build.props for the Aspire repo to platform and arch values? We basically interrogate `[MSBuild]::IsOsPlatform()` and `[System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture` to build our own RID value.
> 
> That is another approach, but feels more like a Band-Aid than an actual fix to me. This will fall back to use the same logic that workloads used to do, which will always be more complete. It also means that it is future-proof as it is using the bundled runtime json file included in the SDK. It does add a bit of complexity, but feels like a more complete fix IMO. Open to feedback as always though 😃

To flip my question around, is it possible to update that old logic in Directory.Build.props to use this new approach to keep things consistent?

dsplaisted:
Looking at the dependencies this adds, I think it would be better to put the `FindBestRidForPlatform` task or something similar in the .NET SDK.  You have to worry about having the right versions of the dependencies that will load in Visual Studio, including the right dependencies with your task, etc.  This is kind of a pain to manage over time, so I'd recommend taking advantage of the fact that the .NET SDK already has to deal with that rather than dealing with these dependencies yourself.

I understand a reason not to do this is that you want this to run on the .NET 8 SDK which wouldn't have the new task.  I would consider adding the task in .NET 9 and using it from the SDK if running on the .NET 9 SDK.  Then when you move to requiring the .NET 9 SDK, you can drop the extra dependencies and the copy of the SDKs code for handling the RID graph.

eng/BuildTask.targets
```diff
@@ -0,0 +1,100 @@
...
+  <Target Name="_AddBuildOutputToPackageDesktop" DependsOnTargets="Publish" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+    <PropertyGroup>
+      <BuildTaskTargetTfmSpecificFolder Condition="'$(BuildTaskTargetTfmSpecificFolder)' == ''">$(TargetFramework)</BuildTaskTargetTfmSpecificFolder>
+    </PropertyGroup>
+    <ItemGroup>
+      <TfmSpecificPackageFile Include="$(PublishDir)*.*" PackagePath="$(BuildTaskTargetFolder)/$(BuildTaskTargetTfmSpecificFolder)/%(FileName)%(Extension)"/>
+    </ItemGroup>
+  </Target>
+  
+  </Project>
```
radical, line 100:
```suggestion
</Project>
```

src/Aspire.Hosting.Sdk/Aspire.Hosting.Sdk.csproj
```diff
@@ -1,10 +1,19 @@
-<Project Sdk="Microsoft.Build.NoTargets">
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
joperezr: Correct, this will need to happen at the SDK level in order to be able to run pre-restore.

joperezr:
Thanks for the review folks. After chatting with @dsplaisted more on this offline, we agreed that in order to not have to deal  with the task of making sure dependencies used by the task match the ones built-in on the SDK/Visual Studio the best approach should be to run out-of-proc. I will push some more changes here that will essentially move from an msbuild task to a cli tool instead which is invoked by our targets. The long-term goal for 10.0 will be to move this logic into dotnet/sdk instead.

src/Aspire.Hosting.Sdk/Aspire.Hosting.Sdk.csproj
```diff
@@ -1,10 +1,19 @@
...
+    <TargetFrameworks>$(DefaultTargetFramework);$(NetFrameworkToolCurrent);</TargetFrameworks>
     <PackageTags>aspire hosting sdk</PackageTags>
     <Description>.NET Aspire Hosting SDK. Enabled via &lt;IsAspireHost&gt;true&lt;/IsAspireHost&gt;.</Description>
+    <UsePublicApiAnalyzers>false</UsePublicApiAnalyzers>
+    <!--
+      CS1591: Missing XML comment for publicly visible type or member 'Type_or_Member'
+      CS8002: Referenced assembly does not have a strong name
+    -->
+    <NoWarn>$(NoWarn);CS1591;CS8002</NoWarn>
+    <AppendTargetFrameworkToOutputPath>true</AppendTargetFrameworkToOutputPath>
```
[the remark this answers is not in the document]
joperezr: This project imports Wokload.targets which defaults this to false, hence the need to flip it  back here.

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Aspire.RuntimeIdentifier.Tool.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: Nesting .csprojs under folders that already contain a .csproj is typically not a good idea.

joperezr: Agreed, though in this case the parent project is a `NoTargets` SDK project meaning it doesn't compile anything, and the child project doesn't have any targets that could be packed by the parent by accident either. Main reason why I made it a child project, is that the parent is the one that packs this. I could change it of course, but this "feels" better. There is also precedent in dotnet/runtime where source generator projects are nested under the package that packs them.

joperezr: For example: https://github.com/dotnet/runtime/tree/main/src/libraries/System.Text.Json/gen

eerhardt: in dotnet/runtime, it is the `src` project that packs them, which is a sibling, not in a parent directory.

joperezr: Oh -- fair enough. I don't feel strongly about this but I can change it if you want me to.

eng/dashboardpack/Sdk.in.targets
```diff
@@ -2,15 +2,15 @@
 
   <!-- *** BEGIN *** -->
 
-  <PropertyGroup Condition="'$(NETCoreSdkRuntimeIdentifier)' == '@PlatformRuntime@'">
+  <PropertyGroup>
```
eerhardt, line 5: Do we still need this to be an `in.targets` file anymore now that we are removing these?

eng/dcppack/Aspire.Hosting.Orchestration.in.targets
```diff
@@ -1,6 +1,6 @@
 <Project>
 
-  <PropertyGroup Condition="'$(NETCoreSdkRuntimeIdentifier)' == '@PlatformRuntime@'">
+  <PropertyGroup>
```
eerhardt, line 3: Same question - does this file still need to be `in.targets`?

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+sealed class Program
+{
+    static int Main(string[] args)
+    {
+        CliRootCommand rootCommand = new("Aspire.RuntimeIdentifier.Tool v" + FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).ProductVersion)
+        {
+            TreatUnmatchedTokensAsErrors = true
+        };
+
+        CliOption<string?> runtimeGraphPathOption = new("--runtimeGraphPath", "-rgp")
```
eerhardt, line 22:
```suggestion
        CliOption<string?> runtimeGraphPathOption = new("--runtimeGraphPath")
```

I would drop these "short" option names. We don't need 2 of them, and they aren't following Unix conventions (short options names should only have a single letter so you can chain them together - like `git clean -xdf` sets `-x` and `-d` and `-f`).

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            Console.WriteLine(bestRidForPlatform);
+            return 0;
+        });
+
+        return rootCommand.Parse(args).Invoke();
+    }
+
+    private static string[]? ParseSupportedRidsArgument(ArgumentResult result)
```
eerhardt, line 77: I'm surprised we need to write custom code for this.

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            Description = "RID to use for finding the best applicable RID from mapping.",
+            Required = true
+        };
+
+        CliOption<string[]> supportedRidsOption = new("--supportedRids", "-sr")
+        {
+            Description = "List of RIDs that are supported. Comma-separated.",
+            Required = true,
+            AllowMultipleArgumentsPerToken = true,
```
eerhardt, line 38: I don't think we need this. From the docs:

```
        /// <example>
        /// If set to <see langword="true"/>, the following command line is valid for passing multiple arguments:
        /// <code>
        /// > --opt 1 2 3
        /// </code>
        /// The following is equivalent and is always valid:
        /// <code>
        /// > --opt 1 --opt 2 --opt 3
        /// </code>
        /// </example>
```

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            var supportedRids = parseResult.GetValue(supportedRidsOption);
+
+            string? bestRidForPlatform = NuGetUtils.GetBestMatchingRid(graph, ridToUse!, supportedRids!, out bool wasInGraph);
+
+            if (!wasInGraph)
+            {
+#pragma warning disable CA2201 // Do not raise reserved exception types
+                throw new ApplicationException("Unable to find the best rid to use");
+#pragma warning restore CA2201 // Do not raise reserved exception types
```
eerhardt, line 67:
```suggestion
                throw new InvalidOperationException("Unable to find the best rid to use");
```

eerhardt: Or probably even better:

```suggestion
                Console.WriteLine("Unable to find the best rid to use");
                return -1;
```

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Aspire.RuntimeIdentifier.Tool.csproj
```diff
@@ -0,0 +1,16 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <RollForward>Major</RollForward>
+    <UsePublicApiAnalyzers>false</UsePublicApiAnalyzers>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
eerhardt, line 9: We probably want to disable the app host `app.exe` file since we are never going to use it, and it won't work on RIDs that aren't what we built this for.

```suggestion
    <UseAppHost>false</UseAppHost
  </PropertyGroup>
```

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
joperezr: I followed precedent that has been done on other tools: https://github.com/dotnet/sdk/blob/f007eb76461e4fff93c54e3ea2af67ad4eefeb32/src/Compatibility/ApiCompat/Microsoft.DotNet.ApiCompat.Tool/Program.cs#L362-L371

src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
joperezr: I remember trying without this and things didn't work, so not sure what was wrong in that case.

eerhardt (approved):
LGTM. Thanks!

Issue #5645, opened 2024-09-10
WaitFor: Add health checks to remaining resources


mitchdenny:
With the merge of #5394 and some of the subsequent enhancements around health check integration (#5515) its now time to scale out the support for WaitFor for all of our resources. Essentially each `AddXYZ` method for a resource needs to register a health check with DI and add a matching `HealthCheckAnnotation`. Here are two examples:

**Redis**

https://github.com/dotnet/aspire/blob/1b2e640a366afd68c99b915762416774b8a9be91/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs#L36-L57

**Postgres**

https://github.com/dotnet/aspire/blob/1b2e640a366afd68c99b915762416774b8a9be91/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs#L38-L98

# Remaining resources

## Pri0

- [x] RabbitMQ (done #5718)
- [x] MongoDB (done: #5697)
- [x] Kafka (done #5719)
- [x] MySql (done: #5705)
- [x] SQL Server (done: #5669)
- [x] Azure Cosmos DB emulator (done: #5729)
- [x] Elastic Search (done: #5725)
- [x] Oracle (done: #5734)
- [x] PgAdmin (done: #6081)

## P1

- [x] Azure EventHubs emulators (done: #6079)
- [x] Azure Storage emulator (done: #5761)
- [x] Garnet (done: #5698)
- [x] ValKey (done: #5706)
- [x] Milvus (done: #5707)
- [x] Nats (done: #5753)

Some of these resources don't have health checks and possibly won't need them but we should go through the list and very each one and create issues. We also need test coverage for the WaitFor behavior to verify that it is working as expected for each resource (example):

https://github.com/dotnet/aspire/blob/1b2e640a366afd68c99b915762416774b8a9be91/tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs#L21-L68

Alirexaa:
@mitchdenny , I could help you to completing this one.

davidfowl:
Lets do the azure resources last.

mitchdenny:
@Alirexaa of course! Any help is welcome with this.

Alirexaa:
I will work on Valkey

Alirexaa:
I will work on Milvus.

davidfowl:
Kakfa is high on the priority list.

davidfowl:
I sorted the list in priority order

[...]

PR #5707, opened 2024-09-13
WaitFor Milvus

Alirexaa:
Adds a health check to the host for Milvus resource so that it works with WaitFor.

Related #5645

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5707)

### DIFF
--- a/src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
@@ -4,6 +4,9 @@
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Milvus;
 using Aspire.Hosting.Utils;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Milvus.Client;
 
 namespace Aspire.Hosting;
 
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -51,6 +54,38 @@ public static IResourceBuilder<MilvusServerResource> AddMilvus(this IDistributed
...
+            foreach (var databaseName in milvus.Databases)
+            {
+                if (!lookup.TryGetValue(databaseName.Key, out var databaseResource))
+                {
+                    throw new DistributedApplicationException($"Database resource '{databaseName}' under Milvus Server resource '{milvus.Name}' was not found in the model.");
+                }
+                var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(databaseResource, @event.Services);
+                await builder.Eventing.PublishAsync<ConnectionStringAvailableEvent>(connectionStringAvailableEvent, ct).ConfigureAwait(false);
+                var beforeResourceStartedEvent = new BeforeResourceStartedEvent(databaseResource, @event.Services);
+                await builder.Eventing.PublishAsync(beforeResourceStartedEvent, ct).ConfigureAwait(false);
```
davidfowl, line 77:
```suggestion
                var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(databaseResource, @event.Services);
                await builder.Eventing.PublishAsync<ConnectionStringAvailableEvent>(connectionStringAvailableEvent, ct).ConfigureAwait(false);


                var beforeResourceStartedEvent = new BeforeResourceStartedEvent(databaseResource, @event.Services);
                await builder.Eventing.PublishAsync(beforeResourceStartedEvent, ct).ConfigureAwait(false);
```
```suggestion
                var connectionStringAvailableEvent = new ConnectionStringAvailableEvent(databaseResource, @event.Services);
                await builder.Eventing.PublishAsync<ConnectionStringAvailableEvent>(connectionStringAvailableEvent, ct).ConfigureAwait(false);
                var beforeResourceStartedEvent = new BeforeResourceStartedEvent(databaseResource, @event.Services);
                await builder.Eventing.PublishAsync(beforeResourceStartedEvent, ct).ConfigureAwait(false);
```

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -190,4 +247,27 @@ private static void ConfigureAttuContainer(EnvironmentCallbackContext context, M
...
+    private static async Task<HealthCheckResult> HealthCheck(string? connectionString)
+    {
+        if (connectionString is null)
+        {
+            throw new InvalidOperationException("Connection string is unavailable");
+        }
+
+        try
+        {
+            var milvusClient = new MilvusClient(connectionString);
```
davidfowl, line 260: Maybe the client should be created once? Do we want to have one client per call to health check? These run every 5 seconds...

davidfowl:
Looks like the health check isn't working

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -190,4 +259,44 @@ private static void ConfigureAttuContainer(EnvironmentCallbackContext context, M
         // This will need to be refactored once updated service discovery APIs are available
         context.EnvironmentVariables.Add("MILVUS_URL", $"{resource.PrimaryEndpoint.Scheme}://{resource.Name}:{resource.PrimaryEndpoint.TargetPort}");
     }
+    internal static MilvusClient CreateMilvusClient(IServiceProvider sp, string? connectionString)
+    {
+        if (connectionString is null)
```
davidfowl, line 264: Ideally we would share this logic: https://github.com/dotnet/aspire/blob/65c69024e5ebfe0c7914e6fbcbfea12aa359bd71/src/Components/Aspire.Milvus.Client/MilvusClientSettings.cs#L11

Alirexaa: I agree with @davidfowl we should do this because the created Milvus connection string in app-host isn't compatible with MilvusClient and those properties (database, endpoint, key) should be extracted before creating MilvusClient.

davidfowl (approved):

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
```diff
@@ -51,6 +57,43 @@ public static IResourceBuilder<MilvusServerResource> AddMilvus(this IDistributed
 
         var milvus = new MilvusServerResource(name, apiKeyParameter);
 
+        string? connectionString = null;
```
mitchdenny, line 60: Probably technically don't need the connection string here since you have the client.

mitchdenny (approved):

src/Aspire.Hosting.Milvus/MilvusBuilderExtensions.cs
eerhardt: Was there a reason this wasn't done?

PR #5712, opened 2024-09-14
Enable replicaSet support for MongoDb

### LINKED ISSUE #5238: MongoDB  Transactions  requires replicas
Hello,  we are using MongoDB and recently we have integrated with aspire.  
However, running MongoDB transactions requires replicas.  As I checked I couldn't find any solution right now.

Can you assist me if its possible please?

twsouthwick:
## Description

This change adds support for replica sets in MongoDb. In order to support this a couple changes are required:

- Connection strings have query parameters as options that define the replica set. To support this, the expression builder is now called manually to build it up, as well as changing the database connection string to no long reference the server connection string directly (as the server connection string may have the replicaSet option)
- The target port and the exposed port must be the same for replica sets to work. This adds a check that updates the MongoDb server to not be proxied and have the same port. Ideally, if no port is set, we should use a random one but I couldn't figure out how to do that in an idiomatic way.

Fixes #5238

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5712)


src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -230,6 +165,83 @@ public static IResourceBuilder<MongoDBServerResource> WithInitBindMount(this IRe
...
+    /// Adds a replica set to the MongoDB server resource.
+    /// </summary>
+    /// <param name="builder">The MongoDB server resource.</param>
+    /// <param name="replicaSetName">The name of the replica set. If not provided, defaults to <c>rs0</c>.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<MongoDBServerResource> WithReplicaSet(this IResourceBuilder<MongoDBServerResource> builder, string? replicaSetName = null)
+    {
+        if (builder.Resource.TryGetLastAnnotation<MongoDbReplicaSetAnnotation>(out _))
+        {
+            throw new InvalidOperationException("A replica set has already been added to the MongoDB server resource.");
```
davidfowl, line 178: Why throw instead of noop?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -230,6 +165,83 @@ public static IResourceBuilder<MongoDBServerResource> WithInitBindMount(this IRe
...
+            }
+
+            throw new InvalidOperationException("No endpoint found for the MongoDB server resource.");
+        }
+
+        // See the conversation about setting up replica sets in Docker here: https://github.com/docker-library/mongo/issues/246
+        static string GetReplicaSetInitDockerfileDir(string replicaSet, string host, int port)
+        {
+            var dir = Path.Combine(Path.GetTempPath(), "aspire.mongo", Path.GetRandomFileName());
+            Directory.CreateDirectory(dir);
```
davidfowl, line 227: Directory.CreateTempSubdirectory()

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -247,4 +259,63 @@ private static void ConfigureMongoExpressContainer(EnvironmentCallbackContext co
         context.EnvironmentVariables.Add("ME_CONFIG_MONGODB_URL", sb.ToString());
         context.EnvironmentVariables.Add("ME_CONFIG_BASICAUTH", "false");
     }
+
+    /// <summary>
+    /// Same as <see cref="ResourceBuilderExtensions.WaitFor{T}(IResourceBuilder{T}, IResourceBuilder{IResource})"/> but with a few options we need.
+    /// </summary>
+    private static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, IResourceBuilder<IResource> dependency, bool includeHealthChecks = true) where T : IResource
```
davidfowl, line 266: I'd rather make it possible to disable the health check in the WaitFor call than to copy all of this source code.

twsouthwick: Yeah, I agree - I can make that change as part of this

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
twsouthwick: cool - never seen that api

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
twsouthwick: yeah. I can check if there's already one with the same name and just operate on that if you'd like

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -161,11 +165,157 @@ public static IResourceBuilder<MongoDBServerResource> WithInitBindMount(this IRe
...
+            throw new InvalidOperationException("No endpoint found for the MongoDB server resource.");
+        }
+
+        // See the conversation about setting up replica sets in Docker here: https://github.com/docker-library/mongo/issues/246
+        static string GetReplicaSetInitDockerfileDir(string replicaSet, string host, int port)
+        {
+            var dir = Path.Combine(Path.GetTempPath(), "aspire.mongo", Path.GetRandomFileName());
+            Directory.CreateDirectory(dir);
+
+            var rsInitContents = $$"""rs.initiate({ _id:'{{replicaSet}}', members:[{_id:0,host:'localhost:{{port}}'}]})""";
```
twsouthwick, line 229: @davidfowl I'd prefer to be able to just docker exec into the container after it started, but couldn't figure out how to do that so I'm running a container and configuring things to ensure it completes before anything else needs to use the db. If there is a better way to do this with aspire things, let me know

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
```diff
@@ -21,9 +21,61 @@ public class MongoDBServerResource(string name) : ContainerResource(name), IReso
...
+            AppendConnectionString(builder);
+            AppendSuffix(builder);
+
+            return ReferenceExpression.Create(builder);
+        }
+    }
+
+    internal void AppendConnectionString(in ReferenceExpression.ExpressionInterpolatedStringHandler builder)
+    {
+        builder.AppendLiteral("mongodb://");
```
davidfowl, line 41: There's a type called [ReferenceExpressionBuilder](https://learn.microsoft.com/en-us/dotnet/api/aspire.hosting.applicationmodel.referenceexpressionbuilder?view=dotnet-aspire-8.0) that you can use instead.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
davidfowl: We're adding support for this in Aspire 9 very soon. Might be better to wait until that support is available in the app model.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -632,7 +663,7 @@ public static IResourceBuilder<T> WaitFor<T>(this IResourceBuilder<T> builder, I
 
             // If our dependency resource has health check annotations we want to wait until they turn healthy
             // otherwise we don't care about their health status.
-            if (dependency.Resource.TryGetAnnotationsOfType<HealthCheckAnnotation>(out var _))
+            if (includeHealthChecks && dependency.Resource.TryGetAnnotationsOfType<HealthCheckAnnotation>(out var _))
```
eerhardt, line 666: This probably deserves a test that calls this with `includeHealthChecks: false` and makes sure it works correctly.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -161,11 +163,96 @@ public static IResourceBuilder<MongoDBServerResource> WithInitBindMount(this IRe
...
+                }
+            }
+
+            throw new InvalidOperationException("No endpoint found for the MongoDB server resource.");
+        }
+
+        // See the conversation about setting up replica sets in Docker here: https://github.com/docker-library/mongo/issues/246
+        static string GetReplicaSetInitDockerfileDir(string replicaSet, string host, int port)
+        {
+            var dir = Directory.CreateTempSubdirectory("aspire.mongo").FullName;
```
eerhardt, line 224: A couple notes here:

1. Who cleans up this directory?
2. On Unix, this is going to create a directory that only the current user can access. If the container is running on non-root, the container won't be able to access this folder.

src/Aspire.Hosting.MongoDB/MongoDBDatabaseResource.cs
```diff
@@ -27,7 +27,17 @@ public MongoDBDatabaseResource(string name, string databaseName, MongoDBServerRe
     /// Gets the connection string expression for the MongoDB database.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression
-        => ReferenceExpression.Create($"{Parent}/{DatabaseName}");
+    {
+        get
+        {
+            var builder = new ReferenceExpression.ExpressionInterpolatedStringHandler(10, 2);
```
eerhardt, line 33: Why `10`?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
twsouthwick: cool - is there a tracking issue for that?

src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs
twsouthwick: oh much cleaner. thanks

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
twsouthwick: is there a pattern that already exists to handle creating/cleaning up directories? Maybe another reason as @davidfowl mentioned above to maybe wait to exec into a container

src/Aspire.Hosting.MongoDB/MongoDBDatabaseResource.cs
twsouthwick: out of date now because I learned about the `ReferenceExpressionBuilder` but it was the length of the scheme and a bit more for a server.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
twsouthwick: added

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -94,7 +95,8 @@ public static IResourceBuilder<T> WithMongoExpress<T>(this IResourceBuilder<T> b
                                                         .WithImageRegistry(MongoDBContainerImageTags.MongoExpressRegistry)
                                                         .WithEnvironment(context => ConfigureMongoExpressContainer(context, builder.Resource))
                                                         .WithHttpEndpoint(targetPort: 8081, name: "http")
-                                                        .ExcludeFromManifest();
+                                                        .ExcludeFromManifest()
+                                                        .WaitFor(builder);
```
mitchdenny, line 99: I'm not sure about using `WaitFor` invisibly like this. I can see the benefit but if one of the mongo db instances files to start then you won't be able to inspect any of the other databases.

twsouthwick: I added that because once you have a replica set, the existing connection string may cause express to fail if it tries to connect before the replica set is initialized (a race condition) and it doesn't recover from it.

creyke:
+1 for this. Missing replica sets doesn't just block transactions, it also blocks change streams.

These are two very commonly utilised features.

This looks close to done, but is not active... is this likely to get merged @twsouthwick?

Thanks in advance.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
twsouthwick: @davidfowl any progress on being able to exec into the container?

twsouthwick:
@creyke I'd like to get this in, but in order to set up the replica set, we must run some commands in the container. sounds like docker exec support will be available soon, so this is waiting on that. Once that is available, I'll update it so this can be merged.

feritzcan2:
> @creyke I'd like to get this in, but in order to set up the replica set, we must run some commands in the container. sounds like docker exec support will be available soon, so this is waiting on that. Once that is available, I'll update it so this can be merged.

Still no luck? Thats highly blocker for integration testing /(

alexjamesbrown:
@twsouthwick was there anything left to do on this? Currently using the workaround on this issue
https://github.com/dotnet/aspire/issues/6811 - but definitely feels like this support should be in Aspire core.

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
```diff
@@ -161,11 +163,96 @@ public static IResourceBuilder<MongoDBServerResource> WithInitBindMount(this IRe
...
+                ENTRYPOINT ["mongosh", "--port", "{port}", "--host", "{host}", "rs.js"]
+                """);
+            return dir;
+        }
+    }
+
     private static void ConfigureMongoExpressContainer(EnvironmentCallbackContext context, MongoDBServerResource resource)
     {
-        // Mongo Exporess assumes Mongo is being accessed over a default Aspire container network and hardcodes the resource address
+        var sb = new StringBuilder($"mongodb://{resource.Name}:{resource.PrimaryEndpoint.TargetPort}/?directConnection=true");
```
davidfowl, line 243: ReferenceExpressionBuilder

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
davidfowl: So after some thought, I think a pattern we can use here is:
- Put the docker file in the nuget package
- Mark it as content so it gets copied to the output (we'lll want to name it so it doesn't conflict)
- Use build arguments to parameterize the docker file and pass that state in

danmoseley:
@twsouthwick is this still in progress?

davidfowl:
@twsouthwick @eerhardt and I have an offline thread about this one.

alexjamesbrown:
Just commenting again as not to close this issue

@davidfowl / @eerhardt  was there any update on this at all?

src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs
eerhardt: With #7136, we have now introduced a new "IAspireStore" API:

https://github.com/dotnet/aspire/blob/acdf5012407c2ab0259f4b6f075acb6c668949f3/src/Aspire.Hosting/ApplicationModel/IAspireStore.cs#L17-L22

This file could be written there and my above concerns would be resolved.

twsouthwick:
I took a stab at updating it, but have found that since I did the first attempt that, password auth has been added. That adds a complication for enabling replica sets as that then requires a certificate to be used as well.

The place I'm getting stuck on is that I can generate a cert, but getting it into the container is weird.

- If I do it as one of the mongo init scripts, I can generate it, but then I can't set the permissions correctly
- I've seen people attempt to do it by doing it in the entrypoint (i.e. `[some commands to create cert and set permissions]; call_original_entrypoint.sh $@`)  but it appears the Aspire doesn't allow setting an entrypoint to an arbitrary command like that.

danmoseley:
next action here seems to be a question for @eerhardt and I know he's pretty engaged right now.

eerhardt:
@twsouthwick @davidfowl and I met offline and had some ideas how to move this forward. We think if we had something like https://github.com/dotnet/aspire/pull/6481, we could make this simpler.

danmoseley:
OK I'm going to close this as it sounds like we're blocked on that. Feel free to reopen, or create a new PR, when ready to continue @twsouthwick

Issue #5700, opened 2024-09-13
Flaky tests in `Aspire.Dashboard.Tests.TelemetryRepositoryTests.LogTests`


radical:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=807275
Build error leg or test failing: Aspire.Dashboard.Tests.TelemetryRepositoryTests.LogTests.Subscription_MultipleUpdates_MinExecuteIntervalApplied
Pull request: https://github.com/dotnet/aspire/pull/5684

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "Elapsed time .* should be greater than min interval",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=807275
**Error message validated:** `[Elapsed time .* should be greater than min interval`]
**Result validation:** :white_check_mark: Known issue matched with the provided build.
**Validation performed at:** 9/13/2024 6:15:01 AM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[809118](https://dev.azure.com/dnceng-public/public/_build/results?buildId=809118)|dotnet/aspire|[Aspire.Dashboard.Tests.TelemetryRepositoryTests.LogTests.Subscription_MultipleUpdates_MinExecuteIntervalApplied](https://dev.azure.com/dnceng-public/public/_build/results?buildId=809118&view=ms.vss-test-web.build-test-results-tab&runId=20932568&resultId=100301)|dotnet/aspire#5719|
|[2539220](https://dev.azure.com/dnceng/internal/_build/results?buildId=2539220)|dotnet-aspire|[Aspire.Dashboard.Tests.TelemetryRepositoryTests.LogTests.Subscription_MultipleUpdates_MinExecuteIntervalApplied](https://dev.azure.com/dnceng//internal/_build/results?buildId=2539220&view=ms.vss-test-web.build-test-results-tab&runId=54196671&resultId=100412)||
[...]

PR #5720, opened 2024-09-15
Record time in callback and more logging in min interval test

JamesNK:
More for https://github.com/dotnet/aspire/issues/5700

Maybe dispatching the channel read/writes is interfering with the time? Track elapsed time in callback.
Also, more logging.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5720)


mitchdenny (approved):

davidfowl:
We should still disable the test.

JamesNK:
Failure only happens in CI and as far as I know there is no quartine run of tests. I don't how to test if it is fixed unless it is enabled.

davidfowl:
By making a PR where the test is enabled and iterating on it there.

JamesNK:
The test has a 95% pass rate. I can easily make a PR where the test passes. For example, this PR.

davidfowl:
Anecdotally if it affects more than one PR then we need to disable it.

PR #5725, opened 2024-09-15
WaitFor for Elasticsearch

Alirexaa:
## Description

Adds `WaitFor` support for Elasticsearch

Related: #5645

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5725)


mitchdenny:
Looks good to me (based on our current pattern).

mitchdenny (approved):

davidfowl:
The timeout in the test might be too low. I don't know how long elastic search takes to boot but its notoriously slow and unreliable from our other tests?

davidfowl:
cc @radical note the changes

tests/helix/send-to-helix-basictests.targets
```diff
@@ -32,7 +32,7 @@
       <!-- runsettings timeout in ms, default to 15 mins -->
       <_DefaultWorkItems TimeoutMs="900000" />
 
-      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Hosting.Elasticsearch.Tests'" TimeoutMs="1200000" />
+      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Hosting.Elasticsearch.Tests'" TimeoutMs="3600000" />
```
eerhardt, line 35: Is this really an hour? We are going to wait an hour for these tests? Isn't that just going to make us wait longer when the thing isn't responsive?

radical: This will become a problem whenever helix is backed up. The total timeout for the whole job is 90mins, so this will definitely timeout. Do we need to break this up?
https://github.com/dotnet/aspire/blob/2344aaf77fadfecd3186a42205882bea56885b8a/eng/pipelines/azure-pipelines-public.yml#L102-L103

radical: The test run on this PR took ~8mins. If this is taking up to an hour sometimes then maybe we need to investigate why that is the case.

PR #5739, opened 2024-09-16
[ci] Fix codeql build

### LINKED ISSUE #5738: [CI] `codeql` build broken
```
  You must install or update .NET to run this application.
  App: D:\a\_work\1\s\artifacts\bin\Aspire.Workload.Tests\Release\net8.0\testhost.exe
  Architecture: x64
  Framework: 'Microsoft.NETCore.App', version '8.0.0' (x64)
  .NET location: C:\hostedtoolcache\windows\dotnet
  The following frameworks were found:
    9.0.0-preview.7.24405.7 at [C:\hostedtoolcache\windows\dotnet\shared\Microsoft.NETCore.App]
```

radical:
Avoid doing test specific tasks when not building to run tests.

Fixes https://github.com/dotnet/aspire/issues/5738 .
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5739)


radical:
Validation build - https://dev.azure.com/dnceng/internal/_build/results?buildId=2539843

tests/Directory.Build.targets
```diff
@@ -35,7 +35,7 @@
 
   <!-- Used for running one helix job per test class -->
   <Target Name="_ExtractTestClassNames"
-          Condition="'$(ExtractTestClassNamesForHelix)' == 'true'"
+          Condition="'$(ExtractTestClassNamesForHelix)' == 'true' and '$(ArchiveTests)' == 'true'"
```
eerhardt, line 38: Why are we building anything in the `tests` directory in this build configuration? If we don't want to do "test things", do we need to build the test assemblies?

radical: That's a good question. AFAICS, even if we have `-p:Tests=false` like used for the codeql build, it still builds the test assemblies because they are referenced from the sln.

eerhardt: I didn't think it uses the sln, but instead uses:

https://github.com/dotnet/aspire/blob/a8b0bed9bcf0e09ca6274e08b2ecdc145ae6df81/eng/Build.props#L2-L8

radical: Oh then we should definitely make building the tests conditional. I'll open an issue to track that work.

radical:
@joperezr this needs a review.

eerhardt (approved):
This is good enough to unblock the codeql build. But I'd like to understand why we are building tests at all in that build.

radical:
> This is good enough to unblock the codeql build. But I'd like to understand why we are building tests at all in that build.

Opened this to track work to make building the tests conditional - https://github.com/dotnet/aspire/issues/5769

PR #5750, opened 2024-09-17
Add metric highlights

### LINKED ISSUE #5287: Create DashpageChart component to render a single chart in a dashpage

### LINKED ISSUE #5289: Populate dashpages from configuration

### LINKED ISSUE #5295: Dashpages with no available metrics should not be shown in the navigation tree
The list of available dashpages in the Metrics UI is dynamic, depending on the selected resource.

A dashpage will only be shown if:

- metrics are available for at least one chart, and
- metrics are available for all "required" charts.

### LINKED ISSUE #5298: Improve deep linking to dashpages
Currently it's possible to deep link to a metric/instrument:

https://localhost:15887/metrics/resource/frontend?meter=OpenTelemetry.Instrumentation.Runtime&instrument=process.runtime.dotnet.assemblies.count

We need the same ability to link to dashpages.

Per #5297, this should also work with global dashpages.

### LINKED ISSUE #5303: Define a default set of dashpages to ship with the dashboard
The dashboard will ship with a default set of dashpages, configured via a `dashpages.json` file (per #5289 and #5302).

We need to define that initial set.

@samsp-msft the file format is still in flux, but if you list out the titles and what metrics they contain, I can do the rest.

### LINKED ISSUE #5347: Dashpages UX improvements
Looking at the feature so far, I think we should make some subtle UX changes to the implementation:

- [ ] Create a separate logical section for DashPages from All Metrics
image

- [ ] Remove the description field from each metric and reduce the amount of vertical space in the title of each metric panel

- [ ] Replace the filter panel on each metric with a dropdown in the title bar of each metric view.
image
Clicking the button will show either a popup or side panel with the filter UI for controlling the view.

Q: If you change the filter view for a dashpage - how long does that persist for? If you switch dashpage and come back are the filters still applied? If you switch to another resource will that have the same view. I think today that switching away will reset the views. We should have a decision on the level of persistence, and have an ability to reset the view.

drewnoakes:
## Description

For the last month, @adamint and I have been working on `feature/dashpages`. We're now preparing it for review for merge into `main`.

A dashpage is a composite metric view, displayed in the metrics page when data for the relevant instruments are available.

Here's an example dashpage, showing multiple instruments in separate tiled charts:

image

Currently dashpages are defined in a JSON file that ships with the dashboard. We include a default configuration.

Fixes #5289
Fixes #5295
Fixes #5303
Fixes #5298
Fixes #5287
Fixes #5347

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No, not initially.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5750)


adamint (approved):

maddymontaquila (commented):
Can I get a build of this or at least a demo vid before I give it a formal blessing???

leslierichardson95:
> Can I get a build of this or at least a demo vid before I give it a formal blessing???

@adamint and @drewnoakes if possible, can either of you make it to a UX board review meeting and/or sync with David Culbertson this week?  I wanted to do one earlier but one of us was always OOF in the past few weeks. :)

JamesNK:
I haven't looked at any code yet. This is from testing.

Add a spacer between view dropdown and duration dropdown:
image
I think other toolbars use FluentDividers.

I expected the filters popup to be closable by clicking anyway on the page outside of the popup. This is how popups work everywhere else in Aspire (and web pages in general). However, only clicking in the panel will close the popup:
dash-close-filter-popup

Multiple popups can be opened at the same time (would be fixed by fixing the issue above, because clicking outside of the popup would close it). Also, switching between views leaves the popup open (also would be fixed by the above):
dash-popup-stays-open

Some other state is preserved when switching between views. "Show count" stays enable.

I don't know what happened, but while testing one of the charts stopped updating:
dash-frozen
http.client.request.duration has frozen.

It might be related to this error in the console:
```
fail: Aspire.Hosting.Dashboard.Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost[111]
      Unhandled exception in circuit 'hYYAo8UMq3S58Ff2g215fTIFV_By0GPW3yIyuAYCwjo'.
      System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
         at System.Collections.Generic.List`1.Enumerator.MoveNext()
         at Aspire.Dashboard.Model.InstrumentViewModel.UpdateDataAsync(OtlpInstrumentSummary instrument, List`1 matchedDimensions) in C:\Development\Source\aspire\src\Aspire.Dashboard\Model\InstrumentViewModel.cs:line 23
         at Aspire.Dashboard.Components.ChartContainer.UpdateInstrumentDataAsync(OtlpInstrumentData instrument) in C:\Development\Source\aspire\src\Aspire.Dashboard\Components\Controls\Chart\ChartContainer.cs:line 113
         at Aspire.Dashboard.Components.ChartContainer.UpdateDataAsync() in C:\Development\Source\aspire\src\Aspire.Dashboard\Components\Controls\Chart\ChartContainer.cs:line 93
         at Aspire.Dashboard.Components.ChartContainer.DisposeAsync() in C:\Development\Source\aspire\src\Aspire.Dashboard\Components\Controls\Chart\ChartContainer.cs:line 70
         at Microsoft.AspNetCore.Components.RenderTree.Renderer.<>c__DisplayClass85_0.<<Dispose>g__HandleAsyncExceptions|1>d.MoveNext()
```

Opening the attribute popup inside the filter popup doesn't work:
image

I think there has been a regression on standalone metrics pages. The graph extends to fill the entire page. I believe it use to have a max width:
image

I see the view title is in the query string. Is there some kind of internal identifer for views? It should be what is added to the query string. If there isn't an internal identifer for views then there probably should be one. For example, a view has a `Name` and a `DisplayName`. That way people can give views a unique identifer without worrying about conflicts, e.g. two people adding views titled "Database":
image

This button for a chart has filters and options. I don't think showing the filter button (highlighted in green) is right, specially since we're using it for another reason on the same screen. Since the popup is a mix of graph options, I think it should be a cog. A cog makes it clear that you're configuring the graph options.
image

The graph panels are hard up against the bottom of the page. The spacing between rows of panels should be added to the bottom:
image

adamint:
> How are dashpages sorted in the tree? Meters and instruments are alphabetical. Should dashpages be alphabetical?

We are displaying dashpages in the order that they are defined in `dashpages.json`. We could do either

adamint:
> An instrument page shows the instrument title and description. A dashpage doesn't, and you can only see what page you're on by looking at the tree. Should the dashpage title (and a description if there is one) be displayed? That would make it more consistent with other pages in the metrics area. And if we're going for a Grafana like feel, Grafana also displays the title at the top of a dashboard.

Previously, the title was just shown on mobile, now it's shown on desktop too. How do you like this?

image

adamint:
> Clicking on dashpages menu option doesn't show anything, just the default "Select an instrument". That doesn't make sense in this situation. It should show a list of pages like clicking on a meter does for instruments:

image

I reused our existing data grid for this!

adamint:
> The graph panels are hard up against the bottom of the page. The spacing between rows of panels should be added to the bottom:

Fixed. 
image

> This button for a chart has filters and options. I don't think showing the filter button (highlighted in green) is right, specially since we're using it for another reason on the same screen. Since the popup is a mix of graph options, I think it should be a cog. A cog makes it clear that you're configuring the graph options.

Changed to a cog

adamint:
> I see the view title is in the query string. Is there some kind of internal identifer for views? It should be what is added to the query string. If there isn't an internal identifer for views then there probably should be one. For example, a view has a Name and a DisplayName. That way people can give views a unique identifer without worrying about conflicts, e.g. two people adding views titled "Database":

Currently, there is no internal identifier for views. That display name is also unique

adamint:
> http.client.request.duration has frozen. Closing and opening the filter popup causes values to update. I'm guessing an error was thrown and the timer that regularly updates data and renders stopped working. Closing and opening the filter causes a render and so data updates once.

I was unable to reproduce this issue. Do you have repro steps?

adamint:
> Add a spacer between view dropdown and duration dropdown:

Done. 
image

> Some other state is preserved when switching between views. "Show count" stays enable.

I was able to reproduce this, now show count works properly inside the card.

adamint:
> I think there has been a regression on standalone metrics pages. The graph extends to fill the entire page. I believe it use to have a max width:

This has now been fixed. The new resize observer was resizing the plot to be larger than the maximum width we had set.
image

JamesNK:
> Currently, there is no internal identifier for views. That display name is also unique

Please add one. Even if titles are unique now, this is going in the URL so there is the chance that people will bookmark it. It doesn't seem like a difficult change and it helps [avoid the URL changing in the future](https://www.w3.org/Provider/Style/URI).

JamesNK:
> I think this due to how cards work, and thus can't be changed without adding a document click listener. Is this a correct assumption @vnbaaij?

What are you using cards for that is special? If cards are a problem, why not use a regular div and apply the same styling?

JamesNK:
> I was unable to reproduce this issue. Do you have repro steps?

No. It looks like a thread safety problem: collection being modified at the same time as it is being enumerated

vnbaaij:
> > I think this due to how cards work, and thus can't be changed without adding a document click listener. Is this a correct assumption @vnbaaij?
> 
> What are you using cards for that is special? If cards are a problem, why not use a regular div and apply the same styling?

I generally don't use cards for a popup. For DataGrid resize popup we indeed just use a div with some script to detect clicking outside.

adamint:
> Multiple popups can be opened at the same time (would be fixed by fixing the issue above, because clicking outside of the popup would close it). Also, switching between views leaves the popup open (also would be fixed by the above):

This no longer seems possible

adamint:
> > Currently, there is no internal identifier for views. That display name is also unique
> 
> Please add one. Even if titles are unique now, this is going in the URL so there is the chance that people will bookmark it. Changing it to not use the title will break URLs.
> 
> It doesn't seem like a difficult change and it helps [avoid the URL changing in the future](https://www.w3.org/Provider/Style/URI).

Added

PR #5773, opened 2024-09-18
Remove SDK Workload from Aspire 9.0

### LINKED ISSUE #5444: Remove Aspire.Hosting.SDK from the Workload
For 9.0, we want to be able to build Aspire 9.0 projects without having to install the SDK Workload, and the one remaining piece for that to be possible is to remove the Aspire.Hosting.SDK from the workload. We will then have to ship a standalone MSBuild SDK which will be referenced by AppHost projects which will be the new way of getting this logic to projects, since the workload will not be bringing it any longer.

joperezr:
## Description

One of the goals for Aspire 9.0 is to no longer require an SDK Workload to be installed in order to create/build/run .NET Aspire applications, especially given after 8.2.0 all it has is the .NET Aspire templates (which can also be inserted into Visual Studio or installed separately) and the Aspire.Hosting.SDK (which could also just be an SDK reference in AppHost projects). Given that, with Aspire 9.0 we are no longer planning to ship a workload, and instead just slightly modify how the AppHost projects reference the SDK in order to allow consumers to build and run applications without needing to install a workload.

This PR contains the required changes to the templates, as well as it removes the production of the Aspire workload and deals with all of the fallout from that. The goal is to merge this in time for .NET Aspire's 9.0 RC1 release later this month, so that we can better test and ensure everything is working as expected. 

cc: @radical @eerhardt @DamianEdwards @maddymontaquila @mitchdenny @davidfowl 

Fixes https://github.com/dotnet/aspire/issues/5444

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1667
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5773)


eng/Build.props
```diff
@@ -3,8 +3,8 @@
     <ProjectToBuild Include="$(RepoRoot)src\**\*.csproj" Exclude="$(RepoRoot)src\Aspire.ProjectTemplates\templates\**\*.csproj" />
     <ProjectToBuild Include="$(RepoRoot)eng\dcppack\**\*.csproj" />
     <ProjectToBuild Include="$(RepoRoot)eng\dashboardpack\**\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)playground\**\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)tests\**\*.csproj"  />
+    <!-- <ProjectToBuild Include="$(RepoRoot)playground\**\*.csproj" />
```
joperezr, line 6: Playground apps and tests were disabled to allow the repo to build. I will continue iterating on this until ready and will remove this change as well as mark the PR as ready for review.

playground/Directory.Build.targets
```diff
@@ -10,7 +10,7 @@
     <SkipEventHubsEmulation Condition="'$(SkipEventHubsEmulation)' == '' and ('$(RepoRoot)' == '' or '$(ContinuousIntegrationBuild)' == 'true' or '$(CODESPACES)' == 'true')">true</SkipEventHubsEmulation>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(IsAspireHost)' == 'true' and '$(SkipDashboardProjectReference)' != 'true'">
+  <ItemGroup Condition="$(MSBuildProjectName.EndsWith('AppHost')) and '$(SkipDashboardProjectReference)' != 'true'">
```
DamianEdwards, line 13: Can we make the `Aspire.Hosting.AppHost` package add something, e.g. property or project capability, to make conditions like a bit more concrete?

joperezr: We could, but note that this is only needed inside our repo. For Aspire consumers scenario, it is straight forward to determine whether or not their projects are AppHost projects as well as if they are using the SDK or not and which version. In the repo it's different, as we consume things directly as opposed to doing it via a package ref, which is what makes us have to write conditions like this.

playground/AWS/AWSCDK.AppHost/AWSCDK.AppHost.csproj
```diff
@@ -1,11 +1,10 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <IsAspireHost>true</IsAspireHost>
```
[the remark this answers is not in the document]
joperezr: At least from my initial testing of the changes on this PR things seem to be working as expected, but I could be missing things. Apart from dogfooding, what ways do we have to ensure that all things in visual studio light up as expected? Do we have them listed out somewhere so I can do a thorough validation?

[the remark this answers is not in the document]
davidfowl: I think is this needed for azd.

cc @ellismg @vhvb1989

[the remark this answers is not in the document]
joperezr: I think you are correct, I'm having issues when trying to deploy with azd, and in the logs I do see azd looking for the project that has isAspireHost set to true. VS experience seems good FYI.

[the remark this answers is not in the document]
joperezr: I have a plan B approach which I think will work. Trying that now and validating VS and publishing work as expected.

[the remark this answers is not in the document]
eerhardt: They are checking for the `IsAspireHost` property:

https://github.com/Azure/azure-dev/blob/3b48ae58706ea4c0feb1943d92caef9b9a24a12e/cli/azd/internal/appdetect/dotnet_apphost.go#L47

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
```diff
@@ -255,10 +255,10 @@ namespace Projects%3B
     <MakeDir Directories="$(AspireManifestPublishOutputPath)" />
   </Target>
 
-  <Target Name="__ErrorOnMininumWorkloadVersionMissing" BeforeTargets="PrepareForBuild"
-    Condition="'$(IsAspireHost)' == 'true' and ('$(AspireHostingSDKVersion)' == '' or $([MSBuild]::VersionLessThan('$(AspireHostingSDKVersion)', '8.2.0')))">
-    <Error Code="ASPIRE005"
-      Text="$(MSBuildProjectName) project requires a newer version of the .NET Aspire Workload to work correctly. Please run `dotnet workload update`." />
+  <Target Name="__ErrorOnMininumSdkVersionMissing" BeforeTargets="PrepareForBuild"
+  Condition="'$(IsAspireHost)' == 'true' and ('$(AspireHostingSDKVersion)' == '' or $([MSBuild]::VersionLessThan('$(AspireHostingSDKVersion)', '9.0.0')))">
```
joperezr, line 259: @DamianEdwards we should think about whether or not we want to do this. The question is whether or not we want to support the following scenario:

1. User has an existing Aspire 8.x project, and has the workload installed (which is required for them to build the project.)
2. After we release Aspire 9.0, they notice there is a new version of the AppHost package, so they bump their reference to 9.0 without adding a line in their project to use the new MSBuild SDK.

We have two options here:

1. We could just let this succeed, as they would already have the workload installed which practically has the same targets from the SDK so we could let them build. The main pro here is reducing friction for existing customers while upgrading. The main con I see is that we are allowing tearing, and also that this project will succeed or not depending on the machine it is running in.
2. We force them to also set the MSBuild SDK line in their project in order to be able to build. That is the way things are with these changes, but we could change this. The pro here is we don't allow tear reducing our testing matrix, and also that we have consistency across machines as the behavior of the build would be the same whether the 8.x workload is installed or not.

Thoughts?

DamianEdwards: I think we force them to add the MSBuild SDK. To get into this situation they have to change their project file anyway (upgrade to new package version) so it seems best to force them to be fully aligned with the new idiom rather than some odd combination we don't test.

joperezr: > To get into this situation they have to change their project file anyway (upgrade to new package version)

True but only if you assume people update their packages manually, which I'm not so sure that is the most common case. For people doing it via Nuget package manager in VS or some other tool, we would now be forcing them to hand edit the project (as I don't believe there is any UI for doing it)

DamianEdwards: The point was it requires a project modification, so I think it's the right time to force this change.

davidfowl: No more forcing users to break, we learned our lesson in 8.2

joperezr: That was different, in that case users were ending up in a broken state without them doing any changes on their side (so we broke them by the only fact of shipping new packages) This one, as @DamianEdwards pointed out, requires a user action to be on this state.

DamianEdwards: Right, this is the key difference. Last time, we broke folks after updating a globally installed asset (the workload). This time, we're saying, if you update your package ref to 9.0, you should update your project file to match what we expect it to look like, i.e. removing the property. We could make it less severe than an error too if that makes us feel better too.

davidfowl: I am anti this change. Worst case, there will be turds left in old project files. That seems like a problem that will fix itself as new templates become the default.

radical (approved):

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
DamianEdwards: For full transparency, so we're super clear, the move to 9.0 will require the following changesin the AppHost project:
- Updating the package reference for `Aspire.Hosting.AppHost` to version `9.0`
- Adding the SDK to the project file via an `<Sdk>` item

After doing that, this specific warning would be saying, hey you don't need that `<IsAspireHost>` property. If we think that's a stretch too far and we're OK with that unused property hanging around in some project files, then I'm fine to remove this (but let's do it in a separate PR so we can get this one in!!).

DamianEdwards: Now I've reread the original comment I realize I misunderstood this one. This is about the `<Sdk>` entry itself, for folks who already have the old workload installed. I **really** don't think we should rely on the 8.x workload targets for 9.0 projects, but rather we should force people to move to specifying the SDK specifically in their project files, so that their projects **no longer require the workload**. That is the key change we're trying to make here so we should enforce it IMO. The `<IsAspireHost>` property is a true leftover that has no functional impact so not instructing them to remove that has no impact.

DamianEdwards:
@joperezr is this ready to merge now!?!

src/Aspire.Hosting.AppHost/build/Aspire.Hosting.AppHost.targets
joperezr: Exactly, that describes the state. IsAspireHost will remain and won't hurt (and will help tools like azd as well as our internal conditions in the repo). The question was whether or not we should allow people with existing 8.x projects and workload installed to just bump their AppHost package version and have them be able to build, which we totally could but it does complicate our compat matrix as well as it is prone to break in places where workload isn't installed. Personally, I agree with @DamianEdwards and think we should go ahead with this change.

tests/Shared/WorkloadTesting/BuildEnvironment.cs
```diff
@@ -34,36 +33,42 @@ public class BuildEnvironment
...
-                    string workloadsProjString = Path.Combine("tests", "workloads.proj");
-                    throw new XunitException(
-                        $"Could not find a sdk with the workload installed at {sdkFromArtifactsPath} computed from {nameof(RepoRoot)}={RepoRoot}." +
-                        $" Build all the packages with '{buildCmd} -pack'." +
-                        $" Then install the sdk+workload with 'dotnet build {workloadsProjString}'." +
-                        " See https://github.com/dotnet/aspire/tree/main/tests/Aspire.Workload.Tests#readme for more details.");
-                }
-            }
-            else
+            // if (!useSystemDotNet)
```
joperezr, line 52: @radical should this be removed? IF not I can log an issue to make sure we are tracking this follow-up work.

tests/helix/send-to-helix-inner.proj
```diff
@@ -92,14 +92,17 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(NeedsWorkload)' == 'true'">
-    <HelixPreCommand Condition="'$(OS)' == 'Windows_NT'" Include="set PATH=%HELIX_CORRELATION_PAYLOAD%\dotnet-latest%3B%PATH%" />
-    <HelixPreCommand Condition="'$(OS)' != 'Windows_NT'" Include="export PATH=$HELIX_CORRELATION_PAYLOAD/dotnet-latest:$PATH" />
+    <!--<HelixPreCommand Condition="'$(OS)' == 'Windows_NT'" Include="set PATH=%HELIX_CORRELATION_PAYLOAD%\dotnet-latest%3B%PATH%" />-->
```
joperezr, line 95: @radical this too. Should we just remove this?

tests/Shared/WorkloadTesting/BuildEnvironment.cs
radical: I'm already working on the follow up in the templates9 PR, which includes changes from this branch.

tests/helix/send-to-helix-inner.proj
radical: I want to decide on what exactly needs to be removed, or retained in the templates9 PR, so I avoided removing stuff here.

tests/Shared/WorkloadTesting/BuildEnvironment.cs
joperezr: Ok great, then I'll leave as-is.

tests/helix/send-to-helix-inner.proj
joperezr: Sounds good, I'll leave it as-is then.

src/Aspire.ProjectTemplates/templates/aspire-apphost/Aspire.AppHost1.csproj
```diff
@@ -1,5 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <Sdk Name="Aspire.AppHost.Sdk" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
```
eerhardt, line 3: Why is this not just

```suggestion
<Project Sdk="Aspire.AppHost.Sdk">
```

?

joperezr: We could with some work, but I opted not for several reasons:

- Aspire.AppHost.Sdk doesn't import Microsoft.NEt.Sdk sdk, we could do it, but it doesn't do it now.
- The way that version is specified on top level sdk's is not using metadata which may not work in case you want to have the same property to control the reference to the SDK and the version of the AppHost package ref.
- To do so now would also make us sort of commit to the fact that AppHost projects will always be of this type of SDK, and it would be really hard if in the future there is a scenario where that is not the case.

Mainly the issue is that while it does look slightly cleaner, it would make us commit to a path that we may want to change in the future and we wouldn't be able to without breaking people again.

joperezr:
> @joperezr is this ready to merge now!?!

As soon as internal test build passes and CI is green I'm hitting merge.

src/Aspire.ProjectTemplates/templates/aspire-apphost/Aspire.AppHost1.csproj
eerhardt: > To do so now would also make us sort of commit to the fact that AppHost projects will always be of this type of SDK, and it would be really hard if in the future there is a scenario where that is not the case.

> Mainly the issue is that while it does look slightly cleaner, it would make us commit to a path that we may want to change in the future and we wouldn't be able to without breaking people again.

I don't see how the current approach is any different w.r.t. to that aspect. We are embedding the SDK name and version on line 3 instead of line 1. How could we change it without breaking people?

eerhardt: cc @baronfel - any input here? Is using the top `<Project Sdk=""` syntax not recommended for a reason like:

> The way that version is specified on top level sdk's is not using metadata

baronfel: For end users if you wanted to manage NuGet-delivered SDK versions uniformly your only real option today is global.json. I'd love to have something more user-friendly here - global.json MSBuild SDK versions aren't exactly integrated with the rest of our tooling, but all of that is owned by the NuGet team entirely.

Issue #1209, opened 2023-12-05
Add Aspire related analyzers for improved developer experience


timheuer:
There are likely patterns that are emerging here, especially in AppHost scenario, that could benefit from a set of code analyzers to warn/enforce/fix some of them.  For example in #1202 an analyzer could have produced a squiggle + build warning alerting me to a conflict.

![image](https://github.com/dotnet/aspire/assets/4821/a8b6ea9a-e581-4c09-9666-c2ee95036d6b)

mitchdenny:
Setting the first analyzer size to medium because we'll need to setup the infrastructure in the repo.

danmoseley:
I suggest moving this to the GA or later milestones as it can be out of band, and we are almost complete. Anyone see differently?

danmoseley:
moving to backlog

DamianEdwards:
FYI I have a branch adding the required infrastructure and the first analyzer (invalid resource/endpoint names) at https://github.com/dotnet/aspire/tree/damianedwards/model-name-analyzer. Hope to send a PR soon.

PR #5775, opened 2024-09-19
Add initial app host analyzers to the Aspire.Hosting.AppHost package

DamianEdwards:
## Description

This adds initial set of analyzers to `Aspire.Hosting.AppHost` and infrastructure to build more out in the future. Initial analyzers included are:

- ASPIRE006: Application model items must have valid names

Diagnostics are reported as errors as they'll result in runtime exceptions.

Contributes to #1209

The analyzer is based on hosting method parameters being decorated with a new `ResourceNameAttribute` or `EndpointNameAttribute`, which both implement the new interface `IModelNameParameter`. All relevant hosting methods (including in hosting integration packages) have been updated with these attributes as appropriate to enable the analyzer.

If we think requiring the attribute to enable the analyzers for a method is too burdensome, we could update the analyzer to also look for method operations invoked on `IDistributedApplicationBuilder` for methods named like `AddXXX` that pass a parameter `string name`. We could keep the attributes to enable other methods to participate where that pattern doesn't suffice, e.g. `IResourceBuilder<PostgresResource>.AddDatabase(string name)`, as it's likely not correct to always assume methods like that result in resources being added to the application model with the specified name.

The analyzer infrastructure is copied from [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore/tree/main/src/Framework/AspNetCoreAnalyzers/src/Analyzers).

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1658

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5775)


mitchdenny:
The unique name detection is going to need to handle branches:

```csharp
    [Fact]
    public async Task ResourceNameUniqueCanHandleTernaryBranches()
    {
        var test = AnalyzerTest.Create<AppHostAnalyzer>("""
            using Aspire.Hosting;

            var parameter = builder.ExecutionContext.IsPublishMode
                ? builder.AddParameter("x")
                : builder.AddParameter("x");
            """, []);

        await test.RunAsync();
    }

    [Fact]
    public async Task ResourceNameUniqueCanHandleIfBranches()
    {
        var test = AnalyzerTest.Create<AppHostAnalyzer>("""
            using Aspire.Hosting;

            IResourceBuilder<ParameterResource>? parameter;

            {
                parameter = builder.AddParameter("x");
            }
            else
            {
                parameter = builder.AddParameter("x");
            }
            """, []);

        await test.RunAsync();
    }

    [Fact]
    public async Task ResourceNameUniqueCanHandleSwitchBranches()
    {
        var test = AnalyzerTest.Create<AppHostAnalyzer>("""
            using Aspire.Hosting;

            var parameter = builder.ExecutionContext.Operation switch {
                DistributedApplicationOperation.Run => builder.AddParameter("x"),
                _ => builder.AddParameter("x")
            };
            """, []);

        await test.RunAsync();
    }
```

JamesNK:
> ASPIRE0001: Resource names must be unique

I wrote an analyzer like this for ASP.NET Core routing and reporting duplicate routes. It took forever to get right.

With this kind of analyzer you have to be **_really_** conservative about finding duplicates. You can only report a duplicate if they are both in the same context: different methods, if branches, switch statements, ternary expressions, null coalesing, etc, etc.

JamesNK:
My analyzer: https://github.com/dotnet/aspnetcore/blob/d05f3586adc0439595436ff6b8677548302fd64d/src/Framework/AspNetCoreAnalyzers/src/Analyzers/RouteHandlers/DetectAmbiguousRoutes.cs

My analyzer is very complex because it's possible to add extension methods to MapGet and friends, and as soon as you do that, you don't know if it is truely a duplicate or not. But you should be able to copy logic about only analyzing in the same context.

mitchdenny:
Funny you say that @JamesNK as I was looking at this I was remembering that analyzer thinking what a good job you did with it around handling these corner cases (although some of these will be pretty common patterns in Aspire).

JamesNK:
I didn't do that good a job. There were a bunch of backports during .NET 8 (9?) to fix issues.

mitchdenny:
... eventually. :)

DamianEdwards:
> With this kind of analyzer you have to be really conservative about finding duplicates. You can only report a duplicate if they are both in the same context: different methods, if branches, switch statements, ternary expressions, null coalesing, etc, etc.

Hmm OK. I think I'll just remove it for now and we can revisit it later. This is mostly about getting the infra in and the first analyzer to verify the resource name is valid.

eng/NullablePolyfill.targets
```diff
@@ -0,0 +1,26 @@
...
+        C# 8.0's nullable annotations only work for .NET Core 3.0+ by default. However it can be made to work for other target frameworks
+        by compiling the annotations (as internal types).
+        The code block that follows sets it up so projects in this repo that target ns2.0 or netfx can compile when Nullable is configured.
+        Based on https://github.com/dotnet/runtime/blob/93b6c449d4f31ddd7d573d1d3769e681d5ebceb9/src/libraries/Directory.Build.targets#L215-L222
+
+        SuppressNullableAttributesImport to disable. Used when attributes are already present from another project because of InternalsVisibleTo.
+    -->
+    <When Condition=" ('$(Nullable)' == 'annotations' OR '$(Nullable)' == 'enable') AND
+        '$(SuppressNullableAttributesImport)' != 'true' AND
+        (('$(TargetFrameworkIdentifier)' == '.NETStandard' AND $([MSBuild]::VersionLessThanOrEquals('$(TargetFrameworkVersion)', '2.1'))) OR '$(TargetFrameworkIdentifier)' == '.NETFramework')">
```
eerhardt, line 14: We can really simplify this condition given that we know we never target old/end of life TFMs.

```suggestion
    <When Condition="'$(Nullable)' != '' and
                        '$(Nullable)' != 'disable' and
                        '$(SkipIncludeNullableAttributes)' != 'true' and
                        '$(MSBuildProjectExtension)' == '.csproj' and
                        '$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
```

This is what currently is in `main` of dotnet/runtime:
https://github.com/dotnet/runtime/blob/4d37f40c49ed7f03d557e96323835bb87638b6ad/src/libraries/Directory.Build.targets#L171-L175

DamianEdwards: Yeah I lifted this straight from dotnet/aspnetcore: https://github.com/dotnet/aspnetcore/blob/542f2b9877da67e90cca6ce1b65aa9557fbb2aa8/eng/targets/CSharp.Common.targets#L112-L132

Happy to change to simpler version.

Directory.Packages.props
```diff
@@ -141,6 +141,7 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.9.0" />
     <!-- build dependencies -->
     <PackageVersion Include="MicroBuild.Plugins.SwixBuild.Dotnet" Version="1.1.87-gba258badda" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.11.0" />
```
eerhardt, line 144: We are going to have to be careful on versioning now. Previously we only referenced Roslyn in our internal build for the Configuration JSON schema generator. So we always updated the version, because it was only used for our build. That's why the existing CodeAnalysis dependences were listed under `<!-- build dependencies -->`.

But now that we are building analyzers that we give out to customers, we will need to be more thoughtful when this version gets updated.

I suggest at least splitting this new PackageVersion into a separate section that isn't under `<!-- build dependencies -->`. I also think it might make sense for us to not use CPM for this dependency and instead have an explicit MSBuild property for the version, and use `PackageOverride` in the analyzer. (Although I've never attempted to build an analyzer in a repo that uses CPM and dependabot before, so there might be better approaches.)

DamianEdwards: I'll try detaching the analyzer project from CPM and setting it specifically for the reasons you point out.

playground/TestShop/TestShop.AppHost/TestShop.AppHost.csproj
```diff
@@ -19,6 +19,12 @@
     <AspireProjectOrPackageReference Include="Aspire.Hosting.Redis" />
     <AspireProjectOrPackageReference Include="Aspire.Hosting.PostgreSQL" />
 
+    <ProjectReference Include="..\..\..\src\Aspire.Hosting.Analyzers\Aspire.Hosting.Analyzers.csproj" IsAspireProjectResource="false"
```
eerhardt, line 22: Should we put this in the root playground targets, so they all get it?

https://github.com/dotnet/aspire/blob/3bcf24fe7d24fba323753f548f4f606714bbb20c/playground/Directory.Build.targets#L13-L15

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            return new ModelNameOperation(operation, modelType, modelNameToken);
+        }
+    }
+
+    private enum ModelType
+    {
+        Unknown,
+        Resource,
+        Endpoint
```
eerhardt, line 145: Where is `ModelType.Endpoint` used? Can we remove this and the attribute until it is used? We can easily add it back when we need it.

DamianEdwards: Yeah it's not actually used currently, I can remove.

src/Aspire.Hosting.Analyzers/Infrastructure/BoundedCacheWithFactory.cs
```diff
@@ -0,0 +1,79 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Analyzers.Infrastructure;
+
+// This type is copied from https://github.com/dotnet/roslyn-analyzers/blob/9b58ec3ad33353d1a523cda8c4be38eaefc80ad8/src/Utilities/Compiler/BoundedCacheWithFactory.cs
```
eerhardt, line 6: This is unfortunate that it needs to be copied to everywhere that wants to be an analyzer.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
```diff
@@ -0,0 +1,28 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace Aspire.Hosting.Analyzers;
+
+public partial class AppHostAnalyzer
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("MicrosoftCodeAnalysisReleaseTracking", "RS2008:Enable analyzer release tracking")]
```
eerhardt, line 11: What does this mean and why can we just suppress it?

DamianEdwards: Copied from dotnet/aspnetcore, no idea :smile:

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
```diff
@@ -0,0 +1,28 @@
...
+    internal static class Diagnostics
+    {
+        internal static readonly DiagnosticDescriptor s_modelNameMustBeValid = new(
+            id: "ASPIRE0000",
```
eerhardt, line 15: We already have

* ASPIRE001
* ASPIRE002
* ...
* ASPIRE005

We should probably rectify these warning IDs. It might make sense to put them in a shared place so we can see them and know what each of them are.

Example: https://github.com/dotnet/runtime/blob/main/docs/project/list-of-diagnostics.md

src/Aspire.Hosting/ApplicationModel/ModelNameAttribute.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
eerhardt, line 1: (nit) file name and type name don't match.

src/Aspire.Hosting/ApplicationModel/EndpointNameAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+/// <code lang="csharp">
+/// public static IResourceBuilder&lt;TResource&gt; WithCustomEndpoint&lt;TResource&gt;(this IResourceBuilder&lt;TResource&gt; builder, [EndpointName] string name)
+/// </code>
+/// </para>
+/// </summary>
+/// <remarks>
+/// This API supports analyzers in Aspire.Hosting.Analyzers.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+public class EndpointNameAttribute : Attribute, IModelNameParameter
```
eerhardt, line 19: Can we remove this attribute until it is needed? I'm wondering if we actually even need this attribute, or if the analyzer would just hard-code the ~5 methods that take an EndpointName.

DamianEdwards: hard code, ewww. Could go back to what I had originally which is just a single attribute `ModelNameAttribute` that aligns with the `ModelName` type as now we're only checking for valid model name anyway.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
DamianEdwards: Only 3 digits? Man we better use them sparingly I guess!

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.cs
DamianEdwards: OK it *is* used it turns out, in that the model type is passed as string to `ModelName.TryValidateName`. That's why this is there.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
DamianEdwards: Added .\docs\list-of-diagnostics.md

src/Aspire.Hosting/ApplicationModel/EndpointNameAttribute.cs
DamianEdwards: This is needed as the analyzer needs to pass a "target" to `ModelName.TryValidate` which becomes part of the message. Originally I had it like `[ModelName("Resource")]` or `ModelName("Endpoint")` but that looked super repetitive and jank so I made it strongly typed.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
DamianEdwards: Should I change all of them to use the more typical four-digit format?

src/Aspire.Hosting.Analyzers/Infrastructure/BoundedCacheWithFactory.cs
DamianEdwards: Yeah. I was talking about another case where we wanted to share sources earlier today and day-dreamed about built-in support for producing source packages directly from the CLI, e.g. `dotnet pack --id MySourcePackage --version 0.0.1 --description "My Great source package" --sources ./**/*.cs,./**/*.resx`, so we could avoid having to create special projects everywhere we want to share sources.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\Aspire.Hosting\ApplicationModel\ModelName.cs" Link="ModelName.cs" />
+    <Compile Include="..\Shared\StringComparers.cs" Link="StringComparers.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.11.0" PrivateAssets="All" />
```
eerhardt, line 32: Is 4.11 too new? It was only shipped a month ago.

If someone is using net8 and an older VS, this might break them.

DamianEdwards: How do we determine what the correct version to use is?

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
eerhardt: > Only 3 digits?

Can always add `ASPIRE1001` in the future.

> Should I change all of them to use the more typical four-digit format?

I believe that would be a breaking change. Probably not worth it.

src/Aspire.Hosting.Analyzers/Infrastructure/BoundedCacheWithFactory.cs
eerhardt: See also https://github.com/dotnet/aspire/tree/main/src/Tools/ConfigurationSchemaGenerator/RuntimeSource where we copied a TON of the Configuration.Binder source generator code. 😞

eng/NullablePolyfill.targets
```diff
@@ -0,0 +1,28 @@
...
+
+        SuppressNullableAttributesImport to disable. Used when attributes are already present from another project because of InternalsVisibleTo.
+    -->
+    <When Condition="'$(Nullable)' != '' and
+                     '$(Nullable)' != 'disable' and
+                     '$(SkipIncludeNullableAttributes)' != 'true' and
+                     '$(MSBuildProjectExtension)' == '.csproj' and
+                     '$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
+      <PropertyGroup>
+        <DefineConstants>$(DefineConstants),INTERNAL_NULLABLE_ATTRIBUTES</DefineConstants>
```
eerhardt, line 18: We never want to make these attributes public. I would just get rid of this DefineConstant and delete all the `public` modifiers in `NullableAttributes.cs`.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
DamianEdwards: ASP.NET Core seems to be using 3.3.1 because it says some analyzers need to be used in VS2019, but perhaps that's out of date? https://github.com/dotnet/aspnetcore/blob/e25539b691b0425f94f66818ea0a0b4820f44804/eng/Versions.props#L247-L253

DamianEdwards: @jaredpar @dsplaisted what's the right way to determine which version of `Microsoft.CodeAnalysis.CSharp` our analyzer project should reference?

DamianEdwards:
@JamesNK @joperezr @captainsafia I believe you all have experiences authoring/shipping analyzers before? Would be good to get someone with analyzer experience to sign-off on this.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
eerhardt: I think you take the version that aligns with the minimum VS or .NET SDK dependency.

https://learn.microsoft.com/dotnet/aspire/fundamentals/setup-tooling says VS 17.10 is the minimum VS. That aligns with Roslyn 4.10. And .NET 8 SDK, I assume 8.0.100, which aligns with Roslyn 4.8.

So a good starting point would be `4.8.0`, I'd think. I don't imagine us needing to go back further since we only support .NET 8+.

jaredpar: Do your analyzers ship in a package or as part of the .NET SDK? If it's a separate package then @eerhardt is correct. If it's part of the .NET SDK then you can start taking a dependency on latest.

DamianEdwards: A package, thanks.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
DamianEdwards: Should we enable this? @eerhardt @joperezr

src/Aspire.Hosting.Analyzers/Infrastructure/WellKnownTypes.cs
```diff
@@ -0,0 +1,157 @@
...
+        if (types.Length == 1)
+        {
+            return types[0];
+        }
+
+        // Multiple types match the name. This is most likely caused by someone reusing the namespace + type name in their apps or libraries.
+        // Workaround this situation by prioritizing types in System and Microsoft assemblies.
+        foreach (var type in types)
+        {
+            if (type.ContainingAssembly.Identity.Name.StartsWith("System.", StringComparison.Ordinal)
```
captainsafia, line 102: We had to apply this workaround in the ASP.NET Core repo because the `Microsoft.AspNetCore.Builder.EndpointRouteBuilderExtensions` type name was used both by the shared framework and a third-party dependency in some cases. This would cause the `GetTypeByMetadataName` to throw. See https://github.com/dotnet/aspnetcore/issues/50836#issuecomment-1813301014 for more info.

You might be able to get away with removing this workaround given the set of types in your WellKnownTypes cache.

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.Diagnostics.cs
captainsafia: We've had these disabled in the ASP.NET Core repo since forever and I think we disabled them when we started shipping analyzers in the shared framework out of habit as well.

If we could enable them from the get-go here, I'd be in favor of it since it'll force better practices around documenting analyzers.

That being said, I can't comment on how good the infrastructure is with working with these analyzer releasing tracking documents. For example, I'm not sure if there are codefixes for these in the same way there are for PublicApi.txt files.

DamianEdwards: Seems like it does and seems to work pretty well:

https://github.com/user-attachments/assets/e13fc10e-7281-4af5-a294-d63376224aee

DamianEdwards: I went ahead and enabled this.

DamianEdwards:
I think all feedback has been addressed. Just looking for an approval now 😃 
@eerhardt @JamesNK @mitchdenny @captainsafia

src/Aspire.Hosting.Analyzers/AppHostAnalyzer.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            modelNameOperations.TryAdd(ModelNameOperation.Create(invocation, parameterData.Value.ModelType, token), value: default);
+        }
+    }
+
+    private static bool IsModelNameInvocation(WellKnownTypes wellKnownTypes, IMethodSymbol targetMethod, [NotNullWhen(true)] out (IParameterSymbol ModelNameParameter, ModelType ModelType)? parameterData)
+    {
+        // Look for first string parameter annotated with attribute that implements IModelNameParameter
+        ModelType modelType = default;
+        var candidateParameter = targetMethod.Parameters.FirstOrDefault(ps =>
```
JamesNK, line 78: Is only the first parameter with a model name attribute validated? If all methods only have one at the moment, it's problem just a matter of time until we have a situation with multiple.

DamianEdwards: Right now, yes. I could update it to support multiple.

DamianEdwards: I updated it to support multiple parameters in a single method and multiple attributes on a single parameter.

src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
```diff
@@ -32,4 +32,16 @@
     <PackageDownload Include="Microsoft.DeveloperControlPlane.$(BuildOs)-$(BuildArch)" Version="[$(MicrosoftDeveloperControlPlanedarwinamd64PackageVersion)]" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!-- Reference the analyzer to ensure it's built but don't reference its output -->
+    <ProjectReference Include="..\Aspire.Hosting.Analyzers\Aspire.Hosting.Analyzers.csproj" OutputItemType="AnalyzerAssemblyPath" ReferenceOutputAssembly="false" Private="true" />
```
eerhardt, line 37: Did you happen to `dotnet pack` this project and make sure the package looks right? (No new dependencies, the analyzer is in the right spot, etc.)

DamianEdwards: image

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
```diff
@@ -0,0 +1,42 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
eerhardt, line 5: I think this line can be removed since this csproj doesn't support packing - IsPackable=false.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <UsePublicApiAnalyzers>false</UsePublicApiAnalyzers>
+    <IsPackable>false</IsPackable>
+    <IsShipping>false</IsShipping>
```
eerhardt, line 10: This feels like a lie. This is shipping. (I'm not sure what this property actually does.)

DamianEdwards: I think it's used by build infra to organize outputs into the Shipping/NotShiping directories?

DamianEdwards: Again, copied from dotnet/aspnetcore https://github.com/dotnet/aspnetcore/blob/843f7b011dae5687c77f3570542325b01301f8db/src/Framework/AspNetCoreAnalyzers/src/Analyzers/Microsoft.AspNetCore.App.Analyzers.csproj#L4-L7

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
DamianEdwards: Copied from dotnet/aspnetcore https://github.com/dotnet/aspnetcore/blob/843f7b011dae5687c77f3570542325b01301f8db/src/Framework/AspNetCoreAnalyzers/src/Analyzers/Microsoft.AspNetCore.App.Analyzers.csproj#L4-L7

tests/Aspire.Hosting.Analyzers.Tests/Aspire.Hosting.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+
+    <RunTestsOnHelix>false</RunTestsOnHelix>
```
eerhardt, line 6: Is there a reason these can't be run on Helix? I think our goal is to get them all running there, if we can.

cc @radical

eerhardt (approved):
LGTM on the infra side. Just some random questions.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
eerhardt: This is different. You copied some properties, and then set other properties that nullify these copied properties.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
eerhardt: > copied from dotnet/aspnetcore

Your link points to `IsShippingPackage`, not `IsShipping`.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
DamianEdwards: Ah OK, yes `<IsPackable>false</IsPackable>` came from the Roslyn analyzer template IIRC. We're structuring this a bit different to how the template does so very likely we can simplify this a bit.

tests/Aspire.Hosting.Analyzers.Tests/Aspire.Hosting.Analyzers.Tests.csproj
DamianEdwards: Hmmm I'm not sure where that came from actually. I can try removing.

src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj
DamianEdwards: Doh. I can remove. Do we know what it does?

eerhardt: > Do we know what it does?

I don't. I see:

https://github.com/dotnet/aspnetcore/blob/843f7b011dae5687c77f3570542325b01301f8db/docs/ProjectProperties.md

not sure if those are dotnet/aspnetcore specific properties or what.

radical:
For analyzers in general, we should install these for the workload/template tests too, have them turned on, and use `warnAsError` so we can catch issues that users might run into.

radical:
Created https://github.com/dotnet/aspire/issues/5876 .

PR #5779, opened 2024-09-19
Make unnamed volume names more unique

### LINKED ISSUE #5413: Implicitly named volumes created by the WithDataVolume() method can collide across AppHost projects with the same name
The various `WithDataVolume` methods create a container volume that's implicitly named based on the App Host project name and resource name, e.g. if the App Host project is named "TestShop.AppHost" and the resource the method is called on is named "postgres", the implicit volume name will be "TestShop.AppHost-postgres-data".

This can lead to collisions of these implicit volume names when the App Host project is not named uniquely between different solutions. I'm seeing more customer solutions where the App Host project is simply named "AppHost". If there are resources named simply such as "postgres" across those App Host projects, their implicit data volume names will collide.

This issue can be worked around currently by the developer by passing a custom volume name to the `WithDataVolume` method, but then it's the responsibility of the developer to ensure volume names don't clash between any container resource with a volume (or indeed any other non-Aspire created volume).

We should consider updating the implicit volume naming logic to include a hash of the App Host project path, or some other deterministic, stable value derived from the App Host project. Note this would be a breaking behavior change although we could consider quirking it (i.e. allowing the user to specify what implicit volume naming strategy is used) or some other mitigation.

davidfowl:
## Description

- Follow a scheme similar to what we do for persistent containers. Use the first 10 characters of the SHA256 of the app host's physical path as the volume name prefix.
- This is a breaking change and the only workaround is the manually specify the volume name.

Fixes #5413

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5779)

### DIFF
--- a/src/Shared/VolumeNameGenerator.cs
+++ b/src/Shared/VolumeNameGenerator.cs
@@ -20,8 +20,10 @@ internal static class VolumeNameGenerator
             throw new ArgumentException($"The suffix '{suffix}' contains invalid characters. Only [a-zA-Z0-9_.-] are allowed.", nameof(suffix));
         }
 
-        // Create volume name like "myapplication-postgres-data"
-        var applicationName = builder.ApplicationBuilder.Environment.ApplicationName;
+        // Create volume name like "{sha256}-postgres-data"
+
+        // Compute a short hash of the content root path to differentiate between multiple AppHost projects with similar volume names
+        var applicationName = builder.ApplicationBuilder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
         var resourceName = builder.Resource.Name;
         return $"{(HasOnlyValidChars(applicationName) ? applicationName : "volume")}-{resourceName}-{suffix}";
     }
--- a/src/Shared/VolumeNameGenerator.cs
+++ b/src/Shared/VolumeNameGenerator.cs
@@ -7,12 +7,6 @@ namespace Aspire.Hosting.Utils;
 
 internal static class VolumeNameGenerator
 {
-    /// <summary>
-    /// Creates a volume name with the form <c>$"{applicationName}-{resourceName}-{suffix}</c>, e.g. <c>"myapplication-postgres-data"</c>.
-    /// </summary>
-    /// <remarks>
-    /// If the application name contains chars that are invalid for a volume name, the prefix <c>"volume-"</c> will be used instead.
-    /// </remarks>
     public static string CreateVolumeName<T>(IResourceBuilder<T> builder, string suffix) where T : IResource
     {
         if (!HasOnlyValidChars(suffix))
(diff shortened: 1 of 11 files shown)


src/Shared/VolumeNameGenerator.cs
```diff
@@ -20,8 +20,10 @@ public static string CreateVolumeName<T>(IResourceBuilder<T> builder, string suf
             throw new ArgumentException($"The suffix '{suffix}' contains invalid characters. Only [a-zA-Z0-9_.-] are allowed.", nameof(suffix));
         }
 
-        // Create volume name like "myapplication-postgres-data"
-        var applicationName = builder.ApplicationBuilder.Environment.ApplicationName;
+        // Create volume name like "{sha256}-postgres-data"
+
+        // Compute a short hash of the content root path to differentiate between multiple AppHost projects with similar volume names
+        var applicationName = builder.ApplicationBuilder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
```
eerhardt, line 26: As a user, how am I supposed to know which volume maps to which app in Docker Desktop?

davidfowl: You don't know the SHA256 off the top of your head?!? I was waiting for feedback to say we should keep the assembly name as part of the prefix. I think instead of what it does today around dropping the name if the characters are invalid, we should instead remove those characters.

eerhardt: Is there a length limit? Maybe we put the 5-10 character SHA at the end of the volume name?

davidfowl: I think the limit is OS specific (these end up being paths). I also think we should lowercase the name.

eerhardt: > I also think we should lowercase the name.

I don't necessarily object to that. I just want to be able to understand what each volume is used for.

DamianEdwards: Yeah I think we should keep the application name in here and append the SHA to disambiguate.

DamianEdwards: Also you'll need to update the doc-comment to reflect whatever you end up making the format.

davidfowl: Doc comments on internals APIs 👀

DamianEdwards: Then delete it 😄

JamesNK, line 26: Who sets `AppHost:Sha256`? Do we need to be resilent against `AppHost:Sha256` not having a value? Or someone setting a value shorter than 10?

JamesNK: We do: https://github.com/dotnet/aspire/blob/4d80745955ccc0c9b31b1a8fe94e8d5b5b8a1d77/src/Aspire.Hosting/DistributedApplicationBuilder.cs#L154-L169

Nevermind.

davidfowl: Done

DamianEdwards (approved):
Looks good!

mitchdenny:
@davidfowl I just noticed this whilst doing some experimentation with container app customization. In the playground for containerapps the redis container uses a data volume. I noticed that the data volume name changes based on the path you are publishing from. This probably isn't desirable at deployment time because if you change your source layout on the build agent (for example) then you'll end up with a different path.

For deployment scenarios instead of using the AppHost:Sha256 value we probably need to use something based on the resource group name (like we do for resource names).

davidfowl:
That's a good catch. We need a sha based on the apphost name itself and it shouldn't include the path (ephemeral CI environments etc).

PR #5784, opened 2024-09-19
[dashpages] Dashpage layout improvements

drewnoakes:
Make dashpage responsive to container, not viewport

`FluentGrid` uses `@media` queries for breakpoints, which work on the window size (viewport area). However we have a splitter in this UI, and dragging it should cause the dashpage area to relayout as the container size changes.

This change removes `FluentGrid` and instead uses flexbox layout directly.

> [!NOTE]
> This PR is to the `feature/dashpages` branch, not `main`.

## Before

dashpage-layout-before

## After

dashpage-layout-after
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5784)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
@@ -13,7 +13,7 @@
 <FluentGrid Spacing="3" AdaptiveRendering="true" Justify="JustifyContent.FlexStart" Class="dashpage-chart-grid">
     @foreach (var chart in SelectedDashpage.Charts)
     {
-        if (Instruments.FirstOrDefault(i => i.Name == chart.InstrumentName) is { } instrument)
+        if (Instruments.FirstOrDefault(i => StringComparers.OtlpInstrumentName.Equals(i.Name, chart.InstrumentName)) is { } instrument)
         {
             @if (ViewportInformation.IsDesktop)
             {
--- a/src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
@@ -10,44 +10,27 @@
     <h3 title="@SelectedDashpage.Name">@SelectedDashpage.Name</h3>
 }
 
-<FluentGrid Spacing="3" AdaptiveRendering="true" Justify="JustifyContent.FlexStart" Class="dashpage-chart-grid">
+<section class="dashpage-chart-grid">
     @foreach (var chart in SelectedDashpage.Charts)
     {
         if (Instruments.FirstOrDefault(i => StringComparers.OtlpInstrumentName.Equals(i.Name, chart.InstrumentName)) is { } instrument)
         {
-            @if (ViewportInformation.IsDesktop)
-            {
-                <FluentGridItem sm="12" md="6" lg="4" Class="dashpage-chart-grid-item">
-                    <FluentCard AreaRestricted="false">
-                        <DashpageChartContainer
-                            ApplicationKey="@(SelectedApplication.Id!.GetApplicationKey())"
-                            MeterName="@(instrument.Parent.MeterName)"
-                            InstrumentName="@(instrument.Name)"
-                            Duration="@SelectedDuration.Id"
-                            ActiveView="@(SelectedViewKind ?? Metrics.MetricViewKind.Graph)"
-                            OnViewChangedAsync="@OnViewChangedAsync"
-                            Applications="@Applications" />
-                    </FluentCard>
-                </FluentGridItem>
-            }
-            else
-            {
-                <div class="dashpage-chart-grid-item">
-                    <FluentCard AreaRestricted="true">
-                        <DashpageChartContainer
-                            ApplicationKey="@(SelectedApplication.Id!.GetApplicationKey())"
-                            MeterName="@(instrument.Parent.MeterName)"
-                            InstrumentName="@(instrument.Name)"
-                            Duration="@SelectedDuration.Id"
-                            ActiveView="@(SelectedViewKind ?? Metrics.MetricViewKind.Graph)"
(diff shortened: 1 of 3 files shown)


src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
```diff
@@ -10,44 +10,27 @@
     <h3 title="@SelectedDashpage.Name">@SelectedDashpage.Name</h3>
 }
 
-<FluentGrid Spacing="3" AdaptiveRendering="true" Justify="JustifyContent.FlexStart" Class="dashpage-chart-grid">
+<section class="dashpage-chart-grid">
```
JamesNK, line 13: Why is a section element instead of div?

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
```diff
@@ -10,44 +10,27 @@
...
     @foreach (var chart in SelectedDashpage.Charts)
     {
-        if (Instruments.FirstOrDefault(i => i.Name == chart.InstrumentName) is { } instrument)
+        if (Instruments.FirstOrDefault(i => StringComparers.OtlpInstrumentName.Equals(i.Name, chart.InstrumentName)) is { } instrument)
```
JamesNK, line 16: nit: string.Equals + StringComparison.OtlpInstrumentName is more idomatic in this scenario

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor.css
```diff
@@ -0,0 +1,11 @@
+.dashpage-chart-grid {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 10px;
+}
+
+.dashpage-chart-grid .dashpage-chart-tile {
+    flex: 1 1 25%;
```
JamesNK, line 8: Could you add a small comment here that says what this does

adamint (approved):

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
drewnoakes: `section` gives a bit more semantics, suggesting it's a larger area of content. Similarly, `article` is used for each chart. This theoretically gives tools that inspect the DOM (such as screen readers) more info about the DOM. `div`s get used for non-semantic layout.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
drewnoakes: Seems subjective. I prefer this form as it's less code and feels more natural to me to use the right object for the operation, than passing an enum to a method to modify its behaviour. I feel like `StringComparison` only exists for historical reasons, and could be removed from the library if it weren't for backcompat requirements.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor.css
drewnoakes: Can you explain what's unclear? Do you want me to explain how `flex` works?

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
JamesNK: I wouldn't worry about being semantically correct. We're using div everywhere for block content and span for inline content.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
JamesNK: StringComparer is for non-string scenarios, e.g. the key comparer in a dictionary:
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#methods-that-perform-string-comparison-indirectly

StringComparison is for string operations:
https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage

JamesNK: This SO answer summerizies it well: https://stackoverflow.com/a/8918112

I feel like we've had this conversation before.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor.css
JamesNK: I don't know what `1 1 25%` means. I'm more experienced with css grid than flex box and I'm guessing most people who see this won't know what it does.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor
adamint: It is preferred to use the semantically correct element. I would not change to div just for the sake of it

JamesNK: We aren't using semantically correct elements in other places. It seems like a non-goal and makes it different from everywhere else in the app which is always to use div. Often being consistently correct is better than being technically correct.

But arguing over html elements is the least productive thing in the world so if you think this is best then go ahead.

src/Aspire.Dashboard/Components/Controls/Chart/DashpageCharts.razor.css
drewnoakes: Ok I see. This doc summarizes it pretty well: https://developer.mozilla.org/en-US/docs/Web/CSS/flex

I'm not sure what comment we'd add, beyond linking to docs. There'll always be someone who isn't familiar with a language/platform/framework feature we're using, and we can't explain all of them.

PR #5789, opened 2024-09-19
Add method to select the OpenAI component based on configuration

### LINKED ISSUE #5755: `AddOpenAIRestApiClient` to wrap both AzureOpenAI and OpenAI integrations
## Background and Motivation

Between the existing **Aspire.Azure.AI.OpenAI** and the new **Aspire.OpenAI** integration implemented in https://github.com/dotnet/aspire/pull/5621 we can now register an `AzureOpenAIClient` or `OpenAIClient` with `AddAzureOpenAIClient` or `AddOpenAIClient`. Also `AzureOpenAIClient` inherits from `OpenAIClient`.

We want to create a new method in the Aspire.Azure.AI.OpenAI component, named `AddOpenAIRestApiClient` that will be able to return the correct instance, either `AzureOpenAIClient` or `OpenAIClient` based on the connection string that is provided.

The logic to detect which concrete implementation to return will be based on the presence of an `Endpoint` in the connection string:
- If it's empty it will use the OpenAI service. 
- Otherwise:
  - If `IsAzure` is set in the connection string, if it's `true` return `AzureOpenAIClient`, otherwise `OpenAIClient`.
  - If `Endpoint` contains `azure.com` or `azure.cn`return the `AzureOpenAIClient` instance
  - Otherwise return an `OpenAIClient` instance.

## Proposed API

```diff
+ namespace Microsoft.Extensions.Hosting;

+ public static class AspireOpenAIRestAPIExtensions
+ {
+     /// <summary>
+     /// Registers a concrete implementation of <see cref="OpenAIClient"/> as a singleton in the services provided by the <paramref name="builder"/> based on the provided connection string.
+     /// </summary>
+     /// <param name="builder">The <see cref="IHostApplicationBuilder" /> to read config from and add services to.</param>
+     /// <param name="connectionName">A name used to retrieve the connection string from the ConnectionStrings configuration section.</param>
+     public static void AddOpenAIRestApiClient(this IHostApplicationBuilder builder, string connectionName)
+ }
```

## Usage Examples

```csharp
builder.AddServiceDefaults();

builder.AddOpenAIRestApiClient("openai");
```

These AppHost configurations would register an `AzureOpenAIClient` instance:

```json
{
  "ConnectionStrings": {
    "openai": "https://{account_name}.openai.azure.com/"
  }
}
```

```json
{
  "ConnectionStrings": {
    "openai": "Endpoint=https://{account_name}someotherazuredomain.com/;IsAzure=true"
  }
}
```

These AppHost configurations would register an `OpenAIClient` instance:

```json
{
  "ConnectionStrings": {
    "openai": "Key={openai_account_key}"
  }
}
```

```json
{
  "ConnectionStrings": {
    "openai": "Endpoint=http://localhost/myollama"
  }
}
```

Each Aspire service would then provide the configuration matching the endpoint they provided, using either `Aspire:Azure:AI:OpenAI` or `Aspire:OpenAI` keys.

## Alternative Designs

N/A

## Risks

This is not a breaking change, and it builds on top of the existing OpenAI integrations such that there is no code change in these either.

sebastienros:
Fixes https://github.com/dotnet/aspire/issues/5755
Contributes to https://github.com/dotnet/aspire/issues/4567
Completes https://github.com/dotnet/aspire/pull/5621
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5789)


sebastienros:
The method name is still an open discussion.

sebastienros:
> Choosing the best name for a method is crucial for clarity and maintainability. Both names you provided have their merits, but let's break them down:
AddOpenAIClientFromConfiguration: This name clearly indicates that the method adds an OpenAI client based on some configuration settings. It's straightforward and tells the user exactly what the method does.
AddAzureOrDefaultOpenAIClient: This name suggests that the method adds an OpenAI client, either from Azure or a default configuration. It implies a fallback mechanism, which might be useful if the primary configuration is not available.
Given these options, I would recommend AddOpenAIClientFromConfiguration. This name is more explicit about the source of the configuration and avoids potential confusion about the fallback mechanism. It keeps the method's purpose clear and concise.

src/Components/Aspire.Azure.AI.OpenAI/AspireConfigurableOpenAIExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+        };
+
+        if (connectionBuilder.TryGetValue(ConnectionStringEndpoint, out var endpoint) && endpoint != null && Uri.TryCreate(endpoint.ToString(), UriKind.Absolute, out var endpointUri))
+        {
+            serviceUri = endpointUri;
+        }
+
+        if (serviceUri == null)
+        {
+            throw new InvalidOperationException($"An OpenAIClient could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:{connectionName}' or specify a '{nameof(AzureOpenAISettings.Endpoint)}' or '{nameof(AzureOpenAISettings.Key)}' in the '{connectionName}' configuration section.");
```
eerhardt, line 96: If there is no URI and only a Key, doesn't that mean `OpenAIClient` should be used?

src/Components/Aspire.Azure.AI.OpenAI/README.md
```diff
@@ -36,10 +36,21 @@ public CognitiveController(AzureOpenAIClient client)
...
 
+## Azure-agnostic client resolution
+
+You can retrieve the `AzureOpenAIClient` object using the base `OpenAIClient` service type. This allows for code that is not dependent on Azure OpenAI-specific features to not depend directly on Azure types.
+
+Additionally this package provides the `AddOpenAIClientFromConfiguration` extension method to register an `OpenAIClient` instance based on the connection string that is provided. This allows your application
+to register the best implementation for the OpenAI Rest API it connects. The following rules are followed:
+
+- If the `Endpoint` attribute is empty, the OpenAI service is used and an `OpenAIClient` instance is registered.
+- If the attribute `IsAzure` is provided, `AzureOpenAIClient` is registered if the value is `true`, `OpenAIClient` otherwise.
```
eerhardt, line 49: We might want to document what the connection string looks like in the README.

sebastienros:
I validated it's working in eShop.

src/Components/Aspire.Azure.AI.OpenAI/AspireConfigurableOpenAIExtensions.cs
sebastienros: Correct, I fixed it after testing in eShop.

src/Components/Aspire.Azure.AI.OpenAI/AspireConfigurableOpenAIExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using OpenAI;
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// Provides extension methods for registering <see cref="OpenAIClient"/> as a singleton in the services provided by the <see cref="IHostApplicationBuilder"/>.
+/// </summary>
+public static class AspireConfigurableOpenAIExtensions
```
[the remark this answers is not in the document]
sebastienros: I prefer to keep it separate for readability, and also to map to the separate test class.

davidfowl (approved):

eerhardt (approved):

PR #5802, opened 2024-09-20
Inject commands on all DCP based resources

davidfowl:
## Description

Follow up to https://github.com/dotnet/aspire/pull/5538. Wire up commands later in the cycle so that we get all resources that DCP considers valid (this fixes emulators and other dynamically added resources that don't go through extension methods).

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5802)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
@@ -12,9 +12,9 @@ internal static class CommandsConfigurationExtensions
     internal const string StopType = "stop";
     internal const string RestartType = "restart";
 
-    internal static IResourceBuilder<T> WithLifeCycleCommands<T>(this IResourceBuilder<T> builder) where T : IResource
+    internal static void AddLifeCycleCommands(this IResource resource)
     {
-        builder.WithCommand(
+        resource.Annotations.Add(new ResourceCommandAnnotation(
             type: StartType,
             displayName: "Start",
             executeCommand: async context =>
@@ -40,9 +40,9 @@ internal static class CommandsConfigurationExtensions
                 }
             },
             iconName: "Play",
-            isHighlighted: true);
+            isHighlighted: true));
 
-        builder.WithCommand(
+        resource.Annotations.Add(new ResourceCommandAnnotation(
             type: StopType,
             displayName: "Stop",
             executeCommand: async context =>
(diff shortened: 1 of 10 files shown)


mitchdenny (approved):

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1165,8 +1165,11 @@ private void PrepareProjectExecutables()
         }
     }
 
-    private static void EnsureReplicaInstancesAnnotation(IResource resource)
+    private static void EnsureRequiredAnnotations(IResource resource)
     {
+        // Add the default lifecycle commands (start/stop/restart)
+        resource.AddLifeCycleCommands();
```
JamesNK, line 1171: One problem with adding them here is there isn't a way for someone to modify annotations to remove the start/stop/restart commands.

I don't know if anyone would ever want to do that 🤷

davidfowl: It's possible to support by checking and no overriding, but I want to fix the gaps we currently have.

src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
```diff
@@ -12,9 +12,9 @@ internal static class CommandsConfigurationExtensions
     internal const string StopType = "stop";
     internal const string RestartType = "restart";
 
-    internal static IResourceBuilder<T> WithLifeCycleCommands<T>(this IResourceBuilder<T> builder) where T : IResource
+    internal static void AddLifeCycleCommands(this IResource resource)
```
JamesNK, line 15: nit: This type is called CommandsConfigurationExtensions. Since this is no longer an extension method type, consider a different name.

JamesNK (approved):

src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
davidfowl: It’s still an extension method

JamesNK: Oh yeah. But not on builder.

Anyway, it doesn't matter.

PR #5806, opened 2024-09-20
Uses values from constants in documentation

sebastienros:
There were some discrepancies between the code documentation and the actual values. This suggested pattern will prevent that, or at least make it obvious to update or catch in PRs by having the summary close to the actually value.

This is how it looks after:

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5806)


eerhardt (approved):
Assuming this is all of them, this looks great! Thanks

PR #5807, opened 2024-09-20
[release/8.2] Fix RID regression by adding a task that calculates the best matching RID for platform

joperezr:
Manual backport of #5695 to release/8.2

cc: @eerhardt 

## Customer Impact

In Aspire, we have always just cross-compiled our Dashbarod and DCP packages for a limited set of RIDs (win-x64, win-x86, win-arm64, linux-x64, linux-arm64, osx-x64, and osx-arm64). When running on platforms that had special RIDs that we weren't cross-compiling for (for example, rhel.8) then we were relying on the SDK's logic to automatically pick the best closest Dashboard and DCP pack for that RID.

However, in Aspire 8.2.0 we removed Dashboard and DCP from the workload and instead just added some sdk targets that were adding these references on the fly, but we were not doing this special matching logic the the SDK used to do and instead we were simply trying to use the current RID, which caused a regression in all of those platforms that have special RIDs which we don't cross-compile for.

In this PR, we are copying the logic that the SDK does for doing that matching, and we are calling it inside of our targets in order to ensure that we are going to get a valid package reference, which will fix the regression in those systems. This has been manually validated in RedHat 8, which was the original report for this regression. 

## Testing

Automated testing to catch regressions with new versions as well as manual exploratory testing

## Risk

Low.

## Regression?

yes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5807)


src/Aspire.Hosting.Sdk/Aspire.RuntimeIdentifier.Tool/Program.cs
```diff
@@ -0,0 +1,86 @@
...
+        rootCommand.Options.Add(runtimeGraphPathOption);
+        rootCommand.Options.Add(netcoreSdkRuntimeIdentifierOption);
+        rootCommand.Options.Add(supportedRidsOption);
+        rootCommand.SetAction((ParseResult parseResult) =>
+        {
+            string rgp = parseResult.GetValue(runtimeGraphPathOption) ?? throw new InvalidOperationException("The --runtimeGraphPath argument is required.");
+
+            if (!File.Exists(rgp))
+            {
+                throw new FileNotFoundException("File {0} does not exist. Please ensure the runtime graph path exists.", rgp);
```
sebastienros, line 51: I don't think the pattern will work here

joperezr: You are right, in fact I don't think throwing an exception is best in this case as this will show up in build logs so I'll change this for a Console WriteLine and use a non-zero return code.

joperezr:
Merging to get a build ready for validation for 8.2.1. Will address any other feedback in a follow-up.

PR #5809, opened 2024-09-20
Allow for customization of the Azure ProvisioningContext + Better Azure resource name scheme

### LINKED ISSUE #5341: Invalid Cloud deployment of Azure Service Bus topics - name truncation

### Describe the bug

**Issue**
Topics names are truncated in bicep scripts based on fluent distributed application definition.
**Code**
```csharp

var serviceBus = builder.AddAzureServiceBus("serviceBus");

serviceBus.AddTopic("device-connection-state-events", ["device-management"]);
serviceBus.AddTopic("device-lifecycle-events", ["device-management"]);
serviceBus.AddTopic("device-twin-change-events", ["device-management"]);

...
```
Bicep script from temp
```json
resource serviceBusTopic_*** 'Microsoft.ServiceBus/namespaces/topics@2021-11-01' = {
  parent: serviceBusNamespace_***
  name: 'device-connection-state-'
  location: location
  properties: {
  }
}
```
and
```json
resource serviceBusTopic_*** 'Microsoft.ServiceBus/namespaces/topics@2021-11-01' = {
  parent: serviceBusNamespace_***
  name: 'device-twin-change-event'
  location: location
  properties: {
  }
}
```

### Expected Behavior

- Topic is defined correctly up to 260 chars: https://learn.microsoft.com/en-us/azure/templates/microsoft.servicebus/namespaces/topics?pivots=deployment-language-bicep#property-values

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

.net 8

### Anything else?

_No response_

### LINKED ISSUE #5756: Better Azure resource name scheme
## Problem

In 8.x, we used an early/alpha version of Azure.Provisioning. This version used a naming scheme for Azure resources that tried to be the least common denominator of all resources.

```C#
        protected string GetGloballyUniqueName(string resourceName)
            => $"toLower(take('{resourceName}${{uniqueString(resourceGroup().id)}}', 24))";
```

This causes problems like [Invalid Cloud deployment of Azure Service Bus topics - name truncation (dotnet/aspire#5341)](https://github.com/dotnet/aspire/issues/5341).

The new version of Azure.Provisioning has a [better scheme](https://github.com/Azure/azure-sdk-for-net/blob/60109aee76c8e70143efebc1e304a550273aceae/sdk/provisioning/Azure.Provisioning/src/Primitives/ResourceNamePropertyResolver.cs#L201-L214) that has information about what names the resource type allows (max length, valid characters, etc.).

```C#
    public override BicepValue<string>? ResolveName(
        ProvisioningContext context,
        Resource resource,
        ResourceNameRequirements requirements)
    {
        string prefix = SanitizeText(resource.ResourceName, requirements.ValidCharacters);
        string separator =
            requirements.ValidCharacters.HasFlag(ResourceNameCharacters.Hyphen) ? "-" :
            requirements.ValidCharacters.HasFlag(ResourceNameCharacters.Underscore) ? "_" :
            requirements.ValidCharacters.HasFlag(ResourceNameCharacters.Period) ? "." :
            "";
        BicepValue<string> suffix = GetUniqueSuffix(context, resource);
        return BicepFunction.Take(BicepFunction.Interpolate($"{prefix}{separator}{suffix}"), requirements.MaxLength);
    }
```

However, we are not able to take advantage of this new scheme. It is a breaking change to take this new naming scheme. If someone used 8.x to deploy to Azure, then updated to 9.0 and deployed again they would get duplicate resources since the naming scheme changed. For example, there is a separator now and `toLower` is no longer called in bicep but instead the `resource.ResourceName` is sanitized first before putting it in the bicep. If the resource doesn't allow upper case characters, the ResourceName is lowered. For resources that support upper case characters, this means a different resource name.

The new naming scheme is better than the old, but maintaining compatibliity is not allowing us to use it.

## Proposed Solution

Allow for a user to customize the Azure CDK `ProvisioningContext` object that we use to build bicep. This will allow users to "opt in" to different naming conventions (including using the .NET Aspire 8.x naming scheme).

To do this, add a new class

```csharp
public class AzureProvisioningOptions
{
    public ProvisioningContext ProvisioningContext { get; } 
}
```

This will be an `IOptions` class, which can be configured as a typical IOptions object. Users can configure it to and modify the ProvisioningContext as necessary.

We also add a ResourceNameResolver that encodes the 8.x behavior:

```
public sealed class AzureResourceNamePropertyResolverAspireV8 : DynamicResourceNamePropertyResolver {}
```

Users who want to keep the 8.x naming scheme can write the following code at the beginning of their AppHost:

```csharp

builder.Services.Configure<AzureProvisioningOptions>(options =>
{
    options.ProvisioningContext.PropertyResolvers.Insert(0, new AzureResourceNamePropertyResolverAspirev8());
});
```

~We should add a compatibility switch to allow developers who upgrade to the new version of .NET Aspire to continue using the same "deployment version" as a previous version.~

~One way we could do this is to allow `Annotations` on the `IDistributedApplicationBuilder`. One of the annotations would be an `AzureDeploymentVersion` annotation that would have values like:~

* 8
* 9
* Current

~`Current` would be the default if there is no annotation specified. We would add new versions as we needed. And new breaking changes to the bicep would be gated behind a check if the specified AzureDeploymentVersion was "new enough".~

~This allows people to use the `9.0` .NET Aspire, but continue to use the old naming scheme.~

~We would also use this in other behavior changes - like when we change Azure PostgreSQL and Redis to use managed identity by default. Other "better defaults, but breaking changes" would do the same.~

eerhardt:
## Description

Add IOptions<AzureProvisioningOptions> which contains a ProvisioningContext. Users can customize the ProvisioningContext by configuring the AzureProvisioningOptions like a typical IOptions.

This allows us to use the new CDK default naming scheme. Users can opt into the old scheme by customizing the ProvisioningContext.

Fix #5756
Fix #5341

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1666
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5809)


src/Aspire.Hosting.Azure/AzureConstructResource.cs
```diff
@@ -55,8 +56,7 @@ public override BicepTemplateFile GetBicepTemplateFile(string? directory = null,
         var generationPath = Directory.CreateTempSubdirectory("aspire").FullName;
         var moduleSourcePath = Path.Combine(generationPath, "main.bicep");
 
-        var provisioningContext = GetProvisioningContext();
-        var plan = resourceModuleConstruct.Build(provisioningContext);
+        var plan = resourceModuleConstruct.Build(ProvisioningContext);
```
davidfowl, line 59: You still need a fallback

eerhardt: It allows `null`.

https://github.com/Azure/azure-sdk-for-net/blob/c8bdcfb11c5283c34da4693d5a700e565d80c384/sdk/provisioning/Azure.Provisioning/src/Infrastructure.cs#L193

It uses the `context ??= ProvisioningContext.Provider.GetProvisioningContext();`.

@tg-msft - is that going to change at all with the changes to remove the DefaultInfrastructure you were mentioning?

src/Aspire.Hosting.Azure/AzureConstructResource.cs
```diff
@@ -24,6 +24,7 @@ public class AzureConstructResource(string name, Action<ResourceModuleConstruct>
     /// Callback for configuring construct.
     /// </summary>
     public Action<ResourceModuleConstruct> ConfigureConstruct { get; internal set; } = configureConstruct;
+    internal ProvisioningContext? ProvisioningContext { get; set; }
```
davidfowl, line 27: nit:
```suggestion

    internal ProvisioningContext? ProvisioningContext { get; set; }
```

eerhardt:
This PR is now ready for review.

src/Aspire.Hosting.Azure/AzureProvisioningOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Aspire.Hosting.Azure;
+
+/// <summary>
+/// Options to configure the behavior of Azure resources.
+/// </summary>
+/// <remarks>
+/// These options can be configured using https://learn.microsoft.com/dotnet/core/extensions/options#use-di-services-to-configure-options
+/// to customize how Azure resources are provisioned.
+/// </remarks>
+public sealed class AzureProvisioningOptions
```
davidfowl, line 15: Not sure this should be provisioning. 
```suggestion
public sealed class AzureResourceOptions
```

eerhardt: Done.

davidfowl (approved):

PR #5832, opened 2024-09-22
Rename the dcp resource on restart

### LINKED ISSUE #5803: Logs break when restarting projects
Logs for projects are lost after restarting. 

https://github.com/user-attachments/assets/3c7da8f0-b458-46ba-a210-a284522813cc

davidfowl:
## Description

This change makes the old dcp resource as hidden and makes a new resource. The UI can be a bit jarring as items are removed and then added.

Fixes #5803

PS: It's unclear if we want this change. I just wanted to understand the behavior and potential fix. Instead DCP could reset the log stream state when a new resource is created.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
@@ -21,7 +21,7 @@ internal static class CommandsConfigurationExtensions
             {
                 var executor = context.ServiceProvider.GetRequiredService<ApplicationExecutor>();
 
-                await executor.StartResourceAsync(context.ResourceName, context.CancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


JamesNK:
I think this should be fixed in DCP. I logged an issue there.

The complete name is visible for replicas and changing the name will be visible in UI. Also, the telemetry uses the complete name so there will be two instances of the app in the telemetry pages.

PR #5835, opened 2024-09-22
Support Container References in `WithImage`

### LINKED ISSUE #4355: Support combined image and tags syntax in WithImage extension method
In many cases of using docker, we usually specify `image` + `tag` in one continious string in a following format:
"{image}:{tag}", for example like `postgres:14.3`. This notation is used with direct docker command, as well as on dockerhub or in Dockerfiles and Docker-compose files, and it might be a little bit better to be able just to copy one string and specify it in extension method. 

But current extension method specifies `tag = "latest"` directly, so you can't just set "postgres:14.3" in image. For me that was a surprise, and I've got runtime error: 

```json
      could not create the container    {"Container": {"name":"database"}, "Reconciliation": 3, "error": "CreateContainer command returned error code 1 Stdout: '' Stderr: 'invalid reference format\n'"}
```

In my opinion, change of extension method will add a little bit to a better developer experience:

Before:
```cs

builder.AddPostgres("database")
    .WithImage("postgres", "16.2")

builder.Build().Run();
```

After:
```cs

builder.AddPostgres("database")
    .WithImage("postgres:16.2")
    .WithPgAdmin();

builder.Build().Run();
```

### LINKED ISSUE #5834: `WithImage` overwrites existing tag

### Describe the bug

If you call `WithImage` on a resource that already has a tag specified, it ovewrites the existing tag with `latest`.

This is because `WithImage` has an optional parameter with a default value of `"latest"`.
https://github.com/dotnet/aspire/blob/af5fee35d59bb2a802aa0b3e2110623d4e7f4e9b/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L141

### Expected Behavior

Given that I didn't provide a tag to `WithImage()`, I'd expect the container to continue using my existing tag.

### Steps To Reproduce

```cs
[Fact]
public void WithImageShouldNotOverrideTag()
{
    var redis = builder.AddContainer("redis", "original-tag").WithImage("redis-stack");
    Assert.Equal("original-tag", redis.Resource.Annotations.OfType<ContainerImageAnnotation>().Single().Tag);
}
```

### Exceptions (if any)

```
    Assert.Equal() Failure: Strings differ
           ↓ (pos 0)
Expected: "original-tag"
Actual:   "latest"
           ↑ (pos 0)
```

### .NET Version info

Repro'd against commit af5fee35d59bb2a802aa0b3e2110623d4e7f4e9b (latest `main` at the time of writing)

### Anything else?

One solution to fix this would be to break up `WithImage(string image, string tag = "latest")` into two methods, however this would be a breaking change according to the PublicApi analysers:
```cs
public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image) where T : ContainerResource
{
   // Existing implementation
}

public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image, string tag = "latest") where T : ContainerResource
  => builder.WithImage(image).WithTag(tag);
```

A side effect of this change is that the `ContainerImageAnnotaiton.Tag` property will now be null for such.  The property already is nullable so consumers should be able to handle this possibility.  This also makes it consistent with `ContainerImageAnnotaiton.Repository` which we leave as null rather than defaulting to `docker.io`.

Fixing this will also provide a partial solution to #4355 as it is the addition of an additional `latest` tag that breaks container references given to `WithImage()`.  (Note, this is only a partial solution to that problem as whilst DCP will start the correct container, the `ContainerResourceAnnotation` is not correctly broken down to it's constituent parts.)

afscrome:
## Description

Updated `WithImage()` to properly handle full image references, parsing out out registry, tag & images.  This means you can use any string you pass to `docker run` for an image e.g. `WithImage("mcr.microsoft.com/dotnet/aspire-dashboard:8.2")`.

Fixes #4355
Fixes #5834

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [X] Yes
      - [ ] Yes
      - [X] No
      - [X] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5835)


src/Aspire.Hosting/Utils/ContainerReferenceParser.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        //TODO: Sort host name normalisation from https://github.com/distribution/reference/blob/main/normalize.go#L66
+        return new(domain, path, tag, digest);
+
+        static string? GetGroupValueOrNull(Group group)
+            => group.Success ? group.Value : null;
+    }
+
+    // Based on https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/regexp.go
+    public class ContainerParserRegexes
```
davidfowl, line 55: this should be using the regex source generator

src/Aspire.Hosting/Utils/ContainerReferenceParser.cs
```diff
@@ -0,0 +1,43 @@
...
+        if (string.IsNullOrEmpty(input))
+        {
+            throw new ArgumentOutOfRangeException(nameof(input), "repository name must have at least one component");
+        }
+
+        var match = ImageNameRegex().Match(input);
+
+        if (!match.Success)
+        {
+            throw new ArgumentOutOfRangeException(nameof(input), input, "invalid reference format");
```
afscrome, line 23: The error message isn't great, but it matches the error docker gives off, so 
https://github.com/distribution/reference/blob/8c942b0459dfdcc5b6685581dd0a5a470f615bff/reference.go#L72

src/Aspire.Hosting/Utils/ContainerReferenceParser.cs
```diff
@@ -0,0 +1,43 @@
...
+            match.Groups["image"].Value,
+            GetGroupValueOrDefault(match.Groups["tag"]),
+            GetGroupValueOrDefault(match.Groups["digest"])
+            );
+
+        static string? GetGroupValueOrDefault(Group group)
+            => group.Success ? group.Value : default;
+    }
+
+    [GeneratedRegex("^((?<registry>((localhost|\\[(?:[a-fA-F0-9:]+)\\]|([\\w-]+(\\.[\\w-]+)+))(:\\d+)?)|([\\w-]+:\\d+))\\/)?(?<image>[\\w-./<>]+)(:(?<tag>[\\w-.<>]+))?(@(?<digest>.+))?$")]
```
afscrome, line 37: Based on https://github.com/dotnet/aspire/blob/ea985ec4fd70b9eb8cb2f8c1a567372f5475c89c/src/Aspire.Hosting/Utils/ContainerReferenceParser.cs#L40 with addition for IPv6 addresses

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
```diff
@@ -138,18 +139,57 @@ public static IResourceBuilder<T> WithImageRegistry<T>(this IResourceBuilder<T>
...
+            imageAnnotation.Tag = parsedReference.Tag;
+        }
+        else if (tag is { })
+        {
+            imageAnnotation.Tag = tag;
+        }
+
+        if (imageAnnotation.Tag is null && imageAnnotation.SHA256 is null)
+        {
+            imageAnnotation.Tag = "latest";
```
afscrome, line 190: This bit could instead be done by defaulting the tag in `ContainerImageAnnotation`

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
```diff
@@ -138,18 +139,57 @@ public static IResourceBuilder<T> WithImageRegistry<T>(this IResourceBuilder<T>
     /// <param name="image">Image value.</param>
     /// <param name="tag">Tag value.</param>
     /// <returns></returns>
-    public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image, string tag = "latest") where T : ContainerResource
+    public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image, string? tag = null) where T : ContainerResource
```
afscrome, line 142: I've had to make an API breaking change here - specifically changing `string tag = "latest"` to `string? tag = null`.

Without this, `WithImage("foo:tag")` would become `WithImage("foo:tag", "latest")` which would be ambiguous.

Open to other suggestions

afscrome:
@danegsta @mitchdenny Do you mind taking a look at this?  The two test failures seem to be known issues - #6163

tests/Aspire.Hosting.RabbitMQ.Tests/AddRabbitMQTests.cs
```diff
@@ -163,7 +163,6 @@ public void WithManagementPluginThrowsForUnsupportedContainerImageTag(string ima
     }
 
     [Theory]
-    [InlineData(" ")]
```
mitchdenny, line 166: Why this change?

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
mitchdenny: If we are going to break it might be better to break it by removing the tag argument altogether.

Thoughts @davidfowl / @danegsta ?

tests/Aspire.Hosting.RabbitMQ.Tests/AddRabbitMQTests.cs
afscrome: `rabbitmq.WithImage(" ")` now throws `invalid reference format` before the test can even get as far as testing if `rabbitmq.WithManagementPlugin` throws.

This scenario is covered by https://github.com/dotnet/aspire/blob/03b07232818f172246dbc274b3fc1fa6a9a07cb6/tests/Aspire.Hosting.Tests/Utils/ContainerReferenceParserTests.cs#L20

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
afscrome: Would splitting the method up into two be a better option than changing the default value?
```cs
public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image) where T : ContainerResource

// Possibly with addition of [Obsolete("Use 'WithImage(image).WithTag(tag) instead.")]
public static IResourceBuilder<T> WithImage<T>(this IResourceBuilder<T> builder, string image, string tag) where T : ContainerResource
    => builder.WithImage(image).WithImageTag(tag);
```

For my understanding, what exactly is the breaking change the public api analyser is complaining about here?  Is it the `string!` to `string?` - I'd have thought that would be OK as we're widening allowed inputs, not narrowing or is it the change of the default value itself.  Or is it the change of the default value from `"latest"` to `null`?  I know changing the default value is problematic for binary compatibility as anyone who previously called the method will still have the old `"latest"` default baked in, however in this particular case we ultimately set the tag to `"latest"` if the tag is `null` so the impact of changing the default should be mostly non observable.

afscrome: Any thoughts @davidfowl / @danegsta ?

src/Aspire.Hosting/Utils/ContainerReferenceParser.cs
danegsta: Maybe something like "invalid reference format: could not parse container image name"?

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
davidfowl: I like the idea of making the default tag null and we’d internally still use latest only if the tag wasn’t set. Feels like we should make this change now if we’re going to do it

afscrome: @davidfowl  Just to check, do you mean handle the default value within the `WithImage` method as done in the PR (i.e. setting ContainerImageAnnotation.Tag to `latest` method if the tag argument is null / missing), or do you mean handle the default elsewhere (i.e. leave `ContainerImageAnnotation.Tag` as null and then handling that later on.)

davidfowl: probably as delayed as possible. Though I am concered at the number of places that might need to update the handling of a null tag.

tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
```diff
@@ -86,7 +86,7 @@ public async Task AddContainerWithArgs()
         """
         {
           "type": "container.v0",
-          "image": "none:latest",
+          "image": "none",
```
davidfowl, line 89: This seems bad

afscrome: If a user explicitly gives us a tag of `latest`, then we should include that in the manifest (i.e. `none:latest`).  If a user doesn't' give us a tag that's more open question.  My take was that the images `none` and `none:latest` are equivalent, so if a user doesn't give us a tag we should treat that by not explicitly including a tag.  Happy to put `latest` back in the manifest if you think that's better.

That said, I haven't put huge amounts of thought into this latest commit - mainly done the obvious change to see what tests break.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
afscrome: 3 types of tests broke

1. Tests asserting that `ContainerImageAnnotation.Tag` was `latest` when they didn't pass a tag. (expected to break as that's what we changed)
2. Mainfests for images with tags no longer have `:latest` in theim  (i.e. `nginx:latest` has become just `nginx`)
3. Image in bicep files without a tag no longer have `:latest` on them.

Both 2 & 3 share the same root cause with `TryGetContainerImageName` which currently has explicitly logic to not include the tag if one isn't set  https://github.com/dotnet/aspire/blob/0b562b8eb039343df763d6e9083da8fbb938604e/src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs#L242-L246

This change also removes the `:latest` from the dashboard UI if no tag is specified:

image

```cs
builder.AddContainer("no-tag", "nginx");
builder.AddContainer("explicit-latest", "nginx:latest");
```

So the question is should aspire explicitly set the tag to latest before passing to the manifest / container runtime, or should it just leave the tag blank and let that system deal with it?  The PR currently has the latter, but changing the former should be a matter of changing `TryGetContainerImageName`, backing out the test changes from b93a3f2f59c8e0804e887222576e9ab32f3a6f26 and 81a77b3b245a6cffe16d778af6608be6a0ce0f95 and adding some additional test coverage.

tests/Aspire.Hosting.RabbitMQ.Tests/AddRabbitMQTests.cs
afscrome: @mitchdenny does my explanation resolve your question about the test change?

davidfowl:
@afscrome if we want to take another swing at this, I think we should go with the new method, non breaking approach.

afscrome:
Superseded by #7254 7254

PR #5840, opened 2024-09-23
Remove lifecycle commands from dashboard resource

JamesNK:
## Description

Tweak to remove start/stop/restart from dashboard resource in the dashboard. Prevent us from clicking stop on the dashboard and breaking everything.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5840)

### DIFF
--- a/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
@@ -57,8 +57,14 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
     {
         if (appModel.Resources.SingleOrDefault(r => StringComparers.ResourceName.Equals(r.Name, KnownResourceNames.AspireDashboard)) is { } dashboardResource)
         {
-            var commands = dashboardResource.Annotations.OfType<ResourceCommandAnnotation>().ToList();
-            foreach (var command in commands)
+            // The dashboard is visible during development. Stopping the dashboard with a command breaks the dashboard.
+            // Remove the lifecycle commands so they're not accidently clicked in during development.
+            var lifecycleCommands = dashboardResource.Annotations
+                .OfType<ResourceCommandAnnotation>()
+                .Where(a => a.Type is CommandsConfigurationExtensions.StartType or CommandsConfigurationExtensions.StopType or CommandsConfigurationExtensions.RestartType)
+                .ToList();
+
+            foreach (var command in lifecycleCommands)
             {
                 dashboardResource.Annotations.Remove(command);
             }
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -31,28 +31,48 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
...
         }
 
         _dashboardLogsTask = WatchDashboardLogsAsync(_shutdownCts.Token);
 
         return Task.CompletedTask;
     }
 
+    public Task AfterResourcesCreatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
+    {
+        if (appModel.Resources.SingleOrDefault(r => StringComparers.ResourceName.Equals(r.Name, KnownResourceNames.AspireDashboard)) is { } dashboardResource)
```
drewnoakes, line 58: #5750 adds the concept of a "system" resource annotation, which could be used to filter here in a way that doesn't have special knowledge about the dashboard. We will also want to filter out stop/restart for dashboard extensions here too I suspect, which are also identifiable by annotation.

drewnoakes (approved):
Approving as #5750 hasn't merged. I think we should consider revisiting this when that's merged.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -31,28 +31,48 @@ internal sealed class DashboardLifecycleHook(IConfiguration configuration,
...
+        {
+            // The dashboard is visible during development. Stopping the dashboard with a command breaks the dashboard.
+            // Remove the lifecycle commands so they're not accidently clicked in during development.
```
JamesNK, line 61:
```suggestion
            // Remove the lifecycle commands so they're not accidently clicked during development.
```

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: A systems resource to where? The resource server? Or the app model?

drewnoakes: Both. It's an annotation on the resource in the app model, and exposed to clients of the resource service in the proto.

Issue #5558, opened 2024-09-05
Unable to detect what resources used `WaitFor`


afscrome:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

The `WaitOn` implementation in main doesn't seem to leave behind any observable marker of it's use on the resource.

It would be helpful if there was a way to interrogate which resources make use of `WaitFor`.  For example in a `BeforeStartEvent`, find all resources with `.WaitFor(db)` and add an additional `.WaitFor(migrator)`

### Describe the solution you'd like

I'd like some way to find out from an `IResource` that had `WaitFor` configured - likely through some kind of annotation.

davidfowl:
cc @mitchdenny

mitchdenny:
With the implementation of `WaitFor` in main we avoided introducing another annotation and instead just made use of the event subscriptions. The downside of this is that we don't have a lot of introspection ability which could be useful.

An example of where it could be useful is where we might want to redirect health check error details in logs to dependent resources to make the cause of a resource in a `Waiting` state more useful.

I'm contemplating introducing a `WaitForAnnotation` or a `DependencyAnnotation` and storing a reference to the subscription on it. Alternatively, @JamesNK was introducing a relationship annotation in #5311 (but has other priorities right now).

I think this is a good suggestion no matter which way we tackle it.

mitchdenny:
I'll work on this after #5515 merges in.

mitchdenny:
https://github.com/dotnet/aspire/pull/5842 should address this issue. As part of the PR we add some additional logging to make it clear when a dependent resource has finished waiting. Additionally the eventing hookup has been changed so that the waits occur concurrently so you'll see log entries for all dependency resources.

mitchdenny:
Fixed in #5842 but wasn't closed automatically. /cc @afscrome

PR #5842, opened 2024-09-23
WaitAnnotation and ResourceNotificationService.WaitForDependenciesAsync

### LINKED ISSUE #5557: Hard to see what `WaitFor` is really waiting on

With the `WaitFor` implementation in main, it's hard to see what a resource is actually waiting on.  Logs are logged when a resource starts waiting on a resource, but nothing gets logged once that wait is complete.

In the following example, it looks like we're waiting for both `sql` and `migrator`, however we're actually only waiting on `migrator` as `sql` has started up.
image

Similarly with a longer chain of dependencies, You may see that you're waiting on `B`, but `B` is in turn waitign on `C` (which could in turn be waiting on `D`).  

A quick solution is to include a log entry when a wait is complete

```
2024-09-05T19:16:18 Waiting for resource 'sql' to enter the 'Running' state.
2024-09-05T19:16:19 Waiting for resource 'migrator' to complete.
2024-09-05T19:16:20 Finished waiting for resource 'sql' to enter the 'Running' state.
```
That would at least allow you to work out what is outstanding.

A slightly better solution would be if there was an easy way to see what was outstanding from a single place, without having to calculate across multiple log entries.   Without adding additional UI, there's limited options with the append only logs - you could add a log along the lines of `Waiting on 2/3 dependencies - 'migrator' and 'foo' remain` log each time a dependency is removed.  Even better is if this could include transitive dependencies, but that likely gets messy quickly.

The ideal would be some kind of UI on the resource that could present the dependencies as some kind of tree which you could follow to see how far the dependencies have gotten.

### Additional context

_No response_

mitchdenny:
## Description

This PR introduces some changes related to `WaitFor`. Here is the high-level summary:

1. `WaitFor` and `WaitForCompletion` add a new annotation to the resource which captures whether to wait for completion or just until healthy.
2. When the DAB creates the app host it wires up to the `BeforeResourceStartedEvent` which calls out to RNS to wait for the resource dependencies.

Fixes #5557 #5558 

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No (existing test coverage is adequate)
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5842)


src/Aspire.Hosting/ApplicationModel/WaitAnnotation.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// The resource that will be waited on.
+    /// </summary>
+    public IResource Resource { get; } = resource;
+
+    /// <summary>
+    /// The subscription that is invoked to handle wait logic.
+    /// </summary>
+    public DistributedApplicationEventSubscription Subscription { get; } = subscription;
```
davidfowl, line 28: Why store this?

mitchdenny: I was initially thinking it might be useful to enable unsubscribing down the track. Its been removed now anyway.

afscrome (commented):
Awesome

davidfowl (approved):
This is a good change.

PR #5852, opened 2024-09-23
Add stop time to resource properties

### LINKED ISSUE #5623: Include end time for resources that complete

When a resource finishes, it would be nice if an `End Time` field could be added to the resource along side the `Start Time`

I'd like to see an `End Time` alongside the `Start Time`.  A `Duration` would also be nice.
image

### Additional context

Sample code: 
```cs
builder.AddExecutable("oneSec", "pwsh", ".",
    "-NoProfile",
    "-NoLogo",
    "-NonInteractive",
    "-Command",
    "Write-Host 'Starting Migration'; Start-Sleep -Seconds 1; Write-Host 'Done'");
```

JamesNK:
## Description

Add start and stop time to dashboard resource properties. The dashboard was previously using the created date as the start date.

Implications of changing to use start date:

* Waiting resources will have a blank start date. This seems like a Good Improvement.
* Anyone who has implemented the resource service will need to update their service to have the start date
* ~The PostgresDatabaseResource doesn't have a start date. I think custom resources will need some changes to set the start and stop dates correctly. @mitchdenny Do you know this area?~

Fixes https://github.com/dotnet/aspire/issues/5623

Demo:

start-time

image

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5852)


src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -27,6 +27,16 @@ public sealed record CustomResourceSnapshot
     /// </summary>
     public DateTime? CreationTimeStamp { get; init; }
```
JamesNK, line 28: We already have a creation time stamp which is different from the startup timestamp. The creation timestamp is displayed in the "Start time" column. Seems odd. Wouldn't the `StartupTimestamp` make more sense?

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -625,7 +627,9 @@ private CustomResourceSnapshot ToSnapshot(Container container, CustomResourceSna
                 new(KnownProperties.Container.Ports, GetPorts()),
             ],
             EnvironmentVariables = environment,
-            CreationTimeStamp = container.Metadata.CreationTimestamp?.ToLocalTime(),
+            CreationTimeStamp = container.Metadata.CreationTimestamp?.ToUniversalTime(),
+            StartupTimeStamp = container.Status?.StartupTimestamp?.ToUniversalTime().UtcDateTime,
```
JamesNK, line 631: @karolz-ms CreationTimestamp is a DateTime and StartupTimestamp is a DateTimeOffset. Do you know why the property types are different? Can they and should they be made to have a consistent type?

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -74,6 +74,7 @@ protected override void OnInitialized()
             new KnownProperty(KnownProperties.Resource.DisplayName, Loc[Resources.Resources.ResourcesDetailsDisplayNameProperty]),
             new KnownProperty(KnownProperties.Resource.State, Loc[Resources.Resources.ResourcesDetailsStateProperty]),
             new KnownProperty(KnownProperties.Resource.CreateTime, Loc[Resources.Resources.ResourcesDetailsStartTimeProperty]),
+            new KnownProperty(KnownProperties.Resource.StopTime, Loc[Resources.Resources.ResourcesDetailsStopTimeProperty]),
```
JamesNK, line 77: Should we display create and startup time in the properties, or would that just be confusing?

I think they would only display different values if we support starting a docker container without recreating it.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -625,7 +627,9 @@ private CustomResourceSnapshot ToSnapshot(Container container, CustomResourceSna
                 new(KnownProperties.Container.Ports, GetPorts()),
             ],
             EnvironmentVariables = environment,
-            CreationTimeStamp = container.Metadata.CreationTimestamp?.ToLocalTime(),
+            CreationTimeStamp = container.Metadata.CreationTimestamp?.ToUniversalTime(),
```
JamesNK, line 630: @mitchdenny @davidfowl Why are you converting to local time here? My experience is timestamps are UTC. Values are always converted to UTC when sent to the dashboard and the local browser time is used to convert it to a local time. Keeping it UTC up until that point seems good, but maybe there is a reason that I don't see.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
karolz-ms: I think that makes more sense, yes

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -74,6 +74,7 @@ protected override void OnInitialized()
             new KnownProperty(KnownProperties.Resource.DisplayName, Loc[Resources.Resources.ResourcesDetailsDisplayNameProperty]),
             new KnownProperty(KnownProperties.Resource.State, Loc[Resources.Resources.ResourcesDetailsStateProperty]),
             new KnownProperty(KnownProperties.Resource.CreateTime, Loc[Resources.Resources.ResourcesDetailsStartTimeProperty]),
```
drewnoakes, line 76: The distinction between create/start applies here too.

src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
```diff
@@ -352,6 +352,14 @@ public void Complete()
             _logStreamCts.Cancel();
         }
 
+        public void Reset()
```
drewnoakes, line 355: A comment on this method would be helpful.

It doesn't seem like the race condition within this method is too serious, but you still might consider enclosing the test-and-write within `lock (_lock)`.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
JamesNK: `CreationTimestamp` comes from `V1ObjectMeta`. Since the built-in types use DateTime, should our custom properties be consistent with them?

JamesNK: The date doesn't have an offset, so I'm guessing it comes back as UTC. Deserializing to DateTime should be fine. What do you think?

image

karolz-ms: I think whatever makes things more consistent is fine. I suggested `DateTimeOffset` initially because the K8s code is using Go time, which always has a location associated with it https://pkg.go.dev/time@go1.23.1#Time. But if C# Kubernetes client uses DateTime for timestamps, we should probably follow that.

JamesNK:
All done. Ready for review

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
davidfowl: No we don't care about the creation time

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: I don't recall who wrote this

JamesNK: Mitch couldn't remember either.

I haven't seen any regression from changing it.

davidfowl:
Aside: Can we make the detail view update live as well ?

davidfowl (approved):

JamesNK:
> Aside: Can we make the detail view update live as well ?

I think that might be a reasonable amount of work (half a day). Will long a new issue.

PR #5870, opened 2024-09-24
Replace health check publisher and scheduler with ResourceHealthCheckService and introduce ResourceReadyEvent.

### LINKED ISSUE #5602: Resource health check polling (cloud traffic concerns)
In #5601 we reduce the period for health check execution. We go from 30 seconds to 5 seconds. Some cloud hosted resources could be more expensive for health check intervals.

Do we need to introduce a `Period` property on the `HealthCheckAnnotation` and then improve the logic in `ResourceHealthCheckScheduler` to enable or disable a check based on whether the per resource internal has been reached. We might want to have two intervals, one for startup and one for normal running with the AddXYZResource method able to set the sensible default.

This would make the scheduler more complex.

### LINKED ISSUE #5638: ResourceHealthCheckPublisher should publish ResourceHealthyEvent
@mitchdenny This is where it would be useful to have a resource healthy event.

_Originally posted by @davidfowl in https://github.com/dotnet/aspire/pull/5227#discussion_r1751063723_

### LINKED ISSUE #5871: Add ability to dispatch DAE events concurrently and fire and forget
We need an API like the following:

```csharp
eventing.PublishAsync(@event, EventDispatch.FireAndForget);
```

The driver for introducing this is that there are some points in execution where we need to dispatch events but not block execution for timing critical code (health check publisher).

### LINKED ISSUE #5943: Synchronise health state between parent and child resources
After https://github.com/dotnet/aspire/pull/5936 merged in there is a brief inconsistency in the dashboard UX where a child resource can be in a Running state and the parent resource can be in a Running (Not ready) state. It would be better if the child resource copied the health status from the parent or considered the health status of the parent in its health check.

We could centralise some logic to copy this state but it would need to take into consideration health checks that are applied to the child resource directly (we don't do this by default but we need to allow for it) or we could have some kind of synthetic "parent health check" which follows the parents health status.

mitchdenny:
## Description

This PR removes the health check publisher and scheduler and combines it into a single `ResourceHealthCheckService` which runs in the background doing health checks on a per resource basis. 

Fixes #5602 
Fixes #5638
Fixes #5871
Fixes #5943
## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No (existing tests provide coverage - internal mechanism now relies on this event)
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5870)


src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
```diff
@@ -2,36 +2,78 @@
...
+                    var pendingSubscriptionCallback = subscription.Callback(@event, cancellationToken);
+                    pendingSubscriptionCallbacks.Add(pendingSubscriptionCallback);
+                }
+
+                if (dispatchBehavior == EventDispatchBehavior.NonBlockingConcurrent)
+                {
+                    // Non-blocking concurrent.
+                    _ = Task.Run(async () =>
+                    {
+                        await Task.WhenAll(pendingSubscriptionCallbacks).ConfigureAwait(false);
```
davidfowl, line 42: Log failures here

src/Aspire.Hosting/Health/ResourceNotificationHealthCheckPublisher.cs
```diff
@@ -2,20 +2,30 @@
...
 using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Eventing;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 
 namespace Aspire.Hosting.Health;
 
-internal class ResourceNotificationHealthCheckPublisher(DistributedApplicationModel model, ResourceNotificationService resourceNotificationService) : IHealthCheckPublisher
+internal class ResourceNotificationHealthCheckPublisher(DistributedApplicationModel model, ResourceNotificationService resourceNotificationService, IDistributedApplicationEventing eventing, IServiceProvider services) : IHealthCheckPublisher
 {
+    private readonly List<string> _alreadyHealthy = new();
```
davidfowl, line 12: HashSet

src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs
mitchdenny: We can't :( Logging isn't necessarily setup yet.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            var resourceEvents = resourceNotificationService.WatchAsync(stoppingToken);
+
+            await foreach (var resourceEvent in resourceEvents.ConfigureAwait(false))
+            {
+                _latestEvents[resourceEvent.Resource.Name] = resourceEvent;
+
+                if (!_resourceTasks.ContainsKey(resourceEvent.Resource.Name))
+                {
+                    var pendingResourceTask = Task.Run(() => MonitorResourceHealthAsync(resourceEvent, stoppingToken), stoppingToken);
```
davidfowl, line 29: We shouldn't start unless the state is running.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,119 @@
...
+            // NOTE: If there are no transitive health check annotations then there
+            //       is currently nothing to monitor. At this point in time we don't
+            //       dynamically add health checks at runtime. If this changes then we
+            //       would need to revisit this and scan for transitive health checks
+            //       on a periodic basis (you wouldn't want to do it on every pass.
+            logger.LogTrace("Stopping monitoring resource '{Resource}' has not health check annotations.", resource.Name);
+            return;
+        }
+
+        var registrationKeysToCheck = annotations.DistinctBy(a => a.Key).Select(a => a.Key).ToImmutableHashSet();
```
davidfowl, line 58: Just use a normal hashset or a frozen one.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,119 @@
...
+                        entry.Value.Status,
+                        entry.Value.Description
+                        );
+                }
+
+                if (_latestEvents[resource.Name].Snapshot.HealthStatus == report.Status)
+                {
+                    // If the last health status is the same as this health status then we don't need
+                    // to publish anything as it just creates noise.
+                    logger.LogTrace("Resource '{Resource}' health status has not changed from '{Status}'.", resource.Name, report.Status);
```
davidfowl, line 93: Maybe don't log either 😄

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,119 @@
...
+                        await resourceNotificationService.PublishUpdateAsync(resource, id, s => s with
+                        {
+                            HealthStatus = report.Status
+                        }).ConfigureAwait(false);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _ = ex;
```
davidfowl, line 115: ?

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -313,10 +313,7 @@ private void ConfigureHealthChecks()
             {
                 if (ExecutionContext.IsRunMode)
```
davidfowl, line 314: Delete this completely.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting.Health;
+
+internal class ResourceHealthCheckService(ILogger<ResourceHealthCheckService> logger, ResourceNotificationService resourceNotificationService, HealthCheckService healthCheckService) : BackgroundService
+{
+    private readonly Dictionary<string, Task> _resourceTasks = new();
+    private readonly Dictionary<string, ResourceEvent> _latestEvents = new();
```
davidfowl, line 15: I don't love these being fields if they aren't used outside of these 2 methods.

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: Need to keep it there for publish mode. But I've removed the other side of the branch.

davidfowl: Why? Who is doing health checks?

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
```diff
@@ -54,6 +54,47 @@ public static bool TryGetAnnotationsOfType<T>(this IResource resource, [NotNullW
...
     }
 
+    /// <summary>
+    /// Attempts to retrieve all annotations of the specified type from the given resource including transitively from parents.
+    /// </summary>
+    /// <typeparam name="T">The type of annotation to retrieve.</typeparam>
+    /// <param name="resource">The resource to retrieve annotations from.</param>
+    /// <param name="result">When this method returns, contains the annotations of the specified type, if found; otherwise, null.</param>
+    /// <returns>true if annotations of the specified type were found; otherwise, false.</returns>
+    public static bool TryGetTransitiveAnnotationsOfType<T>(this IResource resource, [NotNullWhen(true)] out IEnumerable<T>? result) where T : IResourceAnnotation
```
davidfowl, line 64: nit: This isn't transitive, it's including parents. Transitivity would be the other way.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
mitchdenny: We need them. I've changed the resource tasks one to a hashset since I don't (currently) need access to the tasks. But I do need the latest snapshot for each resource to determine whether I should publish a status update (to avoid noise).

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
mitchdenny: Cleaned this up a bit. When debugging its possible for a cancellation to occur but we don't want to stop the loop. So I watch for an operation cancelled exception but ignore it if it isn't from the stopping token for the background service.

src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs
mitchdenny: Renamed.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
davidfowl: Make sure there’s no concurrency issues

mitchdenny: Just realized I could move one into a local variable due to some refactoring I did.

The remaining one is only updated from the main loop, the monitoring loops just read from it periodically so its safe. We don't ever remove resources so we don't need to worry about an event not being present by the time the monitoring loop for a resource is up and running.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        catch (OperationCanceledException)
+        {
+            // This was expected as the token was canceled
+        }
+    }
+
+    private async Task MonitorResourceHealthAsync(ResourceEvent initialEvent, CancellationToken cancellationToken)
+    {
+
```
davidfowl, line 43: nit:
```suggestion
```

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -0,0 +1,127 @@
...
+        var resource = initialEvent.Resource;
+        var resourceReadyEventFired = false;
+
+        if (!resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out var annotations))
+        {
+            // NOTE: If there are no transitive health check annotations then there
```
davidfowl, line 49:
```suggestion
            // NOTE: If there are no health check annotations then there
```

davidfowl (approved):
Follow up: Make ResourceHealthCheckService more testable using the time provider and add more logs.

Issue #5563, opened 2024-09-05
Azure Functions + Aspire initial integration follow-ups


captainsafia:
This is a meta-issue to track some of the pending tasks from the initial work to add Azure Functions support to Aspire.

- [x] - Dual connection strings dilemma (https://github.com/dotnet/aspire/pull/5827) 
- [x] - Handling port defined in launch profile (https://github.com/dotnet/aspire/pull/5873)
- [x] - Add support for Service Bus resources (https://github.com/dotnet/aspire/pull/5593)
- [ ] -  Ensure implementation is resilient to storage configured with same name as default

## Dual connection strings dilemma (https://github.com/dotnet/aspire/pull/5827)

The current implementation relies on custom mappings of connection strings from the Aspire format to a format that the Azure Functions host (which uses the underlying Azure SDKs) can consume. This enables resource references to be passed through to the Azure Functions host, which facilitates interactions with the resources. However, Aspire-based connection strings are also required if we want to enable access to Aspire-injected resources in Azure Functions _workers_.

Currently, an uninvestigated bug occurs when both connect string key/value formats are injected in to the Azure Functions host. Azure Functions workers currently inherit connection strings from the host so the host must be able to successfully handle both formats in order to support the entire E2E.

 ## Handling port defined in launch profile (https://github.com/dotnet/aspire/pull/5873)

The default Azure Functions templates allow the user to define the port the Azure Functions host will listen on for processing HTTP triggers. Aspire's launch profile infrastructure doesn't expose APIs for being able to read this property value and expose the user-configured port as an HTTP endpoint. This means that we can only support ports that are randomly assigned and injected into the host service by Aspire. For an application to work successfully, there must be no port defined in the launch profile.

We should explore making changing to the launch profile APIs to support being able to introspect launch settings.

## Add support for Service Bus resources (https://github.com/dotnet/aspire/pull/5593)

Service Bus resources are not currently supported because there is no support for emulation of Service Bus, which makes end-to-end testing of the integration difficult (all the other testing issues aside 😅). There's also miscellaneous issues with referencing the ServiceBus extension for Azure Functions locally in our repo that need to be resolved.

## Ensure implementation is resilient to storage configured with same name as default

The implementation currently auto-configures a storage resource with the name `azure-functions-default-storage` when no storage resource is configured. This can present a problem if the user tries to allocate a storage resource with the same name:

```
builder.AddAzureStorage("azure-functions-default-storage");

[...]

PR #5873, opened 2024-09-24
Support loading port info from launch profile for Azure Functions

captainsafia:
## Description

- Make `IProjectMetadata.LaunchSettings`, `LaunchSettings`, and `LaunchProfile` public APIs
- Move launch profile-related logic to shared sources
- Add support for loading port info from launch profile for Azure Functions

Contributes towards https://github.com/dotnet/aspire/issues/5563 and https://github.com/dotnet/aspire/issues/4877

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [X] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [X] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5873)


davidfowl:
@mitchdenny take a look at the public APIs please

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -93,6 +95,29 @@ public static class AzureFunctionsProjectResourceExtensions
...
+    {
+        var launchProfile = builder.Resource.GetEffectiveLaunchProfile();
+        int? port = null;
+        if (launchProfile is not null)
+        {
+            var commandLineArgs = CommandLineArgsParser.Parse(launchProfile.LaunchProfile.CommandLineArgs ?? string.Empty);
+            if (commandLineArgs is { Count: > 0 })
+            {
+                var indexOfPort = commandLineArgs.IndexOf("--port");
+                port = indexOfPort > -1 ? int.Parse(commandLineArgs[indexOfPort + 1], CultureInfo.InvariantCulture) : null;
```
davidfowl, line 115: Make this a bit more robust (argument out of range etc).

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -93,6 +95,29 @@ public static class AzureFunctionsProjectResourceExtensions
...
+            }
+        }
+        return builder.WithHttpEndpoint(port: port, targetPort: port, isProxied: port == null);
```
davidfowl, line 118: Not understanding why you are setting the target port and proxied.

```suggestion
        return builder.WithHttpEndpoint(port: port);
```

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -93,6 +95,29 @@ public static class AzureFunctionsProjectResourceExtensions
...
+    /// <summary>
+    /// Configures the Azure Functions project resource to use the specified port as its HTTP endpoint.
+    /// This method queries the launch profile of the project to determine the port to
+    /// use based on the command line arguments configure in the launch profile,
+    /// </summary>
+    /// <param name="builder">The resource builder for the Azure Functions project resource.</param>
+    /// <returns>An <see cref="IResourceBuilder{AzureFunctionsProjectResource}"/> for the Azure Functions project resource with the endpoint configured.</returns>
+    private static IResourceBuilder<AzureFunctionsProjectResource> WithFunctionsHttpEndpoint(this IResourceBuilder<AzureFunctionsProjectResource> builder)
+    {
+        var launchProfile = builder.Resource.GetEffectiveLaunchProfile();
```
davidfowl, line 107: We don't need to block this PR but I think we'll also want to figure out how to let support the entire launch profile (application url etc).

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
captainsafia: It looks like Azure Functions prioritizes the port defined in the launch settings over the one that it receives in the command line args when it's configured, so the target port that we configure in the `WithArgs` callback is disregarded.

I believe this is the right change we want to make if we want to allow Functions to set the port the underlying worker will listen on based on launchsettings and mapping the same port value out.

davidfowl: You are right. Since it's listening already that makes sense.

src/Aspire.Hosting/Aspire.Hosting.csproj
```diff
@@ -33,13 +33,12 @@
...
     <Compile Include="$(SharedDir)SchemaUtils.cs" Link="Utils\SchemaUtils.cs" />
-    <Compile Include="$(SharedDir)LaunchSettings.cs" Link="LaunchSettings.cs" />
-    <Compile Include="$(SharedDir)LaunchProfile.cs" Link="LaunchProfile.cs" />
-    <Compile Include="$(SharedDir)LaunchSettingsSerializerContext.cs" Link="LaunchSettingsSerializerContext.cs" />
     <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
     <Compile Include="$(SharedDir)ConsoleLogs\LogEntries.cs" Link="Utils\ConsoleLogs\LogEntries.cs" />
     <Compile Include="$(SharedDir)ConsoleLogs\LogEntry.cs" Link="Utils\ConsoleLogs\LogEntry.cs" />
     <Compile Include="$(SharedDir)ConsoleLogs\TimestampParser.cs" Link="Utils\ConsoleLogs\TimestampParser.cs" />
+    <Compile Include="$(SharedDir)CommandLineArgsParser.cs" Link="Utils\CommandLineArgsParser.cs" />
+    <Compile Include="$(SharedDir)LaunchProfiles\*.cs" />
```
mitchdenny, line 41: This doesn't seem right. There is code in `ApplicationExecutor` that relies on `LaunchProfileAnnoation`. This means we'll have two different `LaunchProfileAnnotation` types - the the on that `Aspire.Hosting` understands and the one that `Aspire.Hosting.Azure.Functions` understands.

captainsafia: This annotation seems to only store the name of the launch profile to be used. Is there a particular concern for the divergence, especially given they reference the same shared source for this implementation?

mitchdenny: Right now I think it is OK because we aren't using the LaunchProfileAnnotation in a way that the divergence will hurt us, but I'm thinking of how much of a pain to debug this might be in the future when we are looking for the launchprofileannotation type and its not showing up :)

davidfowl:
Aspire.Hosting.ProjectUtilities
Aspire.Hosting.Project.Utilities
Aspire.Hosting.Projects

tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs
```diff
@@ -106,6 +105,7 @@ private sealed class TestProject : IProjectMetadata
                 ["funcapp"] = new()
                 {
                     CommandLineArgs = "--port 7071",
+                    LaunchBrowser = false,
```
davidfowl, line 108: ?

captainsafia: This came because I did the revert [here](https://github.com/dotnet/aspire/pull/5873/commits/cbd20d582300759c81c8219360aa058be7f5d8be). The original test data set `LaunchBrowser = false` for no reason really. :/

davidfowl (approved):

PR #5899, opened 2024-09-24
Fix ClientOptions configuration binding for Event Hubs clients

### LINKED ISSUE #5903: Event Hubs component does not pass client options to constructed clients in all conditions

### Describe the bug

It seems that when binding an Aspire component (in this case EventProcessorClient) to configuration like this:

"Aspire": {
  "Azure": {
    "Messaging": {
      "EventHubs": {
        "EventProcessorClient": {
          "EventHubName": "hub1",
          "BlobContainerName": "checkpoints",            
          "ClientOptions": {
            "Identifier": "Processor1"
          }
        },
... EventHubs:* binds, and will correctly set EventHubName and BlobContainerName but Identifier is not assigned.

The problem is that the logic that constructs clients does not pass the ClientOptions instance in all branches :/

e.g.

                if (!string.IsNullOrEmpty(settings.ConnectionString))
                {
                    if (!string.IsNullOrEmpty(settings.EventHubName))
                    {
                        return new EventProcessorClient(containerClient,
                                                consumerGroup,
                                                settings.ConnectionString,
                                                settings.EventHubName); // << missing options
                    }
                    else
                    {
                        return new EventProcessorClient(containerClient,
                                                consumerGroup,
                                                settings.ConnectionString); // << missing options
                    }
                }
                else
                {
                    return new EventProcessorClient(containerClient,
                        consumerGroup,
                        settings.FullyQualifiedNamespace,
                        settings.EventHubName, cred, options); // includes options
                }

### Expected Behavior

Options should be included in all cases.

### Steps To Reproduce

Tests will be added in PR

### Exceptions (if any)

_No response_

### .NET Version info

n/a

### Anything else?

_No response_

oising:
Fixes #5903 

### Changes

- All components had slightly different patterns for constructing clients according to configuration, although the logic was identical for all. I have normalized the construction logic for all components to encompass the pattern of one of the following:
  - connectionstring (includes EntityPath)
  - connectingstring & settings.EventHubName (overrides or provides EntityPath)
  - fullyqualifiednamespace & settings.EventHubName (provides EntityPath)
  - partitionreceiver ctor did not follow the same logic as the other clients. Now it does.

- Tests added for all branches of the above logic

- All constructors now provide the correct ClientOptions in the ctor, so bound config is correctly provided for all branches of client construction

### Problem(s)

When binding an Aspire component (in this case EventProcessorClient) to configuration like this:

```json
"Aspire": {
  "Azure": {
    "Messaging": {
      "EventHubs": {
        "EventProcessorClient": {
          "EventHubName": "hub1",
          "BlobContainerName": "checkpoints",            
          "ClientOptions": {
            "Identifier": "Processor1"
          }
        },
```

... EventHubs:* binds, and will correctly set `EventHubName` and `BlobContainerName` but `Identifier` is **not** assigned, and instead a random UUID is returned.

The problem is that the logic that constructs clients does not pass the ClientOptions instance in all branches :/

e.g.

```csharp
                if (!string.IsNullOrEmpty(settings.ConnectionString))
                {
                    if (!string.IsNullOrEmpty(settings.EventHubName))
                    {
                        return new EventProcessorClient(containerClient,
                                                consumerGroup,
                                                settings.ConnectionString,
                                                settings.EventHubName); // << missing options
                    }
                    else
                    {
                        return new EventProcessorClient(containerClient,
                                                consumerGroup,
                                                settings.ConnectionString); // << missing options
                    }
                }
                else
                {
                    return new EventProcessorClient(containerClient,
                        consumerGroup,
                        settings.FullyQualifiedNamespace,
                        settings.EventHubName, cred, options); // includes options
                }
```

/cc @sebastienros 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5899)

### DIFF
--- a/tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
+++ b/tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
@@ -108,6 +108,47 @@ public class AspireEventHubsExtensionsTests
     }
 
     [Theory]
+    [InlineData(false, EventProcessorClientIndex)]
+    [InlineData(true, EventProcessorClientIndex)]
+    public async Task BindsEventProcessorClientOptionsIdentifier(bool useKeyed, int clientIndex)
+    {
+        var builder = Host.CreateEmptyApplicationBuilder(null);
+
+        ConfigureBlobServiceClient(useKeyed, builder.Services);
+
+        var key = useKeyed ? "eh" : null;
+        builder.Configuration.AddInMemoryCollection([
+            new KeyValuePair<string, string?>(
+                CreateConfigKey(
+                    $"Aspire:Azure:Messaging:EventHubs:{s_clientTypes[clientIndex].Name}",
+                    key, "ClientOptions:Identifier"), "customidentifier"),
+            new KeyValuePair<string, string?>(
+                CreateConfigKey(
+                    $"Aspire:Azure:Messaging:EventHubs:{s_clientTypes[clientIndex].Name}",
+                    key, "BlobContainerName"), "checkpoints"),
+            new KeyValuePair<string, string?>("ConnectionStrings:eh", EhConnectionString),
+        ]);
+
+        if (useKeyed)
+        {
+            s_keyedClientAdders[clientIndex](builder, "eh", null);
+        }
+        else
+        {
+            s_clientAdders[clientIndex](builder, "eh", null);
+        }
+
+        using var host = builder.Build();
+
+        await Task.Delay(500);
+
+        var client = RetrieveClient(key, clientIndex, host) as EventProcessorClient;
+
+        Assert.NotNull(client);
+        Assert.Equal("customidentifier", client.Identifier);
+    }
+
+    [Theory]
     [InlineData(false, EventHubProducerClientIndex)]
     [InlineData(true, EventHubProducerClientIndex)]
     [InlineData(false, EventHubConsumerClientIndex)]
--- a/tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
+++ b/tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
@@ -11,11 +11,12 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Xunit;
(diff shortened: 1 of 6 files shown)


oising:
This PR contains deliberate failing tests and will not the checks above :)

tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
```diff
@@ -107,6 +108,209 @@ public void ProcessorClientShouldNotTryCreateContainerWithBlobContainerSpecified
...
+                CreateConfigKey(
+                    AspireEventHubsSection + s_clientTypes[clientIndex].Name, key, "PartitionId"), "foo"),            
+            new KeyValuePair<string, string?>(
+                CreateConfigKey(
+                    $"Aspire:Azure:Messaging:EventHubs:{s_clientTypes[clientIndex].Name}",
+                    key, "EventHubName"), "MyHub"),
+            new KeyValuePair<string, string?>("ConnectionStrings:eh", EhConnectionString),
+        ]);
+
+        output.WriteLine(builder.Configuration.GetDebugView());
```
sebastienros, line 287: Do you want to keep these or did it just matter during dev.

sebastienros (approved):

tests/Aspire.Azure.Messaging.EventHubs.Tests/AspireEventHubsExtensionsTests.cs
oising: Personally, I prefer to keep it to make the test results more understandable for anyone else who wants to extend or clone them, but I guess it was only a dev thing.

sebastienros: @radical Do you think this could be an issue in terms of space for logs or do we not care about it?

oising: @sebastienros Let me just kill the line if it's going to delay the merge. I'd like to make the 9.x cutoff.

oising: I've removed the lines and pushed. I'll resolve this now.

sebastienros:
@oising hopefully [this PR](https://github.com/dotnet/aspire/pull/5927) will unblock the build

PR #5906, opened 2024-09-25
Fix ReplicasAndProxylessEndpointThrows

### LINKED ISSUE #5898: Regression: ReplicasAndProxylessEndpointThrows test fails
`ReplicasAndProxylessEndpointThrows` is disabled in CI per #4651, but I was running it frequently and it was passing reliably.

It became broken a few days ago due to a5b3bcfddb3478ab34f207251929d531322d1a85. The root of the problem is that the new code never calls `ExecutableReplicaSet.Create`, and no `ExecutableReplicaSet` instance is ever created.

Yet `ApplicationExecutor` has error handling logic that tests for it, and this is why the test is failing:

```c#
if (appResource.DcpResource is ExecutableReplicaSet ers && ers.Spec.Replicas > 1)
{
    throw new InvalidOperationException($"Resource '{modelResourceName}' uses multiple replicas and a proxy-less endpoint '{ea.Name}'. These features do not work together.");
}
```

It seems like a valid error detection situation, which has regressed.

JamesNK:
## Description

Added annotations to exe resources to identify if they're part of an unofficial replica set.

Fixes https://github.com/dotnet/aspire/issues/5898

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5906)

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -5,6 +5,7 @@ using System.Collections.Concurrent;
 using System.Collections.Immutable;
 using System.Data;
 using System.Diagnostics;
+using System.Globalization;
 using System.Net.Sockets;
 using System.Text.Json;
 using System.Threading.Channels;
@@ -1102,6 +1103,8 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                 exeSpec.Annotate(CustomResource.OtelServiceNameAnnotation, project.Name);
                 exeSpec.Annotate(CustomResource.OtelServiceInstanceIdAnnotation, nameSuffix);
                 exeSpec.Annotate(CustomResource.ResourceNameAnnotation, project.Name);
+                exeSpec.Annotate(CustomResource.ResourceReplicaCount, replicas.ToString(CultureInfo.InvariantCulture));
+                exeSpec.Annotate(CustomResource.ResourceReplicaIndex, i.ToString(CultureInfo.InvariantCulture));
 
                 SetInitialResourceState(project, exeSpec);
 
@@ -1817,7 +1820,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
             }
             else if (!ea.IsProxied)
             {
-                if (appResource.DcpResource is ExecutableReplicaSet ers && ers.Spec.Replicas > 1)
+                if (HasMultipleReplicas(appResource.DcpResource))
                 {
                     throw new InvalidOperationException($"Resource '{modelResourceName}' uses multiple replicas and a proxy-less endpoint '{ea.Name}'. These features do not work together.");
                 }
@@ -1837,7 +1840,7 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
                         $"The endpoint '{ea.Name}' for resource '{modelResourceName}' requested a proxy ({nameof(ea.IsProxied)} is true). Non-container resources cannot be proxied when both {nameof(ea.TargetPort)} and {nameof(ea.Port)} are specified with the same value.");
                 }
 
-                if (appResource.DcpResource is ExecutableReplicaSet && ea.TargetPort is int)
+                if (HasMultipleReplicas(appResource.DcpResource) && ea.TargetPort is int)
                 {
                     throw new InvalidOperationException(
                         $"Resource '{modelResourceName}' can have multiple replicas, and it uses endpoint '{ea.Name}' that has {nameof(ea.TargetPort)} property set. Each replica must have a unique port; setting {nameof(ea.TargetPort)} is not allowed.");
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1837,7 +1840,7 @@ private void AddServicesProducedInfo(IResource modelResource, IAnnotationHolder
                         $"The endpoint '{ea.Name}' for resource '{modelResourceName}' requested a proxy ({nameof(ea.IsProxied)} is true). Non-container resources cannot be proxied when both {nameof(ea.TargetPort)} and {nameof(ea.Port)} are specified with the same value.");
                 }
 
-                if (appResource.DcpResource is ExecutableReplicaSet && ea.TargetPort is int)
+                if (HasMultipleReplicas(appResource.DcpResource) && ea.TargetPort is int)
```
JamesNK, line 1843: I'm not sure about this change. Why did this throw if the resource could have replicas rather than check if there are multiple replicas?

JamesNK:
cc @davidebbo

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: I do not think an ApplicationExecutor change or DCP custom annotation is necessary here. The problem should be easy to detect at Aspire model level. 

The gist of the check is that if someone has more than one instance of a program that uses a "proxy-less" endpoint, that will surely end up in a port conflict.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1849,6 +1852,19 @@ private void AddServicesProducedInfo(IResource modelResource, IAnnotationHolder
...
             appResource.ServicesProduced.Add(sp);
         }
+
+        static bool HasMultipleReplicas(CustomResource resource)
+        {
+            if (resource is ExecutableReplicaSet ers && ers.Spec.Replicas > 1)
+            {
+                return true;
+            }
+            if (resource is Executable exe && exe.Metadata.Annotations.TryGetValue(CustomResource.ResourceReplicaCount, out var value) && int.TryParse(value, CultureInfo.InvariantCulture, out var replicas) && replicas > 1)
```
mitchdenny, line 1862: nit: add a line here

mitchdenny:
I think given that the previous check was already in `ApplicationExecutor` ... it makes sense to reenable it there. Picking up what @davidebbo was saying in #5898 about this being a validation problem. I agree and I wonder whether we need to provide some kind of mechanism to allow each resource and its constituent annotations to be validated at startup (or every time a change is made to a resource).

src/Aspire.Hosting/Dcp/Model/ModelCommon.cs
```diff
@@ -25,6 +25,8 @@ internal abstract class CustomResource : KubernetesObject, IMetadata<V1ObjectMet
     public const string OtelServiceNameAnnotation = "otel-service-name";
     public const string OtelServiceInstanceIdAnnotation = "otel-service-instance-id";
     public const string ResourceStateAnnotation = "resource-state";
+    public const string ResourceReplicaCount = "resource-replica-count";
```
mitchdenny, line 28: Can we achieve this without the custom DCP annotation? If you can access the IResource you should be able to validate that here.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
karolz-ms: James and I chatted. This is fine for now, and will help if ExecutableReplicaSets are used again someday.

karolz-ms (approved):

src/Aspire.Hosting/Dcp/Model/ModelCommon.cs
JamesNK: Karol and I talked about this. It could come from the model. But we agreed that it was kind of useful to have this info directly on a DCP resource, and it's similar to true replica sets which has this info on the DCP resource.

Will leave as annotations for now, but it is internal logic so can be changed in the future if desired.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
davidfowl: We should just use the ModelResource. It's much simpler 😄

Issue #5908, opened 2024-09-25
Wait for and resource health logging isn't visible


JamesNK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

`WaitUntilHealthyAsync` and `WaitUntilCompletionAsync` write to a logger to report status, but the logs aren't visible in the console logs.

The problem is they call `GetLogger(IResource)` to get a logger, but this logger doesn't include the full resource name with the unique suffix.

e.g. the wait methods write to a logger called `basketservice` but DCP logs go to, and the dashboard UI gets logs from `basketservice-abcxyz`.

### Expected Behavior

Wait for logs are visible in console logs from the dashboard.

JamesNK:
Things work great if you fetch a logger with the complete resource name, e.g. basketservice-abcxyz, like [is done when executing a resource command](https://github.com/dotnet/aspire/blob/90bf558bcc3106e85950f934b1423612c6e5a350/src/Aspire.Hosting/Dashboard/DashboardServiceData.cs#L98-L100). But it doesn't work when getting a logger using IResource, which just uses name, e.g. basketservice. Because the dashboard uses the complete name, logs written to basketservice are never viewed. [WaitFor is getting a logger like this](https://github.com/dotnet/aspire/blob/90bf558bcc3106e85950f934b1423612c6e5a350/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L125-L126) so its logs aren't visible once the resource actually starts. I logged an issue about it: https://github.com/dotnet/aspire/issues/5908

**Idea to fix:**

The problem is resources have a name that changes when DCP starts. Before: basketservice, After: basketservice-abcxyz. I suggest we fix this by not waiting for a DCP resource to start before generating the complete name. The random suffix [is already generated in the host](https://github.com/dotnet/aspire/blob/90bf558bcc3106e85950f934b1423612c6e5a350/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L1061). We can generate this earlier when the model type is created and store it in an annotation. If something is a replica, we can generated multiple at once. This would be like doing what [ReplicaInstancesAnnotation](https://github.com/dotnet/aspire/blob/90bf558bcc3106e85950f934b1423612c6e5a350/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs#L669-L672) does, except in advance instead of waiting for the resource to be returned from DCP watch.

[...]

PR #5909, opened 2024-09-25
Fix console logs after restart

### LINKED ISSUE #5803: Logs break when restarting projects

JamesNK:
PR changes:

* Don't complete a resource logger when a resource is deleted. The only scenario where resources are deleted right now is restart, and we still want logs to be available after a restart.
* Correctly display in-memory logs, i.e. logs written by resource logger service, when viewing the console logs page. They're no longer cleared away when console logs watch stops.

Fixes https://github.com/dotnet/aspire/issues/5803

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5909)

### DIFF
--- a/src/Aspire.Dashboard/Otlp/Model/MetricValues/DimensionScope.cs
+++ b/src/Aspire.Dashboard/Otlp/Model/MetricValues/DimensionScope.cs
@@ -3,7 +3,6 @@
 
 using System.Diagnostics;
 using Aspire.Dashboard.Configuration;
-using Aspire.Dashboard.Otlp.Storage;
 using Google.Protobuf.Collections;
 using OpenTelemetry.Proto.Metrics.V1;
 
--- a/src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceLoggerService.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Threading.Channels;
+using Aspire.Dashboard.Otlp.Storage;
 using Aspire.Hosting.ConsoleLogs;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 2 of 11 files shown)


mitchdenny:
@drewnoakes this is relevant to the conversation we were having today about health checks UX in the dashboard. Specifically this bit:

> Correctly display in-memory logs, i.e. logs written by resource logger service, when viewing the console logs page. They're no longer cleared away when console logs watch stops.

For 9.0 we might be best to just write the resource log with error messages from health checks. Previously it wasn't that great of an experience, but by persisting the log messages between visits to the logs page it changes things.

mitchdenny:
Hrm ... this still doesn't preserve the logs from before the process starts. The `WaitFor` implementation uses the resource logger to output some messages like the following:

image

The process can finish starting up and be running and not be showing the Kestrel log outputs, but if I navigate away and come back I see this:

image

You can use the _waitfor_ playground to verify this. You need to look at the console logs quickly though before it has a chance to start (only takes about 3 seconds once the dashboard is visible).

mitchdenny:
Subtle observation, if you catch it early enough the text in the drop down is different:

image

But once the process is running it changes to the screenshot I have up above (the first one).

JamesNK:
The problem is in how WaitFor does logging. I created an issue for it: https://github.com/dotnet/aspire/issues/5908

The problem is that IResource doesn't know about the unique name that is generated by DCP. Maybe the value could be generated ealier and stored in an annotation. Then GetLogger(IResource) would have the right name. That wouldn't work for resources with replicas where there are multiple names.

mitchdenny:
Hrm ... realistically if we had two replicas, both replicas would be getting held up by the same set of health checks - so we would want the log to show up in both places. We might need an API that allows us to do some kind of "aggregate logging" across replicas.

adamint (approved):

PR #5929, opened 2024-09-25
Add unknown state, add to transitory states

### LINKED ISSUE #5550: Unknown container state shows up green in the dashboard
I thought we fixed this @adamint ?

image

Repro is to delete the container using docker desktop and leave the dashboard running. 

PS @karolz-ms @danegsta, shouldn't DCP auto recreate the container in this case? The object hasn't been deleted at the DCP level, just the backing container in the container runtime.

adamint:
## Description

Fixes unknown state appearing as success in the resource state column.

Fixes #5550

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5929)

### DIFF
--- a/src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
+++ b/src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
@@ -34,7 +34,7 @@
                     Class="severity-icon" />
     }
 }
-else if (Resource.IsUnusableTransitoryState())
+else if (Resource.IsUnusableTransitoryState() || Resource.IsUnknownState())
 {
     <FluentIcon Icon="Icons.Regular.Size16.CircleHint"
                 Color="Color.Info"
--- a/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
+++ b/src/Aspire.Dashboard/Extensions/ResourceViewModelExtensions.cs
@@ -29,8 +29,10 @@ internal static class ResourceViewModelExtensions
 
     public static bool IsUnusableTransitoryState(this ResourceViewModel resource)
     {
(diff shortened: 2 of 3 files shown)


JamesNK:
I believe start/stop/restart will be disabled when unknown with this change. There should be an option to stop/restart when in an unknown state to see if that fixes it.

Also add some tests for the lifecycle command states to verify unknown allows this.

adamint:
> I believe start/stop/restart will be disabled when unknown with this change. There should be an option to stop/restart when in an unknown state to see if that fixes it.

It shouldn't, as commands are obtained in `Partials.ToViewModel`, which does not invoke this method. However, you're right that Unknown isn't a good match for the `IsUnusableTransitoryState` method. Moved into its own method.

adamint:
> Also add some tests for the lifecycle command states to verify unknown allows this.

I don't see the benefit of this, if commands are obtained and filtered elsewhere.

JamesNK:
It's not directly related to this change, but since we're special casing Unknown state here, it would be good to also test that the state and lifecycle commands are what are expected.

It's just 3 lines here: https://github.com/dotnet/aspire/blob/624b042ccab9ddbde854d294d3ad9a6115d2fbd0/tests/Aspire.Hosting.Tests/ResourceCommandAnnotationTests.cs#L33

JamesNK (approved):

Issue #3299, opened 2024-03-30
Put all database connection strings in keyvault


davidfowl:
Today database connection string always grab the parent connection string (the database server) and they append the database portion of the connection string. This is fine for resources that don't put their connection strings in key vault, but we can optimize the experience so that the deployment tool never needs to grab keyvault secrets to create container apps.

davidfowl:
Huge benefit as the secrets are auto pulled from keyvault every 30 mins https://learn.microsoft.com/en-us/azure/container-apps/manage-secrets?tabs=azure-portal

mitchdenny:
Need to have some more discussion about this.

PR #5930, opened 2024-09-25
Add Azure PostgreSQL managed identity

### LINKED ISSUE #5793: Azure PostgreSQL should support managed identity
When provisioning a PostgreSQL database server today, we use an admin user name and password. This approach isn't as secure as using managed identity.

See

* https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.PostgreSQL.AAD/
* https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.PostgreSQL.AADOnly/

We should have an option for enabling managed identity / Entra ID when provisioning PostgreSQL database servers.

One could argue that we should do this by default, in the name of security. However, the problem is that the app connecting to this database is going to have to change to follow the steps in https://devblogs.microsoft.com/dotnet/using-postgre-sql-with-dotnet-and-entra-id/. So it would break apps when they published to Azure, if the app wasn't set up to work with Entra ID.

### Proposal

Add a new enum that gets passed into `.AsAzurePostgresFlexibleServer()` that switches which kind of auth is used.

```C#

public enum AzurePostgresAuth
{
    Password = 0,
    ActiveDirectory = 1 // Note this is the name used in ARM, Bicep, CDK
}

// user code
var pgsqldb = builder.AddPostgres("pgsql")
    .AsAzurePostgresFlexibleServer(AzurePostgresAuth.ActiveDirectory)
    .AddDatabase("pgsqldb");
```

### Alternative

We could add a new extension method `.WithActiveDirectoryAuth()` that disables the password based auth and switches the administrator to use managed identity.

```C#
var pgsqldb = builder.AddPostgres("pgsql")
    .AsAzurePostgresFlexibleServer()
    .WithActiveDirectoryAuth()
    .AddDatabase("pgsqldb");
```

The downfall of this is that it would need to hang off `IResourceBuilder<PostgresServerResource>` and if someone called it without calling `AsAzurePostgresFlexibleServer` first it wouldn't work well.

cc @mitchdenny @davidfowl @tg-msft @sebastienros @radical

eerhardt:
## Description

- Add a new API to replace AsAzure and PublishAsAzure. This creates an Azure resource to start, so it can be configured like other azure resources. Since this is a new API, we can use managed identity by default.

- Add an API to convert managed identity to password based auth, so users can still go that route if they need to.

- Add RunAsContainer following the RunAsEmulator model. This allows local development against a PostreSQL container and publish to go to a manged Azure PostgreSQL flexible server.

Fix #5793

## Checklist

- Is this feature complete?
  - [x] No. Follow-up changes expected. Still need https://github.com/dotnet/aspire/issues/5998
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [x] Yes
  - [x] Yes
      - [x] Yes
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1704

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5930)


davidfowl:
@eerhardt Can you do this as well https://github.com/dotnet/aspire/issues/3299? I think postgres is the only database that we need to fix.

mitchdenny:
@eerhardt is this something you want to get in for 9.0 or 9.x?

mitchdenny:
The downside of this approach is composability of resources. Let's say I wanted to have a resource that depended on a Postgres database instance. With this change the implementor of that would need to have extensions for each cloud provider (unless they just allow any IResourceWithConnectionString).

davidfowl:
> The downside of this approach is composability of resources. Let's say I wanted to have a resource that depended on a Postgres database instance. With this change the implementor of that would need to have extensions for each cloud provider (unless they just allow any IResourceWithConnectionString).

That can be solved with an IPostgresResource interface right?

eerhardt:
> @eerhardt is this something you want to get in for 9.0 or 9.x?

9.0. [Azure PostgreSQL should support managed identity (dotnet/aspire#5793)](https://github.com/dotnet/aspire/issues/5793) is in the 9.0 milestone.

eerhardt:
> The downside of this approach is composability of resources. Let's say I wanted to have a resource that depended on a Postgres database instance. With this change the implementor of that would need to have extensions for each cloud provider (unless they just allow any IResourceWithConnectionString).

We already have a "composability of resources" problem with the current API. You can't call `.ConfigureConstruct` on the Azure resource for Postgres. Even worse - there is no Azure resource for Postgres in the model. So if you wanted to add an event to inspect all the resources (and possibly modify them), you won't see the `AzurePostgresResource` resource in the model.

> That can be solved with an IPostgresResource interface right?

Yes, we can add that, if necessary.

eerhardt:
This PR is ready for review. I believe I have everything ready for this to be merged.

tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs
```diff
@@ -98,7 +99,7 @@ param outputs_azure_container_apps_environment_id string
           }
         }
         """;
-
+        output.WriteLine(bicep);
```
davidfowl, line 102: ?

mitchdenny:
I believe it would be good to add an `IPostgresResource` interface (or add a follow-up for this).

playground/bicep/BicepSample.AppHost/Program.cs
```diff
@@ -37,8 +37,8 @@
 
 var administratorLogin = builder.AddParameter("administratorLogin");
 var administratorLoginPassword = builder.AddParameter("administratorLoginPassword", secret: true);
-var pg = builder.AddPostgres("postgres2", administratorLogin, administratorLoginPassword)
-                .AsAzurePostgresFlexibleServer()
+var pg = builder.AddAzurePostgresFlexibleServer("postgres2")
+                .WithPasswordAuth(administratorLogin, administratorLoginPassword)
```
mitchdenny, line 41: Should this be `WithPasswordAuthentication`?

mitchdenny:
Overall I think this looks good. I do think that there is another approach we could take here (just putting it out there for sake of completeness.

We could do this:

```csharp
var builder = DistributedApplication.Create(args);
var db = builder.AddPostgres("pgsql").AsAzure(out var azpgsql).AddDatabase("db");
azpgsql.DoAzureyApiStuff();
```

Folks have historically avoided out vars but we see them a lot more these days. All that said I think your approach here is fine.

mitchdenny (approved):

tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs
eerhardt: You want this for when the bicep changes and you need to update it. You can copy and paste the new output from the test log

davidfowl (approved):
lgtm

playground/bicep/BicepSample.AppHost/Program.cs
eerhardt: The reason I chose these names (WithPasswordAuth here and WithAccessKeyAuth in Reds) is because they match the APIs from Azure.Provisioning (which I assume just match the names in the rest API).

https://github.com/dotnet/aspire/blob/e3e903ab9f648e7780abb1a5bd351bb7d393b6c2/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs#L192-L196

But looking again at Redis, the name is `DisableAccessKeyAuthentication`, which is inconsistent with the above Postgres APIs. So it is probably better we just completely follow .NET Design Guidelines with our APIs here and use full words.

eerhardt: FYI - @tg-msft. I'm assuming this isn't something we can change/fix in Azure.Provisioning since it is just generated from the service APIs / typespec.

davidfowl: We should do an API review, `WithAccessKey` is not intuitive. We should take the liberty here to add some opinions to this layer of the stack. We can do this in a follow up PR.

davidfowl: As far as redis is concerned its a password.

eerhardt: As far as Azure is concerned, it is an "access key":

image

davidfowl: Specified as the password in the redis connection string.

eerhardt: @philon-msft - is there a reason that the portal / ARM use the term "access key"?

philon-msft: Azure Redis doesn't let you choose your own (like a "password") - instead it generates one for you (more like an "access key"). 
So in general Redis terminology it's a password, but for Azure Redis it's an access key.

tg-msft: Correct, it's [`passwordAuth` on the wire](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2022-12-01/FlexibleServers.json#L744) and we're not looking to deviate from our automated translation of [the existing ARM experience](https://learn.microsoft.com/en-us/dotnet/api/azure.resourcemanager.postgresql.flexibleservers.models.postgresqlflexibleserverauthconfig.passwordauth?view=azure-dotnet#azure-resourcemanager-postgresql-flexibleservers-models-postgresqlflexibleserverauthconfig-passwordauth).  There's no way Azure.Provisioning would scale to all of Azure otherwise.

eerhardt:
> I believe it would be good to add an `IPostgresResource` interface (or add a follow-up for this).

Opened [Add interfaces to PostgreSQL, Redis, and SqlServer Hosting Resources (dotnet/aspire#6055)](https://github.com/dotnet/aspire/issues/6055)

playground/bicep/BicepSample.AppHost/Program.cs
eerhardt: > Azure Redis doesn't let you choose your own (like a "password") - instead it generates one for you (more like an "access key"). So in general Redis terminology it's a password, but for Azure Redis it's an access key.

Does Azure Redis let me add users with their own passwords and ACLs?

philon-msft: Not standard Redis users+passwords, but you can add *Entra* users/apps/ManagedIdentities and assign them to roles for ACL functionality. See [doc](https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-azure-active-directory-for-authentication) for the Entra experience

PR #5505, opened 2024-08-31
[IGNORE][tests] Add run scripts for tests

radical:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5505)

PR #5932, opened 2024-09-25
Add net9.0 support for EFCore components

### LINKED ISSUE #5207: Upgrade EF components to 9.0

EF 9.0 introduced [engine-specific API for SQL Server, Azure SQL and Azure Synapse](https://github.com/dotnet/efcore/pull/34052).  If `UseAzureSynapse` or `UseAzureSql` is called by the user in `AddDbContext`, then `EnrichSqlServerDbContext` will throw, because it calls `UseSqlServer` and now that results in conflicting configuration.

EF 9.0 also has a [number of significant improvements](https://github.com/dotnet/efcore/issues?q=sort%3Aupdated-desc+is%3Aissue+label%3Aarea-aspire+milestone%3A9.0.0+is%3Aclosed+) for migrations scenarios relevant to Aspire.

`EnrichSqlServerDbContext` should call `ConfigureSqlEngine` with `EnableRetryOnFailureByDefault` instead of `UseSqlServer`.

Additionally, it and other EF components should call `ConfigureDbContext` instead of `PatchServiceDescriptor` as that depends on implementation details.

### Additional context

_No response_

sebastienros:
Fixes #5207

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5932)


global.json
```diff
@@ -1,11 +1,11 @@
...
-    "version": "9.0.100-preview.7.24407.12",
+    "version": "9.0.100-rc.1.24452.12",
     "rollForward": "major",
     "allowPrerelease": true
   },
   "tools": {
-    "dotnet": "9.0.100-preview.7.24407.12",
+    "dotnet": "9.0.100-rc.1.24452.12",
     "runtimes": {
       "dotnet/x64": [
```
sebastienros, line 10: I feel like these might need some updates, @joperezr ?

radical: Yes, but we should do it in a separate PR.

AndriySvyryd (approved):
Not blocking, but for futureproofing also change `EnrichNpgsqlDbContext`, `EnrichMySqlDbContext`, `EnrichOracleDatabaseDbContext` and `EnrichCosmosDbContext` to use `ConfigureDbContext`

Consider renaming `AspireAzureEFCoreCosmosExtensions`

eng/Versions.props
```diff
@@ -66,5 +67,25 @@
     <MicrosoftEntityFrameworkCoreSqlServerPackageVersion>8.0.8</MicrosoftEntityFrameworkCoreSqlServerPackageVersion>
     <MicrosoftEntityFrameworkCoreToolsPackageVersion>8.0.8</MicrosoftEntityFrameworkCoreToolsPackageVersion>
     <MicrosoftNETRuntimeWorkloadTestingInternalVersion>9.0.0-preview.5.24272.3</MicrosoftNETRuntimeWorkloadTestingInternalVersion>
+
+    <NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>8.0.4</NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net9.0'">
```
eerhardt, line 74: I don't understand why we need to do this split. Microsoft.EntityFrameworkCore 9.0-rc1 targets net8.0. Why do we need to keep referencing the version 8 packages when targeting `net8.0`? Why can't we just continue to only target `net8.0` and depend on EF v9.0?

sebastienros: I talked about it with @joperezr before starting. efcore 9.0 brings extensions dependencies on 9.0 which we would have to bump (logging, ...), and customers would then need these too, even when targeting 8.0. If we didn't do that users who would like to stay on 9.0 for support reasons would not be able to.

davidfowl:
> Not blocking, but for futureproofing also change EnrichNpgsqlDbContext, EnrichMySqlDbContext, EnrichOracleDatabaseDbContext and EnrichCosmosDbContext to use ConfigureDbContext

We want to use these new APIs as part of these EF upgrade so that we can actually use the new methods and make sure they work in the scenarios they were designed for.

eng/Versions.props
eerhardt: I don't like this situation we are putting people (especially library authors) into.

You can reference the 9.0 extension libraries on net8.0. If we cared about this, we should make EF multi-target and not pull in the 9.0 extensions when on net8.0.

We shouldn't have some teams caring about this, and some teams not.

cc @ericstj @SamMonoRT

davidfowl: Lots of people care about this because it'll mean going from using extensions from the shared framework, to bin deploying all of extensions for .NET 8 apps.

eerhardt: Should we fix EF then? Why does their version `9.0` packages support `net8.0` but lift of MS.Ext references to version `9.0`?

davidfowl: Im surprised it doesn't

AndriySvyryd: Would this work? https://github.com/dotnet/efcore/pull/34769

Wouldn't it cause downgrade warnings when used in conjunction with other 9.x packages?

sebastienros:
@radical Please check the build changes. I updated the `ZipTestArchive` target to handle the multi-target build step which doesn't have `OutDir`. And also solved a concurrency issue when the two tfms are built in parallel and would write the same file. The non-default TFM is added as a prefix on the zip filename.

tests/Directory.Build.targets
```diff
@@ -25,11 +25,15 @@
...
       <TestsArchiveSourceDir Condition="'$(TestsArchiveSourceDir)' == ''">$(OutDir)</TestsArchiveSourceDir>
+      <ZipTestArchiveTfm></ZipTestArchiveTfm>
+      <ZipTestArchiveTfm Condition="'$(TargetFramework)' != '$(DefaultTargetFramework)'">-$(TargetFramework)</ZipTestArchiveTfm>
     </PropertyGroup>
 
-    <MakeDir Directories="$(TestArchiveTestsDir)" />
-    <ZipDirectory SourceDirectory="$(TestsArchiveSourceDir)"
-                  DestinationFile="$([MSBuild]::NormalizePath('$(TestArchiveTestsDir)', '$(MSBuildProjectName).zip'))"
+    <!-- TestsArchiveSourceDir is empty (OutDir) for multi-target build step -->
+    <MakeDir Condition="$(TestsArchiveSourceDir) != ''" Directories="$(TestArchiveTestsDir)" />
```
radical, line 33: If it is empty then how why doesn't `ZipDirectory` fail with empty `SourceDirectory`?

radical: oh, you are skipping it in that case. I'll do a local build, and look at the binlog.

sebastienros: It failed before these changes. Now this is skipped when it is the case, when it's invoking the cross-target build.

radical: You can use `and '$(IsInnerBuild)' == 'true'` in the condition for the target. This will be true when it is doing the inner build - which is one per target framework. And the target will get skipped for the outer build.

tests/Directory.Build.targets
```diff
@@ -25,11 +25,15 @@
...
+    <ZipDirectory Condition="$(TestsArchiveSourceDir) != '' and '$(TargetFramework)' == '$(DefaultTargetFramework)'"
```
radical, line 34: Also, the `'$(TargetFramework)' == '$(DefaultTargetFramework)'` means that we would get only the default zip file for `net8.0`, and not for `net9.0`.

sebastienros: I could use some help for this one, when I ran helix without this it failed because it would not find a trx file matching the net90.zip file (I rephrased the error message).

tests/Directory.Build.targets
sebastienros: Thanks, will try that instead

tests/Directory.Build.targets
radical: Um I can work on that bit.

radical:
This needs the workload testing bits to be re-enabled, which would give us a `9.0 sdk with 8.0 runtime` to allow running the net9.0 tests here on helix.

radical:
The net9.0 tests here will need a 9 sdk to run on helix. We could normally get that from the workload-tests setup, but since that is disabled in the repo right now, and waiting for on https://github.com/dotnet/aspire/pull/5932 - we can't test it right now.

I hope to have the templates PR merged this week. If this needs to merged sooner, then an alternative would be to get the test infra (just msbuild *should* be enough) changes here.

radical:
net9.0 tests are running succesfully:
- `Aspire.Microsoft.EntityFrameworkCore.Cosmos.Tests-net9.0`
- `Aspire.Microsoft.EntityFrameworkCore.SqlServer.Tests-net9.0`
- `Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.Tests-net9.0
- `Aspire.Oracle.EntityFrameworkCore.Tests-net9.0`

sebastienros:
I have tested it works locally. 
- service with net9.0 
- both `AddNpgsqlDbContext` and `EnrichNpgsqlDbContext`
- enabling/disabling metrics from code

sebastienros:
Two test failed. One seems to be a random issue with te TestHost.App playground and the fact that RabbitMQ management was assigned the same port as CatalogDb  web app.

The second is related to this change, but the tests timed out. Increasing it for now since none show up as a problem.

src/Components/Aspire.Microsoft.EntityFrameworkCore.SqlServer/AspireSqlServerEFCoreSqlClientExtensions.cs
```diff
@@ -113,6 +113,42 @@ void ConfigureRetry()
 #pragma warning disable EF1001 // Internal EF Core API usage.
             if (!settings.DisableRetry || settings.CommandTimeout.HasValue)
             {
+#if NET9_0_OR_GREATER
+                builder.CheckDbContextRegistered<TContext>();
```
eerhardt, line 117: Why is this only inside the if NET9 check? In the other EF libraries, it is for both.

src/Components/Aspire.Microsoft.EntityFrameworkCore.SqlServer/AspireSqlServerEFCoreSqlClientExtensions.cs
```diff
@@ -113,6 +113,42 @@ void ConfigureRetry()
...
+
+                builder.Services.ConfigureDbContext<TContext>(optionsBuilder => optionsBuilder.ConfigureSqlEngine(options =>
+                {
+                    var extension = optionsBuilder.Options.FindExtension<SqlServerOptionsExtension>();
```
eerhardt, line 121: Why does this one have all this code duplicated between the 2 TFMs? The other ones are able to share code.

eng/Versions.props
DamianEdwards: Can we bump these to the rc.2 versions? Does that line up with the respective planned release dates?

tests/Aspire.Pomelo.EntityFrameworkCore.MySql.Tests/.runsettings
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <!-- Timeout in ms, 20 minutes -->
```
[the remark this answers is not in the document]
sebastienros: I am trying to figure out if the MySQL tests are failing because of the time it takes only. There aren't any failures in the log but it stops at 15min, which was the timeout. Locally they all run in 5s. Two tests (with different inline data) are taking [1 m 25 s]: HealthCheckReportsExpectedStatus and LoggerFactoryIsUsedByRegisteredClient. They account for all the time on Helix. They take 100ms locally. And this should not be about downloading the image because these delays appear at different wall clock times in the log.

https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-5932-merge-b3281ba7f783453882/Aspire.Pomelo.EntityFrameworkCore.MySql.Tests-net9.0/1/console.0d22ff48.log?helixlogtype=result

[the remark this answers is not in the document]
sebastienros: Note that this run didn't take my change into account : `-- RunConfiguration.TestSessionTimeout=900000`, I expected 1200000 here.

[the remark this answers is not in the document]
sebastienros: Finally, this is not specific to this branch, I took the most recent passing build and this test set took 17m12.

```
[xUnit.net 00:17:12.28]   Finished:    Aspire.Pomelo.EntityFrameworkCore.MySql.Tests&#xD;
```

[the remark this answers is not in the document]
sebastienros: It's random:

image

@radical can you see what I need to add on top of the .runsettings to increase this test timeout? Did something change in this PR related to it, since another successful run on another PR took 17m which is more than the default timeout I am seeing in this PR.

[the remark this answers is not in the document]
radical: The timeout for the helix job has to be set in the helix targets (till we get https://github.com/dotnet/aspire/pull/5505 merged).

tests/helix/send-to-helix-basictests.targets
```diff
@@ -41,9 +41,9 @@
       <!-- runsettings timeout in ms, default to 15 mins -->
       <_DefaultWorkItems TimeoutMs="900000" />
 
-      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Hosting.Elasticsearch.Tests'" TimeoutMs="3600000" />
-      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Hosting.Oracle.Tests'" TimeoutMs="1200000" />
-      <_DefaultWorkItems Condition="'%(FileName)' == 'Aspire.Pomelo.EntityFrameworkCore.MySql.Tests'" TimeoutMs="1200000" />
+      <_DefaultWorkItems Condition="$([System.String]::new('%(FileName)').StartsWith('Aspire.Hosting.Elasticsearch.Tests'))" TimeoutMs="3600000" />
+      <_DefaultWorkItems Condition="$([System.String]::new('%(FileName)').StartsWith('Aspire.Hosting.Oracle.Tests'))" TimeoutMs="1200000" />
+      <_DefaultWorkItems Condition="$([System.String]::new('%(FileName)').StartsWith('Aspire.Pomelo.EntityFrameworkCore.MySql.Tests'))" TimeoutMs="1200000" />
```
sebastienros, line 46: Was about to comment that I had found this line setting a different timeout. So that was configured like so for some time already.

I don't remember if we have looked at the reason why this test is so long.

radical:
One of the workload tests failed with:
```
  Failed to download package 'OpenTelemetry.Instrumentation.Http.1.9.0' from 'https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/45bacae2-5efb-47c8-91e5-8ec20c22b4f8/nuget/v3/flat2/opentelemetry.instrumentation.http/1.9.0/opentelemetry.instrumentation.http.1.9.0.nupkg'.&#xD;
  The HTTP request to 'GET https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/45bacae2-5efb-47c8-91e5-8ec20c22b4f8/nuget/v3/flat2/opentelemetry.instrumentation.http/1.9.0/opentelemetry.instrumentation.http.1.9.0.nupkg' has timed out after 100000ms.&#xD;
```

We can re-try.

sebastienros:
@radical workload test failure? This one was already failing in the previous run btw.

eng/pipelines/templates/BuildAndTest.yml
```diff
@@ -86,6 +86,7 @@ steps:
     - template: /eng/pipelines/templates/send-to-helix.yml
       parameters:
         HelixProjectPath: '$(Build.SourcesDirectory)/tests/helix/send-to-helix-ci.proj'
+        # Code coverage on windows is disabled: <issue>
```
eerhardt, line 89: Was this intentional? Was this supposed to have an issue linked?

eerhardt (approved):
LGTM. Let's get this in.

We should also consider targeting rc2 now that it has shipped.

eng/pipelines/templates/BuildAndTest.yml
sebastienros: We should remove it, I will remove it in the backport PR if that's fine, I don't want to wait another build for that

PR #5936, opened 2024-09-26
Pass thru WaitFor to parent resourcs.

### LINKED ISSUE #5868: Remove health checks from child resources
In the current work we've done for `WaitFor` we have resource checks for child resources. Whilst this is desirable in the long term, for .NET Aspire 9.0 we are going to remove them because we aren't automatically creating databases.

This issue is to go through all the resources in our tree and make the call on whether the child resource should have its health check removed.

mitchdenny:
## Description

This PR removes health checks from child resources (e.g. SQL Server Database). Instead what we do here is that when `WaitFor` is invoked we determine whether we are adding a reference to a resource that has a parent, and then we add a `WaitAnnotation` for both the child resource and the parent resource. This combined with the check removal in the child resources has a few effects.

First, dependents will not block waiting for a database to be created, they'll only wait for the server to be created. After some experimentation with the way people use Aspire this seemed like a more natural experience.

Leaving the `WaitAnnotation` in place of the child resources is important though for future scenarios where we want to enable something like this:

```csharp
var db = builder.AddSqlServer("sql")
                .AddDatabase("db")
                .WithCreation(); // This extension does not exist yet.
```

What the `WithCreation(...)` extension would do in this case is add a health check to the child resource and register a event subscription for the resource healthy event for the parent.

This is just an explanantion as to why we are keeping the `WaitAnnotation` on the child resource ;)

Fixes #5868 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5936)


mitchdenny:
fyi @Alirexaa

davidfowl (approved):

PR #5938, opened 2024-09-26
Fail to start when health check registration is missing for resource.

### LINKED ISSUE #5910: AppHost should crash if health check annotations reference a health check which does not exist.
When the AppHost starts up, if a health check is not registered in DI, but is in a health check annotation then the AppHost should crash.

mitchdenny:
## Description

This PR adds an options validator which checks to ensure that all health checks registrations referenced by resources in `HealthCheckAnnotation` instances are present. If they are not then we log a critical warning and throw blocking startup.

Fixes #5910 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5938)

### DIFF
--- a/src/Aspire.Hosting/DistributedApplicationBuilder.cs
+++ b/src/Aspire.Hosting/DistributedApplicationBuilder.cs
@@ -278,6 +278,35 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
 
     private void ConfigureHealthChecks()
     {
+        _innerBuilder.Services.AddSingleton<IValidateOptions<HealthCheckServiceOptions>>(sp =>
+        {
+            var appModel = sp.GetRequiredService<DistributedApplicationModel>();
+            var logger = sp.GetRequiredService<ILogger<DistributedApplicationBuilder>>();
+
+            // Generic message (we update it in the callback to make it more specific).
+            var failureMessage = "A health check registration is missing. Check logs for more details.";
+
+            return new ValidateOptions<HealthCheckServiceOptions>(null, (options) =>
+            {
+                var resourceHealthChecks = appModel.Resources.SelectMany(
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -278,6 +278,35 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
 
     private void ConfigureHealthChecks()
     {
+        _innerBuilder.Services.AddSingleton<IValidateOptions<HealthCheckServiceOptions>>(sp =>
```
davidfowl, line 281: Do we log from the other configure options code?

mitchdenny: No ... we don't log in other places. However the reason I did it here is that I felt that if we just gave a generic error that a health check was missing it wouldn't be very helpful to the developer. By outputting to the logs we make it easier to find.

Happy to remove this though because this check is pretty aggressive and should show up for folks as soon as they make the mistake.

drewnoakes (approved):
LGTM. I assume the intent is not to include details in an exception message in case that message leaks during a failure? Is that a concern here?

mitchdenny:
It's more that you can't. You can only give a static failure message. But I do write out the name of the check that is missing as a critical log entry.

PR #5944, opened 2024-09-26
Copy parent health checks into child to drive sync status.

### LINKED ISSUE #5943: Synchronise health state between parent and child resources

mitchdenny:
## Description

This is one possible solution for #5943. When using `WaitFor` on a resource that has a parent we'll automatically copy the health check annotations from the parent into the child. It works well because the same loop that sets the health status on the parent, sets it on the child. The copy occurs just before the app host starts so we can pick up any health checks that may have been manually added after the `WaitFor` is invoked.

Fixes #5943

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5944)

### DIFF
--- a/src/Aspire.Hosting/ResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ResourceBuilderExtensions.cs
@@ -612,9 +612,19 @@ public static class ResourceBuilderExtensions
             // the child resource itself does not have any health checks setup.
             var parentBuilder = builder.ApplicationBuilder.CreateResourceBuilder(dependencyResourceWithParent.Parent);
             builder.WaitFor(parentBuilder);
+
+            var parentHealthCheckAnnotations = parentBuilder.Resource.Annotations.OfType<HealthCheckAnnotation>().Select(hca => hca.Key);
+            var resourceHealthCheckAnnotations = builder.Resource.Annotations.OfType<HealthCheckAnnotation>().Select(hca => hca.Key);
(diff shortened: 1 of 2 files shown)


davidfowl:
I'm not sure I like this. How about in the Publisher, we just propagate the health status to child resources?

mitchdenny:
You know I think you are right. There is another issue with this - if you don't use `WaitFor` on a child resource won't replicate its parents` status.

PR #5951, opened 2024-09-26
Allow WithReference for AWS CDK Stack resources

vlesierse:
cc @normj

## Description

AWS CDK construct resources are only supported using `WithReference` & `WithEnvironment` on Project resources. A common scenario would be that you are adding an existing AWS CDK stack to Aspire and that you would like to reference exposed constructs and values from that stack to projects in Aspire. This fix changes the contract from `IResourceBuilder<IConstructResource<TConstruct>>` to `IResourceBuilder<IResourceWithConstruct<TConstruct>>` which also applies to `IStackResource`

```csharp

var stack = builder.AddAWSCDKStack("custom", scope => new CustomStack(scope, "Aspire-custom"));
builder.AddProject<Projects.Frontend>("frontend")
    .WithReference(stack, s => s.Bucket.BucketName, "BucketName")

builder.Build().Run();
```

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x]  No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5951)


normj (approved):
So this adds the use case where before I could add references with delegates to individual constructs like a topic but now I can also add the reference with the delegate referencing the top level stack and then in the delegate navigate to what I want to reference.

vlesierse:
> So this adds the use case where before I could add add references with delegates to individual constructs like a topic but now I can also add the reference with the delegate referencing the top level stack and then in the delegate navigate to what I want to reference.

Yes, it makes using stacks referenced from other projects and referencing the outputs a lot easier.

davidfowl (approved):

davidfowl:
Tks!

PR #5976, opened 2024-09-27
Fix filter encoding with colon and plus characters

JamesNK:
## Description

I noticed that log and trace filters with serialization control characters (`+`, `:`) were broken.

Blazor's `[SupplyParameterFromQuery]` feature unescapes the value before giving it to us, which causes deserialization to break. For example:

* Query string value: `name:condition:before%3Aafter`
* Parameter value: `name:condition:before:after`

If you have a filter with one of these characters in it and you refresh the browser, the filter disappears because it expects 3 colons.

This PR:

* Double encodes and decodes these special characters
* Adds tests

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5976)

### DIFF
--- a/src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
+++ b/src/Aspire.Dashboard/Components/Dialogs/FilterDialog.razor.cs
@@ -68,11 +68,11 @@ public partial class FilterDialog
             _parameters = knownFields;
         }
 
-        if (Content.Filter is { } logFilter)
+        if (Content.Filter is { } filter)
         {
-            _formModel.Parameter = _parameters.SingleOrDefault(c => c.Id == logFilter.Field);
-            _formModel.Condition = _filterConditions.Single(c => c.Id == logFilter.Condition);
-            _formModel.Value = logFilter.Value;
+            _formModel.Parameter = _parameters.SingleOrDefault(c => c.Id == filter.Field);
+            _formModel.Condition = _filterConditions.Single(c => c.Id == filter.Condition);
+            _formModel.Value = filter.Value;
         }
         else
         {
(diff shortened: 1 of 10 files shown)


davidfowl:
cc @javiercn  @SteveSandersonMS @MackinnonBuck

SteveSandersonMS:
@JamesNK Yes, I think it will be necessary to do something like this. Any value that you put in a querystring needs to be encoded for querystrings.

To be honest I would look at `DashboardUrls.StructuredLogsUrl` and try to make it a bit less coupled to particular expectations for `filters` being pre-encoded. If this took care of encoding as part of constructing the URI, then the caller wouldn't need to know about this. In theory as long as the encoding happens at the point where you build the URI, then whenever you code later receives the querystring value, it will decode to whatever was supplied to this method, so you don't need to hardcode rules about what gets encoded at other places.

JamesNK:
The reason we didn't encode the whole filter query string when adding it to the URL is it becomes a bit of a dissater to read:

Before:

> filters=log.message:contains:Application%20started.%20Press%20Ctrl%252BC%20to%20shut%20down.%2Blog.message:contains:Hosting%20environment%253A%20Development

After:

> filters=log.message%3Acontains%3AApplication%2520started.%2520Press%2520Ctrl%252BC%2520to%2520shut%2520down.%20log.message%3Acontains%3AHosting%2520environment%253A%2520Development

Although I guess once values require a lot of escaping neither are really readable.

JamesNK:
Updated. I think this is better and safer, although URLs don't look quite as good. Given the trouble that special characters have caused in this query string value, it's better to just be safe.

Could improve in the future by using an encoder which doesn't escape `:` and `+` characters, but we'd need to copy `AddQueryString` out of Blazor source. Not important.

mitchdenny (approved):
LGTM. More of a side comment renames from *LogFilter to Filter (non blocking)?

JamesNK:
The rename of "Log filter" -> "Filter" is because filters are used on structured logs and traces pages now. I didn't feel like creating a new PR just to rename stuff 😄

PR #5986, opened 2024-09-27
Clean up how we write parameters to the manifest

davidfowl:
- Don't use the parameters on the resource, instead write them directly to the construct. Since we're generating them and referencing them directly.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5986)

### DIFF
--- a/playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
+++ b/playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
@@ -37,8 +37,8 @@ resource api 'Microsoft.App/containerApps@2024-03-01' = {
         }
         {
           name: 'connectionstrings--account'
-          value: account_secretoutputs_connectionstring
           identity: outputs_azure_container_registry_managed_identity_id
+          keyVaultUrl: account_secretoutputs_connectionstring
         }
         {
           name: 'value'
--- a/src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
+++ b/src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
@@ -2,6 +2,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics;
 using System.Globalization;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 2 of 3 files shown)


tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs
```diff
@@ -520,6 +520,8 @@ public async Task PublishAsContainerAppInfluencesContainerAppDefinition()
         builder.AddContainer("api", "myimage")
             .PublishAsAzureContainerApp((module, c) =>
             {
+                Assert.Contains(c, module.GetResources());
```
eerhardt, line 523: This wasn't true before?

playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
```diff
@@ -37,8 +37,8 @@ resource api 'Microsoft.App/containerApps@2024-03-01' = {
         }
         {
           name: 'connectionstrings--account'
-          value: account_secretoutputs_connectionstring
           identity: outputs_azure_container_registry_managed_identity_id
+          keyVaultUrl: account_secretoutputs_connectionstring
```
eerhardt, line 41: Was this already out of date? I don't see how this could have changed with your changes.

src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
```diff
@@ -127,6 +128,26 @@ record struct EndpointMapping(string Scheme, string Host, int Port, int? TargetP
...
 
+            public void WriteToManifest(ManifestPublishingContext context, AzureConstructResource construct)
+            {
+                construct.WriteToManifest(context);
+
+                // We're handling custom resource writing here instead of in the AzureConstructResource
+                // this is because we're tracking the BicepParameter instances as we process the resource
+                if (Parameters.Count > 0)
+                {
+                    context.Writer.WriteStartObject("params");
```
eerhardt, line 139: Could we end up in a situation where `AzureConstructResource` wrote the `params` object, and then now we are writing another one to the same object?

davidfowl: That would be a bug in our code. Code outside never gets access to the `AzureConstructResource` so its not a concern.

tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs
davidfowl: Yes. Thats' why I am asserting it 😄

playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
davidfowl: Yes

src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
davidfowl: Done

eerhardt (approved):

PR #5987, opened 2024-09-27
Update ready state check for Azure Functions tests

captainsafia:
Try checking for the host started message to avoid races between the HTTP client request sent from the API service and when the host is actually ready to accept requests.

cc: @radical 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5987)


tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
```diff
@@ -68,11 +68,11 @@ public async Task AzureFunctionsTest()
         await app.StartAsync();
         await app.WaitForResources().WaitAsync(TimeSpan.FromMinutes(2));
 
-        // Wait for the blobTrigger to be discovered as an indication that the host and worker
-        // has been successfully launched
+        // Wait for the 'Job host started' message as an indication
```
eerhardt, line 71: Do you want to remove the ActiveIssue attribute?

davidfowl:
I think you need to make sure the http client is retying failed requests.

tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
captainsafia: Done.

captainsafia:
> I think you need to make sure the http client is retying failed requests.

Didn't realize an HTTP client configured with resilience was already set up in the test infrastructure. 😅 

Updated the test to use it.

captainsafia:
@radical Does this look good to you?

tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
```diff
@@ -68,26 +67,27 @@ public async Task AzureFunctionsTest()
...
+        // that the Functions host has initialized correctly
         await WaitForAllTextAsync(app,
             [
-                "MyAzureBlobTrigger: blobTrigger"
+                "Job host started"
```
radical, line 74: Should we wait for both? Though I think this combined with the http resilience will be enough.

radical (approved):
LGTM! Thank you!

tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
captainsafia: IMO, checking for `"MyAzureBlobTrigger: blobTrigger"` doesn't have as much value as running the actual blob trigger scenario so that's why I opted to change the check that we do here.

PR #5999, opened 2024-09-27
Add metrics and improve telemetry support in Azure.AI.OpenAI

### LINKED ISSUE #5451: OpenAI component doesn't setup telemetry correctly
I have been fiddling with the playground sample for the playground OpenAI component trying to get it to emit telemetry. I think the following things are missing;

- Updating Directory.Packages.props to use beta3 for the Azure.AI.OpenAI package

``` xml
<PackageVersion Include="Azure.AI.OpenAI" Version="2.0.0-beta.3" />
```

I also had to add a source to nuget as that version is pulling in a more recent OpenAI SDK version than is available internally at the moment

- In the Aspire.Azure.AI.OpenAI component:
  -  Update the activity source names to include OpenAI.*

``` csharp
protected virtual string[] ActivitySourceNames => new[] { $"{typeof(TClient).Namespace}.*", "OpenAI.*" }; 
```

  - Set the app context switch for the OpenAI instrumentation

``` csharp
        if (GetTracingEnabled(settings))
        {
            AppContext.SetSwitch("OpenAI.Experimental.EnableOpenTelemetry", true);
            builder.Services.AddOpenTelemetry()
                .WithTracing(traceBuilder => traceBuilder.AddSource(ActivitySourceNames));

        }
```

When these are all in place, we get the metrics and telemetry that @lmolkova added to OpenAI recently. 

image

I believe we should include the app context switch - its there because the semanic conventions are not stable, but IMHO for Aspire we should be showing what's available rather than hiding it.

sebastienros:
## Description

- Enables telemetry app switch when metrics or tracing is configured.
- Adds support for metrics in Azure components.
- Enables metrics in Azure.AI.OpenAI.

Covers #5451

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5999)


src/Components/Aspire.OpenAI/AspireOpenAIExtensions.cs
```diff
@@ -107,12 +107,16 @@ private static void AddOpenAIClient(
 
         if (!settings.DisableTracing)
         {
+            AppContext.SetSwitch("OpenAI.Experimental.EnableOpenTelemetry", true);
```
eerhardt, line 110: I think we should be following the same pattern as the other experimental telemetry in the Azure SDK - 

https://github.com/dotnet/aspire/blob/25bbf0543a5915d2ff4b5055431bce176bf1c650/src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs#L64-L80

cc @lmolkova

sebastienros: Applied this pattern here and also did that for OpenAI from which I copied it initially.
Note that it doesn't make it automatic then and users still have to nabe the flag even though we advertise telemetry is on by default. But I assume it's on purpose because it's experimental. Not sure the issue that is linked should be marked as fixed then after that.

eerhardt: > and users still have to nabe the flag even though we advertise telemetry is on by default. But I assume it's on purpose because it's experimental.

Correct. It is experimental, so we should let the users decide when they want to enable it (and become dependent on an experimental feature).

tests/Aspire.Azure.AI.OpenAI.Tests/ConformanceTests.cs
```diff
@@ -80,24 +81,28 @@ void ConfigureCredentials(AzureOpenAISettings settings)
...
         => options.DisableTracing = !enabled;
 
     protected override void TriggerActivity(AzureOpenAIClient service)
         => service.GetChatClient("dummy").CompleteChat("dummy gpt");
+
+    private static RemoteInvokeOptions EnableTelemetry()
+    => new()
+    {
+        RuntimeConfigurationOptions = { { "OpenAI.Experimental.EnableOpenTelemetry", true } }
+    };
```
eerhardt, line 107: (nit) the whitespacing/indent here could be cleaned up.

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
```diff
@@ -39,13 +42,56 @@ public sealed class AzureOpenAISettings : IConnectionStringSettings
...
+    /// <summary>
+    /// Gets or sets a boolean value that indicates whether the OpenTelemetry metrics are enabled or not.
+    /// </summary>
+    /// <remarks>
+    /// Azure AI OpenAI telemetry support is experimental, the shape of traces may change in the future without notice.
+    /// It can be enabled by setting "OpenAI.Experimental.EnableOpenTelemetry" <see cref="AppContext"/> switch to true.
+    /// Or by setting "OPENAI_EXPERIMENTAL_ENABLE_OPEN_TELEMETRY" environment variable to "true".
+    /// </remarks>
+    /// <value>  
+    /// The default value is <see langword="false"/>.
```
eerhardt, line 54: It's not necessarily `false` - it defaults to the value of the env var / appcontext switch. I think we could just leave this `value` section off until the env var goes away.

eerhardt (approved):
LGTM. Thanks!

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
sebastienros: applied

sebastienros:
Oops, I think I should not check auto-merge unless it's only waiting for a build to pass, and not for also an approval. Following up with a PR to fix the doc.

PR #6000, opened 2024-09-27
Fix roles on default Storage resource  and endpoint mappings for Functions

captainsafia:
This pull request resolves two issues with Functions apps on Aspire to ACA.

The. first commit is because blob triggers require `StorageAccountContributor` access on the host storage account at startup in order to properly initialize the blob trigger listener. This role assignment isn't configured by default on the Azure Storage resource so we add it to the default resource when under publish mode.

If users are configuring their own host storage, they'll need to configure the role assignments themselves in the construction callback.

We could be more conservative here and only apply this logic if we detect a blob resource is configured in the environment, but applying this behavior by default in all cases seems the least likely to cause spookiness.

The second commit configures the container image used by Azure Functions to listen on and expose the non-priviliged port :8080 by default to avoid permissions issues when initializing the Azure Functions host. The [Functions image used](https://github.com/Azure/azure-functions-docker/blob/dev/host/4/bookworm/dotnet-isolated/dotnet8-isolated/dotnet8-isolated.Dockerfile#L26-L29) extends the default .NET images but overrides the port used for back-compat reasons. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6000)

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -35,12 +35,15 @@ public static class AzureFunctionsProjectResourceExtensions
             // account when deployed. We assign this role to the host storage resource when running in publish mode.
             if (builder.ExecutionContext.IsPublishMode)
             {
-#pragma warning disable AZPROVISION001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
-                storage = builder.AddAzureStorage(DefaultAzureFunctionsHostStorageName, (builder, construct, storageAccount) =>
+                var configureConstruct = (ResourceModuleConstruct construct) =>
                 {
+#pragma warning disable AZPROVISION001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+                    var storageAccount = construct.GetResources().OfType<StorageAccount>().FirstOrDefault(r => r.ResourceName == DefaultAzureFunctionsHostStorageName)
+                        ?? throw new InvalidOperationException($"Could not find storage account with '{DefaultAzureFunctionsHostStorageName}' name.");
                     construct.Add(storageAccount.AssignRole(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
-                }).RunAsEmulator().Resource;
 #pragma warning restore AZPROVISION001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+                };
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -65,6 +79,13 @@ public static class AzureFunctionsProjectResourceExtensions
                 context.EnvironmentVariables["OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY"] = "in_memory";
                 context.EnvironmentVariables["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true";
                 context.EnvironmentVariables["FUNCTIONS_WORKER_RUNTIME"] = "dotnet-isolated";
+                // Set ASPNETCORE_URLS to use the non-privileged port 8080 when running in publish mode.
+                // We can't use the newer ASPNETCORE_HTTP_PORTS environment variables here since the Azure
+                // Functions host is still initialized using the classic WebHostBuilder.
+                if (context.ExecutionContext.IsPublishMode)
+                {
+                    context.EnvironmentVariables["ASPNETCORE_URLS"] = "http://+:8080";
+                }
```
davidfowl, line 88: You don't want to hardcode 8080. You want to use the target port.

```suggestion
                if (context.ExecutionContext.IsPublishMode)
                {
                   var endpoint = resource.GetEndpoint("http");
                    context.EnvironmentVariables["ASPNETCORE_URLS"] = ReferenceExpression.Create($"http://+:{endpoint.Property(EndpointProperty.TargetPort)}");
                }
```

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -65,6 +79,14 @@ public static class AzureFunctionsProjectResourceExtensions
                 context.EnvironmentVariables["OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY"] = "in_memory";
                 context.EnvironmentVariables["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true";
                 context.EnvironmentVariables["FUNCTIONS_WORKER_RUNTIME"] = "dotnet-isolated";
+                // Set ASPNETCORE_URLS to use the non-privileged port 8080 when running in publish mode.
+                // We can't use the newer ASPNETCORE_HTTP_PORTS environment variables here since the Azure
+                // Functions host is still initialized using the classic WebHostBuilder.
+                if (context.ExecutionContext.IsPublishMode)
+                {
+                   var endpoint = resource.GetEndpoint("http");
```
davidfowl, line 87:
```suggestion
                    var endpoint = resource.GetEndpoint("http");
```

davidfowl:
I think this is a good place to start. When we are threat modeling this, we need to raise the added role assignment on the host storage.

davidfowl (approved):

PR #6005, opened 2024-09-28
Fix how we handle secret outputs in the aca code gen logic

davidfowl:
- Originally we were leaning on azd to pass in the either a url or a secret value but that would require azd to understand the context in which a secret output was used or would require us to invent new syntax for how to represent secret values vs secret uris. With this new approach, we ask for the keyvault name and resolve the secret uri in the bicep directly. This means that we can use {resource.secretOutputs.name} as the secret value instead of the url.
- Bonus, we should be able to remove the extra logic from azd that attempts to resolve the key vault secret and delegate that to ARM.
- Added tests

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6005)


src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
```diff
@@ -698,9 +712,36 @@ BicepValue<string> GetHostValue(string? prefix = null, string? suffix = null)
...
+                    // Now we resolve the secret
+                    secret = KeyVaultSecret.FromExisting($"{kv.ResourceName}_{secretOutputReference.Name}");
+                    secret.Name = secretOutputReference.Name;
+                    secret.Parent = kv;
+
+                    KeyVaultSecretRefs[secretOutputReference.ValueExpression] = secret;
+                }
+
+                // REVIEW: Is there a better way to do this?
+                return new MemberExpression(
```
davidfowl, line 741: cc @tg-msft @eerhardt

tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs
```diff
@@ -307,10 +307,10 @@ public async Task ProjectWithManyReferenceTypes()
           "path": "api.module.bicep",
           "params": {
             "api_containerport": "{api.containerPort}",
-            "mydb_secretoutputs_connectionstring": "{mydb.secretOutputs.connectionString}",
+            "mydb_secretoutputs": "{mydb.secretOutputs}",
```
mitchdenny, line 310: So is the assumption here if we have a placeholder which ends in ...secretOutputs that we are referring to the keyvault name?

davidfowl: Not assumption, that's the behavior we spec for azure bicep resources. 

```
{resource.secretOutputs} = secret output keyvault name
{resource.secretOutputs.{name}} - secret value
```

We have some existing conventions that are similar from the GA release (but aren't as clean) e.g. if a resource has a paramter called "keyVaultName" and an empty string value, we will pass the secret output keyvault name for that resource. In the future, that would just be:

```JSON
{
    "params":  {
        "keyVaultName": "{resource.secretOutputs}"
     }
}
```

That's the more explicit way to define without assuming anything about the names of parameters.

playground/AzureContainerApps/AzureContainerApps.AppHost/postgres.module.bicep
```diff
@@ -0,0 +1,63 @@
+@description('The location for the resource(s) to be deployed.')
```
eerhardt, line 1: Where's this file come from?

davidfowl: Not my codez

src/Aspire.Hosting.Azure.ContainerApps/AzureContanierAppsInfrastructure.cs
eerhardt: Since this property is nested under "Properties", there isn't a better way to do this right now. But @tg-msft will fix it so this can be `return secret.SecretUri;`.

Instead of `REVIEW` can you change this to `TODO`? That's how I've been tracking all the things that should be fixed when the CDK 1.0 is out.

playground/AzureContainerApps/AzureContainerApps.AppHost/postgres.module.bicep
eerhardt: You added the file, and this playground app. I don't see anything in the app using postgres. I think this is a left over file that should be deleted.

playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
```diff
@@ -3,10 +3,10 @@ param location string = resourceGroup().location
 
 param api_containerport string
 
+@secure()
 param storage_outputs_blobendpoint string
```
eerhardt, line 7: Why's this parameter now considered "secure"? It is just a URI, right?

eerhardt (approved):

playground/AzureContainerApps/AzureContainerApps.AppHost/api.module.bicep
davidfowl: We're currently marking everything that is a connection string as secure. We can be a bit more specialized but that can be a follow up.

playground/AzureContainerApps/AzureContainerApps.AppHost/postgres.module.bicep
davidfowl: Oh I might have undone that change.

PR #6010, opened 2024-09-29
Update daily installation instructions to account for workload removal.

afscrome:
## Description

- Removed the `installLatestfromReleaseBranch` scripts that dealt with workload installation
- Updated installation instructions to no longer reference workload installation

Probably also want to update the `Visual Studio 2022 version 17.9 Preview` reference - either to non preview 17.9, or to a newer VS version.  

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6010)


docs/using-latest-daily.md
```diff
@@ -6,33 +6,24 @@ These instructions will get you set up with the latest build of Aspire. If you j
...
 
+If you use package source mapping, you'll also need to add the following mappings to your NuGet.config
+
+```xml
+<packageSourceMapping>
+  <packageSource key="dotnet8">
+    <package pattern="Aspire.*" />
+    <package pattern="Microsoft.Extensions.*" />
+  </packageSource>
+</packageSourceMapping>
```
gabynevada, line 24: Could you consider adding the entire NuGet.config example?

I did not have one and had to look at the .NET 9 repo for how it was configured, I finished with something like this at the end.

```suggestion
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
    <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
  <packageSourceMapping>
    <packageSource key="dotnet8">
      <package pattern="Aspire.*" />
      <package pattern="Microsoft.Extensions.*" />
    </packageSource>
  </packageSourceMapping>
</configuration>

```

afscrome: The `dotnet nuget add source` should have added the appropriate Package Source to your nuget.config file, although perhaps it did it in your user profile's file (E.g. `%appdata%\NuGet\nuget.config`)

The Package Source Mappings are only needed if you use package source mappings - if you didn't already have a nuget.config file, then you probably don't need the mappings.  If you do have mappings, then you'll need to carefully merge the mappings into your existing mappings.  Because of this, I'm hesitant to provide a complete `nuget.config`.  If I used your suggested nuget.config, I would be in violation of my company's internal policies which prohibit the use of nuget.org, instead requiring us to go through an internal mirror which is configured in a higher level nuget.config.

I did add a link to Package Source Mappings however to make it easier to find out what they are and how to configure them.

afscrome:
I also think there's a bit missing from here to get the latest templates - the above instructions create an app host w/out the required `<Sdk />` element which is in the templates - https://github.com/dotnet/aspire/blob/main/src/Aspire.ProjectTemplates/templates/aspire-apphost/Aspire.AppHost1.csproj

docs/using-latest-daily.md
gabynevada: Understood, it was probably an issue on my side

docs/using-latest-daily.md
```diff
@@ -6,33 +6,24 @@ These instructions will get you set up with the latest build of Aspire. If you j
...
+```
+
```
davidfowl, line 26: Where's the command to install the templates?

joperezr:
Thanks so much for opening this @afscrome, this was on my to-do list for this week but it is great that you helped and contributed with this 😃 

I apologize for the delay in reviewing but was swamped today and didn't get a chance to do it yet, but it is next on my to-do list so I'll be able to review very soon. In case I find things in review, and of course assuming you are okay with it, I can probably just go ahead and push changes directly into this PR. Let me know if you are okay with that, and thanks again for putting this up.

davidfowl:
Here's the command @afscrome 
```
dotnet new install Aspire.ProjectTemplates::9.0.0-preview.* --force
```

We'll need to update this when the version changes.

docs/using-latest-daily.md
DamianEdwards: I think we should recommend creating a local nuget.config file as updating your global feeds to include these CI feeds is generally not a great idea.

afscrome:
@joperezr  Please go ahead and make any changes you wish to my branch - otherwise I may not get a chance to work on this again until the weekend.

joperezr:
Pushed a few changes to add a couple of sections per feedback. Let me know if that looks good to you @afscrome

afscrome:
Looks good to me @joperezr .  Does the reference to `Visual Studio 2022 version 17.9 Preview` be updated to a newer version?  (I'd imagine it should at least be 17.9 non preview at this point.)

joperezr:
Good call, I'll update it to 17.10 as that is the minimum required version.

joperezr (approved):

joperezr:
Thank you so much for the contribution @afscrome!

Issue #6007, opened 2024-09-29
[WebToolsE2E][Aspire] Select any resource on the Console page of the dashboard and click Refresh the browser. The page is blank.


v-reinawang:
**REGRESSION INFO**: Worked in Aspire 8.2.1

**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2 ENU
2. Install SDK 8.0.402
3. Install Aspire 9.0 RC1 [9.0.0-preview.4.24478.1]

**REPRO STEPS**

1. Open CMD, run following command to create an Aspire Starter App project
```
dotnet new aspire-starter -o AspireStarterApp
cd AspireStarterApp/AspireStarterApp.AppHost
```
2. Dotnet run the project in CMD, and open URL in the browser.
`dotnet run`
3. Go to Console page, choose any resource and refresh the browser

**Note**: This issue only repro on Console page, not repro on other pages.

**ACTUAL**
Click refresh once or twice, the page is blank without any content.
![dashboard](https://github.com/user-attachments/assets/6995615f-b793-403a-b6c1-ac8297d04739)

**EXPECTED**
Refresh the browser and the page will display normally.
![image](https://github.com/user-attachments/assets/38628273-c951-47a4-981e-ba2321f36ad6)

v-reinawang:
Verify this issue has been fixed in Aspire 9.0.0-rc.1.24506.1.

PR #6018, opened 2024-09-30
Add a maximum log batch size to avoid exceeding max message size limit

JamesNK:
## Description

While trying to repo https://github.com/dotnet/aspire/issues/6007 I managed to generate enough logs to exceed the max receive size limit (4MB). I've had a feeling we'd see this one day with initial snapshots of existing data, but this is the first time I've seen it.

This PR forces console logs into batches with a max size of 500. It doesn't guarentee we don't exceed the size because we don't limit log line size, but it should prevent most cases of big data.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6018)

### DIFF
--- a/src/Aspire.Hosting/Dashboard/DashboardService.cs
+++ b/src/Aspire.Hosting/Dashboard/DashboardService.cs
@@ -21,6 +21,9 @@ namespace Aspire.Hosting.Dashboard;
 internal sealed partial class DashboardService(DashboardServiceData serviceData, IHostEnvironment hostEnvironment, IHostApplicationLifetime hostApplicationLifetime, ILogger<DashboardService> logger)
     : Aspire.ResourceService.Proto.V1.DashboardService.DashboardServiceBase
 {
+    // gRPC has a maximum receive size. Force logs into batches to avoid exceeding receive size.
+    public const int LogMaxBatchSize = 200;
+
     // Calls that consume or produce streams must create a linked cancellation token
     // with IHostApplicationLifetime.ApplicationStopping to ensure eager cancellation
     // of pending connections during shutdown.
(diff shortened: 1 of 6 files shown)


src/Aspire.Hosting/Dashboard/DashboardService.cs
```diff
@@ -21,6 +21,9 @@ namespace Aspire.Hosting.Dashboard;
 internal sealed partial class DashboardService(DashboardServiceData serviceData, IHostEnvironment hostEnvironment, IHostApplicationLifetime hostApplicationLifetime, ILogger<DashboardService> logger)
     : Aspire.ResourceService.Proto.V1.DashboardService.DashboardServiceBase
 {
+    // gRPC has a maximum receive size. Force logs into batches to avoid exceeding receive size.
+    public const int LogMaxBatchSize = 200;
```
JamesNK, line 25:
```suggestion
    public const int LogMaxBatchSize = 500;
```

mitchdenny (approved):
LGTM. Do we need to worry about excessively long individual log lines?

JamesNK:
> This PR forces console logs into batches with a max size of 500. It doesn't guarentee we don't exceed the size because we don't limit log line size, but it should prevent most cases of big data.

PR #6036, opened 2024-09-30
Only make things secrets that contain secrets

davidfowl:
- Only make the environment variable a secret if it has secret parameters

PS: Not required, but we can detect when there's a "secret" in the value and we can only make them secret if that's the case.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6036)


mitchdenny (approved):
LGTM

PR #6037, opened 2024-09-30
Remove default values from AzureOpenAIDeployment ctor

### LINKED ISSUE #5375: Remove primary constructor from AzureOpenAIDeployment
The `AzureOpenAIDeployment` has a primary constructor with default values in it.
We should get rid of these optional arguments in the constructor as this change the public API when we need to update the defaults.

Current API:
```c#
public class AzureOpenAIDeployment(string name, string modelName, string modelVersion, string skuName = "Standard", int skuCapacity = 8)
```

Options are:

1- Use nullable arguments, define default on properties
```c#
public class AzureOpenAIDeployment(string name, string modelName, string modelVersion, string? skuName, int? skuCapacity)
```

2- Remove optional arguments from the ctor
```c#
public class AzureOpenAIDeployment(string name, string modelName, string modelVersion)
```

3- Remove all properties from the ctor, like we do for settings classes
```c#
public class AzureOpenAIDeployment
{
   ...
}
```

sebastienros:
## Description

Remove default values from the public API (the constructor) so we don't update the API surface when changing defaults.

image

Fixes #5375

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6037)

### DIFF
--- a/src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIDeployment.cs
+++ b/src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIDeployment.cs
@@ -9,10 +9,14 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <param name="name">The name of the deployment</param>
 /// <param name="modelName">The name of the model.</param>
 /// <param name="modelVersion">The version of the model.</param>
-/// <param name="skuName">The name of the SKU.</param>
-/// <param name="skuCapacity">The capacity of the SKU.</param>
-public class AzureOpenAIDeployment(string name, string modelName, string modelVersion, string skuName = "Standard", int skuCapacity = 8)
+public class AzureOpenAIDeployment(string name, string modelName, string modelVersion)
 {
+    /// <value>"Standard"</value>>
+    private const string DefaultSkuName = "Standard";
+
+    /// <value>8</value>
+    private const int DefaultSkuCapacity = 8;
+
     /// <summary>
     /// Gets the name of the deployment.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIDeployment.cs
```diff
@@ -31,10 +35,16 @@ public class AzureOpenAIDeployment(string name, string modelName, string modelVe
...
+    /// The default value is <inheritdoc cref="DefaultSkuName"/>.
+    /// </value>
+    public string SkuName { get; set; } = DefaultSkuName;
 
     /// <summary>
     /// Gets the capacity of the SKU.
     /// </summary>
-    public int SkuCapacity { get; } = skuCapacity;
+    /// <value>
+    /// The default value is <inheritdoc cref="DefaultSkuCapacity"/>.
```
sebastienros, line 47: This is maybe too much here, if anyone thinks so that I will remove the constants and just put the values in the property and the comment directly.

davidfowl:
This is a breaking change right?

sebastienros:
@davidfowl correct, but we almost did one on the same thing in [here](https://github.com/dotnet/aspire/pull/5374). And this is what this PR is for, prevent having the make breaking changes when we just want to change a default value.

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIDeployment.cs
```diff
@@ -9,10 +9,14 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <param name="name">The name of the deployment</param>
 /// <param name="modelName">The name of the model.</param>
 /// <param name="modelVersion">The version of the model.</param>
-/// <param name="skuName">The name of the SKU.</param>
-/// <param name="skuCapacity">The capacity of the SKU.</param>
-public class AzureOpenAIDeployment(string name, string modelName, string modelVersion, string skuName = "Standard", int skuCapacity = 8)
+public class AzureOpenAIDeployment(string name, string modelName, string modelVersion)
```
eerhardt, line 12: Would it be "less breaking" if we did:

```suggestion
public class AzureOpenAIDeployment(string name, string modelName, string modelVersion, string? skuName = null, int? skuCapacity = null)
```

eerhardt (approved):
LGTM.

@davidfowl @mitchdenny - any objections here? Yes this is an API breaking change, but in a very niche area.

davidfowl:
cc @aaronpowell

aaronpowell:
I was just discussing default/optional arguments with Mitch this morning 🤣

PR #6049, opened 2024-10-01
Improve the handling of Container to Container references

### LINKED ISSUE #3735: Support replacing the container host within ReferenceExpression

davidebbo:
## Description

Replace the `ReplaceLocalhostWithContainerHost` hack with proper expression handling that covers all cases of (Exe, Host) x (Exe, Host) referencing.

Fixes #3735

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6049)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
@@ -19,8 +19,11 @@ internal class ExpressionResolver
 
             return (property, target.IsContainer()) switch
             {
+                // If Container -> Container, we go directly to the container name, bypassing the host
                 (EndpointProperty.Host or EndpointProperty.IPV4Host, true) => target.Name,
+                // If Container -> Exe, we need to go through the container host
                 (EndpointProperty.Host or EndpointProperty.IPV4Host, false) => endpointReference.ContainerHost,
+                // If Container -> Container, we use the target port, since we're not going through the host
                 (EndpointProperty.Port, true) => await endpointReference.Property(EndpointProperty.TargetPort).GetValueAsync(cancellationToken).ConfigureAwait(false),
                 (EndpointProperty.Url, true) => $"{endpointReference.Scheme}://{target.Name}:{endpointReference.TargetPort}",
                 _ => await endpointReference.Property(property).GetValueAsync(cancellationToken).ConfigureAwait(false)
@@ -56,7 +59,9 @@ internal class ExpressionResolver
     internal static async ValueTask<string?> Resolve(bool sourceIsContainer, IValueProvider valueProvider, CancellationToken cancellationToken) =>
         sourceIsContainer switch
         {
-            true => await ResolveWithContainerSource(valueProvider, cancellationToken).ConfigureAwait(false),
-            false => await valueProvider.GetValueAsync(cancellationToken).ConfigureAwait(false)
+            // Exe -> Exe and Exe -> Container cases
+            false => await valueProvider.GetValueAsync(cancellationToken).ConfigureAwait(false),
+            // Container -> Exe and Container -> Container cases
+            true => await ResolveWithContainerSource(valueProvider, cancellationToken).ConfigureAwait(false)
         };
 }
--- a/src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
@@ -9,7 +9,7 @@ internal class ExpressionResolver
 {
     // Resolve an expression when it is being used from inside a container
     // This means that if the target is also a container, we're dealing with container-to-container communication
-    static async ValueTask<string?> ResolveWithContainerSource(object? value, string defaultContainerHostName, CancellationToken cancellationToken)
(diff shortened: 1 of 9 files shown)


src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -0,0 +1,79 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Globalization;
+
+namespace Aspire.Hosting.ApplicationModel;
+
+internal class ExpressionResolver
+{
+    // Resolve an expression when it is being used from inside a container
```
mitchdenny, line 10: Just use actual XML doc comments here.

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -0,0 +1,79 @@
...
+    // This means that if the target is also a container, we're dealing with container-to-container communication
+    static async ValueTask<string?> ResolveWithContainerSource(object? value, HashSet<EndpointReference> processedEndpoints, CancellationToken cancellationToken)
```
mitchdenny, line 12: ResourceWithContainerSourceAsync?

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        async Task<string?> EvalEndpoint(EndpointReference endpointReference, EndpointProperty property)
```
mitchdenny, line 14: EvalEndpointAsync?

mitchdenny (approved):
A few minor pieces of feedback but overall this looks good. I tested it out locally with Azure PGSQL just to verify that scenario isn't impacted.

davidebbo:
@mitchdenny thanks for taking the time to try it with Azure PGSQL.

I made small additional tweaks after chatting with @davidfowl. One challenge is that when the expression only uses the host, or only the port, if we change it, we can break things. This was happening with Kafka, which has `PLAINTEXT_HOST://localhost:{primaryEndpoint.Property(EndpointProperty.Port)}`. Here, we can't change port to targetPort, because the host is hard coded.

I had an initial fix when you reviewed it, but made it more general by using a 2 phase process:
1. Do a walk just to collect whether we use host and/or port for each binding
2. Normal evaluation phase, using data from 1. to decide whether to modify things

I also evolved the test to cover quite a few more scenarios.

davidfowl:
This looks really good! 

We also need to change the cosmos connection string:

https://github.com/dotnet/aspire/blob/d0d3ce55c07baed98891a6e95e294ae6348c347e/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorConnectionString.cs#L10

This needs to use the host and port expression as well.

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -0,0 +1,113 @@
...
+            EndpointReferenceExpression ep => await EvalEndpointAsync(ep.Endpoint, ep.Property).ConfigureAwait(false),
+            IValueProvider vp => await vp.GetValueAsync(cancellationToken).ConfigureAwait(false),
+            _ => throw new NotImplementedException()
+        };
+    }
+
+    internal static async ValueTask<string?> ResolveAsync(bool sourceIsContainer, IValueProvider valueProvider, CancellationToken cancellationToken)
+    {
+        // For each endpoint, store two bools, to track if the host and port are in use
+        var endpointUsage = new Dictionary<EndpointReference, bool[]>();
```
davidfowl, line 100: Instead of an array this could be a tuple:

```suggestion
        var endpointUsage = new Dictionary<EndpointReference, (bool, bool)>();
```

davidebbo: @davidfowl that was my first thoughts, but tuples are immutable, so didn't seem ideal here, since we need to update the values piecemeal.

davidfowl: I just realized how painful it would be to do that yep. A small class would be a little better but this is fine.

davidebbo:
> We also need to change the cosmos connection string

Done.

PR #6059, opened 2024-10-01
Put all database connection strings in keyvault

### LINKED ISSUE #3299: Put all database connection strings in keyvault

eerhardt:
## Description

When using password authentication and child databases in Azure, don't create a connection string based on the parent's connection string (which is a secretOutputs). The reason being that the deployment tool needs to grab the parent server's connection string from the keyvault and add to it.

Instead, put the database connection strings into the keyvault as well. This way all the secrets are contained in the keyvault and they can be referenced the same way as the parent - via keyvault.

Fix #3299

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6059)


davidfowl:
Can you change this test https://github.com/dotnet/aspire/blob/d13467e2c020155aba2b6e29add2f0bb7228129d/tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs#L249

src/Aspire.Hosting.Azure/Utils/BicepIdentifierHelpers.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Azure.Utils;
+
+internal static partial class BicepIdentifierHelpers
+{
+    // See rules from Bicep's highlightjs implementation:
+    // https://github.com/Azure/bicep/blob/a992bdf2d4d7c5c7dec684b7d0de4db9cb260f8a/src/highlightjs/src/bicep.ts#L12
+    [GeneratedRegex("^[A-Za-z_][A-Za-z0-9_]*$")]
```
eerhardt, line 13: Note that this regex changed to no longer allow the Kelvin Sign (which a-z, ignore case allows).

src/Aspire.Hosting.Azure/Utils/BicepIdentifierHelpers.cs
```diff
@@ -0,0 +1,61 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Aspire.Hosting.Azure.Utils;
+
+internal static partial class BicepIdentifierHelpers
```
davidfowl, line 9: cc @captainsafia

davidfowl (approved):

PR #6062, opened 2024-10-02
Limit max size of displayed filters on structured logs and traces pages

JamesNK:
## Description

Add maximum width to filter buttons. Goal is to keep the UI pretty with big data. The full value is available via a tooltip or by clicking filter button to open the dialog.

Before:
image

After:
image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6062)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
+++ b/src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor
@@ -71,7 +71,7 @@
                     {
                         <FluentDivider slot="end" Role="DividerRole.Presentation" Orientation="Orientation.Vertical" />
                     }
-                    <FluentButton slot="end" Appearance="Appearance.Outline" OnClick="() => OpenFilterAsync(filter)">@filter.GetDisplayText(FilterLoc)</FluentButton>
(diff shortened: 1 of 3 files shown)


drewnoakes:
Why is there space here?

image

How did you pick the width of 350px?

An issue in the original screenshot is that the label and textbox don't break together:

image

Even with a width of 350px, is it possible that the line breaks as in the first screenshot? Would be good to keep these two together.

JamesNK:
> Why is there space here?
> 
> image

The content is right aligned. It's still right aligned when wrapped to a new line.

> How did you pick the width of 350px?

I picked it to balance showing most or all filter content most of the time, and showing too much and breaking the UI.

> An issue in the original screenshot is that the label and textbox don't break together:
> 
> image
> 
> Even with a width of 350px, is it possible that the line breaks as in the first screenshot? Would be good to keep these two together.

Wrapping splits, them apart if the width is too large or the browser window is too small. Limiting the filter width like the PR does is the best way I know to keep them together.

adamint (approved):

PR #6069, opened 2024-10-02
Show health-check data in dashboard

### LINKED ISSUE #5569: Expose health check status in the dashboard
In #5515 we are introducing health checks into resources so that when `WaitFor` is used and a health check is associated with a resource it blocks dependent resources from starting up until the check enters either a terminal state or is healthy.

We want to light up this feature in the dashboard so that developers can reason why a particular resource is not healthy. The key features are:

1. [x] The status on each resource should have its status and its health status ... e.g. `Running (Healthy)` or `Running (Unhealthy)`. If the resource does not have any health checks associated with it then the health status can be ommitted.
2. [ ] If the status is unhealthy, the text `Unhealthy` should be a hyperlink that brings up the resource details panel and displays the health checks section.
3. [ ] The health checks section in the panel should list all the checks in the resource (typically just one but there can be more).
4. [ ] The health check will need to show not just the health check entry status, but also any messag/exception details.

drewnoakes:
## Description

Fixes #5569

This change plumbs through data about health-checks to the dashboard.

Values are displayed in the resource detail view, in their own accordian section, in a new three-column grid:

The icon and popup used in the structured log message column to show exceptions was pulled into a new `<ExceptionDetails>` component and reused for health check error data.

image

Tooltip handling on the resource grid's state column cells was previously inconsistent. We would sometimes add a tooltip to the icon and text, and other times to the cell's entire bounds. Formatting logic for the resource grid's state column is moved into code-behind, simplifying the razor and ensuring consistent treatment. Tooltips are now always at the cell level, making them easier to hit with the mouse, and making the accessible text consistent.

image

As part of this change, a new `ContentBeforeValue` parameter was added to `GridValue`, as a complement to `ContentAfterValue`.

A new "HealthChecksSandbox" playground was added to make testing of health check data easier.

Resources without health checks appear immediately healthy.

health-state

A placeholder message is shown in the resource details until the first health data arrives:

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6069)


JamesNK:
image

It would be much easier to tell at a glance the health status by having an icon here to go with the text.

Also, I think the description/exception should be in a popup. That avoids overloading the column with too much information.

You probably don't want to use the PropertyGrid control because of the custom content, but having a custom data grid in a details control is simple enough: https://github.com/dotnet/aspire/blob/d0d3ce55c07baed98891a6e95e294ae6348c347e/src/Aspire.Dashboard/Components/Controls/SpanDetails.razor#L94-L113

JamesNK:
image

What happens if wait for is depending on something with multiple instances? Are multiple rows displayed? Also, is the resource name here the full name, e.g. `message-abcxyz` but resolved to message, or the display name, e.g. `message`?

The dashboard tends to use the full name everywhere but it's not obvious because it shortens it when possible. It should keep doing that if possible. But I don't know if the full name is known at this point.

davidfowl:
Lots of new test falures.

src/Aspire.Hosting/Dashboard/ResourceSnapshot.cs
```diff
@@ -26,8 +25,9 @@ internal abstract class ResourceSnapshot
     public required ImmutableArray<EnvironmentVariableSnapshot> Environment { get; init; }
     public required ImmutableArray<VolumeSnapshot> Volumes { get; init; }
     public required ImmutableArray<UrlSnapshot> Urls { get; init; }
-    public required HealthStateKind? HealthState { get; set; }
+    public required ImmutableArray<HealthReportSnapshot> HealthReports { get; init; }
```
davidfowl, line 28: We want the aggregate state top level as well as exposing individual heath states.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -71,12 +74,25 @@ public sealed record CustomResourceSnapshot
...
+    /// </summary>
+    public ImmutableArray<WaitForSnapshot> WaitFors { get; init; } = [];
+
+    /// <summary>
+    /// Gets the aggregate health status of the resource.
+    /// </summary>
+    /// <remarks>
+    /// This property's value is derived from <see cref="HealthReports"/> and is provided for convenience.
+    /// </remarks>
+    public HealthStatus HealthStatus => HealthReports.MinBy(r => r.Status)?.Status ?? HealthStatus.Healthy;
```
davidfowl, line 89: This should be a cheap check. Let's just store the aggregate and not make this computed.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -53,7 +55,23 @@ public sealed record CustomResourceSnapshot
...
+        }
+    }
+
+    /// <summary>
+    /// Gets the aggregate health status of the resource.
+    /// </summary>
+    /// <remarks>
+    /// This property's value is derived from <see cref="HealthReports"/> and is provided for convenience.
+    /// </remarks>
+    public HealthStatus HealthStatus { get; private set; } = HealthStatus.Healthy;
```
davidfowl, line 74: Null by default?

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -53,7 +55,23 @@ public sealed record CustomResourceSnapshot
...
     /// May be zero or more. If there are no health reports, the resource is considered healthy.
     /// </remarks>
-    public ImmutableArray<HealthReportSnapshot> HealthReports { get; init; } = [];
+    public ImmutableArray<HealthReportSnapshot> HealthReports
+    {
+        get => _healthReports;
+        init
+        {
+            _healthReports = value;
+            HealthStatus = value.MinBy(r => r.Status)?.Status ?? HealthStatus.Healthy;
```
davidfowl, line 64: Why compute it at all? The code that sets this has the precomputed value?

davidfowl:
> I think we should have an icon here. It would be much easier to tell at a glance the health status by having an icon here to go with the text.

+ 100, spoke to @drewnoakes about this too.

> Also, I think the description/exception should be in a popup. The structured logs grid does that with exception information. There isn't enough room for 3 columns with text, so a popup avoids overloading the column with too much information.

The popup we have for the exception message today is very annoying as it goes away on hover. Maybe we can make them be the same and then fix both 😄

drewnoakes:
@davidfowl @JamesNK see the updated PR description for info on changes.

- Removed "waits for" info
- Updated presentation of health checks (screenshots in description)

src/Aspire.Dashboard/Resources/ControlsStrings.resx
```diff
@@ -299,6 +305,9 @@
   <data name="ResourceDetailsVolumesHeader" xml:space="preserve">
     <value>Volumes</value>
   </data>
+  <data name="ResourceHealthChecksHeader" xml:space="preserve">
+    <value>Health Checks</value>
```
JamesNK, line 309: Change to `Health checks`. Other section titles capitalize the first word only.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -108,6 +111,60 @@
...
+                                GridTemplateColumns="1fr 1fr 1.5fr"
+                                ShowHover="true">
+                    <AspireTemplateColumn Title="@ControlStringsLoc[nameof(ControlsStrings.NameColumnHeader)]" Class="nameColumn">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.NameColumnHeader)]"
+                                   Value="@context.Name"
+                                   EnableHighlighting="@(!string.IsNullOrEmpty(_filter))"
+                                   HighlightText="@_filter" />
+                    </AspireTemplateColumn>
+                    <AspireTemplateColumn Title="@ControlStringsLoc[nameof(ControlsStrings.StateColumnHeader)]" Class="stateColumn">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.StateColumnHeader)]"
```
JamesNK, line 136: I noticed when I searched for the state name, e.g. `Healthy`, the healthcheck disappeared. I think either don't highlight state when searching, or have the state be detected by searching.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -108,6 +111,60 @@
...
+                                   HighlightText="@_filter"
+                                   TextVisualizerTitle="@context.Name">
+                            <ContentBeforeValue>
+                                @{
+                                    (Icon icon, Color color) = context.HealthStatus switch
+                                    {
+                                        HealthStatus.Healthy => ((Icon)new Icons.Filled.Size16.Heart(), Color.Success),
+                                        HealthStatus.Degraded => (new Icons.Filled.Size16.HeartBroken(), Color.Warning),
+                                        HealthStatus.Unhealthy => (new Icons.Filled.Size16.HeartBroken(), Color.Error),
+                                        _ => (new Icons.Regular.Size16.Heart(), Color.Success)
```
JamesNK, line 148: Could health status be something else?

I think either throw an error (which would break the whole UI so probably not a good idea) or have an unknown icon. Could reuse what the resources grid shows which I think is a grey circle.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -252,3 +252,51 @@ public bool MatchesFilter(string filter) =>
...
+
+    string? IPropertyGridItem.Value => HealthStatus.ToString();
+
+    string? IPropertyGridItem.ValueToCopy => _compositeValue;
+
+    string? IPropertyGridItem.ValueToVisualize => _compositeValue;
+
+    private static string GetCompositeValue(HealthStatus HealthStatus, string? Description, string? Exception)
+    {
+        StringBuilder builder = new();
```
JamesNK, line 274: var

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -252,3 +252,51 @@ public bool MatchesFilter(string filter) =>
...
+
+        builder.Append(HealthStatus.ToString());
```
JamesNK, line 276: Other places call humanize on healthstatus. Should be consistent.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -252,3 +252,51 @@ public bool MatchesFilter(string filter) =>
...
+
+        return builder.ToString();
+    }
+
+    public bool MatchesFilter(string filter)
+    {
+        return
+            Name?.Contains(filter, StringComparison.CurrentCultureIgnoreCase) == true ||
+            Description?.Contains(filter, StringComparison.CurrentCultureIgnoreCase) == true ||
+            Exception?.Contains(filter, StringComparison.CurrentCultureIgnoreCase) == true;
```
JamesNK, line 300: I think the problem with including exception here is it isn't visible on the page. While searching you don't see text it being highlighted. It could be confusing why the health check is still visible. IMO remove it.

Also, add a check on the humanized result of the healthstatus enum (see earlier comment about it being filtered out). Maybe `HealthStatus.Humanize()` should be cached somewhere. Don't want to repeatedly process it in a Where statement. I don't know how expensive it is.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -108,6 +111,60 @@
...
+                                    };
+                                }
+                                <FluentIcon Value="@icon" Color="@color" Class="severity-icon" />
+                            </ContentBeforeValue>
+                        </GridValue>
+                    </AspireTemplateColumn>
+                    <AspireTemplateColumn Title="@ControlStringsLoc[nameof(ControlsStrings.DetailsColumnHeader)]" Class="detailsColumn">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.DetailsColumnHeader)]"
```
JamesNK, line 156: What do you think about having a placeholder value if there aren't details?

image

Something like `(None)` or `(Empty)` might feel better than an empty column.

We don't do that for other values on these grids yet so probably shouldn't start in this PR.

src/Aspire.Dashboard/ResourceService/Partials.cs
```diff
@@ -48,19 +48,31 @@ public ResourceViewModel ToViewModel(BrowserTimeProvider timeProvider, IKnownPro
...
 
+        HealthReportViewModel ToHealthReportViewModel(HealthReport healthReport)
+        {
+            return new HealthReportViewModel(healthReport.Key, Convert(healthReport.Status), healthReport.Description, healthReport.Exception);
+
+            static Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus Convert(HealthStatus healthStatus)
+            {
+                return healthStatus switch
+                {
+                    HealthStatus.Unknown or HealthStatus.Healthy => Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus.Healthy,
```
JamesNK, line 68: Why is unknown state considered healthy? Intentional or mistake?

If it is intentional, could you add a comment saying why.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -159,6 +172,15 @@ internal void Deconstruct(out string name, out object? value, out bool isSensiti
...
 public sealed record ResourceCommandSnapshot(string Type, ResourceCommandState State, string DisplayName, string? DisplayDescription, object? Parameter, string? ConfirmationMessage, string? IconName, IconVariant? IconVariant, bool IsHighlighted);
 
+/// <summary>
+/// A report produced by a health check about a resource.
+/// </summary>
+/// <param name="Name">The name of the health check that produced this report.</param>
+/// <param name="Status">The state of the resource, according to the report.</param>
+/// <param name="Description">An optional description of the report, for display.</param>
+/// <param name="Exception">An optional string containing exception details.</param>
+public sealed record HealthReportSnapshot(string Name, HealthStatus Status, string? Description, string? Exception);
```
JamesNK, line 182: `Exception` -> `ExceptionDetails`? It is surprising to me to have a property called Exception that is a string and not an `System.Exception`.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -191,22 +209,17 @@ message Resource {
...
-    optional HealthStateKind HealthState = 16;
+    // The aggregate health state of the resource. Generally reflects data from health_reports, but may differ.
+    optional HealthStatus health_status = 16;
+    // Reports from health checks, about this resource.
+    repeated HealthReport health_reports = 19;
 
     // The resource start time.
     optional google.protobuf.Timestamp started_at = 17;
     // The resource stop time.
     optional google.protobuf.Timestamp stopped_at = 18;
```
JamesNK, line 220: I believe these are all new in 9.0 so you can adjust the field numbers to not be out of order.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
drewnoakes: Looking at the implementation, it's moderately expensive.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
drewnoakes: We're inconsistent on empty values. Even in the resource grid, we have show "(None)" for some columns (like endpoints) but not others (like source).

src/Aspire.Dashboard/ResourceService/Partials.cs
drewnoakes: Innocent until proven guilty. Will add a comment.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
drewnoakes: Agree. I've used `ExceptionText` elsewhere so will be consistent with that.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
drewnoakes: We use a dashed circle for unknown resource status. There's no dashed heart icon, so I'll use the same dashed circle (fluent calls it a hint).

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
drewnoakes: Will do. FYI @snehapar9, in case you've already implemented anything in ACA for this proto.

src/Aspire.Dashboard/ResourceService/Partials.cs
JamesNK: What does wait for do when a dependency has a health check that returns unknown? This should be consistent with it @mitchdenny 

It would be odd to have a green tick with an unknown status in the UI, but wait for is still waiting for it to enter a health state.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
JamesNK: This is partially done. Code in the app host like this snapshot is still `Exception`.

src/Aspire.Dashboard/ResourceService/Partials.cs
JamesNK: I looked into this more, and we're using the extensions health check enum on the server and the client (dashboard) with an unknown status inbetween on the proto layer.

It's impossible for the app host to return unknown, but a custom resource service implementation could return it.

I think the dashboard should stop using `Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus` and use the enum that reflects all the values that can be returned from the server. That way we don't have to guess what is meant by unknown.

JamesNK (commented):
A couple more things

JamesNK:
It's possible to have an exception with no details:

image

This looks a little odd. How about in this situation (there are exception details but no text details) default the details text to "Exception details"?

src/Aspire.Dashboard/ResourceService/Partials.cs
JamesNK: Alternatively, maybe there shouldn’t be an unknown option in the service layer?

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
```diff
@@ -50,13 +48,7 @@ static GenericResourceSnapshot CreateResourceSnapshot(IResource resource, string
                     ExitCode = snapshot.ExitCode,
                     State = snapshot.State?.Text,
                     StateStyle = snapshot.State?.Style,
-                    HealthState = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _) ? snapshot.HealthStatus switch
```
mitchdenny, line 53: As discussed in our call, this is likely the cause of the incorrect state issues we are seeing in the dashboard. Thinking about it a bit more, its possibly also the reason why we were polling at increased intervals as well (not 100% sure on that though).

Just a reminder to fix this before merging.

davidfowl: What did you see?

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -45,6 +45,11 @@ public partial class ResourceDetails
             .Where(vm => vm.MatchesFilter(_filter))
             .AsQueryable();
 
+    private IQueryable<HealthReportViewModel> FilteredHealthReports =>
+        Resource.HealthReports
```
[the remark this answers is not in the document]
drewnoakes: Ordered when we create `ResourceViewModel` from the protobuf so we don't order twice per render.

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
drewnoakes: When the resource is created, it has no health reports so appears as "Running". Then the reports arrive, and it transitions to "Running (Unhealthy)", then finally back to "Running".

I'm going to ensure that a resource with a health check annotation that has no reports is still considered unhealthy.

There was an additional behaviour we saw where the health check polling was happening more frequently than we expected it to. I'll fix this and we can check whether things return to what we expected to see.

davidfowl: Don't have any health reports should not be the same as the health reports have not arrived yet. So treat null different to 0.

Mrxx99:
Was the displaying of WaitFor removed from this PR because
a) It will be done in a separate PR
b) The ideas was scrapped?

I hope it's a) because it seems quite useful, esp. if it would show the wait status as icon (either waiting/not waiting or the status icon of the resource which it is waiting on would be a good solution)

davidfowl:
Yes it was removed, we want to design a relationships feature and waiting will be one kind of relationship.

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
drewnoakes: This has been fixed in the latest push. Resources that expect health checks have a null health state, which the UI shows as unhealthy.

Resources without health checks appear immediately healthy.

I've also added a placeholder message in the resource details area that appears until the first health data arrives.

See the PR description for new video/screenshots.

src/Aspire.Dashboard/ResourceService/Partials.cs
drewnoakes: I believe the most recent push addresses the concerns here.

drewnoakes:
> It's possible to have an exception with no details:
> 
> image
> 
> This looks a little odd. How about in this situation (there are exception details but no text details) default the details text to "Exception details" or "Error thrown", etc.

I updated the column to show the first line of the exception text when an exception exists. The icon won't look lonely any more.

image

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -48,9 +48,26 @@ public sealed record CustomResourceSnapshot
...
+    /// Defaults to <see cref="HealthStatus.Healthy"/>.
+    /// </para>
+    /// </remarks>
+    public HealthStatus? HealthStatus { get; init; } = Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus.Healthy;
+
+    /// <summary>
+    /// The health reports for this resource.
+    /// </summary>
+    /// <remarks>
+    /// May be zero or more. If there are no health reports, the resource is considered healthy.
```
drewnoakes, line 68:
```suggestion
    /// May be zero or more. If there are no health reports, the resource is considered healthy
    /// so long as no heath checks are registered for the resource.
```

JamesNK (approved):
The only outstanding thing is guessing a status for unknown. It is still happening. I think the simplest fix is remove UNKOWN from the enum. It's new in 9 so it isn't a breaking change.

There are also a bunch of parameters with init instead of set.

Otherwise, it looks good to me.

drewnoakes:
> The only outstanding thing is guessing a status for unknown. It is still happening. I think the simplest fix is remove UNKOWN from the enum. It's new in 9 so it isn't a breaking change.

@JamesNK the item having a zero value seems important, as an omitted field would get that default. I can see a default of both healthy or unhealthy making sense, and having an UNKNOWN item avoids that decision. Do you have a view on that?

JamesNK:
This is a best practice guide about having a zero value on an enum: https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

Personally I think we should rely on implementors to always set a value, and if they don't then they'll observe that not setting a value will default to Healthy. We're not adding this enum to an existing message so we don't need to keep compatibility with someone not setting it because they have an old proto spec.

Or we could follow the guide and have a zero value option that is `UNSET`/`UNSPECIFIED`/`UNDEFINED`/`UNKNOWN`. But the guide says not to assign semantic meaning to it having the unspecified value which I agree with. Either we display a health check in the UI with a status of unknown (ok), or the dashboard throws an error (bad, because it breaks everything else). But I don't like guessing a real status from a missing status from the server.

src/Aspire.Dashboard/ResourceService/Partials.cs
drewnoakes: See 68578f85006decbd721df7d2017a18cf9489ff99 and 14d55aca335107abbe937a52ec2f14921084ec68.

drewnoakes:
CI failure is a failing test in `MySqlFunctionalTests.VerifyWaitForOnMySqlBlocksDependentResources` which seems unrelated to this PR.

PR #6080, opened 2024-10-03
Improve resource logging, and using wait for with replicas

### LINKED ISSUE #5626: Console Log Viewer is weird with `WaitFor` / `WaitForCompletion`

### Describe the bug

If you view console logs of a resource whilst it's `WaitFor` or `WaitForCompletion` on another resource, then it appears as though the resource never starts as the page hangs for ever and never starts showing the container logs

image

### Expected Behavior

Once the resource has finished waiting, application logs should start streaming e.g.
image

### Steps To Reproduce

```cs
var pause = builder.AddExecutable("pause", "pwsh", ".",
    "-NoProfile",
    "-NoLogo",
    "-NonInteractive",
    "-Command",
    "Write-Host 'Starting Migration'; Start-Sleep -Seconds 10; Write-Host 'Done'");

var redis = builder.AddRedis("redis")
    .WaitForCompletion(pause);
```

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

The problem seems to be related to replicas.  Once a resource starts after waiting, the view console logs url changes from https://localhost:17206/consolelogs/resource/redis to https://localhost:17206/consolelogs/resource/redis-xzsjzykw .

The UI is aware that something has happened as the resource drop down changes from having the resource name selected, to `(None)`.

### LINKED ISSUE #5908: Wait for and resource health logging isn't visible

JamesNK:
## Description

Fixes a number of issues:

* WaitFor displays one resource in the dashboard while it waits to start.
* Resource name sent to dashboard changes when a DCP resource goes from Starting/Waiting to Running.
* WaitFor logging wasn't visible in dashboard console logs.
* Events were raised for each replica for a resource. This would cause wait for to run for each replica (this probably wasn't bad because once the tests pass for one replica, they pass for all, but it wasn't intented behavior, and you'd see multiple logs)

PR changes:

* Capture project/container/exe instance ID when first publishing resource rather than waiting for resource to start.
* Use instance IDs to fan out resource notifications and logs.
* Fix events being raised multiple times for replicas.

Fixes https://github.com/dotnet/aspire/issues/5626
Fixes https://github.com/dotnet/aspire/issues/5908

wait-for-replicas-logs2

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6080)


src/Aspire.Hosting/ApplicationModel/ReplicaInstancesAnnotation.cs
```diff
@@ -10,5 +10,5 @@ namespace Aspire.Hosting.ApplicationModel;
 /// </summary>
 internal class ReplicaInstancesAnnotation : IResourceAnnotation
 {
-    public ConcurrentDictionary<string, string> Instances { get; } = [];
+    public ConcurrentDictionary<string, int> Instances { get; } = [];
```
JamesNK, line 13: int = replica index.

I was using this at one point, but it can be reverted. I left it in in case there is a reason we want to know replica instance from the `IResource` layer.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -468,9 +468,25 @@ static ResourceCommandSnapshot CreateCommandFromAnnotation(ResourceCommandAnnota
...
     {
-        return PublishUpdateAsync(resource, resource.Name, stateFactory);
+        var resourceNames = ResolveResourceNames(resource);
+        foreach (var resourceName in resourceNames)
+        {
+            await PublishUpdateAsync(resource, resourceName, stateFactory).ConfigureAwait(false);
+        }
+    }
+
+    internal static string[] ResolveResourceNames(IResource resource)
```
JamesNK, line 480: This is used by a couple of times. I'll move it to a better place before merge.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
```diff
@@ -17,7 +17,8 @@
 builder.AddProject<Projects.WaitForSandbox_ApiService>("api")
        .WithExternalHttpEndpoints()
        .WithReference(db).WaitFor(db)
-       .WaitForCompletion(dbsetup);
+       .WaitForCompletion(dbsetup)
+       .WithReplicas(2);
```
davidfowl, line 21: You haven't tested waiting on projects that have replicas. I think that will wait on the first one to be ready.

JamesNK: `WaitForCompletion` for a replica doesn't work: https://github.com/dotnet/aspire/blob/05cbb1cf5e446d5e6fd0c5ebac63fa708af3f964/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L188-L191

This is point 1 from the teams chat. Is it supposed to work? @mitchdenny

davidfowl: It’s blocked explicitly but it’s possible to make it work in theory with this annotation

davidfowl:
What happens if you log to a project/executable resource before the annotation is added? They will be missed right?

JamesNK:
> What happens if you log to a project/executable resource before the annotation is added? They will be missed right?

Yes. If it is a problem then we could move adding the annotation earlier.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -411,12 +410,6 @@ private async Task ProcessResourceChange<T>(WatchEventType watchEventType, T res
                         _logger.LogTrace("Updating application model resource {ResourceName} with {ResourceKind} resource {ResourceName}", appModelResource.Name, resourceKind, resource.Metadata.Name);
                     }
 
-                    if (_hiddenResources.TryAdd(appModelResource, true))
-                    {
-                        // Hide the application model resource because we have the DCP resource
-                        await notificationService.PublishUpdateAsync(appModelResource, s => s with { State = "Hidden" }).ConfigureAwait(false);
-                    }
```
davidfowl, line 418: This isn't a small change. I like that we can remove it but I don't understand why.

JamesNK: Yes, it's always right.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
JamesNK: Issue: https://github.com/dotnet/aspire/issues/6083

src/Aspire.Hosting/Dcp/DcpNameGenerator.cs
```diff
@@ -0,0 +1,70 @@
...
+
+internal sealed class DcpNameGenerator
+{
+    // A random suffix added to every DCP object name ensures that those names (and derived object names, for example container names)
+    // are unique machine-wide with a high level of probability.
+    // The length of 8 achieves that while keeping the names relatively short and readable.
+    // The second purpose of the suffix is to play a role of a unique OpenTelemetry service instance ID.
+    private const int RandomNameSuffixLength = 8;
+    private readonly IConfiguration _configuration;
+    private readonly IOptions<DcpOptions> _options;
```
davidfowl, line 19: nit:
```suggestion
    private readonly DcpOptions _options;
```

src/Aspire.Hosting/Dcp/DcpNameGenerator.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    public DcpNameGenerator(IConfiguration configuration, IOptions<DcpOptions> options)
+    {
+        _configuration = configuration;
+        _options = options;
```
davidfowl, line 24: nit:

```suggestion
        _options = options.Value;
```

src/Aspire.Hosting/Dcp/DcpNameGenerator.cs
```diff
@@ -0,0 +1,70 @@
...
+            _ => _configuration["AppHost:Sha256"]!.Substring(0, RandomNameSuffixLength).ToLowerInvariant(),
+        };
+
+        return (GetObjectNameForResource(container, _options.Value, nameSuffix), nameSuffix);
+    }
+
+    public (string Name, string Suffix) GetExecutableName(IResource project)
+    {
+        var nameSuffix = GetRandomNameSuffix();
+        return (GetObjectNameForResource(project, _options.Value, nameSuffix), nameSuffix);
```
davidfowl, line 42:
```suggestion
        return (GetObjectNameForResource(project, _options, nameSuffix), nameSuffix);
```

src/Aspire.Hosting/ApplicationModel/ReplicaInstancesAnnotation.cs
```diff
@@ -6,9 +6,11 @@
 namespace Aspire.Hosting.ApplicationModel;
 
 /// <summary>
-/// Keep track of active replicas ids from DCP.
+/// Keep track of instances from DCP.
 /// </summary>
 internal class ReplicaInstancesAnnotation : IResourceAnnotation
 {
-    public ConcurrentDictionary<string, string> Instances { get; } = [];
+    public ConcurrentDictionary<string, Instance> Instances { get; } = [];
```
davidfowl, line 13: Is there any benefit to this being a dictionary? Seems like it could be an `ImmutableArray<Instance>`

src/Aspire.Hosting/Dcp/DcpNameGenerator.cs
JamesNK: I left it as it is because getting the value causes validation to run, and if DcpOptions doesn't have a value then it errors.

davidfowl (approved):

PR #6081, opened 2024-10-03
Add WithHttpHealthCheck/WithHttpsHealthCheck.

mitchdenny:
## Description

This PR adds `WithHttpHealthCheck` and `WithHttpsHealthCheck`. This enables any resource that implements `IResourceWithEndpoints` to quickly have a health check added to it so that it can paticipate in the `WaitFor(...)` API pattern (and show readiness status on the dashboard).

Related #5645 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6081)


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,67 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+    {
+        path = path ?? "/";
+        endpointName = endpointName ?? "http";
+        statusCode = statusCode ?? 200;
+
+        Uri? uri = null;
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, (@event, ct) =>
+        {
+            var endpoint = builder.Resource.GetEndpoint(endpointName);
+            var uriString = $"{endpoint.AllocatedEndpoint.UriString}{path}";
```
davidfowl, line 731: endpoint.Url

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,67 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+            uri = new Uri(uriString, UriKind.Absolute);
```
davidfowl, line 732: You don't want to mess with slashes right?

```suggestion
            uri = new Uri(endpoint.Url, path, UriKind.Absolute);
```

DamianEdwards, line 732: Should we be more defensive WRT the value of `path` here before we append it to the URI string and pass it to the `Uri` constructor? Or are we happy letting invalid path values cause an exception in the `Uri` ctor and having that bubble out?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,67 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+    /// <summary>
+    /// Adds a health check to the resource which is mapped to a specific endpoint.
+    /// </summary>
+    /// <typeparam name="T">A resource type that implements <see cref="IResourceWithEndpoints" />.</typeparam>
+    /// <param name="builder">A resource builder.</param>
+    /// <param name="path">The relative path to test.</param>
+    /// <param name="statusCode">The result code to interpret as healthy.</param>
+    /// <param name="endpointName">The name of the endpoint to derive the base address from.</param>
+    /// <returns>A resource builder.</returns>
+    public static IResourceBuilder<T> WithHttpHealthCheck<T>(this IResourceBuilder<T> builder, string? path = null, int? statusCode = null, string? endpointName = null) where T : IResourceWithEndpoints
```
DamianEdwards, line 721: This overloading feels a little strange, i.e. `WithHttpHealthCheck` allowing specifying an endpoint name that isn't "http", and vice-versa for `WithHttpsHealthCheck`. Should we consider having a `WithEndpointHealthCheck` that these two methods just call down into passing the appropriate endpoint name instead?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,67 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
         return builder;
     }
 
+    /// <summary>
+    /// Adds a health check to the resource which is mapped to a specific endpoint.
+    /// </summary>
+    /// <typeparam name="T">A resource type that implements <see cref="IResourceWithEndpoints" />.</typeparam>
+    /// <param name="builder">A resource builder.</param>
+    /// <param name="path">The relative path to test.</param>
+    /// <param name="statusCode">The result code to interpret as healthy.</param>
```
DamianEdwards, line 718: Would be good to document the defaults on this and other parameters as applicable.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
mitchdenny: What protocol would be using for WithEndpointHealthCheck. I know HTTP rules the world but ....

src/Aspire.Hosting/ResourceBuilderExtensions.cs
mitchdenny: I'm happy for Uri ctor to do the work here for us to detect invalid URL syntax.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: It would obtain it from the endpoint specified and if it's not an HTTP-based scheme it would throw.

davidfowl: This should be coupled to http, not endpoints in general. Throw if the scheme isn't http as well.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,66 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+    {
+        path = path ?? "/";
+        endpointName = endpointName ?? "http";
+        statusCode = statusCode ?? 200;
+
+        Uri? uri = null;
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, (@event, ct) =>
+        {
+            var endpoint = builder.Resource.GetEndpoint(endpointName);
+            uri = new Uri(endpoint.Url, path, UriKind.Absolute);
```
davidfowl, line 731:
```suggestion
            uri = new Uri(new Uri(endpoint.Url), path);
```

davidfowl: OR:

```
var uri = new UriBuilder(endpoint.Url) { Path = path }.Uri;
```

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: So should `WithHttpsHealthCheck` throw if the scheme of the endpoint specified isn't `https`? Or is it just a shortcut for specifying the endpoint named "https" and it throws if the scheme of that endpoint isn't "http" or "https"?

mitchdenny: Yes. I'm doing that now and have covering tests.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,123 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+    public static IResourceBuilder<T> WithHttpHealthCheck<T>(this IResourceBuilder<T> builder, string? path = null, int? statusCode = null, string? endpointName = null) where T : IResourceWithEndpoints
+    {
+        path = path ?? "/";
+        endpointName = endpointName ?? "http";
+        statusCode = statusCode ?? 200;
+
+        if (builder.Resource.GetEndpoint(endpointName) is not { Scheme: "http" })
+        {
+            throw new DistributedApplicationException($"The endpoint '{endpointName}' on resource '{builder.Resource.Name}' was not using the 'http' scheme.");
+        }
```
davidfowl, line 750: I think this check should be deferred, so you can just call it later and it'll fail when it needs to be resolved.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -706,6 +709,123 @@ public static IResourceBuilder<T> WithHealthCheck<T>(this IResourceBuilder<T> bu
...
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> WithHttpsHealthCheck<T>(this IResourceBuilder<T> builder, string? path = null, int? statusCode = null, string? endpointName = null) where T : IResourceWithEndpoints
+    {
+        endpointName = endpointName ?? "https";
+
+        if (builder.Resource.GetEndpoint(endpointName) is not { Scheme: "https" })
+        {
+            throw new DistributedApplicationException($"The endpoint '{endpointName}' on resource '{builder.Resource.Name}' was not using the 'https' scheme.");
+        }
```
davidfowl, line 824: This should be deferred as well. Instead of having 2 checks we should just check for either http or https in the `WithHttpHealthCheck` method.

davidfowl (approved):

PR #6085, opened 2024-10-03
Enhance Azure Event Hubs hosting integration to support AzureEventHubResource child

### LINKED ISSUE #5561: Aspire.Hosting.Azure.EventHubs should support configuring Consumer Groups
Today, Aspire.Hosting.Azure.EventHubs allows users to add EventHub instances:

https://github.com/dotnet/aspire/blob/7b1fd6d6fe5aa0c6fc3680f0e23ec127c9743789/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs#L80-L85

However, the developer can't add consumer groups to the EventHub.

We should allow for consumer groups to be modeled.

One issue is the `AddEventHub` method doesn't return an `EventHubResource`, but instead returns the `AzureEventHubsResource` (which represents the Event Hubs Namespace). We will need to figure out how to rectify that, as we would need to model EventHubResource as a class we can add consumer groups to.

We may also want to be able to configure partition count as well.

cc @sebastienros @davidfowl

oising:
## Description

Addresses: #5561 

- [x] Add AzureEventHubResource
  - [x] Independent ConnectionString from namespace, includes EntityPath
  - [x] Support Azure Functions bits
- [ ] Allow configuration of Consumer Groups
  - [ ] Support in Emulator
  - [ ] Support live / Azure
    - [ ] AzureEventHub CDK -- requires support (currently missing)
- [ ] Allow configuration of Partition Count
  - [ ] Support in Emulator
  - [ ] Support live / Azure
- [x] Update playground code

This is a _breaking change_ as you can no longer chain `.AddEventHub(...)` calls, but the connection string will now include the EntityPath. The latter shouldn't cause any issues as the current workaround to this missing element is to override in the settings callback. Settings will continue to override whatever is returned. I believe any breaking changes are limited to the AppHost. 

Also, the ordering of `RunAsEmulator()` is important (as it is in other child-resource hosting components like Azure Storage), e..g

```csharp
        var resource = builder.AddAzureEventHubs("resource")
                              .AddEventHub("hubx")
                              .RunAsEmulator()
                              .WithHealthCheck("blocking_check");
```

This used to work because `AddEventHub` returned the parent `AzureEventHubsResource`. This is no longer the case. The RunAsEmulator() needs to move to directly after `AddAzureEventHubs(...)`.

Fixes #5561

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6085)


src/Aspire.Hosting.Azure.EventHubs/PublicAPI.Shipped.txt
```diff
@@ -6,5 +6,3 @@ Aspire.Hosting.Azure.AzureEventHubsResource.EventHubsEndpoint.get -> Aspire.Host
 Aspire.Hosting.AzureEventHubsExtensions
 static Aspire.Hosting.AzureEventHubsExtensions.AddAzureEventHubs(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>!
 static Aspire.Hosting.AzureEventHubsExtensions.AddAzureEventHubs(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>!, Aspire.Hosting.ResourceModuleConstruct!, Azure.Provisioning.EventHubs.EventHubsNamespace!>? configureResource) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>!
-static Aspire.Hosting.AzureEventHubsExtensions.AddEventHub(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>! builder, string! name) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Azure.AzureEventHubsResource!>!
```
eerhardt, line 9: This isn't how this is supposed to work. You don't modify the "Shipped.txt" file. Instead you put a `*REMOVED*` line in the .Unshipped.txt file.

oising: I have no idea how this undocumented, opaque and loathsome analyzer is meant to work. I went along with the Rosyln Analyzer's suggestion to correct the problem and here we are 😆

eerhardt: You aren't alone.

cc @agocke

oising:
@mitchdenny @eerhardt @davidfowl 

Do you want me to do anything with this? Do I wait until you've invented something similar for ESB topics and queue child resources, or is there a pattern I can follow now?

eerhardt:
@sebastienros - any thoughts here? This seems the same as https://github.com/dotnet/aspire/pull/6737

davidfowl:
Fixed in 9.1

PR #6088, opened 2024-10-03
Rename ContainerApps assembly to AppContainers

eerhardt:
This matches the Azure.Provisioning name, which is the pattern we follow for our Hosting.Azure packages.

See

https://github.com/dotnet/aspire/blob/a18868f4bf0e3e739eea5480af1feeef926e869b/src/Aspire.Hosting.Azure.ContainerApps/Aspire.Hosting.Azure.ContainerApps.csproj#L16

We replace `Azure.Provisioning` with `Aspire.Hosting.Azure` and keep the same last term.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6088)


davidfowl:
Saw this coming but I didn't want to to do it 😃

eerhardt:
I dislike this name, but being consistent is higher pri IMO.

tg-msft (approved):

PR #6091, opened 2024-10-03
Rename components to integrations in templates

captainsafia:
Just a small thing I noticed when creating a project with the templates. We're still using the old "components" verbiage in the code comment.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6091)

### DIFF
--- a/playground/Qdrant/Qdrant.ApiService/Program.cs
+++ b/playground/Qdrant/Qdrant.ApiService/Program.cs
@@ -3,7 +3,7 @@ using Qdrant.Client.Grpc;
 
 var builder = WebApplication.CreateBuilder(args);
 
-// Add service defaults & Aspire components.
+// Add service defaults & Aspire integrations.
 builder.AddServiceDefaults();
 
 // Add services to the container.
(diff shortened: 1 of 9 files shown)


eerhardt (approved):

playground/Qdrant/Qdrant.ApiService/Program.cs
```diff
@@ -3,7 +3,7 @@
 
 var builder = WebApplication.CreateBuilder(args);
 
-// Add service defaults & Aspire components.
```
joperezr, line 6: Is this right though? cc @DamianEdwards. Service Defaults won't really add Aspire Integrations to your services, it does add ServiceDiscovery which will help in finding those, but doesn't seem like the comment was accurate here unless "Components" here really meant the "Aspire things"

DamianEdwards: The comment should read `// Add service defaults & Aspire client integrations.` now I think. The comment mentions both because the intent was this is where you call methods that extend `IHostApplicationBuilder` directly, which is what service defaults and Aspire client integrations do.

captainsafia: > which is what service defaults and Aspire client integrations do.

Yep, this was my understanding of the comments intent.

But good point that we should be specific about these being "client integrations" so users can discern what parts of the docs they should be reading when adding code here.

DamianEdwards (approved):

PR #6095, opened 2024-10-04
Show persistent containers in the dashboard

### LINKED ISSUE #5947: Add dashboard UX showing that a container is persistent
We added persistent containers in the model and they don't show up any differently in the dashboard. It would be nice to have a glyph with a tooltip that explained that the lifetime is managed outside of aspire.

cc @danegsta @drewnoakes @JamesNK @leslierichardson95

drewnoakes:
## Description

Fixes #5947

Adds a property to container resources that models their lifetime.

Shows this value in the resource details view.

Adds an icon to the container name column for persistent containers, with a tooltip.

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6095)

### DIFF
--- a/src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
+++ b/src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
@@ -3,17 +3,37 @@
 @using Aspire.Dashboard.Model
 @using Aspire.Dashboard.Otlp.Model
 @using Aspire.Dashboard.Resources
+
 @inject IStringLocalizer<Columns> Loc
+
+@if (Resource.Properties.TryGetValue(KnownProperties.Container.Lifetime, out var value) &&
+    value.Value.HasStringValue &&
+    StringComparer.Ordinal.Equals(value.Value.StringValue, KnownContainerLifetimes.Persistent))
+{
+    <FluentIcon
+        Icon="Icons.Regular.Size16.Shield"
+        Title="@Loc[nameof(Columns.PersistentContainerIconTooltip)]"
+        Color="Color.Neutral"
+        Class="severity-icon" />
+}
 
 <span title="@FormatName(Resource)"><FluentHighlighter HighlightedText="@FilterText" Text="@FormatName(Resource)"/></span>
 
 @code {
     [Parameter, EditorRequired]
-    public required ResourceViewModel Resource { get; set; }
+    public required ResourceViewModel Resource { get; init; }
 
     [Parameter, EditorRequired]
-    public required Func<ResourceViewModel, string> FormatName { get; set; }
+    public required Func<ResourceViewModel, string> FormatName { get; init; }
 
     [Parameter, EditorRequired]
-    public required string FilterText { get; set; }
+    public required string FilterText { get; init; }
+
+    private static class KnownContainerLifetimes
+    {
+        // Matches members from Aspire.Hosting.ApplicationModel.ContainerLifetime
+
+        public const string Default = nameof(Default);
+        public const string Persistent = nameof(Persistent);
+    }
 }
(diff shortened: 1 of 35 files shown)


src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
```diff
@@ -3,17 +3,37 @@
...
 
     [Parameter, EditorRequired]
-    public required string FilterText { get; set; }
+    public required string FilterText { get; init; }
+
+    private static class KnownContainerLifetimes
+    {
+        // Matches members from Aspire.Hosting.ApplicationModel.ContainerLifetime
+
+        public const string Default = nameof(Default);
```
JamesNK, line 36: If a lifetime isn't set to persistent the UI will show Default. Do we think that is discriptive enough? I could imagine people seeing the value and no knowing what default means.

Maybe show "Session" or "Transient" as the lifetime instead of default.

src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
```diff
@@ -3,17 +3,37 @@
...
 @using Aspire.Dashboard.Resources
+
 @inject IStringLocalizer<Columns> Loc
 
+@if (Resource.Properties.TryGetValue(KnownProperties.Container.Lifetime, out var value) &&
+    value.Value.HasStringValue &&
+    StringComparer.Ordinal.Equals(value.Value.StringValue, KnownContainerLifetimes.Persistent))
+{
+    <FluentIcon
+        Icon="Icons.Regular.Size16.Shield"
```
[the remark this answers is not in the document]
drewnoakes: @mitchdenny suggested a pin too on Teams and I agree it's a better choice. Going to move it to the right side as well.

src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
drewnoakes: Agree that "Default" isn't super helpful. I don't love hardcoding the intepretation of the default value into the dashboard, but we could to the translation in the app host. Either way, the value isn't localised.

Another option is to change the property to boolean `IsPersistent`.

JamesNK: I don't think we have any true/false properties.

Maybe `ContainerLifetime` enum itself should change from Default to Session? @mitchdenny @davidfowl

JamesNK: I started a teams discussion. Doesn't need to block this PR.

JamesNK (approved):

src/Aspire.Dashboard/Resources/Columns.resx
```diff
@@ -189,4 +189,7 @@
   <data name="EndpointsColumnDisplayOverflowTitle" xml:space="preserve">
     <value>Endpoints</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="PersistentContainerIconTooltip" xml:space="preserve">
+    <value>This resource is persistent and won't be stopped when the app host is shutdown.</value>
```
drewnoakes, line 193: This should be 

```suggestion
    <value>This resource is persistent and won't be stopped when the app host is shut down.</value>
```

(xlf files need regenerating, so can't just merge this on github.com)

PR #6096, opened 2024-10-04
Fix bad ordering of resources by name in some situations

JamesNK:
## Description

There is a problem with ordering resource by "Name" in the UI.

Imagine resources called "database" and another is called "database-admin". The full names could end up:
* "database-xyz" and "database-dashboard-xyz", or
* "database-abc" and "database-dashboard-abc"

The order can change depending on what the randomly generated full name is.

One run of dashboard:
image

Another:
image

PR adds a comparer that orders by display name first, then uses the full name if there are multiple instances with the same display name (replicas)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
@@ -310,7 +310,7 @@ public partial class Resources : ComponentBase, IAsyncDisposable
                 continue;
             }
 
-            if (item.DisplayName == resource.DisplayName)
+            if (string.Equals(item.DisplayName, resource.DisplayName, StringComparisons.ResourceName))
             {
                 count++;
                 if (count >= 2)
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
@@ -141,7 +141,7 @@ public partial class Resources : ComponentBase, IAsyncDisposable
 
     private IQueryable<ResourceViewModel>? FilteredResources => _resourceByName.Values.Where(Filter).OrderBy(e => e.ResourceType).ThenBy(e => e, ResourceViewModelNameComparer.Instance).AsQueryable();
 
-    private readonly GridSort<ResourceViewModel> _nameSort = GridSort<ResourceViewModel>.ByAscending(p => p.Name);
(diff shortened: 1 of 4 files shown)


src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -73,6 +73,29 @@ public enum ReadinessState
     Ready
 }
 
+public sealed class ResourceViewModelNameComparer : IComparer<ResourceViewModel>
```
davidfowl, line 76: internal?

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -139,9 +139,9 @@ static bool UnionWithKeys(ConcurrentDictionary<string, bool> left, ConcurrentDic
         }
     }
 
-    private IQueryable<ResourceViewModel>? FilteredResources => _resourceByName.Values.Where(Filter).OrderBy(e => e.ResourceType).ThenBy(e => e.Name).AsQueryable();
+    private IQueryable<ResourceViewModel>? FilteredResources => _resourceByName.Values.Where(Filter).OrderBy(e => e.ResourceType).ThenBy(e => e, ResourceViewModelNameComparer.Instance).AsQueryable();
```
davidfowl, line 142: Do we need to do this anywhere else?

davidfowl (approved):

src/Aspire.Dashboard/Model/ResourceViewModel.cs
JamesNK: Doesn't matter. No one should reference the dashboard

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
JamesNK: Console logs app selector could have be wrong as well. Fixed.

JamesNK:
https://github.com/dotnet/aspire/issues/6099

Issue #6076, opened 2024-10-02
Templates 9.0 follow up work


radical:
Once https://github.com/dotnet/aspire/pull/6058 is merged:

- [x] ~`unsupported hosts - vs version` `Does this need to be bumped higher? Are we doing any testing of Aspire 9 with VS 17.10 or 17.11?` https://github.com/dotnet/aspire/pull/6058#discussion_r1784890643~ **UPDATE** We're happy to leave docs saying 17.12 is required but keep the template constraint set to 17.9
- [x] Update `Microsoft.Extensions.HttpResilience` version in the net9 template to net9 nuget (PR #6419)
- [x] Check if the Web template needs any updates (PR #6124)
- [x] Make sure `https://aka.ms/dotnet/aspire/third-party-notices` is valid, used in the templates
- [x] Update app host templates to use `WaitFor` where applicable
- [x] Update Service Defaults template to configure tracing for an activity source with the application name
- [x] ~#4206~ (moved to 9.1)
- [ ] [tests] Add `new-and-build` tests for `servicedefaults`, and `apphost` templates
- [x] [tests] Add test coverage for new `--aspire-version` template option

cc @DamianEdwards @joperezr

DamianEdwards:
I ensured `https://aka.ms/dotnet/aspire/third-party-notices` works

joperezr:
Marking as High. @DamianEdwards is going to do item 2 which is the only one required for 9

DamianEdwards:
Moving to Backlog now that the 9.0 tasks are done.

PR #6115, opened 2024-10-04
Update starter template to use WaitFor & add application tracing source

DamianEdwards:
This updates the starter template to use the new `WaitFor` feature and register tracing for a source named for the application.

Contributes to #6076

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6115)

### DIFF
--- a/src/Aspire.ProjectTemplates.9.0.net8/templates/aspire-servicedefaults/Extensions.cs
+++ b/src/Aspire.ProjectTemplates.9.0.net8/templates/aspire-servicedefaults/Extensions.cs
@@ -58,7 +58,8 @@ public static class Extensions
             })
             .WithTracing(tracing =>
             {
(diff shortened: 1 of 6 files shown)


radical:
Do we want to add/update any tests to check that the changes work as expected in the generated projects?

radical:
Would it make sense to follow the same convention in the playground apps, or other test apps?

DamianEdwards:
> Do we want to add/update any tests to check that the changes work as expected in the generated projects?

I'm not sure what we'd add honestly. As long as we test WaitFor and we have tests that cover that starter works in general (i.e. the projects startup and load), we should be fine I think.

radical (approved):

PR #6116, opened 2024-10-04
Sanitize bicep indentifiers

### LINKED ISSUE #6074: Need to sanitize bicep names from Aspire resource names
Today, Aspire resource names support hyphen `-` characters.

We take those resource names and use them for bicep variable names. For example:

https://github.com/dotnet/aspire/blob/aa34436aa9d168c3858a5e2fec231f4a746103da/src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs#L43-L49

This doesn't work well because when you use a `-` in the name, you don't get an error until you try to `azd up`.

```C#
var storage = builder.AddAzureStorage("my-storage").RunAsEmulator(container =>
{
    container.WithDataBindMount();
});

var blobs = storage.AddBlobs("blobs");

builder.AddProject<Projects.AzureStorageEndToEnd_ApiService>("api")
       .WithExternalHttpEndpoints()
       .WithReference(blobs);
```

When you `azd up`, it produces:

```
Packaging services (azd package)


Provisioning Azure resources (azd provision)
Provisioning Azure resources can take some time.


ERROR: error executing step command 'provision': initializing provisioning manager: failed to compile bicep template: failed running bicep build: exit code: 1, stdout: , stderr: C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(2,7) : Warning no-unused-params: Parameter "location" is declared but never used. [https://aka.ms/bicep/linter/no-unused-params]
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(4,7) : Warning no-unused-params: Parameter "principalId" is declared but never used. [https://aka.ms/bicep/linter/no-unused-params]
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(6,7) : Warning no-unused-params: Parameter "principalType" is declared but never used. [https://aka.ms/bicep/linter/no-unused-params]
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(8,10) : Error BCP028: Identifier "my" is declared multiple times. Remove or rename the duplicates.
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(8,12) : Error BCP068: Expected a resource type string. Specify a valid resource type of format "<types>@<apiVersion>".
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(8,12) : Error BCP029: The resource type is not valid. Specify a valid resource type of format "<types>@<apiVersion>".
C:\Users\eerhardt\AppData\Local\Temp\azd-infra2898406436\my-storage\my-storage.module.bicep(8,71) : Error BCP018: Expected the "=" character at this location.
```

We need to sanitize these names, probably by replacing `-` with `_` (which is safe to do because bicep supports underscore, and Aspire doesn't support underscore).

cc @davidfowl @mitchdenny @tg-msft

eerhardt:
## Description

When we create resources in bicep, we need to ensure the identifier name we use is valid. One common complication is that Aspire resources only allow letters, numbers and a hyphen. But bicep doesn't allow hyphen, but instead an underscore.

Add a new convenience API to get a bicep identifier from an IAzureResource.

Fix https://github.com/dotnet/aspire/issues/6074

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] No
  - [x] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6116)

### DIFF
--- a/src/Aspire.Hosting.Azure/AzureResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure/AzureResourceExtensions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Azure.Utils;
 
 namespace Aspire.Hosting;
 
@@ -22,4 +23,21 @@ public static class AzureResourceExtensions
         ParameterResourceBuilderExtensions.ConfigureConnectionStringManifestPublisher((IResourceBuilder<IResourceWithConnectionString>)builder);
         return builder;
     }
+
+    /// <summary>
+    /// Gets the Bicep identifier for the Azure resource.
+    /// </summary>
+    /// <param name="resource">The Azure resource.</param>
+    /// <returns>A valid Bicep identifier.</returns>
+    public static string GetBicepIdentifier(this IAzureResource resource) =>
+        NormalizeBicepIdentifier(resource.Name);
+
+    /// <summary>
+    /// Normalizes the given identifier name to make it a valid Bicep identifier name.
+    /// </summary>
+    /// <param name="identifierName">The identifier name to normalize.</param>
+    /// <returns>A valid Bicep identifier.</returns>
+    // TODO: This API should be deleted when we get the API from Azure.Provisioning
+    public static string NormalizeBicepIdentifier(string identifierName) =>
+        BicepIdentifierHelpers.Normalize(identifierName);
 }
--- a/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/openai.module.bicep
+++ b/playground/OpenAIEndToEnd/OpenAIEndToEnd.AppHost/openai.module.bicep
@@ -32,7 +32,7 @@ resource openai_CognitiveServicesOpenAIContributor 'Microsoft.Authorization/role
   scope: openai
 }
 
-resource gpt-4o 'Microsoft.CognitiveServices/accounts/deployments@2024-10-01' = {
+resource gpt_4o 'Microsoft.CognitiveServices/accounts/deployments@2024-10-01' = {
   name: 'gpt-4o'
   properties: {
     model: {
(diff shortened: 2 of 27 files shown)


src/Aspire.Hosting.Azure/AzureResourceExtensions.cs
```diff
@@ -22,4 +23,21 @@ public static IResourceBuilder<T> PublishAsConnectionString<T>(this IResourceBui
...
+    public static string GetBicepIdentifier(this IAzureResource resource) =>
+        NormalizeBicepIdentifier(resource.Name);
+
+    /// <summary>
+    /// Normalizes the given identifier name to make it a valid Bicep identifier name.
+    /// </summary>
+    /// <param name="identifierName">The identifier name to normalize.</param>
+    /// <returns>A valid Bicep identifier.</returns>
+    // TODO: This API should be deleted when we get the API from Azure.Provisioning
+    public static string NormalizeBicepIdentifier(string identifierName) =>
```
eerhardt, line 41: @tg-msft is going to add an API for this in Azure.Provisioning. I'm just putting this API here for now until that API comes in.

There are problems with using shared source files and internal classes. The unit test references them all and fails to compile since it has IVT to all the libraries and so it sees multiple copies. This was easiest for now.

tg-msft (approved):

src/Aspire.Hosting.Azure/AzureResourceExtensions.cs
davidfowl: The fact that we can’t do this in the provisioning context is kinda depressing

eerhardt: > The fact that we can’t do this in the provisioning context is kinda depressing

How/when would you expect that to happen? On insertion? All at the end?

There are places where we reference these names, or find by these names, where the names need to line up.

PR #6124, opened 2024-10-05
Update starter template from .NET 9 app template updates

DamianEdwards:
This updates the starter template with changes from the relevant ASP.NET Core 9 templates.

- net8 version gets just a Bootstrap update & some accessibility fixes
- net9 version gets all changes

Contributes to #6076 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6124)


eng/Versions.props
```diff
@@ -69,5 +69,6 @@
     <!-- for templates -->
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet8>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet8>
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
+    <MicrosoftAspNetCorePackageVersionForNet9>9.0.0-rc.2.24470.14</MicrosoftAspNetCorePackageVersionForNet9>
```
eerhardt, line 72: How does this version get updated? Is it automated?

DamianEdwards: Nope. We'll have to bump it to match depending on our release schedule, etc.

DamianEdwards: Are there any other packages we're referencing where the version is tracking .NET 9 releases?

eerhardt: I think https://github.com/dotnet/aspire/pull/5932 is the only one I know of.

cc @sebastienros

eerhardt:
>                 Endpoints: ["http://localhost:\\d+", "https://localhost:\\d+"]),

Why are we changing this test? Shouldn't we still be hitting this endpoint?

---
Refers to: tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs:122 in f3d49cf. [](commit_id = f3d49cf3a263d587411945eab6fe08207124f977, deletion_comment = False)

DamianEdwards:
> Why are we changing this test? Shouldn't we still be hitting this endpoint?

Because the launchUrl was changed and the test was failing as it was attempting to verify the endpoints showed up in the dashboard.

eerhardt:
> > Why are we changing this test? Shouldn't we still be hitting this endpoint?
> 
> Because the launchUrl was changed and the test was failing as it was attempting to verify the endpoints showed up in the dashboard.

@radical - shouldn't the test be hitting this endpoint and ensure it "works" (returns a success status code)?

radical:
> > > Why are we changing this test? Shouldn't we still be hitting this endpoint?
> > 
> > 
> > Because the launchUrl was changed and the test was failing as it was attempting to verify the endpoints showed up in the dashboard.
> 
> @radical - should the test be hitting this endpoint and ensure it "works" (returns a success status code)? Or does that happen elsewhere?

The endpoint tried is what it finds on the dashboard, and checks that it is the `weather` page. For the redis cache case, this method checks the returned results too.

https://github.com/dotnet/aspire/blob/3a341a20a93322941c1a5908103a1f327e076a54/tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs#L52-L111

radical:
We'll need to add a check for the `/weatherforecast` endpoint explicitly.

radical:
I think we should test the 8.0 starter template too, so it doesn't regress.

DamianEdwards:
>We'll need to add a check for the /weatherforecast endpoint explicitly.
OK. Where would that go? Another test method on `StarterTemplateRunTestsBase<T>` ?

radical:
> > We'll need to add a check for the /weatherforecast endpoint explicitly.
> 
> @radical OK. Where would that go? Another test method on `StarterTemplateRunTestsBase<T>` ?

Since this would be just checking that the endpoint exists, I would add it to `AssertStarterTemplateRunAsync` in `WorkloadTestsBase`.

DamianEdwards:
> Since this would be just checking that the endpoint exists

@radical not quite following. Wouldn't we want it to send a request and verify the result? What do you mean by "checking the endpoint exists"?

radical:
> > Since this would be just checking that the endpoint exists
> 
> @radical not quite following. Wouldn't we want it to send a request and verify the result? What do you mean by "checking the endpoint exists"?

Sure, that makes sense. You can add a method to that class, and call from `AssertStarterTemplateRunAsync` so it gets used by any test checking the starter template.

radical:
> I think we should test the 8.0 starter template too, so it doesn't regress.

Using the `DefaultTargetFramework` thing that I mentioned in the other comment, you can skip this part, as it will get covered in running all the tests with tfm=net8.0 .

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
```diff
@@ -112,22 +140,17 @@ static async Task<List<string[]>> GetAndValidateCellTexts(ILocator tableLoc)
...
-                Type: "Project",
+            new(Type: "Project",
                 Name: "apiservice",
                 State: "Running",
                 Source: $"{project.Id}.ApiService.csproj",
-                Endpoints: ["http://localhost:\\d+", "https://localhost:\\d+"]),
-
-            new ResourceRow(
-                Type: "Project",
-                Name: "webfrontend",
```
radical, line 126: The test will fail because it won't recognize the `webfrontend` resource that it would find on the dashboard.

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
```diff
@@ -112,22 +140,17 @@ static async Task<List<string[]>> GetAndValidateCellTexts(ILocator tableLoc)
...
-                State: "Running",
-                Source: $"{project.Id}.Web.csproj",
-                Endpoints: ["https://localhost:\\d+", "http://localhost:\\d+"])
+                Endpoints: BuildEnvironment.DefaultTargetFramework == TestTargetFramework.Current
+                    ? ["http://localhost:\\d+", "https://localhost:\\d+"]
+                    : ["http://localhost:\\d+/weatherforecast", "https://localhost:\\d+/weatherforecast"])
```
radical, line 151: Also, I just realized that we should change these to:
```csharp
                    ? ["^http://localhost:\\d+$", "^https://localhost:\\d+$"]
                    : ["^http://localhost:\\d+/weatherforecast$", "^https://localhost:\\d+/weatherforecast$"]
```

tests/Aspire.Workload.Tests/StarterTemplateRunTestsBase.cs
DamianEdwards: I might have to sync up with you in a call. Sorry, first time seeing these tests and I'm struggling following how they're all structured.

radical: I can fix the tests a bit for you, if you want.

radical:
I have added tfm=net8.0 variants of the starter template tests. I think the templates that explicitly differ should get checked on the PRs to, so I added them here, instead of depending on an OOB test run.

DamianEdwards:
@radical yay this is passing! Now we can focus on the content getting approved :smile:

eng/Versions.props
```diff
@@ -72,6 +72,7 @@
     <!-- for templates -->
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet8>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet8>
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
+    <MicrosoftAspNetCorePackageVersionForNet9>9.0.0-rc.2.24474.3</MicrosoftAspNetCorePackageVersionForNet9>
```
joperezr, line 75: In order for this version to flow or get updated automatically, We should probably use an actual package here that ships from aspnetcore so that each time we update package versions it will get associated to a real package and get bumped. Can't we reuse any package that we already depend on?

src/Aspire.ProjectTemplates/templates/aspire-starter-net8/Aspire-StarterApplication.1.ApiService/Properties/launchSettings.json
```diff
@@ -4,8 +4,7 @@
     "http": {
       "commandName": "Project",
       "dotnetRunMessages": true,
-      "launchBrowser": true,
-      "launchUrl": "weatherforecast",
+      "launchBrowser": false,
```
joperezr, line 7: Is this what we have now in our .NET 8 templates?

src/Aspire.ProjectTemplates.net8/templates/aspire-starter/Aspire-StarterApplication.1.Web/wwwroot/lib/bootstrap/dist/css/bootstrap-grid.css
```diff
@@ -0,0 +1,4085 @@
+/*!
```
joperezr, line 1: I think all of the files in this directory are not based on main yet and should all be removed. We no longer have a Aspire.ProjectTemplates.net8 project :smiley:

src/Aspire.ProjectTemplates/templates/aspire-starter-net8/Aspire-StarterApplication.1.ApiService/Properties/launchSettings.json
DamianEdwards: This is what the ASP.NET Core 9 templates have now, yes. I figured it's best to align rather than do something different.

src/Aspire.ProjectTemplates.net8/templates/aspire-starter/Aspire-StarterApplication.1.Web/wwwroot/lib/bootstrap/dist/css/bootstrap-grid.css
DamianEdwards: Ah they all moved under the one templates project right? Let me move these files.

joperezr (approved):
Changes look good, other than the two comments I left. Other than that this seems like good to go, thanks @DamianEdwards

eng/Versions.props
DamianEdwards: I don't see any others in this file.

DamianEdwards: Chatted offline, this is the only ASP.NET Core 9 ref we have and we'll revisit automating updating of this later on.

DamianEdwards:
Looks like a Helix timeout for the Linux leg 😞

eng/Versions.props
```diff
@@ -72,6 +72,7 @@
     <!-- for templates -->
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet8>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet8>
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
```
eerhardt, line 74: What's the story with `MicrosoftExtensionsHttpResiliencePackageVersionForNet9`? It has the same version as "ForNet8". Shouldn't these 2 versions be separate, so when I get a .NET 9 app, I get the 9.0 Resilience package?

DamianEdwards: Yep, it's listed in #6076 as a TODO

DamianEdwards: I'm going to address the remaining content items in #6076 as part of the next PR which refactors the templates down to a single template you can select TFM and Aspire version from.

PR #6135, opened 2024-10-05
Improve the experience when docker isn't running

### LINKED ISSUE #5533: Throw a friendly error from Aspire if Docker isn't running
This is what a user sees today if they run Aspire and docker isn't running.

image

Should display a nice, friendly error message. Something like, "Failed to connect to Docker. Ensure that Docker is running."

davidfowl:
## Description

- When the container runtime isn't running, log a critical error but continue running and sending requests to dcp. DCP will wait for the container runtime to be healthy (started) and will run the containers.

Fixes https://github.com/dotnet/aspire/issues/5533

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6135)

### DIFF
--- a/playground/Redis/Redis.AppHost/Program.cs
+++ b/playground/Redis/Redis.AppHost/Program.cs
@@ -16,4 +16,14 @@ builder.AddProject<Projects.Redis_ApiService>("apiservice")
     .WithReference(garnet).WaitFor(garnet)
     .WithReference(valkey).WaitFor(valkey);
 
+#if !SKIP_DASHBOARD_REFERENCE
+// This project is only added in playground projects to support development/debugging
+// of the dashboard. It is not required in end developer code. Comment out this code
+// or build with `/p:SkipDashboardReference=true`, to test end developer
+// dashboard launch experience, Refer to Directory.Build.props for the path to
+// the dashboard binary (defaults to the Aspire.Dashboard bin output in the
+// artifacts dir).
+builder.AddProject<Projects.Aspire_Dashboard>(KnownResourceNames.AspireDashboard);
+#endif
+
 builder.Build().Run();
--- a/playground/Redis/Redis.AppHost/Redis.AppHost.csproj
+++ b/playground/Redis/Redis.AppHost/Redis.AppHost.csproj
@@ -9,6 +9,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\KnownResourceNames.cs" Link="KnownResourceNames.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
     <AspireProjectOrPackageReference Include="Aspire.Hosting.AppHost" />
     <AspireProjectOrPackageReference Include="Aspire.Hosting.Redis" />
     <AspireProjectOrPackageReference Include="Aspire.Hosting.Garnet" />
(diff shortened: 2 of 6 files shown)


davidfowl:
When you launch the app host, this is what it says:

image

This is what the dashboard looks like when the container runtime isn't running.

image

We show logs in each of the containers that show the error message from DCP:

image

@danegsta I think dcp should set the state of these containers to something that reflects that. I can do it in the apphost but it feels like the source of truth for what the container state is should be dcp. We could use a new state "Error" or "Container runtime unhealthy" etc.

Here's what it looks like when you run docker:

https://github.com/user-attachments/assets/b72c11bc-ad51-48d2-9d05-080cc7034067

danegsta:
First thought is that we could add something like MissingRequirement as the new state and a status property for the specific related error to make it easier to surface in the Dashboard.

davidfowl:
> First thought is that we could add something like MissingRequirement as the new state and a status property for the specific related error to make it easier to surface in the Dashboard.

I think we should do something like that. But we should make it specific to the actual state instead of a generic `MissingRequirement` state (unless there's some other way to show the specific reason).

mitchdenny:
I think that @danegsta is right about having a `MissingRequirements` state. The reason is that this PR will need to include an update to `KnownResourceStates.TerminalStates` to account for getting into this state. This is used by `WaitForCompletion(...)` (ultimately).

mitchdenny (approved):

JamesNK:
This is what I see in the UI:

image

And this is in the console logs:

image

Overall I like it.

Suggestions for improvement:
1. There is duplication between error messages printed to app host console and resource console. Maybe share code? I like that the app host console includes the name of the runtime, and I like that the console logs includes the error message from DCP. One set of code that does both?
2. Putting myself in the shoes of an Aspire beginner, I don't know what unhealthy means. Do I need to reinstall Docker? IMO if the container runtime is unhealth we could always suggest restarting it. Or we could look at the error message from DCP and if there is matching content, log in the UI that the container runtime isn't running.

>Get "[http://%2F%2F.%2Fpipe%2FdockerDesktopLinuxEngine/v1.47/containers/json?limit=1"](http://%2F%2F.%2Fpipe%2Fdockerdesktoplinuxengine/v1.47/containers/json?limit=1%22);: open //./pipe/dockerDesktopLinuxEngine: The system cannot find the file specified.

Look for "open //./pipe/dockerDesktopLinuxEngine"? It might be better than "The system cannot find the file specified." which could be localized.
Would need to do something similar for podman which no doubt has a different address.

davidfowl:
I moved the code to a shared location, and I improved the error message as much as I could (using chatgpt 😄 )

JamesNK (approved):

maddymontaquila:
yessssss 🔥

PR #6145, opened 2024-10-07
Add IResourceWithWaitSupport.

### LINKED ISSUE #5997: Custom resource with `WaitFor` behaves very differently if it implements `IResourceWithParent`

### Describe the bug

If I have a custom `IResourceWithParent`, it seems to get set to `Ready` once it's parent is ready, ignoring any other `WaitFor` I've specified on the resource, and even though like the other custom resource I don't have any other code indicating it's initial state.

image

### Expected Behavior

Given that `custom1 ` has a `WaitFor(migrator)` and `migrator` is still waiting, I'd expect `custom1` to still be in the Waiting state.  

Also since it's a custom resource and I haven't implemented anything to mark the resource as ready, I'd expect the state after it's finished Waiting to be `Unknown` like `custom2` rather than `Ready`.  (Or for `custom2` to default it's initial state to `Ready` like `custom1` does.

I can "fix" these issues by not implementing `IResourceWithParent` on `MigratedDatabase`, but it seemed surprising to me that the presence of this caused such different behaviour.

### Steps To Reproduce

```cs
var sql = builder.AddSqlServer("sql");
var migrator = builder.AddExecutable("migrator", "pwsh", ".",
    "-NoProfile",
    "-NoLogo",
    "-NonInteractive",
    "-Command",
    "Write-Host 'Starting Migration'; Start-Sleep -Seconds 10; Write-Host 'Done'")
    .WaitFor(sql);

var db = sql.AddDatabase("whatever")
    .WaitFor(migrator);

var custom1 = builder.AddResource(new MigratedDatabase("custom1", db))
    .WaitFor(migrator);

var custom2 = builder.AddResource(new CustomResource("custom2"))
    .WaitFor(migrator);

public class MigratedDatabase(string name, IResourceBuilder<IResourceWithConnectionString> inner)
    : Resource(name), IResourceWithParent<IResourceWithConnectionString>, IResourceWithConnectionString
{
    public IResourceWithConnectionString Parent => inner.Resource;
    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
}


public class CustomResource(string name) : Resource(name)
{
}
```

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

This example is based on how I've done DB migrations in 8.x with https://github.com/davidfowl/WaitForDependenciesAspire.

mitchdenny:
## Description

This change constrains `WaitFor` and `WaitForCompletition` to resources which implement `IResourceWithWaitSupport`. The reason for this change is that not all resources contain the backend logic to block their initialization based on other resources. A good example of this is child resources of databases. The way these resources currently work is that they adopt their parents state and don't implement any blocking logic.

In future releases we hope to be able to add this interface event to child resources on a case by case basis but for now this constraint helps avoid potential confusion that exposing a non-functional `WaitFor` method call on child resources might cause.

Fixes #5997

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes (adds a constraint so existing tests covered)
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No (discussed with @davidfowl)
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6145)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ContainerResource.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ContainerResource.cs
@@ -8,7 +8,7 @@ namespace Aspire.Hosting.ApplicationModel;
 /// </summary>
 /// <param name="name">The name of the resource.</param>
 /// <param name="entrypoint">An optional container entrypoint.</param>
-public class ContainerResource(string name, string? entrypoint = null) : Resource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithEndpoints
+public class ContainerResource(string name, string? entrypoint = null) : Resource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithEndpoints, IResourceWithWaitSupport
 {
     /// <summary>
     /// The container Entrypoint.
--- a/src/Aspire.Hosting/ApplicationModel/ExecutableResource.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ExecutableResource.cs
@@ -9,7 +9,7 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <param name="name">The name of the resource.</param>
 /// <param name="command">The command to execute.</param>
 /// <param name="workingDirectory">The working directory of the executable.</param>
-public class ExecutableResource(string name, string command, string workingDirectory) : Resource(name), IResourceWithEnvironment, IResourceWithArgs, IResourceWithEndpoints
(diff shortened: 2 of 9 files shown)


davidfowl:
None of the azure resources?

davidfowl (approved):

PR #6164, opened 2024-10-07
Share Annotations with inner resources in Azure PostgreSQL, Redis, and SQL Server resources.

### LINKED ISSUE #6146: AddAzurePostgresFlexibleServer(..).RunAsContainer(...).AddDatabase(...) results in missing annotations on child resources
When using the new `AddAzurePostgresFlexibleServer(...)` method with `RunAsContainer(...)` the annotations from child resources are not copied to the internally instanced `PostgresDatabaseResource` instances. This would result in some mechanisms like WithCommand(...) not working.

Picked it up whilst looking to support `WaitFor` with child resources.

eerhardt:
This allows for things like WaitFor to work correctly when using RunAsContainer.

Fix #6146
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6164)


src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -289,6 +289,11 @@ public static IResourceBuilder<AzurePostgresFlexibleServerResource> RunAsContain
         }
 
         var azureResource = builder.Resource;
+        var azureDatabases = builder.ApplicationBuilder.Resources
+            .OfType<AzurePostgresFlexibleServerDatabaseResource>()
+            .Where(db => db.Parent == azureResource)
+            .ToDictionary(db => db.Name);
+
         RemoveAzureResources(builder.ApplicationBuilder, azureResource);
```
mitchdenny, line 297: Not a big issue, but this results in scanning the resource collection twice to get the same set of data from above. Why not just iterate over the resources inline or do the deletes in the loop below.

tests/Aspire.Hosting.Azure.Tests/AzurePostgresExtensionsTests.cs
```diff
@@ -323,4 +323,52 @@ public async Task WithPasswordAuthenticationBeforeAfterRunAsContainer(bool befor
         Assert.Equal("Host=localhost;Port=12455;Username=user;Password=p@ssw0rd1;Database=db1", await db1.Resource.ConnectionStringExpression.GetValueAsync(CancellationToken.None));
         Assert.Equal("Host=localhost;Port=12455;Username=user;Password=p@ssw0rd1;Database=db2Name", await db2.Resource.ConnectionStringExpression.GetValueAsync(CancellationToken.None));
     }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void RunAsContainerAppliesAnnotationsCorrectly(bool before)
```
mitchdenny, line 330: What happens if I do this?

```csharp
var db = builder.AddAzurePostgresFlexibleServer("pg")
                .WithPasswordAuthentication()
                .RunAsContainer(c =>
                {
                    c.WithPgAdmin();
                })
                .AddDatabase("db")
                .WithAnnotation(new DummyAnnotation());
```

Based on what I'm seeing in this branch the DummyAnnotation will be lost by the time this gets down into the application executor.

mitchdenny:
I just realized that you are going to need to update `AddDatabase(...)` as well to copy annotations if the parent resource has an InnerResource.

https://github.com/dotnet/aspire/blob/770b3dfeeab32afc5403d4cee5131be723faf2c0/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs#L256

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
eerhardt: I can pass the dictionary into the Remove method.

tests/Aspire.Hosting.Azure.Tests/AzurePostgresExtensionsTests.cs
eerhardt: Good catch! I've fix this and added tests for it.

eerhardt:
> I just realized that you are going to need to update `AddDatabase(...)` as well to copy annotations if the parent resource has an InnerResource.
> 
> https://github.com/dotnet/aspire/blob/770b3dfeeab32afc5403d4cee5131be723faf2c0/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs#L256

Yep, that was the fix for https://github.com/dotnet/aspire/pull/6164#discussion_r1791031810.

mitchdenny (approved):
LGTM ... verified that commands annotations carry across.

Issue #6163, opened 2024-10-07
[tests] MSBuild crashes randomly on test runs `error MSB4166: Child node "2" exited prematurely. Shutting down.`


radical:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=829385
Build error leg or test failing: Aspire.Workload.Tests.StarterTemplateProjectNamesTests.Aspire.Workload.Tests.StarterTemplateProjectNamesTests.StarterTemplateWithTest_ProjectNames
Pull request: N/A

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "MSBUILD : error MSB4166: Child node.* exited prematurely.",
  "BuildRetry": true,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=829385
**Error message validated:** `[MSBUILD : error MSB4166: Child node.* exited prematurely.`]
**Result validation:** :white_check_mark: Known issue matched with the provided build.
**Validation performed at:** 10/7/2024 10:16:53 PM UTC

#### Summary
|24-Hour Hit Count|7-Day Hit Count|1-Month Count|
|---|---|---|
|0|0|0|

joperezr:
@radical is there an easy way to check if this is still an issue? seems like it hasn't get linked to for a while now.

radical:
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/1472?v=4&amp;size=80">@radical](https://github.com/radical) is there an easy way to check if this is still an issue? seems like it hasn't get linked to for a while now.

Closing this for now as it hasn't been hit in quite some time.

PR #6181, opened 2024-10-08
Retry tests if msbuild crashes

radical:
And get the msbuild debug files from crashes to aid debugging.

Issue: https://github.com/dotnet/aspire/issues/6163

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6181)

### DIFF
--- a/eng/test-configuration.json
+++ b/eng/test-configuration.json
@@ -4,6 +4,7 @@
   "localRerunCount": 1,
   "retryOnRules": [
     { "testName": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Tests.ConformanceTests.HealthCheckReportsExpectedStatus" } },
-    { "testAssembly": { "contains": "Aspire.Elastic.Clients.Elasticsearch.Tests" }, "failureMessage": { "contains": "RemoteExecutionException : Half-way through waiting for remote process." } }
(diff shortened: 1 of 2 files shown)


joperezr:
Do we understand why the crash is happening though? How confident are we that when in a failure state, retrying will make this pass?

radical:
It doesn't crash like this on every build, but it does crash sometimes. So, this is just to ease the pain of having to do retries or having red builds on internal pipeline. Meanwhile, we are tracking the occurrence of these with https://github.com/dotnet/aspire/issues/6163 .

joperezr (approved):
LGTM, thanks @radical. If there is a way of not only retrying, but also keeping trace of when that needed to happen it would be great, as that way we won't get false-negatives indicating that this is no longer a problem, but that shouldn't block this.

radical:
> LGTM, thanks @radical. If there is a way of not only retrying, but also keeping trace of when that needed to happen it would be great, as that way we won't get false-negatives indicating that this is no longer a problem, but that shouldn't block this.

I'm setting `MSBUILDDEBUGPATH` here, so we should get some debug logs from msbuild which we can use to further debug this.

PR #6183, opened 2024-10-08
Initial integration of codespace URL rewriting logic into hosting.

### LINKED ISSUE #1178: Aspire failing in GitHub Codespace
Hi, Team.

When the current AppHost is running in GitHub Codespaces (and possibly in devcontainer), I found a few issues:
 
1. **The dashboard page is not automatically opening** - (Related to #655 ) Either Blazor app or API app automatically opens a new tab when I run it individually. But, the dashboard doesn't automatically open a new page. Instead I have to manually find the page and open it through the port forwarding tab.

- [x] #1296

    ```json
    "applicationUrl": "http://localhost:<random_port_number>",
    "environmentVariables": {
      "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "http://localhost:<random_port_number>"
    }
    ```

   I initially expected that the `applicationUrl` attribute will be used to launch the dashboard page, but it uses `http://localhost:18888`. I'm not sure what these `applicationUrl` and `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL` are for.

- [x] #1297

- [x] #1298

mitchdenny:
## Description

Adds support for GitHub Codespaces by automatically rewriting the URLs in the dashboard to point to the GitHub Codespaces proxied events.

Fixes #1178 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6183)

### DIFF
--- a/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
+++ b/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
@@ -36,10 +36,7 @@ internal class CodespacesUrlRewriter(ILogger<CodespacesUrlRewriter> logger, ICon
 
                     if (!originalUrlSnapshot.IsInternal && (uri.Scheme == "http" || uri.Scheme == "https") && uri.Host == "localhost")
                     {
-                        if (remappedUrls is null)
-                        {
-                            remappedUrls = new();
-                        }
+                        remappedUrls ??= new();
 
                         var newUrlSnapshot = originalUrlSnapshot with
                         {
--- a/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
+++ b/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
@@ -11,16 +11,26 @@ namespace Aspire.Hosting.Codespaces;
 
 internal class CodespacesUrlRewriter(ILogger<CodespacesUrlRewriter> logger, IConfiguration configuration, ResourceNotificationService resourceNotificationService) : BackgroundService
 {
+    private const string CodespacesEnvironmentVariable = "CODESPACES";
+    private const string CodespaceNameEnvironmentVariable = "CODESPACE_NAME";
+    private const string GitHubCodespacesPortForwardingDomain = "GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN";
+
+    private string GetRequiredCodespacesConfigurationValue(string key)
+    {
+        ArgumentNullException.ThrowIfNullOrEmpty(key);
+        return configuration.GetValue<string>(key) ?? throw new DistributedApplicationException($"Codespaces was detected but {key} environment missing.");
+    }
+
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
-        if (!configuration.GetValue<bool>("CODESPACES", false))
+        if (!configuration.GetValue<bool>(CodespacesEnvironmentVariable, false))
         {
             logger.LogTrace("Not running in Codespaces, skipping URL rewriting.");
             return;
         }
 
-        var gitHubCodespacesPortForwardingDomain = configuration.GetValue<string>("GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN") ?? throw new DistributedApplicationException("Codespaces was detected but GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN environment missing.");
-        var codespaceName = configuration.GetValue<string>("CODESPACE_NAME") ?? throw new DistributedApplicationException("Codespaces was detected but CODESPACE_NAME environment missing.");
+        var gitHubCodespacesPortForwardingDomain = GetRequiredCodespacesConfigurationValue(GitHubCodespacesPortForwardingDomain);
+        var codespaceName = GetRequiredCodespacesConfigurationValue(CodespaceNameEnvironmentVariable);
 
         do
         {
--- a/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
+++ b/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
@@ -1,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Immutable;
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Hosting.Codespaces;
+
+internal class CodespacesUrlRewriter(IConfiguration configuration, ResourceNotificationService resourceNotificationService) : BackgroundService
+{
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        var gitHubCodespacesPortForwardingDomain = configuration.GetValue<string>("GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN") ?? throw new DistributedApplicationException("Codespaces was detected but GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN environment missing.");
+        var codespaceName = configuration.GetValue<string>("CODESPACE_NAME") ?? throw new DistributedApplicationException("Codespaces was detected but CODESPACE_NAME environment missing.");
+
+        do
+        {
+            var resourceEvents = resourceNotificationService.WatchAsync(stoppingToken);
+
+            await foreach (var resourceEvent in resourceEvents.ConfigureAwait(false))
+            {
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,69 @@
...
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        if (!configuration.GetValue<bool>("CODESPACES", false))
+        {
+            logger.LogTrace("Not running in Codespaces, skipping URL rewriting.");
+            return;
+        }
+
+        var gitHubCodespacesPortForwardingDomain = configuration.GetValue<string>("GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN") ?? throw new DistributedApplicationException("Codespaces was detected but GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN environment missing.");
+        var codespaceName = configuration.GetValue<string>("CODESPACE_NAME") ?? throw new DistributedApplicationException("Codespaces was detected but CODESPACE_NAME environment missing.");
```
JamesNK, line 23: Add a helper method for getting these values that removes duplication of keys and error messages.

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,69 @@
...
+                foreach (var originalUrlSnapshot in resourceEvent.Snapshot.Urls)
+                {
+                    var uri = new Uri(originalUrlSnapshot.Url);
+
+                    if (!originalUrlSnapshot.IsInternal && (uri.Scheme == "http" || uri.Scheme == "https") && uri.Host == "localhost")
+                    {
+                        if (remappedUrls is null)
+                        {
+                            remappedUrls = new();
+                        }
```
JamesNK, line 42:
```suggestion
                        remappedUrls ??= new();
```

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                        var newUrlSnapshot = originalUrlSnapshot with
+                        {
+                            Url = $"{uri.Scheme}://{codespaceName}-{uri.Port}.{gitHubCodespacesPortForwardingDomain}{uri.AbsolutePath}"
```
JamesNK, line 46: Add a comment or link that explains this url

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            await foreach (var resourceEvent in resourceEvents.ConfigureAwait(false))
+            {
+                Dictionary<UrlSnapshot, UrlSnapshot>? remappedUrls = null;
+
+                foreach (var originalUrlSnapshot in resourceEvent.Snapshot.Urls)
+                {
+                    var uri = new Uri(originalUrlSnapshot.Url);
+
+                    if (!originalUrlSnapshot.IsInternal && (uri.Scheme == "http" || uri.Scheme == "https") && uri.Host == "localhost")
```
JamesNK, line 37: Why these values

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,69 @@
...
+                    await resourceNotificationService.PublishUpdateAsync(resourceEvent.Resource, resourceEvent.ResourceId, s => s with
+                    {
+                        Urls = transformedUrls.ToImmutableArray()
+                    }).ConfigureAwait(false);
+                }
+            }
+
+            // Short delay if we crash just to avoid spinning CPU.
+            await Task.Delay(5000, stoppingToken).ConfigureAwait(false);
+        } while (!stoppingToken.IsCancellationRequested);
```
JamesNK, line 67: What are you doing here? If `WatchAsync` errors then won't it throw an error right out of this method? Also, there should be a log message if there was an error and you need to recover.

Add a test that covers this situation to verify it works.

davidfowl:
I assume this is also post 9.0

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
mitchdenny: Honestly I don't know what `IsInternal' really means - do you?

As for only exposing `http` and `https` schemed endpoints - from what I could tell that is the only option it supported. It doesn't seem to support anything other than HTTP.

mitchdenny:
> I assume this is also post 9.0

I think so. This change is about the app host changes required to support codespaces better. But there are still some gotchas around the codespace devcontainer itself which I don't think are fully fleshed out.

For example you need a devcontainer which supports Docker-in-Docker (ideally) which the default universal one does. But then you want to install the C# DevKit. I don't know if we've got options to streamline this experience for customers.

mitchdenny:
One option would be to have a GitHub template with the DevContainer pre-configured with the right decontainer file and the recommended VSCode extensions and then the workflow would be that people create their repo based on that template, and then do dotnet new themselves.

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,89 @@
...
+        if (!configuration.GetValue<bool>(CodespacesEnvironmentVariable, false))
+        {
+            logger.LogTrace("Not running in Codespaces, skipping URL rewriting.");
+            return;
+        }
+
+        var gitHubCodespacesPortForwardingDomain = GetRequiredCodespacesConfigurationValue(GitHubCodespacesPortForwardingDomain);
+        var codespaceName = GetRequiredCodespacesConfigurationValue(CodespaceNameEnvironmentVariable);
+
+        do
```
JamesNK, line 35: Maybe add a comment that this runs for the lifetime of the host so it can react to resources as they start and new resources being added?

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using System.Collections.Immutable;
+using Aspire.Hosting.ApplicationModel;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Aspire.Hosting.Codespaces;
+
+internal class CodespacesUrlRewriter(ILogger<CodespacesUrlRewriter> logger, IConfiguration configuration, ResourceNotificationService resourceNotificationService) : BackgroundService
```
JamesNK, line 12: nit
```suggestion
internal sealed class CodespacesUrlRewriter(ILogger<CodespacesUrlRewriter> logger, IConfiguration configuration, ResourceNotificationService resourceNotificationService) : BackgroundService
```

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -0,0 +1,89 @@
...
+                        await resourceNotificationService.PublishUpdateAsync(resourceEvent.Resource, resourceEvent.ResourceId, s => s with
+                        {
+                            Urls = transformedUrls.ToImmutableArray()
+                        }).ConfigureAwait(false);
+                    }
+                }
+            }
+            catch (Exception ex) when (!stoppingToken.IsCancellationRequested)
+            {
+                // When debugging sometimes we'll get cancelled here but we don't want
```
JamesNK, line 82: The delay was removed. Intentional?

JamesNK (approved):

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
mitchdenny: Yeah, I looked at how this was done in other areas and the delay isn't really necessary. The Watch is a natural throttling issue. The reason why we catch and retry is mostly for debuggability. I've found that in certain cases if I am debugging these loops the cancellation can occur and the background service stops.

When I tracked the cancellation back at one point it appeared to be the dashboard hanging up after a timeout. In those cases we just want to restart watching.

PR #6194, opened 2024-10-08
Update versions for latest .NET release

eerhardt:
9.0 RC2 shipped today. Along with a servicing release of .NET 8. Updating to these versions.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6194)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -195,7 +195,7 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
     <PackageVersion Include="Azure.Identity" Version="1.12.1" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -195,11 +195,11 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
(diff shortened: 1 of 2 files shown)


Directory.Packages.props
```diff
@@ -195,7 +195,7 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
```
joperezr, line 198: This was pinned for CG so I'm hesitant to keep it as 8.0.4. The intention would be that at some point, whoever is bringing that transitively would update themselves to something past 8.0.4 and we can remove the transitive pinning. If we keep bumping this, then that is unlikely to happen.

eerhardt: 8.0.4 is now marked vulnerable

https://www.nuget.org/packages/System.Text.Json/8.0.4

image

joperezr: gotcha, ok that totally makes sense then, sorry wasn't aware.

eng/Versions.props
```diff
@@ -39,56 +39,64 @@
...
     <MicrosoftNETRuntimeWorkloadTestingInternalVersion>9.0.0-preview.5.24272.3</MicrosoftNETRuntimeWorkloadTestingInternalVersion>
 
     <NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>8.0.8</NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>
     <!-- for templates -->
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet8>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet8>
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
+
+    <!-- System dependencies -->
+    <SystemFormatsAsn1PackageVersion>8.0.1</SystemFormatsAsn1PackageVersion>
+    <SystemTextJsonPackageVersion>8.0.5</SystemTextJsonPackageVersion>
```
sebastienros, line 78: Should we use the 9.0-rc one and this one for 8?

sebastienros: Ok, it's below, sorry

sebastienros (approved):

PR #6195, opened 2024-10-08
Prevent preview packages to ship as RC

joperezr:
## Description

@DamianEdwards @eerhardt, as discussed offline, these changes will make it so that packages that are going to stay in preview once 9.0 ships, won't go to get an RC.1 label, and instead continue shipping as preview.4.

Making this change directly on the rc1 branch and will backport to main.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6195)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -5,7 +5,8 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <PreReleaseVersionLabel>rc.1</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel Condition="'$(SuppressFinalPackageVersion)' != 'true'">rc.1</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel Condition="'$(PreReleaseVersionLabel)' == ''">preview.4</PreReleaseVersionLabel>
     <!--
       When running package validation as part of the build, we want to ensure we didn't break the API against the previous
       version of each package. The following property points to the package version that should be used as baseline.
--- a/src/Components/Aspire.Azure.AI.OpenAI/Directory.Build.props
+++ b/src/Components/Aspire.Azure.AI.OpenAI/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
       <!-- This needs to be set to true before importing parent Directory.Build.props -->
-       <!-- This library can't ship stable until Azure.AI.OpenAI ships stable. -->
+      <!-- This library can't ship stable until Azure.AI.OpenAI ships stable. -->
       <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
     </PropertyGroup>
   
--- a/src/Components/Aspire.Azure.AI.OpenAI/Directory.Build.props
+++ b/src/Components/Aspire.Azure.AI.OpenAI/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+      <!-- This needs to be set to true before importing parent Directory.Build.props -->
+       <!-- This library can't ship stable until Azure.AI.OpenAI ships stable. -->
+      <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    </PropertyGroup>
+  
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+  </Project>
--- a/src/Components/Aspire.OpenAI/Directory.Build.props
+++ b/src/Components/Aspire.OpenAI/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
       <!-- This needs to be set to true before importing parent Directory.Build.props -->
-       <!-- This library can't ship stable until OpenAI ships stable. -->
+      <!-- This library can't ship stable until OpenAI ships stable. -->
       <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
     </PropertyGroup>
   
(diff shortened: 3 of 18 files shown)


joperezr:
Screenshot of packages produced locally after these changes:

image

src/Components/Aspire.Azure.AI.OpenAI/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+      <!-- This needs to be set to true before importing parent Directory.Build.props -->
+       <!-- This library can't ship stable until Azure.AI.OpenAI ships stable. -->
```
radical, line 4:
```suggestion
      <!-- This library can't ship stable until Azure.AI.OpenAI ships stable. -->
```

src/Components/Aspire.OpenAI/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+      <!-- This needs to be set to true before importing parent Directory.Build.props -->
+       <!-- This library can't ship stable until OpenAI ships stable. -->
```
radical, line 4:
```suggestion
      <!-- This library can't ship stable until OpenAI ships stable. -->
```

radical:
IMHO, we should add a test for this somewhere. One place could be next to this target https://github.com/dotnet/aspire/blob/658eed954f7af61465f31c7060881f0505ebebe7/tests/Shared/Aspire.Workload.Testing.targets#L73

.. or in template tests.

joperezr:
> MHO, we should add a test for this somewhere. One place could be next to this target

I don't disagree, but it wouldn't be straight forward as pr and local builds override this value for all packages to `-dev` or `-ci`. This means this could only really be tested in an official build. We can add an MSBuild target that runs after pack only on Official builds to validate that the prerelease tag is the right one, but that wouldn't cover us for CI

eng/Versions.props
```diff
@@ -5,7 +5,8 @@
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <PreReleaseVersionLabel>rc.1</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel Condition="'$(SuppressFinalPackageVersion)' != 'true'">rc.1</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel Condition="'$(PreReleaseVersionLabel)' == ''">preview.4</PreReleaseVersionLabel>
```
eerhardt, line 9: Can we instead do this in a `.targets` so we don't have to split properties out of the .csprojs?

joperezr: Unfortunately no, this needs to happen in a props file as it's values are read in the Common props from the SDK, so setting it later would be too late. I am adding a test like @radical suggested, so at least we'll know if this is broken in a project because they aren't using a d.b.props file.

eerhardt: Have 2 properties here - named whatever.

And then in the Directory.Build.targets set the `PreReleaseVersionLabel`.

This is better ordering in MSBuild IMO

eerhardt: This is how arcade consumes this property - in a .targets.

https://github.com/dotnet/arcade/blob/382c993e9ab742d2292c9330e1b952c5b57aaa2c/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L131-L137

joperezr: That is technically possible, but I would be worried about the potential impact of those changes, especially when making these in a release branch as we don't know what changing the ordering of those properties could do. What would you say about doing this less risky change here, and then in main (which will have more bake time and validation) do the change you suggest?

src/Components/Directory.Build.targets
```diff
@@ -27,6 +27,12 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="ValidatePreReleaseLabelIsCorrectForPreviewPackages"
+          AfterTargets="Pack"
+          Condition="'$(OfficialBuildId)' != '' and '$(SuppressFinalPackageVersion)' == 'true' and !$(PreReleaseVersionLabel.StartsWith('preview'))">
```
eerhardt, line 32: Would it be better to actually look at the outputted nupkg and verify its version?

eng/Versions.props
eerhardt: >  this needs to happen in a props file as it's values are read in the Common props from the SDK, so setting it later would be too late

I don't think this is true. Doing a `dotnet build /bl` and looking at the preprocess, the only places I see that use this property are the .csproj setting it, and the arcade line above.

> but I would be worried about the potential impact of those changes

If you are really concerned about the risk, we can clutter the 8 projects here like we are doing. But I don't think this is ideal. If you want to do it the correct way in main, I'm fine with that.

joperezr: Yeah, at this point my top priority is to de-risk RC1, so unless you feel strongly about it, I'd opt for keeping this as-is and then getting this refactored in main

src/Components/Directory.Build.targets
joperezr: This would be possible I think, yes, via a Regex that extracts the label. Unless you feel strongly about it, I can make those changes in main along with your refactoring suggestion in favor of getting rc1 build ready.

eerhardt (approved):

Issue #6040, opened 2024-10-01
Orchestrator & Operators


mitchdenny:
Since .NET Aspire 8.0 the start-up of resources has occurred during the AppHost start-up process blocking the return of `StartAsync(...)` until all resources are running. This is because as part of the startup process components like `ApplicationExecutor` need to block startup of particular resources until environment variables can be fully resolved and with the addition of `WaitFor` in .NET Aspire 9.0 - until the health checks associated with the resource return healthy.

Additionally, we have inconsistent behavior between resources that are handled by specific code (such as containers, processes, projects in `ApplicationExecutor` and Azure resources via Azure Provisioner) and custom resources that sit inside the application model in their default starting state.

At the moment if you are implementing your own custom resource you need to create a lifecycle hook or wireup to an event to trigger a background process to handle doing things like publishing events and updating status via the `ResourceNotificationService`.

This proposal recommends introducing a primitive in the the .NET Aspire application model called an _operator_ which is a component which is given responsibility for managing a resource. `ApplicationExecutor` and `AzureProvisioner` are examples of operators but are implemented in different ways.

The goal would be to introduce a mechanism where when a resource is added to the application model an operator is assigned based on a set of rules (or can be explicitly specified in annotations). From that point forward that operator is fully responsible for managing the lifecycle for that resource.

As part of this change, we would move from a blocking process in startup to a model where operators start executing after the application has started.

In turn this change would open up the possibility of the application model being more mutable during the app hosts lifetime. With the introduction of command annotations, we have the ability to trigger code execution from the dashboard. Often this code can just be imperative C# code - however in some cases it would be useful to be able to launch a container resource or a process as a way of fulfilling the requirements of a command being executed. By making the application model mutable at runtime (e.g. adding resources after startup) we may be able to facilitate this.

A scenario might be having a load test command on a API project which triggers an associated k6 load test inside a container image and being able to view the outputs of the test.

I see operators as being singletons which are registered in the container. There would be a single operator for container resources, one for executables, one for projects. All of these would probably delegate a lot of their processing to the application executor which would need work to enable it to be less "single shot" - some of this has already been done with the introduction of start/stop for resources.

There would be a single "orchestrator" resource who is responsible for selecting the most appropriate operator for a given resource. This would be a cooperative process where each operator is given an opportunity to nominate itself for a specific resource. If more than one operator says it can handle a resource then we would need some mechanism to pin a resource to a particular operator - this could be an annotation.

```csharp
[...]

PR #6201, opened 2024-10-09
Don't block on application executor.

### LINKED ISSUE #6038: Distributed application started message in app host console is delayed based on wait fors

### Describe the bug

App host displays this message when it starts:

> Distributed application started. Press Ctrl+C to shut down.

But it is delayed based on wait for checks. That means the dashbaord can be opened in a browser window and loaded before this message is shown.

Also, (I haven't checked this) if a resource waits forever then could the message never be shown?

start-message-delayed

### Expected Behavior

`Distributed application started. Press Ctrl+C to shut down.` is displayed promptly once app host startup is done, similar to ASP.NET Core web apps and other apps built with hosing.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

mitchdenny:
## Description

This is an onion peeling PR to see what happens if we simply no longer block on `ApplicationExecutor` when we start up. I'm expecting lots of things to break in unit tests but have verified that test shop works. This is an exploration based on #6038 but is something we are going to need to tackle as part of #6040.

Fixes #6038 

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6201)

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpHostService.cs
+++ b/src/Aspire.Hosting/Dcp/DcpHostService.cs
@@ -72,7 +72,7 @@ internal sealed class DcpHostService : IHostedLifecycleService, IAsyncDisposable
 
         EnsureDcpContainerRuntime(dcpInfo);
         EnsureDcpHostRunning();
-        await _appExecutor.RunApplicationAsync(cancellationToken).ConfigureAwait(false);
+        _ = Task.Run(() => _appExecutor.RunApplicationAsync(cancellationToken).ConfigureAwait(false), cancellationToken);
     }
 
     private void EnsureDcpContainerRuntime(DcpInfo? dcpInfo)


davidfowl:
Biggest regression so far is the fact that endpoints are not guaranteed to be allocated after calling StartAsync. That also means resolving connection strings changes. Now that we have the eventing model that's less of a concern, but a big breaking change.

PR #6209, opened 2024-10-09
Remove HealthStatus defaulting to healthy when there is no value from a health check

### LINKED ISSUE #6125: Resources in a waiting state should not show up as healthy
image

adamint:
## Description

The proposed changes are

- give resources a null health status until health reports start to be received
- give resources a healthy status if they do not have health checks and are running
- create initial health reports with a null status so that the dashboard immediately knows which health checks will be running and that they have not returned data yet (make `HealthReportSnapshot.Status` nullable
- remove the "Waiting for data..." text and immediately render the grid of health reports, with an indeterminate loading circle while we wait for health check data

Animation

@davidfowl @drewnoakes 

Fixes #6125

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6209)


src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -493,9 +493,7 @@ private static CustomResourceSnapshot GetCurrentSnapshot(IResource resource, Res
             {
                 ResourceType = resource.GetType().Name,
                 Properties = [],
-                HealthStatus = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _)
-                    ? null // Indicates that the resource has health check annotations but the health status is unknown.
-                    : HealthStatus.Healthy
```
adamint, line 498: It seems like an incorrect assumption to deem any resource that does not have a health check healthy?

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -52,14 +52,11 @@ public sealed record CustomResourceSnapshot
...
-    /// is known to have a health check and no reports exist, then this value is <see langword="null"/>.
-    /// </para>
-    /// <para>
-    /// Defaults to <see cref="HealthStatus.Healthy"/>.
+    /// This value is derived from <see cref="HealthReports"/>. If a resource is known to have a health check
+    /// and no reports exist, or if a resource does not have a health check, then this value is <see langword="null"/>.
     /// </para>
     /// </remarks>
-    public HealthStatus? HealthStatus { get; init; } = Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus.Healthy;
+    public HealthStatus? HealthStatus { get; init; }
```
davidfowl, line 59: 👍🏾

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: Correct. We needed to differentiate between the, I don't have any health status ever and I will have one eventually (it hasn't come in yet). I think it just means, no health reports means the health status isn't available, and at least one means there is.

adamint: Is differentiation needed? In both cases (I don't have any health status ever) and (I will have one eventually), the current health status are both null

davidfowl:
I think the tests are failing

JamesNK (changes requested):
Needs test. Could probably add an assert on health status to an existing test that looks at custom resource snapshot properties after publish.

drewnoakes:
I'm not sure the approach here is correct. With the changes here, I expect the UI to show "Running (Unhealthy)" for resources that don't have health checks.

In `main`, a `null` value is used so that the dashboard's state column shows a resource as unhealthy (i.e. "Running (Unhealthy)") when we are expecting health reports to arrive. The `null` applies only when a health check exists, and so carries that information.

This distinction was added as a fix for the following scenario:

- A resource that has a health check starts, but we have no health reports.
  - It appears in the UI as "Running" and green.
- A health report arrives showing that the resource is unhealthy.
  - It changes to show as "Running (Unhealthy)" with a white icon.
- The resource becomes healthy.
  - It changes to "Running" and green again.

The `null` value allows us to prevent that first state, which incorrectly suggests the resource is healthy.

davidfowl:
> I'm not sure the approach here is correct. With the changes here, I expect the UI to show "Running (Unhealthy)" for resources that don't have health checks.

No, it should show Running, not unhealthy. If we don't have health checks then we're in a "we don't know any better and I have to assume the resource is ready" state.

drewnoakes:
> No, it should show Running, not unhealthy. If we don't have health checks then we're in a "we don't know any better and I have to assume the resource is ready" state.

Sorry I wasn't clear. I agree with you, and that's what happens today. I should have written:

> I'm not sure the approach here is correct. With the changes here, I expect the UI to show "Running (Unhealthy)" for resources that have health checks but don't yet have health reports.

adamint:
> I think the tests are failing

It is a helix issue

adamint:
> I'm not sure the approach here is correct. With the changes here, I expect the UI to show "Running (Unhealthy)" for resources that don't have health checks.
> 
> In `main`, a `null` value is used so that the dashboard's state column shows a resource as unhealthy (i.e. "Running (Unhealthy)") when we are expecting health reports to arrive. The `null` applies only when a health check exists, and so carries that information.

That's a confusing distinction. It carries the meaning of "no current health status available." This meaning also applies to resources that do not contain health checks, because we do not have a health status available for them. So when resources aren't defaulting to healthy, `null` can be simplified to "no current health status available" for all resources, regardless of health checks being present.

adamint:
> Needs test. Could probably add an assert on health status to an existing test that looks at custom resource snapshot properties after publish.

Could you clarify? I also added a null assert in `ResourceNotificationTests` but I'm pretty sure you mean something else.

tests/Aspire.Dashboard.Components.Tests/Controls/StateColumnDisplayTests.cs
```diff
@@ -0,0 +1,112 @@
...
+    [InlineData(
+        /* state */ KnownResourceState.Running, null, "Unhealthy", null,
+        /* expected output */ RunningAndUnhealthyTooltip, "CheckmarkCircleWarning", Color.Neutral, "Running (Unhealthy)")]
+    [InlineData(
+        /* state */ KnownResourceState.Running, null, null, "warning",
+        /* expected output */ null, "Warning", Color.Warning, "Running")]
+    [InlineData(
+        /* state */ KnownResourceState.Running, null, null, "NOT_A_VALID_STATE_STYLE",
+        /* expected output */ null, "Circle", Color.Neutral, "Running")]
+    public void ResourceViewModel_ReturnsCorrectIconAndTooltip(
```
adamint, line 54: @JamesNK I also added tests for these methods since there are so many different icon/tooltip/text possibilities.

JamesNK:
> > Needs test. Could probably add an assert on health status to an existing test that looks at custom resource snapshot properties after publish.
> 
> Could you clarify? I also added a null assert in `ResourceNotificationTests` but I'm pretty sure you mean something else.

Write a test for each scenario that people are talking about here and assert the health status. That draws a line in the sand for how each scenario would work.

* Resource doesn't have health checks
* Resource has health checks but no result yet.

Maybe others? I didn't look at each comment closely.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
mitchdenny: If a resource has no heath checks but is running, isn't it by definition healthy based on the criteria supplied by the hosting integration author?

Maybe no health checks + running = healthy
Not running + no health checks = null state.

mitchdenny: I think by changing this code we would break WaitUntilHealthyAsync(...)

adamint: > If a resource has no heath checks but is running, isn't it by definition healthy based on the criteria supplied by the hosting integration author?

That could be an assumption we make, since without healthchecks we don't have any way to tell the state of the resource. If so, we could default to `Healthy` but keep `HealthReportSnapshot.HealthStatus` nullable.

adamint: I've made this change.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -492,13 +492,16 @@ private static CustomResourceSnapshot GetCurrentSnapshot(IResource resource, Res
...
-                Properties = [],
-                HealthStatus = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _)
-                    ? null // Indicates that the resource has health check annotations but the health status is unknown.
-                    : HealthStatus.Healthy
+                Properties = []
             };
         }
 
+        // A resource is also healthy if it has no health check annotations and is in the running state.
+        if (previousState.HealthStatus is null && !resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _) && previousState.State?.Text == KnownResourceStates.Running)
```
adamint, line 500: @mitchdenny

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -389,6 +390,20 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
         return Task.CompletedTask;
     }
 
+    /// <summary>
+    /// Update resource snapshot health status if the resource is running with no health checks.
+    /// </summary>
+    private static CustomResourceSnapshot UpdateHealthStatus(IResource resource, CustomResourceSnapshot previousState)
+    {
+        // A resource is also healthy if it has no health check annotations and is in the running state.
+        if (!resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _) && previousState.State?.Text == KnownResourceStates.Running)
```
mitchdenny, line 399: I don't think this is sufficient. If you look at `ResourceHealthCheckService` it considers the health checks of parent resources when determining child resource health. What we have here could introduce subtle bugs around child resource status.

We are better off leaving the true health status to be determined by the `ResourceHealthCheckService`.

mitchdenny: Actually upon reflection this is fine. You aren't interpreting health status here you are just asserting that none of the resources or their parents have health check annotations.

Coincidently this should be rare because in most cases where we have a parent and a child resource we have implemented health checks in the parent at least.

JamesNK: Should there be tests around parent/child resources and health status?

adamint: I wouldn't think so, but @mitchdenny?

mitchdenny: It would be good to have some from a dashboard perspective. We have the health status itself covered on the app model side of things.

adamint:
This is ready for review.

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor.cs
```diff
@@ -25,31 +25,40 @@ public partial class StateColumnDisplay
...
+    }
+
     /// <summary>
     /// Gets the tooltip for a cell in the state column of the resource grid.
     /// </summary>
     /// <remarks>
     /// This is a static method so it can be called at the level of the parent column.
     /// </remarks>
-    public static string? GetResourceStateTooltip(ResourceViewModel resource, IStringLocalizer<Columns> Loc)
+    internal static string? GetResourceStateTooltip(ResourceViewModel resource, string exitedUnexpectedlyTooltip, string exitedTooltip, string runningAndUnhealthyTooltip)
```
drewnoakes, line 43: Did you consider mocking an `IStringLocalizer` rather than breaking this into two methods? Doing that would ensure the correct resources are being used, and would avoid an extra method with a bit of a strange arg list.

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor.cs
```diff
@@ -25,31 +25,40 @@ public partial class StateColumnDisplay
     [Inject]
     public required IStringLocalizer<Columns> Loc { get; init; }
 
+    internal static string? GetResourceStateTooltip(ResourceViewModel resource, IStringLocalizer<Columns> loc)
+    {
+        return GetResourceStateTooltip(
+            resource,
+            loc[Columns.StateColumnResourceExitedUnexpectedly].Value,
+            loc[Columns.StateColumnResourceExited].Value,
+            loc[nameof(Columns.RunningAndUnhealthyResourceStateToolTip)]);
```
drewnoakes, line 34: This looks wrong. Actually, the bug existed before, but we can fix it here.

Without the `nameof` this will only ever return the neutral (`en-US`) value:

```suggestion
        return GetResourceStateTooltip(
            resource,
            loc[nameof(Columns.StateColumnResourceExitedUnexpectedly)],
            loc[nameof(Columns.StateColumnResourceExited)],
            loc[nameof(Columns.RunningAndUnhealthyResourceStateToolTip)]);
```

This would make a good analyzer.

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor
```diff
@@ -1,7 +1,7 @@
-﻿@using Aspire.Dashboard.Model
+﻿@using Aspire.Dashboard.Resources
 
 @{
-    var vm = GetStateViewModel();
+    var vm = GetStateViewModel(Resource, Loc[Columns.UnknownStateLabel]);
```
drewnoakes, line 4:
```suggestion
    var vm = GetStateViewModel(Resource, Loc[nameof(Columns.UnknownStateLabel)]);
```

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -117,64 +117,61 @@
...
+                    <AspireTemplateColumn Title="@ControlStringsLoc[nameof(ControlsStrings.StateColumnHeader)]" Class="stateColumn">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.StateColumnHeader)]"
+                                   Value="@(context.HealthStatus?.Humanize() ?? Loc[nameof(Resources.WaitingHealthDataStatusMessage)])"
+                                   EnableHighlighting="@(!string.IsNullOrEmpty(_filter))"
+                                   HighlightText="@_filter"
+                                   TextVisualizerTitle="@context.Name">
+                            <ContentBeforeValue>
+                                @if (context.HealthStatus is null)
+                                {
+                                    <FluentProgressRing Width="16px" Style="display: inline-block;" Class="severity-icon" />
```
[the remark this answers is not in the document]
adamint: FluentProgressRing is displayed like an icon here, and so should use the same styling, but with the added complication of it needing a different display value. I'm of the opinion that it could be bug-prone to extract it to a separate class if we change severity-icon behavior in the future

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor.cs
adamint: Done

JamesNK:
I tested and one thing I noticed is now there is no state displayed in resource details for "Unknown" state.

image

However, there have just been changes to return a new status for containers if there is no runtime. Should rebase and check it works as expected.

JamesNK:
This is what I see when I merge latest changes from main:

image

That status shouldn't be green. I don't _think_ that is a regression in this PR.

JamesNK:
New issue: https://github.com/dotnet/aspire/issues/6351

adamint:
> This is what I see when I merge latest changes from main:
> 
> image
> 
> That status shouldn't be green. I don't _think_ that is a regression in this PR.

Yup, not a regression, just an unknown enum. Since the code that determines state icon is changed here, I also added `RuntimeUnhealthy` to fix this issue
image

adamint:
@mitchdenny `HealthCheckIntervalSlowsAfterSteadyHealthyState` has failed twice on windows, succeeded always on Linux. Is it flaky?

davidfowl:
It's flaky

src/Aspire.Dashboard/Components/ResourcesGridColumns/StateColumnDisplay.razor.cs
adamint: filed https://github.com/dotnet/aspire/issues/6352

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl, line 399: This doesn't need to run every publish, it never changes. Can the caller do this instead?

drewnoakes, line 399: The new `HasAnnotationIncludingAncestorsOfType<T>` method in https://github.com/dotnet/aspire/pull/6357 would make this cheaper to call.

JamesNK:
FYI I have some changes to make in the next hour. Nothing too contraversial.

In the interest of time, I'm going to add them directly to this branch.

JamesNK:
Changes:

* Moved most of `StateColumnDisplay` logic into the view model
* Moved tests to components tests to dashboard tests. The actual control isn't being tested with BUnit (just static methods on the control) so it didn't fit there. Now that they're on a view model.
* Changed state tests to use a shared localizer. It's a bit more robust because you don't have to declare the translation again. Previously if a formatted param was added or removed, the translation in tests could get out of sync.
* Treating `RuntimeUnhealthy` state as stopped produced a sub-optimal tooltip:
    image
    Now there is a tooltip that explains the most common problem (container runtime isn't running) and includes a URL for more info.

JamesNK (approved):
Thumbs up for dashboard changes

JamesNK:
I tested more and found a situation I'm not sure is correct.

`messaging` hasn't reported a health status but the icon is green:

image

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
adamint: It becomes much cheaper to call if the first condition is `previousState.HealthStatus is not HealthStatus.Healthy` - changed

adamint: It becomes much cheaper to call if the first condition is `previousState.HealthStatus is not HealthStatus.Healthy` - would you have concerns about that?

davidfowl (approved):

drewnoakes:
There's a regression in the status column.

image

The status for the `unhealthy` and `degraded` resources should read `Running (unhealthy)` and `Running (degraded)` respectively.

davidfowl:
Maybe we can start writing dashboard view model tests

JamesNK:
I tested and the problem is in `ResourceNotificationService` (@davidfowl how dare you blame the dashboard 😋 )

https://github.com/dotnet/aspire/blob/f19600fd0ecb6b2938011403867e7b4a3afe59d2/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L398-L402

There isn't a `HealthCheckAnnotation` so it is marked as healthy. The health report with a degraded status isn't taken into account.

I'm guessing this problem is unique to the health checks sandbox test app because the health report is manually specified in the snapshot and isn't driven by the annotation.

It seems like the snapshot health reports is the better place to calculate the health status from. But this is the first time I've looked at health checks so someone with more knowledge here should decide.

drewnoakes:
@adamint will be opening a PR shortly

drewnoakes:
https://github.com/dotnet/aspire/pull/6367

adamint:
> I tested and the problem is in `ResourceNotificationService` (@davidfowl how dare you blame the dashboard 😋 )
> 
> https://github.com/dotnet/aspire/blob/f19600fd0ecb6b2938011403867e7b4a3afe59d2/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L398-L402
> 
> There isn't a `HealthCheckAnnotation` so it is marked as healthy. The health report with a degraded status isn't taken into account.
> 
> I'm guessing this problem is unique to the health checks sandbox test app because the health report is manually specified in the snapshot and isn't driven by the annotation like typical usage.
> 
> https://github.com/dotnet/aspire/blob/f19600fd0ecb6b2938011403867e7b4a3afe59d2/playground/HealthChecks/HealthChecksSandbox.AppHost/Program.cs#L43-L49
> 
> It seems like the snapshot health reports is the better place to calculate the health status from. But this is the first time I've looked at health checks so someone with more knowledge here should decide.

The bug is a little more generalized than just the health check test app. The root cause is that there should be a check to see if a resource has a `ResourceSnapshotAnnotation`; if it does, we should not attempt to override the health status. This is correct because sans a health check, there is nothing else that could modify that initial `HealthStatus`

JamesNK:
> This is correct because sans a health check, there is nothing else that could modify that initial HealthStatus

I don't think that's true. `ResourceNotificationService` is public. Someone could add a health report after creating the resource.

davidfowl:
maybe we make this property internal. The only thing publishing health state would be via the annotations

PR #6190, opened 2024-10-08
[release/9.0-rc1] Add net9.0 support for EFCore components (#5932)

sebastienros:
Backport of #5932 to release/9.0-rc1

## Customer Impact

We want developers to use EF Core RC1 packages. They contain configuration methods inspired by Aspire.

## Testing

Existing tests + new TFM. Manual testing on eShop targeting net9.0.

## Regression?

No
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6190)

PR #6213, opened 2024-10-09
[release/9.0-rc1] Update versions for latest .NET release

eerhardt:
Backport of #6194 to release/9.0-rc1

/cc @eerhardt

## Customer Impact

Our .NET Aspire 9.0-rc1 builds are still using the 9.0-rc1 builds of .NET - specifically EF Core. Customers will be using month old versions if we don't update to 9.0 rc2.

Also, System.Text.Json 8.0.4, which we depend on, was marked vulnerable. So updating to the new version of that important so users don't get vulnerable warnings.

## Testing

Automated tests in repo.

## Risk

Low. These are new versions of .NET that were released publicly. Getting more test coverage on them is helpful.

## Regression?

No

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -195,7 +195,7 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
     <PackageVersion Include="Azure.Identity" Version="1.12.1" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -195,11 +195,11 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
     <PackageVersion Include="Azure.Identity" Version="1.12.1" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="System.Formats.Asn1" Version="8.0.1" />
+    <PackageVersion Include="System.Formats.Asn1" Version="$(SystemFormatsAsn1PackageVersion)" />
+    <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonPackageVersion)" />
   </ItemGroup>
 </Project>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -16,9 +16,9 @@
     <DefaultTargetFramework>net8.0</DefaultTargetFramework>
     <AllTargetFrameworks>$(DefaultTargetFramework);net9.0</AllTargetFrameworks>
     <!-- dotnet 8.0 versions for running tests -->
-    <DotNetRuntimePreviousVersionForTesting>8.0.8</DotNetRuntimePreviousVersionForTesting>
+    <DotNetRuntimePreviousVersionForTesting>8.0.10</DotNetRuntimePreviousVersionForTesting>
     <!-- dotnet 8.0 versions for running tests - used for workload tests -->
-    <DotNetSdkPreviousVersionForTesting>8.0.401</DotNetSdkPreviousVersionForTesting>
+    <DotNetSdkPreviousVersionForTesting>8.0.403</DotNetSdkPreviousVersionForTesting>
     <UseVSTestRunner>true</UseVSTestRunner>
     <!-- Enable to remove prerelease label. -->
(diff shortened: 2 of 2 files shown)


sebastienros (approved):

sebastienros:
Failed twice on elasticsearch, potentially true positive.

Directory.Packages.props
```diff
@@ -195,11 +195,11 @@
     <PackageVersion Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
 
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
     <PackageVersion Include="Azure.Identity" Version="1.12.1" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="System.Formats.Asn1" Version="8.0.1" />
+    <PackageVersion Include="System.Formats.Asn1" Version="$(SystemFormatsAsn1PackageVersion)" />
```
joperezr, line 202: Similar comment to your PR in main, but should we avoid using properties and instead spell out the versions here? This is the reasoning behind my question: Next time that we are bumping dependencies, if the version bump that I see is under this section of Pinned for CG, then I'd likely undo the change, as the long term goal is to remove these as opposed to continue bumping them (once the direct dependency updates the transitive dependency version). If the update is in Versions.props in a property, I'm unlikely to make that link and decide not to update, causing is to always be pinning this transitive dependency.

eng/Versions.props
```diff
@@ -40,56 +40,64 @@
     <MicrosoftDotNetXUnitExtensionsPackageVersion>9.0.0-beta.24473.1</MicrosoftDotNetXUnitExtensionsPackageVersion>
     <MicrosoftDotNetBuildTasksInstallersPackageVersion>9.0.0-beta.24473.1</MicrosoftDotNetBuildTasksInstallersPackageVersion>
     <MicrosoftDotNetBuildTasksWorkloadsPackageVersion>9.0.0-beta.24473.1</MicrosoftDotNetBuildTasksWorkloadsPackageVersion>
-    <MicrosoftExtensionsHttpResiliencePackageVersion>8.9.1</MicrosoftExtensionsHttpResiliencePackageVersion>
-    <MicrosoftExtensionsDiagnosticsTestingPackageVersion>8.9.1</MicrosoftExtensionsDiagnosticsTestingPackageVersion>
+    <MicrosoftExtensionsHttpResiliencePackageVersion>8.10.0</MicrosoftExtensionsHttpResiliencePackageVersion>
```
joperezr, line 43: Not a blocker, but in reality we probably should be also editing Version.details.xml as part of these updates to signal which version of a component we actually depend on.

eng/Versions.props
```diff
@@ -40,56 +40,64 @@
...
-    <MicrosoftEntityFrameworkCoreToolsPackageVersion>9.0.0-rc.1.24451.1</MicrosoftEntityFrameworkCoreToolsPackageVersion>
+    <MicrosoftEntityFrameworkCoreCosmosPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreCosmosPackageVersion>
+    <MicrosoftEntityFrameworkCoreDesignPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreDesignPackageVersion>
+    <MicrosoftEntityFrameworkCoreSqlServerPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreSqlServerPackageVersion>
+    <MicrosoftEntityFrameworkCoreToolsPackageVersion>9.0.0-rc.2.24474.1</MicrosoftEntityFrameworkCoreToolsPackageVersion>
 
     <NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>9.0.0-rc.1</NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>
+
+    <!-- System dependencies -->
+    <SystemFormatsAsn1PackageVersion>9.0.0-rc.2.24473.5</SystemFormatsAsn1PackageVersion>
```
joperezr, line 99: These two are getting bumped from an 8.0-era package to a 9.0-era package, which will need to be undone in main anyway. Just as FYI.

joperezr: In fact, given this is an RC release, this might even be considered as a breaking change when we make the right change in GA

eerhardt:
> Failed twice on elasticsearch, potentially true positive.

It passed in main. So I'm confused if it is truly a true positive.

Directory.Packages.props
eerhardt: I honestly don't know when we are supposed to be doing logic in Directory.Packages.props and when we are supposed to be doing it in Versions.props. I was just following how we did the split with EF based on TFMs. See https://github.com/dotnet/aspire/pull/6190#discussion_r1792522231

eng/Versions.props
eerhardt: > which will need to be undone in main anyway

why?

> given this is an RC release, this might even be considered as a breaking change when we make the right change in GA

I am not understanding this point.

When you are targeting net8.0, you get the `8.0.x` versions of these packages. This line only applies to our packages that target net9.0 (3 of them - the EF libraries). And for those, this change must happen or else we get build errors saying that we are downgrading System.Text.Json and System.Formats.Asn1. EF v9 depends on the 9.0 versions of these packages. Since we are using Central Package Management, we can't say we depend on the 8.0 versions in the net9 build of the EF packages. So we _have_ to update this version in our build for those 3 packages.

sebastienros:
Maybe Elastic is just really flaky. We had issues by the past and there is already a specific timeout AFAIR.

joperezr (approved):

joperezr:
Talked offline, we need to take this. Approved.

PR #6216, opened 2024-10-09
Add 8.x template pack that can run sideXside with Aspire 9 templates

joperezr:
## Description

This will start shipping an Aspire.ProjectTemplates.8 NuGet package which will be able to be installed side by side with Aspire 9 templates, and will allow users to use Aspire 8.x in Visual Studio.

cc: @radical @phenning @DamianEdwards 

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6216)


joperezr:
As an example, here are the diff comparison between the template config from the pack being installed via the workload, and the one that will be shipped in this new template pack:

image

The diffs are the same with the rest of the templates.

joperezr:
@phenning now that I think of it, should we also bump the precedence of these templates to 9000 so that they "win" over the one installed from the workload?

radical:
Should we add a run for the workload tests where the projects are created with the new suffix?

phenning:
> @phenning now that I think of it, should we also bump the precedence of these templates to 9000 so that they "win" over the one installed from the workload?

Sure, couldn’t hurt.

phenning (approved):
Approving this now without the precedence changes in case I lose power.

radical:
`EndToEnd` tests are [failing with](https://helixre107v0xd1eu3ibi6ka.blob.core.windows.net/dotnet-aspire-refs-pull-6216-merge-290218a7658b4bc38e/Aspire.EndToEnd.Tests-basicservices/1/console.4a150d9c.log?helixlogtype=result): `TestProject.AppHost depends on Aspire.Dashboard.Sdk.linux-x64 (>= 8.2.1) but Aspire.Dashboard.Sdk.linux-x64 8.2.1 was not found. An approximate best match of Aspire.Dashboard.Sdk.linux-x64 8.2.2 was resolved.`

joperezr:
Thanks for the review folks, given the new approach we decided to take where these are includded with the new templates, I'm closing this for now as it is no longer needed.

PR #6222, opened 2024-10-10
Standardise presentation of empty cells in grids

drewnoakes:
## Description

Previously, when a cell is empty, we would be inconsistent in our presentation. Sometimes it would appear empty, other times we showed "None".

This change introduces a [en dash](https://en.wikipedia.org/wiki/Dash#En_dash) for empty cells.

---

Here's TestShop when it's starting up:

image

Extended this treatment to the "Trace" column in structured logs:

image
## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6222)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor
@@ -105,7 +105,7 @@
                                 <AspireTemplateColumn ColumnId="@SourceColumn" ColumnManager="@_manager" Title="@Loc[nameof(Dashboard.Resources.Resources.ResourcesSourceColumnHeader)]" Tooltip="true" TooltipText="@(ctx => GetSourceColumnValueAndTooltip(ctx)?.Tooltip)">
                                     @if (GetSourceColumnValueAndTooltip(context) is { } columnDisplay)
                                     {
-                                        <SourceColumnDisplay Resource="context" FilterText="@_filter" Value="@columnDisplay.Value" ContentAfterValue="@columnDisplay.ContentAfterValue" ValueToCopy="@columnDisplay.ValueToCopy" Tooltip="@columnDisplay.Tooltip" />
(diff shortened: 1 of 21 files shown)


JamesNK:
I think I disagree. The problem with showing nothing is it could create confusion about whether there are no endpoints, or data failed to load.

I had a look around for what other people do and found this: https://www.reddit.com/r/UXDesign/comments/1cow603/nullempty_fields_in_tables_what_are_your_thoughts/. They know more about UX than me.

Maybe the change that we should be making is consistently showing `-` in table cells when there is no value.

drewnoakes:
I'm not too concerned what we show, just that it should be consistent.

Can try with a dash. Personally I think an empty cell is fine.

Here's what we show for the HealthChecks playground right nowL

image

The start time and source columns are empty, and endpoints shows "None". Personally I think the horizontal gridlines make it easy to parse empty cells, and a `-` doesn't really differentiate between "no value" and "value not loaded" to me.

Interested what others think. Can share some screenshots with `-` or other ideas tomorrow.

drewnoakes:
Here's how it looks with all dashes. I think it's ok like this so will update the PR to reflect this:

image

drewnoakes:
Extended this treatment to the "Trace" column in structured logs:

image

davidfowl (approved):
Looks good

davidfowl:
Should it be centered

JamesNK:
> Should it be centered

Absolutely, definitely not 😋

PR #6225, opened 2024-10-10
Add AddAzureOpenAIChatClient

SteveSandersonMS:
## Description

This is a starting point based on the discussion earlier this week with @eerhardt, @sebastienros, and @luisquintanilla.

It allows developers to register an `IChatClient` in client projects via:

```cs
builder.AddAzureOpenAIChatClient("openai");

// Or with custom pipeline:
builder.AddAzureOpenAIChatClient("openai", pipeline => pipeline
    .UseChatOptions(...)
    .UseFunctionCalling());
```

So far I only did this for Azure OpenAI but once we've agreed on the shape would do the equivalent for OpenAI too.

### Lifetime

Currently this registers the `IChatClient` as **singleton**. That differs from prior assumptions in the MEAI repo that it would be scoped (cc: @stephentoub). The reason I did singleton here is that the underlying `OpenAIClient` is singleton, and being singleton and thread-safe is cited as a core tenet of Azure client libraries.

I suspect we should try to create an expectation that `IChatClient` implementations should be designed to be singleton and thread-safe. Anyone who finds they need per-usage-site state can store it in `ChatOptions` (e.g., `AdditionalProperties`). If this turns out to be impractical then we will have to reconsider this design. But it's better to start with singleton and later relax it to be scoped than to go in the other direction.

Also filed https://github.com/dotnet/extensions/issues/5499 to track on the MEAI side validating this for the existing `IChatClient` implementations.

If I'm misjudging this and Aspire is totally OK with client libraries registering as scoped, we can reconsider.

### APIs for chaining on `AddAzureOpenAIClient`

It's also possible we might want to enable an API pattern like this:

```cs
builder.AddAzureOpenAIClient("openai")
    .AddChatClient()          // Optionally also pass model/deployment or we'll get it from config
    .AddEmbeddingGenerator(); // Likewise
```

... instead of requiring two top-level calls (`AddAzureOpenAIChatClient` and `AddAzureOpenAIEmbeddingGenerator`). However that would involve changing the return type of `AddAzureOpenAIClient` from `void` to some builder we can put `AddChatClient` and `AddEmbeddingGenerator` extensions onto.

I'm fine with doing this if we like it. There's an argument for still also supporting the top-level calls for equivalence with `AddOllamaChatClient` etc.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No but will do when API shape is agreed
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No (justification: it's just registering DI services, which the developer would otherwise do manually and get the same end result)
  - [x] Yes
    - Link to aspire-docs issue: We need to confirm the API first
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6225)


SteveSandersonMS:
BTW does anyone know why `Microsoft.Extensions.AI` is not on the `dotnet-public` feed at `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json`? Since prerelease builds of M.E.DI.Abstractions are there, I'd expect M.E.AI to be there too.

stephentoub:
> Currently this registers the IChatClient as singleton.

Thank you :smile:

sebastienros:
@stephentoub, "/azp run"?

image

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+        this IHostApplicationBuilder builder,
+        string connectionName,
+        Func<ChatClientBuilder, ChatClientBuilder>? configurePipeline = null,
+        Action<AzureOpenAISettings>? configureSettings = null,
+        Action<IAzureClientBuilder<AzureOpenAIClient, AzureOpenAIClientOptions>>? configureClientBuilder = null,
+        string? deploymentName = null)
+    {
+        builder.AddAzureOpenAIClient(connectionName, configureSettings, configureClientBuilder);
+
+        builder.Services.AddSingleton(services =>
```
stephentoub, line 45: I assume if we change AddChatClient to be a singleton, this would instead use that?

stephentoub:
> @stephentoub, "/azp run"?
> 
> image

Possibly. I'm used to that not being appropriate in dotnet/runtime.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
SteveSandersonMS: Yes, it would be equivalent, so it might as well do.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+using OpenAI;
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// Provides extension methods for registering <see cref="IChatClient"/> as a singleton in the services provided by the <see cref="IHostApplicationBuilder"/>.
+/// </summary>
+public static class AspireAzureOpenAIChatClientExtensions
+{
+    private const string DeployentKey = "Deployment";
```
stephentoub, line 20: Deployent => Deployment

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private static string GetRequiredDeploymentName(IConfiguration configuration, string connectionName)
+    {
+        string? deploymentName = null;
+
+        if (configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            var connectionBuilder = new DbConnectionStringBuilder { ConnectionString = connectionString };
+            deploymentName = (connectionBuilder[DeployentKey] ?? connectionBuilder[ModelKey]).ToString();
```
stephentoub, line 67: If `connectionBuilder[DeploymentKey]` might be null, what guarantees `connectionBuilder[ModelKey]` will be non-null? If it's null, the ToString will null ref

SteveSandersonMS:
```suggestion
            deploymentName = (connectionBuilder[DeployentKey] ?? connectionBuilder[ModelKey])?.ToString();
```

stephentoub: Actually, it looks like DbConnectionStringBuilder's indexer will throw rather than return null if the keyword isn't found?

SteveSandersonMS: Really? I thought I checked that. Maybe I somehow didn't. But this will certainly come up when adding unit tests once the API design is agreed.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private static string GetRequiredDeploymentName(IConfiguration configuration, string connectionName)
+    {
+        string? deploymentName = null;
+
+        if (configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            var connectionBuilder = new DbConnectionStringBuilder { ConnectionString = connectionString };
+            deploymentName = (connectionBuilder[DeployentKey] ?? connectionBuilder[ModelKey])?.ToString();
```
stephentoub, line 67: It looks like DbConnectionStringBuilder's indexer actually throws rather than returning null if the keyword isn't found. Should these instead use TryGetValue?

stephentoub: https://github.com/dotnet/runtime/blob/e32148a8bd322f0b7cd5368d4c3810354d9fc14a/src/libraries/System.Data.Common/src/System/Data/Common/DbConnectionStringBuilder.cs#L78-L90

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        var configurationSectionName = AspireAzureOpenAIExtensions.DefaultConfigSectionName;
```
stephentoub, line 70: I assume we don't need to be concerned about conflicting deployment/model names being specified in the two different places, and it's ok for us to just silently prefer one over the other rather than error'ing out?

sebastienros:
I think the builder alternative was suggested instead of a lambda in the extension method because we already have two optional lambdas in it (to customize option and settings).

```c#
    public static void AddAzureOpenAIClient(
        this IHostApplicationBuilder builder,
        string connectionName,
        Action<AzureOpenAISettings>? configureSettings = null,
        Action<IAzureClientBuilder<AzureOpenAIClient, AzureOpenAIClientOptions>>? configureClientBuilder = null)
```

stephentoub (approved):
Design looks reasonable to me.

src/Components/Aspire.OpenAI/MEAIPackageOverrides.targets
```diff
@@ -0,0 +1,13 @@
...
+  <ItemGroup>
+    <!--
+      Microsoft.Extensions.AI depends on 9.x packages, even on net8.0, so we have to override central package management
+      to avoid "package downgrade" build errors. This is only used when referencing Aspire.OpenAI and doesn't break
+      compatibility with net8.0.
+    -->
+    <PackageReference Include="Microsoft.Extensions.Primitives" VersionOverride="9.0.0-*" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" VersionOverride="9.0.0-*" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" VersionOverride="9.0.0-*" />
+    <PackageReference Include="System.Text.Json" VersionOverride="9.0.0-*" />
```
sebastienros, line 11: [This PR](https://github.com/dotnet/aspire/pull/6194) should simplify that, right? Or do we also want to use the package in apps targeting 8.0?

SteveSandersonMS:
> I think the builder alternative was suggested instead of a lambda in the extension method because we already have two optional lambdas in it (to customize option and settings).

Can you give an example of how that would look?

If you mean [something like this](https://github.com/dotnet/temp-ai-abstractions/issues/296), we can certainly look at going that way. However it would prohibit this sort of calling pattern:

```cs
builder.AddAzureOpenAIClient("openai")
    .AddChatClient()
    .AddEmbeddingGenerator();
```

... because if `AddChatClient` returned something you can chain pipeline builder calls onto, then you wouldn't later be able to call `AddEmbeddingGenerator`.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
sebastienros: Example: we use `connectionBuilder.TryGetValue(ConnectionStringEndpoint, out var endpoint)` in `AddOpenAIClientFromConfiguration`.

src/Components/Aspire.OpenAI/MEAIPackageOverrides.targets
SteveSandersonMS: > Or do we also want to use the package in apps targeting 8.0?

Yes. We have no choice, because M.E.AI.Abstractions targets the 9.x versions of these packages (even on `net8.0`).

eerhardt: We shouldn't be using wildcards `*`s pacakge versions.

As for using STJ 9 on net8.0, I think that is something to think hard about in MEAI before shipping, since 9 is STS and 8 is LTS.

stephentoub: @SteveMCarroll

SteveSandersonMS:
I'm keen to finish off this PR but to avoid any wasted work can someone from the Aspire side acknowledge if you're happy with the API shape so far? cc @sebastienros @eerhardt @davidfowl

Specifically, this sort of registration usage:

```cs
builder.AddAzureOpenAIChatClient("openai");
```

Or with custom pipeline:

```cs
builder.AddAzureOpenAIChatClient("openai", pipeline => pipeline
    .UseChatOptions(...)
    .UseFunctionCalling()); // ... plus any other pipeline builder calls
```

### Alternative

The only other main proposal is the following, which chains the IChatClient pipeline onto the top-level call instead of using a lambda:

```cs
builder.AddAzureOpenAIClient("openai").AddChatClient()
    .UseChatOptions(...)
    .UseFunctionCalling(); // ... plus any other pipeline builder calls
```

This looks better in this case but has the drawback that you can't also call `AddEmbeddingGenerator` without capturing the `builder.AddAzureOpenAIClient("openai")` result into a separate variable, plus we'd have to change the return type of `AddAzureOpenAIClient`/`AddOpenAIClient` from `void` to some builder we can put `AddChatClient` and `AddEmbeddingGenerator` onto.

davidfowl:
The alternative looks better.

SteveSandersonMS:
@davidfowl Hold on, I updated the code above, after realising I was coupling together two unrelated designs. Can you re-review?

SteveSandersonMS:
As for the question of whether the IChatClient pipeline should be configured via a lambda or should be chained onto the `AddAzureOpenAIChatClient`/`AddChatClient` call, this is a further design pivot covered by https://github.com/dotnet/temp-ai-abstractions/issues/296. Whichever way we go on https://github.com/dotnet/temp-ai-abstractions/issues/296 should result in the same choice on that point here.

stephentoub:
> It has a drawback

We also need to revisit the Use method, which someone might chain at the end (assuming the previous methods return the same ChatClientBuilder) not realizing it's an expensive nop. Renaming it back to Build wood help with that.

SteveSandersonMS:
> We also need to revisit the Use method, which someone might chain at the end

I agree. We should remove that method completely and make it a constructor parameter to `ChatClientBuilder`.

sebastienros:
I like both alternatives (with or without builder). But if we expect users to use mostly use chat or embedding clients directly instead of the `OpenAIClient` then would prefer the non-builder alternative and no lambda.

If we use a builder on `AddAzureOpenAIClient` then use the lambda instead of having to capture the builder (like we do for `withCosmosDBEmulator(p => p.WithVolume())`

davidfowl:
> If we use a builder on AddAzureOpenAIClient then use the lambda instead of having to capture the builder (like we do for withCosmosDBEmulator(p => p.WithVolume())

Except this isn't an app host API.

eerhardt:
One concern I have is this method:

https://github.com/dotnet/aspire/blob/4f7dbd485c260e25f92f4cacdef724bddb93967a/src/Components/Aspire.Azure.AI.OpenAI/AspireConfigurableOpenAIExtensions.cs#L21-L27

Following the pattern of `AddAzureOpenAIChatClient`, we would need to duplicate this method for each type of client we wanted: chat, embedding, audio, etc. So we'd have:

* `AddAzureOpenAIChatClientFromConfiguration`
* `AddAzureOpenAIEmdeddingClientFromConfiguration`

Where as if adding the OpenAIClient was separated from adding the "child" chat, embedding, assistant, audio, etc. clients we wouldn't need to duplicate the "from configuration" part.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIChatClientExtensions.cs
SteveSandersonMS: Just to be safe, I'm updating this to throw if both are specified.

src/Components/Aspire.OpenAI/MEAIPackageOverrides.targets
stephentoub: > The current status for this is that MEAI.Abstractions will depend on STJ v8, whereas MEAI will depend on STJ v9, right?

When targeting < net9.0, yes.

> If so then the Aspire integration package will still have to depend on STJv9 (even on net8.0)... Is this going to be a problem?

Not sure. Who do we think will be concerned?

SteveSandersonMS: @eerhardt Your comment above gave me the impression this was a concern. Can you let me know if it is?

eerhardt: Generally there are 2 concerns with lifting libraries out of the shared framework:

1. (technical) It is another .dll that needs to be distributed with your application. Also, it is no longer ReadyToRun'd because we aren't using the assembly in the shared framework (which is RID-specific and R2R'd).
2. (policy) the servicing lifetime of System.Text.Json 9.0 ends before the servicing lifetime of System.Text.Json 8.0. If you are staying on `net8.0` TFM, you probably want to keep on a long-term support framework. But using MEAI (not .Abstractions) means that you are signing up for being on the short-term support System.Text.Json. Meaning you will need to move up to the v10 System.Text.Json faster than you would have if you stayed on the v8.

SteveSandersonMS:
## Updates

Following discussion here, the latest proposed shape for the Aspire integration is:

```cs
builder.AddAzureOpenAIClient("openai").AddChatClient("gpt-4o-mini", pipeline => pipeline
    .UseFunctionCalling()
    .UseOpenTelemetry());
```

I think that's fine but am still keen to consider the non-lambda alternative:

```cs
builder.AddAzureOpenAIClient("openai").AddChatClient("gpt-4o-mini")
    .UseFunctionCalling()
    .UseOpenTelemetry();
```

... or with @kzu's [proposed simplification](https://github.com/dotnet/extensions/issues/5572#issuecomment-2443281499):

```cs
builder.AddAzureOpenAIClient("openai").AddChatClient("gpt-4o-mini")
    .UseDefaultPipeline();
```

I like `UseDefaultPipeline` because it avoids either being too labouriously explicit (`.UseFunctionCalling().UseOpenTelemetry()`) or too magic (eliding both of those and making them implicit, which is then confusing if you want to put something before or after them).

### More complex case

Arguably this is also better even if you want to register multiple services. Before you'd have:

```cs
builder.AddAzureOpenAIClient("openai")
    .AddChatClient("gpt-4o-mini", pipeline => pipeline
        .UseFunctionCalling()
        .UseOpenTelemetry())
    .AddEmbeddingGenerator("deployment-name", pipeline => pipeline
        .UseOpenTelemetry());
```

That's a bit scary but still works as a single logical statement. It could conceptually be reduced to something like this to reduce the sense of nested lambdas:

```cs
builder.AddAzureOpenAIClient("openai")
    .AddChatClient("gpt-4o-mini", ChatClient.DefaultPipeline)
    .AddEmbeddingGenerator("deployment-name", EmbeddingGenerator.DefaultPipeline);
```

With the chaining above, you'd need a further variable:

```cs
var openai = builder.AddAzureOpenAIClient("openai");
openai.AddChatClient("gpt-4o-mini").UseDefaultPipeline();
openai.AddEmbeddingGenerator("deployment-name").UseDefaultPipeline();
```

I think this is at least equally clean, and given how this makes clear how you can chain in custom stuff before and after the default pipeline pieces, it feels good to me.

SteveSandersonMS:
I think we want to move forwards with this as soon as we can, so I'm particularly looking for feedback from @eerhardt on the STJv9 dependency question and @stephentoub on the API shape.

src/Components/Aspire.OpenAI/MEAIPackageOverrides.targets
SteveSandersonMS: Thanks @eerhardt. Based on this I'm going to complete the PR using the API design [here](https://github.com/dotnet/aspire/pull/6225#issuecomment-2470704222).

If anyone doesn't think this is the best API shape then please speak up asap!

eerhardt:
> I think that's fine but am still keen to consider the non-lambda alternative:

I like the non-lambda alternative personally. Yes it means you may have to capture something in a variable in order to do multiple, but IMO that is OK, you can only be so fluent.

> I like UseDefaultPipeline because it avoids either being too labouriously explicit (.UseFunctionCalling().UseOpenTelemetry()) or too magic (eliding both of those and making them implicit, which is then confusing if you want to put something before or after them).

A principle of .NET Aspire is that telemetry is enabled by default, so I'm a little concerned that you would need to opt-in to telemetry by calling UseDefaultPipeline. We don't have anything else like this in .NET Aspire. Instead everything else has an opt-out. Would it be possible to have a `.ClearPipeline()` for the cases where you wanted to not use the default, or add something before the default? For example:

```C#
builder.AddAzureOpenAIClient("openai").AddChatClient("gpt-4o-mini")
    .ClearPipeline()
    .UseFunctionCalling();
```

would turn off telemetry, but keep function calling.

> I think we want to move forwards with this as soon as we can

Have we given any more thought to the Hosting integration APIs? Specifically @sebastienros's proposal at [Configure OpenAI models in the app host (dotnet/aspire#6577)](https://github.com/dotnet/aspire/pull/6577/).

SteveSandersonMS:
OK, since we got the updated build of `Microsoft.Extensions.AI` that supports the API patterns needed for the design we want, this PR is now hopefully done.

> A principle of .NET Aspire is that telemetry is enabled by default

This PR now aligns with that principle. I've made all the new APIs enable otel for `IChatClient` and `IEmbeddingGenerator<T, U>` by default. You can explicitly opt out, e.g.:

```cs
builder.AddAzureOpenAIClient("openai").AddChatClient(disableOpenTelemetry: true);
```

... and then if you want you can call `UseOpenTelemetry()` manually at some other point in the `IChatClient` pipeline. By default the open telemetry step is immediately before the inner client in the pipeline, which is almost always what a developer would want.

> Have we given any more thought to the Hosting integration APIs? Specifically @sebastienros's proposal at https://github.com/dotnet/aspire/pull/6577/.

I'm totally happy with that and think we should do it. I've already given comments on that PR.

I would like to merge this PR first and then @sebastienros is free to rebase on that. This has already been an extremely long-lived PR so it would be great to close this out.

CC @eerhardt @sebastienros @stephentoub for any final review. If there aren't objections I'll merge this soon since I think we've already been through enough design iterations to believe people are on board with this, and there was nothing surprising that came up in the final implementation.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIClientBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+using Aspire.OpenAI;
+using Azure.AI.OpenAI;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Azure.AI.OpenAI;
+
+/// <summary>
+/// A builder for configuring an <see cref="AzureOpenAIClient"/> service registration.
+/// </summary>
+public class AspireAzureOpenAIClientBuilder : AspireOpenAIClientBuilder
```
SteveSandersonMS, line 13: The point of this subclass is just so that, if we want to add **Azure** OpenAI-specific info to the builder in the future, we can do so non-breakingly.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// <param name="builder">An <see cref="AspireOpenAIClientBuilder" />.</param>
+    /// <param name="serviceKey">The service key with which the <see cref="IChatClient"/> will be registered.</param>
+    /// <param name="deploymentName">Optionally specifies which model deployment to use. If not specified, a value will be taken from the connection string.</param>
+    /// <param name="disableOpenTelemetry">Optional. If <see langword="true"/>, skips registering open telemetry support in the <see cref="IChatClient"/> pipeline.</param>
+    /// <returns>A <see cref="ChatClientBuilder"/> that can be used to build a pipeline around the inner <see cref="IChatClient"/>.</returns>
+    public static ChatClientBuilder AddKeyedChatClient(
+        this AspireOpenAIClientBuilder builder,
+        string serviceKey,
+        string? deploymentName = null,
+        bool disableOpenTelemetry = false)
```
stephentoub, line 49: I find the double-negatives a bit more challenging to reason about. Should this be enableOpenTelemetry = true?

Either way is ok, though.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+    private const string ModelKey = "Model";
+
+    /// <summary>
+    /// Registers a singleton <see cref="IChatClient"/> in the services provided by the <paramref name="builder"/>.
+    /// </summary>
+    /// <param name="builder">An <see cref="AspireOpenAIClientBuilder" />.</param>
+    /// <param name="deploymentName">Optionally specifies which model deployment to use. If not specified, a value will be taken from the connection string.</param>
+    /// <param name="disableOpenTelemetry">Optional. If <see langword="true"/>, skips registering open telemetry support in the <see cref="IChatClient"/> pipeline.</param>
+    /// <returns>A <see cref="ChatClientBuilder"/> that can be used to build a pipeline around the inner <see cref="IChatClient"/>.</returns>
+    public static ChatClientBuilder AddChatClient(
```
stephentoub, line 28: Should this add an OpenAI.ChatClient in addition to an IChatClient?

stephentoub (approved):

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: I was following the "optional flags should default to false" guidebook here. Even though the negation is clumsy, it makes it more obvious what the default is. A lot of ASP.NET Core flags work this way.

Totally willing to change this round if that's your preference. If you don't feel strongly I'll leave it as-is.

stephentoub: > If you don't feel strongly I'll leave it as-is.

I don't.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// Provides extension methods for registering <see cref="IChatClient"/> as a singleton in the services provided by the <see cref="IHostApplicationBuilder"/>.
+/// </summary>
+public static class AspireOpenAIClientBuilderChatClientExtensions
+{
+    private const string DeploymentKey = "Deployment";
+    private const string ModelKey = "Model";
```
eerhardt, line 19: Why are we reading both `Deployment` and `Model` keys? Should we just pick one and go with it?

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderEmbeddingGeneratorExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        deploymentName ??= GetRequiredDeploymentName(builder);
+        var result = openAiClient.AsEmbeddingGenerator(deploymentName);
+
+        return disableOpenTelemetry
+            ? result
+            : new OpenTelemetryEmbeddingGenerator<string, Embedding<float>>(result);
+    }
+
+    private static string GetRequiredDeploymentName(AspireOpenAIClientBuilder builder)
```
eerhardt, line 74: Can we refactor this code so it isn't duplicated between these two builder classes?

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
sebastienros: This is also what we decided for all our settings. Telemetry is enabled by default in Aspire components. Here is what we have:

```
    /// <summary>
    /// Gets or sets a boolean value that indicates whether the OpenTelemetry tracing is disabled or not.
    /// </summary>
    /// <value>
    /// The default value is <see langword="false"/>.
    /// </value>
    public bool DisableTracing { get; set; }
```

eerhardt, line 49: How does this `disableOpenTelemetry` parameter compose with with the property on the root `OpenAISettings`?

https://github.com/dotnet/aspire/blob/ef415697283fff284e908f983b8d132d714a6630/src/Components/Aspire.OpenAI/OpenAISettings.cs#L44-L52

If I set `OpenAISettings.DisableTracing = true`, should that affect what happens here?

sebastienros, line 49: For consistency with other components this might be called `disableTracing`.
Though we usually don't have these in extension methods but in settings classes (configuration binding) with an extra lambda to configure by code.

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/UseIChatClient.razor
```diff
@@ -0,0 +1,37 @@
+﻿@page "/useichatclient"
```
eerhardt, line 1: Can this just go in the existing `Home.razor`? I'm not sure why we'd have/need both.

SteveSandersonMS: `Home.razor` is for using `OpenAIChatClient`. This one is for using `IChatClient`.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderEmbeddingGeneratorExtensions.cs
SteveSandersonMS: I can factor out a shared utility for it.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: Because OpenAI prefers "model" terminology while Azure OpenAI prefers "deployment". The intent is that developers don't need to worry about getting the wrong one, because at any given time they can specify either but not both. If we picked a single one, it may seem weird in cases where it doesn't match the backend you're using.

This isn't something I feel very strongly about, and we could force people to pick the key that corresponds to the backend they're using. It just seems like adding an extra pit of failure for no reason.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
sebastienros: You can skip this argument if the method respects the setting from the higher level registration. I think it's fine to set it once for all clients.

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/UseIChatClient.razor
eerhardt: I think we should just move this logic into Home.razor. I don't see a need to have both.

Also, I pulled the changes locally and this page doesn't appear to work:

```
2024-11-26T12:32:38       System.InvalidOperationException: An IChatClient could not be configured. Ensure a 'Deployment' or 'Model' value is provided in 'ConnectionStrings:openai', or specify a 'Deployment' in the 'Aspire:Azure:AI:OpenAI' configuration section, or specify a 'deploymentName' in the call.
```

I assume we need to change the AppHost

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
sebastienros: I get confused every time I work on my related PR. Thanks to the explanation.

eerhardt: Maybe at least stick this commentary into a code comment for the next person so they understand why both names are used.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: I'll see if I can get the `OpenAISettings.DisableTracing` flag to control this.

That seems reasonable though it's arguably a bit clunky because the meaning doesn't align perfectly. We're never going to **disable** tracing in the sense of "block it from working". We only want a way to signal "don't auto-add the tracing middleware, and then I might add it manually in a custom place".

I think the slight mismatch of meaning isn't a big problem and if we can control this via `OpenAISettings.DisableTracing` it seems good to me. Not certain yet whether that fits the layering though.

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/UseIChatClient.razor
sebastienros: > I assume we need to change the AppHost

My PR is definitely changing how these settings are set in the app host.

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: Done

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderEmbeddingGeneratorExtensions.cs
SteveSandersonMS: Done

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/UseIChatClient.razor
SteveSandersonMS: > Also, I pulled the changes locally and this page doesn't appear to work:

Thanks. I've updated the E2E code so that it uses the same technique to get the deployment name in the new E2E page as it did in the old one, so it should work consistently for the two.

> My PR is definitely changing how these settings are set in the app host.

Makes sense!

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: Done. This is definitely better. Thanks for the suggestion!

SteveSandersonMS: > You can skip this argument if the method respects the setting from the higher level registration

That is now done.

SteveSandersonMS: No longer applicable as it now uses the higher-level config for this (instead of an explicit parameter).

src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderChatClientExtensions.cs
SteveSandersonMS: I tried implementing this (and we can do so) but am not sure it's the right thing to do. Two issues:

1. `Aspire.OpenAI` and `Aspire.Azure.AI.OpenAI` don't have a convention that there should be a `OpenAI.ChatClient` in DI. Instead they expect there only to be an `OpenAIClient`. We would be making up a new convention for that, and it's not clear from the API shape that the developer is asking us to register anything other than an `IChatClient`. 
3. It may go wrong if the developer registers multiple unkeyed (or same-keyed) `IChatClient` instances in DI. The registration for `OpenAI.ChatClient` would have to be a factory that reads the current `IChatClient` from DI and then call `GetService<OpenAI.ChatClient>()` on it (because the `OpenAI.ChatClient` instance is produced internally by MEAI's `OpenAIChatClient`). If the developer overwrote the OpenAI-specific `IChatClient` with a non-OpenAI `IChatClient` in their DI registry, that lookup would fail. That is, they would have a service collection entry for the `OpenAI.ChatClient` type, but resolving it would fail for what seems like an unrelated reason.

Altogether, given that it's not an existing convention to register a singleton `OpenAI.ChatClient` in DI, and since any developer who wants to get the `OpenAI.ChatClient` corresponding to their `IChatClient` easily can do so [1], I would suggest we don't add a further behavior to document and maintain here. At least not now - it's easier to add this in the future than it would be to remove it.

[1] Either by calling `chatClient.GetService<OpenAI.ChatClient>()` at the usage site, or by registering it as a singleton with `builder.Services.AddSingleton(s => s.GetService<IChatClient>().GetService<OpenAI.ChatClient>())`.

SteveSandersonMS:
I believe this addresses all the remaining feedback (except possibly [this](https://github.com/dotnet/aspire/pull/6225#discussion_r1858792224)). We've been through enough design iterations to be sure everyone's on board, and there's an approval already. Since most people will now be out for most of a week, I'll merge now rather than hanging around potentially for a long time for further approvals.

If anyone has further comments please let me know!

eerhardt:
Thanks for the great contribution here, @SteveSandersonMS. I think we are in a much better place w.r.t. supporting OpenAI in .NET Aspire, and I'm excited to see what people build with this functionality.

PR #6230, opened 2024-10-10
Show exception details in text visualizer dialog

### LINKED ISSUE #6130: Remove exception details UX now that we can view log message
image

We have https://github.com/dotnet/aspire/pull/6064. We can consider a top level button to view the log message

drewnoakes:
## Description

Fixes #6130

Exception details are shown behind a button in two places:

1. In the "Message" column of the structured logs table, when an exception exists.
2. In the "Health reports" grid in resource details, when a health report includes exception details.

Previously, the stack trace would be shown in a popup when the mouse hovered over the icon, sort of like a tooltip.

This commit changes the behaviour so that exception details are shown in the existing text visualizer dialog whenever the icon is clicked.

The icon becomes a `FluentButton` for consistency with the menu button to the right of the cell.

Also some optimisations in `GridValue`:

- Don't use `FluentHighlighter` when the text to highlight is empty, which will be the common case.
- Remove `MaxDisplayLength`. We never set this property, and it used space on every `GridValue` instance, and we create a lot of these objects.
- Remove some redundant DOM elements
- Support adding arbitrary buttons to the right-hand side of the area (used for the "exception details" button)

---

Here's how it looks for health checks:

image

Here's how it looks on mouse-over, along with a tooltip. The button appears too close to the text here, but that's consistent with other inline lightweight buttons. It only shows on mouse over. If we add more margin, it looks bad when the mouse isn't hovering over it.

image

Clicking the button opens the text visualizer we use throughout the dashboard:

image

Here's how it looks in structured logs:

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6230)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
+++ b/src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
@@ -1,52 +1,35 @@
 ﻿@namespace Aspire.Dashboard.Components
 
 @using Aspire.Dashboard.Extensions
+@using Aspire.Dashboard.Components.Dialogs
 @using Aspire.Dashboard.Otlp.Model
 @using Aspire.Dashboard.Resources
 
 @inject IStringLocalizer<ControlsStrings> Loc
 
-@{
-    var iconId = Guid.NewGuid().ToString();
-    @*
-     * We don't want a browser tooltip to display when hovering over the error icon.
-     * Use alt instead of title for screen readers. Also, the column has a title set
-     * so add a whitespace title to prevent a browser tooltip from displaying.
-     *@
-    var title = " ";
-}
-
 @if (!string.IsNullOrWhiteSpace(ExceptionText))
 {
-    <FluentIcon alt="@Loc[nameof(ControlsStrings.ExceptionDetailsTitle)]"
-                title="@title"
-                Style="flex-shrink: 0"
-                Id="@iconId"
-                Icon="Icons.Filled.Size16.DocumentError"
-                Color="Color.Accent"
-                Class="severity-icon" />
-
-    <FluentTooltip Anchor="@iconId" MaxWidth="650px">
-        <div style="padding: 10px">
-            <h4>@Loc[nameof(ControlsStrings.ExceptionDetailsTitle)]</h4>
-
-            <pre style="white-space: pre-wrap; overflow-wrap: break-word; max-height: 300px; overflow: auto;">@ExceptionText</pre>
-
-            <div style="float: right; margin-bottom: 10px;">
-                <FluentButton Id="@Guid.NewGuid().ToString()"
-                              Appearance="Appearance.Lightweight"
-                              AdditionalAttributes="@FluentUIExtensions.GetClipboardCopyAdditionalAttributes(ExceptionText, Loc[nameof(ControlsStrings.GridValueCopyToClipboard)].ToString(), Loc[nameof(ControlsStrings.GridValueCopied)].ToString())">
-                    @Loc[nameof(ControlsStrings.GridValueCopyToClipboard)]
-
-                    <FluentIcon Class="copy-icon align-text-bottom" Style="display:inline;" Icon="Icons.Regular.Size16.Copy" />
-                    <FluentIcon Class="checkmark-icon align-text-bottom" Style="display:none;" Icon="Icons.Regular.Size16.Checkmark" />
-                </FluentButton>
-            </div>
-        </div>
-    </FluentTooltip>
+    <FluentButton Appearance="Appearance.Lightweight"
+                  OnClick="OnExceptionDetailsClicked"
+                  Title="@Loc[nameof(ControlsStrings.ExceptionDetailsTitle)]"
+                  Class="exception-details-button">
+        <FluentIcon Icon="Icons.Filled.Size16.DocumentError"
+                    Color="Color.Accent" />
+    </FluentButton>
 }
 
 @code {
     [Parameter, EditorRequired]
     public required string ExceptionText { get; set; }
+
+    [CascadingParameter]
+    public required ViewportInformation ViewportInformation { get; init; }
+
+    [Inject]
+    public required IDialogService DialogService { get; init; }
+
+    private void OnExceptionDetailsClicked(MouseEventArgs e)
+    {
+        _ = TextVisualizerDialog.OpenDialogAsync(ViewportInformation, DialogService, Loc[nameof(ControlsStrings.ExceptionDetailsTitle)], ExceptionText);
+    }
 }
--- a/src/Aspire.Dashboard/Components/Controls/GridValue.razor.css
+++ b/src/Aspire.Dashboard/Components/Controls/GridValue.razor.css
@@ -9,7 +9,7 @@
     grid-template-columns: 1fr auto auto;
 }
 
-::deep .cellText {
+::deep .grid-value {
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
@@ -19,9 +19,10 @@
     opacity: 0;
     cursor: pointer;
     width: 0;
+    display: inline-block;
 }
 
-::deep:hover .defaultHidden, ::deep:focus-within .defaultHidden  {
+::deep:hover .defaultHidden, ::deep:focus-within .defaultHidden {
     opacity: 1;  /* safari has a bug where hover is not always called on an invisible element, so we use opacity instead */
     width: auto;
 }
--- a/src/Aspire.Dashboard/Components/Controls/GridValue.razor
+++ b/src/Aspire.Dashboard/Components/Controls/GridValue.razor
@@ -3,25 +3,27 @@
 @inject IStringLocalizer<Dialogs> DialogsLoc
 
 <div class="@GetContainerClass()" style="width: inherit;">
+
+    @* Value area *@
+
     @if (EnableMasking && IsMasked)
     {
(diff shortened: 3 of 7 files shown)


davidfowl:
Yes please!

davidfowl (approved):

src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
```diff
@@ -1,52 +1,35 @@
...
-                              AdditionalAttributes="@FluentUIExtensions.GetClipboardCopyAdditionalAttributes(ExceptionText, Loc[nameof(ControlsStrings.GridValueCopyToClipboard)].ToString(), Loc[nameof(ControlsStrings.GridValueCopied)].ToString())">
-                    @Loc[nameof(ControlsStrings.GridValueCopyToClipboard)]
-
-                    <FluentIcon Class="copy-icon align-text-bottom" Style="display:inline;" Icon="Icons.Regular.Size16.Copy" />
-                    <FluentIcon Class="checkmark-icon align-text-bottom" Style="display:none;" Icon="Icons.Regular.Size16.Checkmark" />
-                </FluentButton>
-            </div>
-        </div>
-    </FluentTooltip>
+    <FluentButton Appearance="Appearance.Lightweight"
```
JamesNK, line 12: It doesn't look good on mouse over with the button hard up against the text:

image

Reduce the width of the button (maybe have 4px?) and add a right margin to provide some space between the button and text.

Example from the inline error button on resources page:

https://github.com/dotnet/aspire/blob/fe53c9874e5db7e64be5221b5f41c0ea1d9203b0/src/Aspire.Dashboard/Components/Pages/Resources.razor.css#L6-L11

JamesNK, line 12: Clicking on the button opens and closes the log details because the click event propagates to the row click handle. Stop that with stopPropagation. Example:

https://github.com/dotnet/aspire/blob/fe53c9874e5db7e64be5221b5f41c0ea1d9203b0/src/Aspire.Dashboard/Components/ResourcesGridColumns/UnreadLogErrorsBadge.razor#L12

If the attribute doesn't work on the button, then add a span/div around the button with the attribute.

src/Aspire.Dashboard/Components/Controls/GridValue.razor.css
```diff
@@ -9,19 +9,23 @@
...
 }
 
-::deep .cellText {
+::deep .grid-value {
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
 }
 
+::deep .grid-value > .exception-details-button {
```
JamesNK, line 18: IMO add `exception-details-button` styles to global app.css.

The component isn't specific to grid value - it could be placed anywhere - it just happens to only be used in grid value at the moment.

JamesNK (changes requested):

src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
drewnoakes: I called this out in:

> Here's how it looks on mouse-over, along with a tooltip. The button appears too close to the text here, but that's consistent with other inline lightweight buttons. It only shows on mouse over. If we add more margin, it looks bad when the mouse isn't hovering over it.

Making this button smaller means it no longer matches the equivalent button on the right. We should be consistent.

JamesNK: Do you mean the `...` button that appear on hover? I wouldn't call it inline. It appears over the top of content on hover. The existing prior art of a clickable icon that is inline in text is the unread logs errors button. It doesn't have padding, just a margin to give it space from text (CSS above).

The problem with having the large padding is content isn't consistently aligned with the start of the column. It doesn't look good. And it breaks consistency with other columns that start with an icon (e.g. the level column also in the structured logs table).

Aspire today:

image

Current PR:

image

Appearance wise it should stay the same, with the change in behavior from hover to click.

cc @davidfowl @leslierichardson95 for input.

davidfowl: Honestly, I would prefer if we removed that little box and moved it to where the actions are on the left. Similar to how start/stop look on the resources view.

JamesNK: Hmm, that's an interesting idea. One problem I see with that is if there is an error then the `...` actions button would have to always be shifted right to make room for the "View exception details" button. But a lot of the time there won't be any logs exception details visible in the UI, so the `...` button will look like it is in the wrong place.

Here is an alternative idea:

What about changing the button in this PR to just a regular icon. No hovering, no clicking. Then in the actions menu, there is a "Exception details" option. There is already an menu option for viewing the message in the text visualizer. Would be simple to do the same thing for the exception details.

Imagine an extra option for "Exception details" with its matching icon in the picture below:
image

davidfowl: Prefer if it was aligned. A button without a tooltip feels bad though. It's hard to tell what that icon even is until you hover over it. @DamianEdwards was lamenting that the experience of viewing errors in the structured logs felt difficult compared to looking at console logs.

I think the core fix here is to add the menu item to view the exception details in the grid view *and* in the details view. If we can do the same with the icon then great, but its less discoverable IMO.

DamianEdwards: How about changing the status icon instead of resources with unread errors, rather than trying to fit an extra icon in somewhere else?

davidfowl:
I would love this in 9

src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
drewnoakes: Moved to a container for all buttons in `GridValue`, so it applies consistently.

src/Aspire.Dashboard/Components/Controls/GridValue.razor.css
drewnoakes: Further changes made this rule unnecessary.

src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
drewnoakes: I moved the button to the right hand side of the "cell" so that it sits alongside the masking button and the ... dropdown menu button. There's a generic facility for us to add other buttons there in future.

drewnoakes: See updated screenshots in description.

drewnoakes:
@JamesNK @davidfowl @DamianEdwards I moved the icon to the right side, so it sits with the masking button and the "..." dropdown button. See updated screenshots in PR description.

JamesNK (approved):
I like it.

One extra suggestion is to add an action to the menu here:
image

If there is an exception message then add a menu item with `<icon> Exception details` and launches the same visualizer. The `Log message` menu item already does the same thing.

If it is difficult then create a follow up issue. Stick with this for now and add the menu item in 9.1.

drewnoakes:
Adding a menu item makes sense, though it might reasonably be expected in the `GridValue`'s own drop-down menu. That'd make it work in the Health Checks area too. Any thoughts?

I was tempted to pull out an abstraction for these buttons and menu items, to take a more data-driven approach. I would rather do that in 9.1, to keep the scale of the change down.

JamesNK:
I found a regression:
* Previously the content in the grid cell would use the full width. Content would shrink when needed to make room for the `...` button
* Now the content is always strunk to make room for the `...` button.

Before:
log-button-main

After:
log-button-branch

JamesNK:
> Adding a menu item makes sense, though it might reasonably be expected in the `GridValue`'s own drop-down menu. That'd make it work in the Health Checks area too. Any thoughts?

We'd need to add a way to add items to the grid value menu. It's doable, and I like the consistency, but not as high value. Structured logs will get a lot of use.

Could do it properly in 9.1.

drewnoakes:
> Previously the content in the grid cell would use the full width. Content would shrink when needed to make room for the ... button. Now the content is always strunk to make room for the ... button.

Fixed in 340124ae04b989794cb14bb7d01f4a031d7f9170.

drewnoakes:
@JamesNK what do you think of this: 

image

The filled icon looks a bit strange, but there's no regular version of that icon as far as I can tell.

JamesNK:
* I prefer the text as `Exception details`. That's consistent with `Log message`. I don't know why I like this system, but not having `View details` or consistently having `View xxx` for all items felt worse.
* Move `Log details` above `Exception details`. That way there is consistent order of items. Always `View details` and `LogMessage` then an optional `Exception details`

drewnoakes:
I'll do the menu item in a follow-up PR for 9.1.

davidfowl (approved):

PR #6241, opened 2024-10-11
Ensure grid item keys are unique

JamesNK:
## Description

There are places where collections of attributes are merged together and can have the same name. This is a problem because the name is used as a row key, and duplicate row keys will cause errors.
There are also some places where "known" attributes exclude user defined attributes with the same name from being displayed. In other words, the first attribute with a key wins, and others are discarded.

This PR introduces internal keys that are used by attributes. The internal key allows known and unknown attributes to sit side-by-side.

Fix for 9.0.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6241)


src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -281,12 +286,15 @@ public UrlViewModel(string name, Uri url, bool isInternal)
...
+public sealed record class VolumeViewModel(int index, string Source, string Target, string MountType, bool IsReadOnly) : IPropertyGridItem
 {
-    string? IPropertyGridItem.Name => Source;
-
+    string IPropertyGridItem.Name => Source;
     string? IPropertyGridItem.Value => Target;
 
+    // Source could be empty for an anomymous volume so it can't be used as a key.
+    // Because there is no good key in data, use index of the volume in results.
+    object IPropertyGridItem.Key => index;
```
JamesNK, line 296: Confirmed that using source is a problem when there are multiple anonymous volumes:

image

I looked at the data turned from DCP today, and the k8s API spec - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#volumemount-v1-core - and there doesn't seem to be a better id.

drewnoakes (approved):
LGTM. Involves a few more allocations in places (boxing, temp strings), but I can't see an easy way around that.

PR #6250, opened 2024-10-11
Add CORS allowed origins config override to app host

### LINKED ISSUE #6249: AppHost Dashboard CORS Configuration

There was a discussion here [Issue 4917](https://github.com/dotnet/aspire/issues/4917) about adding CORS configuration to the Dashboard when run Standalone and via the AppHost.  The work appears to have been actioned here [PR 5212](https://github.com/dotnet/aspire/pull/5212).  From my reading of this, it looks like when running Standalone the DASHBOARD__OTLP__CORS__ALLOWEDORIGINS env var (see [Docs](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/configuration?tabs=bash#otlp-cors)) can be used to set the origins to something suitable for the use case.  However when the AppHost starts the Dashboard, ALLOWEDORIGINS is automatically set based on the endpoints configured in the AppHost's resources.

If the above is true, this creates an unsurmountable problem when using a Resource in the AppHost which has a custom local domain name.  It is currently not possible to configure the AppHost to use a custom domain name when defining resources - they will get a localhost:[RandomPort] instead (this limitation has been discussed several times in the Issues list).  This will presumably be auto-added to the Dashboard CORS origins list when the AppHost starts up.  But if that Resource is a SPA with a custom domain (manually added to the local hosts file, and using a self-signed cert), telemetry sent to the Dashboard will be blocked by CORS.  This prevents the use of browser-sourced telemetry in these scenarios.

Tagging @JamesNK due to the nature of the work and his existing involvement, question was asked originally on the docs repo here - [Issue 1588](https://github.com/dotnet/docs-aspire/issues/1588#issuecomment-2405073806).

Many thanks.

Possible options - allow the use of custom domains in WithEndpoint() style builder methods (which would presumably then be passed on in the auto-CORS behaviour).  Or allow the use of DASHBOARD__OTLP__CORS__ALLOWEDORIGINS (or similar) to **add** to the auto-CORS origin list when running the Dashboard via AppHost.

### Additional context

_No response_

JamesNK:
## Description

The dashboard supports OTLP HTTP and CORS to get telemetry from browser apps. CORS allowed origins are automatically calculated based on resource endpoints.

However, there are some scenarios where automatic CORS endpoints don't work:

- When browsing to the internal, proxy endpoint for a resource.
- When a custom domain is added to the host file and pointed to localhost, browsing to the custom domain.

These origins aren't automatically added and so sending telemetry fails.

The fix in this PR is to add manual dashboard CORS configuration to the app host. When the dashboard resource is configured, the app host manual configuration is simply forwarded to the dashboard. If there is no dashboard configuration then the existing logic is used.

Fixes https://github.com/dotnet/aspire/issues/6249

Fixes customer bug with existing feature. ~Intended for 9.0.~

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6250)

### DIFF
--- a/playground/BrowserTelemetry/BrowserTelemetry.AppHost/Program.cs
+++ b/playground/BrowserTelemetry/BrowserTelemetry.AppHost/Program.cs
@@ -4,7 +4,8 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
 builder.AddProject<Projects.BrowserTelemetry_Web>("web")
-    .WithExternalHttpEndpoints();
+    .WithExternalHttpEndpoints()
+    .WithReplicas(2);
 
 #if !SKIP_DASHBOARD_REFERENCE
 // This project is only added in playground projects to support development/debugging
--- a/playground/BrowserTelemetry/BrowserTelemetry.AppHost/Properties/launchSettings.json
+++ b/playground/BrowserTelemetry/BrowserTelemetry.AppHost/Properties/launchSettings.json
@@ -11,7 +11,8 @@
         "DOTNET_ENVIRONMENT": "Development",
         "DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL": "https://localhost:16175",
         "DOTNET_RESOURCE_SERVICE_ENDPOINT_URL": "https://localhost:17037",
-        "DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES": "true"
+        "DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES": "true",
+        "DOTNET_DASHBOARD_CORS_ALLOWED_ORIGINS": "*"
       }
     },
     "http": {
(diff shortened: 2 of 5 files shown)


davidfowl:
This is a feature....

JamesNK:
> This is a feature....

Because it is a new config setting? IMO it's a bug fix for scenarios where browser telemetry feature doesn't work.

If it doesn't meet the bar then it can be moved to post 9.0.

adamint (approved):

JamesNK:
@mitchdenny Can you take a look? The changes are all in the app host.

mitchdenny:
Implementations wise I have no problem with this. The only thing that gives me pause is whether it might be better to have some kind of mechanism to add to the computed origins list rather than completely replacing it.

That said, I think that given this only impacts the OTLP endpoint means its probably not much of a security concern. After all most telemetry endpoints are fairly permissive anyway and this is still just locally bound.

We should probably mention this in the threat model?

JamesNK:
> The only thing that gives me pause is whether it might be better to have some kind of mechanism to add to the computed origins list rather than completely replacing it.

I think it is simplest to understand if it is a replace. That means setting `DOTNET_DASHBOARD_CORS_ALLOWED_ORIGINS` on the host is equivilent to setting `DOTNET_DASHBOARD_CORS_ALLOWED_ORIGINS` on the dashboard.

If it turns out we want a third option we could look at that as a different setting.

PR #6255, opened 2024-10-11
[release/9.0-rc1] Template changes to consolidate to single package

joperezr:
## Description

The current experience from the CLI for creating new Aspire 9.0 projects is not ideal, especially if you already have the workload installed. The first problem, is that the new Aspire 9.0 templates are not overriding the ones from the workload, and instead they are installed side by side. This presents an issue when trying to create new projects from the command line, as the template engine doesn't know which one to pick as it can be seen in the following example:

image

This is confusing to users as it is not obvious for them knowing which one to pick. The second problem, is that the list of templates now shows two entries for templates, and it is also not clear which one is the one that the user wants:

image

The third problem, is that the new 9.0 templates only bring support for `net9.0` TFM, so if users want to stay targeting a .NET LTS version, they need to install an additional Template pack (`Aspire.ProjectTemplates.net8`) which is not ideal as it is not discoverable. Once they install that second template pack, then they can create templates targeting `net8.0` using the `-f net8.0` syntax like:

image

### Solution

The fix in this PR is making it such that all of the templates are in a single package, meaning the `net9.0` and `net8.0` templates for Aspire 9 (fixing problem no. 3 above), but additionally, it also provides 'downlevel' templates that will fix the conflict on shortnames (causing problem no. 1 above) as well as the description so that it is easier to understand which one to pick (problem no. 2 above). The result after these changes is this:

image
(Aspire templates description now clearly call out the version of Aspire so user knows which one to pick)

image
(Since template shortnames no longer collide, then the engine knows which template to use)

image
(Without needing to install an additional template pack, users can now also target `net8.0`)

cc: @davidfowl @DamianEdwards @radical @eerhardt @phenning @timheuer 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1667
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6255)


joperezr:
@balachir FYI

eerhardt (approved):

radical (approved):

Issue #6174, opened 2024-10-07
Clean up Azure Hosting APIs


eerhardt:
We should do an API Review for the Azure Hosting APIs. Specifically:

- Remove the Experimental callbacks, for example
https://github.com/dotnet/aspire/blob/a8b1a619d67549243342dfd2c92cfaf26905cb65/src/Aspire.Hosting.Azure.Storage/AzureStorageExtensions.cs#L42-L43
    - Instead, callers will call `builder.AddAzureStorage("storage").ConfigureConstruct(c => ...)` if they want to customize the Azure resource

- Rename `ResourceModuleConstruct` to `AzureResourceInfrastructure` since it inherits from `Azure.Provisioning.Infrastructure`.

- Rename `AzureConstructResource` to `AzureProvisioningResource`

- Rename `ConfigureConstruct` to `ConfigureInfrastructure`
    - This matches the Azure.Provisioning naming of this type.

We should also take a holistic view of all our Azure Hosting APIs and do an API Review on them to make sure all the things we are shipping for 9.0 make sense for the long term.

cc @mitchdenny @davidfowl @tg-msft

eerhardt:
* Expose `ContainerAppCustomizationAnnotation` publicly - cc @davidfowl

davidfowl:
Rename `AddContainerAppsInfrastructure` to `AddAzureContainerAppsInfrastructure`.

PR #6258, opened 2024-10-11
Remove AZPROVISION001 and Experimental Callbacks

eerhardt:
## Description

With the current version of Azure.Provisioning, there is no longer an experimental warning for AZPROVISION001. We can remove our suppressions everywhere for this warning.

In doing this, we are also removing the experimental callbacks in the Aspire.Hosting.Azure APIs. These APIs need more design. We have ideas on how to make them better in the future, but their current shape isn't it.

The places where the configure callbacks stay:

* PublishAsAzureContainerApp extensions. These APIs are no longer experimental.

Contributes to #6174

## Checklist

- Is this feature complete?
  - [x] No. Follow-up changes expected. More Azure API cleanup expected. See issue.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] Yes

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6258)


davidfowl (approved):

tests/Aspire.Hosting.Azure.Tests/AzureBicepResourceTests.cs
```diff
@@ -1412,22 +1390,12 @@ public async Task AsAzurePostgresFlexibleServerViaPublishMode()
...
-            Assert.NotNull(resource);
-            azurePostgres = resource;
-        });
+        var postgres = builder.AddPostgres("postgres", usr, pwd).AsAzurePostgresFlexibleServer();
         postgres.AddDatabase("db", "dbName");
 #pragma warning restore CS0618 // Type or member is obsolete
 
         var manifest = await ManifestUtils.GetManifestWithBicep(postgres.Resource);
 
-        // Setup to verify that connection strings is acquired via resource connectionstring redirct.
```
eerhardt, line 1426: Should do the same thing here as the AsAzureSqlDatabase tests - get the azure resource from the annotation, and set its outputs.

Issue #6155, opened 2024-10-07
Ensure all containers are password protected


eerhardt:
Where possible, we should be enabling passwords on our containers.

The goal here is to provide a layer of redundancy. If ACA firewalling (or other system to from external access) fails, an external client will still not be able to access the container without knowledge of the authentication password.

The following containers don't have password protection:

- [ ] Garnet
- [ ] Kafka
- [x] MongoDB [Add authentication (username & password) to AddMongoDB (dotnet/aspire#5788)](https://github.com/dotnet/aspire/pull/5788)
- [ ] Nats
- [ ] Redis [Revisit Redis password options (dotnet/aspire#3838)](https://github.com/dotnet/aspire/issues/3838)
- [ ] Seq
- [ ] Valkey

Mrxx99:
For NATS I would like to contribute the password implementation, I already have explored and started work on this

Mrxx99:
> > For NATS I would like to contribute the password implementation, I already have explored and started work on this
> 
> Sounds great! Looking forward to the PR.

@eerhardt The PR #6259 is ready for review.

eerhardt:
@liammclennan @nblumhardt - any interest in adding password support for Seq.

nblumhardt:
@eerhardt sounds good!

The Seq container supports a couple of environment variables that would make this pretty easy.

One possible wrinkle; the variables are:

 * `SEQ_FIRSTRUN_ADMINUSERNAME` - username for the automatically-created admin user account; default is `admin`
 * `SEQ_FIRSTRUN_ADMINPASSWORD` - plain text password for same
 * `SEQ_FIRSTRUN_ADMINPASSWORDHASH` - alternative to the above, provided as a salted cryptographic hash, for marginally better security when passing through configuration systems
 * `SEQ_FIRSTRUN_REQUIREAUTHENTICATIONFORHTTPINGESTION - sets the "require authentication for ingestion" flag

But, these only apply when the container's storage is first created, so modifying them once the Seq container has been initialized/has persistent data will have no effect. Would this be surprising/break expectations compared with the way other containers are doing it?

eerhardt:
> But, these only apply when the container's storage is first created, so modifying them once the Seq container has been initialized/has persistent data will have no effect. Would this be surprising/break expectations compared with the way other containers are doing it?

[...]

PR #5788, opened 2024-09-19
Add authentication (username & password) to AddMongoDB

eerhardt:
## Description

Adds authentication (username & password) to AddMongoDB as optional parameters.

This is a continuation of #4370. I am not able to update the existing PR because it is from a protected branch (`main` of a fork).

Fix #4348

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [x] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [x] Yes
      - [ ] No
  - [ ] No
- Does the change make any security assumptions or guarantees?
  - [x] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [x] No - follows existing patterns
  - [ ] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5788)

PR #6259, opened 2024-10-11
Add password protection to NATS

Mrxx99:
## Description

This adds parameter support to NATS. The postgres implementation was used as inspiration for the implementation.
It uses the user info of the connection string URL to add user name and password, so the connections string looks like
`nats://nats:MGBKxjTAjX2stg2wzzjtVW@localhost:52166`.
Some NATS clients support this form natively (as can bee seen [here](https://docs.nats.io/using-nats/developer/connecting/userpass)) but the .NET one unfortunately does not. It would be nice if this support would be added directly to NATS.Net then that code could be removed from here again.

Contributes to: #6155

~~The `NatsServerResource` constructor is changed from `(string name)` to `(string name, ParameterResource? userName, ParameterResource password)` to be the same as the `PostgresServerResource`~~

Currently there is a inconsistency between the parameters of `AddNats(...)` and `AddPostgres(...)`.
The postgres method looks like `(name, userName = null, password = null, port = null)` and the Nats one is changed from `(name, port = null)` to (name, port = null, userName = null, password = null)', otherwise `AddNats("nats", 4222)` would break.

So there are 3 options:
a) keep it like it currently is, not breaking, but inconsistent to postgres and forces to use named arguments when user wants to specify password but no port.
b) change the order to be consistent with postgres, breaks existing code that specifies the port without named arguments, but is nicer to use when specifying password and no port
c) keep the existing constructor and add a new one that mimics the postgres one, the existing one would call into that. This would have all the benefits of the two options above and will only break on `AddNats("nats", default)` (which is unlikely to be used I think).

Edit: seems like all resources have the port at the end in this scenario, except MongoDb which has two constructors, but both have the ports after the name. Maybe it would make sense to also make it consistent for MongoDb and to change MongoDb to option c), should still be possible as the second constructor is not yet shipped.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/pull/1817
  - [ ] No


src/Components/Aspire.NATS.Net/AspireNatsClientExtensions.cs
```diff
@@ -92,7 +92,26 @@ NatsConnection Factory(IServiceProvider provider)
...
+                {
+                    AuthOpts = options.AuthOpts with
+                    {
+                        Username = user,
+                        Password = password
+                    }
+                };
+
+                // redact password so it is not visible in logs
+                redactedConnectionString = redactedConnectionString.Replace(password, "***");
```
Mrxx99, line 111: Now after I thought more about this, this can lead to issues if the password is "nats" or another string that is part of the URL. Will change this.

eerhardt:
> It uses the user info of the connection string URL to add user name and password, so the connections string looks like
nats://nats:MGBKxjTAjX2stg2wzzjtVW@localhost:52166.
Some NATS clients support this form natively (as can bee seen [here](https://docs.nats.io/using-nats/developer/connecting/userpass)) but the .NET one unfortunately does not. It would be nice if this support would be added directly to NATS.Net then that code could be removed from here again.

One interesting behavior is that now that the NATS Hosting has a health check, I am getting the user name and password printed out in my console log:

image

Is this expected behavior? That doesn't seem right.

cc @mtmk

src/Aspire.Hosting.Nats/PublicAPI.Shipped.txt
```diff
@@ -4,7 +4,6 @@ Aspire.Hosting.ApplicationModel.NatsServerResource.ConnectionStringExpression.ge
 Aspire.Hosting.ApplicationModel.NatsServerResource.NatsServerResource(string! name) -> void
 Aspire.Hosting.ApplicationModel.NatsServerResource.PrimaryEndpoint.get -> Aspire.Hosting.ApplicationModel.EndpointReference!
 Aspire.Hosting.NatsBuilderExtensions
-static Aspire.Hosting.NatsBuilderExtensions.AddNats(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name, int? port = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.NatsServerResource!>!
```
eerhardt, line 7: We shouldn't be / can't be making breaking changes. You can only add new API. See https://github.com/dotnet/aspire/pull/5788 for how this can be done.

eerhardt:
> Currently there is a inconsistency between the parameters of AddNats(...) and AddPostgres(...).

Yes, this is an issue, but not one that has a good solution. With MongoDB we decided to go with option (a) keep it like it currently is, not breaking, but inconsistent. This decision was made because we don't make breaking changes to public APIs we've already shipped.

src/Aspire.Hosting.Nats/NatsServerResource.cs
```diff
@@ -10,25 +10,65 @@ namespace Aspire.Hosting.ApplicationModel;
...
+        PasswordParameter = password;
+    }
+
     /// <summary>
     /// Gets the primary endpoint for the NATS server.
     /// </summary>
     public EndpointReference PrimaryEndpoint => _primaryEndpoint ??= new(this, PrimaryEndpointName);
 
+    /// <summary>
+    /// user name for the NATS server
```
eerhardt, line 39:
```suggestion
    /// Gets or sets the user name for the NATS server.
```

src/Aspire.Hosting.Nats/NatsServerResource.cs
```diff
@@ -10,25 +10,65 @@ namespace Aspire.Hosting.ApplicationModel;
...
+    /// </summary>
+    public ParameterResource? UserNameParameter { get; set; }
+
+    internal ReferenceExpression UserNameReference =>
+        UserNameParameter is not null ?
+            ReferenceExpression.Create($"{UserNameParameter}") :
+            ReferenceExpression.Create($"{DefaultUserName}");
+
+    /// <summary>
+    /// password for the NATS server
```
eerhardt, line 49:
```suggestion
    /// Gets or sets the password for the NATS server.
```

tests/Aspire.Hosting.Nats.Tests/AddNatsTests.cs
```diff
@@ -1,18 +1,108 @@
...
+        var connectionStringResource = natsResource as IResourceWithConnectionString;
+        Assert.NotNull(connectionStringResource);
+        var connectionString = await connectionStringResource.GetConnectionStringAsync();
+
+        Assert.Equal($"nats://nats:{natsResource.PasswordParameter?.Value}@localhost:4222", connectionString);
+        Assert.Equal("nats://nats:{nats-password.value}@{nats.bindings.tcp.host}:{nats.bindings.tcp.port}", connectionStringResource.ConnectionStringExpression.ValueExpression);
+    }
+
+    [Fact]
+    public async Task AddNatsCreatesConnectionStringWithUserAndPassword()
```
eerhardt, line 84: There's not much difference between this test and `AddNatsSetsUserNameAndPasswordOnNatsServerResource`. Can they be combined into a single test?

tests/Aspire.Hosting.Nats.Tests/AddNatsTests.cs
```diff
@@ -38,15 +128,30 @@ public void AddNatsContainerWithDefaultsAddsAnnotationMetadata()
...
     {
         var path = OperatingSystem.IsWindows() ? @"C:\tmp\dev-data" : "/tmp/dev-data";
 
         var appBuilder = DistributedApplication.CreateBuilder();
-        appBuilder.AddNats("nats", 1234).WithJetStream().WithDataBindMount(path);
+        appBuilder.Configuration["Parameters:user"] = "usr";
+        appBuilder.Configuration["Parameters:pass"] = "pass";
+
+        var user = appBuilder.AddParameter("user");
+        var pass = appBuilder.AddParameter("pass");
```
eerhardt, line 152:
```suggestion
        var user = appBuilder.AddParameter("user", "usr");
        var pass = appBuilder.AddParameter("pass", "pass");
```

tests/Aspire.Hosting.Nats.Tests/NatsFunctionalTests.cs
```diff
@@ -57,6 +58,100 @@ public async Task VerifyNatsResource()
...
+    public async Task AuthenticationShouldWork(string? user, string? password)
+    {
+        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+      
+        using var builder = TestDistributedApplicationBuilder.CreateWithTestContainerRegistry(testOutputHelper);
+        builder.Configuration["Parameters:user"] = user;
+        builder.Configuration["Parameters:pass"] = password;
+
+        var usernameParameter = user is null ? null : builder.AddParameter("user");
+        var passwordParameter = password is null ? null : builder.AddParameter("pass");
```
eerhardt, line 76:
```suggestion

        var usernameParameter = user is null ? null : builder.AddParameter("user", user);
        var passwordParameter = password is null ? null : builder.AddParameter("pass", password);
```

tests/Aspire.Hosting.Tests/Utils/CommandLineArgumentsEvaluator.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Tests.Utils;
+
+public static class CommandLineArgumentsEvaluator
```
eerhardt, line 6: Can we combine this with 

https://github.com/dotnet/aspire/blob/e516d8250ab4a507f03148b894e4ae0c5c398751/tests/Aspire.Hosting.Tests/Utils/ArgumentEvaluator.cs#L6-L8

It looks duplicated.

src/Components/Aspire.NATS.Net/AspireNatsClientExtensions.cs
```diff
@@ -92,7 +92,27 @@ NatsConnection Factory(IServiceProvider provider)
...
+                {
+                    AuthOpts = options.AuthOpts with
+                    {
+                        Username = uriBuilder.UserName,
+                        Password = uriBuilder.Password
+                    }
+                };
+
+                // redact password so it is not visible in logs
+                uriBuilder.Password = "***";
```
eerhardt, line 111: I think this should be logged as an issue on the NATS.Net library and fixed there. This doesn't make sense at this level.

Mrxx99: Since the NATS.Net library currently just ignores the user/password info in the url we could also just remove both here completely before passing it to the library. I just thought maybe seeing that it is a authenticated connection could be interesting in the logs, so I redacted instead of removed. For the health check connection we could also remove that user information before passing it to the connection.

Mrxx99:
> > Currently there is a inconsistency between the parameters of AddNats(...) and AddPostgres(...).
> 
> Yes, this is an issue, but not one that has a good solution. With MongoDB we decided to go with option (a) keep it like it currently is, not breaking, but inconsistent. This decision was made because we don't make breaking changes to public APIs we've already shipped.

So should I also use a) for Nats?. The password to mongodb was only introduced in aspire 9 RC, so not yet released, so could still be changed to c). Or is having an new amibguity for AddNats/Mongo("name", default) such a problem? Or (for mongo) is breaking change not possible even in RC?

playground/nats/Nats.AppHost/aspire-manifest.json
```diff
@@ -3,9 +3,13 @@
   "resources": {
     "nats": {
       "type": "container.v0",
-      "connectionString": "nats://{nats.bindings.tcp.host}:{nats.bindings.tcp.port}",
+      "connectionString": "nats://nats:{nats-password.value}@{nats.bindings.tcp.host}:{nats.bindings.tcp.port}",
```
mtmk, line 6: what's the point of adding user:pass in the url? NATS .NET client doesn't support this notation so it doesn't have a functional purpose.

src/Aspire.Hosting.Nats/NatsBuilderExtensions.cs
```diff
@@ -43,7 +49,15 @@ public static IResourceBuilder<NatsServerResource> AddNats(this IDistributedAppl
...
             };
 
-            options = options with { Url = connectionString };
+            options = options with
+            {
+                Url = connectionString,
+                AuthOpts = new()
+                {
+                    Username = await nats.UserNameReference.GetValueAsync(ct).ConfigureAwait(false),
+                    Password = nats.PasswordParameter!.Value,
```
mtmk, line 58: this is good correct way of passing user, pass to the client. no need to have it in the url

mtmk:
> > It uses the user info of the connection string URL to add user name and password, so the connections string looks like
> > nats://nats:MGBKxjTAjX2stg2wzzjtVW@localhost:52166.
> > Some NATS clients support this form natively (as can bee seen [here](https://docs.nats.io/using-nats/developer/connecting/userpass)) but the .NET one unfortunately does not. It would be nice if this support would be added directly to NATS.Net then that code could be removed from here again.
> 
> One interesting behavior is that now that the NATS Hosting has a health check, I am getting the user name and password printed out in my console log:
> 
> image
> 
> Is this expected behavior? That doesn't seem right.
> 
> cc @mtmk

I might be missing something here but there is no reason to pass user:pass in the url since it won't be used by the client, plus the client doesn't have safeguards to redact the password since it's not a supported method of passing username/password or tokens.

src/Components/Aspire.NATS.Net/AspireNatsClientExtensions.cs
mtmk: ok I think I am understanding this now. so there is no other way to pass the user pass but in the connection string?

Mrxx99:
@mtmk
> 
> I might be missing something here but there is no reason to pass user:pass in the url since it won't be used by the client, plus the client doesn't have safeguards to redact the password since it's not a supported method of passing username/password or tokens.

It is used because that's how aspire forwards information between resources. Environmental variables would also be possible, but the default aspire approach is using connection strings. The username and password are parsed out in the Aspire.NATS.Net integration package. The user/password information could be stripped out before passing the URL to NATS.Net. An advantage of having this connection string is also easy interopability with clients in other languages that already support this format. So it would be nice if NATS.Net would also support it.

src/Components/Aspire.NATS.Net/AspireNatsClientExtensions.cs
Mrxx99: Yes, connection string or extra environmental variables.

mtmk:
> @mtmk
> 
> > I might be missing something here but there is no reason to pass user:pass in the url since it won't be used by the client, plus the client doesn't have safeguards to redact the password since it's not a supported method of passing username/password or tokens.
> 
> It is used because that's how aspire forwards information between resources. Environmental variables would also be possible, but the default aspire approach is using connection strings. The username and password are parsed out in the Aspire.NATS.Net integration package. The user/password information could be stripped out before passing the URL to NATS.Net. An advantage of having this connection string is also easy interopability with clients in other languages that already support this format. It would be nice if NATS.Net would also support it directly.

@Mrxx99 thanks for the explanation 👍 
sounds like something to be implemented on the client also as @eerhardt suggested. feel free to start a [nats.net](https://github.com/nats-io/nats.net) PR I'm happy to prioritize its release.

mtmk:
@Mrxx99 NATS .NET with URL auth support released in v2.5.3 thank you for your contribution 👍
https://github.com/nats-io/nats.net/releases/tag/v2.5.3

Mrxx99:
@eerhardt The 2.5.3 version of Nats.NET (https://www.nuget.org/packages/NATS.Net/2.5.3) needs to be added to the dotnet-public feed so I can use it in this PR.

sebastienros:
> The 2.5.3 version of Nats.NET needs to be added to the dotnet-public feed so I can use it in this PR.

Started job, should be done in 10 minutes

eerhardt (approved):

PR #6260, opened 2024-10-12
Allow aspire dashboard run behind proxy with SSL terminate

### LINKED ISSUE #5134: Make it possible to run the dashboard behind a reverse proxy or load balancer

I would like to be able to host Aspire Dashboard behind a reverse proxy such as NGINX. Unfortunately, Aspire Dashboard doesn't take `X-Forwarded-*` headers into consideration, and doesn't allow to configure a path base.

As an example, assume Aspire Dashboard is accessible internally from http://aspire-dashboard-service:18888/ and externally from https://example.com/dev/aspire-dashboard/ through a reverse proxy.

The issue of the path base has already been described in #4528 and #4159. Additionally, authentication through OpenID Connect fails to set the correct value for `redirect_uri`, as it uses the internal schema (HTTP) rather than the external one (HTTPS).

It's possible to add support for reverse proxies and load balancers as described here:
https://learn.microsoft.com/aspnet/core/host-and-deploy/proxy-load-balancer

Ideally, all the featured described in previous page (except for certificate forwarding) should be configurable by the user, namely:
- Forwarded headers from the reverse proxy or load balancer
- Header names, known proxies and forward limit
- Forcing the assumption that external address is always HTTPS (necessary when the reverse proxy does not support `X-Forwarded-*` headers)
- Path base

### Additional context

_No response_

wu-yafeng:
## Description

When aspire run behind a proxy / load balancer with ssl terminate, aspire oidc authn will redirect user to IdP with parameter `redirect_uri=http://aspire.host`.

This PR allows aspire redirect with https `redirect_uri`.

Fixes #5134 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -283,6 +283,8 @@ public sealed class DashboardWebApplication : IAsyncDisposable
             .AddSupportedCultures(supportedCultures)
             .AddSupportedUICultures(supportedCultures));
 
+        _app.UseForwardedHeaders();
+
         WriteVersion(_logger);
 
         _app.Lifetime.ApplicationStarted.Register(() =>


davidfowl:
You can just set configuration and it’ll enable this middleware

wu-yafeng:
> You can just set configuration and it’ll enable this middleware

Thanks. I found the solution:

ForwardedHeader can be enabled by set environment `ASPNETCORE_FORWARDEDHEADERS_ENABLED` to `true`, more info see https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-8.0#forward-the-scheme-for-linux-and-non-iis-reverse-proxies

PR #6262, opened 2024-10-12
Reject spans that create a circular loop

JamesNK:
## Description

Spans have a parent. A span is not allowed reference itself or one of its anscetors as a parent.

Data sent to the dashboard could contain an invalid circular reference. Rendering the bad data in the UI could cause errors. This PR adds validation logic when spans are received to reject the bad data.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6262)


src/Aspire.Dashboard/Otlp/Model/OtlpTrace.cs
```diff
@@ -96,6 +101,31 @@ static string BuildFullName(OtlpSpan existingSpan)
         }
     }
 
+    private static bool HasCircularReference(OtlpSpan span)
```
mitchdenny, line 104: Is this called every single time a span is sent to the OTEL endpoint?

JamesNK: Yes

src/Aspire.Dashboard/Otlp/Model/OtlpTrace.cs
```diff
@@ -96,6 +101,31 @@ static string BuildFullName(OtlpSpan existingSpan)
...
+        {
+            return false;
+        }
+
+        // Walk up span ancestors to check there is no loop.
+        var stack = new List<OtlpSpan> { span };
+        var currentSpan = span;
+        while (currentSpan.GetParentSpan() is { } parentSpan)
+        {
+            if (stack.Contains(parentSpan))
```
drewnoakes, line 117: Is a `List<>` the right collection here? Seems like a `HashSet<>` would be a better fit, avoiding O(N) lookups.

Concerned about the space and time complexity of this change. As called out above, the method's called for every span, and it then walks the whole ancestry, and does an O(N) lookup within that. That's a lot of nested loops.

JamesNK: Changed to a keyed collection. Will adaptively switch to a dictionary when it exceeds a certain number of spans.

JamesNK:
PR feedback addressed.

drewnoakes (approved):

PR #6275, opened 2024-10-13
Experimental custom domain support for ACA.

mitchdenny:
## Description

Adds custom domain support to the playground sample. The approach used here is that you deploy first without custom domains enabled (just hit enter on the `certificateName` value but provide the desired custom domain name.

Then once you've deployed for the first time go into the portal and go through the binding process for the custom domain and create the appropriate domain ownership verification steps (creating TXT/CNAME records etc).

Once the domain is up and running then you can deploy again, but this time edit the `config.json` file with the name of the certificate (look in the container app environment).

Opening this as a PR for now for discussion since we'll probably want to wrap this up in some kind of API. The API would need to augment the callback for configuring the custom domains but also inject parameters for each project its applied to.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No (shipping as experimental)
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1864
  - [] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6275)


src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+using Azure.Provisioning.Expressions;
+using Azure.Provisioning;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Provides extension methods for customizing Azure Container App resource.
+/// </summary>
+public static class ContainerAppExtensions
```
davidfowl, line 15:
```suggestion
public static class AzureContainerAppExtensions
```

src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            new InterpolatedString(
+                "{0}/managedCertificates/{1}",
+                [
+                 new IdentifierExpression(containerAppManagedEnvironmentIdParameter.IdentifierName),
+                    new IdentifierExpression(certificatNameParameter.IdentifierName)
+                 ]),
+            new NullLiteral()
+            );
+
+        app.Configuration.Value!.Ingress!.Value!.CustomDomains = new BicepList<ContainerAppCustomDomain>()
```
davidfowl, line 92: This will erase anything done before. Is that the intent?

mitchdenny: If we didn't provide anything anything that was added manually would be removed. By doing this we allow folks to specify it as part of the deployment.

src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
mitchdenny: This was intentional, since the type I am extending is `ContainerApp`. Thoughts?

src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
mitchdenny: ... and if someone is customizing those properties they probably aren't using this method anyway they are doing something custom.

src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
davidfowl: That's fine then.

davidfowl (approved):

src/Aspire.Hosting.Azure.AppContainers/ContainerAppExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </example>
+    [Experimental("ASPIREACADOMAINS001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
+    public static void ConfigureCustomDomain(this ContainerApp app, IResourceBuilder<ParameterResource> customDomain, IResourceBuilder<ParameterResource> certificateName)
+    {
+        if (app.ParentInfrastructure is not AzureResourceInfrastructure module)
+        {
+            throw new ArgumentException("Cannot configure custom domain when resource is not parented by ResourceModuleConstruct.", nameof(app));
+        }
+
+        var containerAppManagedEnvironmentIdParameter = module.GetResources().OfType<ProvisioningParameter>().Single(
```
davidfowl, line 65: I don't love this part.

PR #6276, opened 2024-10-13
Add missing port configuration to AddAzurePostgresFlexibleServer RunAsContainer

gabynevada:
## Description

The AddAzurePostgresFlexibleServer RunAsContainer does not have an exposed property to specify the running port like the AddPostgres method allows. This change adds the optional parameter to RunAsContainer and forwards it to the AddPostgres method used internally.

- Added the same port parameter used on the AddPostgres method
- Documented new parameter on RunAsContainer method
- Updated public api reference

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [X] Yes
      - [ ] Yes
      - [X] No
      - [ ] Yes
      - [X] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6276)


gabynevada:
@dotnet-policy-service agree

davidfowl:
@eerhardt Can you review? I think we want this for azure pg,sql and redis.

cc @mitchdenny as well.

mitchdenny:
To be consistent we'd probably have something like WithHostPort

davidfowl:
> To be consistent we'd probably have something like WithHostPort

Are we going to do that for all containers? Just these 3? I Think we expose the actual container resource so that means WithHostPort will show up on all of these resources.

eerhardt:
> > To be consistent we'd probably have something like WithHostPort
> 
> Are we going to do that for all containers? Just these 3? I Think we expose the actual container resource so that means WithHostPort will show up on all of these resources.

I don't think we want it this way. We don't have this `port` parameter for CosmosDB or Azure Storage emulators. But instead you are expected to write:

```C#
builder.AddAzureCosmosDB("cosmos")
    .RunAsEmulator(container =>
    {
        container.WithGatewayPort(9000);
    });

builder.AddAzureStorage("storage")
    .RunAsEmulator(configureContainer: builder =>
    {
        builder.WithBlobPort(9001);
        builder.WithQueuePort(9002);
        builder.WithTablePort(9003);
    });
```

So instead, for PostgreSQL we should have:

```C#
builder.AddAzurePostgresFlexibleServer("pgsql")
    .RunAsContainer(configureContainer: builder =>
    {
        builder.WithHostPort(9004);
    });
```

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -205,6 +205,22 @@ public static IResourceBuilder<T> WithPgAdmin<T>(this IResourceBuilder<T> builde
...
+    /// Configures the host port that the PostgresServer resource is exposed on instead of using randomly assigned port.
+    /// </summary>
+    /// <param name="builder">The resource builder for PostgresServer.</param>
+    /// <param name="port">The port to bind on the host. If <see langword="null"/> is used random port will be assigned.</param>
+    /// <returns>The resource builder for PostgresServer.</returns>
+    public static IResourceBuilder<PostgresServerResource> WithHostPort(this IResourceBuilder<PostgresServerResource> builder, int? port)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        return builder.WithEndpoint("http", endpoint =>
```
mitchdenny: I don't think `http` is the right endpoint name in this case. I think you want 'tcp'.

gabynevada: Yes sorry, noticed right after and pushed the changes.

gabynevada:
@eerhardt @mitchdenny  Added requested changes for the PostgresServerResource.

Some follow up questions:
- Is this the expected design?
- Are other resource changes needed in this PR?

mitchdenny:
@gabynevada apologies for the slow response here. I think we'll look to take this in 9.1.

eerhardt:
Thank you for your contribution and suggestion @gabynevada!

I'm going to close this PR in favor of [Add ability to configure Port and Password for Redis and Postgres and SqlServer when they configured to run by `RunAsContainer` (dotnet/aspire#8439)](https://github.com/dotnet/aspire/pull/8439), since that PR contains the full fix for other resources as well.

PR #6278, opened 2024-10-14
Add argument null validation to Aspire.Hosting

### LINKED ISSUE #2649: Review use of ArgumentException/ArgumentNullException in all public APIs

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/2649

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6278)


JamesNK:
There is a test failure from https://github.com/dotnet/aspire/blob/4e7e0de000bb83f824cca2773892a22b5da89245/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L317

Which is wrong: `WithImageRegistry` not allowing null or `WithDockerfile` calling `WithImageRegistry` with a null value?

tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
```diff
@@ -185,7 +185,7 @@ public async Task EnsureContainerWithVolumesEmitsVolumes()
         builder.AddContainer("containerwithvolumes", "image/name")
             .WithVolume("myvolume", "/mount/here")
             .WithVolume("myreadonlyvolume", "/mount/there", isReadOnly: true)
-            .WithVolume(null! /* anonymous volume */, "/mount/everywhere");
+            .WithVolume("/mount/everywhere");
```
eerhardt, line 188: Should we update the API so allow for `name` to be null? If you follow the call path, it gets passed into `source` below:

```
    /// <param name="source">The source path if a bind mount or name if a volume. Can be <c>null</c> if the mount is an anonymous volume.</param>
    /// <param name="target">The target path of the mount.</param>
    /// <param name="type">The type of the mount.</param>
    /// <param name="isReadOnly">A value indicating whether the mount is read-only.</param>
    public ContainerMountAnnotation(string? source, string target, ContainerMountType type, bool isReadOnly)
```

Also, these `WithVolume` APIs don't meet .NET Design Guidelines:

https://github.com/dotnet/aspire/blob/4e7e0de000bb83f824cca2773892a22b5da89245/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L50

https://github.com/dotnet/aspire/blob/4e7e0de000bb83f824cca2773892a22b5da89245/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L63

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading

❌ AVOID being inconsistent in the ordering of parameters in overloaded members. Parameters with the same name should appear in the same position in all overloads.

eerhardt: Should we update the API so allow for name to be null?

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -168,6 +178,10 @@ public static IResourceBuilder<ProjectResource> AddProject(this IDistributedAppl
     /// </example>
     public static IResourceBuilder<ProjectResource> AddProject(this IDistributedApplicationBuilder builder, [ResourceName] string name, string projectPath, string? launchProfileName)
     {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(name);
```
drewnoakes, line 182: Several of these should probably throw if empty, as in:

```suggestion
        ArgumentException.ThrowIfNullOrEmpty(name);
```

JamesNK: They could. However, there is more validations of names that ensures they're not empty, Technically this null check isn't nesseccary, but I like the consistency of checking all references for null in public API calls.

tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
eerhardt: Because it makes it easier to write code like this:

```csharp
string? volumeName = GetVolumeName(); // may return null

builder.AddContainer("containerwithvolumes", "image/name")
    .WithVolume(volumeName, "/mount/everywhere");
```

eerhardt: See also: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading

✔️ DO allow null to be passed for optional arguments.

JamesNK: Ok

JamesNK:
> Ok. So, that means `WithImageRegistry` should allow a null argument?
> 
> On the surface that seems strange to me. Why would someone call it with null rather than just not call it? If it is changed to accept null, what should the doc comment say if you pass null to it?
> 
> I'm not fimilar with this API. I'd like some actionable feedback.

@mitchdenny What is the answer here?

mitchdenny:
Given the API has shipped allow null to be passed to WithImageRegistry.

JamesNK:
I don't think it being shipped should stop us. It shipped without a nullable argument.

What is the best behavior?

mitchdenny (approved):
LGTM. I think its good to tighten this up.

mitchdenny:
Kicking new build to see what the test failures were.

mitchdenny:
Flagging this as a breaking change as we did technically change the signature, although most folks won't notice if they just recompile.

PR #6300, opened 2024-10-14
Add copy/download console logs button

### LINKED ISSUE #1429: Add a copy console logs button
Make it easy to copy logs to the clipboard with a copy output button

adamint:
## Description

This wasn't very straightforward because we buffer the visible console log lines, so we cannot just take the visible html. The strategy I took was to avoid messing with the `AnsiParser` and instead to strip out the added tags in `ConsoleLogs`.

Fixes #1429

Animation

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No


JamesNK (changes requested):
Unfortunatly there is a fundamental problem with this approach and that's virtualization. A client-side approach can only copy the logs that are on the client, and because of virtualization that will be those that are visible + about 5-10 on either side.

That behavior won't be what people expect. They would expect a top-level button to copy the entire log output.

I see two alternatives:
* Have a copy line button, i.e. you hover over a line and a copy icon appears on the right, similar to how we have the `...` button appear when hovering over a grid cell. Is there value in opening a console log line in the visualizer as well? I don't know. Maybe for consistency have the menu with the standard copy and visualizer options in console logs.
* Change the copy button you have to a download button. The download button could call off to the server and return the entire result as a file. This shouldn't be difficult as the dashboard already has all the console log data in memory. There are docs for how to do this [here](https://learn.microsoft.com/en-us/aspnet/core/blazor/file-downloads?view=aspnetcore-8.0#download-from-a-stream). Logs should always be under 250 MB so the stream approach should be fine.

adamint:
> They would expect a top-level button to copy the entire log output.

it does. It is not a client-side approach @JamesNK

adamint:
> > They would expect a top-level button to copy the entire log output.
> 
> it does. It is not a client-side approach @JamesNK

However, I do like this download idea more

JamesNK:
Ok, I looked again at this PR and I misunderstood where it was getting the data from. I thought it was coming from the text area. Instead it comes from an attribute with all the data, unvirtualized.

The behavior looks like it is good. But I have some technical concerns:

* Is there a limit on how much can be in the clipboard? I couldn't find anyone discuss a limit.
* Every time the console UI updates there is **a lot** of work going on to update the attribute. There could be 10,000 logs. That means the UI is potentially running a regex 10,000 times, creating 10,000 new strings, then concating them into a new giant string, then putting the result in an attribute which is sent to the client over the network. That is a lot of CPU usage, memory usage and network usage.

I haven't attached a profiler to the UI, but when I look at the browser dev tools for test shop and I see a lot of network usage:

image

That's over 1 meg a second.

JamesNK:
@adamint Have you see this: https://wolfgangrittner.dev/how-to-use-clipboard-api-in-firefox/

Perhaps we don't need to have all the content in an attribute.

JamesNK:
> > > They would expect a top-level button to copy the entire log output.
> > 
> > 
> > it does. It is not a client-side approach @JamesNK
> 
> However, I do like this download idea more

I think there is a place for both. Copying to clipboard (if possible) is useful. Also having a download button would be useful.

adamint:
image
image

@JamesNK it now makes a request to a new api `/api/logs/{resourceName}(/download)?` depending on whether you are copying or downloading

JamesNK:
> image

The row download/copy button is quite buggy. On my machine it is only visible if the mouse is just at the end of content, and it pushes the line below down. It's a cute feature but not as useful. I suggest you leave it out of this PR and focus on getting it working well for the entire console output. We can revisit it later.

adamint:
> I suggest you leave it out of this PR and focus on getting it working well for the entire console output. We can revisit it later.

all good with me 👍

src/Aspire.Dashboard/wwwroot/js/app.js
```diff
@@ -114,8 +128,45 @@ function isScrolledToBottom(container) {
...
+    if (typeof ClipboardItem && navigator.clipboard.write) {
+        // NOTE: Safari locks down the clipboard API to only work when triggered
+        //   by a direct user interaction. You can't use it async in a promise.
+        //   But! You can wrap the promise in a ClipboardItem, and give that to
+        //   the clipboard API.
+        //   Found this on https://developer.apple.com/forums/thread/691873
+        const text = new ClipboardItem({
+            "text/plain": fetch(url)
+                .then((response) => response.text())
+                .then((text) => new Blob([text], { type: "text/plain" }))
```
JamesNK, line 142: Does this work on Safari? It would make using the clipboard easier in a lot of situations.

JamesNK:
UI suggestions:

I don't like the action button. How about a gear icon instead? And right align it in the toolbar.

For example:
image

From https://github.com/grpc/grpc-dotnet/actions/runs/11283188129/job/31382129776

Also some icons on the options might look good. Standard copy button and `ArrowDownload`

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
```diff
@@ -25,6 +25,24 @@
                 // This takes up too much horizontal space on mobile, so show on a new line on mobile
                 <FluentLabel Typo="Typography.Body" aria-live="polite" aria-label="@Loc[nameof(Dashboard.Resources.ConsoleLogs.LogStatusLabel)]">@PageViewModel.Status</FluentLabel>
             }
+
+            @if (PageViewModel.SelectedResource is not null)
+            {
+                <FluentMenuButton ButtonAppearance="@Appearance.Neutral"
```
JamesNK, line 31: The menu button should always be visible. If there is no selected resource then disable the menu items.

src/Aspire.Dashboard/ResourceService/DashboardClient.cs
```diff
@@ -517,9 +537,29 @@ async IAsyncEnumerable<IReadOnlyList<ResourceLogLine>> IDashboardClient.Subscrib
             {
                 yield return batch.SelectMany(batch => batch).ToList();
             }
+
+            if (cancelAfterInitialData)
+            {
+                await initialDataCts.CancelAsync().ConfigureAwait(false);
+                break;
+            }
```
JamesNK, line 545: It's possible that big logs will be returned in multiple batches.

src/Aspire.Dashboard/wwwroot/js/app.js
JamesNK: Hmm, I was hoping we could combine this with the technique shown here: https://learn.microsoft.com/en-us/aspnet/core/blazor/file-downloads?view=aspnetcore-8.0#download-from-a-stream

But it requires a roundtrip to the server to get the stream reference.

JamesNK:
@davidfowl How important is copying logs vs downloading logs? Copying logs is difficult because Safari prevents text being added to the clipboard outside of a user event. In other words, we can't have a button make a call to the server, and the server return data, then add it to the clipboard.

To make it work the data must either:
* Already be on the page. Can't do that here because logs could be large and updating.
* Downloaded via a URL. This is technically possible, but it's a fair bit of work. For example, the method for getting logs today subscribes to updates which means we don't know when it has finished returning current logs. We'd need to add a new API to the resource service for getting historical logs (or alternatively, add a flag to subscribe logs to say stop returning new logs, which would be weird on an API called subscribelogs)

On the other hand, downloading logs is quite simple. We can have a button call to the console logs page, get logs as a stream from dashboard state, and trigger a download.

JamesNK:
> > I suggest you leave it out of this PR and focus on getting it working well for the entire console output. We can revisit it later.
> 
> all good with me 👍

Thanks. Can you update the PR to remove code related to line copying/viewing.

There are problems I see around getting logs in a separate API call. I'd like to experiment with the best way to do that.

src/Aspire.Dashboard/ResourceService/DashboardClient.cs
adamint: All batches are returned on line #538 - that hasn't changed

adamint:
> Downloaded via a URL. Complicated. The URL can't access Blazor state so it would need to call an API that downloads logs. However, the method for getting logs today subscribes to updates which means we don't know when it has finished returning current logs. **We'd need to add a new API to the resource service** for getting historical logs then finishing (or alternatively, add a flag to subscribe logs to say stop returning new logs, which would be weird on an API called subscribelogs). This is technically possible, but a fair bit of work.

Does watching and only returning the initial batches not work for this scenario?

JamesNK:
It doesn't work because there is a max message size of 4MB. If logs are bigger than 4MB then they are forcefully broken up into multiple messages.

JamesNK:
@adamint Talked with @davidfowl offline and he prefers log download. That means we can greatly simplify this PR.

Use the technique here: https://learn.microsoft.com/en-us/aspnet/core/blazor/file-downloads?view=aspnetcore-8.0#download-from-a-stream to get data directly from the dashboard. The logs are already in-memory so it should be simple to build them up in a stream, pass back to JS, and have JS trigger the download.

Some things to keep in mind:
* Logs can be formatted with HTML. Need to use raw value, which could mean storing it alongside formatted value.
* The file name should be the resource name + the current time down to the minute. e.g. `myfrontend-abcsdfs-yyyyMMddhhmmss.txt`
* Even thought there is one option, keep the menu. We'll add more options to it in the future (e.g. toggle timestamps on/off, clear logs, etc)

adamint:
@JamesNK simplified, including example gif.
Animation

JamesNK:
UI feedback:
* Right align the menu. Should just be a matter of setting `slot=end`. See other toolbars for examples.
* Change the menu button to a `...` icon
* Add `ArrowDownload` to `Download logs` menu item.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -401,6 +407,19 @@ private async Task StopAndClearConsoleLogsSubscriptionAsync()
...
 
+    private async Task DownloadLogsAsync()
+    {
+        // Strip control sequences from log lines before combining them into a single string.
+        var formattedLines = _logEntries.GetEntries().Select(entry => entry.RawContent is null ? null : AnsiParser.StripControlSequences(entry.RawContent));
+        var logContent = string.Join(Environment.NewLine, formattedLines);
+
+        var stream = new MemoryStream(Encoding.UTF8.GetBytes(logContent));
+        using var streamReference = new DotNetStreamReference(stream);
+        var fileName = $"{PageViewModel.SelectedResource!.DisplayName}-{DateTime.Now.ToString("yyyyMMddhhmmss", CultureInfo.CurrentCulture)}.txt";
```
JamesNK, line 418: Only certain characters are allowed in a file name. `DisplayName` characters could contain something that's invalid.

Quick way to replace invalid chars with `_`: https://github.com/dotnet/maui/blob/33240ea8bccbd398ffb34a4f303dd7301e5f3eaa/src/Essentials/src/FileSystem/FileSystem.uwp.cs#L13

JamesNK, line 418: I don't think there is any reason to use the current culture for this:
```suggestion
        var fileName = $"{PageViewModel.SelectedResource!.DisplayName}-{DateTime.Now.ToString("yyyyMMddhhmmss", CultureInfo.InvariantCulture)}.txt";
```

src/Aspire.Dashboard/ConsoleLogs/AnsiParser.cs
```diff
@@ -21,6 +21,24 @@ public class AnsiParser
     private const int XtermForegroundSequenceCode = 38;
     private const int XtermBackgroundSequenceCode = 48;
 
+    public static string StripControlSequences(string text)
```
JamesNK, line 24: This is very inefficent. Compare it with `ConvertToHtml` which:

* Jumps directly to the next escape char using `IndexOf`. `IndexOf` is very fast and it avoids inspecting every character multiple times
* Adds ranges of text. Avoids appending one char at a time to the StringBuilder

Additionally, if there are no control sequences (will be common) then there is no need to create a StringBuilder. The original string can be returned.

JamesNK (changes requested):

src/Aspire.Dashboard/ConsoleLogs/AnsiParser.cs
adamint: Gotcha, made this more efficient now.

JamesNK: Much better. I made a couple of small changes so that the StringBuilder can be lazy, and the builder is initialized at the existing strings length.

JamesNK:
Looks good. I made some minor changes:

- Lazy StringBuilder when stripping control sequences
- Concat all values using a StreamWriter that writes UTF8 directly to the MemoryStream. Avoid creating an array with all strings, and a joined string.
- Reduce concept of content vs raw content to where it is needed

JamesNK (approved):

PR #6301, opened 2024-10-14
Shorten Functions storage prefix and use AppHost name SHA

### LINKED ISSUE #6257: Azure Functions Default Azure Storage name is too long
When deploying an Azure Functions app using the default/implicit Azure Storage account, the name used is too long and isn't getting a unique string in Azure. The storage account name needs to be unique across all of Azure.

https://learn.microsoft.com/en-us/azure/templates/microsoft.storage/storageaccounts?pivots=deployment-language-bicep#storageaccounts

```C#
var funcApp = builder.AddAzureFunctionsProject<Projects.AzureFunctionsEndToEnd_Functions>("funcapp")
    .WithExternalHttpEndpoints()
    .WithReference(eventHubs)
    .WithReference(blob)
    .WithReference(queue);
```

Produces the following bicep:

```bicep
resource defaultazfuncstorage2380b 'Microsoft.Storage/storageAccounts@2024-01-01' = {
  name: take('defaultazfuncstorage2380b${uniqueString(resourceGroup().id)}', 24)
  kind: 'StorageV2'
  location: location
  sku: {
    name: 'Standard_GRS'
  }
  properties: {
    accessTier: 'Hot'
    allowSharedKeyAccess: false
    minimumTlsVersion: 'TLS1_2'
    networkAcls: {
      defaultAction: 'Allow'
    }
  }
  tags: {
    'aspire-resource-name': 'defaultazfuncstorage2380b'
  }
}
```

Note this line: `name: take('defaultazfuncstorage2380b${uniqueString(resourceGroup().id)}', 24)`

We are already over the 24 char limit with our default name + 5 SHA chars from the AppHost path.

If 2 developers are working on the same app in the same file location (ex. `C:\git`), when they deploy to separate resource groups, the Storage account is going to conflict because they both will try to use `defaultazfuncstorage2380` as the name. This is the same if you are deploying the same app twice from a GitHub Action or Azure Devops pipeline.

We should make the prefix shorter to allow for more unique characters to come from the `uniqueString` portion.

cc @captainsafia

### LINKED ISSUE #6270: DataVolume names should not be derived based on AppHost:Sha256.
@davidfowl I just noticed this whilst doing some experimentation with container app customization. In the playground for containerapps the redis container uses a data volume. I noticed that the data volume name changes based on the path you are publishing from. This probably isn't desirable at deployment time because if you change your source layout on the build agent (for example) then you'll end up with a different path.

For deployment scenarios instead of using the AppHost:Sha256 value we probably need to use something based on the resource group name (like we do for resource names).

_Originally posted by @mitchdenny in https://github.com/dotnet/aspire/issues/5779#issuecomment-2408820153_

captainsafia:
Fixes https://github.com/dotnet/aspire/issues/6270

Fixes https://github.com/dotnet/aspire/issues/6257

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6301)

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -14,7 +14,16 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class AzureFunctionsProjectResourceExtensions
 {
-    internal const string DefaultAzureFunctionsHostStorageName = "defaultazfuncstorage";
+    /// <remarks>
+    /// The prefix used for configuring the name default Azure Storage account that is used
+    /// for Azure Functions bookkeeping. Locally, the name is generated using a combination of this
+    /// prefix, a hash of the AppHost project name. During publish mode, the name generated
+    /// is a combination of this prefix, a hash of the AppHost project name, and  the name of
+    /// the resource group associated with the deployment. We want to keep the total number
+    /// of characters in the name under 24 characters to avoid truncation by Azure and allow
+    /// for unique enough identifiers.
+    /// </remarks>
+    internal const string DefaultAzureFunctionsHostStorageName = "funcstorage";
 
     /// <summary>
     /// Adds an Azure Functions project to the distributed application.
@@ -29,7 +38,7 @@ public static class AzureFunctionsProjectResourceExtensions
 
         // Add the default storage resource if it doesn't already exist.
         var storageResourceName = builder.CreateDefaultStorageName();
-        AzureStorageResource? storage = builder.Resources
+        var storage = builder.Resources
             .OfType<AzureStorageResource>()
             .FirstOrDefault(r => r.Name == storageResourceName);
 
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -14,7 +14,16 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class AzureFunctionsProjectResourceExtensions
 {
-    internal const string DefaultAzureFunctionsHostStorageName = "defaultazfuncstorage";
+    /// <remarks>
+    /// The prefix used for configuring the name default Azure Storage account that is used
+    /// for Azure Functions bookkeeping. Locally, the name is generated using a combination of this
+    /// prefix, a hash of the AppHost project name. During publish mode, the name generated
+    /// is a combination of this prefix, a hash of the AppHost project name, and  the name of
+    /// the resource group associated with the deployment. We want to keep the total number
```
eerhardt, line 22: I think this comment is out of date / inaccurate.

captainsafia: Yes, I was on the fence about whether shortening the prefix was sufficient and we could get away with still including the SHA. I ended up deciding not to include the SHA so that's how we got here.

davidfowl (approved):

davidfowl:
I am a little concerned about using the same field and switching it based on mode but this direction seems right.

Issue #6226, opened 2024-10-10
[WebToolsE2E][Aspire] After adding Aspire.Microsoft.EntityFrameworkCore.SqlServer, a warning appears: Package 'System.Runtime.Caching' 8.0.0 has a known high severity vulnerability


v-reinawang:
**REGRESSION INFO**: Repro on Aspire 8.2.0 and Aspire 8.2.1

**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2 ENU
2. Install Dev17.12 Preview 3 [35409.206.d17.12][Include SDK 9.0.100-rc.2.24474.11]
3. Install Aspire 9.0.0-rc.1.24509.13

**REPRO STEPS**

1. Create a new project > .NET Aspire 9 Starter App > named it 'AspireApp' > .Net 9.0 (Preview) > Create
2. Right click AspireApp.Web > Add > .NET Aspire package > Install Aspire.Microsoft.EntityFrameworkCore.SqlServer (Version 9.0.0-rc.1.24509.13)

**NOTE**: 
1. If TargetFramework selects net8.0, this issue will also be repro.
2. Create Razor Pages ~ individual project. This issue is not repro because the Microsoft.EntityFrameworkCore.SqlServer package references System.Runtime.Caching (6.0.0)
3. Package 'System.Runtime.Caching' Release Date:
![Image](https://github.com/user-attachments/assets/43838e4b-b203-4886-833e-f7f46f7cf73b)

**ACTUAL**
A warning appears: Package 'System.Runtime.Caching' 8.0.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-qj66-m88j-hmgj
![Image](https://github.com/user-attachments/assets/29bcd492-02b3-45fd-a134-62d7f199eab7)

![Image](https://github.com/user-attachments/assets/f075f80e-4e8c-46bb-90ae-45e8e0c09a5a)

**EXPECTED**
No warning

lommez:
I'm having the same problem.
Any solution for this?

eerhardt:
> I'm having the same problem. Any solution for this?

For now, you can add an explicit `<PackageReference Include="System.Runtime.Caching" Version="8.0.1" />` which will remove the warning.

eerhardt:
This package being marked as vulnerable was a mistake. See https://github.com/github/advisory-database/pull/4904/ which is correcting the advisory with the correct package name.

We just need to wait for nuget.org to be updated and this warning will go away.

[...]

PR #6303, opened 2024-10-14
Enable NuGet Audit

joperezr:
## Description

Enabling NuGet Audit so that our build can catch issues when one of our dependencies is marked as vulnerable, so we are able to pin to a higher version  in order to lift the dependency out of the vulnerability.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6303)


sebastienros (approved):

ViktorHofer:
Contributes to https://github.com/dotnet/arcade/issues/15019

@joperezr we recommend also setting the following WarningsNotAsErrors property so that local dev builds / PRs don't stop working on a Patch Tuesday: https://github.com/dotnet/arcade/blob/964e434191bd3ca5675743d08ff742ec0f1e79a9/Directory.Build.props#L17-L18

It's important to keep this enabled for official builds though.

ViktorHofer (approved):

Directory.Packages.props
```diff
@@ -197,6 +197,7 @@
     <!-- Pinned versions for Component Governance - Remove when root dependencies are updated -->
     <PackageVersion Include="Azure.Core" Version="1.44.0" />
     <PackageVersion Include="Azure.Identity" Version="1.12.1" />
+    <PackageVersion Include="System.Runtime.Caching" Version="8.0.1" />
```
eerhardt, line 200:
```suggestion
```

This is not necessary. See https://github.com/dotnet/aspire/issues/6226#issuecomment-2412218950

Directory.Build.props
```diff
@@ -34,6 +34,8 @@
 
     <DashboardPublishedArtifactsOutputDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'DashboardArtifacts', '$(Configuration)'))</DashboardPublishedArtifactsOutputDir>
     <WorkloadsPackageSource>$(ArtifactsShippingPackagesDir)</WorkloadsPackageSource>
+    <!-- Only upgrade NuGetAudit warnings to errors for official builds. -->
+    <WarningsNotAsErrors Condition="'$(OfficialBuild)' != 'true'">$(WarningsNotAsErrors);NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
```
eerhardt, line 38: Do we have a way to alert the team, or a subset of the team, when an OfficialBuild fails?

Maybe this shouldn't be based on OfficialBuild, but instead in our rolling CI build? For sure I don't want PRs to start failing on the day a security release comes out.

eerhardt: Along those lines.... this PR is failing due to NU1903

ViktorHofer: We just found out that there's an underlying msbuild bug which ignores WarningsNotAsErrors when operating on solution files. See https://teams.microsoft.com/l/message/19:3212bf033f4c4b5198643a04fa1048fa@thread.skype/1728680077070?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1728680077070&teamName=.NET%20Developer%20Experience&channelName=MSBuild&createdTime=1728680077070 for more details

Here's the workaround that you would need to add meanwhile: https://github.com/dotnet/templating/pull/8451/files#diff-f428536807848d69e9884e86523c06dc1b05d1b281dadd8dd4c3a6d413171519

ViktorHofer: The workaround doesn't work. I would hold this PR off for a few days until we reached consensus on how to react to the msbuild bug.

eerhardt:
FYI - https://github.com/dotnet/aspire/pull/6319 should fix the current failure in this PR - transitively referencing Microsoft.Extensions.Caching.Memory 8.0.0.

Directory.Build.props
ViktorHofer: Looks like understanding why WarningsNotAsErrors doesn't correctly work with solutions will take longer. If things work for you I would not block this PR on it. You always have the option to turn NuGet Audit on/off if something doesn't work. You can NoWarn individual nuget warnings, suppress specific vulerability reports and/or disable NuGet Audit entirely.

As this feature is enabled by default for our customers starting with .NET 8.0.2xx and .NET 9, please give this a try and give feedback back to the NuGet team and me/us in the Arcade issue.

eerhardt: I don't think we should merge this change until it doesn't fail PRs. Getting all our PRs (and local builds) broken when someone marks a NuGet package as vulnerable isn't a good situation to be in.

danmoseley:
@joperezr is this still a project or should we close

joperezr:
We should at some point enable this, but  it is not clear when, so I'll go ahead and close for now and we can reopen when ready.

PR #6068, opened 2024-10-02
Start using TimeProvider in ResourceHealthCheckService.

mitchdenny:
## Description

Currently `ResourceHealthCheckService` uses the DateTime to control loop intervals. This PR switches over to using `TimeProvider` for improved testability.

Fixes #6014 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [ ] Yes (coming soon to a PR near you ... this one!)
  - [ ] No
- Did you add public API?
  - [x] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [x] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [x] No
  - [ ] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6068)

PR #6309, opened 2024-10-15
Move SlowDownMonitoringAsync.

### LINKED ISSUE #6134: SlowDownMonitoringAsync not called when resource reaches steady state.

### Describe the bug

If the health status reaches a steady state the slow down won't occur because we drop out of the loop before it is called. Found it whilst looking into options for speeding up the health checks here:

https://github.com/dotnet/aspire/issues/6112

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

mitchdenny:
## Description

Previously we were calling this method late in the monitoring loop however after some earlier refactoring we made it less effective. By moving it up to where we are in a steady state it has the chance to have an effect (it internally checks for healthy state and falls through quickly if it isn't in one.

Fixes #6134 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6309)

### DIFF
--- a/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
+++ b/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
@@ -83,10 +83,9 @@ internal class ResourceHealthCheckService(ILogger<ResourceHealthCheckService> lo
                         cancellationToken).ConfigureAwait(false);
                 }
 
-                if (_latestEvents[resource.Name].Snapshot.HealthStatus == report.Status)
+                if (_latestEvents[resource.Name] is { } latestEvent && latestEvent.Snapshot.HealthStatus == report.Status)
                 {
-                    // If the last health status is the same as this health status then we don't need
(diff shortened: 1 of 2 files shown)


davidfowl:
These tests would be MUCH better if we did https://github.com/dotnet/aspire/pull/6068

davidfowl (approved):

PR #6312, opened 2024-10-15
Fix broken tests when running Config Schema Generator on .NET 9

bart-vmware:
## Description

Adapt the Config Schema Generator logic to deal with the breaking change in `System.Text.Json` v9, reported at https://github.com/dotnet/runtime/issues/108790.

More specifically, this PR fixes the `JsonNode` handling inside Config Schema Generator to work with STJ v9. The generator project still targets .NET 8, but now references STJ v9 explicitly. Its test project now multi-targets against .NET 8 and 9. 

Before this PR, the generator and its tests only ran on .NET 8 (referencing STJ v8), which is why there were no test failures.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

/cc @eerhardt 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6312)


joperezr:
Hello @bart-vmware, thanks for the contribution. This PR is currently set to target release/9.0-rc.1 which is no longer an active branch. Assuming you are okay with me re-targeting it against `main` ?

joperezr:
We are pretty locked down at this point for release/9.0 branch, so unfortunately wouldn't be able to take this into there. `main` is already a net9 TFM too, in fact `main` is our 9.1 branch now, so that would be the best place to target for your PR.

joperezr:
Given the above, I'm retargeting this to go against main as the rc.1 branch is frozen anyway.

src/Components/Directory.Build.targets
```diff
@@ -78,7 +78,7 @@
   <Target Name="CalculateConfigurationSchemaGeneratorPath">
     <MSBuild Projects="$(ConfigurationSchemaGeneratorProjectPath)"
              Targets="GetTargetPath"
-             RemoveProperties="TargetFramework">
+             Properties="$(Properties);TargetFramework=$(DefaultTargetFramework)">
```
eerhardt, line 81: I'm not sure this is right. (actually I'm pretty sure it is wrong). Can you explain why you are proposing this change? What doesn't work about the current code?

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+	  <TargetFrameworks>$(AllTargetFrameworks)</TargetFrameworks>
```
eerhardt, line 4: We shouldn't need to multi-target the generator. It should only need to target `net8.0`.

tests/ConfigurationSchemaGenerator.Tests/ConfigurationSchemaGenerator.Tests.csproj
```diff
@@ -18,10 +18,19 @@
...
 
+  <ItemGroup Condition="'$(TargetFramework)' != 'net9.0'">
+    <PackageReference Include="Microsoft.Extensions.Hosting" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+    <PackageReference Include="Microsoft.Extensions.Hosting" VersionOverride="9.0.0-rc.2.24473.5" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" VersionOverride="9.0.0-rc.2.24473.5" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" VersionOverride="9.0.0-rc.2.24473.5" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
eerhardt: This should be fixed now. You should be able to revert these changes because of the following:

https://github.com/dotnet/aspire/blob/2e60ee60398402a51f15e17d1115587ed523510e/eng/Versions.props#L76-L79

This will automatically reference the 9.0 hosting packages when targeting `net9.0`.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -719,6 +715,32 @@ private static string[] CreateExclusionPaths(List<string>? exclusionPaths)
         return result;
     }
 
+    private static void ReplaceNodeWithKeyCasingChange(JsonObject jsonObject, string key, JsonNode value)
+    {
+#if NET9_0_OR_GREATER
```
eerhardt, line 720: @eiriktsarpalis - is this the recommended way to respond to the breaking change? Do we have documentation guidance on the break?

I almost wonder if we should be checkign the System.Text.Json major version number instead. Or doing some sort of "runtime capability test" - for example, execute a method that checks for the new behavior and set a `bool` flag.

The situation that can break this is if you are targeting `net8.0`, but somehow the `9.0` System.Text.Json nuget package gets pulled into the app (ex. through a transitive reference). In that situation, the `#else` code will be compiled into the app, but the 9.0 STJ behavior will happen.

eerhardt:
(Sorry for the late response. We were busy locking down for .NET Aspire 9.0, which will be released next week.)

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -61,7 +61,7 @@ private void GenerateLogCategories(JsonObject parent)
         {
             var categoryNode = new JsonObject();
             categoryNode["$ref"] = "#/definitions/logLevelThreshold";
-            propertiesNode[categories[i]] = categoryNode;
+            ReplaceNodeWithKeyCasingChange(propertiesNode, categories[i], categoryNode);
```
eiriktsarpalis, line 64: Given that this is code copied from the config binder generator, is it appropriate to be making changes locally?

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
eiriktsarpalis: I'm not sure I fully understand what tests were broken by the particular change in STJ 9, but per the discussion in https://github.com/dotnet/runtime/issues/108790#issuecomment-2407508284 the pre-9 behaviour is undocumented, accidental behaviour that is inconsistent with how all other dictionary types work in the same context. 

We've agreed to add a breaking change note in our documentation for completeness, but beyond that I don't think it's appropriate to introduce mitigations down the stack just so that the pre-9 behavior is preserved. Assuming there are failing tests that check against that behavior, then I would recommend just updating the tests.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
eerhardt: This part of the code isn't copied from the config binder generator. The part that is copied is inside the https://github.com/dotnet/aspire/tree/main/src/Tools/ConfigurationSchemaGenerator/RuntimeSource directory.

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
bart-vmware: I added multi-targeting so we can have tests for both the STJ 8.0 and 9.0 behavior.
Alternatively, we could only target net8.0, but reference STJ 9.0 from the generator project (and remove the TFM-conditional code). Would you prefer that?

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
eerhardt, line 4:
```suggestion
    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
```

tests/ConfigurationSchemaGenerator.Tests/ConfigurationSchemaGenerator.Tests.csproj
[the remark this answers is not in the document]
eerhardt: I've fixed this with the latest changes.

src/Components/Directory.Build.targets
eerhardt: Reverted with the latest changes.

src/Tools/ConfigurationSchemaGenerator/ConfigurationSchemaGenerator.csproj
eerhardt: > Alternatively, we could only target net8.0, but reference STJ 9.0 from the generator project (and remove the TFM-conditional code). Would you prefer that?

Yes, that is a good idea. I've done that in the latest commit.

Targeting a single TFM (`net8.0`) will help us when we move to ship this tool in a reusable nuget package outside of the repo. #3309.

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
eerhardt: We've moved to always use STJ `9.0.0` even on `net8.0`. So we no longer have to switch behavior here.

eerhardt:
Thanks for the PR here @bart-vmware. I've pushed some updates to make this ready for review. I took your suggestion of always using System.Text.Json `9.0.0` in the tool. This allows for much simpler code.

bart-vmware:
@eerhardt Thanks for doing the updates, this is basically what I had in mind, so LGTM from me.

I've renamed `ReplaceNode` to `ReplaceNodeWithKeyCasingChange` for clarity, and fixed the broken test on .NET 9. I don't know _why_ the explicit `VersionOverride` solves the problem, a multi-targeting glitch perhaps? Without it, the dependencies look like this:

image

eiriktsarpalis (approved):
Thanks

src/Tools/ConfigurationSchemaGenerator/ConfigSchemaEmitter.cs
```diff
@@ -719,6 +715,22 @@ private static string[] CreateExclusionPaths(List<string>? exclusionPaths)
         return result;
     }
 
+    private static void ReplaceNodeWithKeyCasingChange(JsonObject jsonObject, string key, JsonNode value)
+    {
+        // In System.Text.Json v9, the casing of the new key is not adapted. See https://github.com/dotnet/runtime/issues/108790.
+        // So instead, remove the existing node and insert a new one with the updated key.
+        var index = jsonObject.IndexOf(key);
```
eiriktsarpalis, line 722: This workaround requires dual lookup if the key is missing, which suggests to me that a `JsonObject.TryAdd`method similar to the one in https://github.com/dotnet/runtime/issues/107947 might be necessary.

bart-vmware:
Is this PR ready to be merged?

eerhardt (approved):
LGTM.

Thanks for the contribution @bart-vmware!

PR #6316, opened 2024-10-15
Fix creating temporary dashboard client on startup

JamesNK:
## Description

While testing I noticed that a dashboard client was created and then disposed in some middleware. This happened because the middleware wants to know if the client is enabled or not, so resolves the service.

However, because the client is a scoped service, a new instance is created and then later disposed. Once the dashboard UI is displayed another client is created.

The fix is to split the IsEnabled flag out into its own interface which is resolved to test if the client is enabled in middleware. It's ok to do because it has a singleton lifetime.

Low priority. Post 9.0.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6316)

### DIFF
--- a/src/Aspire.Dashboard/DashboardWebApplication.cs
+++ b/src/Aspire.Dashboard/DashboardWebApplication.cs
@@ -232,6 +232,7 @@ public sealed class DashboardWebApplication : IAsyncDisposable
 
         // Data from the server.
         builder.Services.TryAddScoped<IDashboardClient, DashboardClient>();
+        builder.Services.TryAddSingleton<IDashboardClientStatus, DashboardClientStatus>();
 
         // OTLP services.
         builder.Services.AddGrpc();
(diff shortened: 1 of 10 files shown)


adamint (approved):

drewnoakes (approved):
Nice find and fix.

PR #6323, opened 2024-10-15
Set docker summaries for all extensions.

sebastienros:
## Description

Saw that it was missing on SqlServer during Damian's demo.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6323)


src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -98,7 +98,8 @@ public static IResourceBuilder<AzureCosmosDBResource> AddAzureCosmosDB(this IDis
 
     /// <summary>
     /// Configures an Azure Cosmos DB resource to be emulated using the Azure Cosmos DB emulator with the NoSQL API. This resource requires an <see cref="AzureCosmosDBResource"/> to be added to the application model.
-    /// For more information on the Azure Cosmos DB emulator, see <a href="https://learn.microsoft.com/azure/cosmos-db/emulator#authentication"></a>
+    /// For more information on the Azure Cosmos DB emulator, see <a href="https://learn.microsoft.com/azure/cosmos-db/emulator#authentication"></a>.
+    /// This version of the package defaults to the <inheritdoc cref="CosmosDBEmulatorContainerImageTags.Tag"/> tag of the <inheritdoc cref="CosmosDBEmulatorContainerImageTags.Registry"/>/<inheritdoc cref="CosmosDBEmulatorContainerImageTags.Image"/> container image.
```
eerhardt, line 102: This seems more like `<remarks>` information, and not necessarily `<summary>` information.

Thoughts @IEvangelist?

src/Aspire.Hosting.Redis/RedisContainerImageTags.cs
```diff
@@ -21,7 +21,7 @@ internal static class RedisContainerImageTags
     public const string RedisCommanderImage = "rediscommander/redis-commander";
 
     /// <summary>latest</summary>
-    public const string RedisCommanderTag = "latest";
+    public const string RedisCommanderTag = "latest"; // There aren't better tag than 'latest' which is 3 years old.
```
eerhardt, line 24:
```suggestion
    public const string RedisCommanderTag = "latest"; // There isn't a better tag than 'latest' which is 3 years old.
```

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
sebastienros: I tried to keep the existing pattern, but I would agree with that change.

eerhardt: Yeah, it looks a little mixed. Some places put it in the remarks today.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -79,9 +82,12 @@ public static IResourceBuilder<MySqlDatabaseResource> AddDatabase(this IResource
...
 
     /// <summary>
-    /// Adds a phpMyAdmin administration and development platform for MySql to the application model. This version of the package defaults to the <inheritdoc cref="MySqlContainerImageTags.PhpMyAdminTag"/> tag of the <inheritdoc cref="MySqlContainerImageTags.PhpMyAdminImage"/> container image.
+    /// Adds a phpMyAdmin administration and development platform for MySql to the application model.
     /// </summary>
-    /// <param name="builder">The MySql server resource builder.</param>
+    /// <remarks>
+    /// This version of the package defaults to the <inheritdoc cref="MySqlContainerImageTags.PhpMyAdminTag"/> tag of the <inheritdoc cref="MySqlContainerImageTags.PhpMyAdminImage"/> container image.
+    /// </remarks>
+    /// /// <param name="builder">The MySql server resource builder.</param>
```
eerhardt, line 90:
```suggestion
    /// <param name="builder">The MySql server resource builder.</param>
```

eerhardt (approved):
LGTM

sebastienros:
@eerhardt As I was writing what validation I had just done to ensure it's working, I realized that moving it to `<remarks>` is actually breaking it. Because it will now "inherit" the remarks tags. So the `/// <summary>latest</summary>` on the tags now need to be in `/// <remarks>` too.

eerhardt:
>  So the `/// <summary>latest</summary>` on the tags now need to be in `/// <remarks>` too.

You mean "instead of" right?

sebastienros:
"too" because I changed summary to remarks, so the Tags also need the change. But yes. 
I applied the change on the backport PR, assuming we can then merge release/9.0 back to main. Not sure we need to update it in main in that case.

davidfowl:
backport?

eerhardt:
> backport?

https://github.com/dotnet/aspire/pull/6433

PR #6340, opened 2024-10-16
Mark Aspire.Hosting.Python as experimental

### LINKED ISSUE #6298: Mark AddPythonApp API as experimental
We have a couple of issues with the existing AddPythonApp API. We want to mark it as experimental because it still needs more work.

sebastienros:
## Description

Fixes #6298

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6340)

### DIFF
--- a/src/Aspire.Hosting.Python/Directory.Build.props
+++ b/src/Aspire.Hosting.Python/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+      <!-- This needs to be set to true before importing parent Directory.Build.props -->
+      <!-- In preview until the public API is validated. -->
+      <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+    </PropertyGroup>
+  
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+  </Project>
--- a/src/Aspire.Hosting.Python/PythonAppResource.cs
+++ b/src/Aspire.Hosting.Python/PythonAppResource.cs
@@ -8,7 +8,7 @@ using Aspire.Hosting.ApplicationModel;
 namespace Aspire.Hosting.Python;
 
 /// <summary>
-/// A resource that represents a python executible or app.
+/// A resource that represents a python executable or app.
 /// </summary>
 /// <param name="name">The name of the resource.</param>
 /// <param name="executablePath">The path to the executable used to run the python app.</param>
--- a/src/Aspire.Hosting.Python/PythonAppResourceBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Python/PythonAppResourceBuilderExtensions.cs
@@ -44,8 +44,8 @@ public static class PythonAppResourceBuilderExtensions
     /// </para>
     /// </remarks>
     /// <example>
-    /// Add a python app or executible to the application model. In this example the python code entry point is located in the <c>PythonProject</c> directory
-    /// if this path is relative then it is assumed to be relative to the AppHost directory, and the virtual enviroment path if relative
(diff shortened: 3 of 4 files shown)


davidfowl:
We want to mark the API as experimental. When you call AddPythonApp it should have the attribute on it. Like this https://github.com/eerhardt/aspire/blob/5603e928797a629d972c0084e2d6414169d1295c/src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIExtensions.cs#L39

davidfowl:
Although, maybe pre-release is enough to signal what we want.

sebastienros:
I could not find this attribute used anywhere though.

davidfowl (approved):

src/Aspire.Hosting.Python/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+      <!-- This needs to be set to true before importing parent Directory.Build.props -->
+      <!-- In preview until the public API is validated. -->
+      <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
```
davidfowl, line 5: Why isn't this in the csproj?

eerhardt:
My thoughts:

1. Making a nuget package not stable (i.e. have a `-preview` suffix in the version) should be enough signal to people that the API can change before it becomes stable. Users don't have to install a "pre release" package and then ALSO suppress a warning about using experimental APIs inside that prerelease package.

2. `Aspire.Hosting.Python` already shipped stable in 8.1 and 8.2. If we don't ship a "stable" 9.0.0 package, people will continue using the 8.2.1 Aspire.Hosting.Python package with Aspire 9.0. This maybe won't cause problems, but it will be confusing.

src/Aspire.Hosting.Python/Directory.Build.props
eerhardt: See conversation here: https://github.com/dotnet/aspire/pull/6195#discussion_r1792442697

cc @joperezr

sebastienros: Based on this PR this is the way to do it now
https://github.com/dotnet/aspire/pull/6195/files

davidfowl:
@DamianEdwards thoughts?

src/Aspire.Hosting.Python/Directory.Build.props
joperezr: That is no longer the case, it was only for RC1 as we had things going RC1 and some other things staying in preview, but now everything should fall back to be preview or stable.

eerhardt:
Another problem is that shipped https://www.nuget.org/packages/Aspire.Hosting.Python/9.0.0-rc.1.24511.1. Which means when we ship `9.0.0` GA, the Python package produced with that build will still have `-preview` and will be LOWER than the rc1 package. So even if people move to the 9.0.* package, they will get the rc1 package, which is old.

Maybe we need to ship this as stable and mark the APIs as `[Experimental]`....

DamianEdwards:
> Maybe we need to ship this as stable and mark the APIs as [Experimental]....

I thought that was the proposal?

eerhardt:
> > Maybe we need to ship this as stable and mark the APIs as [Experimental]....
> 
> I thought that was the proposal?

What ended up getting changed was the package is no longer building as "stable", so when it gets built it will have a `-preview` version.

sebastienros:
I will do the change

PR #6347, opened 2024-10-16
Azure Hosting API Review Round 2

### LINKED ISSUE #6174: Clean up Azure Hosting APIs

eerhardt:
## Description

The final round of API review changes for the Azure Hosting APIs.

Easiest to review each commit separately.

- Rename AzureResourceInfrastructure.Resource to AspireResource
- Remove AzureResourceInfrastructure parameter properties

Fix #6174

The final version of the APIs can be viewed here:

https://spa.apiview.dev/review/2c5cd25605b44759b267536b0fcecc84?activeApiRevisionId=7fad9d11cea64516a972b5323432333b

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [x] Yes
  - [x] No
  - [x] Yes

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6347)

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -54,8 +54,6 @@ public static class AzureFunctionsProjectResourceExtensions
 
                     var principalTypeParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalType, typeof(string));
                     var principalIdParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalId, typeof(string));
-                    infrastructure.Add(principalTypeParameter);
-                    infrastructure.Add(principalIdParameter);
 
                     var storageAccount = infrastructure.GetResources().OfType<StorageAccount>().FirstOrDefault(r => r.IdentifierName == storageResourceName)
                         ?? throw new InvalidOperationException($"Could not find storage account with '{storageResourceName}' name.");
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -51,7 +51,6 @@ public static class AzureFunctionsProjectResourceExtensions
             {
                 var configureInfrastructure = (AzureResourceInfrastructure infrastructure) =>
                 {
-
                     var principalTypeParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalType, typeof(string));
                     var principalIdParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalId, typeof(string));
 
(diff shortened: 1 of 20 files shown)


src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -50,9 +51,15 @@ public static class AzureFunctionsProjectResourceExtensions
             {
                 var configureInfrastructure = (AzureResourceInfrastructure infrastructure) =>
                 {
+
+                    var principalTypeParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalType, typeof(string));
+                    var principalIdParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalId, typeof(string));
+                    infrastructure.Add(principalTypeParameter);
+                    infrastructure.Add(principalIdParameter);
```
eerhardt, line 58: This is unfortunate because it results in duplicate parameters in the bicep. @tg-msft - is that something we want to protect against in Azure.Provisioning? Adding duplicate parameters to an `Infrastructure` object?

Maybe we shouldn't be adding these parameters to the infastructure expicitly, and letting our parameter fix up code take care of it, like we were before.

davidfowl: The fixup works well if there's no special handling (e.g secrets or non string types)

eerhardt: What if we had a `GetOrAddParameter(string name, Type? type = null)` method on `Infrastructure`?

eerhardt: These would all be:

```C#
var principalTypeParameter = infrastructure.GetOrAddParameter(AzureBicepResource.KnownParameters.PrincipalType);
var principalIdParameter = infrastructure.GetOrAddParameter(AzureBicepResource.KnownParameters.PrincipalId);
```

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -50,9 +51,13 @@ public static class AzureFunctionsProjectResourceExtensions
             {
                 var configureInfrastructure = (AzureResourceInfrastructure infrastructure) =>
                 {
+
```
eerhardt, line 54:
```suggestion
```

davidfowl (approved):

mitchdenny (approved):

Issue #5980, opened 2024-09-27
[WebToolsE2E][Aspire] In Aspire 9.0 RC1 dashboard there are some strings that are not localized in non-ENU OS.


vmykagapuz:
INSTALL STEPS
1. Clean machine: Win11 x64 22h2 DEU
2. Install VS 17.12 Preview 3
3. Install Aspire 9.0 RC1

REPRO STEPS
1. Create Aspire Starter App
2. Build solution and run the project

ACTUAL

The following are the strings that are not localized

1. The "Close" in Settings
![image](https://github.com/user-attachments/assets/728da70f-5cd1-4fd8-bfd9-c35208abaf0d)

2. In Resources

- The "Actions" & [View Details/Console logs/Structured logs/Traces/Metrics/Restart] under 3 dots button
![image](https://github.com/user-attachments/assets/8abbb76e-a655-4ad0-b9f0-2970ea2067e8)

3. The "None" option in Console and Metrics
![image](https://github.com/user-attachments/assets/7344367a-512c-46d4-84e7-42c472292b74)

4. In Structured and Traces

- The "All" option in Resource and Level under Structured
![image](https://github.com/user-attachments/assets/64d3bcd4-8fbb-4c62-981c-3628f1b88e4f)
![image](https://github.com/user-attachments/assets/3cdede7c-4622-49a1-b5b2-5e77f17a5665)

- The "Filters" & "No filters"
![image](https://github.com/user-attachments/assets/f016f7e3-f992-4745-90c0-0e5d7e575f96)

- The "Add Filter" under Filters section
![image](https://github.com/user-attachments/assets/2868e0b5-abe1-4a92-9fc0-2c7e88c2155d)

- The strings option under Parameter
![image](https://github.com/user-attachments/assets/301d8faa-ac43-40fd-898d-54163b852d06)

- The "contains" & "not contains" in Condition
![image](https://github.com/user-attachments/assets/661d0800-a572-421e-9d64-ab3783568a2e)

5. The "Restart" under 3 dots button in aspire dashboard after publish
![image](https://github.com/user-attachments/assets/519bc7cb-1e8e-4235-b511-9fb6202564c5)

EXPECTED
The mentioned strings should be localized.

vmykagapuz:
I updated the Actual Result adding more items of untranslated strings found in KOR and CHS OS.

drewnoakes:
@vmykagapuz it's helpful to know if a translation is missing in just one language, or in all languages.

[...]

PR #6348, opened 2024-10-17
Fix incorrect resource lookups

drewnoakes:
## Description

I believe this fixes some places where strings would not be localized in the UI for users of supported `en-US` cultures.

When using `IStringLocalizer`, we must use the name of the resource we want to look up. The easiest way to do this is using `nameof` with the generated property for that resource.

Omitting the namespace causes the resource to be looked up and then used as a key for a lookup, which is not correct. The lookup should happen through `IStringLocalizer`, not the generated resource class.

Follows an observation made in https://github.com/dotnet/aspire/pull/6209#discussion_r1803940775

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6348)

### DIFF
--- a/src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
+++ b/src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
@@ -210,8 +210,8 @@ public partial class ChartContainer : ComponentBase, IAsyncDisposable
                 {
                     var text = v switch
                     {
-                        null => Loc[ControlsStrings.LabelUnset],
-                        { Length: 0 } => Loc[ControlsStrings.LabelEmpty],
+                        null => Loc[nameof(ControlsStrings.LabelUnset)],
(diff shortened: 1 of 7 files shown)


JamesNK (approved):

JamesNK:
Check if any of the fixes are raised in this issue: https://github.com/dotnet/aspire/issues/5980

drewnoakes:
> Check if any of the fixes are raised in this issue: #5980

It might help with some, but not all. Looks like some of those just don't have translations yet.

adamint (approved):

PR #6295, opened 2024-10-14
Add Api diff between 8.2.1 -> 9.0 RC1

joperezr:
This is adding a markdown file that has the API Diff between 8.2.1 and 9.0-rc1 release.

@davidfowl @DamianEdwards @eerhardt @mitchdenny @JamesNK PTAL and make sure everything here is expected. Also, please tag anyone that is relevant for areas as you see appropriate.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6295)

PR #6350, opened 2024-10-17
Rename CommandType to Name on resource command types

JamesNK:
## Description

Feedback from https://github.com/dotnet/aspire/pull/6295#discussion_r1803777740.

Note that this is a source breaking change for people who have implemented the proto when they update the proto the identifer will change name. It's not a wire breaking change because the field id (`1` in this case) is the same.

@joperezr Where should the `commandType` -> `name` breaking change be communicated? The model API is new in 9.0 so I believe can change at this point without breaking change notification.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

### DIFF
--- a/tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
+++ b/tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
@@ -118,7 +118,7 @@ public class DashboardServiceTests
         var resourceData = Assert.Single(update.InitialData.Resources);
         var commandData = Assert.Single(resourceData.Commands);
 
-        Assert.Equal("TestType", commandData.Name);
+        Assert.Equal("TestName", commandData.Name);
         Assert.Equal("Display name!", commandData.DisplayName);
         Assert.Equal("Display description!", commandData.DisplayDescription);
         Assert.Equal(Value.ForList(Value.ForString("One"), Value.ForString("Two")), commandData.Parameter);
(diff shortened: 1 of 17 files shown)


tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
```diff
@@ -79,7 +79,7 @@ public async Task WatchResources_ResourceHasCommands_CommandsSentWithResponse()
         using var applicationBuilder = TestDistributedApplicationBuilder.Create();
         var builder = applicationBuilder.AddResource(testResource);
         builder.WithCommand(
-            type: "TestType",
+            name: "TestType",
```
drewnoakes, line 82:
```suggestion
            name: "TestName",
```

drewnoakes (approved):
Looks fine, other than compile error in tests. I think "Id" would be a better term as it indicates the uniqueness requirement more so than "Name".

DamianEdwards:
I think we need to do regular breaking change notifications as we changed this after the rc.

JamesNK:
I prefer name because resources have a name, and endpoints have a name. Command names would be consistent.

And I think name makes more sense because the value is being provided by the user. Id feels more appropriate if the value is auto-generated.

davidfowl (approved):

davidfowl:
It turns out we shipped commands in the resource server protocol so this change breaks ACA and we need to either revert all of it, or deprecate the old fields and add the new fields

PR #6359, opened 2024-10-17
Remove commas between endpoints in grid

drewnoakes:
## Description

The commas between endpoints are redundant as the underline of the anchor visually separates multiple entries. The comma visually collides with the underline, and looks a bit strange when there's only one item and the rest in overflow.

Subjective, but I think it looks a little sharper like this.

### Before

image

image

### After

image

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6359)


JamesNK:
> The commas between endpoints are redundant as the underline of the anchor visually separates multiple entries.

This won't be true for tcp endpoints. They're plain text.

However, maybe the space can be the delimiter. That shouldn't be a situation where an endpoint contains spaces.

drewnoakes:
> This won't be true for tcp endpoints. They're plain text.

True, some URLs won't be hyperlinked or underlined, but the space should still be clear.

DamianEdwards:
Yeah I like this change

adamint (approved):
Agree that space is an acceptable delimiter here

drewnoakes:
@adamint the test failure here seems related to health checks. Did we commit something that breaks tests?

```
Assert.Null() Failure: Value of type 'Nullable<HealthStatus>' has a value
Expected: null
Actual:   Healthy
```

PR #6366, opened 2024-10-18
Fix flaky test and add more test logging

JamesNK:
## Description

Adds more logging to tests.
Also improve flaky test to avoid potential race condition.

Fixes https://github.com/dotnet/aspire/issues/6354

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6366)

PR #6367, opened 2024-10-18
Don't force set health status if there is an initial snapshot

adamint:
## Description

There was one regression and one red herring from the health check playground project.

Regression: we should not force set the health status to healthy if there is an initial snapshot.

Red herring: if a health check is not run, the health status is not recomputed. Because of this, we should be specifying the health status in the initial snapshot in the playground project resources. 

@JamesNK @drewnoakes 

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6367)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -352,7 +352,6 @@ public class ResourceNotificationService
             var newState = stateFactory(previousState);
 
             newState = UpdateCommands(resource, newState);
-            newState = UpdateHealthStatus(resource, newState);
 
             notificationState.LastSnapshot = newState;
 
@@ -388,43 +387,6 @@ public class ResourceNotificationService
         }
 
         return Task.CompletedTask;
-    }
-
-    /// <summary>
-    /// Update resource snapshot health status if the resource is running with no health checks.
-    /// </summary>
-    private static CustomResourceSnapshot UpdateHealthStatus(IResource resource, CustomResourceSnapshot previousState)
-    {
-        // We can compute the resource health status based on its health reports if it
-        // - has no health checks
-        // - has an initial null health status (wasn't set by an initial snapshot)
-        // - is in the running state
-        //
-        // Since health reports may have been manually added
-        // resource health status uses the same algorithm as in ResourceHealthCheckService
-        // computing the health status based on the health reports reported
-        // and defaulting to Healthy
-        if (previousState.State?.Text == KnownResourceStates.Running)
-        {
-            var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
-
-            if (!hasHealthChecks)
-            {
-                var initialHealthStatus = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out var snapshotAnnotation) ? snapshotAnnotation.InitialSnapshot.HealthStatus : null;
-
-                // maybe the user has set a health status in the initial snapshot
-                // if so we should default to the initial health status
-                if (initialHealthStatus is null)
-                {
-                    return previousState with
-                    {
-                        HealthStatus = previousState.HealthReports.MinBy(r => r.Status)?.Status ?? HealthStatus.Healthy
-                    };
-                }
-            }
-        }
-
-        return previousState;
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -395,20 +395,22 @@ public class ResourceNotificationService
     /// </summary>
     private static CustomResourceSnapshot UpdateHealthStatus(IResource resource, CustomResourceSnapshot previousState)
     {
-        var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
-        var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
-
         // A resource is also healthy if it
         // - has a null health status (wasn't set by a health check nor initial snapshot)
         // - has no health check annotations
         // - is in the running state
         // - was not started with an initial snapshot
-        if (previousState.HealthStatus is null
-            && !hasHealthChecks
-            && previousState.State?.Text == KnownResourceStates.Running
-            && !hasInitialSnapshot)
+        if (previousState.HealthStatus is null)
         {
-            return previousState with { HealthStatus = HealthStatus.Healthy };
+            var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
+            var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
+
+            if (!hasHealthChecks
+                && previousState.State?.Text == KnownResourceStates.Running
+                && !hasInitialSnapshot)
+            {
+                return previousState with { HealthStatus = HealthStatus.Healthy };
+            }
         }
 
         return previousState;
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -399,10 +399,11 @@ public class ResourceNotificationService
         var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
 
         // A resource is also healthy if it
+        // - has a null health status (wasn't set by a health check nor initial snapshot)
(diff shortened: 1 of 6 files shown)


tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
```diff
@@ -87,20 +87,14 @@ public async Task WatchResources_ResourceHasCommands_CommandsSentWithResponse()
...
             updateState: c => ApplicationModel.ResourceCommandState.Enabled,
             displayDescription: "Display description!",
-            parameter: new [] {"One", "Two"},
+            parameter: new[] { "One", "Two" },
             confirmationMessage: "Confirmation message!",
             iconName: "Icon name!",
             iconVariant: ApplicationModel.IconVariant.Filled,
             isHighlighted: true);
 
-        await resourceNotificationService.PublishUpdateAsync(testResource, s =>
```
adamint, line 96: diff is messed up, look at side by side

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -395,8 +395,17 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
...
-        // A resource is also healthy if it has no health check annotations and is in the running state.
-        if (previousState.HealthStatus is not HealthStatus.Healthy && !resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _) && previousState.State?.Text == KnownResourceStates.Running)
+        var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
+        var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
+
+        // A resource is also healthy if it
+        // - has no health check annotations
+        // - is in the running state
+        // - was not started with an initial snapshot
+        if (previousState.HealthStatus is not HealthStatus.Healthy
```
JamesNK, line 405: If there is no initial snapshot, and there is no annotation, doesn't that mean `HealthStatus` must be null? That seems more specific test than not healthy.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -395,8 +395,17 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
...
+            && !hasHealthChecks
+            && previousState.State?.Text == KnownResourceStates.Running
```
JamesNK, line 407: Put the previousState checks together? Then inside the if statement, resolve the attributes then. That means they're only resolved when needed.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
adamint: > HealthStatus must be null

Yeah, good point. There's nothing that could have affected the health status. Updated

tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
JamesNK: What do you mean? Is this just whitespace change? Note that it will conflict with a change I've just made here: https://github.com/dotnet/aspire/pull/6366

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
adamint: done, thanks

tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
```diff
@@ -130,6 +124,46 @@ await resourceNotificationService.WaitForResourceAsync(testResource.Name, r =>
...
+
+        // Act
+        var task = dashboardService.WatchResources(
+            new WatchResourcesRequest(),
+            writer,
+            context);
+
+        // Assert
+        var initialRead = await writer.ReadNextAsync();
+        var initialSnapshot = Assert.Single(initialRead.InitialData.Resources);
```
JamesNK, line 160: This will be flaky. There is no guarentee that published resource will be in the initial data returned by the service.

Waiting for the resource mostly fixes it (see above test). However I believe you need to wait on another type to properly fix it. See https://github.com/dotnet/aspire/pull/6366 and the changes I made on the `DashboardDataService`.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -395,10 +395,22 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
...
+        // - is in the running state
+        // - was not started with an initial snapshot
+        if (previousState.HealthStatus is null)
         {
-            return previousState with { HealthStatus = HealthStatus.Healthy };
+            var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
+            var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
+
+            if (!hasHealthChecks
+                && previousState.State?.Text == KnownResourceStates.Running
```
JamesNK, line 409: nit: Check at the same time as health status. It is cheaper that getting annotations.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -395,10 +395,22 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
...
+        // A resource is also healthy if it
+        // - has a null health status (wasn't set by a health check nor initial snapshot)
+        // - has no health check annotations
+        // - is in the running state
+        // - was not started with an initial snapshot
+        if (previousState.HealthStatus is null)
         {
-            return previousState with { HealthStatus = HealthStatus.Healthy };
+            var hasHealthChecks = resource.TryGetAnnotationsIncludingAncestorsOfType<HealthCheckAnnotation>(out _);
+            var hasInitialSnapshot = resource.TryGetLastAnnotation<ResourceSnapshotAnnotation>(out _);
```
JamesNK, line 406: I don't think this works. Someone could add a health report with ResourceNotificationService after starting a resource.

JamesNK: For example:

```cs
await notificationService.PublishUpdateAsync(
    resource,
    state => state with
    {
        State = new("Running", "success"),
        HealthReports = [new HealthReportSnapshot($"{name}_check", status, description, exception)]
    });
```

tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
adamint: so is this dependent on #6366?

JamesNK: You can do what the test above is currently doing. It's 99.9% stable. Create an issue to update it once 6366 is merged.

tests/Aspire.Hosting.Tests/Dashboard/DashboardServiceTests.cs
adamint: No, a new test has been added in that placement - we can fix the merge conflict

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
adamint: What do you think about 
- removing this method entirely
- on the dashboard side, say a resource with a null health status is healthy if it is running with 0 health reports
- not make any assumptions other than above about the resource status - if users are updating health reports themselves, they need to handle health status too

@JamesNK

JamesNK: Here is what you could do:

1. Exit out of the method if `HealthCheckAnnotation` is set. Assume that the system will manage updating health status
2. If state is not running then set HealthStatus to null. If it's not running then there is no status. This is needed for resources that start/stop. Shouldn't have a healthy status if the app isn't running.
3. If state is running then look at health reports to figure out status. No reports = healthy. A degragded report = degradged, etc.

In other words, let aspire manage the health status if `HealthCheckAnnotation` is set, otherwise the health reports on the snapshot are the source of truth.

Having both health status and health reports on the snapshot be mutable is a point of friction. Either they can go out of sync (could be unexpected) or the health reports overwrite what is manually set.

PR #6368, opened 2024-10-18
Make resource HealthStatus computed from HealthReports

adamint:
## Description

The purpose of this PR is to make the `HealthStatus` field of a resource no longer settable, and instead have it computed from the `HealthReports` and `State` fields.

HealthStatus is computed as
- null, if the resource is not `Running`
- unhealthy, if the resource is running, and has a health check that has not returned a value yet
- healthy, if there are no health checks
- otherwise whatever the min of the returned health check values is

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6368)


src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -34,8 +34,22 @@ public sealed class ResourceViewModel
...
+        {
+            if (HealthReports.Length == 0 && KnownState == KnownResourceState.Running)
+            {
+                // If there are no health reports and the resource is running, assume it's healthy.
+                return Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus.Healthy;
+            }
+
+            return HealthReports.MinBy(r => r.HealthStatus)?.HealthStatus;
+        }
+    }
```
davidfowl, line 51: I think we want to compute it once when data is received.

src/Aspire.Dashboard/ResourceService/Partials.cs
```diff
@@ -49,7 +49,6 @@ public ResourceViewModel ToViewModel(BrowserTimeProvider timeProvider, IKnownPro
                 KnownState = HasState ? Enum.TryParse(State, out KnownResourceState knownState) ? knownState : null : null,
                 StateStyle = HasStateStyle ? StateStyle : null,
                 Commands = GetCommands(),
-                HealthStatus = HasHealthStatus ? MapHealthStatus(HealthStatus) : null,
```
davidfowl, line 52: Compute it here.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -52,11 +52,19 @@ public sealed record CustomResourceSnapshot
...
+    public HealthStatus? GetHealthStatus()
+    {
+        if (HealthReports.Length == 0 && State == "Running")
+        {
+            // If there are no health reports and the resource is running, assume it's healthy.
+            return HealthStatus.Healthy;
+        }
+
+        return HealthReports.MinBy(r => r.Status)?.Status;
+    }
```
davidfowl, line 67: I'd like to keep this a property. There's a change @drewnoakes made originally which I prefer:

Set this internally when the HealthReports are set and do the computation there/

src/Aspire.Dashboard/Model/ResourceViewModel.cs
adamint: done

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -34,11 +35,33 @@ public sealed class ResourceViewModel
...
+            _healthReports = value;
+            HealthStatus = ComputeHealthStatus(value, KnownState);
+        }
+    }
+
     public KnownResourceState? KnownState { get; init; }
 
+    private static HealthStatus? ComputeHealthStatus(ImmutableArray<HealthReportViewModel> healthReports, KnownResourceState? state)
+    {
+        if (healthReports.Length == 0 && state is KnownResourceState.Running)
```
JamesNK, line 56: If the resource isn't running then return null.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -86,6 +95,17 @@ public sealed record CustomResourceSnapshot
     /// The commands available in the dashboard for this resource.
     /// </summary>
     public ImmutableArray<ResourceCommandSnapshot> Commands { get; init; } = [];
+
+    private static HealthStatus? ComputeHealthStatus(ImmutableArray<HealthReportSnapshot> healthReports, string? state)
+    {
+        if (healthReports.Length == 0 && state == "Running")
```
JamesNK, line 101: Same as previous comment. Also, there should be a known state constant for "Running"

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -208,8 +208,6 @@ message Resource {
     // The set of volumes mounted to the resource. Only applies to containers.
     repeated Volume volumes = 15;
 
-    // The aggregate health state of the resource. Generally reflects data from health_reports, but may differ.
-    optional HealthStatus health_status = 16;
```
JamesNK, line 212: Add 16 to reserved fields (line 181)

src/Aspire.Dashboard/Model/ResourceViewModel.cs
adamint: AppHost will already be sending null

src/Aspire.Dashboard/Model/KnownPropertyLookup.cs
```diff
@@ -27,7 +27,6 @@ public KnownPropertyLookup(IStringLocalizer<Resources.Resources> loc)
             new(KnownProperties.Resource.StartTime, loc[nameof(ResourcesDetailsStartTimeProperty)]),
             new(KnownProperties.Resource.StopTime, loc[nameof(ResourcesDetailsStopTimeProperty)]),
             new(KnownProperties.Resource.ExitCode, loc[nameof(ResourcesDetailsExitCodeProperty)]),
-            new(KnownProperties.Resource.HealthState, loc[nameof(ResourcesDetailsHealthStateProperty)])
```
JamesNK, line 30: We still want to display calculated health status in resources propertries in details view

src/Aspire.Dashboard/Model/ResourceViewModel.cs
JamesNK: What will be null? Do you mean there won't be any reports if not running? Or the reports will all have null health status if not running?

JamesNK:
I tested your branch on health checks sandbox and got this error:

image

Known state propertiy could be set before health reports.

Could fix by defaulting it to an empty array:

```cs
private readonly ImmutableArray<HealthReportViewModel> _healthReports = [];
```

JamesNK:
Health status is inconsistently being shown in resource details:

Visible:
image

Not visible:
image

JamesNK:
While you're making changes in this area, I noticed the `Health State` casing is consistent with other properties. It should be `Health state`

adamint:
> Health status is inconsistently being shown in resource details:
> 
> Visible: image
> 
> Not visible: image

I was just thinking about this! The property does not appear when the resource has a health report with a null value. We encode that assumption in `ResourceStateViewModel` but it should really be encoded in `HealthStatus` itself, which fixes the bug. Made this change

src/Aspire.Dashboard/Model/ResourceStateViewModel.cs
```diff
@@ -62,12 +62,6 @@ internal static ResourceStateViewModel GetStateViewModel(ResourceViewModel resou
             icon = new Icons.Filled.Size16.Circle();
             color = Color.Info;
         }
-        else if (resource.HealthStatus is null)
```
adamint, line 65: This condition is redundant

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -20,6 +20,8 @@ namespace Aspire.Dashboard.Model;
 [DebuggerDisplay("Name = {Name}, ResourceType = {ResourceType}, State = {State}, Properties = {Properties.Count}")]
 public sealed class ResourceViewModel
 {
+    private readonly ImmutableArray<HealthReportViewModel> _healthReports = [];
+    private readonly KnownResourceState? _knownState;
     public required string Name { get; init; }
```
JamesNK, line 25: nit
```suggestion
    private readonly KnownResourceState? _knownState;

    public required string Name { get; init; }
```

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -34,10 +36,45 @@ public sealed class ResourceViewModel
...
+        }
+    }
+
+    public KnownResourceState? KnownState
+    {
+        get => _knownState;
+        init
+        {
+            _knownState = value;
+            ComputeHealthStatus(HealthReports, value);
```
JamesNK, line 59: ComputeHealthStatus is called here but not assigned to anything.

JamesNK (commented):
Looking pretty good. Needs tests.

adamint:
> Looking pretty good. Needs tests.

`ComputeHealthStatus` should be tested but otherwise it is pretty covered

JamesNK (approved):

adamint:
@JamesNK could you look again with the latest commit? 😄

JamesNK:
> Disambiguate have not received health reports and empty health reports

What is the scenario when this matters?

adamint:
> > Disambiguate have not received health reports and empty health reports
> 
> What is the scenario when this matters?

HealthReports may not be set initially. If the state becomes Running before it is set, the resource will erroneously report Healthy if it thinks there are no health reports.

adamint:
> > HealthReports may not be set initially. If the state becomes Running before it is set in the health check service, the resource will erroneously report Healthy for a little while if it thinks there are no health reports.
> 
> What is the alternative? Won't this cause the problem in the opposite direction: erroneously show a **running resource as unhealthy until it sets no health reports**

`ComputeHealthStatus` returns null health status until health reports are set.

> Are you concerned about properties being initialized and the health reports and state being set seperately?

Yes, exactly. Though they are both init-properties, `HealthReports` is not ever set if there are no health check annotations on a resource. (ResourceHealthCheckService#48)

> They're both `init` properties. They must be set when resource snapshot/viewmodel are created. There isn't an opportunity to get `HealthStatus` inbetween setting them.

They're not required properties, so can be uninitialized. In the split second between the resource running event being posted, and health reports being set on the resource, that's when, for resources with health checks, the health status would be incorrect. Instead, we don't know the health status then so should return null. That means we also need to publish an update on resources that don't have health checks, after the running event, setting HealthReports to []. I think this handles that case

davidfowl:
At this point we need to make sure this doesn't happen in our primary use case. Health checks on resources being updated by code we own in the apphost.

davidfowl:
Not sure about that change. Is that supposed to be the first publish?

adamint:
> Not sure about that change. Is that supposed to be the first publish?

That wouldn't be the first publish. Health reports are also updated after the running event is published, [for resources with health checks](https://github.com/dotnet/aspire/blob/908ad8d51b55f66c7fd2e2edaaca89c55fdd26e5/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs#L100)

adamint:
Linux tests keep failing because of a timeout

davidfowl:
Could it be a test?

adamint:
> Could it be a test?

Possibly but I am not finding anything in logs, and related hosting tests are passing locally. Will look into it

mitchdenny:
In my PR over here I am disabling all the resource health check tests as an experiment.:

https://github.com/dotnet/aspire/pull/6275/checks?check_run_id=31730392032

These tests are timing dependent (and use a few TCS and reset events) to control flow. Its possible that they are the cause of the hangs. I want to rule them in or out of being a factor here.

JamesNK:
> They're not required properties, so can be uninitialized. In the split second between the resource running event being posted, and health reports being set on the resource, that's when, for resources with health checks, the health status would be incorrect. It would be reported as `Healthy`, not `null` - if people listen to the resource running event, and check the health status property, it would be wrong. Since we don't know the health status, it should return null. And then an update on resources that don't have health checks needs to be published after the running event, setting HealthReports to [].

Ok, so is this the scenario? (this is what I mean when I ask what a scenario is, the impact of the change mapped to the real world)

1. Resource with health checks definied is in a starting state with no health reports
2. Resource reaches running state. Notification is pushed. HealthStatus is in a bad state.
3. Health checks logic waits for resource to reach running state, then adds health reports to the resource. Notification is pushed. HealthStatus is in a good state.

Maybe the solution is to have health reports set when the resource is created. We know what they are from the annotation. The health status can just be null.

mitchdenny:
There is no need to run a health check before the resource is in a running state because almost definitionally it would be unhealthy (including ancestors).

mitchdenny:
I'm curious to see if this test passes because the Resource Health Check Tests are disabled. It'll be good to narrow down which one is causing the problem in CI since we can't repro locally. So if this passes we should enable half the tests and trigger CI again.

adamint:
> I'm curious to see if this test passes because the Resource Health Check Tests are disabled. It'll be good to narrow down which one is causing the problem in CI since we can't repro locally. So if this passes we should enable half the tests and trigger CI again.

binary search!

mitchdenny:
OK so try the other half now :)

mitchdenny:
Fixed up the playground so that it continues to work with the API changes however the test case has a build failure. But I'm not sure whether I should be changing that @adamint.

adamint:
> Fixed up the playground so that it continues to work with the API changes however the test case has a build failure. But I'm not sure whether I should be changing that @adamint.

My goodness, @mitchdenny tests are passing now... I don't know what infrastructure-related thing was going on

adamint:
@JamesNK @mitchdenny this is ready for re-review.

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -68,6 +68,12 @@ public ResourceStateSnapshot? State
     /// </remarks>
     public HealthStatus? HealthStatus { get; private set; }
 
+    /// <summary>
+    /// Whether the health reports have been initialized in <see cref="ResourceNotificationService"/> to contain
+    /// the health check annotations for this resource.
+    /// </summary>
+    internal bool HealthAnnotationsInitialized { get; init; }
```
davidfowl, line 75: Why do  you need this if it happens once?

adamint: It's not needed, the usages were removed but I missed the actual property declaration. It's been removed

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
```diff
@@ -50,31 +49,9 @@ static GenericResourceSnapshot CreateResourceSnapshot(IResource resource, string
                     State = snapshot.State?.Text,
                     StateStyle = snapshot.State?.Style,
                     HealthStatus = snapshot.HealthStatus,
-                    HealthReports = GetOrCreateHealthReports(),
+                    HealthReports = snapshot.HealthReports,
                     Commands = snapshot.Commands
                 };
-
-                ImmutableArray<HealthReportSnapshot> GetOrCreateHealthReports()
```
JamesNK, line 57: What is the reason for removing this?

I think we want to see health reports for a resource even if it isn't running yet. The health reports are still in a running state. The health reports would show a waiting status. Is this lost?

Alternatively, is it no longer needed because the snapshot always contains the health reports for health check annotations?

playground/HealthChecks/HealthChecksSandbox.AppHost/Program.cs
```diff
@@ -24,28 +25,22 @@
...
-{
     return builder
         .AddResource(new TestResource(name))
+        .WithHealthCheck($"{name}_check")
         .WithInitialState(new()
         {
             ResourceType = "Test Resource",
             State = "Starting",
             Properties = [],
-            HealthReports = [new HealthReportSnapshot($"{name}_check", status, description, exception)]
```
JamesNK, line 48: Why switch to the annotation? I'm not against the sandbox having health checks driven by the annotation - it's the default experience after all - but it's useful to have some resources set health reports in a snapshot manually to double check that scenario isn't broken.

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
JamesNK: Done now in `GetInitialHealthReports`?

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -93,12 +93,8 @@ await resourceNotificationService.PublishUpdateAsync(resource, s =>
                 {
                     var healthReports = MergeHealthReports(s.HealthReports, report);
 
-                    // Matches the logic in ASP.NET Core's private HealthReport.CalculateAggregateStatus
-                    var healthStatus = healthReports.MinBy(r => r.Status)?.Status ?? s.HealthStatus;
-
                     return s with
                     {
-                        HealthStatus = healthStatus,
                         HealthReports = healthReports
```
JamesNK, line 98: nit: Add a comment that the health status on the snapshot is automatically re-calculated from health reports.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -210,6 +210,7 @@ message Resource {
 
     // The aggregate health state of the resource. Generally reflects data from health_reports, but may differ.
     optional HealthStatus health_status = 16;
```
JamesNK, line 212: Was there offline discussion about adding this back? I'm not against it, just want to make sure that was an agreed decision.

JamesNK:
Looks good. I have some questions, nits and changes to the sandbox.

src/Aspire.Hosting/Dashboard/DashboardServiceData.cs
davidfowl: Yep

playground/HealthChecks/HealthChecksSandbox.AppHost/Program.cs
davidfowl: We don’t care about this for 9. We cns re-evaluate if it becomes a real scenario

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
davidfowl: No we want to remove it

JamesNK: I know I said before I didn't want it, but sending the calculated value does reduce duplication. It lets the dashboard be more dumb and accept whatever status the service decides 😄 

I haven't got a hard opinion either way, but the current code looks fine to me.

playground/HealthChecks/HealthChecksSandbox.AppHost/Program.cs
JamesNK: Fair. But, sandboxes are a place to try out dumb stuff. I'd like to keep some examples that manually set health reports just to see what happens if someone does do it.

adamint: `HealthReports` init is now internal. Users cannot set health reports anymore in snapshots

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
adamint: @davidfowl I had re-added it to the proto to do what james said - avoid re-calculating the status in the dashboard and avoid duplication

adamint:
Is this now in a merge-able state?

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
davidfowl: Remove it, let the dashboard recalculate it.

playground/HealthChecks/HealthChecksSandbox.AppHost/Program.cs
JamesNK: Oh ok. Never mind then.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
adamint: Re-removed :) is everything else ok?

JamesNK (approved):

davidfowl:
@adamint I suggest you do a squash and a manual backport. This has 36 commits.

davidfowl (approved):
Looks great

adamint:
There was a health check scenario that has never been accounted for - 

1. Resource has multiple health checks
2. Resource checks health, one report returns a new status but the overall health status has not changed (ie, before Unhealthy+Healthy=Unhealthy -> Unhealthy+Degraded=Unhealthy)
3. The report health is the same as the prior snapshot health status
4. Resource health reports are not updated, even though the status of one of them changed.

I have updated the sandbox and [ResourceHealthCheckService](https://github.com/dotnet/aspire/pull/6368/files#diff-50754331448632c7053f5ef527d8fb5d35a6eb976d2f7984d493c2ea3f23f76cR86) to account for this.

davidfowl:
Good catch @adamint !

PR #6377, opened 2024-10-18
[release/8.2] Update dependencies

joperezr:
Updating NuGet and docker dependencies to latest.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6377)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -109,7 +109,7 @@
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.10.2" />
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.10.2" />
     <PackageVersion Include="Milvus.Client" Version="2.3.0-preview.1"/>
-    <PackageVersion Include="MongoDB.Driver" Version="3.0.0" />
+    <PackageVersion Include="MongoDB.Driver" Version="2.30.0" />
     <PackageVersion Include="MongoDB.Driver.Core.Extensions.DiagnosticSources" Version="1.5.0" />
     <PackageVersion Include="MySqlConnector.DependencyInjection" Version="2.3.6" />
     <PackageVersion Include="MySqlConnector.Logging.Microsoft.Extensions.Logging" Version="2.1.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -106,8 +106,8 @@
     <PackageVersion Include="Humanizer.Core" Version="2.14.1" />
     <PackageVersion Include="KubernetesClient" Version="15.0.1" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="5.2.2" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.10.2" />
-    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.10.2" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.9.3" />
+    <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.9.3" />
     <PackageVersion Include="Milvus.Client" Version="2.3.0-preview.1"/>
     <PackageVersion Include="MongoDB.Driver" Version="[2.30.0,3.0.0)" />
     <PackageVersion Include="MongoDB.Driver.Core.Extensions.DiagnosticSources" Version="1.5.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -109,7 +109,7 @@
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components" Version="4.10.2" />
     <PackageVersion Include="Microsoft.FluentUI.AspNetCore.Components.Icons" Version="4.10.2" />
     <PackageVersion Include="Milvus.Client" Version="2.3.0-preview.1"/>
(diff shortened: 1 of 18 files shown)


Directory.Packages.props
```diff
@@ -10,29 +10,29 @@
...
+    <PackageVersion Include="AWSSDK.Core" Version="3.7.400.36" />
     <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.301" />
     <PackageVersion Include="AWS.Messaging" Version="0.9.2" />
-    <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.5" />
+    <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.6" />
     <PackageVersion Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
     <!-- Azure SDK for .NET dependencies -->
-    <PackageVersion Include="Azure.AI.OpenAI" Version="2.0.0-beta.5" />
-    <PackageVersion Include="Azure.Data.Tables" Version="12.9.0" />
+    <PackageVersion Include="Azure.AI.OpenAI" Version="2.1.0-beta.1" />
```
eerhardt, line 22:
```suggestion
    <PackageVersion Include="Azure.AI.OpenAI" Version="2.0.0-beta.5" />
```

This needs to be reverted.

Directory.Packages.props
```diff
@@ -144,24 +144,24 @@
...
-    <PackageVersion Include="Testcontainers.Elasticsearch" Version="$(TestcontainersPackageVersion)" />
+    <PackageVersion Include="Testcontainers.MsSql" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.MySql" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.PostgreSQL" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.RabbitMq" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.Redis" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.Nats" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.Milvus" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.Oracle" Version="3.10.0" />
+    <PackageVersion Include="Testcontainers.Elasticsearch" Version="3.10.0" />
```
eerhardt, line 164: Revert this and update the property above.

Directory.Packages.props
```diff
@@ -170,10 +170,12 @@
...
     <!-- Pinned version for Component Governance - Remove when root dependencies are updated -->
-    <PackageVersion Include="System.Text.Json" Version="8.0.4" />
-    <PackageVersion Include="Azure.Core" Version="1.42.0" />
+    <PackageVersion Include="System.Text.Json" Version="8.0.5" />
+    <PackageVersion Include="Azure.Core" Version="1.44.1" />
     <PackageVersion Include="Azure.Identity" Version="1.12.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" /> <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
     <PackageVersion Include="System.Formats.Asn1" Version="8.0.1" />
+    <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Text.RegularExpressions" Version="4.3.1" />
```
eerhardt, line 179: These shouldn't be necessary. Why are they being added?

Directory.Packages.props
joperezr: If you see there is one testcontainer that didn't actually change which is why I didn't, but I could of course do the other way and update the property and just hardcode the one that didn't update

Directory.Packages.props
joperezr: I turned on the nuget audit temporarily on this branch before pushing to make sure we didn't have any vulnerable transitive dependencies, and it was complaining about these two.

eerhardt: We should find where those 2 are coming from and fix the real location. No one should be depending on these packages anymore.

eerhardt: I've removed these lines because we shouldn't be making this change, which could affect our product assets.

I dug into this and the only things that are pulling in old versions of these packages are tests. and that is because we are using a super old xunit.assert version - 2.4.2 - over 2 years old.

We should just update our xunit.assert, but not in this PR. I will open one for main.

src/Components/Aspire.Azure.AI.OpenAI/AspireAzureOpenAIExtensions.cs
```diff
@@ -90,7 +90,7 @@ protected override IAzureClientBuilder<AzureOpenAIClient, AzureOpenAIClientOptio
 
                     if (!string.IsNullOrEmpty(settings.Key))
                     {
-                        var credential = new AzureKeyCredential(settings.Key);
+                        var credential = new ApiKeyCredential(settings.Key);
```
eerhardt, line 93: This is working around a binary breaking change between beta versions of System.ClientModel 1.1.0. See https://github.com/Azure/azure-sdk-for-net/pull/45554. The Azure.AI.OpenAI 2.0.0-beta5 version we are using was using that implicit operator, which is no longer there in the released 1.1.0 GA version. Using the ApiKeyCredential directly side-steps the implicit operator break.

eerhardt (approved):
LGTM. Hopefully CI passes someday.

eerhardt:
@joperezr - looks like this needs 2 approvals.

joperezr:
Thanks for the help pushing this through

PR #6379, opened 2024-10-18
Added support for probes on resources.

GMouaad:
## Description

Fixes # (890)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [ ] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6379)

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/IResourceWithProbes.cs
+++ b/src/Aspire.Hosting/ApplicationModel/IResourceWithProbes.cs
@@ -1,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents a resource that has probes associated with it.
+/// </summary>
+public interface IResourceWithProbes : IResourceWithEndpoints
+{
+}
--- a/src/Aspire.Hosting/ApplicationModel/ProbeAnnotation.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ProbeAnnotation.cs
@@ -1,0 +1,85 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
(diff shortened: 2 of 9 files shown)


davidfowl:
@GMouaad Are you going to go a step further and use WithHttp(s) health check on resources with probes?

mitchdenny:
@davidfowl @maddymontaquila @DamianEdwards I think that this is something we should consider for 9.x. Overall I think this approach is sound. One open question is whether we want to automatically inject probes for .NET projects.

We'll also need logic to translate these probes when `PublishAsAzureContainerApp` is used.

davidfowl:
>We'll also need logic to translate these probes when PublishAsAzureContainerApp is used.



Yep! Now we can actually do that 😄

DamianEdwards:
> One open question is whether we want to automatically inject probes for .NET projects.

There's a few challenges with this at the moment:

1. The health endpoints for projects are defined in the service defaults project and are completely customizable. Ideally we'd have a way for the definitions be shared and thus have a single authority.
2. The default health endpoints added by the service defaults template aren't secured or isolated today, and thus are only added in the development environment. To add them by default we'd need to revisit this, e.g. isolate them such that they're not exposed on external networks by using host filtering, etc.

GMouaad:
@davidfowl from what I understood, moving forward, we want to add Http health checks under the hood when calling `WithProbe(..)` and let the AppHost deal with the probing during development and the whatever infrastructure (ACA,AKS,..) in prod.

danmoseley:
I'm going to close this stale PR. Feel free to reopen if you decide to continue.

davidfowl:
@DamianEdwards

PR #6383, opened 2024-10-18
Set additional environment variable to support Functions telemetry

captainsafia:
We need to set an additional environment variable in order to indicate to the functions host that it should emit OpenTelemetry traces.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6383)

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -100,6 +100,8 @@ public static class AzureFunctionsProjectResourceExtensions
                 context.EnvironmentVariables["OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY"] = "in_memory";
                 context.EnvironmentVariables["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true";
                 context.EnvironmentVariables["FUNCTIONS_WORKER_RUNTIME"] = "dotnet-isolated";
+                // Required to enable OpenTelemetry in the Azure Functions host.
+                context.EnvironmentVariables["AzureFunctionsJobHost__telemetryMode"] = "OpenTelemetry";
                 // Set ASPNETCORE_URLS to use the non-privileged port 8080 when running in publish mode.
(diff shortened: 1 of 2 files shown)


fabiocav (approved):

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -100,6 +100,8 @@ public static class AzureFunctionsProjectResourceExtensions
                 context.EnvironmentVariables["OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY"] = "in_memory";
                 context.EnvironmentVariables["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true";
                 context.EnvironmentVariables["FUNCTIONS_WORKER_RUNTIME"] = "dotnet-isolated";
+                // Required to enable OpenTelemetry in the Azure Functions host.
+                context.EnvironmentVariables["AzureFunctionsJobHost__telemetryMode"] = "OpenTelemetry";
```
eerhardt, line 104: @fabiocav - why does this not follow the same naming pattern as `FUNCTIONS_WORKER_RUNTIME`? Different prefix, double underscores, etc.

eerhardt: Is this the exact casing of the environment variable? I believe that linux environment variables are case sensitive.

davidfowl: IConfiguration

fabiocav: > IConfiguration

Yes. The other variables are explicitly consumed as environment variables by the host, while the new variable being added overrides the configuration typically defined in `host.config`, which is pulled into configuration and mapped to a section named `AzureFunctionsJobHost`.

davidfowl (approved):

eerhardt (approved):

PR #6387, opened 2024-10-19
Hosting test timeouts

JamesNK:
## Description

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6387)


mitchdenny:
I think merging this is a good idea. The only question is about timing (in terms of when to merge AND individual timeouts).

Some of our test cases need to be modified to use time provider but that will take some time to do that so some tests might become flakey initially whilst that gets worked through. So should we wait until 9.0 has a stable build.

(acknowledging that we are currently being impacted by flakey tests).

tests/Aspire.StackExchange.Redis.Tests/RedisContainerFixture.cs
```diff
@@ -34,7 +34,7 @@ public async Task DisposeAsync()
     public static async Task<RedisContainer> CreateContainerAsync()
     {
         var container = new RedisBuilder()
-            .WithImage($"{TestConstants.AspireTestContainerRegistry}/{RedisContainerImageTags.Image}:{RedisContainerImageTags.Tag}")
+            .WithImage($"{ComponentTestConstants.AspireTestContainerRegistry}/{RedisContainerImageTags.Image}:{RedisContainerImageTags.Tag}")
```
mitchdenny, line 37: Unrelated?

JamesNK: I added a TestConstants type and renamed this one to avoid conflict.

JamesNK:
> Some of our test cases need to be modified to use time provider but that will take some time to do that so some tests might become flakey initially whilst that gets worked through. So should we wait until 9.0 has a stable build.

Do you want this to be in the 9.0 branch? While these are just test changes, there are a lot of test changes. There are benefits: it would make backporting test changes easier. Could make 9.0 branch less flaky.

I was imagining these changes would be merged on main and not be backported.

JamesNK:
I don't want this PR to rot. If makes changes all over the place and will gather conflicts easily.

What feedback do you have? What changes are required before it can be merged?

mitchdenny:
If you are suggesting backporting this to release/9.0 I am OK with it ... but we should make sure we can get all the way to release/9.0 for sure before merging this to main otherwise every other backport will become painful because of the number of source files this change touches.

Fact that this is already starting to accumulate merge conflicts kind of makes my point. I'm 100% supportive of this change, it's just about timing (I like how this gives us localized control over wait behavior vs. some global approach).

adamint (approved):

JamesNK:
> If you are suggesting backporting this to release/9.0

No

mitchdenny:
I think this is a good change. Its heavy handed but I think test reliability will improve if we can more easily identify where hangs are occuring.

mitchdenny (approved):

PR #6419, opened 2024-10-21
Bump Microsoft.Extensions.Http.Resilience to 9.x

DamianEdwards:
## Description

Bump version of `Microsoft.Extensions.Http.Resilience` to current 9.x preview version on nuget.org and update version used by templates to track the same. This will get updated to stable as part of internal dependency flow.

Contributes to #6076

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] n/a
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6419)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -39,7 +39,7 @@
     <MicrosoftDotNetXUnitExtensionsPackageVersion>9.0.0-beta.24516.2</MicrosoftDotNetXUnitExtensionsPackageVersion>
     <MicrosoftDotNetBuildTasksInstallersPackageVersion>9.0.0-beta.24516.2</MicrosoftDotNetBuildTasksInstallersPackageVersion>
     <MicrosoftDotNetBuildTasksWorkloadsPackageVersion>9.0.0-beta.24516.2</MicrosoftDotNetBuildTasksWorkloadsPackageVersion>
-    <MicrosoftExtensionsHttpResiliencePackageVersion>8.10.0</MicrosoftExtensionsHttpResiliencePackageVersion>
+    <MicrosoftExtensionsHttpResiliencePackageVersion>9.0.0-preview.9.24507.7</MicrosoftExtensionsHttpResiliencePackageVersion>
     <MicrosoftExtensionsDiagnosticsTestingPackageVersion>8.10.0</MicrosoftExtensionsDiagnosticsTestingPackageVersion>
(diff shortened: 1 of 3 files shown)


eng/Versions.props
```diff
@@ -70,7 +70,7 @@
     <NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>8.0.8</NpgsqlEntityFrameworkCorePostgreSQLPackageVersion>
     <!-- for templates -->
     <MicrosoftExtensionsHttpResiliencePackageVersionForNet8>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet8>
-    <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>9.0.0-preview.9.24507.7</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
+    <MicrosoftExtensionsHttpResiliencePackageVersionForNet9>$(MicrosoftExtensionsHttpResiliencePackageVersion)</MicrosoftExtensionsHttpResiliencePackageVersionForNet9>
```
eerhardt, line 73: Is it possible to just delete these properties?

DamianEdwards: Yes, just a bigger change :smile:

DamianEdwards: Ok I removed them.

DamianEdwards:
@joperezr this change needs the new build since https://github.com/dotnet/extensions/commit/eafdf6e9c40bcd561f38979617405fd2801a46e3 went in to dotnet/extensions (with the TFM-split dependencies). The last version available on the dotnet8 feed is from the day before that change went in.

eerhardt (approved):
LGTM. Thanks!

DamianEdwards:
@eerhardt  @joperezr the test errors are caused by `Aspire.Hosting.Testing` depending on `Microsoft.Extensions.Configuration.Binder/8.0.2` which causes a package downgrade in test projects targeting `net9.0`. Do we need to TFM-split the dependencies of `Aspire.Hosting.Testing`?

eerhardt (approved):

PR #6425, opened 2024-10-22
Make RedisInsight work with WithLifetime(...).

### LINKED ISSUE #6111: WithRedisInsight and ContainerLifetime.Persistent causes multiple instances to show up
If I have the following:

```C#

var redis = builder.AddRedis("redis")
    .WithDataVolume()
    .WithRedisInsight(c => c.WithLifetime(ContainerLifetime.Persistent));

builder.AddProject<Projects.Redis_ApiService>("apiservice")
    .WithReference(redis).WaitFor(redis);

builder.Build().Run();
```

When I run multiple times, the RedisInights container stays running, but each time I run a new Redis instance shows up in it:

image

We should fix this so the same redis instance isn't added on re-run.

mitchdenny:
## Description

With this PR, when using `WithRedisInsight(...)` the list of databases will be fetched from the RedisInsight instance and where they overlap with the names of the Redis Instances running already it will delete them so that the subsequent import won't result in duplicated entries.

Fixes #6111

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6425)

### DIFF
--- a/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
@@ -2,8 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Globalization;
+using System.Net.Http.Json;
 using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Redis;
 using Aspire.Hosting.Utils;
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -192,20 +194,50 @@ public static IResourceBuilder<RedisResource> WithRedisInsight(this IResourceBui
             return builder;
         }
 
-        static async Task ImportRedisDatabases(ILogger resourceLogger, IEnumerable<RedisResource> redisInstances, HttpClient client, CancellationToken ct)
+        static async Task ImportRedisDatabases(ILogger resourceLogger, IEnumerable<RedisResource> redisInstances, HttpClient client, CancellationToken cancellationToken)
```
davidfowl, line 197: Simpler to just clear it everytime instead?

mitchdenny: I was thinking that if someone had manually configured Redis connection to another container then they might prefer if we didn't delete it.

I had contemplated not removing them and just not importing if the name matched, but that wouldn't work with password change occurrences.

eerhardt:
Can you add some tests for this?

mitchdenny:
@eerhardt working on it. Turns out we don't have any tests for WithLifetime(...) so I'm having to tackle a few first of a kind problems:

1. ResourceNameSuffix variations whilst under test.
2. Cleaning up "persistent" containers after the test.

mitchdenny:
OK for the time being I've done a bit of a hack to make sure we have test coverage here. See the introduction of `ApplicationExecutorProxy`. Check out https://github.com/dotnet/aspire/issues/6455 for more details on what I think the post 9.0 solution should look like.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -120,6 +121,111 @@ public async Task VerifyRedisResource()
...
+        //       name as known from a DCP perspective. I then use the ApplicationExecutorProxy (introduced with this
+        //       change to call the ApplicationExecutor stop method. The proxy is a public type with an internal
+        //       constructor inside the Aspire.Hosting.Tests package. This is a short term solution for 9.0 to
+        //       make sure that we have good test coverage for WithRedisInsight behavior, but we need a better
+        //       long term solution in 9.x for folks that will want to do things like execute commands against
+        //       resources to stop specific containers.
+        var rns = app2.Services.GetRequiredService<ResourceNotificationService>();
+        var latestEvent = await rns.WaitForResourceHealthyAsync(redisInsightBuilder.Resource.Name, cts.Token);
+        var executorProxy = app2.Services.GetRequiredService<ApplicationExecutorProxy>();
+        await executorProxy.StopResourceAsync(latestEvent.ResourceId, cts.Token);
```
eerhardt, line 224: Can you get the `Stop` command for the resource and execute it? Simulating someone clicking the "Stop" button in the Dashboard.

mitchdenny: Similar problem. The command executor is internal. I've created an issue to look into this area in 9.x:

https://github.com/dotnet/aspire/issues/6455

eerhardt: It feels like the command executor not being public is a bigger issue. How do I write functional tests for my commands?

cc @JamesNK

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -192,20 +194,50 @@ public static IResourceBuilder<RedisResource> WithRedisInsight(this IResourceBui
...
+
+                var databasesToDelete = new List<Guid>();
+
                 using var writer = new Utf8JsonWriter(stream);
 
                 writer.WriteStartArray();
 
                 foreach (var redisResource in redisInstances)
                 {
+                    if (lookup is { } && lookup.Contains(redisResource.Name))
```
eerhardt, line 228: Why are we deleting the servers that are already there, only to add them again?

Wouldn't the more appropriate approach be "if the server already exists, don't add it again" ?

mitchdenny: Because the credentials could have changed. Using the `api/databases` call we don't have a mechanism to check that the credentials are the same, so we just blow it away and recreate.

eerhardt: It would be good to add this in a comment. I don't think this algorithm is intuitive.

eerhardt (approved):
The change looks fine. I'd just add some comments on why we are doing it this way.

davidfowl (approved):

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
eerhardt: Did this ever happen?

PR #6434, opened 2024-10-22
Mark Aspire.Hosting.Python as experimental

### LINKED ISSUE #6298: Mark AddPythonApp API as experimental

sebastienros:
## Description

Fixes #6298

Diagnostic documented here https://github.com/dotnet/docs-aspire/pull/1880

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
    - Link to aspire-docs issue: 
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6434)

### DIFF
--- a/src/Aspire.Hosting.Python/AssemblyInfo.cs
+++ b/src/Aspire.Hosting.Python/AssemblyInfo.cs
@@ -1,0 +1,6 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: Experimental("ASPIREHOSTINGPYTHON001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
--- a/tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
+++ b/tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
@@ -1,8 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#pragma warning disable ASPIREHOSTINGPYTHON001 // Test for experimental feature
-
 using Xunit;
 using Microsoft.Extensions.DependencyInjection;
 using Aspire.Hosting.Utils;
--- a/tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
+++ b/tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#pragma warning disable ASPIREHOSTINGPYTHON001 // Test for experimental feature
+
 using Xunit;
 using Microsoft.Extensions.DependencyInjection;
 using Aspire.Hosting.Utils;
--- a/playground/python/Python.AppHost/Program.cs
+++ b/playground/python/Python.AppHost/Program.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#pragma warning disable ASPIREHOSTINGPYTHON001 // Test for experimental feature
+
 var builder = DistributedApplication.CreateBuilder(args);
 
 builder.AddPythonApp("script-only", "../script_only", "main.py");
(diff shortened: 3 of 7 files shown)


tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#pragma warning disable ASPIREHOSTINGPYTHON001 // Test for experimental feature
```
eerhardt, line 4: (nit) can probably do this for the whole test assembly since it is testing the Python project.

eerhardt (commented):
We also need to "stablize" the package again.

https://github.com/dotnet/aspire/blob/a4ef97a3c716944aa737f8be48638166584fb7e0/src/Aspire.Hosting.Python/Directory.Build.props#L3-L6

tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs
sebastienros: I applied your suggestion but this is failing locally. Can you notice a mistake?

eerhardt: Don't use a suppressmessage attribute. But instead add it to the NoWarn property.

```xml
<PropertyGroup>
  <NoWarn>$(NoWarn);ASPIREHOSTINGPYTHON001</NoWarn>
```

sebastienros: Interesting, that's what I was just working on. So the attribute actually doesn't work? We should update the docs then. Can do it as part of the related PR, not just for this diagnostics.

src/Aspire.Hosting.Python/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: Experimental("ASPIREHOSTINGPYTHON001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
```
eerhardt, line 6: How do we get `https://learn.microsoft.com/en-us/dotnet/aspire/diagnostics/overview#ASPIREHOSTINGPYTHON001` to be a valid URL?

cc @IEvangelist

eerhardt (approved):
Code change looks good to me. We will need to get the doc site updated for the new diagnostic.

joperezr:
> Code change looks good to me. We will need to get the doc site updated for the new diagnostic.

@sebastienros can you please log an issue for this in the aspire-docs repo and paste it above in the PR description form? That way when filling that doc we'll have a link to the original PR.

eerhardt:
> > Code change looks good to me. We will need to get the doc site updated for the new diagnostic.
> 
> @sebastienros can you please log an issue for this in the aspire-docs repo and paste it above in the PR description form? That way when filling that doc we'll have a link to the original PR.

This happened with [Document experimental attributes for Aspire.Hosting.Python (dotnet/docs-aspire#1880)](https://github.com/dotnet/docs-aspire/pull/1880).

src/Aspire.Hosting.Python/AssemblyInfo.cs
IEvangelist: The `ASPIREHOSTINGPYTHON001` has to either be a heading, or you can use an inline named HTML element:

```markdown
<span name="ASPIREHOSTINGPYTHON001"></span>
```

Then the bookmark will work.

sebastienros: @IEvangelist when I create the doc I use a heading, same as the other sections, but the generated id is lower-cased so the current link doesn't work. I will add the <span> this way there is no change to do in aspire, but it's unfortunate. Maybe the doc generator could also inject a name with the actual case of the heading (though that would leak everywhere).

IEvangelist: Yeah, that's a `dotnet/docfx` concern more than anything else. Bookmarks in URLs just look better in lowercase in my opinion, but knowing that they're always in lowercase, why not just have this code ensure that they're lowercase?

sebastienros: I agree, had I known I would have done it before. But I don't want to block release/9.0 insertion on that. Will do it in main once 9.0 has shipped and is merged back.

PR #6458, opened 2024-10-23
[release/9.0] Make resource HealthStatus computed from HealthReports (#6368)

adamint:
Backport of #6368 to release/9.0

## Customer Impact

- Removes the ability to set HealthStatus directly on a resource, and removes the field from the resource service protocol
- Fixes a bug where health reports will not be updated if the overall health of the resource does not change
- Fixes a bug where the health status is not shown in resource details while waiting for health checks to initially run

## Testing

Much manual testing, as well as additional unit tests for computing health status.

## Risk

Medium. It is possible that this breaking change may impact existing projects that set HealthStatus, and may impact customer assumptions during resource updates - for example, resources now start as unhealthy if they are running and wait on a health check, instead of null, as it previously was.

## Regression?

No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6426)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6458)

### DIFF
--- a/tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
+++ b/tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
@@ -1,18 +1,35 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Immutable;
 using Aspire.Dashboard.Model;
 using Aspire.ResourceService.Proto.V1;
 using Google.Protobuf.WellKnownTypes;
 using Microsoft.Extensions.Logging.Abstractions;
 using Xunit;
-
+using DiagnosticsHealthStatus = Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus;
 namespace Aspire.Dashboard.Tests.Model;
 
 public sealed class ResourceViewModelTests
 {
     private static readonly DateTime s_dateTime = new(2000, 12, 30, 23, 59, 59, DateTimeKind.Utc);
     private static readonly BrowserTimeProvider s_timeProvider = new(NullLoggerFactory.Instance);
+
+    [Theory]
+    [InlineData(KnownResourceState.Starting, null, null)]
+    [InlineData(KnownResourceState.Starting, null, new string[]{})]
+    [InlineData(KnownResourceState.Starting, null, new string?[]{null})]
+    // we don't have a Running + HealthReports null case because that's not a valid state - by this point, we will have received the list of HealthReports
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Healthy, new string[]{})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Healthy, new string?[] {"Healthy"})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Unhealthy, new string?[] {null})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Degraded, new string?[] {"Healthy", "Degraded"})]
+    public void Resource_WithHealthReportAndState_ReturnsCorrectHealthStatus(KnownResourceState? state, DiagnosticsHealthStatus? expectedStatus, string?[]? healthStatusStrings)
+    {
+        var reports = healthStatusStrings?.Select<string?, HealthReportViewModel>((h, i) => new HealthReportViewModel(i.ToString(), h is null ? null : System.Enum.Parse<DiagnosticsHealthStatus>(h), null, null)).ToImmutableArray() ?? [];
+        var actualStatus = ResourceViewModel.ComputeHealthStatus(reports, state);
+        Assert.Equal(expectedStatus, actualStatus);
+    }
 
     [Fact]
     public void ToViewModel_EmptyEnvVarName_Success()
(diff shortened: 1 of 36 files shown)


davidfowl (approved):

tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
```diff
@@ -1,19 +1,36 @@
...
 
+    [Theory]
+    [InlineData(KnownResourceState.Starting, null, null)]
+    [InlineData(KnownResourceState.Starting, null, new string[]{})]
+    [InlineData(KnownResourceState.Starting, null, new string?[]{null})]
+    // we don't have a Running + HealthReports null case because that's not a valid state - by this point, we will have received the list of HealthReports
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Healthy, new string[]{})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Healthy, new string?[] {"Healthy"})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Unhealthy, new string?[] {null})]
+    [InlineData(KnownResourceState.Running, DiagnosticsHealthStatus.Degraded, new string?[] {"Healthy", "Degraded"})]
```
radical, line 26: Do we need tests for other states like `Exited`, or `FailedToStart`, `building` etc?

radical:
> Fixes a bug where health reports will not be updated if the overall health of the resource does not change
> Fixes a bug where the health status is not shown in resource details while waiting for health checks to initially run

Can we have tests for these?

tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
adamint: We don’t, all states other than Running are treated the same

adamint:
> Thanks @adamint, I also understand this is something that ACA team needs. Before we mark it as approved though, do you mind expanding a bit on the risk? So as opposed to just saying Medium, elaborate a bit more on what the potential issues are with this?

@joperezr sorry, the edit didn’t save. I’ve added this information now

adamint:
> > Fixes a bug where health reports will not be updated if the overall health of the resource does not change
> > Fixes a bug where the health status is not shown in resource details while waiting for health checks to initially run
> 
> Can we have tests for these?

There are tests for both of these!

tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
radical: > We don’t, all states other than Running are treated the same

Maybe tests to check for that?

adamint: That's definitionally true, but I can add these tests. Assuming they just need to go into main?

adamint:
> Fixes a bug where health reports will not be updated if the overall health of the resource does not change

I've added this in #6461

> Fixes a bug where the health status is not shown in resource details while waiting for health checks to initially run

There is already a test for this

tests/Aspire.Dashboard.Tests/Model/ResourceViewModelTests.cs
adamint: added this in #6461

joperezr:
Thanks for adding the additional info, @adamint . I'm trying to parse your risk assessment and specifically to understand who might be broken by this. Am I understanding correctly that in order to be in a broken state is if I am explicitly setting a status for a resource inside my AppHost code? If they wanted to continue to do that, how can they go about it? Also, assuming there is no way to do this in a non-breaking way?

adamint:
> Am I understanding correctly that in order to be in a broken state is if I am explicitly setting a status for a resource inside my AppHost code? 

It's a breaking code change; previously, you were able to directly set the `HealthStatus` and `HealthReports` for a resource yourself, either through setting the initial state or publishing updates. Now, you are no longer able to do either. Instead, you must add **health checks** and then the AppHost will compute the health reports and status for you.

> If they wanted to continue to do that, how can they go about it? Also, assuming there is no way to do this in a non-breaking way?

We have made the decision to disallow this behavior, so they can't. Now that the only component that can set the HealthStatus is the AppHost, there is only one source of truth.

joperezr:
Got it. Last question. Do we know how common is it for people to set their own Health status? Given this is a major version, it is technically ok to make this sort of breaking change, but it does feel like the type of change you don't want to take this late after snap as there will be less time (basically none) to react to feedback. I'm wondering if we considered quirking this, i.e. allowing the users to set an app context switch that allows them to continue setting the status and marking the API call as obsolete, and then fully removing it in the next major. This essentially is to reduce friction in adoption from 8 to 9.

davidfowl:
This change needs to go in as it has API changes to both the gRPC proto file and the app model. It also fixes bugs with the current health checks implementation that affect the new dashboard in ACA.

joperezr:
Any comments around whether or not we should quirk this? How concerned are we about breaking existing customers?

davidfowl:
No we're not quirking this. This should go in as is.

joperezr (approved):
Ok. Merging as is, but let's keep an eye out for feedback on this.

PR #6469, opened 2024-10-23
Enables OIDC to be configured in local dev scenarios

los93sol:
This enabled OIDC to be configured in local dev scenarios.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6469)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,7 @@
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <TestcontainersPackageVersion>3.10.0</TestcontainersPackageVersion>
-    <AzureProvisiongVersion>1.0.0-beta.1</AzureProvisiongVersion>
+    <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
--- a/playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
+++ b/playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
@@ -42,7 +42,7 @@ builder.AddProject<Projects.AzureContainerApps_ApiService>("api")
            app.ConfigureCustomDomain(customDomain, certificateName);
 
            // Scale to 0
-           app.Template.Value!.Scale.Value!.MinReplicas = 0;
+           app.Template.Scale.MinReplicas = 0;
        });
 
 #if !SKIP_DASHBOARD_REFERENCE
--- a/playground/bicep/BicepSample.AppHost/redis.module.bicep
+++ b/playground/bicep/BicepSample.AppHost/redis.module.bicep
@@ -15,11 +15,11 @@ resource redis 'Microsoft.Cache/redis@2024-03-01' = {
       capacity: 1
     }
     enableNonSslPort: false
+    disableAccessKeyAuthentication: true
     minimumTlsVersion: '1.2'
     redisConfiguration: {
       'aad-enabled': 'true'
     }
-    disableAccessKeyAuthentication: 'true'
   }
   tags: {
     'aspire-resource-name': 'redis'
--- a/playground/cdk/CdkSample.AppHost/Program.cs
+++ b/playground/cdk/CdkSample.AppHost/Program.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Azure.Provisioning.ApplicationInsights;
-using Azure.Provisioning.Expressions;
 using Azure.Provisioning.KeyVault;
 using Azure.Provisioning.OperationalInsights;
 using Azure.Provisioning.ServiceBus;
@@ -17,7 +16,7 @@ var locationOverride = builder.AddParameter("locationOverride");
 var storage = builder.AddAzureStorage("storage")
     .ConfigureInfrastructure(infrastructure =>
     {
-        var account = infrastructure.GetResources().OfType<StorageAccount>().Single();
+        var account = infrastructure.GetProvisionableResources().OfType<StorageAccount>().Single();
         account.Sku = new StorageSku() { Name = sku.AsProvisioningParameter(infrastructure) };
         account.Location = locationOverride.AsProvisioningParameter(infrastructure);
     });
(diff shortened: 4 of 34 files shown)


los93sol:
@dotnet-policy-service agree [company=""]

los93sol:
@dotnet-policy-service agree

src/Aspire.Hosting/Aspire.Hosting.csproj
```diff
@@ -44,6 +46,7 @@
   <ItemGroup>
     <PackageReference Include="Grpc.AspNetCore" />
     <PackageReference Include="KubernetesClient" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" />
```
davidfowl, line 49: What is this for?

los93sol: The Microsoft.AspNetCore.Authorization.OpenIdConnectOptions that get in AppHpst’s DashbpardOptions.  I couldn’t figure out how to pass the whole object through to the context in the LifetimeHook though so I suppose it may make more sense to drop the dependency and just step supported properties manually.  I only mapped Authority, MetaAddress, ClientId, ClientSecret, and Scope

davidfowl: Can we remove this and just make a poco

los93sol: I'll get it updated

los93sol: Updated

davidfowl:
@JamesNK can you review this?

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
```diff
@@ -11,23 +12,42 @@ internal class DashboardOptions
...
-        options.OtlpGrpcEndpointUrl = configuration["DOTNET_DASHBOARD_OTLP_ENDPOINT_URL"];
-        options.OtlpHttpEndpointUrl = configuration["DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL"];
+        if (Enum.TryParse<FrontendAuthMode>(configuration[DashboardConfigNames.DashboardFrontendAuthModeName.ConfigKey], out var dashboardAuthMode))
+        {
+            options.DashboardAuthMode = dashboardAuthMode;
+        }
+
+        options.DashboardToken = configuration["AppHost:BrowserToken"];
+        configuration.Bind("Dashboard:Frontend:OpenIdConnect", options.OpenIdConnect);
+        configuration.Bind("Authentication:Schemes:OpenIdConnect", options.OpenIdConnectSettings);
```
JamesNK, line 48: I've been saying this for a while docs of hosting configuration is important for a while but now I must insist: It **_needs_** to be documented somewhere. As a reviewer who isn't an expert with hosting code I have no idea what configuration options the host accepts. Is this config accepted somewhere but no one knows about it? Is it following the naming standards of other docs? The current situation is how a big mess is created. Some order needs to be brought to this area.

And, not to meantion, docs need to be added so users know what options they can pass via hosting configuration.

Who is going to own adding it? I'll do it if no one else will...

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -214,15 +216,55 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
-            // Configure frontend browser token
-            if (!string.IsNullOrEmpty(browserToken))
+            // Configure frontend auth
+            context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendAuthModeName.EnvVarName] = dashboardAuthMode.ToString();
+
+            if (dashboardAuthMode == FrontendAuthMode.OpenIdConnect)
             {
-                context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendAuthModeName.EnvVarName] = "BrowserToken";
-                context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendBrowserTokenName.EnvVarName] = browserToken;
+                if (options.OpenIdConnect != null)
```
JamesNK, line 224: Forwarding configuration from the host to the dashboard seems to be a common thing to want to do. Are there any plans to do this in a more structured, organized, reusable way?

src/Aspire.Hosting/PublicAPI.Unshipped.txt
```diff
@@ -1,4 +1,20 @@
 #nullable enable
+Aspire.Dashboard.Configuration.FrontendAuthMode
```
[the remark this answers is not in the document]
los93sol: @JamesNK The OpenIdConnectOptions used to configure the auth policy were similarly impacted so I updated this to revert the moves of FrontendAuthMode and OpenIdConnectOptions and instead just used the same POCO approach as @davidfowl suggested for OpenIdConnectOptions that configure OIDC.  With that adjustment there are no public API changes for this.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
los93sol: @JamesNK I'm not sure if this is exactly what you're looking for, but I did find this [dashboard configuration documentation](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/configuration?tabs=bash#common-configuration) that originally lead to this PR as it left me under the impression this was already a supported scenario.

JamesNK: Dashboard configuration is different from app host configuration. People thinking they're the same is a common point of confusion.

I've started a gist to list possible configuration: https://gist.github.com/JamesNK/a22d95de1580f67abf348b1779c7006d
It will be the source of a future app host configuration doc.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
davidfowl: Feels like we should build something that lets you generically forward settings to the dashboard (like a configuration section) where dashboard config can be specified.

los93sol: I agree this is a friction point.  

In this case the prefix is known and I don't think that Aspire should care what settings you use, if it's configurable for a normal app this way then I'd expect it to just pass it through here as well without needing to manually wire it all up.  Unfortunately, I was not able to come up with a way to achieve that without the POCO that limits the options and creates a model that now needs to be maintained to stay in sync with the actual package.  

I felt the same way about the FrontendAuthMethod enum as well, but that one is a little different because of the thin automation layer to automatically configure BrowserToken by default which seems appropriate for the majority of local dev scenarios.

PR #6472, opened 2024-10-24
Fix feed for daily builds from main branch

joperezr:
cc: @DamianEdwards 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6472)

### DIFF
--- a/docs/using-latest-daily.md
+++ b/docs/using-latest-daily.md
@@ -28,6 +28,7 @@ If you use [Package Source Mapping](https://learn.microsoft.com/en-us/nuget/cons
   <packageSource key="dotnet9">
     <package pattern="Aspire.*" />
     <package pattern="Microsoft.Extensions.ServiceDiscovery*" />
+    <package pattern="Microsoft.Extensions.Http.Resilience" />
   </packageSource>
 </packageSourceMapping>
 ```
--- a/docs/using-latest-daily.md
+++ b/docs/using-latest-daily.md
@@ -18,14 +18,14 @@ dotnet new nugetconfig
 
 The latest builds are pushed to a special feed, which you need to add:
 ```sh
-dotnet nuget add source --name dotnet8 https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json
+dotnet nuget add source --name dotnet9 https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json
 ```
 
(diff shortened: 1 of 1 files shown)


docs/using-latest-daily.md
```diff
@@ -28,6 +28,7 @@ If you use [Package Source Mapping](https://learn.microsoft.com/en-us/nuget/cons
   <packageSource key="dotnet9">
     <package pattern="Aspire.*" />
     <package pattern="Microsoft.Extensions.ServiceDiscovery*" />
+    <package pattern="Microsoft.Extensions.Http.Resilience" />
```
DamianEdwards, line 31: Are we sure all the uses of this package in the repo are using a version that's from the `dotnet9` feed? If some are using 8.0.10 still we'll need to include nuget.org and have a mapping for that too I think.

DamianEdwards: In updating dotnet/aspire-samples to use dailies I'm finding I need a mapping for `Microsoft.Extensions.Resilience` too. Maybe we just make this `Microsoft.Extensions.*`?

joperezr: 8.0.10 is out on nuget.org. Perhaps we should instead not try to guess and remove the package source mapping from our suggested nuget.config, as the actual one that the users end up using will be different depending on the things that their app is using.

joperezr: The way our instructions are now, we are only providing the mappings for the packages produced in this repo, but that is not a guarantee that you won't need more depending on your project.

DamianEdwards: I think we should make it as clear as possible to help folks set their project up for the highest changes of success.

joperezr: The problem with adding `Microsoft.Extensions.*`  is that there may be packages that users will want to get from elsewhere. For instance, they may want to use stable bits from NuGet.org for all of .NET OOBs, except for the Aspire packages, so I wonder if we should try to instead have the least inclusive list and then the user can expand if needed. If Microsoft.Extensions.Resilience is needed, then we can add that one too.

davidfowl: What's the blocker with this PR?

joperezr: I think it is just to make sure that the feeds are complete, but I just haven't had the chance to come back to this as it's been a bit hectic the last few days. If we want to merge as is, we could and I can send a follow-up later to address feedback, or otherwise I can take a look at this in a few days.

davidfowl: yea lets its broken right now

joperezr: Sounds good, merging.

PR #6481, opened 2024-10-24
Some ContainerExec experimentation.

mitchdenny:
Something to look at post 9.0.

This is just a draft PR for some discussion. In the DCP types we have ContainerExec. It allows us to invoke a command against an existing container. I'm thinking it might enable us to do things like run backup/restore commands inside container images.

This would allow for some interesting advanced scenarios such as triggering a backup/restore of a postgres database using something like pg_dump within the container (the file would be copied to a databindmount.

@danegsta @karolz-ms @DamianEdwards @davidfowl @drewnoakes @JamesNK @adamint 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6481)

### DIFF
--- a/playground/waitfor/WaitForSandbox.AppHost/Program.cs
+++ b/playground/waitfor/WaitForSandbox.AppHost/Program.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.Extensions.DependencyInjection;
+
 var builder = DistributedApplication.CreateBuilder(args);
 
 var db = builder.AddAzurePostgresFlexibleServer("pg")
@@ -8,6 +10,7 @@ var db = builder.AddAzurePostgresFlexibleServer("pg")
                 .RunAsContainer(c =>
                 {
                     c.WithPgAdmin();
+                    c.WithPgDumpCommands();
                 })
                 .AddDatabase("db");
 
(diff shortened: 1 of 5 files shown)


mitchdenny:
I should note that PgAdmin allows you to backup and restore databases but the idea here is that you could do things like automate the restore of a database from a known baseline.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
```diff
@@ -35,3 +38,32 @@
...
+    public static IResourceBuilder<PostgresServerResource> WithPgDumpCommands(this IResourceBuilder<PostgresServerResource> builder)
+    {
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeStartEvent>((e, ct) =>
+        {
+            var executor = e.Services.GetRequiredService<ContainerCommandExecutor>();
+
+            var databases = e.Model.Resources.OfType<IResourceWithParent<PostgresServerResource>>().Where(r => r.Parent == builder.Resource).ToList();
+            foreach (var db in databases)
+            {
+                var dbBuilder = builder.ApplicationBuilder.CreateResourceBuilder(db);
```
davidfowl, line 53: This super late model of mutating the resource builder is pretty broken... you know how I feel about this.

mitchdenny: This is BeforeStart. Not late at all.

PR #6491, opened 2024-10-25
Enable devcontainers in repo.

mitchdenny:
## Description

This is an initial PR to enable devcontainer support in the repo. The goal is to enable the various remote development scenarios (Codespaces, SSH, etc) so that we can test them. This is important because we've added logic to the apphost that can do things like detect it is running in Codespaces and using the Aspire repo under Codespaces is a good way of casually validating everything is working.

image

## Outstanding issues

- [ ] Dashboard URL shows as localhost:port/login?t=<token>; whilst Codespaces will automatically remap that URL for us when we click on it, the redirect flow (GitHub auth) looses the token in the URL

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
    - Link to aspire-docs issue: 
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6491)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -83,7 +83,7 @@
 			}
 		}
 	},
-	"postCreateCommand": "dotnet restore",
+	"onCreateCommand": "dotnet restore",
 	"postStartCommand": "dotnet dev-certs https --trust"
 
 	// Configure tool-specific properties.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,12 @@
 	"features": {
 		"ghcr.io/devcontainers/features/azure-cli:1": {},
 		"ghcr.io/azure/azure-dev/azd:0": {},
-		"ghcr.io/devcontainers/features/docker-in-docker": {}
+		"ghcr.io/devcontainers/features/docker-in-docker": {},
+		"ghcr.io/devcontainers/features/dotnet": {
+			"additionalVersions": [
+				"8.0.403"
+			]
+		}
 	},
 
 	"containerEnv": {
@@ -18,12 +23,25 @@
 	// "features": {},
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
-	// "forwardPorts": [5000, 5001],
-	// "portsAttributes": {
-	//		"5001": {
-	//			"protocol": "https"
-	//		}
-	// }
+	"forwardPorts": [
+		15887, // playground/waitfor/WaitForSandbox.AppHost
+		5180, // playground/waitfor/WaitForSandbox.ApiService
+		7024 // playground/waitfor/WaitFor.Frontend
+	],
+	"portsAttributes": {
+		"15887": {
+			"label": "WaitForSandbox.AppHost",
+			"protocol": "https"
+		},
+		"5180": {
+			"label": "WaitForSandbox.ApiService",
+			"protocol": "http"
+		},
+		"7024": {
+			"label": "WaitFor.Frontend",
+			"protocol": "https"
+		}
+	},
 
 	// Use 'postCreateCommand' to run commands after the container is created.
 	"postCreateCommand": "rm -rf .dotnet && sudo chmod +x dotnet.sh && ./dotnet.sh restore",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -26,27 +26,43 @@
 
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	"forwardPorts": [
-		15887, // playground/waitfor AppHost
-		5180, // playground/waitfor ApiService
-		7024, // playground/waitfor Frontend
-		15551 // playground/waitfor PGAdmin
+		15887,
+		5180,
+		7024,
+		15551,
+		33803,
+		5350,
+		41567,
+		15306
 	],
 	"portsAttributes": {
-		"15887": {
-			"label": "WaitFor Playground: AppHost",
-			"protocol": "https"
-		},
 		"5180": {
 			"label": "WaitFor Playground: ApiService",
 			"protocol": "http"
+		},
+		"5350": {
+			"label": "Redis Playground: Api Service"
 		},
 		"7024": {
 			"label": "WaitFor Playground: Frontend",
 			"protocol": "https"
 		},
+		"15306": {
+			"label": "Redis Playground: App Host"
+		},
 		"15551": {
 			"label": "WaitFor Playground: PGAdmin",
 			"protocol": "http"
+		},
+		"15887": {
+			"label": "WaitFor Playground: AppHost",
+			"protocol": "https"
+		},
+		"33803": {
+			"label": "Redis Playground: Redis Commander"
+		},
+		"41567": {
+			"label": "Redis Playground: Redis Insight"
 		}
 	},
 	"otherPortsAttributes": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -56,10 +56,12 @@
 				"ms-azuretools.azure-dev"
 			],
 			"settings": {
-				"remote.autoForwardPorts": false
+				"remote.autoForwardPorts": false,
+				"dotnet.defaultSolution": "Aspire.sln"
 			}
 		}
 	},
+	"postCreateCommand": "dotnet restore",
 	"postStartCommand": "dotnet dev-certs https --trust"
 
 	// Configure tool-specific properties.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,43 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+	"name": "C# (.NET)",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:1-9.0",
+	"features": {
+		"ghcr.io/devcontainers/features/azure-cli:1": {},
+		"ghcr.io/azure/azure-dev/azd:0": {},
+		"ghcr.io/devcontainers/features/docker-in-docker": {}
+	},
+
+	"containerEnv": {
+		"DOTNET_ROOT": "/workspaces/aspire/.dotnet"
+	},
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [5000, 5001],
+	// "portsAttributes": {
+	//		"5001": {
+	//			"protocol": "https"
+	//		}
+	// }
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "rm -rf .dotnet && sudo chmod +x dotnet.sh && ./dotnet.sh restore",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-dotnettools.csdevkit"
+			]
+		}
+	}
+
+	// Configure tool-specific properties.
+	// "customizations": {},
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
+}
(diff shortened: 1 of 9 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+	"name": "C# (.NET)",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:1-9.0",
+	"features": {
+		"ghcr.io/devcontainers/features/azure-cli:1": {},
+		"ghcr.io/azure/azure-dev/azd:0": {},
+		"ghcr.io/devcontainers/features/docker-in-docker": {},
+		"ghcr.io/devcontainers/features/dotnet": {
```
mitchdenny, line 11: Current default .NET download is .NET 9 RC2 build (close to GA). But our builds also need .NET 8.0 installed. For non containerized dev we can often assume that this is installed bu we need to explicitly install it here.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+	"name": "C# (.NET)",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:1-9.0",
+	"features": {
+		"ghcr.io/devcontainers/features/azure-cli:1": {},
+		"ghcr.io/azure/azure-dev/azd:0": {},
+		"ghcr.io/devcontainers/features/docker-in-docker": {},
```
mitchdenny, line 10: We add this feature so that we can do docker inside the container.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+	"name": "C# (.NET)",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/dotnet:1-9.0",
+	"features": {
+		"ghcr.io/devcontainers/features/azure-cli:1": {},
+		"ghcr.io/azure/azure-dev/azd:0": {},
```
mitchdenny, line 9: Useful for testing deployments.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
+		15887, // playground/waitfor AppHost
+		5180, // playground/waitfor ApiService
+		7024, // playground/waitfor Frontend
+		15551 // playground/waitfor PGAdmin
```
mitchdenny, line 26: We can let devcontainers automatically map ports, however in the case of .NET Aspire we open so many ports that it can be overwhelming. Current approach is to explicitly map ports here. We probably should go through and make sure we have no overlapping ports which will probably help in other areas too around cookies etc.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+			"label": "WaitFor Playground: Frontend",
+			"protocol": "https"
+		},
+		"15551": {
+			"label": "WaitFor Playground: PGAdmin",
+			"protocol": "http"
+		}
+	},
+	"otherPortsAttributes": {
+		"onAutoForward": "ignore"
```
mitchdenny, line 47: This is supposed to stop autoforwarding although it doesn't seem to be reliable.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-dotnettools.csdevkit",
+				"ms-azuretools.vscode-bicep",
+				"ms-azuretools.azure-dev"
+			],
+			"settings": {
+				"remote.autoForwardPorts": false,
```
mitchdenny, line 59: This DOES seem to be a reliable way of disabling auto port forwarding.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+				"dotnet.defaultSolution": "Aspire.sln"
```
mitchdenny, line 60: This is to stop one of the nags popping up.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}
+	},
+	"postCreateCommand": "dotnet restore",
```
mitchdenny, line 64: Because we control the versions of .NET installed via the `.devcontainer.json` file we can just do a `dotnet restore` here.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,72 @@
...
+	"postStartCommand": "dotnet dev-certs https --trust"
```
mitchdenny, line 65: Trust certs automatically so we don't have to worry about it the first time we debug.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
```diff
@@ -7,7 +7,15 @@
                 .WithPasswordAuthentication()
                 .RunAsContainer(c =>
                 {
-                    c.WithPgAdmin();
+                    c.WithPgAdmin(c =>
+                    {
+                        c.WithHostPort(15551);
```
mitchdenny, line 12: Pinning the PGAdmin port so that it is stable for forwarding.

playground/waitfor/WaitForSandbox.AppHost/Program.cs
```diff
@@ -7,7 +7,15 @@
...
+
+                        // Variables that we are going to need to set when we detect that
+                        // we are run run mode with codespaces.
+                        c.WithEnvironment("PGADMIN_CONFIG_PROXY_X_HOST_COUNT", "1");
+                        c.WithEnvironment("PGADMIN_CONFIG_PROXY_X_PREFIX_COUNT", "1");
```
mitchdenny, line 17: To be discussed. We may want to provide a mechanism for extensions to query whether they are working in codespaces and if so inject some additional configuration values. This is required for PGAdmin. This applies to Codespaces only - for local container dev and probably remote container dev (over SSH) this is not an issue.

.devcontainer/devcontainer.json
davidfowl: What version of azd is this?

mitchdenny: Don't know but we might be able to override it.

mitchdenny: @ellismg @vhvb1989 we are looking at enabling dev containers / codespaces in the dotnet/aspire repo. We are preinstalling azd and it looks like I can pin to a specific version as long as it has been uploaded to the GitHub releases page for azd, but is there a way to make the container image install a version from a PR too?

mitchdenny: @davidfowl to answer your earlier question this is what I found in my spelunking:

https://github.com/Azure/azure-dev/blob/b15ccaf35f60e30095740123f6beaa14c3c0f0fb/ext/devcontainer/src/azd/README.md

vhvb1989: Using the azd feature to pre-install will give yo the latest version at the momento of building the codespace.
The feature does not support installing versions from PRs, but you can just run the azd-install script from the codespace to change whatever version.
Note: if you ever want to use a [prebuilt codespace](https://docs.github.com/en/codespaces/prebuilding-your-codespaces) you might want to change from using the azd-feature to pre-install azd for using a postCreate action in your devcontainer to run azd's install script instead.  I mention this b/c prebuilt codespaces are getting more and more popular (for how fast you load your workspace) but keeping dependencies from features up to date becomes a little messy.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,78 @@
...
+		"ghcr.io/azure/azure-dev/azd:0": {},
+		"ghcr.io/devcontainers/features/docker-in-docker": {},
+		"ghcr.io/devcontainers/features/dotnet": {
+			"additionalVersions": [
+				"8.0.403"
+			]
+		}
+	},
+
+	"hostRequirements": {
```
mitchdenny, line 18: In the interests of performance/usability setting a reasonably high minimum for memory.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -333,6 +335,16 @@ private static void SetPgAdminEnvironmentVariables(EnvironmentCallbackContext co
...
+
+        // When running in the context of Codespaces we need to set some additional environment
+        // varialbes so that PGAdmin will trust the forwarded headers that Codespaces port
+        // forwarding will send.
+        var codespaceOptions = context.ExecutionContext.ServiceProvider.GetRequiredService<IOptions<CodespacesOptions>>();
+        if (context.ExecutionContext.IsRunMode && codespaceOptions.Value.IsCodespace)
+        {
+            context.EnvironmentVariables["PGADMIN_CONFIG_PROXY_X_HOST_COUNT"] = "1";
+            context.EnvironmentVariables["PGADMIN_CONFIG_PROXY_X_PREFIX_COUNT"] = "1";
+        }
```
davidfowl, line 347: Can we invert the dependency so that integrations don't need to worry about code spaces?

mitchdenny: The code that figures out what extra environment variables need to be injected into which resource needs to go somewhere, given the field of integrations is unbounded, and the specific environment variables tied to the integration it makes sense to do it this way I think.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,94 @@
...
+		"cpus": 8,
+		"memory": "32gb",
+		"storage": "64gb"
+	},
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
```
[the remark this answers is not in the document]
davidfowl: I was thinking we should generate these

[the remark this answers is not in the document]
mitchdenny: We can make port forwarding complete automatic. Basically, it just watches the processes and the ports they have open and forwards them. However, I don't like that for a few reasons:

1. Sometimes you don't want something forwarded automatically (security concerns)
2. It does a really bad job of protocol detection (normally detects HTTPS as HTTP which results in a 502 from Codespaces port forwarding).

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
mitchdenny: I think that is a possibility, but the characteristics of each proxy is probably going to have an impact on these variables anyway. So you might end up with a compound condition anyway, at which point it hasn't helped you.

I think that would be premature right now.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
mitchdenny: @davidfowl a thought. What if we, when we detected codespaces or a devcontainer file automatically updated the file based on what is in the apphost? That way people wouldn't really need to think about it, they `.devcontainer` would just organically get configured?

davidfowl (approved):

davidfowl:
@mitchdenny one place where we don't detect the codespace and print the right url is here:

image

We should change this too.

mitchdenny:
Yeah I was looking at this I made some tweaks last night as an experiment. I can put up a PR.

davidfowl:
Also it seems like only certain playground examples work?

PR #6529, opened 2024-10-28
Enable local dev OIDC

los93sol:
This is a continuation of #6469 I needed to move the fork over to a branch to work on some other things and inadvertently closed out the original PR
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6529)


src/Aspire.Hosting/Dashboard/DashboardOptions.cs
```diff
@@ -11,23 +11,69 @@ internal class DashboardOptions
...
-        options.OtlpGrpcEndpointUrl = configuration["DOTNET_DASHBOARD_OTLP_ENDPOINT_URL"];
-        options.OtlpHttpEndpointUrl = configuration["DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL"];
+        if (Enum.TryParse<DashboardAuthMode>(configuration[DashboardConfigNames.DashboardFrontendAuthModeName.ConfigKey], out var dashboardAuthMode))
+        {
+            options.DashboardAuthMode = dashboardAuthMode;
+        }
+
+        options.DashboardToken = configuration["AppHost:BrowserToken"];
+        configuration.Bind("Dashboard:Frontend:OpenIdConnect", options.OpenIdConnect);
+        configuration.Bind("Authentication:Schemes:OpenIdConnect", options.OpenIdConnectSettings);
```
halter73, line 74: I agree with @JamesNK comment on the earlier PR that these two related sections seem a bit too disconnected. Just because this is the section normal web apps use to configure OpenIdConnect doesn't mean that we have to do that for the dashboard, right? Could we include everything under "Dashboard:Frontend:OpenIdConnect"? We could forward it to "Authentication:Schemes:OpenIdConnect" in the orchestrated apps.

Another option might be to add the `NameClaimType`/`UsernameClaimType`/`RequiredClaimType`/`RequiredClaimValue` logic from aspire to ASP.NET Core's OIDC config binding logic and include everything under "Authentication:Schemes:OpenIdConnect" except for the `AuthMode=OpenIdConnect` part.

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
```diff
@@ -214,15 +215,55 @@ private void ConfigureAspireDashboardResource(IResource dashboardResource)
...
+                {
+                    context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendOpenIdConnectNameClaimType.EnvVarName] = options.OpenIdConnect.NameClaimType;
+                    context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendOpenIdConnectUsernameClaimType.EnvVarName] = options.OpenIdConnect.UsernameClaimType;
+                    context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendOpenIdConnectRequiredClaimType.EnvVarName] = options.OpenIdConnect.RequiredClaimType;
+                    context.EnvironmentVariables[DashboardConfigNames.DashboardFrontendOpenIdConnectRequiredClaimValue.EnvVarName] = options.OpenIdConnect.RequiredClaimValue;
+                }
+
+                if (options.OpenIdConnectSettings != null)
+                {
+                    var prefix = "AUTHENTICATION__SCHEMES__OPENIDCONNECT__";
```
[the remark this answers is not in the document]
los93sol: Originally it was, but [earlier feedback](https://github.com/dotnet/aspire/pull/6469#discussion_r1813962144) earlier feedback was to switch to POCO to avoid the dependency on it.  If there was a way to copy that entire section in then I guess you don't need the reference and could just copy that section in directly, but since there's currently no way to bring the whole section in I think POCO is a reasonable approach for now that doesn't really bind you in any way long term.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
los93sol: Is it really just the configuration keys you think need to be changed over?

Currently the appsettings config would look something like this...

```
"Dashboard": {
  "Frontend": {
    "AuthMode": "OpenIdConnect",
    "OpenIdConnect": {
      "RequiredClaimType": "",
      "RequiredClaimValue": ""
    }
  }
},
"Authentication": {
  "Schemes": {
    "OpenIdConnect": {
      "Scope": [ "openid", "profile" ],
      "MetadataAddress": "",
      "ClientId": "",
      "ClientSecret": ""
    }
  }
}
```

I'm guessing you'd rather see something like....

```
"Dashboard": {
  "AuthMode": "OpenIdConnect",
  "OpenIdConnect": {
    "Policy": {
      "RequiredClaimType": "",
      "RequiredClaimValue": ""
    },
    "Settings": {
      "Scope": [ "openid", "profile" ],
      "MetadataAddress": "",
      "ClientId": "",
      "ClientSecret": ""
    }
  }
}
```

los93sol: @halter73 Let me know if I’m interpreting this correctly and I will make the changes

los93sol:
@danmoseley You mentioned you would welcome more PR’s and I’m ready to deliver.  What do I need to do to get some traction on this one?

danmoseley:
Looks like we need guidance from @halter73

los93sol:
@halter73 Please clarify what changes need to happen for this to be merged.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
halter73: Sorry for the delay. I just got back from vacation. I don't have much background on this PR, I only dropped a review because @eerhardt requested it. I have a deeper understanding of the inner workings of the OpenIdConnectHandler than the Aspire dashboard.

Is the intention of this PR to only enable/configure OIDC for the dashboard? Or can this configuration also be used by the projects orchestrated by aspire?

If it's the former, I would like to see something like your second example. It makes it much more clear the configuration only applies to the dashboard. The "AuthMode" might not even be necessary since the presence of the "Dashboard:OpenIdConnect" configuration section ought to be enough to indicate that you want to use it.

However, if the idea is to be able to share OIDC configuration between the dashboard and other projects, it does make more sense to have it split out the way you currently have it. But then, I don't see what things like `NameClaimType` and `UsernameClaimType` would be specific to the dashboard. Wouldn't you want to share that between all projects?

I can see why `RequiredClaimType` and `RequiredClaimValue` might be specific to the dashboard, but I'm not sure that would belong in an "OpenIdConnect" subsection. What if we added support for another authentication mode later that also provides claims? Would it be more future proof to put `RequiredClaimType` and `RequiredClaimValue` in a "Dashboard:Frontend:AuthorizationPolicy" section to avoid duplication?

src/Aspire.Hosting/Dashboard/DashboardLifecycleHook.cs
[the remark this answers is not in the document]
halter73: @davidfowl Can you explain your earlier feedback? I don't like that it's using the same default configuration section `OpenIdConnectConfigureOptions` does but only supports a subset of the options supported by that. If the goal is just to remove Microsoft.AspNetCore.Authentication.OpenIdConnect PackageReference, I would just forward the entire "Authentication:Schemes:OpenIdConnect" `IConfigurationSection` via environment variables.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
los93sol: This is just for the dashboard, that’s why I thought it made sense to put it in the normal location you’d use for any app instead of encapsulating it.  If there was a way to forward the entire section to the context then I think we can avoid taking a dependency on the OpenIdConnect package and it should support every option in whatever version of the package the user is consuming.  I tried to sort that out but came up empty.  It seemed like it would be there already since you can take config and bind it, but I don’t see a clear path to just copying the config section since the config builder in the context doesn’t even exist at that point

halter73: > If there was a way to forward the entire section to the context then I think we can avoid taking a dependency on the OpenIdConnect package and it should support every option in whatever version of the package the user is consuming.

It would be nice if there was a one-liner like `Bind` that supported `Dictonary<string, object>` and created nested dictionaries when necessary, but that's not currently supported. Plus, that's not exactly what we need here since it wouldn't create `UNDERSCORE__DELIMITED__SECTION__NAMES` for the keys.

It doesn't seem like it would be too hard to write custom code that calls `IConfguration.GetChildren()` and builds up the dictionary with the right keys though.

los93sol: Yes!  That’s exactly what I was trying originally but the only way I could think of to do it was feed a root like authentication__ and go grab all the env variables that start with that and bring them over directly.  Doesn’t seem like a very good solution, but it might be doable

danmoseley:
@halter73 could you answer @los93sol question above?

danmoseley:
@los93sol could you merge main in?

@halter73 could you help with the question?

davidfowl:
@los93sol sorry we let this PR go super stale. Do you want to send another PR with this implementation? Biggest thing missng from this beyond the implementation was testing (and understanding the scenarios).

I'm going to close this for now. Feeel free to reopen with that information.

PR #6537, opened 2024-10-28
Use rc.2 branding on EF Npgsql integration

### LINKED ISSUE #6492: Do a pass over packages that shipped as .rc1 to ensure they aren't shipping as -preview
We need to do a check on all packages that are not going stable in November and ensure that, if they shipped as "rc" branded in October, they are not getting a "downgrade" version to preview. This will need to happen at the very least for the NPGsql ef core integration package.

joperezr:
## Description

Having EF NpgSql integration produce RC.2 branded prerelease in order to not downgrade from rc1 version during GA release.

cc: @eerhardt 

Fixes #6492 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6537)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,6 +6,9 @@
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <PreReleaseVersionLabel>preview.5</PreReleaseVersionLabel>
+    <!-- Aspire.Npgsql.EntityFrameworkCore.PostgreSQL will ship right after our November release, but it has already shipped as rc.1 so
+    we can't 'downgrade' it to preview.5. We will ship it as rc.2 in November and will mark it stable once EF Npgsql dependency ships stable. -->
+    <PrereleaseVersionLabel Condition="'$(MSBuildProjectName)' == 'Aspire.Npgsql.EntityFrameworkCore.PostgreSQL'">rc.2</PrereleaseVersionLabel>
     <!--
       When running package validation as part of the build, we want to ensure we didn't break the API against the previous
       version of each package. The following property points to the package version that should be used as baseline.


eng/Versions.props
```diff
@@ -6,6 +6,9 @@
     <PatchVersion>0</PatchVersion>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <PreReleaseVersionLabel>preview.5</PreReleaseVersionLabel>
+    <!-- Aspire.Npgsql.EntityFrameworkCore.PostgreSQL will ship right after our November release, but it has already shipped as rc.1 so
+    we can't 'downgrade' it to preview.5. We will ship it as rc.2 in November and will mark it stable once EF Npgsql dependency ships stable. -->
+    <PrereleaseVersionLabel Condition="'$(MSBuildProjectName)' == 'Aspire.Npgsql.EntityFrameworkCore.PostgreSQL'">rc.2</PrereleaseVersionLabel>
```
eerhardt, line 11: Ideally, this would be in the `Aspire.Npgsql.EntityFrameworkCore.PostgreSQL.csproj`.

eerhardt (approved):

eng/Versions.props
joperezr: Agreed. I thought about it too but in this case I wanted to reduce the risk of properties getting overridden at the wrong time so to de-risk I opted for setting this at the exact same time as it would've been set before. Given this is a one-time change that  will only ever be done in this branch, I think doing this is ok. That said, let me know if you have a strong preference on me moving things around.

eerhardt: This is fine for now. But we should really fix our build logic to be "right" so individual libraries can choose their versions independently.

joperezr:
Merging to get the release ready given the ongoing issues with CI builds.

PR #6564, opened 2024-10-30
[ci] playground tests - show live output from the tests

radical:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6564)

### DIFF
--- a/tests/Aspire.EndToEnd.Tests/.runsettings
+++ b/tests/Aspire.EndToEnd.Tests/.runsettings
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RunSettings>
-  <xunit>
+  <xUnit>
     <ShowLiveOutput>true</ShowLiveOutput>
-  </xunit>
+  </xUnit>
   <RunConfiguration>
     <!-- Timeout in ms, 15 minutes -->
     <TestSessionTimeout>900000</TestSessionTimeout>
--- a/tests/Aspire.EndToEnd.Tests/.runsettings
+++ b/tests/Aspire.EndToEnd.Tests/.runsettings
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RunSettings>
+  <xunit>
+    <ShowLiveOutput>true</ShowLiveOutput>
+  </xunit>
   <RunConfiguration>
     <!-- Timeout in ms, 15 minutes -->
     <TestSessionTimeout>900000</TestSessionTimeout>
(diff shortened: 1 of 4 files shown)


danmoseley (approved):

DamianEdwards:
@radical should we update the xunit runner settings to [enable live output too](https://xunit.net/docs/runsettings#ShowLiveOutput)?

radical:
> @radical should we update the xunit runner settings to [enable live output too](https://xunit.net/docs/runsettings#ShowLiveOutput)?

Is this sufficient to show the output in VS? If so, this sounds like a great idea.

DamianEdwards:
> Is this sufficient to show the output in VS? If so, this sounds like a great idea.

Unfortunately I don't think it works in VS @bradwilson

bradwilson:
Also, if you want to see live output with `dotnet test`, you need to pass `--logger "console;verbosity=normal"`:

https://github.com/user-attachments/assets/c7af5cc7-c899-400e-827b-ec4aba3ee86a

tests/helix/send-to-helix-buildonhelixtests.targets
```diff
@@ -35,7 +35,7 @@
       <_TestBlameArguments Include="--blame-crash-dump-type full" />
 
       <!-- Using `dotnet test` for the project directly here -->
-      <_TestRunCommandArguments Include="dotnet test -s .runsettings --results-directory $(_HelixLogsPath)" />
+      <_TestRunCommandArguments Include="dotnet test -s .runsettings --results-directory $(_HelixLogsPath) -v:n" />
```
eerhardt, line 38: Do we need this when the .runsettings also has:

```
      <Logger friendlyName="console">
        <Configuration>
          <Verbosity>normal</Verbosity>
        </Configuration>
      </Logger>
```

?

bradwilson: Passing `-v:n` to `dotnet test` changes the verbosity of the MSBuild operations that `dotnet test` does, which is different than the verbosity of the console logger.

tests/Aspire.EndToEnd.Tests/.runsettings
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RunSettings>
+  <xunit>
```
bradwilson, line 3: Is this actually working for you? My understanding was that these are case sensitive (because thanks XML), and these _should_ be `xUnit` (with the capital U). https://xunit.net/docs/runsettings

bradwilson: image

tests/helix/send-to-helix-buildonhelixtests.targets
radical: We need `-v:n` so the msbuild logger shows the live output.

tests/Aspire.EndToEnd.Tests/.runsettings
radical: Ah, thank you. I will fix that.

PR #6572, opened 2024-10-31
Change URL detection regex to exclude trailing content

JamesNK:
## Description

URL detection regex was including trailing content in the URL after semi-colon. In some connection strings this would pick up extra values, e.g. content after the URL would be included in `Endpoint=http://myaccount.example.com/;key=abc`.

The new URL is based on GitHub's URL detection for source code. It excludes content after `;` from the URL. This is the same behavior as VS. Technically `;` can be part of a URL, but more commonly it is used as a delimiter for connection strings and collections of URLs.

There isn't a perfect answer here, but the new regex match seems the better alternative. We should be conservative when assuming what content should be in the URL rather than aggressively including too much.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No **TBD**
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6572)

### DIFF
--- a/playground/Stress/Stress.ApiService/ConsoleStresser.cs
+++ b/playground/Stress/Stress.ApiService/ConsoleStresser.cs
@@ -82,6 +82,7 @@ public static class ConsoleStresser
         Console.WriteLine("https://www.example.com/path/?query;string");
         Console.WriteLine("https://;www.example.com/");
         Console.WriteLine("https://www;.example.com/");
+        Console.WriteLine("https://www.exa;mple.com/");
 
         Console.Write("\x1b[0m"); // reset color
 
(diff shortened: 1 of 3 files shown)


mitchdenny (approved):
LGTM

Issue #2395, opened 2024-02-24
Support pre-provisioned configuration for azure resources


davidfowl:
We removed the ability to specify a connection string directly for arbitrary resources now that we have the `AddConnectionString` method. This adds a bit of friction when you want to directly specify the information for pre-provisioned to be used in development, while still allowing them to be provisioned during deployment.

davidfowl:
Moving this to preview 5. The preview 4 mode of operation will be something like this:

```C#
var keyVault = builder.ExecutionContext.IsPublishMode ? builder.AddKeyVault("kv") : builder.AddConnectionString("kv");
```

We'll add the ability to use an existing resource in preview 5

mitchdenny:
For GA we'll do use AddConnectionString(...).

captainsafia:
I'm taking a look at this in reaction to the feedback we received from the Aspire + Azure Functions work.

captainsafia:
Closing this out for 9.1 as a bulk of the work has been done here.

Docs issue for this is over at https://github.com/dotnet/docs-aspire/issues/2549.

We can consider additional enchancements like https://github.com/dotnet/aspire/issues/7514 to this in the future.

PR #6577, opened 2024-10-31
Configure OpenAI models in the app host

sebastienros:
Building over [@SteveSandersonMS PR](https://github.com/dotnet/aspire/pull/6225) to demonstrate the end-to-end experience when ai models/deployments are configured at the host level only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6577)


src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIExtensions.cs
```diff
@@ -103,14 +106,74 @@ public static IResourceBuilder<AzureOpenAIResource> AddAzureOpenAI(this IDistrib
...
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<AzureOpenAIResource> AddDeployment(this IResourceBuilder<AzureOpenAIResource> builder, AzureOpenAIDeployment deployment)
+    public static IResourceBuilder<IHasDeploymentResource> WithDeployment(this IResourceBuilder<IHasDeploymentResource> builder, AzureOpenAIDeployment deployment)
     {
         builder.Resource.AddDeployment(deployment);
+
         return builder;
     }
+
+    /// <summary>
```
sebastienros, line 140: Copy-paste, ignore

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIConnectionResource.cs
```diff
@@ -0,0 +1,36 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents an Azure OpenAI resource.
+/// </summary>
+public class AzureOpenAIConnectionResource(ParameterResource innerResource, string? environmentVariableName) : IHasDeploymentResource
```
sebastienros, line 8: Resource to represent an open ai resource as a connection string only.

src/Aspire.Hosting.Azure.CognitiveServices/PublicAPI.Shipped.txt
```diff
@@ -14,4 +14,3 @@ Aspire.Hosting.ApplicationModel.AzureOpenAIResource.Deployments.get -> System.Co
...
 static Aspire.Hosting.AzureOpenAIExtensions.AddAzureOpenAI(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.AzureOpenAIResource!>!
 static Aspire.Hosting.AzureOpenAIExtensions.AddAzureOpenAI(this Aspire.Hosting.IDistributedApplicationBuilder! builder, string! name, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.AzureOpenAIResource!>!, Aspire.Hosting.ResourceModuleConstruct!, Azure.Provisioning.CognitiveServices.CognitiveServicesAccount!, System.Collections.Generic.IEnumerable<Azure.Provisioning.CognitiveServices.CognitiveServicesAccountDeployment!>!>? configureResource) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.AzureOpenAIResource!>!
-static Aspire.Hosting.AzureOpenAIExtensions.AddDeployment(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.AzureOpenAIResource!>! builder, Aspire.Hosting.ApplicationModel.AzureOpenAIDeployment! deployment) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.AzureOpenAIResource!>!
```
eerhardt, line 17: Are we able to do this without making breaking changes? For example, can we obsolete something and add new APIs?

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIConnectionResource.cs
eerhardt: Why do we need this? Why can't we just use the AzureOpenAIResource?

sebastienros: I wanted a way to configure an existing resource without provisioning. Similar to https://github.com/dotnet/aspire/issues/2395.

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/Home.razor
```diff
@@ -1,39 +1,36 @@
...
         }
 
         <button @onclick="GenerateNextParagraph" autofocus>Generate</button>
     </div>
 
 @code {
-    private ChatClient? chatClient = null;
-    private List<ChatMessage> chatMessages = new List<ChatMessage>();
+    [Inject(Key = "modelA1")]
+    public required IChatClient chatClient { get; set; }
```
SteveSandersonMS, line 18: I think in this page we should only use `ChatClient` and not `IChatClient`, because it's meant to demonstrate usage of the concrete OpenAI client types directly rather than going through MEAI abstractions. It's the other page, `UseIChatClient`, that's meant to show usage of MEAI abstractions.

src/Aspire.Hosting.Azure.CognitiveServices/IHasDeploymentsResource.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents an Azure OpenAI resource.
+/// </summary>
+public interface IHasDeploymentResource: IResourceWithConnectionString, IResourceWithEnvironment
```
SteveSandersonMS, line 8: This is a surprisingly general type name - `Aspire.Hosting.ApplicationModel.IHasDeploymentResource` doesn't sound specific to Azure OpenAI but from the implementation it is. Could this be named something more specific to AzureOpenAI?

src/Components/Aspire.Azure.AI.OpenAI/AzureOpenAISettings.cs
```diff
@@ -70,6 +70,11 @@ public bool DisableTracing
         set { _disableTracing = value; }
     }
 
+    /// <summary>
+    /// Gets the models to use for the Azure OpenAI resource.
```
SteveSandersonMS, line 74: Can the comment explain what the keys and values represent? It's not clear from the property signature alone.

SteveSandersonMS:
Overall looks great. @eerhardt's point about the breaking change is probably the main thing to consider. Besides that it looks pretty close to being ready.

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIConnectionResource.cs
sebastienros: Removed, out of scope

src/Aspire.Hosting.Azure.CognitiveServices/PublicAPI.Shipped.txt
sebastienros: I reverted as I didn't remember why the change was done. Do we want to keep `AddDeployment` or replaced it with `WithDeployment`?

playground/OpenAIEndToEnd/OpenAIEndToEnd.WebStory/Components/Pages/Home.razor
sebastienros: Reverted the changes. Now testing with the `/useichatclient` page you added

sebastienros:
I updated the PR. @SteveSandersonMS is this something we could merge in your PR as it's impacting how configuration is passed to the clients. Other option is I wait for your PR to be done and then react to the changes.

SteveSandersonMS:
@sebastienros I'm fine with it being merged into my PR, but my PR itself can't be completed until we get the next build of Microsoft.Extensions.AI, because the agreed API shape depends on API changes we've just merged into `dotnet/extensions` today.

sebastienros:
@eerhardt does it make sense to split this PR in two: the deployment API changes creating the ENVs, and the MEAI part that is using it and would be part of Steve's.

Or just move everything to Steve's branch and ship everything at once.

eerhardt:
> @eerhardt does it make sense to split this PR in two: the deployment API changes creating the ENVs, and the MEAI part that is using it and would be part of Steve's.
> 
> Or just move everything to Steve's branch and ship everything at once.

Completely up to you and Steve. I don't see a reason why they would need to be separate, but if you think they should be, I wouldn't argue.

danmoseley:
@SteveSandersonMS is this PR still relevant/in progress?

SteveSandersonMS:
@danmoseley It's @sebastienros's PR, so he should comment, but I think the functionality could still be relevant. I don't think it's urgent or a pre-req for anything else we have going on though.

eerhardt:
@dotnet-policy-service - please keep this PR open.

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIExtensions.cs
```diff
@@ -14,6 +14,8 @@ namespace Aspire.Hosting;
 /// </summary>
 public static class AzureOpenAIExtensions
 {
+    internal const string DefaultConfigSectionName = "Aspire:Azure:AI:OpenAI";
+
```
eerhardt, line 18:
```suggestion
```

src/Aspire.Hosting.Azure.CognitiveServices/AzureOpenAIExtensions.cs
```diff
@@ -116,8 +100,44 @@ public static IResourceBuilder<AzureOpenAIResource> AddAzureOpenAI(this IDistrib
...
+        }
+
+        dependency = cdkDeployment;
+        return dependency;
+    }
+
     /// <summary>
-    /// Adds an Azure OpenAI Deployment resource to the application model. This resource requires an <see cref="AzureOpenAIResource"/> to be added to the application model.
+    /// Adds an Azure OpenAI Deployment to the <see cref="AzureOpenAIResource"/> resource. This resource requires an <see cref="AzureOpenAIResource"/> to be added to the application model.
     /// </summary>
```
sebastienros, line 141: Can we obsolete this one now or does it still server a purpose?

eerhardt: That is a good topic for discussion. On one hand, we don't necessarily _have to_ obsolete it. But telling people "hey you should move from this old API to new AI" seems useful. Thoughts?

captainsafia (approved):

PR #6591, opened 2024-11-02
Rewrite dashboard URL in the console.

### LINKED ISSUE #6586: Display port forwarded URL when running app host in Codespace
Currently in `main` we have code that automatically translates URLs in the dashboard to the port forwarded URLs that GitHub Codespaces provides. One of the remaining things to do to round out our Codespaces support is modifying the URL that is printed to the console so that it uses the port forwarded URL instead of the localhost host URL.

This will also fix the issue that when clicking on the local host URL that the auth token is lost meaning that people need to go digging through the console logs to get it.

mitchdenny:
## Description

Restructures the codespace URL rewriting a bit to extract the functionality out into a service in DI (which is then used by a resource rewriter service each time a URL resource is updated). We also use this new service to translate the dashboard launch URL so you can just click on the hyperlink.

Fixes #6586 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6591)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -76,7 +76,9 @@
 				"ms-dotnettools.vscodeintellicode-csharp",
 				"ms-azuretools.vscode-bicep",
 				"EditorConfig.EditorConfig",
-				"ms-azuretools.azure-dev"
+				"ms-azuretools.azure-dev",
+				"GitHub.copilot",
+				"GitHub.copilot-chat"
 			],
 			"settings": {
 				"remote.autoForwardPorts": false,
--- a/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
+++ b/src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
@@ -13,7 +13,7 @@ internal sealed class CodespacesUrlRewriter(IOptions<CodespacesOptions> options)
 
         if (!options.Value.IsCodespace)
         {
-            throw new InvalidOperationException("Cannot translate URL because not currently running in a GitHub Codespace.");
+            return url;
         }
 
         return RewriteUrl(new Uri(url, UriKind.Absolute));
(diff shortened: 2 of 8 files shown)


.devcontainer/devcontainer.json
```diff
@@ -76,7 +76,9 @@
 				"ms-dotnettools.vscodeintellicode-csharp",
 				"ms-azuretools.vscode-bicep",
 				"EditorConfig.EditorConfig",
-				"ms-azuretools.azure-dev"
+				"ms-azuretools.azure-dev",
+				"GitHub.copilot",
+				"GitHub.copilot-chat"
```
davidfowl, line 81: Is this supposed to be usable by anyone or just the team 😄

mitchdenny: Just because the extension is installed doesn't mean people have to use it :)

src/Aspire.Hosting/Codespaces/CodespacesUrlRewriter.cs
```diff
@@ -1,76 +1,32 @@
...
 
-                    if (remappedUrls is not null)
-                    {
-                        var transformedUrls = from originalUrl in resourceEvent.Snapshot.Urls
-                                              select remappedUrls.TryGetValue(originalUrl, out var remappedUrl) ? remappedUrl : originalUrl;
+    public string RewriteUrl(Uri uri)
+    {
+        if (!options.Value.IsCodespace)
+        {
+            throw new InvalidOperationException("Cannot translate URL because not currently running in a GitHub Codespace.");
```
davidfowl, line 26: Why is this throwing? We don't want to stop execution, just log and skip

mitchdenny: This method shouldn't get called unless we are running in a codespace.

davidfowl: Or it could check and return the input url

davidfowl (approved):

PR #6596, opened 2024-11-03
Update schema

### LINKED ISSUE #6581: Publish 9.0 Aspire schema to schema store
We currently have the 8.0 manifest schema doc uploaded to https://json.schemastore.org/aspire-8.0.json.

We made changes in 9.0, see https://github.com/dotnet/aspire/pull/6333.

We should add the new version to the schema store.

cc @mitchdenny

mitchdenny:
## Description

Updating the schema to aspire-9.0. Just a rename of everything.

Fixes #6581

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

### DIFF
--- a/Aspire.sln
+++ b/Aspire.sln
@@ -135,6 +135,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{991DB378-6CB5-4441-BFC3-657400690FC3}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		src\Schema\aspire-8.0.json = src\Schema\aspire-8.0.json
 		Directory.Build.props = Directory.Build.props
 		Directory.Build.targets = Directory.Build.targets
 		Directory.Packages.props = Directory.Packages.props
--- a/src/Schema/aspire-8.0.json
+++ b/src/Schema/aspire-8.0.json
@@ -488,7 +488,7 @@
   "definitions": {
     "resource.azure.bicep.v0": {
       "type": "object",
-      "default": "Represents a resource that is deployed using Azure Bicep.",
+      "description": "Represents a resource that is deployed using Azure Bicep.",
       "required": [ "path" ],
       "properties": {
         "type": {
@@ -523,18 +523,19 @@
           "$ref": "#/definitions/resource.azure.bicep.v0"
         },
         {
+          "type": "object",
           "properties": {
             "type": {
               "const": "azure.bicep.v1"
             },
             "scope": {
-                "type": "object",
-                "properties": {
-                  "resourceGroup": {
-                    "type": "string",
-                    "description": "The name of the resource group to deploy the resource to."
-                  }
+              "type": "object",
+              "properties": {
+                "resourceGroup": {
+                  "type": "string",
+                  "description": "The name of the resource group to deploy the resource to."
                 }
+              }
             }
           }
         }


src/Shared/SchemaUtils.cs
```diff
@@ -5,5 +5,5 @@ namespace Aspire.Hosting.Utils;
 
 internal static class SchemaUtils
 {
-    public const string SchemaVersion = "https://json.schemastore.org/aspire-8.0.json";
+    public const string SchemaVersion = "https://json.schemastore.org/aspire-9.0.json";
```
eerhardt, line 8: Should this go into the release/9.0 branch in our first servicing?

eerhardt (approved):

src/Shared/SchemaUtils.cs
mitchdenny: Need to discuss strategy here with @davidfowl, @vhvb1989, and @ellismg

eerhardt: It's unfortunate that using the 9.0.0 bits today, can produce a manifest that doesn't conform to the schema it says it conforms to. For example:

1. Call the `PublishAsAzureContainerApp` method
2. This will produce a manifest that points to `https://json.schemastore.org/aspire-8.0.json` and also has `project.v1`, which isn't valid for that schema version.

Our playground app does this today:

https://github.com/dotnet/aspire/blob/11c51e7a34d068eb1881ee69e2dd3668511718d2/playground/AzureContainerApps/AzureContainerApps.AppHost/aspire-manifest.json#L2

https://github.com/dotnet/aspire/blob/11c51e7a34d068eb1881ee69e2dd3668511718d2/playground/AzureContainerApps/AzureContainerApps.AppHost/aspire-manifest.json#L105-L106

mitchdenny: Yeah so there are two approaches we can take here. We can update the -8.0 schema on schema store with the one we have in our repo and just not bump the version, or we can ship a -9.0 schema and make sure it works everywhere.

eerhardt: I'd assume bumping the version is the "right" thing to do. It doesn't really make sense to add things and not change the version. Because what's the point of versioning it then?

davidfowl: It can be done dynamically

danmoseley: can someone clarify what's the necessary remaining work here. 

I see there is only the 8.0 on schema store. So far as I can tell the changes in the schema in main would not invalidate anything valid against the 8.0 schema. Also this PR removes all mention of the 8.0 schema in the repo. 

I guess the concern is folks using 9.0 (which shipped still using 8.0 schema but created manifests that do not validated against it). That goes away when 9.1 ships. those manifests were already created, but they are invalid today and remain invalid right? Unless manually updated to point to a 9.0 schema.

Assuming I understand correctly, I think we should just merge this as it's a strict improvement, and update schemastore to 9.0 using the 9.1 GA version of the schema. Mitigations for 9.0-era manifests can be separate. Right?

cc @davidfowl

davidfowl: version stays 8 forever we will continue to update it and we version individual resource types

eerhardt: Can we just drop the version number in the name? `https://json.schemastore.org/aspire.json`. Why do we have a version number at all?

danmoseley: > Why do we have a version number at all?

as I understand @davidfowl point, the changes from 8.0 schema are all additive, ie., all 8.0 manifests validate successfully against this latest schema. So it's still useful to those on older versions (albeit possibly misleading)
if and when that changes (eg., we add a new required section) then we'd increment the version

danmoseley:
OK, schema is now fixed so that it passes the strict validation at schemastore on both 8.0 and latest manifests.

danmoseley:
@davidfowl easy one to signoff & merge

davidfowl:
@danmoseley @captainsafia made schema changes in 9.1. So, we'll need to do this again.

captainsafia:
> @danmoseley @captainsafia made schema changes in 9.1. So, we'll need to do this again.

I assume you mean just a rename? I've already updated the schema with the changes needed for existing resources work.

danmoseley:
Yes, if you mean 4e1a0f23ebd11fa9b601e8b0b1bafb72c0bb0b81 then it's already in here and what I uploaded

PR #6601, opened 2024-11-04
NO MERGE: Open browser tab from command

mitchdenny:
## Description

This is just a spike on opening a browser tab in response to command defined in the app model executing. The biggest issue with this so far is that opening the tab is blocked by the popup blocker (for good reason).

To solve this problem, I think we are going to want to define a kind of `ResourceUrlCommandAnnotation` or something like that which results in the tab being opened in response to the user initiated click in a browser.

Another issue this PR has which is specific to codespaces it that the URL we insert isn't Codespace forwarding URL aware. So we'll need to be able to inject the URL command annotations after endpoints have been allocated.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6601)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
@@ -353,6 +353,10 @@ public partial class Resources : ComponentBase, IAsyncDisposable
             toastParameters.Intent = ToastIntent.Success;
             toastParameters.Icon = GetIntentIcon(ToastIntent.Success);
         }
+        else if (response.Kind == ResourceCommandResponseKind.Action && response.ActionKind == ResourceCommandResponseActionKind.OpenExternal)
+        {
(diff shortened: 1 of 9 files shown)


JamesNK:
We could choose violence and just call `Process.Start("http://theurl.com");` in the command. That launches a new browser with the URL.

(I think doing it in the browser is probably better, even if it requires showing a dialog with the link)

PR #6636, opened 2024-11-09
API for role assignments

### LINKED ISSUE #6161: Improve Azure Managed Identity Defaults
.NET Aspire’s default is to create a single managed identity used by all ACA Container Apps and to grant a highly privileged role from that managed identity to the Azure services.

We should reconsider using a single Managed Identity for all Container Apps by default. And instead create one per Container App, and only assign roles to the services that Container App needs.

davidfowl:
This adds support for APIs for modeling a more secure default security configuration when deploying to azure container apps. It includes a few things:

1. A new set of APIs that azure resources to expose a way to define per reference role assignments. This allows for a least privilege access to resources:

```C#
using Azure.Provisioning.Storage;

var storage = builder.AddAzureStorage("storage")
                   .RunAsEmulator(c => c.WithLifetime(ContainerLifetime.Persistent));
var blobs = storage.AddBlobs("blobs");

builder.AddProject<Projects.AzureContainerApps_ApiService>("api")
       .WithExternalHttpEndpoints()
       .WithReference(blobs)
       .WithRoleAssignments(storage, StorageBuiltInRole.StorageBlobDataContributor);

builder.Build().Run();
```

In the above `WithRoleAssignments` allows the user to specify which roles this project/container requires what roles for a specific storage account.

2. Each compute resource that becomes a container app, gets a roles bicep resource that allocates the managed identity and sets up specific role assignments for that identity based on both the either the per reference role assignments or the resource defaults.
    a. If the targeted resource is an existing resource, it needs to go into its own bicep resource so we can set the "Scope" correctly.

3. Role assignment scopes are resource references. This means we need the ability to get a reference to the underlying azure resource by name, so we can use an existing bicep resource reference in "roles" bicep module. This led to a strange API that was added to `AzureProvisioningResource` to do this in a way that avoids the compute resource needing to know how to get a reference to the underlying azure resource.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6636)


playground/AzureContainerApps/AzureContainerApps.AppHost/api-roles.module.bicep
```diff
@@ -0,0 +1,27 @@
+@description('The location for the resource(s) to be deployed.')
+param location string = resourceGroup().location
+
+param storage_outputs_name string
+
+resource identity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' = {
```
davidfowl, line 6: We generate a module per container/project and that module has both the user assigned identity and role assignments for that compute resource.

playground/AzureContainerApps/AzureContainerApps.AppHost/api-roles.module.bicep
```diff
@@ -0,0 +1,27 @@
...
+    principalId: identity.properties.principalId
+    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')
+    principalType: 'ServicePrincipal'
+  }
+  scope: storage
+}
+
+output id string = identity.id
+
+output clientId string = identity.properties.clientId
```
davidfowl, line 27: These end up as outputs for the container app to use for both runtime and to associate this user assigned identity with the container app.

playground/AzureContainerApps/AzureContainerApps.AppHost/api-roles.module.bicep
```diff
@@ -0,0 +1,27 @@
...
+param storage_outputs_name string
+
+resource identity 'Microsoft.ManagedIdentity/userAssignedIdentities@2023-01-31' = {
+  name: take('identity-${uniqueString(resourceGroup().id)}', 128)
+  location: location
+}
+
+resource storage 'Microsoft.Storage/storageAccounts@2024-01-01' existing = {
+  name: storage_outputs_name
+}
```
davidfowl, line 13: We need to reference the set of resources we are creating role assignments for.

src/Aspire.Hosting.Azure/RoleDefinition.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Aspire.Hosting.Azure;
+
+/// <summary>
+/// 
+/// </summary>
+/// <param name="Id"></param>
+/// <param name="Description"></param>
+
+// REVIEW: This should be part of the Azure.Provisioning APIs
+public record struct RoleDefinition(string Id, string Description);
```
davidfowl, line 13: I think we need a single type for role definitions and then per resource types that have the well known ones. That would reduce the API explosion.

cc @tg-msft

src/Aspire.Hosting.Azure.Storage/AzureStorageResource.cs
```diff
@@ -83,4 +86,16 @@ void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDiction
...
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="name"></param>
+    /// <returns></returns>
+    public override ProvisionableResource GetExistingResource(BicepValue<string> name)
+    {
+        var account = StorageAccount.FromExisting(Infrastructure.NormalizeBicepIdentifier(Name));
+        account.Name = name;
```
davidfowl, line 98: This the strangest part of this change. We need the ability to import the underlying resource use for the role assignment given a name. Look at [this](https://github.com/dotnet/aspire/pull/6636/files#diff-5c542bf0932589f40e044926646e54c8fed701df0595fcdf18025abbe80aaea3R11) to understand more.

This method is called from [here](https://github.com/dotnet/aspire/pull/6636/files#diff-40fca6106cdf3968888c9d2bf55bb47a99dd1087a005c291248d2783dcd5dc07R330)

mitchdenny:
Have you tried deploying this? Based on the Bicep being generated it looks like you are deploying the `-roles.module.bicep` file first and then taking the user assigned identity out of that. I think that is fine for the case where container apps are only accessing other resources - but we just need to double check this model work for container app to container app deployments.

I think it'll be OK because for container app to container app communication the container app identity (resource side) is not what matters, there would need to be a different Azure Entra app registration which represents the resource (which we are going to need to support defining in the app model).

You know if we get this working it's going to take a lot of the friction out of doing S2S auth.

davidfowl:
> Have you tried deploying this? Based on the Bicep being generated it looks like you are deploying the -roles.module.bicep file first and then taking the user assigned identity out of that. I think that is fine for the case where container apps are only accessing other resources - but we just need to double check this model work for container app to container app deployments.

Not yet, there was a bug in azd, [fixed in this pr](https://github.com/Azure/azure-dev/pull/4524). It'll work because the identity and role assignments are part of the provision phase. I'll update the description with the deployment details.

davidfowl:
Deployment works!

src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
```diff
@@ -198,5 +198,16 @@ private static ProvisioningParameter GetOrAddParameter(AzureResourceInfrastructu
 
         return parameter;
     }
+
+    private static string GetNameFromValueExpression(IManifestExpressionProvider ep)
```
eerhardt, line 202: @davidfowl - do you remember why you made this change? Is it still necessary?

eerhardt:
@davidfowl @mitchdenny @sebastienros - let me know what you think about the latest iteration. It feels like an OK solution, but I'm sure there are improvements that can be made.

eerhardt:
Taking this out of draft, as the code changes are ready for review. I still need to write more tests before I want to merge this. But the code changes should all be there.

src/Aspire.Hosting.Azure/AzureProvisioningOptions.cs
```diff
@@ -19,4 +19,10 @@ public sealed class AzureProvisioningOptions
     /// functionality for building Azure resources.
     /// </summary>
     public ProvisioningBuildOptions ProvisioningBuildOptions { get; } = new ProvisioningBuildOptions();
+
+    /// <summary>
+    /// Gets or sets a value indicating whether Azure resources should include default role assignments
+    /// in their bicep templates.
+    /// </summary>
+    public bool UseDefaultRoleAssignments { get; set; } = true;
```
davidfowl, line 27: Should we flip this to make default bool be false?

src/Aspire.Hosting.Azure/AzureResourcePreparer.cs
```diff
@@ -0,0 +1,248 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private async Task<HashSet<IAzureResource>> GetAzureReferences(IResource resource, CancellationToken cancellationToken)
+    {
+        if (resource.TryGetEnvironmentVariables(out var environmentCallbacks))
```
davidfowl, line 163: You need to do arguments as well.

tests/Aspire.Hosting.Azure.Tests/AzureManifestUtils.cs
```diff
@@ -0,0 +1,41 @@
...
+    public static async Task<(JsonNode ManifestNode, string BicepText)> GetManifestWithBicep(IResource resource, bool skipPreparer = false)
+    {
+        if (!skipPreparer)
+        {
+            var executionContext = new DistributedApplicationExecutionContext(DistributedApplicationOperation.Publish);
+            var azurePreparer = new AzureResourcePreparer(Options.Create(new AzureProvisioningOptions()), executionContext);
+            var resources = (IResourceCollection)Activator.CreateInstance(
+                typeof(IResourceCollection).Assembly.GetType("Aspire.Hosting.ApplicationModel.ResourceCollection")!)!;
+            resources.Add(resource);
+            await azurePreparer.BeforeStartAsync(new DistributedApplicationModel(resources), cancellationToken: default);
```
davidfowl, line 22: What is this jank?

eerhardt: Which part do you hate?
* Having to call AzurePreparer before generating the bicep?
* Reusing the internal ResourceCollection class via Reflection? If this is an issue, I can create a copied implementation of IResourceCollection.

eerhardt: I've updated the code to no longer use reflection. I added a convienence ctor on DistributedApplicationModel. Let me know if you think this is acceptable.

eerhardt (approved):
LGTM. Approving because @davidfowl opened this PR and can't approve his own PR.

PR #6654, opened 2024-11-13
Move dashboard login URL log message last

JamesNK:
## Description

While trying out the standalone dashboard I found that the log message with the login link was hard to find because it was mixed in the middle of other log messages.

This PR moves the login link to the last log message at startup.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6654)

### DIFF
--- a/playground/TestShop/BasketService/Repositories/RedisBasketRepository.cs
+++ b/playground/TestShop/BasketService/Repositories/RedisBasketRepository.cs
@@ -36,7 +36,7 @@ public class RedisBasketRepository(ILogger<RedisBasketRepository> logger, IConne
             return null;
         }
 
-        return JsonSerializer.Deserialize<CustomerBasket>(data!, s_jsonSerializerOptions);
+        return JsonSerializer.Deserialize<CustomerBasket>(data.ToString(), s_jsonSerializerOptions);
     }
 
     public async Task<CustomerBasket?> UpdateBasketAsync(CustomerBasket basket)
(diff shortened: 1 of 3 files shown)


playground/TestShop/BasketService/Repositories/RedisBasketRepository.cs
```diff
@@ -36,7 +36,7 @@ public IEnumerable<string> GetUsers()
             return null;
         }
 
-        return JsonSerializer.Deserialize<CustomerBasket>(data!, s_jsonSerializerOptions);
+        return JsonSerializer.Deserialize<CustomerBasket>(data.ToString(), s_jsonSerializerOptions);
```
JamesNK, line 39: I updated to new VS and got a warning about ambigious overload here between span and string. I believe this is caused by work in the compiler to make span more first class.

Fixed by explicitly calling ToString instead of previously using implicitly string conversion.

mitchdenny (approved):

PR #6682, opened 2024-11-14
Add eventing to test builder.

### LINKED ISSUE #6681: Expose IDistributedApplicationEventing on IDistributedApplicationTestingBuilder
It's not possible to use the new eventing system from tests utilizing `IDistributedApplicationTestingBuilder` right now as the `IDistributedApplicationEventing` instance from the inner app host builder is not exposed.

mitchdenny:
## Description

We missed this. It was added to the test builder we use for our internal tests, but not the public API that we ship. Quick fix.

Fixes #6681 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No (not required)
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6682)


DamianEdwards:
@mitchdenny should we think about adding a test that verifies that members added to `DistributedApplicationBuilder` are exposed on the test builder? I imagine something that will fail whenever a new member is added that then requires either updating the test builder to expose it, or updating the test to explicitly confirm that we don't want to expose it.

mitchdenny:
Or we could look at making the test builder derive?

DamianEdwards:
Would be a bigger (breaking?) change to change inheritance wouldn't it? Open to it, just wondering if there's something more pragmatic for the shorter term.

JamesNK (approved):

JamesNK:
> @mitchdenny should we think about adding a test that verifies that members added to `DistributedApplicationBuilder` are exposed on the test builder? I imagine something that will fail whenever a new member is added that then requires either updating the test builder to expose it, or updating the test to explicitly confirm that we don't want to expose it.

+1. Add a test with reflection over both. If the test builder is missing something from `DistributedApplicationBuilder` then throw an error unless in an allow list. New property on `DistributedApplicationBuilder` requires either adding it to test builder or adding it to allow list.

DamianEdwards:
@mitchdenny are you planning to add a test as suggested, or log a follow-up issue?

mitchdenny:
I'll add a test.

PR #6726, opened 2024-11-19
Use dev 9.0 image which contains GA bits.

mitchdenny:
## Description

There isn't a final build of the dotnet dev container image available yet so until it is out we can use this image which seems to contain the GA bits.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6726)


joperezr (approved):
thanks for fixing Mitch!

PR #6737, opened 2024-11-20
Add Service Bus emulator support

### LINKED ISSUE #6605: Add Service Bus emulator
We should have a Service Bus emulator. The implementation would match the Event Hubs one.

sebastienros:
## Description

Add support for the Service Bus emulator https://techcommunity.microsoft.com/blog/messagingonazureblog/introducing-local-emulator-for-azure-service-bus/4304457

This PR introduces a custom model to configure the emulator and the cloud service. There is another branch that was based on the CDK directly (and use it for the emulator JSON too) but some discussions led to create a custom, simplified one (not really simplified here for the sake of understanding limits), that is then converted to JSON for the emulator, and to CDK resources for provisioning. If the model is not sufficient (can be on purpose to limit it to what users use in most cases) then the emulator configuration can be defined in JSON directly via `WithConfigJson(path)`, customized with `RunAsEmulator(c => c.ConfigJson(JsonNode))`, and the CDK resources can be enhanced with the existing `ConfigureInfrastructure()`.

Fixes https://github.com/dotnet/aspire/issues/6605

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6737)


playground/AzureServiceBus/ServiceBus.AppHost/ServiceBus.AppHost.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
```
IEvangelist, line 2: Should be using the:

```xml
  <Sdk Name="Aspire.AppHost.Sdk" Version="9.0.0" />
```

here?

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/OptionalValue.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Aspire.Hosting.Azure.ServiceBus.ApplicationModel;
+
+/// <summary>
+/// Represents an optional value.
+/// </summary>
+/// <typeparam name="T">The type of the value.</typeparam>
+public sealed class OptionalValue<T>
+{
+
```
IEvangelist, line 12:
```suggestion
```

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusClientAffineProperties.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Indicates the Client ID of the application that created the
+    /// client-affine subscription.
+    /// </summary>
+    public OptionalValue<string> ClientId
+    {
+        get { return _clientId!; }
```
IEvangelist, line 28: Why is the `!` needed here? Seems unnecessary, as the field is `new()`'ed up inline.

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusClientAffineProperties.cs
```diff
@@ -0,0 +1,51 @@
...
+        set { _clientId!.Assign(value); }
```
IEvangelist, line 29: Why is the call to `Assign` needed? Looks like the implicit operator would handle this without the need for a backing field at all.

sebastienros: > Looks like the implicit operator would handle this

Maybe. It was done before I update the operator. I'd keep it this way to make it obvious.

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/OptionalValue.cs
```diff
@@ -0,0 +1,69 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Azure.ServiceBus.ApplicationModel;
+
+/// <summary>
+/// Represents an optional value.
+/// </summary>
+/// <typeparam name="T">The type of the value.</typeparam>
+public sealed class OptionalValue<T>
```
eerhardt, line 10: Why not just use nullable?

src/Aspire.Hosting.Azure.EventHubs/EventHubsEmulatorContainerImageTags.cs
```diff
@@ -5,12 +5,12 @@ namespace Aspire.Hosting.Azure.EventHubs;
 
 internal static class EventHubsEmulatorContainerImageTags
 {
-    /// <summary>mcr.microsoft.com</summary>
+    /// <remarks>mcr.microsoft.com</remarks>
```
eerhardt, line 8: Don't we need to make this change everywhere?

src/Aspire.Hosting.Azure.ServiceBus/Aspire.Hosting.Azure.ServiceBus.csproj
```diff
@@ -6,16 +6,24 @@
     <PackageTags>aspire integration hosting azure</PackageTags>
     <Description>Azure Service Bus resource types for .NET Aspire.</Description>
     <PackageIconFullPath>$(SharedDir)AzureServiceBus_256x.png</PackageIconFullPath>
+    <!--<NoWarn>$(NoWarn);RS0016</NoWarn>--><!-- Ignore API warnings until PR is ready -->
```
eerhardt, line 9:
```suggestion
```

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusEmulatorResource.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <param name="innerResource">The inner resource used to store annotations.</param>
+public class AzureServiceBusEmulatorResource(AzureServiceBusResource innerResource) : ContainerResource(innerResource.Name), IResource
+{
+    // The path to the emulator configuration file in the container.
+    internal const string EmulatorConfigJsonPath = "/ServiceBus_Emulator/ConfigFiles/Config.json";
+
+    private readonly AzureServiceBusResource _innerResource = innerResource;
+
+    /// <inheritdoc/>
+    public override string Name => _innerResource.Name;
```
eerhardt, line 20:
```suggestion
```

This isn't necessary since we pass `innerResource.Name` on the constructor.

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusClientAffineProperties.cs
```diff
@@ -0,0 +1,51 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Azure.ServiceBus.ApplicationModel;
```
eerhardt, line 4: Do we really need an `ApplicationModel` namespace? Can't these types just go under `Aspire.Hosting.Azure.ServiceBus`?

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
```diff
@@ -0,0 +1,60 @@
+//using Aspire.Hosting.Azure.ServiceBus.ApplicationModel;
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var serviceBus = builder.AddAzureServiceBus("sbemulator")
+    .RunAsEmulator() // Comment to deploy and use the Azure cloud
+    ;
+
+serviceBus
+    .AddQueue("myQueue", queue =>
```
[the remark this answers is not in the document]
bbrandt: @eerhardt In this case we would need to return an object that combines behavior of `IResourceBuilder<AzureServiceBusResource>` with what you are describing so that we could either modify the specified queue or start to specify a different queue or topic?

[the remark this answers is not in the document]
eerhardt: (full disclosure, I'm not a super fan of fluent APIs)

If you want to start to specify a different queue, you would start a new code line.

```C#
var serviceBus = builder.AddAzureServiceBus("sbemulator");

var queue1 = serviceBus.AddQueue("queue1");
// set properties on queue1

var queue2 = serviceBus.AddQueue("queue2");
// set properties on queue2

File: src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusEmulatorResource.cs [hunk 11]
```diff
@@ -0,0 +1,24 @@
...
+namespace Aspire.Hosting.Azure;
+
+/// <summary>
+/// Wraps an <see cref="AzureServiceBusResource" /> in a type that exposes container extension methods.
+/// </summary>
+/// <param name="innerResource">The inner resource used to store annotations.</param>
+public class AzureServiceBusEmulatorResource(AzureServiceBusResource innerResource) : ContainerResource(innerResource.Name), IResource
+{
+    // The path to the emulator configuration file in the container.
+    internal const string EmulatorConfigJsonPath = "/ServiceBus_Emulator/ConfigFiles/Config.json";
```
timheuer, line 15: assume this is still WIP as would need a way to supply config.json based ont he docs as well?

sebastienros: This is the internal path for the emulator, this won't change. We do bind your own files to this internal path. Use the `WithConfigJson(yourpath)` for that. And internally it will bind to this path.

src/Aspire.Hosting.Azure.EventHubs/EventHubsEmulatorContainerImageTags.cs
sebastienros: This change was already made everywhere in the release/9.0 branch. It just didn't make its way to main. We need to sync the code bases in a different PR (David Pine's one?).

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusClientAffineProperties.cs
sebastienros: OptionalValue was removed

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/OptionalValue.cs
sebastienros: Removed

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusEmulatorResource.cs
sebastienros: I will add another method to configure a `JsonObject` representing the document in case we want to alter either the generated one of the bound one.

oising: > This is the internal path for the emulator, this won't change. We do bind your own files to this internal path. Use the `WithConfigJson(yourpath)` for that. And internally it will bind to this path.

Oh, that would be nice to get in for Event Hubs emulator to unblock a few things while the API evolves to match this app model idea.

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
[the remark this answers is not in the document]
sebastienros: > Why doesn't .AddQueue return the Queue, and then you can modify it as you want?

This would be a breaking change. Is that something we'd want? But we could then reference the resources when building the model, or do `topic.AddSubscription("sub.1")`

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusEmulatorResource.cs
sebastienros: Done, added `ConfigureJson` in the `RunEmulator`'s lambda.

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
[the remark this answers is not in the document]
eerhardt: > This would be a breaking change. Is that something we'd want?

Any way we could make it without a breaking change? A new method name? `AddServiceBusQueue`? 

cc @mitchdenny @davidfowl - thoughts on using callbacks in the API this way?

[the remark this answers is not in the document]
eerhardt: I had a chat with @DamianEdwards offline. We came to the conclusion that the current API isn't ideal. Ideally our APIs would follow either:

1. `parent.AddChild('name')` - returns the Child type (in this case a Queue)
or
2. `parent.WithChild('name', Action<Child> callback)` - returns the Parent type

The general rule - if you are returning a different type, use `Add`. If you are returning the same chained type, use `With`.

Since we already shipped `serviceBus.AddQueue('name')` returning the ServiceBus type, we could go with option 2 here.

The only drawback to 2 is that if you want to hold onto the `Child` in a variable, it gets messy. But the scenarios for needing a Queue/Topic/Subscription and using it elsewhere are limited. The only one I can think of is to get the name of the Queue and pass it into an environment variable to your app (or some other resource).

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+                ReplyToSessionId = "sessionId",
+                SessionId = "session1",
+                SendTo = "xyz"
+            }
+        };
+        subscription.Rules.Add(rule);
+    })
+    ;
+
+serviceBus.RunAsEmulator(configure => configure.ConfigureJson(document =>
```
eerhardt, line 39: Can we think of a better name than `ConfigureJson` here? That doesn't seem very descriptive of what it is doing.

What does the emulator call the json file? Maybe `ConfigureEmulatorConfiguration`? (maybe that's too many configure words...)

playground/AzureServiceBus/ServiceBusWorker/Program.cs
```diff
@@ -0,0 +1,19 @@
...
+var builder = Host.CreateApplicationBuilder(args);
+
+builder.AddServiceDefaults();
+
+builder.AddAzureServiceBusClient("sbemulator");
+
+Console.WriteLine("ServiceBus producer/consumer test");
+
+builder.Services.AddHostedService<Consumer>();
+Console.WriteLine("Starting Service Bus consumer...");
```
eerhardt, line 12: (nit) these console writelines are kind of incorrect. All we did is add a service, but it is saying "starting...".

Can we just remove the writelines?

playground/AzureServiceBus/ServiceBusWorker/Consumer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        processor.ProcessErrorAsync += ErrorHandler;
+
+        await processor.StartProcessingAsync(cancellationToken);
+    }
+
+    private static Task MessageHandler(ProcessMessageEventArgs args)
+    {
+        // Process the message
+        Console.WriteLine($"Received message: {Encoding.UTF8.GetString(args.Message.Body)}");
```
eerhardt, line 27:
```suggestion
        logger.LogInformation($"Received message: {Encoding.UTF8.GetString(args.Message.Body)}");
```

playground/AzureServiceBus/ServiceBusWorker/Consumer.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+    {
+        var processor = client.CreateProcessor("queue1", new ServiceBusProcessorOptions
+        {
+            AutoCompleteMessages = true,
+            MaxConcurrentCalls = 1, // Process one message at a time
+        });
+
+        // Whenever a message is available on the queue
```
eerhardt, line 16: What does this comment mean?

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        _clientFactory = clientFactory;
+        _queueNamesFactory = queueNamesFactory;
+        _topicNamesFactory = topicNamesFactory;
+    }
+
+    /// <inheritdoc />
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        // ServiceBusAdministrationClient is not available in the emulator
```
eerhardt, line 35: What does this comment mean / what is its purpose?

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        try
+        {
+            _serviceBusClient ??= _clientFactory();
```
eerhardt, line 39: It doesn't really pay to cache this client in a field because the health check instance is newly created every time the health check is executed. The IHealthCheck instances aren't reused.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+
+            connectionString = await builder.Resource.ConnectionStringExpression.GetValueAsync(ct).ConfigureAwait(false);
+
+            if (connectionString == null)
+            {
+                throw new DistributedApplicationException($"ConnectionStringAvailableEvent was published for the '{builder.Resource.Name}' resource but the connection string was null.");
+            }
+
+            queueNames = serviceBusEmulatorResources.SelectMany(x => x.Queues).Select(x => x.Name).ToArray();
+            topicNames = serviceBusEmulatorResources.SelectMany(x => x.Topics).Select(x => x.Name).ToArray();
```
eerhardt, line 313: Do we really need to check every queue and topic each time the health check is run? Don't we really only need to check 1?

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusCorrelationFilter.cs
```diff
@@ -0,0 +1,67 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Azure.ServiceBus;
```
eerhardt, line 4: (nit) can you match the folder name on disk and the namespace?

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusTopic.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Text.Json;
+using System.Xml;
+
+namespace Aspire.Hosting.Azure.ServiceBus;
+
+/// <summary>
+/// Represents a Service Bus Topic.
+/// </summary>
+/// <remarks>
+/// List of properties from the CDK that are not exposed here:
```
eerhardt, line 13:
```suggestion
/// List of properties from Azure.Provisioning.ServiceBus that are not exposed here:
```

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusResource.cs
```diff
@@ -11,22 +12,30 @@ namespace Aspire.Hosting.Azure;
...
     /// <summary>
     /// Gets the connection string template for the manifest for the Azure Service Bus endpoint.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
-        ReferenceExpression.Create($"{ServiceBusEndpoint}");
+        IsEmulator
+        ? ReferenceExpression.Create($"Endpoint=sb://{EmulatorEndpoint.Property(EndpointProperty.Host)}:{EmulatorEndpoint.Property(EndpointProperty.Port)};SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;")
+        : ReferenceExpression.Create($"{ServiceBusEndpoint}");
 
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
```
eerhardt, line 40: Is this Azure Functions stuff going to work with the emulator? I assume it needs to be updated.

cc @captainsafia

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusTopic.cs
```diff
@@ -0,0 +1,126 @@
...
+            topic.RequiresDuplicateDetection = RequiresDuplicateDetection.Value;
+        }
+        return topic;
+    }
+
+    /// <summary>
+    /// Converts the current instance to a JSON object.
+    /// </summary>
+    /// <param name="writer">The Utf8JsonWriter to write the JSON object to.</param>
+    public void WriteJsonObjectProperties(Utf8JsonWriter writer)
```
eerhardt, line 101: Do these methods need to be public?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -96,13 +100,15 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
     /// <param name="builder">The Azure Service Bus resource builder.</param>
     /// <param name="name">The name of the topic.</param>
     /// <param name="subscriptions">The name of the subscriptions.</param>
-    public static IResourceBuilder<AzureServiceBusResource> AddTopic(this IResourceBuilder<AzureServiceBusResource> builder, [ResourceName] string name, string[] subscriptions)
+    /// <param name="configure">An optional method that can be used for customizing the <see cref="ServiceBusTopic"/>.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusResource> AddTopic(this IResourceBuilder<AzureServiceBusResource> builder, [ResourceName] string name, string[] subscriptions, Action<ServiceBusTopic>? configure = null)
```
eerhardt, line 105: Do we need to add the Action here, if we also have it in the With methods?

eerhardt: Should we consider obsoleting (or even removing??) the Add methods?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+        {
+            return builder;
+        }
+
+        // Add emulator container
+        var configHostFile = Path.Combine(Path.GetTempPath(), "AspireServiceBusEmulator", Path.GetRandomFileName() + ".json");
+
+        Directory.CreateDirectory(Path.GetDirectoryName(configHostFile)!);
+
+        var password = PasswordGenerator.Generate(16, true, true, true, true, 0, 0, 0, 0);
```
eerhardt, line 255: Is generating a new password every time going to work well?

Similarly, do people expect to maintain data across runs? Other emulators have a WithDataVolume feature. (I'm not saying I think we need this, I'm just asking.)

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+            .WithAnnotation(new ContainerImageAnnotation
+            {
+                Registry = ServiceBusEmulatorContainerImageTags.Registry,
+                Image = ServiceBusEmulatorContainerImageTags.Image,
+                Tag = ServiceBusEmulatorContainerImageTags.Tag
+            })
+            .WithAnnotation(customMountAnnotation);
+
+        var sqlEdgeResource = builder.ApplicationBuilder
+                .AddContainer($"{builder.Resource.Name}-sqledge",
```
eerhardt, line 274: Do we have to use sqledge? Or can/should we reuse our own Sql hosting package?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+                .WithImageRegistry(ServiceBusEmulatorContainerImageTags.AzureSqlEdgeRegistry)
+                .WithEndpoint(targetPort: 1433, name: "tcp")
+                .WithEnvironment("ACCEPT_EULA", "Y")
+                .WithEnvironment("MSSQL_SA_PASSWORD", password);
+
+        builder.WithAnnotation(new EnvironmentCallbackAnnotation((EnvironmentCallbackContext context) =>
+        {
+            var sqlEndpoint = sqlEdgeResource.Resource.GetEndpoint("tcp");
+
+            context.EnvironmentVariables.Add("ACCEPT_EULA", "Y");
```
eerhardt, line 286: We should check with the ServiceBus team that it is OK to auto accept the EULA here.

cc @DamianEdwards

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+                  () => new ServiceBusClient(connectionString, noRetryOptions) ?? throw new DistributedApplicationException($"{nameof(connectionString)} was not initialized."),
+                  () => queueNames ?? throw new DistributedApplicationException($"{nameof(queueNames)} was not initialized."),
+                  () => topicNames ?? throw new DistributedApplicationException($"{nameof(topicNames)} was not initialized.")),
+              failureStatus: default,
+              tags: default,
+              timeout: default));
+
+        builder.WithHealthCheck(healthCheckKey);
+
+        builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>((e, ct) =>
```
eerhardt, line 345: Does this need to be 2 separate event subscriptions? Can we just do everything (get the connection string and queue names plus write the json file) at one time?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+                    writer.WriteStartObject();                  //           "{ (Topic)"
+                    topic.WriteJsonObjectProperties(writer);
+
+                    writer.WriteStartArray("Subscriptions");    //             "Subscriptions": [
+                    foreach (var subscription in topic.Subscriptions)
+                    {
+                        writer.WriteStartObject();              //               "{ (Subscription)"
+                        subscription.WriteJsonObjectProperties(writer);
+
+                        #region Rules
```
eerhardt, line 405:
```suggestion
```

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+                if (configFileMount == null)
+                {
+                    throw new InvalidOperationException("The configuration file mount is not set.");
+                }
+
+                var configJsonAnnotations = emulatorResource.Annotations.OfType<ConfigJsonAnnotation>();
+
+                foreach (var annotation in configJsonAnnotations)
+                {
+                    using var readStream = new FileStream(configFileMount.Source!, FileMode.Open, FileAccess.Read);
```
eerhardt, line 452: Why write the whole file to disk above, only to read the file here and parse it? Can these 2 things happen at the same time?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithDataBindMount(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string? path = null)
+        => builder.WithBindMount(path ?? $".servicebus/{builder.Resource.Name}", "/data", isReadOnly: false);
+
+    /// <summary>
+    /// Adds a named volume for the data folder to an Azure Service Bus emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureServiceBusEmulatorResource"/>.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithDataVolume(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string? name = null)
```
eerhardt, line 490: What are the scenarios for these methods? Doesn't the actual "data" get stored in the sql container?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+        return builder;
+    }
+
+    /// <summary>
+    /// Configures the gateway port for the Azure Service Bus emulator.
+    /// </summary>
+    /// <param name="builder">Builder for the Azure Service Bus emulator container</param>
+    /// <param name="port">Host port to bind to the emulator gateway port.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithGatewayPort(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, int? port)
```
eerhardt, line 524: Is "gateway port" the right term here?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -124,19 +151,381 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithDataVolume(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string? name = null)
+        => builder.WithVolume(name ?? VolumeNameGenerator.Generate(builder, "data"), "/data", isReadOnly: false);
+
+    /// <summary>
+    /// Adds a bind mount for the configuration file of an Azure Service Bus emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureServiceBusEmulatorResource"/>.</param>
+    /// <param name="path">Path to the file on the AppHost where the emulator configuration is located.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithConfigJson(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string path)
```
eerhardt, line 499: Similar question about this name as `ConfigureJson` - `WithConfigJson` doesn't seem descriptive enough.

oising: Maybe `WithConfigFile(...)` would be more future proof and reuseable?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusResource.cs
captainsafia: Yes, the implementation of `IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration` should contain something like:

```csharp
void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration
{
  if (IsEmulator)
  {
     target[$"{connectionName}__ConnectionString"] = ServiceBusEndpoint;
  }
}
```

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
sebastienros: This is a client that is used to check the service and get queue metadata. It's used in the existing health check. But the emulator doesn't implement this (only realized after trying to figure why it didn't work).

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
sebastienros: Didn't know, thanks. I assume that's why the one in the package is using a helper with a static instance. Should we cache it in the factory? Is it fine to keep the cache in the factory itself then? Otherwise I can try to get it from the service provider.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: Sure, we could expect everything to be ready once one is available, but I didn't want to make that assumption. The implementation in the nuget package is per queue/topic, so it doesn't give that info. Have asked the team for now.

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusCorrelationFilter.cs
sebastienros: Which one to keep, the disk or the namespace (already changed it once based on feedback, so I chose wrong apparently)

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
eerhardt: The "normal" route we take is to put the client in DI, and then it gets cached there. But in the Hosting libraries that doesn't seem to be the pattern. If this is expensive to create, we can cache it in a closure in the AzureServiceBusExtensions. I'm doing that in the RabbitMQ updates I'm working on.

Or if it isn't that expesive, we can just create a new client every check.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: @joperezr was also asking, I am ok to make these obsolete then. Wasn't sure it would count as breaking.

src/Aspire.Hosting.Azure.ServiceBus/ApplicationModel/ServiceBusCorrelationFilter.cs
eerhardt: Decide on the namespace. Then match it with the folders. That is my normal operation.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: I didn't think about supporting this. Would be nice though. Can we punt the feature or do we assume we provide this anytime we can?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
eerhardt: We would still document obsoletion in our "what's new". I think we could obsolete them now. And then in 10 we could do a mass removal of obsolete APIs.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: Good question, will ask the team. I already asked about the warning that sqledge was getting deprecated but did not get an answer AFAIK.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: Consistency with the Event Hub and CosmosDB emulators.

eerhardt: Did we use the term "gateway port" in CosmosDB, copy it to Event Hubs, and now it is going to propogate everywhere? CosmosDB has a "gateway" connection mode. 

In normal containers we call this "WithHostPort":

https://github.com/dotnet/aspire/blob/29a23f3c5eb089c30fc11fcc0b411244b2708a0a/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs#L221
https://github.com/dotnet/aspire/blob/29a23f3c5eb089c30fc11fcc0b411244b2708a0a/src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs#L327
https://github.com/dotnet/aspire/blob/29a23f3c5eb089c30fc11fcc0b411244b2708a0a/src/Aspire.Hosting.MongoDB/MongoDBBuilderExtensions.cs#L161

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
sebastienros: > What does the emulator call the json file

They call it:

> the [config.json](https://github.com/Azure/azure-service-bus-emulator-installer/blob/main/ServiceBus-Emulator/Config/Config.json) configuration file.

And 

> supply any changes in JSON configuration before you run the emulator

sebastienros: Changing to `WithConfigurationFile` when it takes a file and `ConfigureEmulator` when it takes an `Action<JsonNode>`

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusResource.cs
sebastienros: I updated it, but used the connection string expression instead of the suggested code. Let me know if I was wrong, then I will ask why the event hub emulator is right ;)

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: It should work with SQL Server, a differentiating factor is that sqledge supports ARM, and hence MacOS. It also uses less resources, but not sure it's a deciding factor.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: Done, and created an issue in the docs repos.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: The file we are reading here could be the one we just populated, but it could be the one that was  provided as-is by the `WithConfigurationFile()` method. We still need to apply the custom callback from `ConfigureEmulator()`.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: I asked the team if that is a supported scenario (or if the db gets initialized on startup for instance).

davidfowl:
How close is this to merging

sebastienros:
I pushed the wrong branch :/ I wanted to update the EventHubs one and not lose my nice green build ... reverting

sebastienros:
> How close is this to merging

If you ask me it's perfect. But I haven't had an approval yet so I must be wrong

davidfowl:
I want to play with it 🙂. I think this is a good enough strawman to decide on follow ups. Should we mark some of these APIs experimental to get some leeway to change them?

sebastienros:
> I want to play with it

Check the playground app I created. And also the provisioning part, it should provision queues/topics...   There is a callback to customize the CDK objects, adding properties that don't matter locally, or are not available in the "common" model between the emulator and cdk.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -122,21 +132,373 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+    {
+        if (builder.ApplicationBuilder.ExecutionContext.IsPublishMode)
+        {
+            return builder;
+        }
+
+        // Add emulator container
+        var configHostFile = Path.Combine(Path.GetTempPath(), "AspireServiceBusEmulator", Path.GetRandomFileName() + ".json");
+
+        Directory.CreateDirectory(Path.GetDirectoryName(configHostFile)!);
```
eerhardt, line 239: We should be using `Directory.CreateTempSubdirectory` here - https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createtempsubdirectory?view=net-9.0. The temp directory isn't guaranteed to be private to the current user.

davidfowl: Yep, we use it all over aspire.

sebastienros: Done

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -122,21 +132,373 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithDataBindMount(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string? path = null)
+        => builder.WithBindMount(path ?? $".servicebus/{builder.Resource.Name}", "/data", isReadOnly: false);
+
+    /// <summary>
+    /// Adds a named volume for the data folder to an Azure Service Bus emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureServiceBusEmulatorResource"/>.</param>
+    /// <param name="name">The name of the volume. Defaults to an auto-generated name based on the application and resource names.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithDataVolume(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string? name = null)
```
eerhardt, line 461: Is `WithDataVolume` still valid? I didn't think the data got perserved across runs.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -122,21 +132,373 @@ public static IResourceBuilder<AzureServiceBusResource> AddQueue(this IResourceB
...
+        => builder.WithVolume(name ?? VolumeNameGenerator.Generate(builder, "data"), "/data", isReadOnly: false);
+
+    /// <summary>
+    /// Adds a bind mount for the configuration file of an Azure Service Bus emulator resource.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="AzureServiceBusEmulatorResource"/>.</param>
+    /// <param name="path">Path to the file on the AppHost where the emulator configuration is located.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureServiceBusEmulatorResource> WithConfigurationFile(this IResourceBuilder<AzureServiceBusEmulatorResource> builder, string path)
+    => builder.WithBindMount(path, AzureServiceBusEmulatorResource.EmulatorConfigJsonPath, isReadOnly: false);
```
eerhardt, line 471: (nit) readability

```suggestion
        => builder.WithBindMount(path, AzureServiceBusEmulatorResource.EmulatorConfigJsonPath, isReadOnly: false);
```

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: But I removed it ... my brain thinks I did, what did I remove then

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusResource.cs
```diff
@@ -11,28 +12,46 @@ namespace Aspire.Hosting.Azure;
...
     {
-        // Injected to support Azure Functions listener initialization.
-        target[$"{connectionName}__fullyQualifiedNamespace"] = ServiceBusEndpoint;
-        // Injected to support Aspire client integration for Service Bus in Azure Functions projects.
-        target[$"Aspire__Azure__Messaging__ServiceBus__{connectionName}__FullyQualifiedNamespace"] = ServiceBusEndpoint;
+        if (IsEmulator)
+        {
+            // Injected to support Azure Functions listener initialization.
+            target[$"{connectionName}__fullyQualifiedNamespace"] = ConnectionStringExpression;
```
eerhardt, line 45: I think this is supposed to be

```suggestion
            target[$"{connectionName}__ConnectionString"] = ConnectionStringExpression;
```

Can you test this to ensure it works? Maybe add the RunAsEmulator to the Functions playground app, and remove all these skip #ifs:

https://github.com/dotnet/aspire/blob/9d8b13880db667e30a91cd77a9aa8b614420a937/playground/AzureFunctionsEndToEnd/AzureFunctionsEndToEnd.AppHost/Program.cs#L23-L25

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// <inheritdoc />
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        // The emulator doesn't support ServiceBusAdministrationClient which would be the preferred method to check the health of the service.
+        // A ServiceBusClient is used instead.
+
+        try
+        {
+            _serviceBusClient ??= _clientFactory();
```
eerhardt, line 37:
```suggestion
            var serviceBusClient = _clientFactory();
```

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusHealthCheck.cs
```diff
@@ -0,0 +1,56 @@
...
+/// The standard ServiceBus health check is not sufficient since it requires predefined knowledge that there is at least one queue or one topic.
+/// At the time we register the health check, we may not know if there are any of each.
+/// </summary>
+internal sealed class ServiceBusHealthCheck : IHealthCheck
+{
+    private readonly Func<ServiceBusClient> _clientFactory;
+    private readonly Func<string?> _nameFactory;
+
+    private ServiceBusClient? _serviceBusClient;
+
```
eerhardt, line 19:
```suggestion
```

We don't need to cache this.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusResource.cs
sebastienros: I am getting 

```
[2025-01-04T00:31:09.280Z] The listener for function 'Functions.MyServiceBusTrigger' was unable to start. Azure.Messaging.ServiceBus: The value 'Endpoint=sb://localhost:51234;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;' is not a well-formed Service Bus fully qualified namespace. (Parameter 'fullyQualifiedNamespace').
```

Is it possible the functions runtime is not expecting the local url for SB? Hopefully it's just a bug in this PR, but it looks like functions is correctly seeing the connection string.

sebastienros: All working

eerhardt (approved):
This LGTM! Nice work here. This is going to be a very useful feature that our customers will enjoy. Let's get it merged.

davidfowl:
The functions test is failing

captainsafia:
@sebastienros I believe the test failure is related to the key format that we use to inject the connection string for the client integrations:

```diff
// Injected to support Aspire client integration for Service Bus in Azure Functions projects.
- target[$"Aspire__Azure__Messaging__ServiceBus__{connectionName}"] = ConnectionStringExpression;
+ target[$"Aspire__Azure__Messaging__ServiceBus__{connectionName}__ConnectionString"] = ConnectionStringExpression;
```

src/Aspire.Hosting.Azure.ServiceBus/ServiceBusQueue.cs
```diff
@@ -0,0 +1,192 @@
...
+    /// </summary>
+    public int? MaxDeliveryCount { get; set; }
+
+    /// <summary>
+    /// A value indicating if this queue requires duplicate detection.
+    /// </summary>
+    public bool? RequiresDuplicateDetection { get; set; }
+
+    /// <summary>
+    /// A value that indicates whether the queue supports the concept of
```
eerhardt, line 88: Is it weird that this property is named "RequiresSession", but the docs say "whether the queue supports the concept of sessions"?

sebastienros: This is coming from the CDK ... but independently of it being the correct description or not I wonder if instead of copying the value we could use the `<inheritdoc />`.

sebastienros:
Thanks @captainsafia for spotting the issue with the AF functional tests.

huberttrueselftrueme:
What is a status of this ? 

When we can install package ?

sebastienros:
@huberttrueselftrueme you can follow these instructions to use the nightly builds https://github.com/dotnet/aspire/blob/main/docs/using-latest-daily.md

londospark:
> @huberttrueselftrueme you can follow these instructions to use the nightly builds https://github.com/dotnet/aspire/blob/main/docs/using-latest-daily.md

They only seem to be helping me install the final .net 9 version at the moment, and nothing with this enabled. Is there an update to those docs anywhere?

sebastienros:
@londospark the page could be updated but the important data is the nuget feed url: https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json

Add this to your sources and you should see the updated versions, example:

https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet9/NuGet/Aspire.Hosting.Azure.ServiceBus/overview/9.1.0-preview.1.25103.7

kyurkchyan:
@sebastienros  I've pulled the latest nighly version of Aspire, I've created a very basic [sample repo](https://github.com/kyurkchyan/AspireSample) with aspire starter template and used the new `RunAsEmulator` extension method

```csharp
var serviceBus = builder.AddAzureServiceBus("ServiceBus")
    .RunAsEmulator()
    .ConfigureInfrastructure(infra =>
    {
        var resources = infra.GetProvisionableResources();
        var serviceBus = resources.OfType<ServiceBusNamespace>().Single();
        serviceBus.Name = GetResourceName("asb", infra);
    })
    .WithTopic("aspire",
        topic => { topic.Subscriptions.AddRange([new ServiceBusSubscription("TestSubscription")]); });
```

When I run the project both on Mac and Windows, I receive the following exception

```
Unhandled exception. System.ArgumentException: Bind mounts must specify an absolute path. (Parameter 'source')
   at Aspire.Hosting.ApplicationModel.ContainerMountAnnotation..ctor(String source, String target, ContainerMountType type, Boolean isReadOnly) in /_/src/Aspire.Hosting/ApplicationModel/ContainerMountAnnotation.cs:line 33
   at Aspire.Hosting.AzureServiceBusExtensions.<>c__DisplayClass7_0.<RunAsEmulator>b__2(BeforeStartEvent event, CancellationToken ct) in /_/src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs:line 351
   at Aspire.Hosting.Eventing.DistributedApplicationEventing.<>c__DisplayClass4_0`1.<<Subscribe>b__0>d.MoveNext() in /_/src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs:line 82
   at Aspire.Hosting.Eventing.DistributedApplicationEventing.PublishAsync[T](T event, EventDispatchBehavior dispatchBehavior, CancellationToken cancellationToken) in /_/src/Aspire.Hosting/Eventing/DistributedApplicationEventing.cs:line 69
   at Aspire.Hosting.DistributedApplication.ExecuteBeforeStartHooksAsync(CancellationToken cancellationToken) in /_/src/Aspire.Hosting/DistributedApplication.cs:line 351
   at Aspire.Hosting.DistributedApplication.RunAsync(CancellationToken cancellationToken) in /_/src/Aspire.Hosting/DistributedApplication.cs:line 310
   at Program.<Main>$(String[] args) in /Users/gagik/Documents/Projects/Learning/Aspire/AspireSample/AspireSample.AppHost/Program.cs:line 37
   at Program.<Main>(String[] args)
```

Is there any extra configuration that should be done to integrate the emulator support?

kyurkchyan:
For anybody that is developing .NET Aspire on mac and will need to use Service Bus Emulator.

Service Bus Emulator is using `azure-sql-edge` docker image as a dependency. If you are using docker version 4.33.0 you will receive a very strange error and the azure sql edge will crash. Take a look at [this comment](https://github.com/docker/for-mac/issues/7368#issuecomment-2275555064) for more details.

As the comment suggested, I downgraded to docker version 4.33.0, and the issue was resolved

woutervanranst:
@kyurkchyan or anyone else, is there guidance on how to get the local emulator up & running using the nightles?

I'm getting a ` The AMQP transport failed to open because the inner transport tcp3 is closed. (ServiceCommunicationProblem)`

AppHost:
```
var sb = builder.AddAzureServiceBus("servicebus")
        .RunAsEmulator()
```

Sender's Program.cs:
```
builder.AddAzureServiceBusClient("servicebus");

...

sender = serviceBusClient.CreateSender("test");
await sender.SendMessageAsync(new ServiceBusMessage("Hello World"), stoppingToken); // Exception is thrown here
```

eerhardt:
@woutervanranst - check out https://github.com/dotnet/aspire/blob/main/playground/AzureServiceBus/ServiceBus.AppHost/Program.cs which uses the emulator. Some tips to look out for:

1. Make sure you are using `WaitFor` if your app is immediately trying to connect the service bus on startup. That will delay your app until the ServiceBus emulator is running.
2. Make sure you passing the object returned from `builder.AddAzureServiceBus` to `.WithReference(serviceBus).WaitFor(serviceBus)`. Calling `.AddServiceBusQueue("test")` in a fluent fashion returns a new resource named "test", which has a different name than `.AddAzureServiceBus("servicebus")`. It is just like a database server and a database - there is a parent child relationship there, and each resource has their own name. The object you pass into `WithReference` needs to have the same name as you pass into `builder.AddAzureServiceBusClient("some name");`

sebastienros:
I updated the playground app @eerhardt mentioned to match your example and it worked fine for me:

```

var serviceBus = builder.AddAzureServiceBus("sbemulator").RunAsEmulator();

serviceBus.AddServiceBusQueue("queue1");

builder.AddProject<Projects.ServiceBusWorker>("worker")
    .WithReference(serviceBus).WaitFor(serviceBus);

builder.Build().Run();
```

woutervanranst:
My bad -- I messed up the order of `AddAzureServiceBusClient` and `AddHostedService` -- nothing to see here

Thanks for the swift responses

PR #6718, opened 2024-11-18
Fix triple slash usage of `inheritdoc`.

IEvangelist:
## Description

Many of the container-based hosting integrations include an `internal` tags class, that stores the various container registry, image and tags. Since these members are `internal` the .NET Aspire API docs doesn't expose them in the rendered API docs. Nor are they available to IntelliSense.

For example, see [`RedisBuilderExtensions.AddRedis` Method: Remarks](https://learn.microsoft.com/dotnet/api/aspire.hosting.redisbuilderextensions.addredis?view=dotnet-aspire-9.0#remarks).

To fix this, when we rely on `<inheritdoc />` we must specify the `path` (XPath) to the desired value. Before my change:

image

After my change:

image

This will also fix the .NET Aspire API docs too, the next time we release.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Is this introducing a breaking change?
      - [ ] Yes
        - Link to aspire-docs issue (please use this [`breaking-change` template](https://github.com/dotnet/docs-aspire/issues/new?template=04-breaking-change.yml)):
      - [ ] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)):
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6718)

PR #6747, opened 2024-11-20
Fix container tags comments

sebastienros:
## Description

Some changes that were done in release/9.0 were not merged back to main due to conflicts. This is fixing it. This should supersede https://github.com/dotnet/aspire/pull/6718 which was opting to point to the `<summary/>` tag instead, but this would be opposite to what we have in release/9.0.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6747)


sebastienros:
@eerhardt I couldn't understand some of your comments on the other PR, like https://github.com/dotnet/aspire/pull/6718#issuecomment-2483835343 and https://github.com/dotnet/aspire/pull/6718#issuecomment-2483827603. I don't any issue with these.

eerhardt:
> @eerhardt I couldn't understand some of your comments on the other PR, like [#6718 (comment)](https://github.com/dotnet/aspire/pull/6718#issuecomment-2483835343) and [#6718 (comment)](https://github.com/dotnet/aspire/pull/6718#issuecomment-2483827603). I don't any issue with these.

For the comment `(Although, that EventHubs one appears misplaced. It should be on the RunAsEmulator method.)` I've pushed an update to this PR to move the remarks to the correct method. The 3 I pointed to weren't updated in #6718. But they are updated here now (using the other way).

For the 2 in AzureOpenAIDeployment - you can see on the docs site that it isn't working:

https://learn.microsoft.com/en-us/dotnet/api/aspire.hosting.applicationmodel.azureopenaideployment.skuname?view=dotnet-aspire-9.0

image

sebastienros:
Thanks. The missing resolution in the docs site is probably something to fix in the docs generation, and if we can't then we may want to change this in the code and remove these substitutions.

I checked for the <remarks> tags and they don't work either, assuming that the docs are based on release/9.0 and not main.

@IEvangelist do you think there is a possibility this gets fixed in the docs.

sebastienros:
@IEvangelist any opinion? It's a free space, you can be honest ;)

IEvangelist:
> @IEvangelist any opinion? It's a free space, you can be honest ;)

Yeah, we reported a bug in the platform to state that there are limitations with the rendering of those values. I believe there are 3 issues raised, 2 by @gewarren and 1 by me. One thing I believe we should do whenever this is working, is to format the container images as follows:

- `repository/image/name:tag`

As that's the most commonly expected format for a fully qualified container image. Like I've done in a personal project here: https://github.com/IEvangelist/profanity-filter/blob/main/src/ProfanityFilter.Hosting/ProfanityFilterResourceBuilderExtensions.cs#L16

eerhardt (approved):
Since this is already in release/9.0, let's get this in `main`.

We should follow up on the online docs to ensure this gets fixed in learn.microsoft.com.

PR #6770, opened 2024-11-22
Add RabbitMQ.Client v7 client integration package

### LINKED ISSUE #3956: RabbitMQ.Client v7 breaks Aspire.RabbitMQ.Client component
There have been some major binary breaking changes in https://www.nuget.org/packages/RabbitMQ.Client/7.0.0-alpha.5. Our component doesn't even load when using this version. The first error I get is:

```
MissingMethodException: Method not found: 'Void RabbitMQ.Client.ConnectionFactory.set_UseBackgroundThreadsForIO(Boolean)'.
Microsoft.Extensions.Configuration.Binder.SourceGeneration.<BindingExtensions_g>F4A9562478640BAD0160976753D96FF712B41D5A2B91D22C8BB8580DF176A8A3D__BindingExtensions.BindCore(IConfiguration configuration, ref ConnectionFactory instance, bool defaultValueIfNotFound, BinderOptions binderOptions)
Microsoft.Extensions.Configuration.Binder.SourceGeneration.<BindingExtensions_g>F4A9562478640BAD0160976753D96FF712B41D5A2B91D22C8BB8580DF176A8A3D__BindingExtensions.Bind_ConnectionFactory(IConfiguration configuration, object instance) in BindingExtensions.g.cs
Microsoft.Extensions.Hosting.AspireRabbitMQExtensions+<>c__DisplayClass5_0.<AddRabbitMQClient>g__CreateConnectionFactory|0(IServiceProvider sp) in AspireRabbitMQExtensions.cs
```

But I assume there will be many others after this.

We will need to handle this breaking change by splitting our component into 2, one for each major version.

1. For the current `Aspire.RabbitMQ.Client` package, we should put a NuGet version limit on our dependency: `[6.8.1,7.0.0)`. This way people won't be able to update to the `7.0.0` version, which will break their app.
2. If RabbitMQ.Client ships an official `7.0.0` stable package during the .NET Aspire `8.x` lifetime, we can add a new, forked component named `Aspire.RabbitMQ.Client.v7` which will have a dependency on `7.0.0` and contain any updates so the .NET Aspire component will work with v7. People who explicitly want to use version 7 can opt into using this package.
3. When .NET Aspire 9 ships, we can "swap" the dependencies around.
    - The `Aspire.RabbitMQ.Client` package will be updated to depend on version 7 of RabbitMQ.Client.
    - If RabbitMQ.Client v6 is still in support, we can create `Aspire.RabbitMQ.Client.v6` which has the dependency limit `[6.8.1, 7.0.0)` and works with the version 6 of RabbitMQ.Client.
    - `Aspire.RabbitMQ.Client.v7` will be dead-ended. We won't make new .NET Aspire 9 versions of this package.

Note that we shouldn't need to do this for every component when the underlying client library ships a new major version. We will only need to follow this model when major breaking changes come into the underlying client library such that our library doesn't even work anymore (or other major breaks that convince us it is worth doing this).

cc @sebastienros @radical @joperezr @DamianEdwards

eerhardt:
## Description

RabbitMQ.Client shipped a version `7.0.0` that has major breaking changes. Specifically, it removed all sync APIs and only contain async APIs. To continue supporting new versions of RabbitMQ.Client, we need to add a new component `Aspire.RabbitMQ.Client.v7` that will work with the new version of RabbitMQ.Client. See https://github.com/dotnet/aspire/issues/3956 for the strategy to deal with these breaking changes.

Fix #3956

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] No
        - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/2360

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6770)


danmoseley:
So from now on, folks that want to use latest will need to pick package named v7, and to use a future v8 rabbitmq they'll need to pick the package named v7 as well. So we have the v7 name indefinitely, if I understand right. Is it worth instead the breaking change to do the reverse, freeze the existing package renaming it v6?

eerhardt:
> So from now on, folks that want to use latest will need to pick package named v7, and to use a future v8 rabbitmq they'll need to pick the package named v7 as well. So we have the v7 name indefinitely, if I understand right. Is it worth instead the breaking change to do the reverse, freeze the existing package renaming it v6?

See https://github.com/dotnet/aspire/issues/3956 for the strategy to deal with these breaking changes.

danmoseley:
Yeah I already read that and I was confused so I asked 🙂

> When .NET Aspire 9 ships, we can "swap" the dependencies around.
> The Aspire.RabbitMQ.Client package will be updated to depend on version 7 of RabbitMQ.Client

I don't think this change did this?
.
> If RabbitMQ.Client v6 is still in support, we can create Aspire.RabbitMQ.Client.v6 which has the dependency limit [6.8.1, 7.0.0) and works with the version 6 of RabbitMQ.Client.

We didn't do this either.

> Aspire.RabbitMQ.Client.v7 will be dead-ended. We won't make new .NET Aspire 9 versions of this package.

What am I missing?

eerhardt:
> > When .NET Aspire 9 ships, we can "swap" the dependencies around.
> > The Aspire.RabbitMQ.Client package will be updated to depend on version 7 of RabbitMQ.Client
> 
> I don't think this change did this? .

At the time that issue was written (in April), we were assuming RabbitMQ.Client v7 would ship before .NET Aspire 9. The general strategy is "on the next major version of .NET Aspire". So we will do the "swap" when .NET Aspire 10 ships since RabbitMQ.Client v7 and .NET Aspire 9 shipped at basically the same time.

> > If RabbitMQ.Client v6 is still in support, we can create Aspire.RabbitMQ.Client.v6 which has the dependency limit [6.8.1, 7.0.0) and works with the version 6 of RabbitMQ.Client.
> 
> We didn't do this either.

This would be done at the time of the "swap" (i.e. .NET Aspire 10) if RabbitMQ.Client v6 is still in support and used by customers.

tests/Aspire.RabbitMQ.Client.Tests/AspireRabbitMQLoggingTests.cs
```diff
@@ -148,6 +153,38 @@ public void TestException()
...
+            .GetField("Log", BindingFlags.Static | BindingFlags.Public)!
+            .GetValue(null)!;
+#endif
+
+    private static void LogInfo(string message)
+    {
+#if RABBITMQ_V6
+        RabbitMqClientEventSource.Log.Info(message);
+#else
+        s_log.GetType().GetMethod("Info")!.Invoke(s_log, new object[] { message });
```
danmoseley, line 168: Is it worth caching these methods, or is this nearly equivalently fast these days? @steveharter

eerhardt: These are methods used in tests that get called once. Theres no need to cache them.

eerhardt:
@radical @sebastienros - this is now ready for review.

The only thing left undone is that we should be using the `AspNetCore.HealthChecks.Rabbitmq.v6` NuGet package in `Aspire.RabbitMQ.Client`, but that package wasn't published as part of the 9.0.0 HealthChecks release. See https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2358.

sebastienros (approved):

PR #6780, opened 2024-11-25
Improve auto-forwarding for GitHub Codespaces and devcontainers.

mitchdenny:
## Description

This PR improves support for automatically forwarding ports for the following dev container scenarios:

1. GitHub Codespace via browser
2. GitHub Codespace via VSCode
3. Local Devcontainer via VSCode

This works by introducing a lifecycle hook that executes after endpoints are allocated. If execution within a codespace or local devcontainer is detected the hook will trigger the update to the machine wide `settings.json` file based on the configuration of endpoints within resources. In codespaces only http and https endpoints will be forwarded (developers can always set manual forwarding).

There is some timing sensitivity here. The settings file needs to be updated before it sees the endpoint in the console logs, if it doesn't there is a chance that the wrong protocol/label will be applied.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [x] Yes
      - [ ] Yes
      - [s] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6780)


davidfowl:
We can’t mutate global state fro the apphost.

mitchdenny:
> We can’t mutate global machine state from the apphost.

I think this is the only way we can do it unless we are given an API.

mitchdenny:
@davidfowl

src/Aspire.Hosting/Devcontainers/DevcontainerPortForwardingHelper.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Aspire.Hosting.Devcontainers;
+
+internal class DevcontainerPortForwardingHelper
+{
+    private const string CodespaceSettingsPath = "/home/vscode/.vscode-remote/data/Machine/settings.json";
+    private const string LocalDevcontainerSettingsPath = "/home/vscode/.vscode-server/data/Machine/settings.json";
+    private const string PortAttributesFieldName = "remote.portsAttributes";
+
+    public static async Task SetPortAttributesAsync(int port, string protocol, string label, CancellationToken cancellationToken = default)
```
davidfowl, line 15: My one request is to add a semaphore slim here and make sure only one thing is calling this at a time. Added bonus if we use exclusive file locking but that's a bit trickier cross OS I think.

mitchdenny:
Disabling a few flaky tests:
https://github.com/dotnet/aspire/issues/6867
https://github.com/dotnet/aspire/issues/6866

src/Aspire.Hosting/Devcontainers/DevcontainerPortForwardingLifecycleHook.cs
```diff
@@ -0,0 +1,73 @@
...
+    private readonly IOptions<DevcontainersOptions> _devcontainersOptions;
+
+    public DevcontainerPortForwardingLifecycleHook(ILoggerFactory loggerFactory, IOptions<CodespacesOptions> codespacesOptions, IOptions<DevcontainersOptions> devcontainersOptions)
+    {
+        _hostingLogger = loggerFactory.CreateLogger("Aspire.Hosting");
+        _codespacesOptions = codespacesOptions;
+        _devcontainersOptions = devcontainersOptions;
+    }
+
+    public async Task AfterEndpointsAllocatedAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken)
```
davidfowl, line 25: Why do we want to use this event instead of BeforeResourceStart? That way it's not blocking startup of the dashboard.

mitchdenny: Also even though this executes before the dashboard starts its a local I/O operation so not a huge impact to overall start performance. In my experimentation the dashboard is up and running before I've decided to interact with it.

davidfowl (approved):

davidfowl:
Great work @mitchdenny !

PR #6805, opened 2024-11-26
Allow changing display language in dashboard settings

### LINKED ISSUE #5009: Let me change language of dashboard

When I launch Aspire dashboard it showes with Polish language. Most developers in Poland don't use polish language in development tools and all polish terms releted to programming sounds unfamiliar/funny for us. For that reason please provide a way to change UI language to English 

I would like to have UI to change language in Aspire Dashboard or other way to change language

### Additional context

_No response_

adamint:
## Description

Currently, you can change the language of the Aspire dashboard by changing the language of the browser. This change adds a setting in the settings dialog to change the display language. Formatting is not affected, as `CultureInfo.CurrentCulture` is unchanged. Only `CultureInfo.CurrentUICulture` is changed.

The one edge case is if there is a culture (such as es-US or pt-PT) that the user is requesting that is different than our localized culture (es and pt-BR respectively). The resolved cultures will be `es` and `pt-BR`, so we should select these as the active cultures in the dropdown - except in zh, since there are two different `zh` cultures that we localize for.

The approach recommended by the [Blazor globalization docs](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-9.0#dynamically-set-the-culture-from-the-accept-language-header) was taken.

Animation

Fixes #5009 

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No


src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor
```diff
@@ -15,7 +17,15 @@
             <FluentRadio Value="@ThemeManager.ThemeSettingDark">@Loc[nameof(Dialogs.SettingsDialogDarkTheme)]</FluentRadio>
         </FluentRadioGroup>
 
+        <FluentCombobox TOption="CultureInfo"
```
JamesNK, line 20: I think this should be a FluentSelect. There is a known list of values, and it's not that large, so typing in a combo box doesn't seem useful.

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor
```diff
@@ -15,7 +17,15 @@
...
+                        Autocomplete="ComboboxAutocomplete.List"
+                        Immediate="true"
+                        Label="@Loc[nameof(Dialogs.SettingsDialogLanguage)]"
+                        Items="@_languageOptions"
+                        OptionText="@(c => c.NativeName.Humanize())"
```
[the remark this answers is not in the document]
adamint: We could determine this based on the Accept-Language header or whether UI culture differs from culture, but `CultureInfo.CurrentUICulture` would just be set to english so that option would be selected. What do you think?

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor
adamint: changed

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
```diff
@@ -44,6 +70,21 @@ private async Task SettingChangedAsync()
...
+        {
+            return;
+        }
+
+        var uri = new Uri(NavigationManager.Uri)
+            .GetComponents(UriComponents.PathAndQuery, UriFormat.Unescaped);
+
+        NavigationManager.NavigateTo(
+            DashboardUrls.SetLanguageUrl(_selectedUiCulture.Name, uri),
+            forceLoad: true);
```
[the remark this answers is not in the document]
adamint: Setting the cookie isn't the problem. Updating text and loading the new satellite assemblies is the real issue.

> Did you consider setting the cookie in the browser with JS interop like changing the theme? That plus triggering a StateHasChanged on the overall page to rerender everything seems like it could work. Although I wouldn't be surprised if changing the culture in the middle of a HTTP request is problematic.

I did, but this approach doesn't work. I've read more into this the past day, and while some workarounds are possible (see [here](https://github.com/dotnet/runtime/issues/98415)), I would strongly prefer continuing with the reload strategy for the following reasons:
- calling the undocumented internal api is not compatible with our current blazor setup
- we could download all resources, but that would require downloading more, with limited use
- changing the language is probably less likely than changing the theme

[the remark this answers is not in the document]
adamint: Yes, there are no workarounds for what we are doing. Those are the only workarounds I have found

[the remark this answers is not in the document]
JamesNK: I'll take a look and see whether I can find a way to make it work.

If a refresh is required then I think a confirmation should be displayed after changing the language, e.g.

> The dashboard must be refreshed when the language is changed. Are you sure you want to change the language?
> Yes, No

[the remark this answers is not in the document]
adamint: > If a refresh is required then I think a confirmation should be displayed after changing the language, e.g.

I don’t think this would be very useful. The user is pretty clear in their intent, the dialog adds unnecessary friction. Reloading (or opening a new page to select language in) is a common pattern, including on Microsoft websites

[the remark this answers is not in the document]
JamesNK: But it’s inconsistent with the rest of the website. Nothing else (adding filters, changing the theme, etc) triggers the dashboard to completely reload. Some page state could be lost.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -437,6 +438,19 @@ await Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.Si
...
         }
+
+        _app.MapGet("/api/set-language", (string? language, string redirectUrl, HttpContext httpContext) =>
+        {
+            if (language is not null)
+            {
+                httpContext.Response.Cookies.Append(
+                    CookieRequestCultureProvider.DefaultCookieName,
+                    // keep current culture for formatting, etc., but change ui (display) culture
+                    CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(CultureInfo.CurrentCulture.Name, language)));
```
JamesNK, line 449: Right now this is a session cookie and is lost when the browser is complete closed. Add an expiry:

> new CookieOptions { Expires = DateTimeOffset.UtcNow.AddYears(1) }

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
JamesNK, line 85: Add a comment here that a cookie must be set and the dashboard reloaded to use the new culture set by localization middle.

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
```diff
@@ -10,14 +12,38 @@ namespace Aspire.Dashboard.Components.Dialogs;
...
+        _languageOptions = DashboardWebApplication.LocalizedCultures
+            .Select(CultureInfo.GetCultureInfo)
+            .OrderBy(c => c.NativeName)
+            .ToList();
+
+        // User may also be using a variant of one of the supported cultures, in which case we should select the culture we support
+        // this doesn't work for zh as we support two different zh cultures
+        if (!_languageOptions.Contains(CultureInfo.CurrentUICulture)
+                 && !StringComparers.Culture.Equals(CultureInfo.CurrentUICulture.TwoLetterISOLanguageName, "zh")
+                 && _languageOptions.FirstOrDefault(c => StringComparers.Culture.Equals(c.TwoLetterISOLanguageName, CultureInfo.CurrentUICulture.TwoLetterISOLanguageName)) is var matchedCulture)
```
JamesNK, line 37: This seems like it could be prone to mistakes. Tests are needed here. It could be moved to a class and have some unit tests that test different cultures.

I looked at what the middleware does and it uses Parent property on culture. What about using that to match?

https://github.com/dotnet/aspnetcore/blob/81a2bab8704d87d324039b42eb1bab0d977f25b8/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs#L172

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -437,6 +438,19 @@ await Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.Si
         {
             _app.MapPost("/authentication/logout", () => TypedResults.SignOut(authenticationSchemes: [CookieAuthenticationDefaults.AuthenticationScheme, OpenIdConnectDefaults.AuthenticationScheme]));
         }
+
+        _app.MapGet("/api/set-language", (string? language, string redirectUrl, HttpContext httpContext) =>
```
JamesNK, line 442: Return `BadRequest` result if language or redirectUrl are null or empty.

drewnoakes:
My understanding is that `CultureInfo.CurrentUICulture` exists for the process. For server-side Blazor, does this mean that when one user sets a language, all users have that setting pushed on them?

That might not be such a problem in the local development case, but for hosted environments like ACA, it might be problematic. If an issue, can we fix it, or allow this feature to be suppressed in such cases?

JamesNK:
`CultureInfo.CurrentUICulture` is thread local.

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
adamint: Thanks, that's a better idea. Updated

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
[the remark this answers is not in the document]
JamesNK: Yeah, I think it would look better below. And slightly smaller text.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -437,18 +433,22 @@ await Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.Si
...
-    {
-        var supportedCultures = CultureInfo.GetCultures(CultureTypes.AllCultures)
-            .Where(culture => LocalizedCultures.Contains(culture.TwoLetterISOLanguageName) || LocalizedCultures.Contains(culture.Name))
-            .Select(culture => culture.Name)
-            .ToList();
-
-        // Non-standard culture but it is the default in many Chinese browsers. Adding zh-CN allows OS culture customization to flow through the dashboard.
-        supportedCultures.Add("zh-CN");
-        return supportedCultures.ToArray();
+        _app.MapGet("/api/set-language", (string? language, string redirectUrl, HttpContext httpContext) =>
```
JamesNK, line 437:
```suggestion
        _app.MapGet("/api/set-language", (string? language, string? redirectUrl, HttpContext httpContext) =>
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -437,18 +433,22 @@ await Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.Si
...
+        {
+            if (string.IsNullOrEmpty(language) || string.IsNullOrEmpty(redirectUrl))
+            {
+                return Task.FromResult(Results.BadRequest());
+            }
+
+            httpContext.Response.Cookies.Append(
+                CookieRequestCultureProvider.DefaultCookieName,
+                // keep current culture for formatting, etc., but change ui (display) culture
+                CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(CultureInfo.CurrentCulture.Name, language)),
```
JamesNK, line 447: Is it correct to keep the current culture for formatting? If you're viewing french language, shouldn't there also be french formatting of dates, numbers, etc?

src/Aspire.Dashboard/Utils/DashboardUrls.cs
```diff
@@ -136,4 +136,12 @@ public static string LoginUrl(string? returnUrl = null, string? token = null)
 
         return url;
     }
+
+    public static string SetLanguageUrl(string language, string redirectUrl)
+    {
+        var languageEscaped = Uri.EscapeDataString(language);
+        var redirectUrlEscaped = Uri.EscapeDataString(redirectUrl);
+
+        return $"/api/set-language?language={languageEscaped}&redirectUrl={redirectUrlEscaped}";
```
JamesNK, line 145: Is there a reason you chose `EscapeDataString` over `QueryHelpers.AddQueryString` to build up the URL? AddQueryString seems like it's the best tool for the job.

src/Aspire.Dashboard/Utils/GlobalizationHelpers.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace Aspire.Dashboard.Utils;
+
+internal static class GlobalizationHelpers
```
JamesNK, line 9: Nice. Good to have this stuff centralized.

src/Aspire.Dashboard/Utils/GlobalizationHelpers.cs
```diff
@@ -0,0 +1,54 @@
...
+           return true;
+       }
+
+       // this doesn't work for zh as we support two different zh cultures
+       if (matchParent && !StringComparers.Culture.Equals(culture.TwoLetterISOLanguageName, "zh"))
+       {
+           var parent = culture.Parent;
+           while (!Equals(parent, parent.Parent))
+           {
+               if (cultureOptions.Contains(parent))
```
JamesNK, line 43: Can you put a counter here and return false if it exceeds 5? (I think that's the number that the middleware uses)

Just want to avoid going into an infinite loop if someone customizes cultures to have a self referental parent. Maybe write a test here as well.

src/Shared/StringComparers.cs
```diff
@@ -25,6 +25,7 @@ internal static class StringComparers
     public static StringComparer OtlpFieldValue => StringComparer.OrdinalIgnoreCase;
     public static StringComparer OtlpSpanId => StringComparer.Ordinal;
     public static StringComparer HealthReportPropertyValue => StringComparer.Ordinal;
+    public static StringComparer Culture => StringComparer.OrdinalIgnoreCase;
```
JamesNK, line 28: nit
```suggestion
    public static StringComparer CultureName => StringComparer.OrdinalIgnoreCase;
```

src/Shared/StringComparers.cs
```diff
@@ -47,4 +48,5 @@ internal static class StringComparisons
     public static StringComparison OtlpFieldValue => StringComparison.OrdinalIgnoreCase;
     public static StringComparison OtlpSpanId => StringComparison.Ordinal;
     public static StringComparison HealthReportPropertyValue => StringComparison.Ordinal;
+    public static StringComparison Culture => StringComparison.OrdinalIgnoreCase;
```
JamesNK, line 51:
```suggestion
    public static StringComparison CultureName => StringComparison.OrdinalIgnoreCase;
```

src/Aspire.Dashboard/DashboardWebApplication.cs
adamint: > shouldn't there also be french formatting of dates, numbers, etc

There's no "french" formatting, within each language it will vary by locale, ie fr-CH has some differences than fr-CA. We could add the ability to change the formatting culture as well, but since the current formatting behavior reflects what the user sees everywhere else, since it's their OS formatting culture, I think it's ok

src/Aspire.Dashboard/Utils/DashboardUrls.cs
adamint: No reason, changed it.

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor.cs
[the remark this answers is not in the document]
adamint: image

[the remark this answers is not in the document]
adamint: What do you think of this?

src/Aspire.Dashboard/DashboardWebApplication.cs
JamesNK: I know there are different French-like cultures. It was an example.

I think changing the language in the dropdown should be the same as changing the language in the browser. If fr-CH is selected from the `Accept-Language` value then both the language and number/date formatting are changed together.

Likewise, if a language is specified by `culture` query string value then it is set as both by default. They're only different if both `culture` and `ui-culture` query string values are specified.

Having two selections for language and formatting is overkill. I've never seen another website that does this. The question is, what is the behavior that most people expect and what, and I think it is changing them together is best.

JamesNK (approved):
I think there are improvements to make, but can do it in a follow up PR.

PR #6819, opened 2024-11-27
Add commands to console logs page

### LINKED ISSUE #6626: Run Resource Commands from Console Logs Page

I've been doing some iteration of onboarding some db migrations into Aspire, which has involved me 

1. Go to Dashboard
2. Go to Console Logs
3. See Error and tweak migrations scripts to fix
4. Go back to the Dashboard to restart the resource
5. Go back to Console Logs to see why the process errored this time

As I'm able to iterate on the migrations outside of aspire, I can test changes by simply restarting the migration resoruce, rather than rebuilding the whole app host.

I'd like the ability to be able to start / stop and restart a resource (and potentially any other resource command) directly from the Console Logs page, without having to go back to to the dashboard and then back to console logs.  That way I could stay on the console log page the entire time - view errors, fix them then restart the migrations from within the console logs page.

Something like:
Image

### Additional context

_No response_

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/6626

Demo:

consolelogs-commands

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6819)


JamesNK:
Feels good to use. cc @davidfowl @DamianEdwards @leslierichardson95 in case you have any UI feedback.

davidfowl:
This looks really good to me

drewnoakes:
I wonder if some users will expect this to stop the logs, rather than the resource, which could be annoying or even dangerous (in ACA for example).

Some ideas that might help with this:

- Use a button with icon and text
- Move the start/stop button into the ... menu, so it always appears with text
- Move buttons next to resource combo box

JamesNK:
The button icons and text come from the resource commands. Right now they just say `Stop`, `Start`, `Restart`. The text doesn't make it any clearer what the button does.

The buttons on the page could be removed and commands are only in a nested `Resource commands` menu.

e.g. to stop a resource, you would go `...` -> `Resource commands` -> `Stop`

davidfowl:
If we add log specific commands where would they go?

JamesNK:
In the example above? Resource commands

JamesNK:
The tooltip can say what the button does:

image

Is that enough to remove confusion?

drewnoakes:
Only if the user reads the tool tip. Having a button that stops a resource without any confirmation that might be mistaken to do something else seems like it could cause some undue problems.

I'm thinking of the scenario where something is logging rapidly and the user wants to pause the logs to make it easier to read them, so they click the "stop" button thinking that's what it does.

What about splitting that "toolbar" area such that all resource related stuff is on one side (combo box, commands) and all log related stuff is on the other side (status, future pause/resume/clear buttons).

JamesNK:
I don't see a good place to put command buttons on the left. If they were placed between the dropdown and status message it wouldn't look good because there can be a variable number of commands for a resource, and the status message location would jump around.

JamesNK:
How about this:

image

drewnoakes:
> How about this:
> 
> image

This is what I had in mind. Seeing it, I think it's much clearer.

src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
```diff
@@ -44,7 +44,7 @@ internal static void AddLifeCycleCommands(this IResource resource)
                     return ResourceCommandState.Hidden;
                 }
             },
-            displayDescription: null,
+            displayDescription: "Start resource",
```
drewnoakes, line 47: Any ideas on how we could localize these? Options including having the server send all translated values, or sending a known key value that's then looked up in the dashboard.

drewnoakes (approved):

src/Aspire.Hosting/ApplicationModel/CommandsConfigurationExtensions.cs
JamesNK: I don't think there is any plan for this kind of text to be localizable. Like state text, the API for adding commands just takes one string. That would need to change, and then I guess the server would need to send the frontend lang in a cookie or query string.

PR #6820, opened 2024-11-27
Option to hide timestamps on console logs page

### LINKED ISSUE #5457: Add option to control timestamps on dashboard console log page

Aspire automatically adds timestamps to dashboard console logs. It would be useful to control how much detail is shown, or if they're shown at all.

UI to control timestamps on console logs page:

* Full (includes milliseconds and timezone)
* Simple (seconds precision - default)
* Off

### Additional context

_No response_

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/5457

Demo:

consolelogs-dates

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6820)

### DIFF
--- a/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
+++ b/src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
@@ -73,6 +73,10 @@ public sealed partial class ConsoleLogs : ComponentBase, IAsyncDisposable, IPage
     [Parameter]
     public string? ResourceName { get; set; }
 
+    [Parameter]
+    [SupplyParameterFromQuery(Name = "hideTimestamp")]
+    public bool? HideTimestamp { get; set; }
+
     private readonly CancellationTokenSource _resourceSubscriptionCts = new();
     private readonly ConcurrentDictionary<string, ResourceViewModel> _resourceByName = new(StringComparers.ResourceName);
     private ImmutableList<SelectViewModel<ResourceTypeDetails>>? _resources;
@@ -99,7 +103,7 @@ public sealed partial class ConsoleLogs : ComponentBase, IAsyncDisposable, IPage
         _resourceSubscriptionToken = _resourceSubscriptionCts.Token;
         _logEntries = new(Options.Value.Frontend.MaxConsoleLogCount);
         _noSelection = new() { Id = null, Name = ControlsStringsLoc[nameof(ControlsStrings.LabelNone)] };
-        PageViewModel = new ConsoleLogsViewModel { SelectedOption = _noSelection, SelectedResource = null, Status = Loc[nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsLoadingResources)] };
+        PageViewModel = new ConsoleLogsViewModel { SelectedOption = _noSelection, SelectedResource = null, Status = Loc[nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsLoadingResources)], HideTimestamp = false };
 
         var loadingTcs = new TaskCompletionSource();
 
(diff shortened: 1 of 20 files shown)


samsp-msft:
Cut the date, but keep the time. The dashboard is not typically used to inspect data from long time periods ago, but understanding the delta and frequency of the log messages based on time is important. Showing a long date isn't needed.

maddymontaquila:
> Cut the date, but keep the time. The dashboard is not typically used to inspect data from long time periods ago, but understanding the delta and frequency of the log messages based on time is important. Showing a long date isn't needed.

Ya agree with @samsp-msft here.

However this is implemented should play nicely with what we decide to do in the future if we start saving dashboard UI state for a local user - ie, resizing/hiding/adding columns, details window pane size

JamesNK:
Today the timestamp matches the ISO 8601 standard. All developers will recognize what it is. I hesistate to change it to just the time, because then it's no longer immediately recognisable as a timestamp.

I also don't want to overload the user with choice. I think the best choices are on and off, which means show all (full timestamp) or nothing.

adamint:
This looks good to me, but a request if possible would be to add a command to toggle UTC/local time

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -261,6 +265,33 @@ private void UpdateMenuButton()
...
+        else
+        {
+            _menuItems.Add(new()
+            {
+                OnClick = async () =>
+                {
+                    PageViewModel.HideLogDate = true;
+                    await this.AfterViewModelChangedAsync(_contentLayout, waitToApplyMobileChange: false);
+                },
+                Text = "Hide log dates",
```
adamint, line 290: add to resources

maddymontaquila:
> Today the timestamp matches the ISO 8601 standard. All developers will recognize what it is. I hesitate to change it to just the time, because then it's no longer immediately recognisable as a timestamp.
> 
> I also don't want to overload the user with choice. I think the best choices are simple on and off, which means show all (full timestamp) or nothing.

This makes sense. I'd keep it on  by default then, but add the option to change it. I think most people are going to expect timestamps like a regular logfile

JamesNK:
The experience here isn't a log file though, it's console output. A default console doesn't include a timestamp with every line. e.g. Windows command prompt or Windows terminal don't include a timestamp.

maddymontaquila:
Sure - but we call them "console logs" on the dashboard and I feel like they're spiritually logs. I'm willing to bend if you and others (@leslierichardson95 mainly) feel strongly that it's worth changing the default behavior though

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -261,6 +265,33 @@ private void UpdateMenuButton()
...
+        {
+            _menuItems.Add(new()
+            {
+                OnClick = async () =>
+                {
+                    PageViewModel.HideLogDate = false;
+                    await this.AfterViewModelChangedAsync(_contentLayout, waitToApplyMobileChange: false);
+                },
+                Text = "Show log dates",
+                Icon = new Icons.Regular.Size16.Eye()
```
drewnoakes, line 278: Rather than change the icon by state, can we use a latched/checked icon state instead? Or maybe we just need a better icon. We use an eye icon for masking secrets throughout.

JamesNK: Done:

image

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -73,6 +73,10 @@ private sealed class ConsoleLogsSubscription
     [Parameter]
     public string? ResourceName { get; set; }
 
+    [Parameter]
+    [SupplyParameterFromQuery(Name = "hideLogDate")]
+    public bool? HideLogDate { get; set; }
```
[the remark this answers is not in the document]
JamesNK: It could be, but I like leaving it nullable because the query string is missing when false. The explicit conversion from `bool?` to `bool` later in the code makes it obvious what behavior is when the query string isn't present.

JamesNK:
PR feedback applied. Seems like consensus is to leave timestamp by default.

Further changes to tweak page can be in follow up PRs.

JamesNK:
BTW this PR depends on https://github.com/dotnet/aspire/pull/6819 being merged first.

drewnoakes (approved):

PR #6839, opened 2024-11-28
Disables Integrated Security in the ConnectionString when using SQL hosting integrations.

danm-de:
Disables the _Integrated Windows Authentication_ when using `Aspire.Hosting.SqlServer` & `Aspire.Hosting.Azure.Sql` hosting integrations.

- adds the `Integrated Security=false` parameter to the ConnectionString.

## Description

Kerberos/Integrated Windows authentication is typically used in corporate environments. The SQL Management Studio remembers the last (successful) connection setting. As a developer,[ if I copy the connection string into the "_Additional Connection Settings_" tab](https://learn.microsoft.com/en-us/dotnet/aspire/database/sql-server-integration?tabs=dotnet-cli%2Cssms#connect-to-database-resources) and then click "Connect", I get an error message:

image

The additional parameter `Integrated Security=false` ensures that the SQL Management Studio uses the correct connection settings (authentication using username & password). The same probably applies to AzureSqlServer - but I couldn't test that.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes (adapted the previous/existing tests)
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No - I assume that the hosting image (and the connection string) will only be used in development and test environments.
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/playground/DatabaseMigration/DatabaseMigration.AppHost/aspire-manifest.json
+++ b/playground/DatabaseMigration/DatabaseMigration.AppHost/aspire-manifest.json
@@ -3,7 +3,7 @@
   "resources": {
     "sql1": {
       "type": "container.v0",
-      "connectionString": "Server={sql1.bindings.tcp.host},{sql1.bindings.tcp.port};User ID=sa;Password={sql1-password.value};TrustServerCertificate=true",
+      "connectionString": "Server={sql1.bindings.tcp.host},{sql1.bindings.tcp.port};User ID=sa;Password={sql1-password.value};TrustServerCertificate=true;Integrated Security=false",
       "image": "mcr.microsoft.com/mssql/server:2022-latest",
       "env": {
         "ACCEPT_EULA": "Y",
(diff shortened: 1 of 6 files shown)


davidfowl:
cc @IEvangelist

danmoseley:
@roji does this seem appropriate?

danmoseley:
@IEvangelist thoughts?

roji:
@danm-de what's the exact purpose of this PR? As far as I know, Integrated Security is already false by default in SqlClient - why would it be needed to explicitly set it to false in the connection string?

IEvangelist:
If the default is already `false`, I don't understand why setting it to `false` here makes a difference. The specific section called out in the docs demonstrates how to get the resource's configured connection string, and paste it into either SSMS or Azure Data Studio, if .NET Aspire is able to connect with these connection strings - but the tools are not, that's not really something we should try changing, is it?

danmoseley:
thanks @danm-de I"ll close.

I don't know where feedback goes for Management Studio (maybe @roji does) but you could ask in an issue/discussion in this repo: https://github.com/dotnet/SqlClient

roji:
@danmoseley am not sure where SSMS feedback should go, but yeah, an issue/discussion in the SqlClient repo is probably the right way forwards.

PR #6876, opened 2024-12-05
Update package dependencies

joperezr:
Updating transitive dependencies from packages.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6876)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,22 +6,22 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    <TestcontainersPackageVersion>3.10.0</TestcontainersPackageVersion>
+    <TestcontainersPackageVersion>4.0.0</TestcontainersPackageVersion>
     <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
     <PackageVersion Include="AspNetCore.HealthChecks.Oracle" Version="8.0.1" />
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.SQS" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.400.36" />
-    <PackageVersion Include="AWSSDK.Core" Version="3.7.400.36" />
+    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.401.6" />
+    <PackageVersion Include="AWSSDK.SQS" Version="3.7.400.59" />
+    <PackageVersion Include="AWSSDK.SimpleNotificationService" Version="3.7.400.59" />
+    <PackageVersion Include="AWSSDK.Core" Version="3.7.400.59" />
     <PackageVersion Include="AWSSDK.Extensions.NETCore.Setup" Version="3.7.301" />
     <PackageVersion Include="AWS.Messaging" Version="0.9.2" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AWS" Version="1.1.0-beta.6" />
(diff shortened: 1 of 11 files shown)


playground/AzureContainerApps/AzureContainerApps.AppHost/AzureContainerApps.AppHost.csproj
```diff
@@ -6,6 +6,8 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <IsAspireHost>true</IsAspireHost>
+    <!-- This is an apphost project, doesn't need a Newtonsoft.Json reference -->
+    <AzureCosmosDisableNewtonsoftJsonCheck>true</AzureCosmosDisableNewtonsoftJsonCheck>
```
joperezr, line 10: The new Azure Cosmos package added a target that runs on restore and surfaces an error when there is no direct dependency on Newtonsoft.Json. Full details can be found here: https://github.com/Azure/azure-cosmos-dotnet-v3/pull/4839 but TL;DR is that they internally depend on newtonsoft but don't want to flow that transitive dependency and instead allow end-user applications to control the version they need.

In our repo, we disable this warning on all places where Newtonsoft isn't needed (as the Cosmos SDK won't run), and we reference Newtonsoft directly when necessary.

Directory.Packages.props
```diff
@@ -6,22 +6,22 @@
...
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    <TestcontainersPackageVersion>3.10.0</TestcontainersPackageVersion>
+    <TestcontainersPackageVersion>4.0.0</TestcontainersPackageVersion>
     <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- AWS SDK for .NET dependencies -->
     <PackageVersion Include="AspNetCore.HealthChecks.Oracle" Version="8.0.1" />
-    <PackageVersion Include="AWSSDK.CloudFormation" Version="3.7.400.36" />
```
eerhardt, line 15: We shouldn't be shipping AWS packages from this branch anymore.

I'll backport https://github.com/dotnet/aspire/pull/6624 to release/9.0.

eerhardt: [[release/9.0] Remove AWS specific code. (dotnet/aspire#6879)](https://github.com/dotnet/aspire/pull/6879)

src/Directory.Build.props
```diff
@@ -5,6 +5,10 @@
     <SkipArcadeNoWarnCS1591>true</SkipArcadeNoWarnCS1591>
     <EnablePackageValidation Condition="'$(EnablePackageValidation)' == ''">true</EnablePackageValidation>
     <PackageValidationBaselineVersion Condition="'$(EnablePackageValidation)' == 'true' and '$(PackageValidationBaselineVersion)' == ''">$(BaselineVersionForPackageValidation)</PackageValidationBaselineVersion>
+
+    <!-- Azure Cosmos is now emitting a wanring if you are and end-user application and are not referencing Newtonsoft directly.
+    We disable that warning for this project as we are not a consumer application. -->
+    <AzureCosmosDisableNewtonsoftJsonCheck>true</AzureCosmosDisableNewtonsoftJsonCheck>
```
eerhardt, line 11: Let's just put this in the 2 projects that need it - not for every src project.

playground/AzureContainerApps/AzureContainerApps.AppHost/AzureContainerApps.AppHost.csproj
eerhardt: Why is the AppHost project depending on the Cosmos SDK? Is it for the health check? If so, then I think we actually DO want Newtonsoft to be present, or else the AppHost is going to fail the health check when Newtonsoft tries to get loaded and it isn't in the bin directly.

danmoseley:
@joperezr I'm assuming this isn't required any more & doesn't need retargeting to main. The updates are outdated, and apparently we're doing fine without this warning disabled in main the last 2 months?

PR #6878, opened 2024-12-05
Add authentication (username and password) to Aspire.Hosting.Seq containers

nblumhardt:
## Description

Opening a draft for feedback/direction - ping @eerhardt 👋 

I loosely followed the corresponding PR for this feature in the MongoDB integration, initially adding optional `username` and `password` arguments to `AddSeq()`.

I felt a bit uneasy about the difference in semantics between these and the similar arguments in `AddMongoDB()`: in that case, omitting the arguments causes default values to be used, including for the password, so the result is still secured. In the Seq case, because the username and password are UI features (not provided to other resources) there seems to be no way to surface generated default values, so when the parameters are omitted authentication is not enabled at all.

To signal the difference in semantics, and to make room for some more detailed XDOCs on behavior, I sketched an `IResourceBuilder<SeqResource>.WithAuthentication()` extension method instead. Not sure if this aligns with how the API is designed or meets discoverability goals, so let me know if parameters on `AddSeq()` would be preferred instead.

I'm also uncertain if/where to add tests for this, so any pointers in that direction would also be appreciated. Cheers! :-)

See https://github.com/dotnet/aspire/issues/6155, 

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6878)

### DIFF
--- a/src/Aspire.Hosting.Seq/SeqBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Seq/SeqBuilderExtensions.cs
@@ -14,6 +14,7 @@ public static class SeqBuilderExtensions
     // The path within the container in which Seq stores its data
     private const string SeqContainerDataDirectory = "/data";
 
+    private const string AcceptEulaEnvVarName = "ACCEPT_EULA";
     private const string UserEnvVarName = "SEQ_FIRSTRUN_ADMINUSERNAME";
     private const string PasswordEnvVarName = "SEQ_FIRSTRUN_ADMINPASSWORD";
 
@@ -38,7 +39,7 @@ public static class SeqBuilderExtensions
             .WithHttpEndpoint(port: port, targetPort: 80, name: SeqResource.PrimaryEndpointName)
             .WithImage(SeqContainerImageTags.Image, SeqContainerImageTags.Tag)
             .WithImageRegistry(SeqContainerImageTags.Registry)
-            .WithEnvironment("ACCEPT_EULA", "Y");
+            .WithEnvironment(AcceptEulaEnvVarName, "Y");
 
         return resourceBuilder;
     }
--- a/src/Aspire.Hosting.Seq/SeqBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Seq/SeqBuilderExtensions.cs
@@ -12,7 +12,10 @@ namespace Aspire.Hosting;
 public static class SeqBuilderExtensions
 {
     // The path within the container in which Seq stores its data
-    const string SeqContainerDataDirectory = "/data";
+    private const string SeqContainerDataDirectory = "/data";
+
+    private const string UserEnvVarName = "SEQ_FIRSTRUN_ADMINUSERNAME";
+    private const string PasswordEnvVarName = "SEQ_FIRSTRUN_ADMINPASSWORD";
 
     /// <summary>
     /// Adds a Seq server resource to the application model. A container is used for local development.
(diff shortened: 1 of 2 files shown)


eerhardt:
My original intention/thought is to use password based auth by default, like the other integrations - MongoDB, Postgres, RabbitMQ, etc.

>  In the Seq case, because the username and password are UI features (not provided to other resources) there seems to be no way to surface generated default values, so when the parameters are omitted authentication is not enabled at all.

Can you explain this a bit more?

The rough design is for the Hosting integration to generate a password (which gets persisted to User Secrets in between runs), and then it tells both the docker container and the .NET Web Application that password.

The docker container starts up and only accepts connections using that password. The .NET Web Application uses that password to connect to the container.

Is that not possible with Seq?

nblumhardt:
This could be my lack of familiarity with Aspire showing. How would the user obtain one of these generated/default passwords in order to log into the Seq UI?

nblumhardt:
> Is the same password used between a user logging into the UI and an app sending data to Seq? Or is it possible to have separate passwords? No password for the UI and a password based connection string in the app sending data.

Apps use API keys when sending data; these are usually given only `Ingest` permission, and not full read/write/setup access.

Each app or service should be assigned a unique API key so that ingestion can be tracked/managed independently for different sources. Because of this there's not usually a single/fixed API key provisioned when a new Seq instance is configured - API keys have to be generated through the UI, or created by calling the container's HTTP API when authenticated as a user.

Users log in with usernames/passwords or SSO credentials; users generally end up with much broader read/write/setup permissions.

Ideally, both user access and ingestion would be secured. It's possible to secure the UI without securing ingestion, but not vice-versa. In terms overall security benefit, the UI is definitely the more important of the two.

I'm open to trying anything - perhaps the service-level "AddSeq" (as opposed to the hosting one) could try creating API keys on start-up, if user credentials are available? (In production this probably wouldn't be a realistic approach - each service would be assigned a pre-created API key - but IIRC we're not really targeting production deployment with the Seq hosting integration.)

Keen to know what you think. Thanks for taking a look!

src/Aspire.Hosting.Seq/SeqBuilderExtensions.cs
```diff
@@ -35,15 +39,37 @@ public static IResourceBuilder<SeqResource> AddSeq(
...
+    public static IResourceBuilder<SeqResource> WithAuthentication(
+        this IResourceBuilder<SeqResource> builder,
+        IResourceBuilder<ParameterResource> username,
+        IResourceBuilder<ParameterResource> password)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(username.Resource.Value, nameof(username));
+        ArgumentException.ThrowIfNullOrWhiteSpace(password.Resource.Value, nameof(username));
+        return builder
+            .WithEnvironment(UserEnvVarName, username.Resource.Value)
+            .WithEnvironment(PasswordEnvVarName, password.Resource.Value);
```
davidfowl, line 66:
```suggestion
            .WithEnvironment(UserEnvVarName, username)
            .WithEnvironment(PasswordEnvVarName, password);
```

danmoseley:
@eerhardt could you respond to answer given above

danmoseley:
@eerhardt ?

eerhardt:
The other "admin UIs" we have:

* RedisCommander / RedisInsight
* PgAdmin / PgWeb
* Kafka UI
* RabbitMQ Managment
* CosmosDB emulator Data Explorer

don't have user/password protection. But none of these get deployed when the application is deployed. They are all "development time" assistants.

> but IIRC we're not really targeting production deployment with the Seq hosting integration

This is interesting to know. What is the plan for publishing an Aspire app that uses Seq?

My general thinking here is that this kind of authentication isn't critical to implement- since it is only on the dev machine and isn't used in production. The original intention of [Ensure all containers are password protected (dotnet/aspire#6155)](https://github.com/dotnet/aspire/issues/6155) was to ensure containers that were used in production were password protected. But from above it sounds like you wouldn't use the Seq container in production. Is that true?

If you want to move forward with this PR, I think that works. It just needs some tests.

nblumhardt:
Hi @eerhardt, thanks for your reply. Needed to sync up with the team on what our plans are here, sorry it took a while to get back to you!

In production, Seq really needs to be set up separately; things like local storage, authentication, URIs etc. are unlikely to be satisfactory when deployed simplistically using Aspire.

If I'm not mistaken - Seq won't be included in the deployment manifest today, so I think we're all good, and can close this 👍

Issue #5458, opened 2024-08-27
Add clear button to dashboard console logs page


JamesNK:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

There are situations where a developer only wants to see logs for the next action that happens. Being able to clear the logs so they don't have to separate old vs new logs would be useful.

### Describe the solution you'd like

A button on console logs page that clears the logs.

This could work by the dashboard calling the host, and the host calling DCP. Alternatively, we could "clear" the logs in the UI by recording the timestamp when the click happened and filter out logs older than the timestamp.

Felx-B:
Hello,
I think this feature should not be restricted to console logs. It would be nice if all logs/traces and metrics were clearable with a simple button.
When I work on a specific use case, it is usefull to clean everything to focus on the scenario I am debugging

JamesNK:
Clear buttons for structured logs/traces/metrics added in this PR: https://github.com/dotnet/aspire/pull/6898

Clear currently isn't supported on the console logs page.

PR #6898, opened 2024-12-08
Clear traces/structured logs functionality

Daluur:
## Description
Added a button to traces and structured pages, to clear all items.
I find it very useful to be able to easily get a clean slate, as it makes it easier to see what is new items. I have worked with other similar tools in the past, and have had a great use of such buttons.

ClearLogs
ClearTraces

The placement and icon of the buttons might not be perfect.
I utilized the existing `Clear` functionality in the `CircularBuffer` to implement this change.

This is similar to this reported issue: https://github.com/dotnet/aspire/issues/5458, just in different pages.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship. However, there might be feedback on the button placement and icon.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No (at least I don't think it is needed)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6898)


Daluur:
> @Daluur please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

JamesNK:
Nice.

I have some UI thoughts:

* Should there be a confirmation before clearing? i.e. a Yes/No dialog page that pops up and says are you sure you want to clear structured logs/traces. The console logs clear button in Chrome developer tools doesn't, so I lean towards no.
* The location of the button next to the resource dropdown could be confusing about what you're clearing. If `(All)` is selected then obviously you're clearing logs for all resources. If a resource is selected then are you still clearing all logs, or just logs for the selected resources?
* Doesn't need to be in this PR, but we'll want to do the same thing on the metrics page and console logs page. But console logs is more difficult because the data is stored externally. Either we'd need to tell the external source to clear its data, or filter it when it arrives to only logs after the cleared time.

Daluur:
> Nice.
> 
> I have some UI thoughts:
> 
> * Should there be a confirmation before clearing? i.e. a Yes/No dialog page that pops up and says are you sure you want to clear structured logs/traces. The console logs clear button in Chrome developer tools doesn't, so I lean towards no.
> * The location of the button next to the resource dropdown could be confusing about what you're clearing. If `(All)` is selected then obviously you're clearing logs for all resources. If a resource is selected then are you still clearing all logs, or just logs for the selected resources?
> * Doesn't need to be in this PR, but we'll want to do the same thing on the metrics page and console logs page. But console logs is more difficult because the data is stored externally. Either we'd need to tell the external source to clear its data, or filter it when it arrives to only logs after the cleared time.

- I think a pop up would just end up being annoying
- I agree that the button placement is not ideal, I had a hard time finding a better place within the page
- Initially I did not think it would be as useful for metrics, but it might be nice to simply clear everything.

What if, instead of having a button on every page, there is a single button somewhere, that clears everything; traces, metrics, and structured logs (and console logs at some point). The button could be located in the top bar next to the settings button, or perhaps inside the settings dialog, if in the header is too prominent.

I have done some work on clearing metrics and clearing traces when viewing trace details, but I have not been able to finish it today, I will likely continue work tomorrow

Daluur:
I made some updates.

Clear in top menu


- I moved the clear button to the top menu, next to settings
- I added functionality to clear metrics

The button clears all signal data, and can be extended to clear console logs in the future.
Clearing of metrics does not work 100%. I think it might be the way the data is populated from OLTP, where it might get 'old' data again, once it has been cleared, though, I am not certain. I intend to debug in a few days

JamesNK:
I'm not sure having it in the top toolbar is an improvement. That area isn't related to the data you see on the page. Also, maybe it is useful to be able to independently clear certain areas of telemetry (and individual resources) rather than a clear all button. Or maybe we have both, i.e. a clear all button in the dashboard settings sidebar _and_ clear buttons on individual telemetry pages.

I think the Aspire team is going to need to talk amoungst ourselves about how we want this feature to work. I'm not sure if that will happen this year - people are disappearing for holidays - but I'll make sure there is a decision before Aspire 9.1 ships, and if you're not available to finish the PR off when we have a decision then I'll make whatever changes are required to get it in.

I think this is valuable. Something that isn't perfect is better than not shipping.

JamesNK:
I thought about the UI more and talked it over with other Aspire folks offline. Good news is consensus on what we want the UI to look like here:

* Have a clear button next to the resource filter on each telemetry page (and the console logs page in the future). This is similar to what your PR originally had. However, clicking on it opens a menu with a couple of options:
    * `Remove all`
    * `Remove for resource` (if a resource could be selected. If `(All)` is selected then it is disabled)
    * Maybe, in the future, `Remove visible` (allows someone to remove an exact filtered set of telemetry)
This is like what Fiddler has in its UI. I think it is simple but gives lots of flexibility. Also, it is a sneaky way of requiring a confirmation: the user clicks to expand the menu, then clicks to remove. They won't remove data accidently by just clicking on the remove button.
    fiddler-remove
 
* Also have a `Remove all` button in settings sidebar (below where the theme is set). This button would clear all telemetry for all resources in one click. The is like the button you added top menu but moved to settings.

This is more complicated than your original PR, but I think the remove menu on the telemetry pages plus a remove all button in settings handles all scenarios while being intutive.

JamesNK:
For the remove menu, you can use the `AspireMenuButton` control. We use that in a couple of places. The actual button itself can be two icons `Delete` + `ChevronDown`.

If you have trouble with the menu, or logic to remove telemetry for one resource, let me know and I can help with the PR.

JamesNK:
I’m away for a week. I’ll take a look when I’m back.

Daluur:
I created a control for the dropdown menu, to reduce the amount of duplicated code, and simplify the logic. 
I also created a button in the settings sidebar to clear all signals.
I did also change the text from Remove -> Clear, as I believe it could be misinterpreted as removing the entire resource.

Happy Holidays :)

JamesNK:
I pushed some changes to your branch for the buttons:

image

JamesNK (approved):

JamesNK:
@Daluur I made some minor changes on your branch. I think it's in a mergable state.

Are you ok with the changes and merging?

Daluur:
@JamesNK, the visual upgrades are very good :)
Generally it seems good. However, I can see you changed the logic in the clearing of traces, where it now only removes traces where the root span is from the resource. I don't think that is the best behavior, as it might keep a bunch of traces you do not really care about when clearing

Only-removes-roots

JamesNK:
I see what you mean. However, I don't think it's a good idea to remove some spans from traces. If we do that then their child spans will be left without parents. A span with no parent becomes a root span and the waterfall graph becomes confusing to read. 

Instead, I think the behavior should be completely remove traces with a matching span.

JamesNK:
I tested it out and it feels like good behavior. Merging now.

We'll definitely call this out in the next release. I think people will love it. Thanks!

Daluur:
Thanks for helping to push it over the finish line 😊

Issue #3647, opened 2024-04-12
[AzureTools][Aspire][Question] Should filters for the Resources page be remembered?


v-mengwe:
**Clean machine:** Win11 x64 23h2 ENU
**VS Version:** VS 17.10.0 Preview 4.0 [34811.107.d17.10]
**Aspire Version:** 8.0.0-preview.5.24201.12
Apply NuGet Feeds
**Podman Desktop:** 1.9.1
**Podman CLI:** 5.0.1

**REPRO STEPS:**

1. Create an Aspire starter project with "Use Redis for caching (requires Docker) checked.
2. F5.
3. In the dashboard, click filters icon and change the filter criteria.
    ![image](https://github.com/dotnet/aspire/assets/62053665/74cb5f74-98be-45fa-af21-3ea14ebe7aad)
4. Click on other columns on the left, e.g. "Console" -> Return to "Resources" page.

**Actual:**
No filter conditions are remembered and default values are returned.
![Actual1](https://github.com/dotnet/aspire/assets/62053665/59dd1970-2d84-47a2-a751-52ebf3244ae6)

**Question:**
Is it by design? Or should we memorize filters?

mitchdenny:
Looks like we remember the details panel docking setting, but not filters or sort orders. /cc @JamesNK

leslierichardson95:
@leslierichardson95 add filter / sorting tag to this issue later

PR #6925, opened 2024-12-10
Allow filtering by resource state and health status, and update the filters based on visible grid items

### LINKED ISSUE #1190: Filter resources with errors on dashboard
Currently our resources page have filter based on the kind of resources. We should also add another filter for states. Specifically, filter for resources with error can be helpful if you have a lot of resources.

### LINKED ISSUE #4554: Resource Type filter includes hidden resource types
By default, `Parameter` resources are hidden from the dashboard, yet this type of resource shows up in the resource type filter.

Adjusting the filter to show only `Parameter` types, shows "no Resources"

image

It feels a bit odd to be given the option to select `Parameter` resources, and then be given no results when choosing that resource type.

adamint:
## Description

Currently, you can filter resources by type, but it would also be useful to filter by health status or by resource state. I extracted the resource filter UI into its own component `ResourceFilters` and converted `SelectResourceTypes` into a component that was more generic and so could be reused for state and health status (`SelectResourceOptions`). This also meant other refactoring was necessary, such as creating reusable methods `Get/SetFieldVisibility` from the getter and setter logic previously in `AreAllTypesVisible`.

Fixes #1190

In a follow-up, we could persist these values in the URL (#3647)

image

(note that, for example, finished and no state have disappeared from available options because no resource with those values is in the grid anymore)
image

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No (largely just a refactor)
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6925)


src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor
```diff
@@ -0,0 +1,55 @@
...
+    <FluentCheckbox Label="@ControlsStringsLoc[nameof(ControlsStrings.LabelAll)]"
+                    ThreeState="true"
+                    ShowIndeterminate="false"
+                    ThreeStateOrderUncheckToIntermediate="true"
+                    @bind-CheckState:get="@AreAllVisible"
+                    @bind-CheckState:set="@OnAllValuesCheckedChangedInternalAsync"
+    />
+
+    @foreach (var (resourceType, _) in AllValues
+                  .Where(pair => Resources.Any(resource => ResourceContainsValue(resource, pair.Key)))
```
adamint, line 21: This is inefficient but I figured it didn't matter much since this `ResourceContainsValue` isn't that expensive and this is only called on filter open.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor.cs
```diff
@@ -0,0 +1,73 @@
...
+
+using System.Collections.Concurrent;
+
+namespace Aspire.Dashboard.Components;
+
+public partial class SelectResourceOptions<TValue>
+{
+    private async Task OnAllValuesCheckedChangedInternalAsync(bool? newAreAllVisible)
+    {
+        AreAllVisible = newAreAllVisible;
```
adamint, line 12: Previously this line (and the logic in OnValueVisibilityChangedInternalAsync) was done in the callback but can be done directly in this component before invoking the callback.

JamesNK:
What happens if there are a lot of states, resources types and health states? The popup should probably have a maximum height and a scoll bar to handle that situation nicely.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor
JamesNK, line 21: I don't think this can work. The resources here are filtered, so as you filter something out it is no longer available in the list of resources.

IMO keep it simple and display checkboxes for all the values the dashboard has seen.

JamesNK:
Couple of examples of bugs.

Database type isn't in list because its resource is nested. Expanding the parent makes it display:

db-type-not-displayed

Problems with unselecting resource types:

resource-type-not-displayed

These are both caused by filtering available options based on visible resources.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor
adamint: You're right. Reverted

adamint:
> Couple of examples of bugs.
> 
> Database type isn't in list because its resource is nested. Expanding the parent makes it display:
> 
> 
> Problems with unselecting resource types:
> 
> 
> These are both caused by filtering available options based on visible resources.

Reverted the change.

adamint:
> What happens if there are a lot of states, resources types and health states? The popup should probably have a maximum height and a scoll bar to handle that situation nicely.

Added a max height and scroll bar.
image

adamint:
@JamesNK

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+    private async Task OnValueVisibilityChangedInternalAsync(TValue value, bool isVisible)
+    {
+        if (isVisible)
+        {
+            AllValues[value] = true;
+        }
+        else
+        {
+            AllValues[value] = false;
+        }
```
JamesNK, line 25:
```suggestion
        AllValues[value] = isVisible;
```

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                areAllChecked = false;
+            }
+        }
+
+        return areAllChecked
+            ? true
+            : areAllUnchecked
+                ? false
+                : null;
```
JamesNK, line 69: I don't like nested ternary statements. Make at least one of these an if.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Aspire.Dashboard.Components;
+
+public partial class SelectResourceOptions<TValue>
+{
+    private async Task OnAllValuesCheckedChangedInternalAsync(bool? newAreAllVisible)
+    {
+        SetCheckState(newAreAllVisible, AllValues);
+        await OnAllResourceTypesCheckedChangedAsync();
+    }
```
JamesNK, line 14:
```suggestion
    private async Task OnAllValuesCheckedChangedInternalAsync(bool? newAreAllVisible)
    {
        SetCheckState(newAreAllVisible, Values);
        await OnResourceTypesCheckedChangedAsync();
    }
```
Don't need `All` in many of these names anymore (I think this suggestion is right). All should just be used when working with the all button now.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor
```diff
@@ -0,0 +1,39 @@
...
+    {
+        <FluentCheckbox Label="@(resourceType.ToString())"
+                        @bind-Value:get="@isChecked"
+                        @bind-Value:set="@(c => OnValueVisibilityChangedInternalAsync(resourceType, c))" />
+    }
+</FluentStack>
+
+@code {
+    [Parameter, EditorRequired]
+    public required ConcurrentDictionary<TValue, bool> AllValues { get; set; }
```
JamesNK, line 29:
```suggestion
    public required ConcurrentDictionary<TValue, bool> Values { get; set; }
```

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -57,90 +67,55 @@ public partial class Resources : ComponentBase, IAsyncDisposable
...
-        {
-            _visibleResourceTypes[resourceType] = true;
-        }
-        else
-        {
-            _visibleResourceTypes.TryRemove(resourceType, out _);
-        }
+        return _resourceTypesToVisibility.TryGetValue(resource.ResourceType, out var typeVisible) && typeVisible
+               && _resourceStatesToVisibility.TryGetValue(GetStateOrDefaultText(resource.State, Loc), out var stateVisible) && stateVisible
+                && _resourceHealthStatusesToVisibility.TryGetValue(GetStateOrDefaultText(resource.HealthStatus?.Humanize(), Loc), out var healthStateVisible) && healthStateVisible
```
JamesNK, line 96:
```suggestion
               && _resourceHealthStatusesToVisibility.TryGetValue(GetStateOrDefaultText(resource.HealthStatus?.Humanize(), Loc), out var healthStateVisible) && healthStateVisible
```

src/Aspire.Dashboard/Components/Controls/ResourceFilters.razor
```diff
@@ -0,0 +1,45 @@
+﻿@using System.Collections.Concurrent
+
+@inject IStringLocalizer<Dashboard.Resources.Resources> Loc
+
+<FluentStack Orientation="Orientation.Vertical" VerticalGap="15">
+    <div>
+        <h4>@Loc[nameof(Resources.Resources.ResourcesResourceTypesHeader)]</h4>
```
JamesNK, line 7: Titles feel too big. Make them h5.

src/Aspire.Dashboard/Components/Controls/ResourceFilters.razor
```diff
@@ -0,0 +1,45 @@
...
+            AllValues="AllResourceHealthStates"
+            OnAllResourceTypesCheckedChangedAsync="OnAllFilterVisibilityCheckedChangedAsync"
+            OnValueVisibilityChangedAsync="OnResourceFilterVisibilityChangedAsync"/>
+    </div>
+</FluentStack>
+
+@code {
+
+    [Parameter, EditorRequired]
+    public required ConcurrentDictionary<string, bool> AllResourceTypes { get; set; }
```
JamesNK, line 32:
```suggestion
    public required ConcurrentDictionary<string, bool> ResourceTypes { get; set; }
```
Same for the rest

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -33,36 +32,39 @@
             <FluentDivider slot="end" Role="DividerRole.Presentation" Orientation="Orientation.Vertical" />
             @if (ViewportInformation.IsDesktop)
             {
-                <FluentButton id="typeFilterButton" slot="end"
-                              Appearance="@(AreAllTypesVisible is true ? Appearance.Stealth : Appearance.Accent)"
+                <FluentButton id="resourceFilterButton" slot="end"
+                              Appearance="@(AreAllVisibleInAnyFilter ? Appearance.Stealth : Appearance.Accent)"
```
JamesNK, line 36: This doesn't seem to work anymore.

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -33,36 +32,39 @@
...
                               IconEnd="@(new Icons.Regular.Size20.Filter())"
-                              @onclick="() => _isTypeFilterVisible = !_isTypeFilterVisible"
-                              Title="@(AreAllTypesVisible is true ? Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFilterAllVisible)] : Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFiltered)])"
-                              aria-label="@(AreAllTypesVisible is true ? Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFilterAllVisible)] : Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFiltered)])" />
+                              @onclick="() => _isFilterPopupVisible = !_isFilterPopupVisible"
+                              Title="@(AreAllVisibleInAnyFilter ? Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFilterAllVisible)] : Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFiltered)])"
+                              aria-label="@(AreAllVisibleInAnyFilter ? Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFilterAllVisible)] : Loc[nameof(Dashboard.Resources.Resources.ResourcesTypeFiltered)])"></FluentButton>
```
JamesNK, line 40: The resource text doesn't make sense anymore. It should be made generic to "has filters" vs "no filters"

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -182,20 +157,18 @@ protected override async Task OnInitializedAsync()
...
-
-                _allResourceTypes.TryAdd(resource.ResourceType, true);
-
-                if (preselectedVisibleResourceTypes is null || preselectedVisibleResourceTypes.Contains(resource.ResourceType))
-                {
-                    _visibleResourceTypes.TryAdd(resource.ResourceType, true);
-                }
+                _resourceTypesToVisibility.TryAdd(resource.ResourceType, preselectedVisibleResourceTypes is null || preselectedVisibleResourceTypes.Contains(resource.ResourceType));
+                _resourceStatesToVisibility.TryAdd(GetStateOrDefaultText(resource.State, Loc), preselectedVisibleResourceStates is null || preselectedVisibleResourceStates.Contains(resource.State ?? string.Empty));
+                 _resourceHealthStatusesToVisibility.TryAdd(GetStateOrDefaultText(resource.HealthStatus?.Humanize(), Loc), preselectedVisibleResourceHealthStates is null || preselectedVisibleResourceHealthStates.Contains(resource.HealthStatus?.Humanize() ?? string.Empty));
```
JamesNK, line 171: If there is no value then the key can be an empty string. It shouldn't be translated text.

Translated text should only be applied in the UI, and this isn't UI related code.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
JamesNK: Have a local method for getting true/false from these dictionaries that avoids the need for TryGetValue.

src/Aspire.Dashboard/wwwroot/css/app.css
```diff
@@ -734,3 +734,8 @@ fluent-data-grid-cell.no-ellipsis {
     vertical-align: text-bottom;
     margin-right: 3px
 }
+
+.resources-popup {
```
JamesNK, line 738:
```suggestion
.resources-filter-popup {
```

src/Aspire.Dashboard/wwwroot/css/app.css
```diff
@@ -734,3 +734,8 @@ fluent-data-grid-cell.no-ellipsis {
...
+    max-height: 400px;
+    overflow-y: scroll;
```
JamesNK, line 740: This value always display a scrollbar (disabled when not needed). It should be auto

JamesNK (changes requested):

JamesNK:
I think you should add some component tests to test these changes. We don't tests for resources page yet so you'll need to add some setup code for registering required services and JS calls, but there a lot of other pages already setup that you can follow.

src/Aspire.Dashboard/Components/Controls/SelectResourceOptions.razor.cs
davidfowl: You can flatten them  or use switch expressions 😄

src/Aspire.Dashboard/Components/Pages/Resources.razor
adamint: Fixed

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
adamint: Changed

JamesNK:
Do you intend to write component tests for this?

adamint:
> Do you intend to write component tests for this?

Yes! I added a test to simulate changing a resource filter.

JamesNK (approved):

PR #6946, opened 2024-12-14
Update Npgsql dependency to 9.0

### LINKED ISSUE #6720: Update Npgsql dependency to 9.0
The `Aspire.Npgsql` client integration package depends on the `Npgsql` package. `Npgsql` just shipped a `9.0.0` version on nuget:

https://www.nuget.org/packages/Npgsql/9.0.0

The Npgsql `9.0.0` still supports `net8.0`, so we could unconditionally depend on the `9.0.0` version.

However, the `9.0.0` version depends on System.Text.Json `9.0.0`:

Image

If we referenced this package from our `net8.0` TFM, it would mean any .NET Aspire app using our client integration package would get the v9.0 version of System.Text.Json. This means that `System.Text.Json` would always be "app local" for the app, which has the following drawbacks:

* It is no longer serviced by the shared framework. So when a new vulnerability / patch fix is shipped, every app will need to re-deploy in order to take advantage of the new fix - you can't just update the shared framework for servicing.
    * Note that we default to using containers, so this is less of a concern since you always need to re-deploy your container to get a new shared framework version.
* `System.Text.Json` is no longer "ready to run" AOT'd. Which means the app starts slower because the Json code needs to be JIT'd on startup.
* The app size is larger, since it contains an extra version of `System.Text.Json`.

We should decide how to update our dependency on `Npgsql`. I think we have 2 high-level choices:

1. Split which version we depend on, depending on TFM. This is similar to how we support EntityFramework integration packages.
2. Unconditionally depend on the 9.0 version, and live with the above drawbacks.

@DamianEdwards @joperezr @roji

### LINKED ISSUE #6852: AddKeyedNpgsqlDataSource fails on Aspire 9: MissingMethodException

### Describe the bug

After upgrading from .Net 8 to .Net 9 / Aspire 9, the following line fails:

```
public static WebApplication Build(this WebApplicationBuilder builder)
{
   ...
   builder.AddKeyedNpgsqlDataSource("myPostgresInstance");
   ...
}
```

This is called when starting an asp net core app, that needs to talk to postgres through the keyed npgsql dataSource.

### Expected Behavior

The keyed npgsql dataSource is added to the di-container.

### Steps To Reproduce

1. Create an asp net core app
2. host the app with aspire
3. call builder.AddKeyedNpgsqlDataSource(..)

### Exceptions (if any)

System.MissingMethodException: 'Method not found: 'OpenTelemetry.Trace.TracerProviderBuilder Npgsql.TracerProviderBuilderExtensions.AddNpgsql(OpenTelemetry.Trace.TracerProviderBuilder, System.Action`1<Npgsql.NpgsqlTracingOptions>)'.'

StackTrace:

  ```
at Microsoft.Extensions.Hosting.AspirePostgreSqlNpgsqlExtensions.<>c.<AddNpgsqlDataSource>b__3_1(TracerProviderBuilder tracerProviderBuilder)
   at OpenTelemetry.OpenTelemetryBuilderSdkExtensions.WithTracing(IOpenTelemetryBuilder builder, Action`1 configure)
   at OpenTelemetry.OpenTelemetryBuilder.WithTracing(Action`1 configure)
   at Microsoft.Extensions.Hosting.AspirePostgreSqlNpgsqlExtensions.AddNpgsqlDataSource(IHostApplicationBuilder builder, Action`1 configureSettings, String connectionName, Object serviceKey, Action`1 configureDataSourceBuilder)
   at Microsoft.Extensions.Hosting.AspirePostgreSqlNpgsqlExtensions.AddKeyedNpgsqlDataSource(IHostApplicationBuilder builder, String name, Action`1 configureSettings, Action`1 configureDataSourceBuilder)
   at 
```
...

### .NET Version info

.Net 9

### PR DESCRIPTION
**Pull request by eerhardt**
Npgsql supports net8.0 and net9.0, which means we can just update to the latest version and don't need to multi-target or split dependencies.

Fix #6720
Fix #6852
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6946)

### DIFF
--- a/tests/Shared/WorkloadTesting/BuildEnvironment.cs
+++ b/tests/Shared/WorkloadTesting/BuildEnvironment.cs
@@ -54,16 +54,17 @@ public class BuildEnvironment
     public static BuildEnvironment ForCurrentSdkOnly => s_instance_90.Value;
     public static BuildEnvironment ForCurrentSdkAndPreviousRuntime => s_instance_90_80.Value;
 
-    public static BuildEnvironment ForDefaultFramework { get; } = DefaultTargetFramework switch
-    {
-        TestTargetFramework.Previous => ForPreviousSdkOnly,
+    public static BuildEnvironment ForDefaultFramework =>
+        DefaultTargetFramework switch
+        {
+            TestTargetFramework.Previous => ForPreviousSdkOnly,
 
-        // Use current+previous to allow running tests on helix built with 9.0 sdk
-        // but targeting 8.0 tfm
-        TestTargetFramework.Current => ForCurrentSdkAndPreviousRuntime,
+            // Use current+previous to allow running tests on helix built with 9.0 sdk
+            // but targeting 8.0 tfm
+            TestTargetFramework.Current => ForCurrentSdkAndPreviousRuntime,
 
-        _ => throw new ArgumentOutOfRangeException(nameof(DefaultTargetFramework))
-    };
+            _ => throw new ArgumentOutOfRangeException(nameof(DefaultTargetFramework))
+        };
 
     public BuildEnvironment(bool useSystemDotNet = false, TemplatesCustomHive? templatesCustomHive = default, string sdkDirName = "dotnet-tests")
     {
--- a/tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
+++ b/tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
@@ -18,6 +18,19 @@
 
   <ItemGroup>
     <PackageReference Include="Polly.Core" />
+
+    <!--
+    Since this project uses both Npgsql and Npgsql.EntityFrameworkCore, we need to ensure they are both the same major version.
+    Aspire.Npgsql uses Npgsql v9, so we need to use Npgsql.EntityFrameworkCore v9 (and target net9.0 as well) to match.
+    -->
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Primitives" VersionOverride="9.0.0" />
   </ItemGroup>
 
 </Project>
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by sebastienros — APPROVED** [review]

**Review by roji — APPROVED** [review]

**Comment by eerhardt** [comment]
I needed to update the Npgsql EF library so it continued using Npgsql 8 when using EF 8. See https://github.com/npgsql/npgsql/issues/5987

File: tests/Shared/WorkloadTesting/BuildEnvironment.cs [hunk 1]
```diff
@@ -54,16 +54,17 @@ public class BuildEnvironment
     public static BuildEnvironment ForCurrentSdkOnly => s_instance_90.Value;
     public static BuildEnvironment ForCurrentSdkAndPreviousRuntime => s_instance_90_80.Value;
 
-    public static BuildEnvironment ForDefaultFramework { get; } = DefaultTargetFramework switch
-    {
-        TestTargetFramework.Previous => ForPreviousSdkOnly,
+    public static BuildEnvironment ForDefaultFramework =>
```
eerhardt, line 57: FYI - @radical.

The current code evaluates these Lazy properties in the static initializer. I refactored this so it is only touched when this property is touched.

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
```diff
@@ -18,6 +18,19 @@
 
   <ItemGroup>
     <PackageReference Include="Polly.Core" />
+
+    <!--
+    Since this project uses both Npgsql and Npgsql.EntityFrameworkCore, we need to ensure they are both the same major version.
+    Aspire.Npgsql uses Npgsql v9, so we need to use Npgsql.EntityFrameworkCore v9 (and target net9.0 as well) to match.
```
eerhardt, line 24: This situation is getting a little messy. The major versions of Npgsql and Npgsql.EntityFrameworkCore need to match, or else you run into breaking changes like: https://github.com/dotnet/aspire/issues/6852#issuecomment-2528911374.

With this change, we only target Npgsql v9, no matter the TFM you target. So this needs to use EF Core 9, which is only supported with aspire if you target `net9.0`. So that's what I made this test do.

FYI @joperezr @sebastienros @roji

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
```diff
@@ -18,6 +18,19 @@
...
+    -->
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" VersionOverride="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Primitives" VersionOverride="9.0.0" />
```
eerhardt, line 33: I'm going to send out a separate PR which refactors our Directory.Packages.props so this is no longer necessary.

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
roji: Yeah, makes sense - as I wrote in https://github.com/npgsql/npgsql/issues/5987#issuecomment-2576346615, EFCore.PG 9.0 requires Npgsql 9.0 because of some changes in how enums are managed; but in general, the EF provider and Npgsql sometimes need to interoperate tightly, resulting in major version dependencies. So this is expected - I hope this is OK on your side.

eerhardt: I _think_ it is OK. We just need to ensure we match the major versions (and so will any customers).

roji: FWIW EFCore.PG does specify upper nuget versions for EF itself, since they're also tightly coupled; with Npgsql/EFCore.PG I haven't done that so far... Will consider it for the next release if we have known major breaks.

Issue #4528, opened 2024-06-15
Add PathBaseMiddleware to Aspire Dashboard 


mu88:
Hi 👋🏻

**Disclaimer:** Since my issue is more a feature request than an API change, I decided not to use one of the predefined templates, but I'm happy to provide any further information you need.

I'm running the Aspire Dashboard as a standalone app via Docker which works nicely 👌🏻 Since multiple Dockerized .NET apps are running on my host, I separated them by making use of the [`PathBaseMiddleware`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase). This way, I can easily reroute them like this:

```c#
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();
app.UsePathBase("/myapp");
```

Applying this pattern to the Aspire Dashboard would probably mean adding the following code to [`DashboardWebApplication.cs`](https://github.com/dotnet/aspire/blob/main/src/Aspire.Dashboard/DashboardWebApplication.cs):

```c#
if (!string.IsNullOrEmpty(builder.Configuration[DashboardConfigNames.PathBase]))
{
    app.UsePathBase(builder.Configuration[DashboardConfigNames.PathBase]); // maybe sanitize PathBase here
}
```

This way, I could easily reroute the Aspire Dashboard and make it accessible on `https://myhost/aspire-dashboard` instead of using a dedicated port (e.g. `https://myhost:18888/`). A corresponding `docker-compose.yml` could look like this:

```yaml
services:
  aspire-dashboard:
    image: mcr.microsoft.com/dotnet/aspire-dashboard
    container_name: aspire-dashboard
    ports:
      - 4317:18889
      - 18888:18888
    environment:
      - Dashboard__PathBase=/aspire-dashboard
```

However, I didn't do a comprehensive and all-encompassing review of the implications and whether this might break certain features - but at least I want to trigger a discussion 🙂

mitchdenny:
Do you have a reverse proxy in front?

mu88:
yes, nginx

mitchdenny:
OK thanks for the additional context. @kvenkatrajan not sure where this sits for you in terms of prioritization. It's the first request I've seen for this.

kvenkatrajan:
Thanks @mu88 ..Interesting proposal.

@WardenGnaw / @tmeschter any impact to how this gets launched in codespaces?

CC: @karolz-ms , @drewnoakes

[...]

Issue #4159, opened 2024-05-12
Hosting Dashboard from child path (https://domain/my/child/path) does not appear to work correctly


jimcward:
I am trying to host the aspire dashboard from k8s running on AKS using nginx as an ingress controller and running into some issues. I would essentially like to host the dashboard from:

https://mydomainhere/dev/aspire/dashboard

Using aspirate to generate some basic config I have managed to successfully deploy the dashboard out via helm and can access it from the above URL - it however immediately redirects to ```https://mydomainhere/structuredlogs```

I believe it should be trying to redirect to ```https://mydomainhere/dev/aspire/dashboard/structuredlogs```

I can put that directly into the browser and it then fails due to it trying to load js/css resources from ```https://mydomainhere/```

I presume this might be related to it deploying out with a ```<base href="/">```

I tried to see if setting ```ASPNETCORE_URLS``` as mentioned here https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/configuration#common-configuration might help but that just leads to an error on startup:

```
Hosting failed to start
      System.InvalidOperationException: A path base can only be configured using IApplicationBuilder.UsePathBase()
```

I don't think I can "fix" this issue via ingress config as I have multiple other web apps running so can't just, for example, redirect any traffic to ```https://mydomainhere/``` back out to ```https://mydomainhere/dev/aspire/dashboard```

Is this a scenario that is not currently supported? Any advice would be greatly appreciated, thank you!

PlPt-ads-tec:
it is not only the base href tag, all js sources, rely on hosting direclty on root path (/).
Even when i use IIS AAR (Reverse proxy) and replace base path and link strings, it will show a page, but js interactions are not working well.
Seems that this project is not prepared for hosting on subpaths. Maybe we get support in future as well.
For workaround i will use a separate port for it, directly mapping to aspire. Alternatively you can configure a second port in ingress and redirect it via https to aspire. i know it is annoying but i cant see an easy change.

samsp-msft:
This is an issue for the Blazor aspect of the Dashboard. If it's being hosted at a deeper url than expected, then the paths for loading other resources such as css or json will be wrong if they use absolute URLs.
IIRC there is not a standard way for the proxy to tell the application that there is a different base URL. What we could do would be to support a custom header in the proxy and use that as the `<base href="/">`. If the proxy is then configured to add that header, the dashboard can create absolute URLs that will work correctly.

[...]

PR #6963, opened 2024-12-18
Adding basepath to the config

ewassef:
This will allow the aspire app to handle running behind a reverse proxy such as a k8s ingress on a subpath https://github.com/dotnet/aspire/issues/4159
https://github.com/dotnet/aspire/issues/4528
https://github.com/dotnet/aspire/issues/5134

## Description

Added a new dashboard base path option, integrated it into the Dashboard Option, validated and replaced all areas that are creating the links

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6963)


src/Aspire.Dashboard/Utils/DashboardUrls.cs
```diff
@@ -143,10 +145,12 @@ public static string LoginUrl(string? returnUrl = null, string? token = null)
...
     {
-        var url = "/api/set-language";
+        var url = $"{BasePath}api/set-language";
         url = QueryHelpers.AddQueryString(url, "language", language);
         url = QueryHelpers.AddQueryString(url, "redirectUrl", redirectUrl);
 
         return url;
     }
+
+    internal static void SetBasePath(string pathBase) => BasePath = pathBase;
```
adamint, line 155: This doesn't seem like an optimal way to update `BasePath` to me. If `DashboardUrls` is going to store this state, it should be turned into a service and injected.

adamint (commented):
Please add test coverage in `ValidateDashboardOptions`.

adamint:
@JamesNK for review

ewassef:
> Please add test coverage in `ValidateDashboardOptions`.

Tests added

src/Aspire.Dashboard/Utils/DashboardUrls.cs
ewassef: I am following the `SetLanguageUrl` method above

adamint: It’s different, as SetLanguageUrl does not set state. It only returns the url for the set language url path.

ewassef: OK. Is this something that will block the feature from being accepted?

davidfowl:
The options changes look fine, but this is a one liner in middleware https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-9.0.

You can undo all of the other changes you made.

ewassef:
> The options changes look fine, but this is a one liner in middleware https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-9.0.
> 
> You can undo all of the other changes you made.

@davidfowl , I'm afraid I do not follow. I am using that method in the middleware, but everything else is to handle the options and also the way we are generating the link in the blazor app. The links seem to be string interpolation from the base before

adamint:
> > The options changes look fine, but this is a one liner in middleware https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-9.0.
> > You can undo all of the other changes you made.
> 
> @davidfowl , I'm afraid I do not follow. I am using that method in the middleware, but everything else is to handle the options and also the way we are generating the link in the blazor app. The links seem to be string interpolation from the base before

I see what fowler means now. We don't need to store path base inside `DashboardUrls`. You do however need to provide a `<base>` tag in blazor since the base path has been changing.

> the way we are generating the link 

The path base is provided in the dashboard options already. You can use that when writing out the dashboard url.

src/Aspire.Dashboard/Utils/DashboardUrls.cs
adamint: See the comment below, DashboardUrls shouldn't be changed at all

davidfowl:
There should be a single call to the `UsePathBase`  middleware in the code base. You can undo all of the other changes that were made to preprend the path base into various places in the application.

adamint:
> There should be a single call to the `UsePathBase` middleware in the code base. You can undo all of the other changes that were made to prepend the path base into various places in the application.

We need to change the blazor path base as well.

davidfowl:
Why? Also UsePath base shoul be the first middleware (order matters). We should be able to remove the prefix in UseErrorHandler as well.

JamesNK:
Is blazor loading something from a rooted path? Eg `/blazor`.

Can it be changed to a relative path? `./blazor`

JamesNK:
https://learn.microsoft.com/en-nz/aspnet/core/blazor/host-and-deploy/?view=aspnetcore-8.0&tabs=visual-studio#app-base-path

davidfowl:
@JamesNK @adamint have you tried this out?

ewassef:
Im still unclear on the ask. I started with just the one liner, but it doesnt work as expected. we need to handle the url creation helpers in all the places that ive modified. The pipelines keep cycling and im not sure whats not actually working from this code?

davidfowl:
UsePathBase should do most of the heavy lifting. I will take a stab at this and see why that simple one line change doesn’t work.

ewassef:
Thanks @davidfowl . we really just want to install this as a subpath in a k8s  ingress :) 
To save you some time, you can take the first commit from the 5 that are there and run your tests

ewassef:
Any progress?

WeihanLi:
Is it possible to use the forwarded header for it, ASP.NET supports the `X-Forwarded-Prefix` since .NET 8

Maybe configure the ForwardedHeadersOptions is more simple to implement

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.forwardedheadersoptions.forwardedheaders?view=aspnetcore-9.0#microsoft-aspnetcore-builder-forwardedheadersoptions-forwardedheaders

https://github.com/dotnet/aspnetcore/issues/23263

davidfowl:
cc @DamianEdwards @javiercn

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -12,6 +12,7 @@ namespace Aspire.Dashboard.Configuration;
 public sealed class DashboardOptions
 {
     public string? ApplicationName { get; set; }
+    public string PathBase { get; set; } = "/";
```
adamint, line 15: I think this should be an empty string

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -144,6 +144,21 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
             }
         }
 
+        // Validate the Path base and make sure it starts and ends with a forward slash.
+
+        if (options.PathBase != null)
+        {
+            if (!options.PathBase.StartsWith("/", StringComparison.Ordinal))
```
adamint, line 151: We have a `StringComparisons` class. Please use or add a comparison from there

src/Aspire.Dashboard/DashboardEndpointsBuilder.cs
```diff
@@ -18,28 +18,28 @@ public static void MapDashboardApi(this IEndpointRouteBuilder endpoints, Dashboa
     {
         if (dashboardOptions.Frontend.AuthMode == FrontendAuthMode.BrowserToken)
         {
-            endpoints.MapPost("/api/validatetoken", async (string token, HttpContext httpContext, IOptionsMonitor<DashboardOptions> dashboardOptions) =>
+            endpoints.MapPost($"{DashboardUrls.BasePath}api/validatetoken", async (string token, HttpContext httpContext, IOptionsMonitor<DashboardOptions> dashboardOptions) =>
```
adamint, line 21: I don't think these are necessary if you just set the app path base

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -172,6 +172,7 @@ public DashboardWebApplication(
         }
         else
         {
+             DashboardUrls.SetBasePath(dashboardOptions.PathBase);
```
adamint, line 175: dashboardurl changes should be reverted

adamint (changes requested):

DamianEdwards:
I have changes to this PR staged locally which I plan to push later today that address the issues raised. However after updating from main I'm seeing a new issue which I'm having some difficultly tracking down. There's been a lot of changes since this PR so there's a bit to get through but I have some hunches. Pls hold tight 😃

playground/TestShop/TestShop.AppHost/HostingExtensions.cs
```diff
@@ -0,0 +1,56 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace TestShop.AppHost;
+
+internal static class HostingExtensions
+{
+    /// <summary>
+    /// Adds a rference to the dashboard resource so that the required environment variables to allow service discovery to the Aspire dashboard
+    /// are added to this resource. Also configures a reference relationship to the dashboard resource.
```
JamesNK, line 10: Needs proof reading

playground/TestShop/TestShop.AppHost/HostingExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            throw new ArgumentException($"The dashboard resource must be named '{KnownResourceNames.AspireDashboard}'.", nameof(dashboard));
+        }
+
+        var dashboardUris = builder.ApplicationBuilder.Configuration["ASPNETCORE_URLS"]?.Split(';')
+            ?.Select(u => Uri.TryCreate(u, UriKind.Absolute, out var uri) is { } ? uri : null)
+            ?.Where(u => u is not null)
+            ?.Cast<Uri>()
+            ?.GroupBy(u => u.Scheme)
+            ?? [];
```
JamesNK, line 25:
```suggestion
        var dashboardUris = builder.ApplicationBuilder.Configuration["ASPNETCORE_URLS"]?.Split(';')
            .Select(u => Uri.TryCreate(u, UriKind.Absolute, out var uri) is { } ? uri : null)
            .OfType<Uri>()
            .GroupBy(u => u.Scheme)
            ?? [];
```

playground/TestShop/TestShop.AppHost/HostingExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+            builder
+                .WithEnvironment("DASHBOARD_PATH", proxyPath)
+                .WithUrls(context =>
+                {
+                    var https = context.Urls.FirstOrDefault(u => u.Endpoint?.EndpointName == "https");
+                    var http = context.Urls.FirstOrDefault(u => u.Endpoint?.EndpointName == "http");
+                    if (https is not null || http is not null)
+                    {
+                        var endpoint = https ?? http!;
+                        context.Urls.Add(new() { Url = $"{endpoint.Url}{proxyPath}", DisplayText = "Proxied Dashboard" });
```
JamesNK, line 47: A string concat of URL + path could mean there is no forward slash, or multiple forward slashes in the new URL. Use a better way to combine them?

playground/TestShop/TestShop.AppHost/Properties/launchSettings.json
```diff
@@ -26,7 +27,8 @@
         "ASPIRE_DASHBOARD_OTLP_HTTP_ENDPOINT_URL": "http://localhost:16032",
         "ASPIRE_RESOURCE_SERVICE_ENDPOINT_URL": "http://localhost:17031",
         "ASPIRE_SHOW_DASHBOARD_RESOURCES": "true",
-        "ASPIRE_ALLOW_UNSECURED_TRANSPORT": "true"
+        "ASPIRE_ALLOW_UNSECURED_TRANSPORT": "true",
+        "Dashboard__ReverseProxy__ForwardHeaders": "true"
```
JamesNK, line 31: This is probably answered in later code, but how does this env var to the host make its way to the dashboard?

src/Aspire.Dashboard/Components/App.razor
```diff
@@ -39,3 +42,22 @@
     <script src="_content/Microsoft.FluentUI.AspNetCore.Components/Microsoft.FluentUI.AspNetCore.Components.lib.module.js" type="module"></script>
 </body>
 </html>
+
+@code {
+    private string pathBase = "/";
```
JamesNK, line 47:
```suggestion
    private string _pathBase = "/";
```

src/Aspire.Dashboard/Components/App.razor
```diff
@@ -39,3 +42,22 @@
     <script src="_content/Microsoft.FluentUI.AspNetCore.Components/Microsoft.FluentUI.AspNetCore.Components.lib.module.js" type="module"></script>
 </body>
 </html>
+
+@code {
```
JamesNK, line 46: Move this to a code behind file. Refactoring and code analyzers work better with code behind cs vs cs in the razor file.

src/Aspire.Dashboard/Components/Pages/IPageWithSessionAndUrlState.cs
```diff
@@ -92,15 +93,16 @@ async Task SetStateAndNavigateAsync()
     /// </returns>
     public static async Task<bool> InitializeViewModelAsync<TViewModel, TSerializableViewModel>(this IPageWithSessionAndUrlState<TViewModel, TSerializableViewModel> page) where TSerializableViewModel : class
     {
-        if (string.Equals(page.BasePath, page.NavigationManager.ToBaseRelativePath(page.NavigationManager.Uri)))
+        var baseRelativePath = page.NavigationManager.ToBaseRelativePath(page.NavigationManager.Uri);
+        if (string.Equals(page.BasePath, baseRelativePath))
```
JamesNK, line 97: OrdinalIgnoreCase?

src/Aspire.Dashboard/Components/Pages/Login.razor.js
```diff
@@ -1,6 +1,6 @@
 export async function validateToken(token) {
     try {
-        var url = `/api/validatetoken?token=${encodeURIComponent(token)}`;
+        var url = `api/validatetoken?token=${encodeURIComponent(token)}`;
```
JamesNK, line 3: We're doing `./xxx` everywhere else. Should this have no slash at the start or `./api/...` to be consistant?

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -144,6 +144,29 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
             }
         }
 
+        if (!options.ReverseProxy.ForwardHeaders)
+        {
+            // TODO: Ensure other properties are not set if ForwardHeaders is false.
+            if (options.ReverseProxy.ForwardLimit != 1)
+            {
+                
+            }
```
JamesNK, line 153: TODO

src/Aspire.Dashboard/Configuration/ValidateDashboardOptions.cs
```diff
@@ -144,6 +144,29 @@ public ValidateOptionsResult Validate(string? name, DashboardOptions options)
...
+        }
+
+        // Validate the path base if set to make sure it starts and ends with a forward slash.
+        if (options.PathBase is not null)
+        {
+            if (!options.PathBase.StartsWith('/'))
+            {
+                errorMessages.Add($"{nameof(DashboardOptions.PathBase)} must start with a forward slash.");
```
JamesNK, line 161: Shouldn't the name here come from a value on `DashboardConfigNames` so the complete path of the option is in the error message?

src/Aspire.Dashboard/DashboardEndpointsBuilder.cs
```diff
@@ -3,7 +3,7 @@
 
 using Aspire.Dashboard.Configuration;
 using Aspire.Dashboard.Model;
-using Aspire.Dashboard.Utils;
+using Aspire.Dashboard.Utils;
```
JamesNK, line 6:
```suggestion
using Aspire.Dashboard.Utils;
```

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -284,6 +294,27 @@ public DashboardWebApplication(
...
+        //       e.g. when not using a reverse-proxy but instead just configuring an app host project to use a path base.
+        if (dashboardOptions.PathBase is not null)
+        {
+            _app.UsePathBase(dashboardOptions.PathBase);
+
+            if (_app.Environment.IsDevelopment())
+            {
+                // In development we want to ensure that requests outside of the path base return a 404 to mimic what would
+                // happen when the dashboard is being served from behind a reverse-proxy at a configured path. Note we still let
+                // ~/login requests through as that's what VS launches the browser to. In that case we redirect instead.
```
JamesNK, line 307: Need intergration tests of this:

Base path + good path = 200
Base path + bad path = 404
Base path + /login = 302

src/Aspire.Dashboard/Utils/DashboardUrls.cs
```diff
@@ -9,18 +9,35 @@ namespace Aspire.Dashboard.Utils;
...
+    {
+        if (baseRelativePath == targetPagePath)
+        {
+            return true;
+        }
+
+        var queryIndex = baseRelativePath.IndexOf('?');
+        var testPath = queryIndex >= 0 ? baseRelativePath[..queryIndex] : baseRelativePath;
+        var isMatch = testPath.TrimEnd('/') == targetPagePath;
+        return isMatch;
```
JamesNK, line 27: Path comparisons here should ignore case and culture, e.g. OrdinalIgnoreCase

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -36,12 +36,30 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
...
             }
 
             var dashboardOtlpGrpcUrl = configuration.GetString(KnownConfigNames.DashboardOtlpGrpcEndpointUrl, KnownConfigNames.Legacy.DashboardOtlpGrpcEndpointUrl);
             var dashboardOtlpHttpUrl = configuration.GetString(KnownConfigNames.DashboardOtlpHttpEndpointUrl, KnownConfigNames.Legacy.DashboardOtlpHttpEndpointUrl);
+            var dashboardPathBase = configuration[KnownConfigNames.DashboardPathBase];
+
+            if (!string.IsNullOrEmpty(dashboardPathBase))
+            {
+                // Ensure OTLP URLs are rewritten to include the path base.
+                if (Uri.TryCreate(dashboardOtlpGrpcUrl, UriKind.Absolute, out var otlpGrpcUri) && !otlpGrpcUri.AbsolutePath.StartsWith(dashboardPathBase))
```
JamesNK, line 50: OrdinalIgnoreCase?

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -36,12 +36,30 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
...
+                {
+                    var uriBuilder = new UriBuilder(otlpGrpcUri);
+                    uriBuilder.Path = dashboardPathBase + uriBuilder.Path.TrimStart('/');
+                    dashboardOtlpGrpcUrl = uriBuilder.Uri.ToString();
+                }
+                if (Uri.TryCreate(dashboardOtlpHttpUrl, UriKind.Absolute, out var otlpHttpUri) && !otlpHttpUri.AbsolutePath.StartsWith(dashboardPathBase))
```
JamesNK, line 56: OrdinalIgnoreCase?

JamesNK (commented):
I haven't tested the branch, just looked at code. My knowledge of the things in this PR (running an ASP.NET Core with a base path, running behind a proxy, forwarding headers) is low. An expert should take another look.

I mentioned coming URLs with OrdinalIgnoreCase a number of times. FYI we have StringComparers/StringComparisons with known comparison types. One already might be there for URLs. If not then you could add one.

playground/TestShop/TestShop.AppHost/HostingExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        if (dashboard.Resource.Name != KnownResourceNames.AspireDashboard)
+        {
+            throw new ArgumentException($"The dashboard resource must be named '{KnownResourceNames.AspireDashboard}'.", nameof(dashboard));
+        }
+
+        var dashboardUris = builder.ApplicationBuilder.Configuration["ASPNETCORE_URLS"]?.Split(';')
+            ?.Select(u => Uri.TryCreate(u, UriKind.Absolute, out var uri) is { } ? uri : null)
+            ?.Where(u => u is not null)
+            ?.Cast<Uri>()
+            ?.GroupBy(u => u.Scheme)
```
JamesNK, line 24: I think scheme is always lower case. Maybe `ToLowerInvariant()` in case it isn't and someone does:

```
ASPNETCORE_URLS=https://localhost:80;HTTPS://LOCALHOST:8080
```

JamesNK:
This is going to require a lot of new tests. For example, does all the existing authentication functionality work properly when running with a base path.

DamianEdwards:
> This is going to require a lot of new tests. For example, does all the existing authentication functionality work properly when running with a base path.

@JamesNK yep for sure. This is now functionally complete more or less and the next step is to figure out a testing strategy.

src/Aspire.Dashboard/Components/Pages/Login.razor.js
DamianEdwards: If should likely be `./api/...` assuming the JS module loading resolution still works. I can't remember if I tried that. Note the module reference paths from .js files was one of the trickiest things to figure out but I'll try it here.

danmoseley:
@ewassef do you plan to continue with this PR?

ewassef:
> @ewassef do you plan to continue with this PR?

Are you serious? All the maintainers did their best to discourage contributions,  said they would fix it and then let it rot.  Most of the people i know have forked this and run it themselves. 

I would Love to not have to do that,  but the style police is stopping everything. So I'm waiting for them to fulfill their commitment ( see thread )

DamianEdwards:
@ewassef apologies for confusion. My intent is to complete this PR, it's just a priority and sequencing question right now. No asks on you right now.

martinyyyy:
Is this still not possible with the dashboard? This is quite disappointing.

davidfowl:
Originally, we though this would be a simple change but it turned out to make it really complete was a yak shaving exercise and we lacked the right testing infrastructure to find all of the places that needed to change to make this work for realz. I will close this PR but we can take another stab at this now that we have a bit more experience with the problems and have ways to verify it.

PR #6969, opened 2024-12-19
Updated Dashboard to accept/map forwarded headers

danespinosa:
## Description

This change addresses the need to map the forwarded headers in the application when ASPNETCORE_FORWARDEDHEADERS_ENABLED is enabled. Currently even if a developer enables the ASPNETCORE_FORWARDEDHEADERS_ENABLED environment variable, the application does not map the forwarded headers properly because the application doesn't specify which headers to map.

Without this change the dashboard doesn't work well behind a reverse proxy like YARP when doing OpenID Auth since the app doesn't map the Host or Protocol (http/https) properly and the redirect ends up being the address YARP redirects the call to.

I need guidance for the test I was thinking calling the dashboard and do something similar to `ValidateTokenMiddlewareTests` to validate the `Host`,  and `Proto`.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes, it assumes the developer knows that forwarding headers and mapping them is required for the dashboard to work properly when behind a reverse proxy like YARP. This in certain scenarios could pose a security risk, although it's documented that this dashboard shouldn't be used in production environments.
      - [ ] Yes
      - [x] No
  - [ ] No
  - [x] Yes
      - [ ] Yes
      - [x] No, if this change is accepted it would be good to clarify that the only 2 forwarded headers allowed are the Host and Protocol in here [Configuration](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/configuration?tabs=bash)
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6969)


adamint:
@JamesNK @drewnoakes 
I'm not sure whether there are any security implications here.

davidfowl:
I’m that doesn’t seem right. Do you need to forward all headers?

danespinosa:
> I’m that doesn’t seem right. Do you need to forward all headers?
@davidfowl I just need `XForwardedHost`/`X-Forwarded-Host` that maps to host, that unblocks me to use Microsoft as an Identity provider, I wouldn't know if we need any other headers for other OpenId identity provider though, the other option would be to expose the ForwardedHeaderOptions via the config and let the users decide what headers to forward?

danespinosa:
> I’m that doesn’t seem right. Do you need to forward all headers?

@adamint  @davidfowl  i have limited the forwarded headers to host and protocol, let me know your thoughts.

adamint:
> X-Forwarded-Host

I'm still unsure about the security implications of this change. Someone with a better understanding (maybe @drewnoakes) needs to weigh in, but this approach still seems insecure. Why should we implicitly be trusting X-Forwarded-Host?

danespinosa:
Your concerns are valid @adamint, that's why the forwarded headers shouldn't be enabled by default, and should be enabled only by a person that knows how they work with a reverse proxy and why they are needed for their specific scenario, from that perspective this feature is secure by default. In my opinion is that if someone uses this feature, they have some common and basic knowledge like: https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-9.0 and https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/host-filtering?view=aspnetcore-9.0 . 

Like any other feature of .NET or Aspire, this can be used by any bad actor, disabling it by default is a must but for advanced or specific scenarios allowing users to leverage it allows application developers to focus on their business logic while leveraging Aspire's great focus on telemetry. From my perspective and security-wise this feature is no different than allowing an unsecure transport (http) for the dashboard, which is currently supported, but discouraged, and should be used at the users' discretion.

danespinosa:
@adamint answering your comment `Would it change anything to compare the host against an 'expected' host regex? From my perspective, using an identity provider should be a supported but also secure-by-default scenario.`

Currently I am achieving this by only allowing only the hosts that I am expecting my reverse proxy to call e.g. `10.0.0.1` i do this by setting the `AllowedHosts` setting in the appSettings.json to the ip `10.0.0.1`

adamint:
Ok thanks for the responses, @danespinosa - and pardon my ignorance. I'm satisfied with this then, @davidfowl?

danespinosa:
Any additional input @davidfowl  @JamesNK  @drewnoakes? I would really appreciate it! Thanks a lot and happy new year!

danmoseley:
> Any additional input @davidfowl @JamesNK @drewnoakes? I would really appreciate it! Thanks a lot and happy new year!

any more feedback on this one?

davidfowl:
cc @Tratcher gut check please.

Tratcher:
This is an incomplete copy of the logic from the default host. It won't work in real environments without setting or clearing KnownNetworks/Proxies, which is where your security issues really come into play. Staying consistent with default host should be ok, it's still on the admin to know if setting ForwardedHeaders_Enabled is appropriate or not.
https://github.com/dotnet/aspnetcore/blob/6d1573a1aec23de85faca56c3550adb56f60a543/src/DefaultBuilder/src/ForwardedHeadersOptionsSetup.cs#L20-L31

danespinosa:
> This is an incomplete copy of the logic from the default host. It won't work in real environments without setting or clearing KnownNetworks/Proxies, which is where your security issues really come into play. Staying consistent with default host should be ok, it's still on the admin to know if setting ForwardedHeaders_Enabled is appropriate or not. https://github.com/dotnet/aspnetcore/blob/6d1573a1aec23de85faca56c3550adb56f60a543/src/DefaultBuilder/src/ForwardedHeadersOptionsSetup.cs#L20-L31

@danmoseley  @davidfowl  thanks for helping getting traction on this PR.

@Tratcher based on your comment am I understanding right that I should add the lines that also clear the known networks/proxies from the options? Is there any comment about forwarding host? the code that you shared forwards the for header too should I add that too? The changes look like below.



``` C#
        options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedHost | ForwardedHeaders.XForwardedProto;
        // Only loopback proxies are allowed by default. Clear that restriction because forwarders are
        // being enabled by explicit configuration.
        options.KnownNetworks.Clear();
        options.KnownProxies.Clear();
```

danespinosa:
> > This is an incomplete copy of the logic from the default host. It won't work in real environments without setting or clearing KnownNetworks/Proxies, which is where your security issues really come into play. Staying consistent with default host should be ok, it's still on the admin to know if setting ForwardedHeaders_Enabled is appropriate or not. https://github.com/dotnet/aspnetcore/blob/6d1573a1aec23de85faca56c3550adb56f60a543/src/DefaultBuilder/src/ForwardedHeadersOptionsSetup.cs#L20-L31
> 
> @danmoseley @davidfowl thanks for helping getting traction on this PR.
> 
> @Tratcher based on your comment am I understanding right that I should add the lines that also clear the known networks/proxies from the options? Is there any comment about forwarding host? the code that you shared forwards the for header too should I add that too? The changes look like below.
> 
> ```cs
>         options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedHost | ForwardedHeaders.XForwardedProto;
>         // Only loopback proxies are allowed by default. Clear that restriction because forwarders are
>         // being enabled by explicit configuration.
>         options.KnownNetworks.Clear();
>         options.KnownProxies.Clear();
> ```

I have tested the changes mentioned in my previous comment and the dashboard and dashboard auth is working great behind yarp

image

danespinosa:
I have pushed the suggested changes by @Tratcher. 

Please let me know if I can do anything else to improve the PR.

DamianEdwards:
FYI I'm incorporating these changes into the work I'm doing on PR #6963 to fully enable the dashboard running behind a reverse proxy, including at a sub-path if desired.

danespinosa:
> FYI I'm incorporating these changes into the work I'm doing on PR #6963 to fully enable the dashboard running behind a reverse proxy, including at a sub-path if desired.

Thanks @DamianEdwards let me know if I can help in anyway or if I should abandon my PR really appreciate you bringing these changes into aspire 🥳

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -267,6 +268,19 @@ public DashboardWebApplication(
             options.Cookie.Name = DashboardAntiForgeryCookieName;
         });
 
+        if (builder.Configuration.GetBool("ASPNETCORE_FORWARDEDHEADERS_ENABLED") ?? false)
+        {
+            builder.Services.Configure<ForwardedHeadersOptions>(options =>
+            {
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedHost | ForwardedHeaders.XForwardedProto;
```
WeihanLi, line 275: What about the `Prefix`, could we use `All` or parse from the configuration? Think maybe better support from aspnetcore framework

DamianEdwards:
We can close this one as I'm doing this work over in the other one

davidfowl:
Feels like we should decouple this from the other PR.

DamianEdwards:
>Feels like we should decouple this from the other PR.

Why's that? It's part of the changes required to make the dashboard work properly behind a reverse proxy. Are you suggesting we separate the changes for:

- Make it work behind a reverse proxy where there's no sub-path/path-base
- Make it work from a sub-path/path-base

davidfowl:
Make XFF work with the dashboard irrespective of using base path as a routing mechanism. I was trying to route to the dashboard behind yarp using host name routing and it didn't work. I realized that we needed this to even make that work.

DamianEdwards:
Got it, and agreed. Supporting sub-paths is a BIG change, but ensuring the dashboard works well behind a reverse proxy with regular host-based routing should be much simpler.

davidfowl:
@danespinosa can you open up another PR?

Issue #6807, opened 2024-11-26
Learnings from the Community Toolkit


aaronpowell:
It's been a few months since we started building the [.NET Aspire Community Toolkit](https://github.com/CommunityToolkit/Aspire) and I thought I'd share some things that we've learnt trying to build integrations outside of the Aspire repo. When it came to building integrations, I wanted to have them as close to the same level of quality that can be found in the Aspire repo, across API design, documentation, and testing, while also behaving as close as we could make them to integrations from the Aspire repo (I'll refer to these as _native integrations_).

## Integration design

When it comes to creating integrations, especially at the level of quality we wanted to achieve, Aspire provides a lot of the groundwork without complexity. Types are generic enough and there's very few `sealed` types, meaning creating custom `Resource` types is easy.

This brings us to the primary challenge as an integration author - consistent defaults with native integrations.

### Util/Shared classes

There are a several classes that live in the [`Shared`](https://github.com/dotnet/aspire/tree/main/src/Shared) and [`Components/Common`](https://github.com/dotnet/aspire/tree/main/src/Components/Common) folders that provide behaviour to integrations. So far, we have wanted to use four of them, `HealthCheckExtensions`, `PasswordGenerator`, `PathNormalizer`, and `VolumeNameGenerator`.

As these types don't belong to assemblies themselves, and are marked as `internal`, we are left with two options:

1. Disregard that functionality.
2. Copy the code into our repo.

So far, we've gone down the route of option two and copied the code across, but this represents a point-in-time snapshot of how the code works and can result in code drift. For example, I noticed while writing this up that our version of `VolumeNameGenerator` is out of date and thus we are not generating volume names consistently with native integrations.

Also, we are looking into how our client integrations can better support the configuration schema (tracking: https://github.com/CommunityToolkit/Aspire/issues/112) and that would require bringing in the `ConfigurationSchemaAttribute` type (as well as the generating tool which has its own issue #3309).

#### Opportunity

There is an opportunity here in which a standalone NuGet package could be created that provides this functionality, or potentially split as two packages, one for client and one for hosting, specifically targeted to integration authors and not general Aspire use (it'd would be a transient dependency though).

As an integration author, I would add a reference to `Aspire.Hosting.Utils` or `Aspire.Client.Utils` (or whatever they are named, naming is hard!), and then I could use `VolumeNameGenerator.CreateVolumeName` and be confident that the names generated in my integration for volumes is consistent with, say, the Postgres hosting integration.

[...]

PR #6996, opened 2024-12-30
Making PathNormalizer a public class from Aspire.Hosting.Utils

aaronpowell:
## Description

When creating an integration that works with the local file system, it'd be ideal to use the `PathNormalizer` utility class so that custom integrations can behave consistently with Aspire "core" integrations. Currently, this class is internal and the only workaround is to copy it into your own codebase - which is not ideal.

This PR makes the class public, documents it, adds it to the unshipped API file, and removes the shared include from integrations that relied on it.

Contributes to #6807

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [ ] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/6996)

### DIFF
--- a/src/Aspire.Hosting.NodeJs/Aspire.Hosting.NodeJs.csproj
+++ b/src/Aspire.Hosting.NodeJs/Aspire.Hosting.NodeJs.csproj
@@ -12,10 +12,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="$(SharedDir)PathNormalizer.cs" Link="Utils\PathNormalizer.cs" />
-  </ItemGroup>
-
-  <ItemGroup>
     <ProjectReference Include="..\Aspire.Hosting\Aspire.Hosting.csproj" />
   </ItemGroup>
 
--- a/src/Aspire.Hosting.Python/Aspire.Hosting.Python.csproj
+++ b/src/Aspire.Hosting.Python/Aspire.Hosting.Python.csproj
@@ -9,10 +9,6 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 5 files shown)


davidfowl:
err, do we need to make this public? It's like 4 lines of code.

aaronpowell:
I think it's better that integrations don't have to copy code out of the Aspire repo to have consistent behaviour.

Yes, it's four lines of code, but it's four LOC here, a handful there, and you keep having to come back to grab stuff.

davidfowl:
I can’t approve this one, doesn’t feel like something we need to expose. Use shared source in the community toolkit.

eerhardt:
Is there a way we can build a source-only NuGet package in dotnet/aspire that includes a bunch of random `internal` helper classes.

Then in the community toolkit, we extract that NuGet package into the individual `.cs` files, and each community toolkit library can pick and choose individual files they want to include? Then the source wouldn't need to be duplicated, and when we updated these internal helper classes, the toolkit could update to the new version as well.

davidfowl:
> Is there a way we can build a source-only NuGet package in dotnet/aspire that includes a bunch of random internal helper classes.

Yes, we should build a source package

danmoseley:
OK closing in favor of https://github.com/dotnet/aspire/issues/7375.

PR #7021, opened 2025-01-04
Grafana: improved aspnetcore.json

Aaronontheweb:
## Description

I'm going to do a self-review with some screenshots because many of the changes made to the Grafana dashboard here can't be appreciated without the data being visualized, but this PR contains several quality of life improvements for the default Grafana dashboards for ASP.NET Core + Prometheus, such as:

1. Rather than displaying absolute raw counter totals, we display the delta between the beginning / end of the time range
2. Include an `All` option for both `job` and `instance` so it's easier to aggregate metrics by application / across several applications.
3. Changed several metrics to use the time-range, rather than the Grafana refresh interval for calculations since the former is _usually_ what people care about.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/7021)


src/Grafana/dashboards/aspnetcore.json
```diff
@@ -198,7 +198,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "histogram_quantile(0.50, sum(rate(http_server_request_duration_seconds_bucket{job=\"$job\", instance=\"$instance\"}[$__rate_interval])) by (le))",
+          "expr": "histogram_quantile(0.50, sum(rate(http_server_request_duration_seconds_bucket{job=~\"$job\", instance=~\"$instance\"}[$__rate_interval])) by (le))",
```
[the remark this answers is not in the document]
Aaronontheweb: So this chart is the 99% latency chart, and addition to changing it to support multi-select I've also modified it to use the `$__range` value in Grafana, which corresponds to the tailing time window, versus the `$__rate__interval` variable, which is just 4x the Prometheus scraping interval: https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/

It's useful for rate queries, but that's not what we're measuring here - we're instead trying to determine "what was 99% latency over X period of time?" - the `$__range` value is better for that, and I've made that change in several charts.

Let's do a before and after comparison for this chart specifically: same app and time-range, but only a single instance:

## Before

image

## This PR

image

This is probably mostly a taste thing, but what I appreciate about the latter chart are the following:

1. It's possible to "scope in" using cursor selection in Grafana if I want to look at a specific time period on the chart. With the original chart I get "no data" in many scenarios.
2. This chart design scales down to smaller time-scales for applications that have less request data - this application does ~1m requests per day and I still have zero data using the default chart design unless I zoom out to a 24 hour time window or so.

The only drawback of this second design is that it's harder to see really large outliers - that's easier to see with the original chart design. This design is averaging rates over a longer period of time, which is what makes it more reliable at showing data over smaller time intervals - you can still see where spikes occur, such as this instance here:

image

But it's not nearly as pronounced. Happy to take feedback on what's more appropriate here, but making sure the chart worked correctly across a wider range of traffic workloads / time ranges was my objective here.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -424,7 +425,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum(rate(http_server_request_duration_seconds_bucket{job=\"$job\", instance=\"$instance\", http_response_status_code=~\"4..|5..\"}[$__rate_interval]) or vector(0)) / sum(rate(http_server_request_duration_seconds_bucket{job=\"$job\", instance=\"$instance\"}[$__rate_interval]))",
```
Aaronontheweb, line 427: This is the error rate chart, which had some of the same problems that the Requests Duration chart did:

1. Not showing data over many intervals
2. Data it showed didn't always make a lot of sense.

## Before

image

I actually don't know what to make of those error rates - 0% last and 97.5% max? What does that mean?

## This PR

We've changed the error rate chart to use the `$__range` and now the story it's telling is much clearer:

image


I have a roughly 70% 400-level error rate over the past 24 hours - and that makes sense with what I know about this application: it runs a private NuGet package endpoint and we get a lot of 404s from clients looking for packages that this server doesn't host, as a function of how NuGet clients are designed.

If I scope out to the past 6 hours, which mostly covers Saturdays when our customers aren't working:

image


That 404 rate scales down, with the rest of our traffic, to roughly 11%. It's now easier to tell that by looking at the charts than before.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -413,6 +413,7 @@
           "showLegend": true
         },
         "tooltip": {
+          "maxHeight": 600,
```
Aaronontheweb, line 416: This is probably junk and I can remove that - stuff Grafana adds when doing a JSON export

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -543,8 +545,8 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum(kestrel_active_connections{job=\"$job\", instance=\"$instance\"})",
-          "legendFormat": "__auto",
+          "expr": "sum(kestrel_active_connections{job=~\"$job\", instance=~\"$instance\"})",
+          "legendFormat": "active connections",
```
Aaronontheweb, line 549: Changed the legend to use a manual entry because otherwise, now that we can aggregrate Kestrel data across multiple applications, you end up with a very gross and verbose-looking auto-generated tooltip here on the legend.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -638,8 +641,8 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum(http_server_active_requests{job=\"$job\", instance=\"$instance\"})",
-          "legendFormat": "__auto",
+          "expr": "sum(http_server_active_requests{job=~\"$job\", instance=~\"$instance\"})",
+          "legendFormat": "active requests",
```
Aaronontheweb, line 645: Same deal as with the kestrel connections - manually renamed the legend here.

I toyed with the idea of breaking out active connections by service / instance and that's easily doable, but opted for something simpler - the user can get that same data by changing the template variables on the selector.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -707,7 +712,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\"})",
+          "expr": "sum(increase(http_server_request_duration_seconds_count{job=~\"$job\", instance=~\"$instance\"}[$__range]))",
```
Aaronontheweb, line 715: This is the "Total Requests" counter - and the way it's currently working is just counting the total value over all time, which means that regardless of what time range you pick the value doesn't go down.

## Before

Case in point - this is my "24 hour" view

image

This is my "past 6 hours" view with the current chart - I took this screenshot 2nd and it displays a higher value

image

## This PR

So we've made two changes, other than the usual bits with the `=~` selector for multi-value support:

* Use the `increase` function in PromQL to compute the difference between `[time range start, time range end]` in our chart and
* Use the `[$__range]` value to scope the `increase` function to specific points in time.

Now we get what we're really interested in: the total requests that occurred during the time period:

Past 24 hours

image

Past 6 hours

image

That's more like it!

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -777,7 +784,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\", error_type!=\"\"})",
+          "expr": "sum(increase(http_server_request_duration_seconds_count{job=~\"$job\", instance=~\"$instance\", error_type!=\"\"}[$__range]))",
```
Aaronontheweb, line 787: This is the Unhandled Exceptions chart - I made the same changes to it as I did on the "Total Requests" chart.

## Before

Over the last 6 hours

image

## This PR

Over the last 6 hours

image

This is accurate - per my traces and logs, there haven't been any exceptions observed for this application over the past 6 hours.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -845,7 +854,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum by (url_scheme) (\r\n    max_over_time(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\"}[$__rate_interval])\r\n  )",
+          "expr": "sum by (url_scheme) (\r\n    increase(http_server_request_duration_seconds_count{job=~\"$job\", instance=~\"$instance\"}[$__range])\r\n  )",
```
Aaronontheweb, line 857: This is the "Requests Secured" chart - rather than pick out just the max value of this counter over the given rate, which doesn't change, we now compute the delta just like we've done with the "Total Requests" chart.

## Before

image

Notice how the chart is always growing to the right? That's because we're not taking the change in counter values into account - that's what the `increase` function will do for us.

## This PR

image


Now we can see some livelier data, showing dips and valleys in HTTP traffic to boot.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -912,7 +923,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum by (method_route) (\r\n    label_replace(max_over_time(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\"}[$__rate_interval]), \"method_route\", \"http/$1\", \"network_protocol_version\", \"(.*)\")\r\n  )",
+          "expr": "sum by (method_route) (\r\n    label_replace(increase(http_server_request_duration_seconds_count{job=~\"$job\", instance=~\"$instance\"}[$__range]), \"method_route\", \"http/$1\", \"network_protocol_version\", \"(.*)\")\r\n  )",
```
Aaronontheweb, line 926: This is the "Requests HTTP Protocol" chart - made identical changes to it akin to what we did with "Requests Secured".

## Before

image

## This PR

image

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1059,7 +1070,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "  topk(10,\r\n  sum by (http_route, http_request_method, method_route) (\r\n    label_join(max_over_time(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\", http_route!=\"\"}[$__rate_interval]), \"method_route\", \" \", \"http_request_method\", \"http_route\")\r\n  ))",
+          "expr": "  topk(10,\r\n  sum by (http_route, http_request_method, method_route) (\r\n    label_join(floor(increase(http_server_request_duration_seconds_count{job=~\"$job\", instance=~\"$instance\", http_route!=\"\"}[$__range])), \"method_route\", \" \", \"http_request_method\", \"http_route\")\r\n  ))",
```
Aaronontheweb, line 1073: This is the "Top 10 Requested Endpoints" graph - in addition to adding multi-variable support, we've done two important things:

1. Scoped the number of requests per endpoint to the `$__range` - same idea as some of the other changes, now these values are dynamic per the time period.
2. Since we're now computing a delta with `increase`, I decided to `floor` that value so we can saw off the floating point - aesthetic choice.

## Before

I had to scope to 24 hours because the 6 hour chart rendered no data - that issue does not occur in the new version in this PR.

image

These are cumulative counter values - we didn't actually serve this much traffic to these endpoints during this time.

## This PR

image

These values are accurate and I can verify these numbers in our log aggregation system too.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1233,7 +1244,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "  topk(10,\r\n  sum by (http_route, http_request_method, method_route) (\r\n    label_join(max_over_time(http_server_request_duration_seconds_count{job=\"$job\", instance=\"$instance\", http_route!=\"\", error_type!=\"\"}[$__rate_interval]), \"method_route\", \" \", \"http_request_method\", \"http_route\")\r\n  ))",
```
Aaronontheweb, line 1236: This is the unhandled exceptions by endpoint chart - made the same changes as the "10 Requested Endpoints"

## Before

I can't get any data to show up at all in the current incarnation of this chart, even if I scope out to 24 hours.

## This PR

image

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1270,9 +1281,9 @@
       "type": "table"
     }
   ],
-  "refresh": "10s",
+  "refresh": "1m",
```
Aaronontheweb, line 1284: Changed the default refresh interval to 1m - can change it back. Just had it accidentally left that way when I exported.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1281,16 +1292,17 @@
   "templating": {
     "list": [
       {
+        "allValue": ".*",
```
Aaronontheweb, line 1295: This is the `job` template variable, and this `.*` value as a custom `allValue`  property is actually an important performance optimization for Grafana --> Prometheus. PromQL queries can't be any larger than 64kb or so (I forget the limit, but it's not that large.) If we're using template variable expansion across a large number of unique jobs / server instances (i.e. hundreds or thousands), it's very easy to hit this figure.

Substituting the "all" value with this `.*` wildcard in Prometheus eliminates this issue and keeps the size of the PromQL query small when "all" is selected.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1281,16 +1292,17 @@
...
         "current": {},
         "datasource": {
           "type": "prometheus",
           "uid": "${DS_PROMETHEUS}"
         },
         "definition": "label_values(http_server_active_requests,job)",
         "hide": 0,
-        "includeAll": false,
+        "includeAll": true,
```
Aaronontheweb, line 1303: Enables "all" to be used

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1281,16 +1292,17 @@
...
         "label": "Job",
-        "multi": false,
+        "multi": true,
```
Aaronontheweb, line 1305: Enables mutli-select to be used, which is now feasible since all of the PromQL queries have been updated to support it.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1304,16 +1316,17 @@
         "type": "query"
       },
       {
+        "allValue": ".*",
```
Aaronontheweb, line 1319: This is the `instance` template variable - it as all of the same changes as the `job` template variable.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1329,9 +1342,10 @@
     ]
   },
   "time": {
-    "from": "now-5m",
+    "from": "now-24h",
```
Aaronontheweb, line 1345: Changed the default time horizon to 24h - I can change it back to 5m. Just happened to be set that way when I exported it.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1350,6 +1364,6 @@
   "timezone": "",
   "title": "ASP.NET Core",
   "uid": "KdDACDp4z",
-  "version": 1,
+  "version": 2,
```
Aaronontheweb, line 1367: Need to bump the chart version if this gets pushed to Grafana Cloud.

src/Grafana/dashboards/aspnetcore.json
```diff
@@ -1329,9 +1342,10 @@
     ]
   },
   "time": {
-    "from": "now-5m",
+    "from": "now-24h",
     "to": "now"
   },
+  "timeRangeUpdatedDuringEditOrView": false,
```
Aaronontheweb, line 1348: I'll remove this

Aaronontheweb (commented):
Detailed my changes as thoroughly as I could

JamesNK:
Great work. I think the interval and refresh values should be reset back (I think they're good for immediately after loading the chart to verify that it works) but otherwise the changes all look good.

There is also the endpoint dashboard. It has most of the same set of issues and the changes here should be applied to it. You're welcome to do it, or I can copy the changes to it in a follow up PR.

Aaronontheweb:
>  I think the interval and refresh values should be reset back (I think small values are good for immediately after someone imports the chart to verify that it works)

Done!

> There is also the endpoint dashboard. It has most of the same set of issues and the changes here should be applied to it. You're welcome to do it, or I can copy the changes to it in a follow up PR.

I hadn't even seen that one, but I'll take a look at it - ok to do that in a separate PR?

JamesNK:
I tried it out in the metrics samples app - https://github.com/dotnet/aspire-samples/pull/638. I think I did something like:

* Launch the aspire solution
* Visit weather page. Generates an auth error when getting the weather api
* Visit auth page and login
* Visit weather page again and successfully get data

What I see: 

* Fractional number of requests in the counts:

    image

* Another thing is the counts in the top requested and unhandled exception endpoints. There has been at least one request to `/api/login` and one error getting `/api/weather`, but they report zero.

    image

JamesNK:
I think the problem is the unhandled exception panel is `__rate_interval` wasn't replaced with `__range`

Also, I think `ceil` seems to give more correct results with very small numbers than `floor`.

Aaronontheweb:
I think for the numbers ceil might need to be moved to the outside of the query, rather than on the components being summed

JamesNK:
> I think for the numbers ceil might need to be moved to the outside of the query, rather than on the components being summed

Sorry, I don't know what this means. Is ceil + increase the right approach if it creates inaccurate numbers?

JamesNK:
I did some research and `increase` and `rate` look like they're inaccurate by design - https://github.com/prometheus/prometheus/issues/3746 and https://www.innoq.com/en/blog/2019/05/prometheus-counters/

There doesn't seem to be a good way to get an exact count of a range. I guess we have to live with it. I don't think it is too bad because these counters are increased by 1 at a time, which I think means the number won't be too far off.

On the changes to the charts, they look subjective rather than strictly better. I'll test some more to see whether they're a good improvement.

JamesNK:
I think the time series graphs should keep using `$__rate_interval`. All the samples and discussion I've seen say that this is the right value to use. I believe `$__range` attempts to average out all the values in the visible range which doesn't seem what you want in a time series graph.

The output of the graphs makes more sense with interval:

range:
image

rate_interval:
image

Aaronontheweb:
So for the error rates and request duration - change those back to rate_interval?

JamesNK:
Yes, in the time series graphs.

Aaronontheweb:
> Yes, in the time series graphs.

So I did some playing around with this - the reason I wasn't getting any data on these charts is because the Prometheus scraping interval in the Grafana configuration (i.e. we tell Grafana how frequently Prometheus collects metrics) was set too low, to like 1s. What's yours set to in the test environment? I bumped it up to 15s on ours but I'm still missing data at shorter (i.e. 1hr) time resolutions

JamesNK:
My dev environment for the dashboard is just the metrics sample in the Aspire examples repo. I don't have a real world place to do development.

From what I've read, rate_interval is supposed to protect against issues around scape rate and zooming too far in: https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/.

I haven't read about very small scrape intervals and zooming very far out. If very small scape intervals are a problem then I guess people would have to stop doing that.

JamesNK:
Sorry for the long delay here. I want to try to get improvements here merged in the next week.

This is what I see:
image

JamesNK:
@Aaronontheweb If you're busy right now I can finish changes left in this PR and get it merged. What do you think?

JamesNK:
I think all the changes required for merge are done:

* Consistently use `$__rate_interval` in rate function
* Add ceil to some metrics that were returning decimals where it was missing
* Improved link to endpoints dashboard to include job and instance values
* Update `aspnetcode-endpoints.json` with all changes

Are any other changes required here? I'll merge it soon and update dashboards at https://grafana.com/orgs/dotnetteam/dashboards in a few days.

Aaronontheweb:
LGTM

JamesNK (approved):

Aaronontheweb:
Thanks James - I appreciate the collaboration.

PR #7034, opened 2025-01-07
Add MongoDB.Driver v3 client integration package

### LINKED ISSUE #6380: MongoDB.Driver v3 breaks Aspire.MongoDB.Driver integration
https://github.com/mongodb/mongo-csharp-driver/releases/tag/v3.0.0
 
>  The MongoDB .NET/C# driver team is pleased to announce our v3.0.0 release! The 3.0.0 release incorporates many user-requested fixes and improvements that have been deferred as backwards-incompatible, as well as internal improvements to pay down technical debt and improve maintainability. Additional major changes include removal of a large area of the public API (mainly from MongoDB.Driver.Core), which was not intended for public use. Removed APIs are marked as deprecated in [v2.30.0](https://www.nuget.org/packages/MongoDB.Driver/2.30.0) version.
For all the breaking changes and for the upgrade guidlines, please see the [upgrade guide](https://www.mongodb.com/docs/drivers/csharp/v3.0/upgrade/v3/).

We should follow the same approach as outlined in [RabbitMQ.Client v7 breaks Aspire.RabbitMQ.Client component (dotnet/aspire#3956)](https://github.com/dotnet/aspire/issues/3956)

cc @sebastienros @radical @joperezr @DamianEdwards

eerhardt:
## Description

MongoDB.Driver shipped a version `3.0.0` that has binary breaking changes such that we can't build a single library that can load in both v2.x and v3.x. To continue supporting new versions of MongoDB.Driver, we need to add a new component Aspire.MongoDB.Driver.v3 that will work with the new version of MongoDB.Driver. See #3956 for the strategy to deal with these breaking changes.

Fix #6380

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] No
        - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/2360

### DIFF
--- a/Aspire.sln
+++ b/Aspire.sln
@@ -637,6 +637,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ServiceBusWorker", "playgro
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ServiceBus.AppHost", "playground\AzureServiceBus\ServiceBus.AppHost\ServiceBus.AppHost.csproj", "{A7EC9111-F3CC-46E8-B95E-3768481D67B4}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Aspire.MongoDB.Driver.v3", "src\Components\Aspire.MongoDB.Driver.v3\Aspire.MongoDB.Driver.v3.csproj", "{FD53B608-138D-8FB1-AA57-9B8CD42CF13E}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Aspire.MongoDB.Driver.v3.Tests", "tests\Aspire.MongoDB.Driver.v3.Tests\Aspire.MongoDB.Driver.v3.Tests.csproj", "{223AF8EB-3A4E-E778-4EBD-6E4876C308B6}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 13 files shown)


src/Components/Aspire.MongoDB.Driver/Aspire.MongoDB.Driver.csproj
```diff
@@ -20,14 +21,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AspNetCore.HealthChecks.MongoDb" />
-    <PackageReference Include="MongoDB.Driver" />
+    <PackageReference Include="AspNetCore.HealthChecks.MongoDb" VersionOverride="[8.1.0,9.0.0)" />
```
sebastienros, line 24: It's there because Directory.Packages (and/or Versions.props) doesn't handle ranges?

src/Components/Aspire.MongoDB.Driver/Aspire.MongoDB.Driver.csproj
```diff
@@ -7,6 +7,7 @@
     <PackageIconFullPath>$(SharedDir)MongoDB_300px.png</PackageIconFullPath>
     <Description>A generic MongoDB client that integrates with Aspire.</Description>
     <NoWarn>$(NoWarn);CS8002</NoWarn> <!-- MongoDB.Driver.Core.Extensions.DiagnosticSources and AspNetCore.HealthChecks.MongoDb packages are not signed -->
+    <DefineConstants>$(DefineConstants);MONGODB_V2</DefineConstants>
```
sebastienros, line 10: It feels weird to have a `MONGODB_V2` tag and a also `v3` moniker. Is it harder to have `MONGODB_V3` instead so we only think about one special case?

src/Components/Aspire.MongoDB.Driver/Aspire.MongoDB.Driver.csproj
eerhardt: No. This is necessary because Directory.Packages.props defines the `9.0.0` version, but this package needs the previous version - 8.1.0 since it uses MongoDB v2.

src/Components/Aspire.MongoDB.Driver/Aspire.MongoDB.Driver.csproj
eerhardt: It isn't harder. I guess my thinking was to make the new code the "default" (no #define for it). And the old code be under the #define. Then when we remove the old version, we just delete the #ifs. But I can flip it around if you'd like so we use a `_V3` define.

davidfowl (approved):
😭

src/Components/Aspire.MongoDB.Driver/Aspire.MongoDB.Driver.csproj
joperezr: Regarding making the new thing the default, did we also consider having the version-less package be v3, and then having the older version be the one that has the `.v2` on the assembly name? I understand we may want to do the reverse for backwards compat, but given these are mostly referred by end-user apps I'd assume it might be safe?

eerhardt: See https://github.com/dotnet/aspire/issues/3956 for the strategy to deal with these breaking changes.

PR #7037, opened 2025-01-07
Refactor PackageVersions and dependencies

eerhardt:
In the dotnet/aspire repo today, the way we switch out dependency versions between net8 and net9 doesn't work if you only target a single TFM - net9.0. This is because we have an MSBuild Property condition based on TargetFramework, which isn't set at property evaluation time.

To fix this, refactor the condition to be an ItemGroup condition on TargetFramework, which gets evaluated after properties get set, so TargetFramework is set. This follows the same approach used in dotnet/extensions.

- To also follow dotnet/extensions (and so the property names aren't so long), I changed the PackageVersion suffix to just Version.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -49,16 +49,6 @@
     <PackageVersion Include="Azure.ResourceManager.Authorization" Version="1.1.3" />
     <PackageVersion Include="Azure.ResourceManager.KeyVault" Version="1.3.0" />
     <PackageVersion Include="Azure.ResourceManager.Resources" Version="1.9.0" />
-    <!-- ASP.NET Core dependencies -->
-    <PackageVersion Include="Microsoft.AspNetCore.Authentication.Certificate" Version="$(MicrosoftAspNetCoreAuthenticationCertificatePackageVersion)" />
-    <PackageVersion Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="$(MicrosoftAspNetCoreAuthenticationJwtBearerPackageVersion)" />
-    <PackageVersion Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="$(MicrosoftAspNetCoreAuthenticationOpenIdConnectPackageVersion)" />
-    <PackageVersion Include="Microsoft.AspNetCore.OutputCaching.StackExchangeRedis" Version="$(MicrosoftAspNetCoreOutputCachingStackExchangeRedisPackageVersion)" />
-    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="$(MicrosoftAspNetCoreTestHostPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(MicrosoftExtensionsCachingStackExchangeRedisPackageVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Diagnostics.HealthChecks.EntityFrameworkCore" Version="$(MicrosoftExtensionsDiagnosticsHealthChecksEntityFrameworkCorePackageVersion)" />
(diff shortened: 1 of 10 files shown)


tests/Aspire.Oracle.EntityFrameworkCore.Tests/Aspire.Oracle.EntityFrameworkCore.Tests.csproj
```diff
@@ -11,6 +11,9 @@
     <ProjectReference Include="..\..\src\Components\Aspire.Oracle.EntityFrameworkCore\Aspire.Oracle.EntityFrameworkCore.csproj" />
     <ProjectReference Include="..\Aspire.Components.Common.Tests\Aspire.Components.Common.Tests.csproj" />
 
+    <!-- Needed until Oracle supports net9.0 -->
```
joperezr, line 14: NIT: Is there an issue that we can link here to track?

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
```diff
@@ -1,6 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
+    <!--
```
joperezr, line 4: Can you elaborate a bit more on this change? Not sure I'm following why the comment was moved but all package references were removed.

Directory.Packages.props
```diff
@@ -188,7 +163,69 @@
     <PackageVersion Include="Azure.Identity" Version="1.13.1" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="System.Formats.Asn1" Version="$(SystemFormatsAsn1PackageVersion)" />
-    <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonPackageVersion)" />
+  </ItemGroup>
+
+  <!-- The following 2 groups are for packages that need to switch based on the .NET TFM being used. -->
+
+  <ItemGroup>
```
joperezr, line 170: NIT: Should this be conditioned as well for net8 tfm?

joperezr (approved):
Couple of questions but looks great overall. Thanks!

tests/Aspire.Oracle.EntityFrameworkCore.Tests/Aspire.Oracle.EntityFrameworkCore.Tests.csproj
eerhardt: Done.

Directory.Packages.props
eerhardt: My thinking was that these are unconditional, and then the "STS" version "updates" the PackageVersion. Either way could work, I don't have strong opinion. Do you?

tests/testproject/TestProject.IntegrationServiceA/TestProject.IntegrationServiceA.csproj
eerhardt: Previously we had to explicitly VersionOverride all those packages for `9.0.0` even though we targeted `net9.0` because of the problem I'm fixing here. Since this project has a single `TargetFramework`, the 9.0.0 packages weren't referenced by default because we had a PropertyGroup condition on `TargetFramework`, which isn't set at the time the condition is evaluated.

With this change, we don't need the VersionOverrides anymore. We can just target `net9.0` and the right packages are brought in.

The comment moved because all the other projects are targeting `net8.0`, but this one explicitly needs to be `net9.0` for the reason the comment mentions.

PR #7045, opened 2025-01-08
Fix playground apps after Npgsql 9 update

eerhardt:
Npgsql EF requires a matching major version with the Npgsql version being used. There are breaking changes between major versions that don't allow Npgsql EF 8 to load with Npgsql 9, and vice versa.

Workaround this by ensuring that whenever we use Npgsql EF we use a matching Npgsql version.

It looks like the playground tests were disabled in [Improve auto-forwarding for GitHub Codespaces and devcontainers. (dotnet/aspire#6780)](https://github.com/dotnet/aspire/pull/6780).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,6 +8,7 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <TestcontainersPackageVersion>4.0.0</TestcontainersPackageVersion>
     <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
+    <!-- The Npgsql version used when using Npgsql EF Core on net8. The major versions need to match between Npgsql and EF Core. -->
     <Npgsql8Version>8.0.6</Npgsql8Version>
   </PropertyGroup>
   <ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,6 +8,7 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <TestcontainersPackageVersion>4.0.0</TestcontainersPackageVersion>
     <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
+    <Npgsql8Version>8.0.6</Npgsql8Version>
   </PropertyGroup>
   <ItemGroup>
     <!-- Azure SDK for .NET dependencies -->
--- a/playground/PostgresEndToEnd/PostgresEndToEnd.ApiService/PostgresEndToEnd.ApiService.csproj
+++ b/playground/PostgresEndToEnd/PostgresEndToEnd.ApiService/PostgresEndToEnd.ApiService.csproj
@@ -8,6 +8,9 @@
 
   <ItemGroup>
     <AspireProjectOrPackageReference Include="Aspire.Npgsql.EntityFrameworkCore.PostgreSQL" />
+    <!-- Npgsql EF needs to match the same major version as the underlying Npgsql assemblies. -->
+    <PackageReference Include="Npgsql.DependencyInjection" VersionOverride="$(Npgsql8Version)" />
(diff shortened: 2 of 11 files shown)


sebastienros (approved):

Directory.Packages.props
```diff
@@ -8,6 +8,7 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
     <TestcontainersPackageVersion>4.0.0</TestcontainersPackageVersion>
     <AzureProvisiongVersion>1.0.0</AzureProvisiongVersion>
+    <Npgsql8Version>8.0.6</Npgsql8Version>
```
joperezr, line 11: NIT: might be good to add a comment above this property to say that this version should never be updated to 9.0 or greater, since different people will be running tools to update versions and it would be good to make sure we don't update this one passed that. Of course you have the property name saying it is 8 anyway, so if you think that's enough that is good too. For sure don't reset CI just for this, so only consider in case you want to make other changes.

playground/PostgresEndToEnd/PostgresEndToEnd.ApiService/PostgresEndToEnd.ApiService.csproj
```diff
@@ -8,6 +8,9 @@
 
   <ItemGroup>
     <AspireProjectOrPackageReference Include="Aspire.Npgsql.EntityFrameworkCore.PostgreSQL" />
+    <!-- Npgsql EF needs to match the same major version as the underlying Npgsql assemblies. -->
+    <PackageReference Include="Npgsql.DependencyInjection" VersionOverride="$(Npgsql8Version)" />
```
joperezr, line 12: NIT: Should this be conditioned? Mainly thinking that at some point DefaultTargetFramework will change so this would likely break.

playground/TestShop/CatalogService/CatalogService.csproj
```diff
@@ -13,13 +13,13 @@
...
-  <ItemGroup>
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" />
-  </ItemGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\CatalogModel\CatalogModel.csproj" />
     <ProjectReference Include="..\TestShop.ServiceDefaults\TestShop.ServiceDefaults.csproj" />
+
+    <!-- Npgsql EF needs to match the same major version as the underlying Npgsql assemblies. -->
+    <PackageReference Include="Npgsql.DependencyInjection" VersionOverride="$(Npgsql8Version)" />
```
joperezr, line 21: NIT: should we consider doing the override in a single place (like a directory.build.props so all of these testshop projects can benefit) and that way we reduce the number of places that need to be updated come .net 10?

joperezr (approved):
Left only NITs, non are blocking, more just questions, so this LGTM.

playground/TestShop/CatalogService/CatalogService.csproj
eerhardt: We should only override the PackageRefernce if the app is also referencing Npgsql EF Core. There are only a handful of playground apps use it. We can't do it in a single place without affecting the projects that don't want it.

The real problem here is transitive pinning. Since most projects want Npgsql v9, we use that in our Directory.Packages.props (even when targeting net8 - see #6946). But when an app is targeting net8 -> it needs EF Core 8. And if it uses EF Core 8 -> it needs Ngpsql 8.

playground/PostgresEndToEnd/PostgresEndToEnd.ApiService/PostgresEndToEnd.ApiService.csproj
eerhardt: Adding a TFM condition on a project that only targets 1 TFM isn't useful IMO. When we change DefaultTargetFramework (which probably won't be until net10 or beyond), we can adjust the references.

Directory.Packages.props
eerhardt: Added a comment.

Issue #7066, opened 2025-01-10
Azure ServiceBus emulator is not reliable in CI


eerhardt:
When running tests that use the ServiceBus emulator in CI, we are randomly getting failures with the sqledge container not starting successfully.

Examples:

* https://dev.azure.com/dnceng-public/public/_build/results?buildId=912694
* https://dev.azure.com/dnceng-public/public/_build/results?buildId=912738

Or when running Azure Functions playground app in the Playground.Tests:

https://dev.azure.com/dnceng-public/public/_build/results?buildId=913500

https://helixr1107v0xdeko0k025g8.blob.core.windows.net/dotnet-aspire-refs-pull-7048-merge-464d941387084146a7/Aspire.Playground.Tests/1/console.b2dd2d1b.log?skoid=8eda00af-b5ec-4be9-b69b-0919a2338892&sktid=72f988bf-86f1-41af-91ab-2d7cd011db47&skt=2025-01-10T21%3A46%3A19Z&ske=2025-01-10T22%3A46%3A19Z&sks=b&skv=2024-11-04&sv=2024-11-04&st=2025-01-10T21%3A46%3A19Z&se=2025-01-10T22%3A46%3A19Z&sr=c&sp=r&sig=cVb5usSQfhF3fv1uoy5lsT2h8kI84vS5kwfTop3fooY%3D

After the tests exit we see:

```
+ docker container ls --all
CONTAINER ID   IMAGE                                                         COMMAND                  CREATED          STATUS                        PORTS                                                                                NAMES
a1c98b3a374c   mcr.microsoft.com/azure-sql-edge:latest                       "/opt/mssql/bin/perm…"   18 minutes ago   Exited (1) 16 minutes ago                                                                                          messaging-sqledge-egpcscww-23011b3d
7fbf12f93e41   mcr.microsoft.com/azure-storage/azurite:3.33.0                "docker-entrypoint.s…"   18 minutes ago   Up 18 minutes                 127.0.0.1:32778->10000/tcp, 127.0.0.1:32777->10001/tcp, 127.0.0.1:32776->10002/tcp   storage-bcfuwrtt-23011b3d
[...]

PR #5827, opened 2024-09-22
Support connection name-based configuration for Azure components

captainsafia:
## Description

This pull request adds support for named configuration providers to Azure components and updates resource types that implement `IResourceWithAzureFunctions` config to inject values into configuration model.

* Upgraded `Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore` to version `2.0.0-preview1` to use new `FunctionsWebApplicationBuilder` implementation so we can take advantage of `IHostApplicationBuilder`-based extension methods for integrations.
* Updates implementors of `IResourceWithAzureFunctionsConfig` to inject named configuration values into environment variables.
* Updated `AzureComponent` to support binding settings from named configuration providers.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [X] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [X] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [X] No
- Does the change require an update in our Aspire docs?
  - [X] Yes
    - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/1663
  - [ ] No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/5827)

PR #7067, opened 2025-01-10
Disable Azure ServiceBus emulator functional tests

eerhardt:
The emulator isn't reliable / consistent in CI. Disabling for now.

See #7066

### DIFF
--- a/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
+++ b/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
@@ -17,11 +17,17 @@
     <PlaygroundSourceDir>$(MSBuildThisFileDirectory)..\..\playground\</PlaygroundSourceDir>
     <TestsSharedDir>$(MSBuildThisFileDirectory)..\Shared\</TestsSharedDir>
     <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
+    <!-- Skip emulators that don't start consistently when running in CI. -->
+    <SkipUnstableEmulators Condition="'$(SkipUnstableEmulators)' == '' and ('$(RepoRoot)' == '' or '$(ContinuousIntegrationBuild)' == 'true' or '$(CODESPACES)' == 'true')">true</SkipUnstableEmulators>
 
     <!-- on helix this will be available in the source dir -->
     <XunitRunnerJson Condition="'$(RepoRoot)' == ''">xunit.runner.json</XunitRunnerJson>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(SkipUnstableEmulators)' == 'true'">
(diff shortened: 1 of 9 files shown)


sebastienros (approved):

tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
```diff
@@ -17,11 +17,17 @@
...
     <TestsSharedDir>$(MSBuildThisFileDirectory)..\Shared\</TestsSharedDir>
     <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
+    <!-- Skip emulators that don't start consistently when running in CI. -->
+    <SkipUnstableEmulators Condition="'$(SkipUnstableEmulators)' == '' and ('$(RepoRoot)' == '' or '$(ContinuousIntegrationBuild)' == 'true' or '$(CODESPACES)' == 'true')">true</SkipUnstableEmulators>
 
     <!-- on helix this will be available in the source dir -->
     <XunitRunnerJson Condition="'$(RepoRoot)' == ''">xunit.runner.json</XunitRunnerJson>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(SkipUnstableEmulators)' == 'true'">
```
joperezr, line 27: NIT: Can this just be done in the parent's Directory.Build.props so it applies to both playground and tests and that way we don't have to change multiple places later?

Feel free to not block on this one, but just something to consider.

tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
```diff
@@ -17,11 +17,17 @@
     <PlaygroundSourceDir>$(MSBuildThisFileDirectory)..\..\playground\</PlaygroundSourceDir>
     <TestsSharedDir>$(MSBuildThisFileDirectory)..\Shared\</TestsSharedDir>
     <RunSettingsFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsFilePath>
+    <!-- Skip emulators that don't start consistently when running in CI. -->
```
joperezr, line 20: NIT: I know you did it on the facts, but might be good to also add the link to the issue tracking removing this in here too. Also not block on this, if you agree we can add it in your next set of changes.

joperezr (approved):
Thanks @eerhardt !

tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
eerhardt: Note that we already had this code in the repo.

* Orignally added in https://github.com/dotnet/aspire/pull/5418
* Renamed in https://github.com/dotnet/aspire/pull/5827
* Removed in https://github.com/dotnet/aspire/pull/6737
* and now brought back with another different name.

One reason to not put it in the root of the repo is so it doesn't affect other projects - like stuff under `src`.

PR #7072, opened 2025-01-11
Move PublishAsDockerFile to use container resources.

### LINKED ISSUE #4067: Change PublishAsDockerFile to use a ContainerResource
.PublishAsDockerFile() hardcodes the path to the Dockerfile and the name of the Dockerfile. 

The path and name cant be overridden using DockerBuildArgs.

Why this matters: When building a monorepo, such as a Vercel turbo monorepo, the working directory for the docker build must be the parent folder at the root of the monorepo (so the docker build process can access packages and etc. in adjacent folders), but there can only be one file called Dockerfile in that root location, preventing publishing multiple apps from the monorepo.

The problematic line is in ExecutableResourceBuilderExtensions.cs:
```
private static async Task WriteExecutableAsDockerfileResourceAsync(ManifestPublishingContext context, ExecutableResource executable, IEnumerable<DockerBuildArg>? buildArgs = null)
{
    ...
    var appHostRelativePathToDockerfile = Path.Combine(executable.WorkingDirectory, "Dockerfile");
    ...
}

### PR DESCRIPTION
**Pull request by davidfowl**
## Description

This allows full customization of the docker building process and removes a dependency on the limited docker.v0 resource type.

- Updated `DockerfileBuildAnnotation` to allow nullable build arguments.
- Modified `WithBuildArg` to accept nullable values.
- Revised tests in `ManifestGenerationTests.cs` and `PublishAsDockerfileTests.cs` to align with new behavior.
- Adjusted `SchemaTests.cs` to use temporary context paths for Dockerfile testing.

Fixes #4067 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [x] Yes
      - [ ] No
  - [ ] No

### CODE REVIEW

File: tests/Aspire.Hosting.Python.Tests/AddPythonAppTests.cs [hunk 1]
```diff
@@ -269,6 +274,18 @@ private static void PreparePythonProject(ITestOutputHelper outputHelper, string
...
+        var dockerFilePath = Path.Combine(projectDirectory, "Dockerfile");
+        File.WriteAllText(dockerFilePath,
+        """
+        FROM python:3.9
+        WORKDIR /app
+        COPY requirements.txt .
+        RUN pip install --no-cache-dir -r requirements.txt
+        COPY . .
+        CMD ["python", "main.py"]
+        """);
```
eerhardt, line 287:
```suggestion
        File.WriteAllText(dockerFilePath,
            """
            FROM python:3.9
            WORKDIR /app
            COPY requirements.txt .
            RUN pip install --no-cache-dir -r requirements.txt
            COPY . .
            CMD ["python", "main.py"]
            """);
```

It's more readable if this is indented.

tests/Aspire.Hosting.Tests/Schema/SchemaTests.cs
```diff
@@ -109,7 +109,13 @@ public static TheoryData<string, Action<IDistributedApplicationBuilder>> Applica
 
                 { "BasicDockerfile", (IDistributedApplicationBuilder builder) =>
                     {
-                        builder.AddExecutable("foo", "bar", "baz", "one", "two", "three").PublishAsDockerFile();
+                        var tempPath = Path.GetTempPath();
+                        var tempContextPath = Path.Combine(tempPath, Path.GetRandomFileName());
+                        Directory.CreateDirectory(tempContextPath);
+                        var tempDockerfilePath = Path.Combine(tempContextPath, "Dockerfile");
```
eerhardt, line 115:
```suggestion
                        var tempContextPath = Directory.CreateTempSubdirectory().FullName;
                        var tempDockerfilePath = Path.Combine(tempContextPath, "Dockerfile");
```

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -69,30 +79,64 @@ public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistribut
...
-        if (buildArgs is not null)
-        {
-            context.WriteDockerBuildArgs(buildArgs);
-        }
+        var container = new ExecutableContainerResource(builder.Resource);
+        var cb = builder.ApplicationBuilder.AddResource(container);
+        cb.WithImage(builder.Resource.Name);
+        cb.WithDockerfile(contextPath: builder.Resource.WorkingDirectory);
+        // Clear the runtime args
+        cb.WithArgs(c => c.Args.Clear());
```
eerhardt, line 126: Why? Can we say why in the comment?

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -69,30 +79,64 @@ public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistribut
...
+        // in publish mode, it won't be used. This is a limitation of the current design.
+        builder.ApplicationBuilder.Resources.Remove(builder.Resource);
 
-        if (buildArgs is not null)
-        {
-            context.WriteDockerBuildArgs(buildArgs);
-        }
+        var container = new ExecutableContainerResource(builder.Resource);
+        var cb = builder.ApplicationBuilder.AddResource(container);
+        cb.WithImage(builder.Resource.Name);
```
eerhardt, line 123: For my understanding, why do we need to call `WithImage` when we call `WithDockerfile`? Is that just to name the image that will be produced?

eerhardt: Also, doesn't `WithDockerfile` just overwrite what we pass here?

https://github.com/dotnet/aspire/blob/882ae2be3e02edbfc1d222ee97fe2da4eea3fe99/src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs#L366-L371

tests/Aspire.Hosting.Tests/ManifestGenerationTests.cs
```diff
@@ -107,43 +107,6 @@ public async Task WithContainerRegistryUpdatesContainerImageAnnotationsDuringPub
         Assert.Equal(expectedManifest, redisManifest.ToString());
     }
 
-    [Fact]
-    public void EnsureExecutablesWithDockerfileProduceDockerfilev0Manifest()
```
eerhardt, line 111: Why is this test being deleted? Can it just be updated for the new behavior?

davidfowl: This entire file should be deleted except for the uber test. This was before we had split tests for each resource type.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
davidfowl: Good catch!

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
davidfowl: Yea I should add a comment here. The args that apply to the executable *could* in theory apply to the container but for the ones we ship it does not just work.

tests/Aspire.Hosting.Tests/PublishAsDockerfileTests.cs
```diff
@@ -113,4 +197,24 @@ public async Task PublishAsDockerFileConfiguresManifestWithBuildArgsThatHaveNoVa
...
     }
+
+    private static DisposableTempDirectory CreateDirectoryWithNodeDockerFile()
+    {
+        var tempDir = Directory.CreateTempSubdirectory("aspire-docker-test");
+        File.WriteAllText(Path.Join(tempDir.FullName, "Dockerfile"), "FROM node:14");
+        return new DisposableTempDirectory(tempDir);
+    }
+
+    struct DisposableTempDirectory : IDisposable
```
eerhardt, line 208:
```suggestion
    readonly struct DisposableTempDirectory : IDisposable
```

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
davidfowl: Nah you need aan image name.

davidfowl: System.InvalidOperationException : The resource 'foo' does not have a container image specified. Use WithImage to specify the container image and tag.

eerhardt: ah - it's because of the ORDER of the calls above. If you change it to:

```C#
        return builder.WithAnnotation(annotation, ResourceAnnotationMutationBehavior.Replace)
                      .WithImage(imageName)
                      .WithImageRegistry(registry: null)
                      .WithImageTag("latest");
```

it will work. `WithImageRegistry` is throwing because it can't find the `ContainerImageAnnotation`.

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
```diff
@@ -69,30 +79,71 @@ public static IResourceBuilder<ExecutableResource> AddExecutable(this IDistribut
...
-        if (buildArgs is not null)
-        {
-            context.WriteDockerBuildArgs(buildArgs);
-        }
+        var container = new ExecutableContainerResource(builder.Resource);
+        var cb = builder.ApplicationBuilder.AddResource(container);
+        // WithImage makes this a container resource (adding the annotation)
+        cb.WithImage(builder.Resource.Name);
+        cb.WithDockerfile(contextPath: builder.Resource.WorkingDirectory);
+        // Arguments to the executive often contain physical paths that are not valid in the container
```
eerhardt, line 131:
```suggestion
        // Arguments to the executable often contain physical paths that are not valid in the container
```

src/Aspire.Hosting/ExecutableResourceBuilderExtensions.cs
davidfowl: I added a comment

eerhardt: I think we should fix WithDockerfile.

The current proposed code is confusing because you are going to get a different Image name that what is being specified here - since WithDockerfile is going to overwrite it.

davidfowl: not in this PR

eerhardt: I still think the code you are adding in this PR is confusing. But I'm not going to block on it.

eerhardt (approved):

Issue #5956, opened 2024-09-21
v9.0.0-rc.1.24431.7: fails to build aspire in CI=true environments


xen0n:
### Describe the Bug

`aspire` fails to build from source if `CI=true` is present in the environment, which is the case on many CI platforms e.g. Travis CI and GitHub Actions, preventing source builds in such CI environments without kludge.

[Code responsible for the behavior](https://github.com/dotnet/dotnet/blob/v9.0.0-rc.1.24431.7/src/aspire/Directory.Build.targets#L18)

### Steps to Reproduce

```sh
git clone -b v9.0.0-rc.1.24431.7 --recurse-submodules https://github.com/dotnet/dotnet.git
cd dotnet
docker run --rm --platform linux/amd64 -v $(pwd):/vmr -e CI=true -ti mcr.microsoft.com/dotnet-buildtools/prereqs:centos-stream9

# inside docker
cd /vmr
./prep-source-build.sh
./build.sh -so --clean-while-building -c Release /p:PortableBuild=true
```

### Other Information

Example run: https://github.com/loongson-community/dotnet-unofficial-build/actions/runs/10969826008/job/30463397000
Although the project is set up for `linux-loongarch64` cross builds, the failure occurs at Stage 1 so still relevant to `linux-x64`.

Relevant log message:

```
/vmr/src/aspire/eng/Versions.targets(49,67): error MSB4057: The target "GitInfo" does not exist in the project. [/vmr/src/aspire/src/Microsoft.NET.Sdk.Aspire/Microsoft.NET.Sdk.Aspire.csproj]
```

I'm able to work around the problem by simply overriding `CI` for `build.sh` invocations [like this](https://github.com/loongson-community/dotnet-unofficial-build/commit/d8a9ecbf87875051c2d288899ee790ed48651394).

radical:
Is `CI=true` a vmr thing? should that be mapped for `ContinuousIntegrationBuild=true` somewhere?

radical:
cc @joperezr

xen0n:
Hi, thank you for following up on this! While I'm not a .NET dev nor very familiar with .NET or the Microsoft CI offerings, here's my current understanding:

> Is `CI=true` a vmr thing?

[...]

PR #7073, opened 2025-01-12
Trying out gha to run tests

### LINKED ISSUE #4599: [tests] Failing test on CI - `DistributedApplicationTests.ProxylessAndProxiedEndpointBothWorkOnSameResource`
https://github.com/dotnet/aspire/pull/4500 enables extra `Aspire.Hosting.Tests` to run on the build machine. Prior to this they were never run on CI.

`Aspire.Hosting.Tests.DistributedApplicationTests.ProxylessAndProxiedEndpointBothWorkOnSameResource` fails because the first `client.GetStringAsync($"{httpEndPoint}urls"` returned only the `http` endpoint. But the second invocation got both `http`, and `https`. Adding a sufficient delay before hitting the `/urls` endpoints makes the test pass though.

The tests have an existing class for the hook which I tried to use here - https://gist.github.com/radical/7ef257229178c36f6b72ffd09f7e0b79#file-test-cs-L22-L27

But running the test with CWLs added to print the results from `/urls`:
```
Got urls from http://localhost:1234/urls: ["http://localhost:1234"]
Got urls from https://localhost:57048/urls: ["http://localhost:1234","https://localhost:57051"]
```

The first call is missing the `https` endpoint. Did the `AfterEndpointsAllocatedAsync` get invoked too early? Or am I missing something else?

(related thread: https://github.com/dotnet/aspire/pull/4500#discussion_r1639055442)
cc @ReubenBond @DamianEdwards @eerhardt 

## Build Information
Build: 
Build error leg or test failing: Aspire.Hosting.Tests.DistributedApplicationTests.ProxylessAndProxiedEndpointBothWorkOnSameResource
Pull request: 

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "Expected: \"[\"http://localhost:1234\"]\"",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": true
}
```

## Last 10 failures:
| Run date | Test Name |
| --------- | ---- |

-- Updated on 5/6/2025 3:39:57 AM UTC

## Summary

| Last 40 | Total failed / total runs |
| --- | --- |
| 0 | 0 / 100 |

davidfowl:
Run integration tests on github actions as individual jobs for maximum concurrency.

This pull request introduces a variety of changes aimed at enhancing the functionality and robustness of the Aspire Hosting project. The most significant updates include the addition of integration tests, the introduction of Unix file mode settings for bind mounts, and various improvements to the PostgreSQL and Redis components.

Fixes https://github.com/dotnet/aspire/issues/4599


src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -304,6 +306,17 @@ public static IResourceBuilder<PostgresServerResource> WithPgWeb(this IResourceB
...
+                // Need to grant read access to the config file on unix like systems.
+                if (!OperatingSystem.IsWindows())
+                {
+                    // 777
+                    var permissions = UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.UserExecute |
+                                      UnixFileMode.GroupRead | UnixFileMode.GroupWrite | UnixFileMode.GroupExecute |
+                                      UnixFileMode.OtherRead | UnixFileMode.OtherWrite | UnixFileMode.OtherExecute;
+
+                    File.SetUnixFileMode(serverFileMount.Source!, permissions);
+                }
```
eerhardt, line 318: This can be done as part of the call to CreateDirectory above:

```C#
    public static class Directory
    {
        [UnsupportedOSPlatform("windows")]
        public static DirectoryInfo! CreateDirectory(string! path, UnixFileMode unixCreateMode);
    }
```

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -304,6 +306,17 @@ public static IResourceBuilder<PostgresServerResource> WithPgWeb(this IResourceB
                     Directory.CreateDirectory(serverFileMount.Source!);
                 }
 
+                // Need to grant read access to the config file on unix like systems.
```
eerhardt, line 309: Why wasn't this needed before?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,91 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - project: tests/Aspire.Hosting.Elasticsearch.Tests/Aspire.Hosting.Elasticsearch.Tests.csproj
+            name: Elasticsearch
+            filter: "WithDataShouldPersistStateBetweenUsages"
+          - project: tests/Aspire.Hosting.PostgreSQL.Tests/Aspire.Hosting.PostgreSQL.Tests.csproj
+            name: PostgreSQL
+            filter: "Aspire.Hosting"
```
eerhardt, line 20: Are these filters just for now? I wouldn't expect us to use them when we did this for real.

tests/Aspire.Hosting.Tests/Utils/TestDistributedApplicationBuilder.cs
```diff
@@ -61,7 +61,7 @@ public static TestDistributedApplicationBuilder Create(Action<DistributedApplica
         return new TestDistributedApplicationBuilder(configureOptions, testOutputHelper);
     }
 
-    public static TestDistributedApplicationBuilder CreateWithTestContainerRegistry(ITestOutputHelper testOutputHelper) =>
+    public static TestDistributedApplicationBuilder CreateWithTestContainerRegistry(ITestOutputHelper? testOutputHelper = null) =>
```
eerhardt, line 64: I think this should be reverted. We always want an ITestOutputHelper to be passed so we get log output correctly.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
```diff
@@ -566,6 +566,21 @@ public async Task RedisInsightWithDataShouldPersistStateBetweenUsages(bool useVo
...
+                    Directory.CreateDirectory(bindMountPath);
+                }
+
+                if (!OperatingSystem.IsWindows())
+                {
+                    var permissions = UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.UserExecute |
+                                      UnixFileMode.GroupRead | UnixFileMode.GroupWrite | UnixFileMode.GroupExecute |
+                                      UnixFileMode.OtherRead | UnixFileMode.OtherWrite | UnixFileMode.OtherExecute;
+
+                    File.SetUnixFileMode(bindMountPath, permissions);
```
eerhardt, line 581: We shouldn't be doing this in the test - this seems like a product bug.

tests/Aspire.Hosting.Tests/Utils/TestDistributedApplicationBuilder.cs
davidfowl: I was debugging 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60 
+    strategy:
+      fail-fast: false
+      matrix:
```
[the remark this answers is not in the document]
joperezr: That is probably good as it allows us to re-run individual test assemblies. Especially since I believe the plan is to include in here only those which are flaky, IIUC

.github/workflows/ci.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: |
+          docker info
+
+      - name: Restore dependencies
+        run: |
+          dotnet restore ${{ matrix.project }}
+
+      - name: Build projects
+        run: |
+          dotnet build ${{ matrix.project }} -c Release /p:CI=false --no-restore
```
joperezr, line 64: What is `/p:CI=false` for? Also, as a NIT, I'm assuming we could theoretically collapse restore, build and test phase into a single command as I believe dotnet test should do all of that.

joperezr: Might also be good to remove the `-c Release` as Debug will help having more actionable info in callstack in case test fails.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,91 @@
...
+    branches: [ main ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60 
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
joperezr, line 14: Out of curiosity, how where these test assemblies picked? Is this at all related to the work that @JamesNK was going to do to gather stats on most common test failures?

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchFunctionalTests.cs
```diff
@@ -62,23 +62,20 @@ await pipeline.ExecuteAsync(
...
-    [InlineData(true)]
-    [InlineData(false)]
+    [Fact]
+    // [InlineData(true)]
+    // [InlineData(false)]
     [RequiresDocker]
-    public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
+    public async Task WithDataShouldPersistStateBetweenUsages()
     {
+        bool useVolume = false;
```
joperezr, line 71: Is this just a temporary change? is the true case flaky?

.github/workflows/ci.yml
davidfowl: > What is /p:CI=false for? Also, as a NIT, I'm assuming we could theoretically collapse restore, build and test phase into a single command as I believe dotnet test should do all of that.

I get this https://github.com/dotnet/aspire/issues/5956

.github/workflows/ci.yml
davidfowl: I plan to do all , but I just chose the worst ones from the ado stats

tests/Aspire.Hosting.Elasticsearch.Tests/ElasticsearchFunctionalTests.cs
davidfowl: Don't look at the source changes in this PR 😄

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
davidfowl: Not sure how our machines or docker is setup. 

@danegsta ideas?

danegsta: The need for custom file permissions on Mac/Linux really depends on the specific container. If the container is running as root, it’s not really necessary, but if a container is running as a custom user, then we have to jump through hoops to make sure that user will be able to read the mounted file. So if the pgweb image started using a specific user instead of root, that’d necessitate the change.

davidfowl: Its strange we dont need this on our other machine though right? I had to add it here because I saw permission errors

.github/workflows/ci.yml
davidfowl: I was playing with how to enable filters. I don't really want any in the final version.

tests/Aspire.Hosting.Redis.Tests/RedisFunctionalTests.cs
davidfowl: It's something, but I don't know why it happens on this gha run.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
davidfowl: I know this is per container, but I wonder if we can as part of creating a BindMount if we should allow setting these flags..

davidfowl: Take a look at the new changes I made WRT bind mounts in the latest commits. I made it first class on bind mount and calls to `WithBindMount` can opt into what per resource.

I think we solve this on an integration by integration basis depending on what minimum permissions that bind mount needs based on the interaction with the container. I still don't know why I had to make these changes in this pr to make it work though.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1497,6 +1497,23 @@ private void PrepareContainers()
 
                 foreach (var mount in containerMounts)
                 {
+                    if (mount.Type == ContainerMountType.BindMount &&
+                        mount.UnixFileMode is not null &&
```
IEvangelist, line 1501: Evaluate the `mount` with pattern matching.

```suggestion
                    if (mount is { Type: ContainerMountType.BindMount, UnixFileMode: not null } &&
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+          - project: tests/Aspire.Hosting.Redis.Tests/Aspire.Hosting.Redis.Tests.csproj
+            name: Redis
+          - project: tests/Aspire.Hosting.Azure.Tests/Aspire.Hosting.Azure.Tests.csproj
+            name: Azure
+          - project: tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
+            name: Playground
+          # Add more projects as needed
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
IEvangelist, line 34: If the [dependabot.yml](https://github.com/dotnet/aspire/blob/main/.github/dependabot.yml#L65) is configured correctly, this would be updated daily. You could opt-in to using a commit sha instead.

```suggestion
        uses: actions/checkout@v4
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Set up .NET Core
+        uses: actions/setup-dotnet@v2
```
IEvangelist, line 37:
```suggestion
        uses: actions/setup-dotnet@v4
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+          dotnet-version: |
+            8.x
+            9.x
+
+      - name: Trust HTTPS development certificate
+        run: dotnet dev-certs https --trust
+
+      - name: Verify Docker is running
+        run: |
+          docker info
```
IEvangelist, line 48:
```suggestion
        run: docker info
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Install Azure Functions Core Tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y azure-functions-core-tools-4
+
+      - name: Restore dependencies
+        run: |
+          dotnet restore ${{ matrix.project }}
```
IEvangelist, line 57:
```suggestion
        run: dotnet restore ${{ matrix.project }}
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+            --blame-hang-timeout 7m \
+            --results-directory testresults \
+            --no-restore \
+            --no-build \
+            /p:CI=false
+
+      - name: Compress test results
+        if: always()
+        run: |
+          zip -r testresults.zip testresults
```
IEvangelist, line 80:
```suggestion
        run: zip -r testresults.zip testresults
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+            --logger html \
+            --blame \
+            --blame-hang-timeout 7m \
+            --results-directory testresults \
+            --no-restore \
+            --no-build \
+            /p:CI=false
+
+      - name: Compress test results
+        if: always()
```
IEvangelist, line 78: Will there always be results to compress? If not, you could make this conditional on the result of the `run_tests` step.

```suggestion
        if: always()
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Restore dependencies
+        run: |
+          dotnet restore ${{ matrix.project }}
+
+      - name: Build projects
+        run: |
+          dotnet build ${{ matrix.project }} /p:CI=false --no-restore
+
+      - name: Run tests
+        id: run_tests
```
IEvangelist, line 64: You can name it with an underscore, but hyphens are also supported. I see this isn't being used though either, which you might want to use it - or remove it.

```suggestion
        id: run-tests
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+          sudo apt-get update
+          sudo apt-get install -y azure-functions-core-tools-4
+
+      - name: Restore dependencies
+        run: |
+          dotnet restore ${{ matrix.project }}
+
+      - name: Build projects
+        run: |
+          dotnet build ${{ matrix.project }} /p:CI=false --no-restore
```
IEvangelist, line 61: For single line commands, no need to `|` unless you plan on these growing? I only recommend this way since the cert call above was on a single line.

```suggestion
        run: dotnet build ${{ matrix.project }} /p:CI=false --no-restore
```

PR #7014, opened 2025-01-03
Improve Elasticsearch functional tests reliability

sebastienros:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/aspire/pull/7014)

PR #7086, opened 2025-01-13
Make Elasticsearch Function tests more stable

eerhardt:
Don't use polly with 10 second delays for 10 retries (which is too short). Instead use the health check to wait until the resource is healthy, let the test run until the global timeout (10 minutes) expires.


sebastienros (approved):

davidfowl (approved):

eerhardt:
Closing in favor of #7014

PR #7065, opened 2025-01-10
Fix CosmosDb sample endpoint in xml docs

sebastienros:
## Description

I believe this was taken from storage queues. I followed the patter from https://learn.microsoft.com/en-us/rest/api/cosmos-db/cosmosdb-resource-uri-syntax-for-rest instead.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Is this introducing a breaking change?
      - [ ] Yes
        - Link to aspire-docs issue (please use this [`breaking-change` template](https://github.com/dotnet/docs-aspire/issues/new?template=04-breaking-change.yml)):
      - [ ] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)):
  - [x] No

PR #7092, opened 2025-01-13
Add support for managed identity in Azure Cosmos DB hosting component

### LINKED ISSUE #3792: [Cosmos .NET SDK] - Introduce Managed Identity authentication for Cosmos
- Introduce AAD Based Authentication for control plane.

sebastienros:
## Description

Fixes #3792

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -205,7 +156,7 @@ public static IResourceBuilder<AzureCosmosDBEmulatorResource> WithGatewayPort(th
     /// <param name="builder">Builder for the Cosmos emulator container</param>
     /// <param name="count">Desired partition count.</param>
     /// <returns>Cosmos emulator resource builder.</returns>
-    /// <remarks>Not calling this method will result in the default of 25 partitions. The actual started partitions is always one more than specified.
+    /// <remarks>Not calling this method will result in the default of 10 partitions. The actual started partitions is always one more than specified.
```
sebastienros, line 159: 11 (actual) is the default number of partitions when nothing is set, so updated to 10 since the actual is argument + 1

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -224,9 +175,151 @@ public static IResourceBuilder<AzureCosmosDBEmulatorResource> WithPartitionCount
...
+            cosmosAccount.DisableLocalAuth = true;
+
+            var principalIdParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalId, typeof(string));
+            infrastructure.Add(principalIdParameter);
+
+            cosmosAccount.CreateRoleAssignment(CosmosDBBuiltInRole.DocumentDBAccountContributor, RoleManagementPrincipalType.ServicePrincipal, principalIdParameter);
+
+            infrastructure.Add(new ProvisioningOutput("connectionString", typeof(string))
+            {
+                Value = BicepFunction.Interpolate($"https://{cosmosAccount.Name}.document.azure.com")
```
davidfowl, line 321: This doesn't seem like the right thing to do.
```suggestion
                Value = cosmosAccount.DocumentEndpoint
```

sebastienros: Done and tested

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
```diff
@@ -4,8 +4,9 @@
 var builder = DistributedApplication.CreateBuilder(args);
 
 var db = builder.AddAzureCosmosDB("cosmos")
-                .AddDatabase("db")
-                .RunAsEmulator();
+                .WithDatabase("db", database => database.Containers.Add(new("entries", "/Id")))
+                .RunAsEmulator()
+                ;
```
eerhardt, line 9:
```suggestion
                .RunAsEmulator();
```

This looks like a mistake

src/Components/Aspire.Microsoft.Azure.Cosmos/MicrosoftAzureCosmosSettings.cs
```diff
@@ -25,7 +25,7 @@ public sealed class MicrosoftAzureCosmosSettings
 
     /// <summary>
     /// A <see cref="Uri"/> referencing the Azure Cosmos DB Endpoint.
-    /// This is likely to be similar to "https://{account_name}.queue.core.windows.net".
+    /// This is likely to be similar to "https://{account_name}.documents.azure.com".
```
eerhardt, line 28: I thought these were fixed with https://github.com/dotnet/aspire/pull/7065?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -14,15 +16,27 @@ public class AzureCosmosDBResource(string name, Action<AzureResourceInfrastructu
...
+    internal List<CosmosDBDatabase> Databases { get; } = [];
 
     internal EndpointReference EmulatorEndpoint => new(this, "emulator");
 
     /// <summary>
     /// Gets the "connectionString" reference from the secret outputs of the Azure Cosmos DB resource.
+    ///
+    /// This is used when Entra ID authentication is used. The connection string is an output of the bicep template.
     /// </summary>
     public BicepSecretOutputReference ConnectionString => new("connectionString", this);
```
eerhardt, line 28: Isn't this supposed to be:

```suggestion
    public BicepOutputReference ConnectionString => new("connectionString", this);
```

It's not a secret reference anymore - it is just a normal bicep output.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
```diff
@@ -0,0 +1,57 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Aspire.Hosting.Azure.CosmosDB;
+
+/// <summary>
+/// This health check also creates default databases and containers for the Azure CosmosDB Emulator.
```
eerhardt, line 10: This feels like too much responsibility. Everytime we are checking for "health" we are rechecking all the dbs and containers.

I think it would be better if the health check was only a health check. And then we "waited" for it to be healthy and only ensure the db and containers are created once.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
```diff
@@ -0,0 +1,57 @@
...
+        try
+        {
+            var cosmosClient = _clientFactory();
+
+            await cosmosClient.ReadAccountAsync().WaitAsync(cancellationToken).ConfigureAwait(false);
+
+            var databases = _databasesFactory();
+
+            if (databases.Length != 0)
+            {
```
eerhardt, line 38: This isn't necessary. You can just `foreach` and it won't run the loop when it is empty.

src/Aspire.Hosting.Azure.CosmosDB/CosmosDBDatabase.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CosmosDBDatabase"/> class.
+    /// </summary>
+    public CosmosDBDatabase(string name)
+    {
+        Name = name;
+    }
+
+    /// <summary>
+    /// The topic name.
```
eerhardt, line 23:
```suggestion
    /// The database name.
```

tests/Aspire.Hosting.Azure.Tests/AzureCosmosDBEmulatorFunctionalTests.cs
```diff
@@ -248,6 +249,50 @@ await pipeline.ExecuteAsync(async token =>
...
+
+        var cosmos = builder.AddAzureCosmosDB("cosmos")
+                            .WithDatabase(databaseName, db => db.Containers.Add(new(containerName, partitionKeyPath)))
+                            .RunAsEmulator();
+
+        using var app = builder.Build();
+
+        await app.StartAsync(cts.Token);
+
+        await app.WaitForTextAsync("Started").WaitAsync(cts.Token);
```
eerhardt, line 273: What's the point of this line? We don't do this in any other test I've seen.

tests/Aspire.Hosting.Azure.Tests/AzureCosmosDBEmulatorFunctionalTests.cs
```diff
@@ -248,6 +249,50 @@ await pipeline.ExecuteAsync(async token =>
...
+
+        // This needs to be outside the pipeline because when the CosmosClient is disposed,
+        // there is an exception in the pipeline
+        using var cosmosClient = host.Services.GetRequiredService<CosmosClient>();
+
+        var database = cosmosClient.GetDatabase(databaseName);
+        var container = database.GetContainer(containerName);
+
+        Assert.NotNull(database);
+        Assert.NotNull(container);
```
eerhardt, line 294: Is this actually making a connection to the Cosmos DB emulator? There is not `await` here, so I assume this will pass whether the db and container are created or not.

tests/Aspire.Hosting.Azure.Tests/AzureCosmosDBEmulatorFunctionalTests.cs
```diff
@@ -248,6 +249,50 @@ await pipeline.ExecuteAsync(async token =>
...
+        var hb = Host.CreateApplicationBuilder();
+        hb.Configuration[$"ConnectionStrings:{cosmos.Resource.Name}"] = await cosmos.Resource.ConnectionStringExpression.GetValueAsync(default);
+        hb.AddAzureCosmosClient(cosmos.Resource.Name);
+
+        using var host = hb.Build();
+
+        await host.StartAsync(cts.Token);
+
+        // This needs to be outside the pipeline because when the CosmosClient is disposed,
+        // there is an exception in the pipeline
```
eerhardt, line 287: outside of what pipeline?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -136,11 +80,18 @@ public static IResourceBuilder<AzureCosmosDBResource> RunAsEmulator(this IResour
             cosmosClient = CreateCosmosClient(connectionString);
         });
 
+        // Use custom health check that also seeds the databases and containers
         var healthCheckKey = $"{builder.Resource.Name}_check";
-        builder.ApplicationBuilder.Services.AddHealthChecks().AddAzureCosmosDB(sp =>
```
eerhardt, line 140: Do we still need a reference on `AspNetCore.HealthChecks.CosmosDb` if we aren't using it?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -224,9 +175,151 @@ public static IResourceBuilder<AzureCosmosDBEmulatorResource> WithPartitionCount
...
+            infrastructure.Add(secret);
+        }
+        else
+        {
+            cosmosAccount.DisableLocalAuth = true;
+
+            var principalIdParameter = new ProvisioningParameter(AzureBicepResource.KnownParameters.PrincipalId, typeof(string));
+            infrastructure.Add(principalIdParameter);
+
+            cosmosAccount.CreateRoleAssignment(CosmosDBBuiltInRole.DocumentDBAccountContributor, RoleManagementPrincipalType.ServicePrincipal, principalIdParameter);
```
eerhardt, line 317: Why isn't this generating a prinicpalType like all the other Azure integrations?

https://github.com/dotnet/aspire/blob/882ae2be3e02edbfc1d222ee97fe2da4eea3fe99/src/Aspire.Hosting.Azure.AppConfiguration/AzureAppConfigurationExtensions.cs#L38-L43

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
sebastienros: That would be a breaking change. How do you suggest we deal with it?

eerhardt: This is why I opt to not make things public unless we absolutely need to make them public.

I would suggest just making the breaking change, and document it. The fact that we are switching to managed identity by default is going to be an even bigger breaking change.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
sebastienros: The default health check is checking only one db and its containers (because its state is only for one db, I assume one could create more instances for other dbs). At least we could do the same, pick the first db.

But I did what you suggested at first, though this could introduce undesired effects that it's stated as healthy but we haven't created the db/containers yet. So while we are creating them other services could have started using them.

Another option would be to change the implementation to do like it does right now, but then store a bool and only check for the first db afterwards (not even the containers).

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
sebastienros: The method to create the role on cosmos doesn't take any. It's working though, you think it must still be missing?

davidfowl: Yes this should be a parameter, @tg-msft?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
davidfowl: Actually, we can just expose another property

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -14,15 +16,27 @@ public class AzureCosmosDBResource(string name, Action<AzureResourceInfrastructu
...
 
+    /// <summary>
+    /// Gets the "connectionString" secret output reference from the bicep template for the Azure Redis resource.
+    ///
+    /// This is set when access key authentication is used. The connection string is stored in a secret in the Azure Key Vault.
+    /// </summary>
+    internal BicepSecretOutputReference? ConnectionStringSecretOutput { get; set; }
+
+    [MemberNotNullWhen(true, nameof(ConnectionStringSecretOutput))]
+    internal bool UseAccessKeyAuthentication => ConnectionStringSecretOutput is not null;
```
davidfowl, line 38: Maybe this should be public too.

davidfowl:
Do we want this to be a breaking change?

eerhardt:
> Do we want this to be a breaking change?

At some point, yes. We want managed identity to be the default everywhere, and this is the last integration that still uses secrets by default.

So the question in my mind is: "When do we want to make the breaking change?". I think now is a good time. But I can see an argument to wait until the next major version.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
sebastienros: I must have been on drugs, it's right there now. Weird as I remember changing this code as it was not working ...

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
sebastienros: Removed, replaced with `Microsoft.Azure.Cosmos` which was transitive before. Also added `NewtonSoft.Json` since there is a target that checks we include it. It was a dependency of `Microsoft.Azure.Cosmos` in 3.45.2.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
sebastienros: Added a flag to only check/create for the resources once.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
davidfowl: This shouldn't be obsolete right?

sebastienros: I wanted to get some feedback on it. So just add a new property without obsoleting the previous one? Just to have a better type? I personally prefer to obsolete but I may not think about some implications.

davidfowl: This is still valid in the access key mode.

sebastienros: OK, now I understand

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
sebastienros: Checking the other resources with the same model (Redis, Postgres) they don't expose these publicly. Instead the `ConnectionStringExpression` is available and returns the one that corresponds to managed identity / access key.

I don't see how CosmosDB is different. (the emulator is equivalent to redis/postgres local containers, and the three have an Azure resource).

davidfowl:
Can we decide if we are going to make the breaking change in 9.1? If yes, we should make sure it's easy to get back to the previous state (and we should validate it)

sebastienros:
> Can we decide if we are going to make the breaking change in 9.1? If yes, we should make sure it's easy to get back to the previous state (and we should validate it)

This is done with `AddAzureCosmosDB("cosmos").WithAccessKeyAuthentication()`

Here is the test that verifies that adding it behaves the same way as before in terms of bicep resources. I also validated it in playground.

https://github.com/dotnet/aspire/pull/7092/files#diff-1d15fcad74e27b197d015f339f523828ef1f59378caa471c4bb8f19d133095efR250

davidfowl:
OK let's go with it. Open up a breaking change issue?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
davidfowl: This seems like it's the only point of contention in the PR. Is it resolved? The other changes look good.

sebastienros: We should have a live chat about it, I tried different options. Right now it does the creation attempt only once then falls back to the standard check. This is just the apphost health check though to report for `WaithHealthyAsync`, the clients use the standard health check.

davidfowl (approved):

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBEmulatorHealthCheck.cs
davidfowl: I approved the PR, can you open an issue? It's kinda related to  https://github.com/dotnet/aspire/issues/7101

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -24,6 +25,23 @@ public class AzureCosmosDBResource(string name, Action<AzureResourceInfrastructu
     /// </summary>
     public BicepSecretOutputReference ConnectionString => new("connectionString", this);
```
eerhardt, line 26: Is this property used anymore? I don't see it being used now that we have `internal BicepSecretOutputReference? ConnectionStringSecretOutput { get; set; }`.

sebastienros: I don't think it is, but it's public so couldn't remove

sebastienros: As a reminder, it started with you asking to change the type to `BicepOutputReference`, which we couldn't so we don't break, and we created the new one.

eerhardt: Can we make it nullable and log it as a breaking change?

eerhardt: A very minor breaking change that I don't think would affect anyone.

Issue #6678, opened 2024-11-14
Improve test stability issues caused by resources failing to start due to "address in use" errors


DamianEdwards:
Found in the [dotnet/aspire-samples](https://github.com/dotnet/aspire-samples) tests (e.g. [this CI failure](https://github.com/dotnet/aspire-samples/actions/runs/11843098774)).

Sometimes when running app host integration tests using `DistributedApplicationTestingBuilder`, resources can fail to start due to the port that was randomly assigned by DCP already being in use by the time the resource goes to start:

![Image](https://github.com/user-attachments/assets/000a8f9e-d8ed-4362-989b-568a869db363)

This results in a test that fails, but that will most likely pass on re-run. We should look at how we can improve this so that tests are more reliable, e.g. doing automatic retries when resources fail to start.

KennethHoff:
Is there something to be done about the "address in use" problem in general? I've had to resort to creating a script that first executes `kill` on various processes before starting Aspire in order to make restarts more palettable. This could presumably also be solved by making hot reloading more functional; not requiring a full restart when you change/add a new environment variable for example.

DamianEdwards:
> Is there something to be done about the "address in use" problem in general?

The challenge is always that it's inherently racy if you want to know what the port is before you assign it to the process. Also it's not clear when it does fail if it's due to some other random process using the port or a previous instance of something launched by Aspire that wasn't shut down properly, or even random collisions in situations like parallel startup. I think adding some kind of support for startup retries could help here. I might attempt to see if I can add some startup-failure detection and retry logic to the testing infrastructure in dotnet/aspire-samples at the app host level, e.g. when a resource state changes to FailedToStart, scrape its logs and look for "port/address already in use" messages and in those cases, issue a "Start" command.

> Aspire being bad at ending its child processes

This was improved in between 9.0.0-rc.1 and 9.0.0, so I'd love to hear if you're seeing any improvement after updating to the latest version.

> This could presumably also be solved by making hot reloading more functional; not requiring a full restart when you change/add a new environment variable for example.

For sure, this is something we're thinking of improving but likely more in the mid-to-long term rather than short term.

yoDon:
This issue is not just a "test stability issue" Aspire 9 seems to have lots of problems stopping projects. In Aspire 8, I routinely started and stopped solutions without issue. In Aspire 9 on OSX with Podman Desktop and the Rider Aspire plugin, about half the time there are stuck ports preventing Aspire from starting.

[...]

Issue #6561, opened 2024-10-30
Tests keep DCP instances running until the whole test run ends


karolz-ms:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When running multiple tests that use app host/DCP, DCP instances (and associated resources such as containers) are kept running until the test run ends. They are not shut down, and resources are not freed, until all the tests complete running. This puts unnecessary strain on the machine and contributes to test instability.

### Expected Behavior

What should happen instead is that as soon as a test ends, regardless of the outcome, the DCP instance is allowed to do a graceful shutdown and has the opportunity to release all resources.

### Steps To Reproduce

1. Clone `dotnet/Aspire-Samples` GitHub repository
2. Run `build.cmd`
3. Run all tests
4. Open Task Manager and observe `dcp`/`dcpctrl` instances piling up as the tests are executed.

DamianEdwards:
What is the lifetime relationship between the app host process and the launched dcp processes today? Does the app host process request the shutdown of dcp, or does dcp monitor the launching process and shut itself down when the launching process ends?

karolz-ms:
There is no good way for the app host process to request DCP shutdown, so the second part is true (dcp monitor the launching process and shut itself down when the launching process ends). Part of the fix for this will be to add a new type of request to DCP that will allow the (test) app host to say "I am done, please perform resource cleanup and shut yourself down".

davidfowl:
Fixed by https://github.com/dotnet/aspire/pull/7098

PR #7098, opened 2025-01-14
Enable programmatic DCP shut down

karolz-ms:
## Description

This change takes advantage of the just-introduced DCP capability for API-triggered shutdown. This should significantly reduce resource consumption when running tests. Also removes some dead code and increases timeouts for tests using the app orchestrator which were unrealistically short (the latter applies to development test run only).

Note: if the programmatic shutdown fails for some reason (unlikely), the existing mechanism of DCP watching the AppHost process will still kick in. Exactly the same resource cleanup code is executed in either case. 

Fixes # (issue)
https://github.com/dotnet/aspire/issues/6678
https://github.com/dotnet/aspire/issues/6561

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
@@ -190,6 +190,25 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
         {
             _logger.LogDebug(ex, "One or more monitoring tasks terminated with an error.");
         }
+
+        try
+        {
+            // The app orchestrator (represented by kubernetesService here) will perform a resource cleanup
+            // (if not done already) when the app host process exits.
+            // This is just a perf optimization, so we do not care that much if this call fails.
+            // There is not much difference for single app run, but for tests that tend to launch multiple instances
+            // of app host from the same process, the gain from programmatic orchestrator shutdown is significant
+            // See https://github.com/dotnet/aspire/issues/6561 for more info.
+            await kubernetesService.StopServerAsync(Model.ResourceCleanup.Full, cancellationToken).ConfigureAwait(false);
+        }
+        catch (OperationCanceledException)
+        {
+            // Ignore.
+        }
+        catch (Exception ex)
+        {
+            _logger.LogDebug(ex, "Application orchestrator could not be stopped programmatically.");
+        }
     }
 
     private static ILookup<IResource?, IResourceWithParent> GetParentChildLookup(DistributedApplicationModel model)
@@ -479,8 +498,8 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
     {
         IAsyncEnumerable<IReadOnlyList<(string, bool)>>? enumerable = resource switch
         {
-            Container c when c.LogsAvailable => new ResourceLogSource<T>(_logger, kubernetesService, _dcpInfo?.Version, resource),
-            Executable e when e.LogsAvailable => new ResourceLogSource<T>(_logger, kubernetesService, _dcpInfo?.Version, resource),
+            Container c when c.LogsAvailable => new ResourceLogSource<T>(_logger, kubernetesService, resource),
+            Executable e when e.LogsAvailable => new ResourceLogSource<T>(_logger, kubernetesService, resource),
             _ => null
         };
 
@@ -1982,49 +2001,6 @@ internal sealed class ApplicationExecutor(ILogger<ApplicationExecutor> logger,
         // RandomNameSuffixLength of lowercase characters
         var suffix = PasswordGenerator.Generate(RandomNameSuffixLength, true, false, false, false, RandomNameSuffixLength, 0, 0, 0);
         return suffix;
-    }
-
-    public async Task DeleteResourcesAsync(CancellationToken cancellationToken = default)
-    {
-        try
-        {
-            AspireEventSource.Instance.DcpModelCleanupStart();
-            await DeleteResourcesAsync<ExecutableReplicaSet>("project", cancellationToken).ConfigureAwait(false);
-            await DeleteResourcesAsync<Executable>("project", cancellationToken).ConfigureAwait(false);
-            await DeleteResourcesAsync<Container>("container", cancellationToken).ConfigureAwait(false);
-            await DeleteResourcesAsync<Service>("service", cancellationToken).ConfigureAwait(false);
-        }
-        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
-        {
-            // Expected
-            _logger.LogDebug("Cancellation received while deleting resources.");
-        }
-        finally
-        {
-            AspireEventSource.Instance.DcpModelCleanupStop();
-            _appResources.Clear();
-        }
-    }
-
-    private async Task DeleteResourcesAsync<TResource>(string resourceType, CancellationToken cancellationToken) where TResource : CustomResource
-    {
-        var resourcesToDelete = _appResources.Select(r => r.DcpResource).OfType<TResource>();
-        if (!resourcesToDelete.Any())
-        {
-            return;
-        }
-
-        foreach (var res in resourcesToDelete)
-        {
-            try
-            {
-                await kubernetesService.DeleteAsync<TResource>(res.Metadata.Name, res.Metadata.NamespaceProperty, cancellationToken).ConfigureAwait(false);
-            }
-            catch (Exception ex)
-            {
-                _logger.LogInformation(ex, "Could not stop {ResourceType} '{ResourceName}'.", resourceType, res.Metadata.Name);
-            }
-        }
     }
 
     /// <summary>
--- a/src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
+++ b/src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
@@ -1,6 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Net;
+using System.Net.Http.Headers;
+using Aspire.Hosting.Dcp.Model;
 using k8s;
 using k8s.Autorest;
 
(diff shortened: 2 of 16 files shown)


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1984,49 +2003,6 @@ private static string GetRandomNameSuffix()
         return suffix;
     }
 
-    public async Task DeleteResourcesAsync(CancellationToken cancellationToken = default)
```
karolz-ms, line 1987: This is dead code, remnant from early Aspire lifetime. Never worked properly.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -190,6 +190,25 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
+
+        try
+        {
+            // The app orchestrator (represented by kubernetesService here) will perform a resource cleanup
+            // (if not done already) when the app host process exits.
+            // This is just a perf optimization, so we do not care that much if this call fails.
+            // There is not much difference for single app run, but for tests that tend to launch multiple instances
+            // of app host from the same process, the gain from programmatic orchestrator shutdown is significant
+            // See https://github.com/dotnet/aspire/issues/6561 for more info.
+            await kubernetesService.StopServerAsync(Model.ResourceCleanup.Full, cancellationToken).ConfigureAwait(false);
```
davidfowl, line 202: Does this block?

karolz-ms: It starts the cleanup process but does not wait for it to finish. So no.

tests/Aspire.Hosting.Tests/HealthCheckTests.cs
```diff
@@ -29,7 +29,7 @@ public async Task WithHttpHealthCheckThrowsIfReferencingEndpointThatIsNotHttpSch
         var ex = await Assert.ThrowsAsync<DistributedApplicationException>(async () =>
         {
             await app.StartAsync();
-        }).DefaultTimeout();
+        }).DefaultTimeout(TestConstants.DefaultOrchestratorTestTimeout);
```
JamesNK, line 32: Are extra constants needed? LongTimeoutDuration is 20 seconds locally and 120 seconds in CI. That seams like it is enough. What takes over 20 seconds locally? We could make the long duration longer if needed.

I don't know what is the right thing for Default, Long (and a new bigger value if required?) timeout consts vs having values explicitly for one product area. I worry that someone will go though and place orchestrator timeouts everywhere just because they're writing code in that area vs using the defaults which suit them.

tests/Aspire.Hosting.Tests/Dcp/TestKubernetesService.cs
```diff
@@ -143,4 +143,19 @@ public Task<T> PatchAsync<T>(T obj, V1Patch patch, CancellationToken cancellatio
     {
         return Task.FromResult(obj);
     }
+
+    public Task StopServerAsync(string resourceCleanup = "Full", CancellationToken cancellation = default)
+    {
+        cancellation.ThrowIfCancellationRequested();
+
+        lock(CreatedResources)
```
JamesNK, line 151: nit
```suggestion
        lock (CreatedResources)
```

src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
```diff
@@ -79,6 +82,33 @@ public async Task<HttpOperationResponse<Stream>> ReadSubResourceAsStreamAsync(
         return result;
     }
 
+    /// <summary>
+    /// PATCH Execution document (part of the DCP administrative interface) with supplied data. 
+    /// </summary>
+    public Task<HttpResponseMessage> PatchExecutionDocumentAsync(
+        ApiServerExecution apiServerExecution,
+        CancellationToken cancellationToken = default
+    )
```
JamesNK, line 91:
```suggestion
        CancellationToken cancellationToken = default)
```

src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
```diff
@@ -79,6 +82,33 @@ public async Task<HttpOperationResponse<Stream>> ReadSubResourceAsStreamAsync(
...
+            Method = HttpMethod.Patch,
+            RequestUri = new Uri(this.BaseUri, url)
+        };
+        httpRequest.Version = HttpVersion.Version20;
+
+        var content = KubernetesJson.Serialize(apiServerExecution, null);
+        httpRequest.Content = new StringContent(content, System.Text.Encoding.UTF8);
+        httpRequest.Content.Headers.ContentType = MediaTypeHeaderValue.Parse("application/merge-patch+json");
+        return SendRequestRaw(content, httpRequest, cancellationToken);
+
```
JamesNK, line 109:
```suggestion
```

src/Aspire.Hosting/OtlpConfigurationExtensions.cs
```diff
@@ -65,15 +65,7 @@ public static void AddOtlpEnvironment(IResource resource, IConfiguration configu
             // Set the service name and instance id to the resource name and UID. Values are injected by DCP.
             var dcpDependencyCheckService = context.ExecutionContext.ServiceProvider.GetRequiredService<IDcpDependencyCheckService>();
             var dcpInfo = await dcpDependencyCheckService.GetDcpInfoAsync(context.CancellationToken).ConfigureAwait(false);
-            if (dcpInfo?.Version?.CompareTo(DcpVersion.MinimumVersionAspire_8_1) >= 0)
```
davidfowl, line 68: It's so nice we no longer have to worry about this versioning ❤️

tests/Aspire.Hosting.Tests/HealthCheckTests.cs
davidfowl: the right timeout when debugging is none 😄

JamesNK: https://github.com/dotnet/aspire/blob/0f8b4c1ad94d1a1973269abf5a9bc4db34c73f6e/tests/Shared/AsyncTestHelpers.cs#L98-L103

😄

davidfowl: Lets jut increase the default to something high

JamesNK: The default timeout being high on the server is fine, and the long timeout being high is fine locally and on the server, but when running tests you don't want to wait 60 seconds because something will never happen. It's a productivity sink.

Default timeout should fail fast in 5 seconds to not waste the developers time.

davidfowl: No it should waste developers time. It should never happened locally and when it does there are lots tools you can use.

if you don’t make it high then we won’t be using it in certain classes of tests and it’ll defeat the entire purpose

davidfowl:
@karolz-ms can you remove these in this PR:

https://github.com/dotnet/aspire/blob/66465d1e19f20616be209d4d77a019aa4b68ea75/tests/helix/send-to-helix-inner.proj#L125-L139

@joperezr can you review those changes please? I want to make sure we can run without cleaning up to see if things are working well with the latest DCP changes plus these new app model changes.

tests/Aspire.Hosting.Tests/HealthCheckTests.cs
karolz-ms: The primary goal of this change is to eliminate false negatives when I am verifying a change by running multiple tests from the IDE, via TestExplorer. With the current timeout values, on my daily go-to dev machine (64 GB RAM, SSD drive, not too old) running `DistributedApplicationTests` results in a handful of tests timing out pretty much **on every run**.  I am wasting a lot of time re-running these failed tests.

Actual test debugging is not an issue with longer timeouts in my experience. When I need a tight inner loop for debugging a particular test, I can just lower/extend individual timeouts as necessary; that scenario happens much less often than "run all test to verify nothing is broken", so I want to optimize for the latter.

There are many reasons why 5 seconds for "short" and 20 seconds for "long" may not be enough for tests that involve DCP. For example, during startup DCP is invoked not 1, but 3 times and does probing for container runtime. If container runtime is "sleeping" that probe may take a while. Anything involving a container is subject to image download/image build delays, and involves retries that DCP does to work-around container orchestrator flakiness. Tests running in parallel affecting each other. Antivirus having to scan new build of Aspire/DCP. Etc. etc.

src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
karolz-ms: done

src/Aspire.Hosting/Dcp/DcpKubernetesClient.cs
karolz-ms: done

davidfowl (approved):
Lets see if this destabilizes the CI 😄 

cc @eerhardt

tests/Aspire.Hosting.Tests/HealthCheckTests.cs
JamesNK: > No it should waste developers time. It should “never” happened locally and when it does there are lots tools you can use (like a debugger)
> 
> if you don’t make it high then we won’t be using it in certain classes of tests and it’ll defeat the entire purpose

Yeah I know that timeouts should never happen when tests are in a good state, but deving code can break things.

When you run tests locally to double check things are good, you want broken tests to fail in 5 seconds with a timeout rather than a minute.

tests/helix/send-to-helix-inner.proj
```diff
@@ -122,22 +122,6 @@
...
-  <ItemGroup Condition="'$(HasDocker)' == 'true'">
-    <HelixPreCommand Include="docker info" />
-    <HelixPreCommand Include="docker ps" />
-    <HelixPreCommand Include="docker container ls --all" />
-    <HelixPreCommand Include="$(_ShutdownDockerContainersCommand)" />
-    <HelixPreCommand Include="docker volume ls" />
-    <HelixPreCommand Include="$(_DeleteDockerVolumesCommand)" />
-    <HelixPreCommand Include="docker network prune -f" />
-
-    <HelixPostCommand Include="docker container ls --all" />
```
eerhardt, line 134: This is useful in situations where the container crashes, to see what is happening on the machine. It is also nice to guarantee somewhat stable machines.

Do we have to delete this?

cc @radical

davidfowl: While that is true, it's nicer if we can actually find bugs instead of having the CI cover up dcp problems 😄

eerhardt: Can we at least add back the diagnostics lines? If you don't want the prune/clean calls for now, that's fine.

davidfowl: Yep that's reasonable

davidfowl: @karolz-ms can you add the diagnostic ones back?

davidfowl:
Since this change went in I'm seeing containers stuck in the starting state and a large set of new tests failing as a result. Lets add the diagnostics back so we can see if things are not being cleaned up properly in some cases. I wonder if some tests are not calling StopAsync properly and things still aren't being cleaned up.

tests/helix/send-to-helix-inner.proj
karolz-ms: Working on it

karolz-ms: https://github.com/dotnet/aspire/pull/7141

joperezr:
> @karolz-ms can you remove these in this PR:
> 
> https://github.com/dotnet/aspire/blob/66465d1e19f20616be209d4d77a019aa4b68ea75/tests/helix/send-to-helix-inner.proj#L125-L139
> 
> @joperezr can you review those changes please? I want to make sure we can run without cleaning up to see if things are working well with the latest DCP changes plus these new app model changes.

Sorry @davidfowl just saw this. Changes look good other than @eerhardt comments about it being nice to have the diagnostic lines.

davidfowl:
We added them back, well @karolz-ms did 😄

davidfowl:
@karolz-ms Can confirm now that with the diagnostic messages added back, we didn't fully fix the issue with dcp. Containers now don't start on the CI. I'll submit a change that does cleanup on start, that way we can track if there are leaking networks still but tests won't fail.

cc @dbreshears

PR #7105, opened 2025-01-14
Allow specifying an Event Hub to use as the default EntityPath for namespace connection string

### LINKED ISSUE #7093: Allow specifying a default EntityPath for Azure Event Hubs / Azure Service Bus connection string

Can we add support for setting an EntityPath in the connection string now that the application model is stable. Having to set the EventHubName via the settings callback for integrations is bothersome. 

An API extension like the following seems the least intrusive:

`builder.AddAzureEventHubs("ns").WithHub("hub", configure => configure.IsDefaultEntity = true);`

...would result in the following connectionstring being available to integrations:

`Endpoint=sb://<NamespaceName>.servicebus.windows.net/;SharedAccessKeyName=<KeyName>;SharedAccessKey=<KeyValue>;EntityPath=hub`

Not setting a default entity should result in a connection string for the namespace only. The builder should throw [on build?] if more than one hub is set as default. I presume this should also work the same for ASB queues etc.

### Additional context

_No response_

oising:
## Description

- Add new extension method WithDefaultEntity(string)

Only one EventHub can be specified as the default entity path. The entity path can still be overridden by settings in integrations. This method can be called multiple times with the same name. 

- Add test to validate EntityPath is in connection string, and is well-formed
- Add test to validate setting multiple default entities is invalid
- Updated playground to remove settings usage (but left setting usage in other tests)
- Updated documentation in README.md for integrations to mention new functionality

An exception is thrown early if WithDefaultEntity sees that a hub has already been flagged.

### Manual testing (playground project):

- [x] Tested with emulator
- [x] Tested with local dev, deploy remote resources
- [x] Tested via `azd up` full deployment to azure

### Notes

I cleaned up the validation logic and tried to simplify it in the base client shared component `EventHubsComponent.cs` in the method `EnsureConnectionStringOrNamespaceProvided`. I also fixed an edge case bug that already existed for parsing. I added `#define` clauses to the playground for disabling the emulator to make it easier to test. There are also some `#define` conditionals in the projects to allow using AzureCli credential source, but functionally the playground project as it stands is unchanged.  

As the FQNS & token credential method doesn't work with docker and the emulator, I couldn't add local unit testing. That said, all scenarios were fully tested, and the code that extracts the event hub name hint for the FQNS method is shared among all five client types, so I'm confident it's solid. 

For those not super familiar with the quirks of event hub connection strings, this PR ensures that the SAS-key style connection string has the `EntityPath` keyword set if `WithDefaultEntity(string hubName)` is called:

i.e. `Endpoint=sb://localhost:57195;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;EntityPath=hub`

 For the FQNS (Uri) & Token Credential connection, this PR stashes a hint in the FQNS (fully qualified namespace) Uri via a QueryString parameter:

`https://foobar.servicebus.azure.net:443/?EntityPath=hub`

The client integration packages will extract this hint from the Url and assign it to the client's `settings.EventHubName` if it is found to be unset, thus using the prior tested code path of the settings callback. 
 
Fixes #7093

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No


oising:
If we can agree on tihs, I'll send a PR for the same thing in Azure Service Bus for queue/topic in EntityPath

playground/AspireEventHub/EventHubs.AppHost/Program.cs
```diff
@@ -7,7 +7,7 @@
 
 var eventHub = builder.AddAzureEventHubs("eventhubns")
     .RunAsEmulator()
-    .WithHub("hub");
+    .WithHub("hub", configure => configure.IsDefaultEntity = true);
```
davidfowl, line 10: nit: 

```suggestion
    .WithHub("hub", h => h.IsDefaultEntity = true);
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -200,7 +200,9 @@ public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResou
             // an event hub namespace without an event hub? :)
             if (builder.Resource.Hubs is { Count: > 0 } && builder.Resource.Hubs[0] is { } hub)
             {
-                var healthCheckConnectionString = $"{connectionString};EntityPath={hub.Name};";
+                var healthCheckConnectionString = connectionString.Contains(";EntityPath=") ?
```
davidfowl, line 203: Instead of Contains, we could parse it properly and build a new one.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -200,7 +200,9 @@ public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResou
             // an event hub namespace without an event hub? :)
             if (builder.Resource.Hubs is { Count: > 0 } && builder.Resource.Hubs[0] is { } hub)
```
davidfowl, line 201:
```suggestion
            if (builder.Resource.Hubs is [var  hub])
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
```diff
@@ -47,8 +47,29 @@ public class AzureEventHubsResource(string name, Action<AzureResourceInfrastruct
...
+            // Of one or more hubs, only one may be flagged as default
+            var defaultEntity = Hubs.Single(hub => hub.IsDefaultEntity);
+
+            return $";EntityPath={defaultEntity.Name}";
+        }
+        catch (InvalidOperationException ex)
+        {
+            throw new InvalidOperationException("Only one EventHub can be configured as the default entity.", ex);
+        }
+    }
```
davidfowl, line 72: 1. Maybe a bit cleaner to use the ReferenceExpressionBuilder and a single method to build the connection string like https://github.com/dotnet/aspire/blob/3cf908b66566cfb8e3112545d931a24d5d6cfe15/src/Aspire.Hosting.MongoDB/MongoDBServerResource.cs#L56
2. Do we want this to throw here? Or somewhere else?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
oising: Oh, that's a new syntax for me. Nice!

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: 1. Will take a look at this
2. That's my feeling too, but I'm not exactly sure where else and/or how to do that. Suggestions?

davidfowl: Did you change something?

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
oising: Done. FYI, I can parse to test for the entitypath, but there is nothing in dotnet to build connection strings.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: btw, I changed the final exception to DistributedApplicationException as this seems to be the thing everywhere else for top level throws

oising:
> Does this also work when deployed?

It's funny you should ask -- I had just deployed the playground project to test, and I found a bug. I'll report back when fixed.

oising:
Well, sh*t.

It turns out that there's no way to officially pass the event hub name when using the FQNS + TokenCredential for a real Event Hub.  It's fine for the emulator, because the non-tokencredential connection string allows you to pass an EntityPath. I had mistakely believed you could include the hub name as part of the Uri in the non-emulator case. 

One way to do it would be to add Aspire-specific handling of a FQNS that includes the hub name as a query string, e.g. `https://foobar.servicebus.windows.net?hubName`. But this would be setting a new precedent for Aspire I think, n.b. special casing parsing of connection strings in the Aspire clients. As for my thoughts on the matter, at first it felt off (and it does a little still) but it does not break compatibility in the connection strings for non-Aspire wrapped clients. 

Thoughts? 

@davidfowl @mitchdenny @eerhardt

davidfowl:
We can use our own connection string format that the client understands. This is what openai does

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
sebastienros: @oising we already do this in the code, example:

```c#
var props = EventHubsConnectionStringProperties.Parse(connectionString);
if (string.IsNullOrWhiteSpace(props.EventHubName)) ...
```

sebastienros:
> We can use our own connection string format that the client understands. This is what openai does

What about setting the ENV that will be bound to the `EntityPath` property in the settings on the client instead? I am doing this in a PR for openai to split the connection string from the models/deployments.

Another comment on this PR, late one sorry, I personally prefer calling a helper like `WithEntityPath(string?)` instead of setting a flag on any hub. I assume this way you could even set it if you don't create the hubs at the host level too?

oising:
> > We can use our own connection string format that the client understands. This is what openai does
> 
> What about setting the ENV that will be bound to the `EntityPath` property in the settings on the client instead? I am doing this in a PR for openai to split the connection string from the models/deployments.
> 
> Another comment on this PR, late one sorry, I personally prefer calling a helper like `WithEntityPath(string?)` instead of setting a flag on any hub. I assume this way you could even set it if you don't create the hubs at the host level too?

Hi @sebastienros - I had thought of that in an earlier point in time, but given that there are five different clients for eventhub, we'd have to inject five env vars. Also, they would take precedence over client appsettings which might be at best surprising and at worst possibly breaking, right? If we use a custom connection string then it's just a hint and the regular configuration should take precedence, if required

davidfowl:
I prefer a single connection string… This brings back our connection info discussion 🙂

oising:
> Reference in

Alrighty -- so I'll add a query path hint in the FQNS. And are we good with having WithEntityPath(<string>) ? It would just set IsDefaultEntity on the model, and I'd drop the visibility to internal on the field. I take it we don't want to start down the "two ways to do everything path"

davidfowl:
before making the change, can we compare both API samples?

oising:
@davidfowl 
> before making the change, can we compare both API samples?

## A
```csharp
var eventHub = builder.AddAzureEventHubs("eventhubns")
    .RunAsEmulator()
    .WithHub("hub1")
    .WithHub("hub2", h => h.IsDefaultEntity = true);
```

vs

## B
```csharp
var eventHub = builder.AddAzureEventHubs("eventhubns")
    .RunAsEmulator()
    .WithHub("hub1")
    .WithHub("hub2")
    .WithDefaultEntity("hub2"); // WithEntityPath ?
```

At first you might think why not `WithDefaultHub`? And the reasoning is if we bring the same API to service bus, you'd need WithDefaultQueue and WithDefaultTopic, which would be mutually exclusive. Using Entity is clearer, IMO. I prefer WithDefaultEntity over WithEntityPath, personally but could be convinced of either.

sebastienros:
I am fine with both approaches, my preference being an extension method (whatever name makes more sense), or an argument on the `AddAzureEventHubs()`. The value could then be stored in the event hubs resource. We may don't even have to deal with invalid state, in case you want to set the value with a hub name that doesn't need to be defined on the resource (like you can do today).

Does my suggestion to generate a separate ENV to flow to the settings make sense? It would be prefixed with the resource name to bind to the correct settings.

oising:
> Does my suggestion to generate a separate ENV to flow to the settings make sense? It would be prefixed with the resource name to bind to the correct settings.

Yes, I thought I had mentioned this above. It does make sense from a mechanical perspective, but like I had said, there are five different clients that ship in the integration and you'd have to inject an ENV var for all of them since you cannot know in the hosting side which clients will be used. Personally, I prefer to inline a hint in the connectionstring.

sebastienros:
> when someone tries to specify/override the event hub in appsetting.json

Isn't it of higher priority than ENVs? And then the configure callback has the most priority.

> there are five different clients that ship in the integration

I checked the code as I am not familiar with it, I see the issue is each component has its own settings class, but the `EventHubName` is set on the base one.

Would something like this work:

```c#
    protected override void BindSettingsToConfiguration(AzureMessagingEventHubsBufferedProducerSettings settings,
        IConfiguration configuration)
    {
        configuration.Bind((AzureMessagingEventHubsSettings)settings);
        configuration.Bind(settings);
    }
```

If not set in component specific section, it would take the one from the apphost config. `configureSettings` would still work. Bonus, now you can even define the defaults for any component.

oising:
> > when someone tries to specify/override the event hub in appsetting.json
> 
> Isn't it of higher priority than ENVs? And then the configure callback has the most priority.
> 

Nope. AppSettings are loaded first, then environment. Last one loaded wins. See:

https://github.com/aspnet/MetaPackages/blob/586d641934e997045ed98b1978684a7b0c607b39/src/Microsoft.AspNetCore/WebHost.cs#L152

> > there are five different clients that ship in the integration
> 
> I checked the code as I am not familiar with it, I see the issue is each component has its own settings class, but the `EventHubName` is set on the base one.
> 

Sure, it's on the base class in code, but the configuration binding is bound against the flattened view of the class hierarchy. 

> Would something like this work:
> 
> ```cs
>     protected override void BindSettingsToConfiguration(AzureMessagingEventHubsBufferedProducerSettings settings,
>         IConfiguration configuration)
>     {
>         configuration.Bind((AzureMessagingEventHubsSettings)settings);
>         configuration.Bind(settings);
>     }
> ```
> 
> If not set in component specific section, it would take the one from the apphost config. `configureSettings` would still work. Bonus, now you can even define the defaults for any component.

oising:
> > when someone tries to specify/override the event hub in appsetting.json
> 
> Isn't it of higher priority than ENVs? And then the configure callback has the most priority.
> 

Nope. AppSettings are loaded first, then environment. Last one wins. See:

https://github.com/aspnet/MetaPackages/blob/586d641934e997045ed98b1978684a7b0c607b39/src/Microsoft.AspNetCore/WebHost.cs#L152

> > there are five different clients that ship in the integration
> 
> I checked the code as I am not familiar with it, I see the issue is each component has its own settings class, but the `EventHubName` is set on the base one.
> 

Sure, it's on the base class in code, but the configuration binding is bound against the flattened view of the class hierarchy. 

> Would something like this work:
> 
> ```cs
>     protected override void BindSettingsToConfiguration(AzureMessagingEventHubsBufferedProducerSettings settings,
>         IConfiguration configuration)
>     {
>         configuration.Bind((AzureMessagingEventHubsSettings)settings);
>         configuration.Bind(settings);
>     }
> ```
> 
> If not set in component specific section, it would take the one from the apphost config. `configureSettings` would still work. Bonus, now you can even define the defaults for any component.

oising:
Marking this as draft while I update the integration bits and add some more tests. Note that I also found an existing bug in the current connection string validation, I'll add more details after.

davidfowl:
What happens when you use this feature with azure functions, none of this applies right?

cc @captainsafia

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -198,9 +225,27 @@ public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResou
             // For the purposes of the health check we only need to know a hub name. If we don't have a hub
             // name we can't configure a valid producer client connection so we should throw. What good is
             // an event hub namespace without an event hub? :)
-            if (builder.Resource.Hubs is { Count: > 0 } && builder.Resource.Hubs[0] is { } hub)
+            if (builder.Resource.Hubs is [var hub])
```
captainsafia, line 228: I think this will now throw if there is more than one hub registered on the resource? Is that desired behavior now?

davidfowl: Why would it throw?

captainsafia: `[var hub]` is a single-element match. If this fails, we'll fallback to this [else that throws](https://github.com/dotnet/aspire/blob/c5a98fa07c3cddc44600c50dba23559a0a4056f9/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs#L251-L254).

oising: That isn't what we want. I'll test this.

oising:
@davidfowl 
> What happens when you use this feature with azure functions, none of this applies right?
> 
> cc @captainsafia

It only improves the scenario for Functions that use the SASKey style that officially supports EntityPath. For those bound via an FQNS, I don't touch it since we don't have control over the client to parse out the hint.

oising:
Ooops, I accidentally checked in overrides in the playground to use AzureCliCredential for my local testing. I'll remove them in the next commit.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
oising: Good catch @captainsafia -- I'll revert the change.

oising:
Ready for review @davidfowl 

I've updated the PR description with more details to help with the review.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -89,30 +90,44 @@ protected static void EnsureConnectionStringOrNamespaceProvided(AzureMessagingEv
...
                 }
             }
         }
-        // If we have a namespace and no connection string, ensure there's an EventHubName
+        // Live: If we have a namespace and no connection string, ensure there's an EventHubName (also look for hint in FQNS)
         else if (!string.IsNullOrWhiteSpace(settings.FullyQualifiedNamespace) && string.IsNullOrWhiteSpace(settings.EventHubName))
+        {
+            if (Uri.TryCreate(settings.FullyQualifiedNamespace, UriKind.Absolute, out var fqns))
+            {
+                var query = System.Web.HttpUtility.ParseQueryString(fqns.Query);
```
davidfowl, line 115: What? 😄

oising: Old school! Also, blame copilot 😁

oising: Honestly, I didn't even notice that it was SystemWeb until your comment. I'll replace it at some point this afternoon.

oising: Fixed lol

davidfowl:
The thing that feels strange about this to be is that fact that we would mutate the original resource to have a default entity instead of making a child resource (like we had before), which could be a way to "deep link" (to use @mitchdenny's words) into an event hub instance. That way you aren't limited to one and there's no strange default behavior to implement.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -116,6 +117,32 @@ public static IResourceBuilder<AzureEventHubsResource> WithHub(this IResourceBui
...
     }
 
+    /// <summary>
+    /// Specifies that the named EventHub should be used as the EntityPath in the resource's connection string.
+    /// <remarks>Only one EventHub can be set as the default entity. If more than one is configured as default, an Exception will be raised at runtime.</remarks>
+    /// </summary>
+    /// <param name="builder">The Azure Event Hubs resource builder.</param>
+    /// <param name="name">The name of the Event Hub.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<AzureEventHubsResource> WithDefaultEntity(this IResourceBuilder<AzureEventHubsResource> builder, [ResourceName] string name)
```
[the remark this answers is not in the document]
oising: > Is there a reason we'd want to use this without having WithHub? Example: connecting to an existing azure resource without the need to call `WithHub` for each of the existing ones.

When you mentioned this first, I had to think about it and I don't believe there's a scenario where we'd want to set a default entity path without having any associated hubs in the resource. If we want to use a pre-existing resource that already has hubs, we would use AddConnectionString. Anything that uses AddEventHubs needs to provide at least one hub, or health checks will fail. I understand it is possible to want to create a namespace without hubs and have hubs created dynamically (i.e. some dapr patterns encourage dthis but you have to grant it management permissions.)

> After all `WithHub("foo")` doesn't ensure it actually exists on the resource either. Almost like we are asking users twice in that case.

The resource itself doesn't exist either -- all this does is mutate the model. At runtime, it's either projected into the emulator config, or emitted in the manifest. I don't think this is our concern?

> And related to this, as a user it might be nice to be able to call it twice, the second being one winning. That would mean store the hub name in the resource itself, instead of having a bool on each resource to verify the integrity (might be simpler too).

I'm not sure how this scenario ("nice to be able to call it twice") would arise beyond a mistake?

> Is "Default" in `DefaultEntity` something concrete? Why not called `WithEntityPath()`? (ignore me if I already asked ;) )

I say "default" because you can override it in the client settings.

oising:
> The thing that feels strange about this to be is that fact that we would mutate the original resource to have a default entity instead of making a child resource (like we had before), which could be a way to "deep link" (to use @mitchdenny's words) into an event hub instance. That way you aren't limited to one and there's no strange default behavior to implement.

I already started that PR before you guys got entangled in internal debate and switched to this model, so then I spent my time on this instead. This is frustrating as a contributor who isn't paid to do this. Ultimately, this is just a hint to to model which gets projected to emulator config, or into bicep via the manifest.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
```diff
@@ -45,10 +45,43 @@ public class AzureEventHubsResource(string name, Action<AzureResourceInfrastruct
...
+        // Of one or more hubs, only one may be flagged as default
+        var defaultEntity = Hubs.Single(hub => hub.IsDefaultEntity);
+
+        if (IsEmulator)
+        {
+            builder.Append($";EntityPath={defaultEntity.Name}");
+        }
+        else
+        {
+            builder.Append($"?EntityPath={defaultEntity.Name}");
```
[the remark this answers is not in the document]
oising: Good question -- but according to https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsofteventhub only `[a-zA-Z0-9\-]` is allowed for namespaces. But I'm not sure if accented characters are permitted. If so, then you're right.

playground/AspireEventHub/EventHubsApi/Program.cs
```diff
@@ -1,14 +1,19 @@
...
 
 builder.AddServiceDefaults();
 
-builder.AddAzureEventHubProducerClient("eventhubns", settings =>
-{
-    settings.EventHubName = "hub";
-});
+builder.AddAzureEventHubProducerClient("eventhubns"
+#if AZCLI
+    , settings => settings.Credential = new AzureCliCredential()
```
eerhardt, line 14: Why doesn't the DeafultAzureCredential work?

src/Components/Aspire.Azure.Messaging.EventHubs/EventProcessorClientComponent.cs
```diff
@@ -111,6 +111,7 @@ private static BlobContainerClient GetBlobContainerClient(
         }
 
         var containerClient = blobClient.GetBlobContainerClient(settings.BlobContainerName);
+
```
eerhardt, line 114:
```suggestion
```

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -89,30 +90,52 @@ protected static void EnsureConnectionStringOrNamespaceProvided(AzureMessagingEv
                 $"'ConnectionStrings:{connectionName}' or specify a 'ConnectionString' or 'FullyQualifiedNamespace' in the '{configurationSectionName}' configuration section.");
         }
 
-        // If we have a connection string, ensure there's an EntityPath if settings.EventHubName is missing
+        // Emulator: If we have a connection string, ensure there's an EntityPath if settings.EventHubName is missing
```
eerhardt, line 93:
```suggestion
        // If we have a connection string, ensure there's an EntityPath if settings.EventHubName is missing
```

This isn't Emulator specific (although the emulator does go down this path. You can use this library with a connection string to a real azure service.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -89,30 +90,52 @@ protected static void EnsureConnectionStringOrNamespaceProvided(AzureMessagingEv
...
 
-                // if EntityPath is missing, throw
-                if (string.IsNullOrWhiteSpace(props.EventHubName))
+                // if EntityPath is found, capture it
+                if (!string.IsNullOrWhiteSpace(props.EventHubName))
                 {
-                    throw new InvalidOperationException(
-                        $"A {typeof(TClient).Name} could not be configured. Ensure a valid EventHubName was provided in " +
-                        $"the '{configurationSectionName}' configuration section, or include an EntityPath in the ConnectionString.");
+                    // this is used later to create the checkpoint blob container
```
eerhardt, line 105: Is this comment correct? I am not sure it makes sense in the current context.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -89,30 +90,52 @@ protected static void EnsureConnectionStringOrNamespaceProvided(AzureMessagingEv
...
+                    settings.EventHubName = props.EventHubName;
                 }
             }
         }
-        // If we have a namespace and no connection string, ensure there's an EventHubName
+        // Live: If we have a namespace and no connection string, ensure there's an EventHubName (also look for hint in FQNS)
```
eerhardt, line 110:
```suggestion
        // If we have a namespace and no connection string, ensure there's an EventHubName (also look for hint in FQNS)
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
```diff
@@ -45,10 +45,43 @@ public class AzureEventHubsResource(string name, Action<AzureResourceInfrastruct
...
+        }
+
+        if (!Hubs.Any(hub => hub.IsDefaultEntity))
+        {
+            // Of zero or more hubs, none are flagged as default
+            return builder.Build();
+        }
+
+        // Of one or more hubs, only one may be flagged as default
+        var defaultEntity = Hubs.Single(hub => hub.IsDefaultEntity);
```
eerhardt, line 72: Can't we use `SingleOrDefault` and only do 1 check through the loop?

Also, we can do this before the above check, and only have a single `IsEmulator` check.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
eerhardt, line 80: How does the Azure.Messaging.EventHubs library handle having this query on the Uri?

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
oising: Yeah, it is needed later to create the container for checkpoints if it doesn't already exist.

playground/AspireEventHub/EventHubsApi/Program.cs
oising: I was getting tired of adding/removing this for my own scenario -- not everyone uses the same account on their laptop to access azure. I thought I could get away with leaving it as an "example" while serving my own selfish purposes. If you want me to remove it, I can.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: It doesn't care, which is why I ab(use) it.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: SingleOrDefault will return null for two or more event hubs.

playground/AspireEventHub/EventHubs.AppHost/Program.cs
```diff
@@ -1,13 +1,21 @@
+#define EMULATOR
```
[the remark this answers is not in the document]
oising: If it bothers you, i can.

playground/AspireEventHub/EventHubsApi/Program.cs
eerhardt: Yes, let's remove this clutter from the playground apps.

For this case, we have a vision of being able to define this via config (I'm not sure we've written this up publicly yet) - in your case you would use User Secrets to specify that you want to use AzureCliCredential instead of the DefaultAzureCredential. However, that vision hasn't been realized yet. We are trying to design something with the Azure SDK team that works for everyone. [[FEATURE REQ] Allow Microsoft.Extensions.Azure to select more TokenCredentials (Azure/azure-sdk-for-net#47927)](https://github.com/Azure/azure-sdk-for-net/issues/47927) is related to this.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
eerhardt: https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.singleordefault?view=net-9.0

SingleOrDefault throws InvalidOperationException when the input sequence contains more than one element.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
eerhardt: Looks like this was explicitly implemented in https://github.com/Azure/azure-sdk-for-net/pull/41819. FYI - @jsquire in case this wasn't intended to be supported as part of that change. .NET Aspire is going to depend on the Query of the Uri being ignored (or at least not throwing an exception).

playground/AspireEventHub/EventHubs.AppHost/Program.cs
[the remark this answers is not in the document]
eerhardt: Yes, let's remove them. I don't see it being maintainable to add all the different variations someone could want to change.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -116,6 +117,32 @@ public static IResourceBuilder<AzureEventHubsResource> WithHub(this IResourceBui
...
+    {
+        // Only one event hub can be the default entity
+        if (builder.Resource.Hubs.Any(h => h.IsDefaultEntity && h.Name != name))
+        {
+            throw new DistributedApplicationException("Only one EventHub can be configured as the default entity.");
+        }
+
+        // We need to ensure that the hub exists before we can set it as the default entity.
+        if (builder.Resource.Hubs.Any(h => h.Name  == name))
```
jsquire, line 136:
```suggestion
        if (builder.Resource.Hubs.Any(h => h.Name == name))
```

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
```diff
@@ -200,7 +227,13 @@ public static IResourceBuilder<AzureEventHubsResource> RunAsEmulator(this IResou
             // an event hub namespace without an event hub? :)
             if (builder.Resource.Hubs is { Count: > 0 } && builder.Resource.Hubs[0] is { } hub)
             {
-                var healthCheckConnectionString = $"{connectionString};EntityPath={hub.Name};";
+                // Endpoint=... format
+                var props = EventHubsConnectionStringProperties.Parse(connectionString);
```
jsquire, line 231: This is a pretty heavy way to check the string.  Since `EntityName` is a well-known token in this scenario, it may make sense to consider an `IndexOf` check instead.

sebastienros: It's also possible that this code has been written because I suggested to use it 🤷‍♂️. "Heavy" doesn't matter when we try to run the Event Hubs emulator or provision it ...

oising: > It's also possible that this code has been written because I suggested to use it 🤷‍♂️. "Heavy" doesn't matter when we try to run the Event Hubs emulator or provision it ...

What he said :)

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: Ah, sorry -- I was misremembering the reasons why I didn't use this pattern. Yeah, I _expected_ it to return `default` if not single, but you're right -- it throws. You're right that it did need another pass to consolidate the logic. I just pushed another commit to address this.

playground/AspireEventHub/EventHubsApi/Program.cs
oising: Done.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
oising: Done.

playground/AspireEventHub/EventHubs.AppHost/Program.cs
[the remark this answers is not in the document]
oising: Done.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs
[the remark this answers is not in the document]
oising: I validated that while a resource group could contain international/accented characters, event hub names cannot:

image

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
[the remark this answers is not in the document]
oising: @sebastienros -- you're right -- it does make sense to allow someone to call it multiple times with different hubs, with last one winning. I've had a head cold all week and for some reason I didn't quite get it, lol. I'll fix that now.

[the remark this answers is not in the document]
oising: I'm renaming `WithDefaultEntity` to `WithEntityPath` since this now opens up:

```csharp
var ns = builder.AddAzureEventHubs("eventhubns")
    .RunAsEmulator()
    .WithHub("hub")
    .WithHub("hub2");
    
builder.AddProject<Projects.EventHubsConsumerA>("consumera")
    .WithReference(ns.WithEntityPath("hub"));
    
builder.AddProject<Projects.EventHubsConsumerB>("consumerb")
    .WithReference(ns.WithEntityPath("hub2"));
 
 // ...
 ```
 
 @davidfowl @eerhardt ^ ?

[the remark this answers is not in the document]
Comment (oising) [reply src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs:127]: Although ... this approach could be troublesome due to it mutating the parent. Example:

```csharp
ar ns = builder.AddAzureEventHubs("eventhubns")
    .RunAsEmulator()
    .WithHub("hub")
    .WithHub("hub2");

var hub = ns.WithEntityPath("hub");
var hub2 = ns.WithEntityPath("hub2");

builder.AddProject<Projects.EventHubsConsumerA>("consumera")
    .WithReference(hub);
    
builder.AddProject<Projects.EventHubsConsumerB>("consumerb")
    .WithReference(hub2);
```

So here, both get pointed at hub2. Oops. What are your thoughts? I like that With vs Add has well defined semantics. Is it clear enough though to dissuade people doing things like this? Maybe this isn't the only API that could fall foul to this.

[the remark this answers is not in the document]
eerhardt: I think this is getting at the comment @davidfowl made here: https://github.com/dotnet/aspire/pull/7105#issuecomment-2608988310. The problem we are running into is that EventHub "Hub" instances aren't a `Resource` in Aspire. Which means they aren't referencable/addressable. So that's the struggle with the above code. If instead it looked more like a database in typical DBMS systems (like SqlServer or Postgres) the above code would be:

```C#
IResourceBuilder<AzureEventHubsResource> ns = builder.AddAzureEventHubs("eventhubns")
    .RunAsEmulator();

IResourceBuilder<AzureEventHubsHubResource> hub = ns.AddHub("hub");
IResourceBuilder<AzureEventHubsHubResource> hub2 = ns.AddHub("hub2");

builder.AddProject<Projects.EventHubsConsumerA>("consumera")
    .WithReference(hub);
    
builder.AddProject<Projects.EventHubsConsumerB>("consumerb")
    .WithReference(hub2);
```

ConsumerA would get a connection string for `eventhubns?EntityPath=hub` and ConsumerB would get a connection string for `eventhubns?EntityPath=hub2`.

This feels more natural and consistent with the rest of Aspire.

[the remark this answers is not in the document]
oising: Absolutely, and I completely agree. I had another PR open where I was creating child resources that could be deep-linked in this style, but it stalled because of refactoring work by the team in main. The thing is that we have obsolete APIs for AddEventHub that mutate the parent, so a stop-gap solution would be nice until the obsoleted API can be removed and then an AddHub added that does what you show. I guess the question is, do we go with this intermediary solution or do we throw it away and target 10.0 with AddHub. That's for you guys to decide.

davidfowl:
We had an offline discussion about "what is a resource" and decided to make these resources. @eerhardt is going to follow up with a spike.

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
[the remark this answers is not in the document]
eerhardt: @oising - check out 

* [Convert AzureEventHub and ConsumerGroup to Resources (dotnet/aspire#7438)](https://github.com/dotnet/aspire/pull/7438)
* [Allow EventHub connection strings to work with EntityPath and ConsumerGroup keys (dotnet/aspire#7453)](https://github.com/dotnet/aspire/pull/7453)

We are turning EventHub and ConsumerGroup into Resources, so they can be deep linked. This way the application won't have to specify the HubName and ConsumerGroup, but instead the values will flow from the Host, just like the connection information.

Let me know what you think about this approach.

davidfowl:
See https://github.com/dotnet/aspire/pull/7453

src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs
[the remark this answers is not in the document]
oising: I see you're stuffing a new pseudo property into the SAS style connection string that the emulator enforces (i.e. Endpoint=) but I how are you managing it for FQNS + tokencredential?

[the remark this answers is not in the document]
eerhardt: > but I how are you managing it for FQNS + tokencredential?

I'm turning that scenario into a connection string of the form:

```
Endpoint=https://eventhubns-cetg3lr.servicebus.windows.net:443/;EntityPath=myhub;ConsumerGroup=mygroup
```

Which gets parsed into an `AzureMessagingEventHubsSettings` object with:

- FullyQualifiedNamespace = https://eventhubns-cetg3lr.servicebus.windows.net:443/
- ConnectionString = null
- EventHubName = myhub
- ConsumerGroup = mygroup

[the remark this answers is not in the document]
oising: @eerhardt And this will still allow someone to override this in the client part via settings - either configuration or the settings callback?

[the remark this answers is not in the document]
oising: simply put, if you use `https://myeventhub.servicebus.windows.net:443/?EntityPath=hub&ConsumerGroup=foo` as a format, you won't break pure azure sdk, and you can still deconstruct into settings.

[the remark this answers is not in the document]
oising: Urgh -- I'm getting embrace, extend, extinguish vibes here but Microsoft own both sides of the contract, so maybe it's not so bad lol. On a completely different tangent -- if we're prepared to accept Aspire lock-in on both sides of the contract, why not make it a _choice_ and have resources implement an additional "enhanced" binding interface, e.g. IResourceWithConnectionProperties and leave IResourceWithConnectionString alone? The latter would be a downlevel binding for using the Azure SDK natively. The former could do whatever the hell it likes. /cc @davidfowl @eerhardt @ReubenBond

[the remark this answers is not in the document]
eerhardt: > And this will still allow someone to override this in the client part via settings - either configuration or the settings callback? 

You can override it in the settings callback. For configuration, the "client integrations" rules of order are:

* bind settings to configuration
* read connection string information from `ConnectionStrings:<name>`
* settings callback

So if the info is in both config and ConnectionsStrings, ConnectionStrings wins.

> The thing I don't particularly like about this approach is that it renders the connection string incompatible for the pure Azure SDK client

It was already incompatible with the pure Azure SDK client. You can't just pass a URL string (with no shared keys) into the "connectionString" parameter of the client ctor. You already had to check if it wasn't a connection string and pass it into the "fullyQualifiedNamespace" parameter (and fill out "credential") and not the "connectionString" parameter.

> This will be a breaking change for 9.0 clients too.

It isn't technically a breaking change because in 9.0 you could never `.WithReference` on the child EventHub. You could only `.WithReference` on the top-level EH namespace.

> if you use https://myeventhub.servicebus.windows.net:443/?EntityPath=hub&ConsumerGroup=foo as a format, you won't break pure azure sdk, and you can still deconstruct into settings.

This doesn't work for the local emulator. You'd still need to parse out ConsumerGroup of the local emulator connection string. The thinking is to "go all in" on the connection string format instead of one-off formats for each azure service. We already build up our own connection strings for things like Azure OpenAI and Search, as well as Milvus, Qdrant, Elastic. A similar change was made for CosmosDB's Database and Containers. So if we need to encode multiple pieces of information into a single string, we prefer to put it in a connection string format.

> have resources implement an additional "enhanced" binding interface

I agree this isn't perfect. I've been pushing for [Add support for Connection Properties (dotnet/aspire#1765)](https://github.com/dotnet/aspire/issues/1765), but we haven't prioritized it. So single strings is what we have today.

My hope is that in the future the Azure SDK supports IConfiguration natively and we can just bind to a configuration section where the Azure SDK loads whatever information is necessary to make the connection.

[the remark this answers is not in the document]
oising: > This doesn't work for the local emulator. You'd still need to parse out ConsumerGroup of the local emulator connection string. 

I'm very aware of this - this entire PR that is being abandoned implements all this functionality for both SAS connectionstrings and FQNS/tokencredential.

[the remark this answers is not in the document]
davidfowl: The developer experience is going to be really nice when we're done here. We're also discussing exposing an api to parse and reason about connection info on the application side of things.

davidfowl:
We can close this out now right?

eerhardt:
Closing in favor of the changes in #7438, #7453, and #7468.

Thanks for pushing this scenario forward @oising! Even though we didn't take this PR, your work inspired the final solution.

PR #7108, opened 2025-01-14
Don't end resource notifications until service is disposed

DamianEdwards:
## Description

Changes the `ResourceNotificationService` to keep notifications flowing until the service is disposed (i.e. when the app host is disposed), rather than stopping them when the application is stopped.

Contributes to #4878

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -14,13 +14,13 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// A service that allows publishing and subscribing to changes in the state of a resource.
 /// </summary>
-public class ResourceNotificationService
+public class ResourceNotificationService : IDisposable
 {
     // Resource state is keyed by the resource and the unique name of the resource. This could be the name of the resource, or a replica ID.
     private readonly ConcurrentDictionary<(IResource, string), ResourceNotificationState> _resourceNotificationStates = new();
     private readonly ILogger<ResourceNotificationService> _logger;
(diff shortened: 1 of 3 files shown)


davidfowl (approved):

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -58,8 +57,10 @@ public ResourceNotificationService(ILogger<ResourceNotificationService> logger,
     {
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         _serviceProvider = serviceProvider;
-        _applicationStopping = hostApplicationLifetime?.ApplicationStopping ?? throw new ArgumentNullException(nameof(hostApplicationLifetime));
         _resourceLoggerService = resourceLoggerService ?? throw new ArgumentNullException(nameof(resourceLoggerService));
+
+        // The IHostApplicationLifetime parameter is not used anymore, but we keep it for backwards compatibility.
+        // Notfication updates will be cancelled when the service is disposed.
```
eerhardt, line 63:
```suggestion
        // Notification updates will be cancelled when the service is disposed.
```

eerhardt (approved):
Thanks!

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
DamianEdwards: doh

PR #7111, opened 2025-01-14
Handling vscode insiders devcontainers

### LINKED ISSUE #7110: Devcontainer support doesn't work for vscode insiders
The devcontainer work to ensure the settings file exists expects a specific path for the `settings.json`, but that path is for VS Code, not VS Code Insiders: https://github.com/dotnet/aspire/blob/main/src/Aspire.Hosting/Devcontainers/DevcontainerSettingsWriter.cs#L14-L15

For VS Code Insiders it should look at this path `/home/vscode/.vscode-server-insiders/data/Machine/settings.json`

Versions of things:

- Aspire 9.1.0-preview.1.25064.6
- VS Code
```
Version: 1.97.0-insider (system setup)
Commit: 69d97b0773575a75736850370918e1171cfde5c6
Date: 2025-01-13T09:58:56.423Z
Electron: 32.2.7
ElectronBuildId: 10660205
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26120
```
- Remote Containers extension 0.395.0

## Workaround

A current workaround is to symlink the VS Code insiders path to the VS Code one and it seems to work

aaronpowell:
Fixes #7110

### DIFF
--- a/src/Aspire.Hosting/Devcontainers/DevcontainerSettingsWriter.cs
+++ b/src/Aspire.Hosting/Devcontainers/DevcontainerSettingsWriter.cs
@@ -12,7 +12,9 @@ namespace Aspire.Hosting.Devcontainers;
 internal class DevcontainerSettingsWriter(ILogger<DevcontainerSettingsWriter> logger, IOptions<CodespacesOptions> codespaceOptions, IOptions<DevcontainersOptions> devcontainerOptions)
 {
     private const string CodespaceSettingsPath = "/home/vscode/.vscode-remote/data/Machine/settings.json";
-    private const string LocalDevcontainerSettingsPath = "/home/vscode/.vscode-server/data/Machine/settings.json";
+    private const string VSCodeServerPath = "/home/vscode/.vscode-server";
+    private const string VSCodeInsidersServerPath = "/home/vscode/.vscode-server-insiders";
+    private const string LocalDevcontainerSettingsPath = "data/Machine/settings.json";
     private const string PortAttributesFieldName = "remote.portsAttributes";
     private const int WriteLockTimeoutMs = 2000;
     private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1);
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting/Devcontainers/DevcontainerSettingsWriter.cs
```diff
@@ -12,7 +12,9 @@ namespace Aspire.Hosting.Devcontainers;
 internal class DevcontainerSettingsWriter(ILogger<DevcontainerSettingsWriter> logger, IOptions<CodespacesOptions> codespaceOptions, IOptions<DevcontainersOptions> devcontainerOptions)
 {
     private const string CodespaceSettingsPath = "/home/vscode/.vscode-remote/data/Machine/settings.json";
-    private const string LocalDevcontainerSettingsPath = "/home/vscode/.vscode-server/data/Machine/settings.json";
+    private const string VSCodeServerPath = "/home/vscode/.vscode-server";
```
davidfowl, line 15: Don't these install side by side? We want to apply these modifications to both right?

aaronpowell: If you switch between VS Code and VS Code Insiders with the same devcontainer, you will end up with both a `.vscode-server` and `.vscode-insiders-server` in the `/home/vscode` directory, which means that there is the possibility of false positives and that you apply changes to the wrong settings file.

Unfortunately, I can't find any way in which you can detect which edition of VS Code you're running in the devcontainer. Here's the Community Toolkit, opened in a devcontainer in both VS Code and VS Code Insiders

image

No environment variables seem useful enough in determining which one to apply beyond "guessing" using the available paths, short of checking the `PATH` for `vscode-server` or `vscode-server-insiders`.

davidfowl: Why can’t we do both?

aaronpowell: It'd require a bit more invasive a refactor, since the method `GetSettingsPath` (that used the const) only returned a single path, but if that'd be preferred, I can have it go down that path.

davidfowl: Yes I would do that.

aaronpowell: Done

davidfowl (approved):

PR #7121, opened 2025-01-15
Refactor WithRealmImport to accept file or directory

### LINKED ISSUE #5086: KeycloakResourceBuilderExtensions.WithRealmImport is unnecessarily constraining the bind to a directory
## Background and Motivation

`KeycloakResourceBuilderExtensions.WithRealmImport` is unnecessarily constraining the bind to a single directory, constraining the user choices.

## Proposed API

Consider something like this, instead:

```csharp
    public static IResourceBuilder<KeycloakResource> WithRealmImport(
        this IResourceBuilder<KeycloakResource> builder,
        string import,
        bool isReadOnly = false)
    {
        ArgumentNullException.ThrowIfNull(importDirectory);

        if (Directory.Exists(import))
        {
            builder.WithBindMount(import, "/opt/keycloak/data/import", isReadOnly);
        }
        else
        {
            var fileName = Path.GetFileName(import);

            builder.WithBindMount(import, $"/opt/keycloak/data/import/{fileName}", isReadOnly);
        }

        return builder;
    }
```

Do not test if the source file or directory exists, because the user might create them later.

## Usage Examples

```csharp
var keycloak = builder.AddKeycloak(...);
// ...
keycloak.WithRealmImport(<path to file>);
// ...
keycloak.WithRealmImport(<path to another file>);
```

paulomorgado:
## Description

Updated the `WithRealmImport` method to change the parameter name from `importDirectory` to `import`, allowing it to accept either a directory or a single import file. Adjusted the method logic to handle both cases and updated the XML documentation for clarity.

Fixes #5086

/cc @davidfowl @DamianEdwards 

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


Aspire.lutconfig
```diff
@@ -0,0 +1,6 @@
+<LUTConfig Version="1.0">
```
DamianEdwards, line 1: Why was this added?

tests/Aspire.Hosting.Keycloak.Tests/KeycloakPublicApiTests.cs
```diff
@@ -107,26 +107,26 @@ public void WithRealmImportShouldThrowWhenBuilderIsNull()
...
+        var action = () => keycloak.WithRealmImport(import);
 
         var exception = Assert.Throws<ArgumentNullException>(action);
-        Assert.Equal(nameof(importDirectory), exception.ParamName);
+        Assert.Equal(nameof(import), exception.ParamName);
     }
 
     [Fact]
-    public void WithRealmImportShouldThrowWhenImportDirectoryDoesNotExist()
+    public void WithRealmImportShouldThrowWhenImportDoesNotExist()
```
DamianEdwards, line 123: Seems we have no tests that actually assert the happy path behavior! Could add a test/tests to verify that calling it with a directory or file that **does** exist adds the expected annotation?

DamianEdwards (changes requested):

Aspire.lutconfig
paulomorgado: Love LUT, but still struggle with this new version.

tests/Aspire.Hosting.Keycloak.Tests/KeycloakPublicApiTests.cs
paulomorgado: Added!

DamianEdwards (approved):

src/Aspire.Hosting.Keycloak/KeycloakResourceBuilderExtensions.cs
```diff
@@ -146,20 +146,26 @@ public static IResourceBuilder<KeycloakResource> WithDataBindMount(this IResourc
     /// </example>
     public static IResourceBuilder<KeycloakResource> WithRealmImport(
         this IResourceBuilder<KeycloakResource> builder,
-        string importDirectory,
+        string import,
```
davidfowl, line 149: This is a breaking change if you use named parameters but this package is preview so care level is low 😄

paulomorgado: > This is a breaking change if you use named parameters but this package is preview so care level is low 😄

Yes and yes. But I couldn't keep `importDirectory` when the semantics has changed.

PR #7125, opened 2025-01-15
Remove major version updates to runtime, aspnetcore, and efcore packages from dependabot

eerhardt:
These versions are all split in our dependencies based on TFM, and dependabot doesn't handle that situation well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,13 +23,19 @@ updates:
       # Pinned versions that should not be updated
       - dependency-name: "Microsoft.CodeAnalysis.CSharp"
       # don't update across major versions of NetPlatform dependencies. We need to do that manually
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -31,9 +31,6 @@ updates:
       AspNetCoreHealthChecks:
         patterns:
           - "AspNetCore.HealthChecks.*"
-      EntityFrameworkCore:
```
sebastienros, line 34: Is it sufficient to disable these? I would assume that groups are just to create batch update PRs, not to enlist packages. I would have added these in the `ignore` section too.

eerhardt: I changed it to be `update-types`, so we only take minor and patch updates. Not major.

joperezr (approved):
LGTM thanks Eric!

PR #7136, opened 2025-01-17
Fix Azure ServiceBus persistent container support

### LINKED ISSUE #7071: The service bus emulator is never persistent when marked as such
When marking the service bus emulator as persistent it still restarts. This is unfortunate because it takes a little bit to start up.

```C#
var serviceBus = builder.AddAzureServiceBus("bus")
                        .RunAsEmulator(c => c.WithLifetime(ContainerLifetime.Persistent))
                        .WithQueue("q");
```

It seems like the problem is:
- https://github.com/dotnet/aspire/blob/be904514949ba5d94e2e4b6d7b8890ab63b88316/src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs#L249

We generate a new password every time, causing the container hash to change.

sebastienros:
## Description

Fixes #7071

Issues identified:
- the lifetime is not forwarded to the dependent container (sqledge)
- the password was new on every start, invalidating reusability
- the volume was using a new folder on every start, invalidating reusability

Now the password and the folder are stored in user secrets and reused.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -233,9 +235,29 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+            }
+        }
+
         // Create a default file mount. This could be replaced by a user-provided file mount.
+
         var configHostFile = Path.Combine(Directory.CreateTempSubdirectory("AspireServiceBusEmulator").FullName, "Config.json");
 
+        if (lifetime == ContainerLifetime.Persistent && builder.ApplicationBuilder.ExecutionContext.IsRunMode && builder.ApplicationBuilder.AppHostAssembly is not null)
+        {
+            configHostFile = GetOrSetUserSecret(builder.ApplicationBuilder.AppHostAssembly, "Parameters:ServiceBusEmulatorConfigFile", configHostFile);
```
davidfowl, line 258: Can you add a comment as to what this does? It is using the same volume path across runs?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -233,9 +235,29 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+            {
+                lifetime = lifetimeAnnotation.Lifetime;
+            }
+        }
+
         // Create a default file mount. This could be replaced by a user-provided file mount.
+
         var configHostFile = Path.Combine(Directory.CreateTempSubdirectory("AspireServiceBusEmulator").FullName, "Config.json");
 
+        if (lifetime == ContainerLifetime.Persistent && builder.ApplicationBuilder.ExecutionContext.IsRunMode && builder.ApplicationBuilder.AppHostAssembly is not null)
```
davidfowl, line 256: I wonder if we can centralize this logic.

davidfowl:
AddAzureServiceBusWithEmulatorGetsExpectedPort - actual failing test

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -264,20 +253,62 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+            context.EnvironmentVariables.Add("MSSQL_SA_PASSWORD", passwordParameter);
         }));
 
         ServiceBusClient? serviceBusClient = null;
         string? queueOrTopicName = null;
 
+        var lifetime = ContainerLifetime.Session;
+
+        // Create a default file mount. This could be replaced by a user-provided file mount.
+        var configHostFile = Path.Combine(Directory.CreateTempSubdirectory("AspireServiceBusEmulator").FullName, "Config.json");
```
davidfowl, line 276: Do we always want to create the directory up front? In the case where the lifetime of is persistent, we may not want this right?

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -51,6 +51,36 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// <summary>
+    /// Creates a new <see cref="ParameterResource"/> that has a generated value using the <paramref name="value"/>.
+    /// </summary>
+    /// <remarks>
+    /// The value will be saved to the app host project's user secrets store when <see cref="DistributedApplicationExecutionContext.IsRunMode"/> is <c>true</c>
+    /// and the lifetime of the resource is <see cref="ContainerLifetime.Persistent"/>.
+    /// </remarks>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">Name of the parameter</param>
+    /// <param name="value"></param>
```
eerhardt, line 63: The `value` parameter is a bit odd since it is only used on the first run. After the value is persisted once, calling `AddPersistentParameter` again with a different `value` is ignored, since it reads it from user secrets.

Maybe calling this `initialValue` or similar may help.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
```diff
@@ -51,6 +51,36 @@ public static IResourceBuilder<ParameterResource> AddParameter(this IDistributed
...
+    /// <returns>The created <see cref="ParameterResource"/>.</returns>
+    public static ParameterResource AddPersistentParameter(this IDistributedApplicationBuilder builder, string name, string value)
```
eerhardt, line 65: What's the difference between this API and `AddParameter(..., persist: true)`?

src/Aspire.Hosting.Azure.ServiceBus/Aspire.Hosting.Azure.ServiceBus.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)src\Aspire.Hosting\Utils\PasswordGenerator.cs" Link="Utils\PasswordGenerator.cs" />
+    <Compile Include="$(SharedDir)SecretsStore.cs" Link="Utils\SecretsStore.cs" />
```
eerhardt, line 17: Is this still needed?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -233,20 +232,10 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
-                ContainerMountType.BindMount,
-                isReadOnly: true);
-
-        builder.WithAnnotation(defaultConfigFileMount);
-
         // Add emulator container
 
-        var password = PasswordGenerator.Generate(16, true, true, true, true, 0, 0, 0, 0);
+        // The password must be at least 8 characters long and contain characters from three of the following four sets: Uppercase letters, Lowercase letters, Base 10 digits, and Symbols
+        var passwordParameter = ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder.ApplicationBuilder, $"{builder.Resource.Name}-sqledge-pwd", minLower: 1, minUpper: 1, minNumeric: 1);
```
eerhardt, line 238:
```suggestion
        var passwordParameter = ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(builder.ApplicationBuilder, $"{builder.Resource.Name}-sql-pwd", minLower: 1, minUpper: 1, minNumeric: 1);
```

Do we need to explicitly call this "sqledge"? What if we changed it in the future?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -264,20 +253,62 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+
+        sqlEdgeResource = sqlEdgeResource.WithLifetime(lifetime);
+
+        var defaultConfigFileMount = new ContainerMountAnnotation(
+                configHostFile,
+                AzureServiceBusEmulatorResource.EmulatorConfigJsonPath,
+                ContainerMountType.BindMount,
+                isReadOnly: true);
+
+        var hasCustomConfigJson = builder.Resource.Annotations.OfType<ContainerMountAnnotation>().Any(v => v.Target == AzureServiceBusEmulatorResource.EmulatorConfigJsonPath);
```
eerhardt, line 305: Why are we checking this now? We weren't before.

sebastienros: `WithConfigurationFile` is removing an existing one. Before this PR the call back would be executed before `WithConfigurationFile` so it would keep only one. Now the callback is called after `WithConfigurationFile` so it won't have the same effect.

There is a test ensuring a single mount is defined.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
sebastienros: In the end it does the same thing, but you need to use a `ParameterDefault` in the existing one and the one we need is private. So it's either making `ConstantParameterDefault` public or create a new helper method that uses it (which is what I implemented).

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
sebastienros: But the resource is named `$"{builder.Resource.Name}-sqledge"`, that's why it's named `$"{builder.Resource.Name}-sqledge-pwd"` with the extra `-pwd`. If we were to change the type of resource, and the name, we could also use a different one for the password.

Or we can name everything without `sqledge` at all.

eerhardt: I think let's drop the `edge` in all for sure. I think `sql` is fine, but if you want to call it something more generic like `-data` I wouldn't object.

src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs
eerhardt: Can this one be built on top of the other one to show that it is the same thing only with different inputs?

eerhardt: Given your current build errors, it kind of shows why this API isn't working very well. You want to pass a ParameterDefault that will generate a temp file when the parameter value isn't already persisted. Basically a callback func that produces the default value when the value isn't already there.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -264,20 +254,57 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
 
         ServiceBusClient? serviceBusClient = null;
         string? queueOrTopicName = null;
 
+        var lifetime = ContainerLifetime.Session;
+
+        var aspireStore = AspireStore.Create(builder.ApplicationBuilder);
+
+        // Deterministic file path for the configuration file
+        var configHostFile = aspireStore.GetOrCreateFile("Config.json");
```
eerhardt, line 279:
```suggestion
        var configHostFile = aspireStore.GetOrCreateFile($"{builder.Resource.Name}-Config.json");
```

src/Shared/AspireStore.cs
```diff
@@ -0,0 +1,176 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Buffers;
+using System.Security.Cryptography;
+using IdentityModel;
+using Microsoft.Extensions.SecretManager.Tools.Internal;
+
+namespace Aspire.Hosting.Utils;
+
+internal sealed class AspireStore : KeyValueStore
```
eerhardt, line 11: I think we will want to make this API public because any integration (including outside of dotnet/aspire) will want to write to this folder.

src/Shared/SecretsStore.cs
```diff
@@ -12,83 +11,92 @@ namespace Microsoft.Extensions.SecretManager.Tools.Internal;
 /// <summary>
 /// Adapted from dotnet user-secrets at https://github.com/dotnet/aspnetcore/blob/482730a4c773ee4b3ae9525186d10999c89b556d/src/Tools/dotnet-user-secrets/src/Internal/SecretsStore.cs
 /// </summary>
-internal sealed class SecretsStore
+internal abstract class KeyValueStore
```
eerhardt, line 14: I'm not sure we need this abstraction. SecretsStore is just about reading/writing the `secrets.json` file. The new Store should be about reading/writing files to the app directory (wherever it is that we decide).

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -228,6 +228,11 @@ public static IResourceBuilder<AzureServiceBusResource> AddSubscription(this IRe
     /// </example>
     public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IResourceBuilder<AzureServiceBusResource> builder, Action<IResourceBuilder<AzureServiceBusEmulatorResource>>? configureContainer = null)
     {
+        if (builder.Resource.IsEmulator)
+        {
+            throw new InvalidOperationException("The Azure Service Bus resource is already configured to run as an emulator.");
+        }
```
davidfowl, line 234: Shouldn't this override the behavior? Why would it throw?

davidfowl: I think we'll want to handle this differently but this is ok for now.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// Gets the base path of this store.
+    /// </summary>
+    public string BasePath => _basePath;
+
+    /// <summary>
+    /// Creates a new instance of <see cref="AspireStore"/> using the provided <paramref name="builder"/>.
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <returns>A new instance of <see cref="AspireStore"/>.</returns>
+    public static AspireStore Create(IDistributedApplicationBuilder builder)
```
davidfowl, line 48: This is a weird API. Can we make this internal and instead expose something on the builder?

tests/Aspire.Hosting.Azure.Tests/AzureServiceBusExtensionsTests.cs
```diff
@@ -188,6 +188,7 @@ public async Task VerifyWaitForOnServiceBusEmulatorBlocksDependentResources()
     {
         var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
         using var builder = TestDistributedApplicationBuilder.Create(output);
+        builder.Configuration["Aspire:Store:Path"] = Path.GetTempPath();
```
davidfowl, line 191: Set this once in the test builder?

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// <returns>The metadata value if found; otherwise, null.</returns>
+    private static string? GetMetadataValue(IEnumerable<AssemblyMetadataAttribute>? assemblyMetadata, string key) =>
+        assemblyMetadata?.FirstOrDefault(a => string.Equals(a.Key, key, StringComparison.OrdinalIgnoreCase))?.Value;
+
+    /// <summary>
+    /// Gets the application host specific prefix based on the builder's environment.
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <returns>The application host specific prefix.</returns>
+    private static string GetAppHostSpecificPrefix(IDistributedApplicationBuilder builder)
```
eerhardt, line 89: I'm not sure we need this logic. I would be simplier if we just use the folder that is specified directly.

It is only used when the `Aspire:Store:Path` config is used. And in that case we can just tell people to use a different folder for each app, if they are having conflicts.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            throw new FileNotFoundException("The source file '{0}' does not exist.", sourceFilename);
+        }
+
+        EnsureDirectory();
+
+        // Strip any folder information from the filename.
+        filename = Path.GetFileName(filename);
+
+        // Delete existing file versions with the same name.
```
eerhardt, line 119: I'm not sure we want to do this. What if these files are being used?

eerhardt: The point of putting this in "/obj" is that we can leave these orphaned files, and when the user deletes the whole project or cleans their repo, the files will go away.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// <summary>
+    /// Creates a new instance of <see cref="AspireStore"/> using the provided <paramref name="builder"/>.
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <returns>A new instance of <see cref="AspireStore"/>.</returns>
+    public static AspireStore Create(IDistributedApplicationBuilder builder)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        var assemblyMetadata = builder.AppHostAssembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
```
eerhardt, line 52: We don't need to do this if `AspireStorePathKeyName` is set. So we can put it inside an `if`.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        });
+    }
+
+    /// <summary>
+    /// Ensures that the directory for the store exists.
+    /// </summary>
+    private void EnsureDirectory()
+    {
+        if (!string.IsNullOrEmpty(_basePath) && !Directory.Exists(_basePath))
```
eerhardt, line 229: Do we need to check if exists? Why not just call CreateDirectory?

src/Aspire.Hosting/AspireStore.cs
sebastienros: The point was that there is still the fallback folder that is handled, and it's a cheap effort to clean older versions of a file we are about to update. If you think there is some risk associated with it I'll remove it.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+
+    /// <summary>
+    /// Gets a deterministic file path that is a copy of the <paramref name="sourceFilename"/>.
+    /// The resulting file name will depend on the content of the file.
+    /// </summary>
+    /// <param name="filename">A file name to base the result on.</param>
+    /// <param name="sourceFilename">An existing file.</param>
+    /// <returns>A deterministic file path with the same content as <paramref name="sourceFilename"/>.</returns>
+    /// <exception cref="FileNotFoundException">Thrown when the source file does not exist.</exception>
+    public string GetFileNameWithContent(string filename, string sourceFilename)
```
eerhardt, line 104: Should we call `sourceFilename` something like `sourceFilenameTemplate`? Since we will be appending the hash to it?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -287,131 +284,59 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+            try
+            {
                 // Apply ConfigJsonAnnotation modifications
-                var configJsonAnnotations = emulatorResource.Annotations.OfType<ConfigJsonAnnotation>();
+                var configJsonAnnotations = builder.Resource.Annotations.OfType<ConfigJsonAnnotation>();
 
                 foreach (var annotation in configJsonAnnotations)
                 {
-                    using var readStream = new FileStream(configFileMount.Source!, FileMode.Open, FileAccess.Read);
+                    using var readStream = new FileStream(tempConfigFile, FileMode.Open, FileAccess.Read);
```
eerhardt, line 335: Do we need to read and write the file for each configJsonAnnotation? Why not read it once, run each annotation, and then write it back once?

src/Aspire.Hosting/AspireStore.cs
eerhardt: I just don't think we should be messing with any other file than what we are operating on. We don't know who else is using these files. Someone could call this API twice with different contents and actually want 2 different files.

Callers and users who want things cleaned up can do it outside of this core API.

src/Aspire.Hosting/AspireStore.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        return finalFilePath;
+    }
+
+    /// <summary>
+    /// Creates an absolute file name for the provided <paramref name="filename"/> in the store.
+    /// </summary>
+    /// <param name="filename">The file name to use in the store.</param>
+    /// <returns>The absolute file name in the store.</returns>
+    public string GetFileName(string filename)
```
eerhardt, line 191: What's this API used for? Why can't callers just use the BasePath to do whatever they need?

src/Aspire.Hosting/AspireStoreExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+public static class AspireStoreExtensions
+{
+    internal const string AspireStorePathKeyName = "Aspire:Store:Path";
+
+    /// <summary>
+    /// Creates a new App Host store using the provided <paramref name="builder"/>.
+    /// </summary>
+    /// <param name="builder">The <see cref="IDistributedApplicationBuilder"/>.</param>
+    /// <returns>The <see cref="IAspireStore"/>.</returns>
+    public static IAspireStore CreateStore(this IDistributedApplicationBuilder builder)
```
eerhardt, line 20: I know we can iterate on this API, so we don't have to block on this, but I think this is going to clutter up normal dev's intellisense. Normal apphost devs don't need to use this API, so I don't think it should be top-level like "AddContainer" or "AddPostgres", etc.

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -287,152 +284,118 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
-        {
-            var serviceBusEmulatorResources = builder.ApplicationBuilder.Resources.OfType<AzureServiceBusResource>().Where(x => x is { } serviceBusResource && serviceBusResource.IsEmulator);
+        var lifetime = ContainerLifetime.Session;
 
-            if (!serviceBusEmulatorResources.Any())
-            {
-                // No-op if there is no Azure Service Bus emulator resource.
-                return;
-            }
+        var aspireStore = builder.ApplicationBuilder.CreateStore();
```
eerhardt, line 295: Do we need to get this so early? Can we get it inside the BeforeStartEvent instead?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -287,152 +284,118 @@ public static IResourceBuilder<AzureServiceBusResource> RunAsEmulator(this IReso
...
+
+            if (connectionString == null)
+            {
+                throw new DistributedApplicationException($"ConnectionStringAvailableEvent was published for the '{builder.Resource.Name}' resource but the connection string was null.");
             }
+
+            // Retrieve a queue/topic name to configure the health check
+
+            var noRetryOptions = new ServiceBusClientOptions { RetryOptions = new ServiceBusRetryOptions { MaxRetries = 0 } };
+            serviceBusClient = new ServiceBusClient(connectionString, noRetryOptions);
```
eerhardt, line 390: (nit) since we split into 2 different event handlers, where these variables are declared are really far away from where they are used (here). Can we move them right before this event handler for better locality?

src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs
```diff
@@ -297,7 +294,12 @@ private static DistributedApplicationBuilder CreateInnerBuilder(
                 builder.Services.ConfigureHttpClientDefaults(http => http.AddStandardResilienceHandler());
             }
 
-            return builder;
+            if (!builder.Configuration.GetValue("Aspire:Store:Path", false))
+            {
+                builder.Configuration["Aspire:Store:Path"] = _tempAspireStorePath = Path.GetTempPath();
```
eerhardt, line 299: I think we should just do this in our tests, not for customer tests. Customer tests should use the AppHost's `obj` directory, just like a normal App IMO.

eerhardt (approved):

src/Aspire.Hosting/AspireStoreExtensions.cs
sebastienros: Ok to move to `Azure.Hosting.ApplicationModel` instead?

PR #7163, opened 2025-01-20
Allow the ResourceReadyEvent to block waiters

### LINKED ISSUE #7164: Add ability to run tasks as part of a resource lifecycle that block other resources from executing when they use WaitFor
It's common practice to run logic after a resource is ready but before it is marked healthy (potentially unblocking other resources). A common scenario is seeing resources (databases, caches etc), today it requires complex orchestration logic to run before and keep all waiting resources blocked until that task is done.

This should be the default behavior of the ResourceReadyEvent.

davidfowl:
## Description

When trying to write code that runs after a resource becomes healthy but before waiters are unblocked it requires complex coordination and a deep understanding of the resource lifecycle. This defaults to blocking waiters until the event is ready, allows users to build pretty nice experiences without complex coordination logic across resources.

Fixes #7164

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -62,6 +62,11 @@ public sealed record CustomResourceSnapshot
     public int? ExitCode { get; init; }
 
     /// <summary>
+    /// A flag indicating whether the resource ready event has completed.
+    /// </summary>
+    internal bool ResourceCompletedEventFired { get; init; }
+
+    /// <summary>
     /// Gets the health status of the resource.
     /// </summary>
     /// <remarks>
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -64,7 +64,7 @@ public sealed record CustomResourceSnapshot
     /// <summary>
     /// A flag indicating whether the resource ready event has completed.
     /// </summary>
-    internal bool ResourceCompletedEventFired { get; init; }
+    internal Task? ResourceReadyEventTask { get; init; }
 
     /// <summary>
     /// Gets the health status of the resource.
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -62,9 +62,9 @@ public sealed record CustomResourceSnapshot
     public int? ExitCode { get; init; }
 
     /// <summary>
-    /// A flag indicating whether the resource ready event has completed.
+    /// A snapshot of the event that indicates the resource is ready.
     /// </summary>
-    internal Task? ResourceReadyEventTask { get; init; }
+    internal EventSnapshot? ResourceReadyEvent { get; init; }
 
     /// <summary>
     /// Gets the health status of the resource.
@@ -134,6 +134,12 @@ public sealed record CustomResourceSnapshot
                 ?? Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus.Unhealthy;
     }
 }
+
+/// <summary>
+/// A snapshot of an event.
+/// </summary>
+/// <param name="EventTask">The task the represents the result of executing the event.</param>
+internal record EventSnapshot(Task EventTask);
 
 /// <summary>
 /// A snapshot of the resource state
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -163,7 +163,10 @@ public class ResourceNotificationService : IDisposable
 
         // Now wait for the resource ready event to be executed.
         resourceLogger.LogInformation("Waiting for resource ready to execute for '{Name}'.", dependency.Name);
-        await WaitForResourceAsync(dependency.Name, re => re.Snapshot.ResourceCompletedEventFired, cancellationToken: cancellationToken).ConfigureAwait(false);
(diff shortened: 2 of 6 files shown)


src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -61,6 +61,11 @@ public ResourceStateSnapshot? State
     /// </summary>
     public int? ExitCode { get; init; }
 
+    /// <summary>
+    /// A flag indicating whether the resource ready event has completed.
+    /// </summary>
+    internal bool ResourceCompletedEventFired { get; init; }
```
JamesNK, line 67: Why `ResourceCompletedEventFired`? Should it be `ResourceReadyEventFired`?

How would this scale to starting/stopping/restarting a resource multiple times? Should something reset it back to false?

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -52,11 +52,20 @@ private async Task MonitorResourceHealthAsync(ResourceEvent initialEvent, Cancel
             //       dynamically add health checks at runtime. If this changes then we
             //       would need to revisit this and scan for transitive health checks
             //       on a periodic basis (you wouldn't want to do it on every pass.
-            var resourceReadyEvent = new ResourceReadyEvent(resource, services);
-            await eventing.PublishAsync(
-                resourceReadyEvent,
-                EventDispatchBehavior.NonBlockingSequential,
-                cancellationToken).ConfigureAwait(false);
+            _ = Task.Run(async () =>
```
JamesNK, line 55: Is no await here anymore ok?

Should there be something to handle logging errors from the unawaited task?

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -76,11 +85,20 @@ await eventing.PublishAsync(
                 if (!resourceReadyEventFired && report.Status == HealthStatus.Healthy)
                 {
                     resourceReadyEventFired = true;
-                    var resourceReadyEvent = new ResourceReadyEvent(resource, services);
-                    await eventing.PublishAsync(
-                        resourceReadyEvent,
-                        EventDispatchBehavior.NonBlockingSequential,
-                        cancellationToken).ConfigureAwait(false);
+
+                    _ = Task.Run(async () =>
```
JamesNK, line 89: Is this the same as above? What do you think of adding a local method to reduce duplication?

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
davidfowl: I'll rename it. Today these events only fire once.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
davidfowl: The behavior before was non blocking, that means don't block on the event handlers firing.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
davidfowl: Will do

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -161,6 +161,13 @@ private async Task WaitUntilHealthyAsync(IResource resource, IResource dependenc
             await WaitForResourceHealthyAsync(dependency.Name, cancellationToken).ConfigureAwait(false);
         }
 
+        // Now wait for the resource ready event to be executed.
+        resourceLogger.LogInformation("Waiting for resource ready to execute for '{Name}'.", dependency.Name);
+        resourceEvent = await WaitForResourceAsync(dependency.Name, re => re.Snapshot.ResourceReadyEventTask is { } t && t.IsCompleted, cancellationToken: cancellationToken).ConfigureAwait(false);
```
JamesNK, line 166: Does this work properly? I believe `WaitForResourceAsync` re-evaluates the predicate when the snapshot is published. The works when testing other data on the snapshot because it is immutable and requires a new snapshot to be published to update. But the task here is mutable. That means if `ResourceReadyEventTask` completes and there isn't a snapshot update afterwards then the wait doesn't complete.

JamesNK: Would a fix be to update `FireResourceReadyEvent()` to await the eventing `PublishAsync` at the end of the method then do another notification publish?

davidfowl: That's a good catch.

davidfowl: I could instead wait until the task is non-null and then just use it.

davidfowl: Updated, didn't keep IsComplete on the event, it's not needed 😄. I just check for a non-null event.

sebastienros:
> runs after a resource becomes healthy but before waiters are unblocked

This is the reason I had to use a health check to seed the containers/dbs in the event hub host integration. With that I could move the logic there and use the standard health check.

JamesNK (approved):

davidfowl:
I'm led to believe that this PR is either causing problems with these tests or exposing flakiness somewhere.

PR #7211, opened 2025-01-22
Move resource creation in ResourceReadyEvent event

sebastienros:
## Description

I am testing the new `ResourceReadyEvent`. However there is a limitation in this case, I would still prefer the previous solution. Resource creation (db, containers) needs to happen when the emulator is "ready". But dependents should not use cosmosdb until the resources are created.

The previous solution was relying on a custom healthcheck to create and assess the db/containers are created and/or ready. This new solution is hitting a deadlock, that is the `ResourceReadyEvent` is triggered once the resource si healthy, but if the healthcheck ensure the db/containers are available then it can't work since `ResourceReadyEvent` is doing it. So right now no db/containers are passed in the healthcheck, so it will only ensure the service is available before `ResourceReadyEvent` can create the db/containers.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No


src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -12,9 +12,9 @@
 using Azure.Provisioning.CosmosDB;
 using Azure.Provisioning.Expressions;
 using Azure.Provisioning.KeyVault;
+using HealthChecks.CosmosDb;
```
davidfowl, line 15:
```suggestion
```

eerhardt:
This looks more like what I expect. But maybe thinking about it more .... not considering the resource to be "healthy" until the database is created might be the right thing to do....

davidfowl:
We can do that as a follow up to this https://github.com/dotnet/aspire/pull/7163

davidfowl (approved):

PR #7221, opened 2025-01-23
Remove app bar item horizontal padding

### LINKED ISSUE #6935: [WebToolsE2E][Aspire] The displayed string for Structured is not fully displayed for some localization.
**INSTALL STEPS**
1. Clean machine: Win11 x64 23h2
2. Install VS 17.12 P4 VAL (with Aspire 9.0.0)
3. Install latest Aspire 9.1.0-preview.1.24611.1

**REPRO STEPS**
1. Create .NET Aspire Starter App > .NET 9.0 (Standard Term Support) > Create
2. Build and F5
3. Change the display language in Settings, check [Spanish, Polish, Turkey and Russian]
4. Check the displayed string for Structured logs

**NOTE**
1. This issue is reproduced with the following language: Spanish, Polish, Turkey and Russian.
2. Resizing the page doesn't fully displayed the localized Structured string.

**ACTUAL**
The localized Structured string is not fully displayed.
Image

**EXPECTED**
The localized Structured string should be fully displayed.

adamint:
## Description

Nav bar items were being cut off due to a default app bar item width. Since there's a wide range of text lengths for these strings based on language, I un-set the width to allow the app bar width to flex to fit content.

**Before**
english:
image

spanish:
image

**After**
english (slightly narrower):
image

spanish (wider):
image

There is one extreme (Russian), I'm not sure what to do with this. Maybe setting a max width? @JamesNK 
image

Fixes #6935 

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

### DIFF
--- a/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
+++ b/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
@@ -233,3 +233,11 @@
         margin-left: 0;
     }
 }
+
+::deep .desktop-nav-menu > .fluent-appbar-item {
+    /* FluentAppBarItems have a default width. However, we have a wide range of widths
+       in content depending on the language (ie, from traces to seguimientos) so should allow
+       the app bar to fit its content. This means that the width of the app bar will be different for some
+       languages. */
+    width: unset;
+}
--- a/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
+++ b/src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
@@ -234,10 +234,9 @@
     }
 }
 
-::deep .desktop-nav-menu > .fluent-appbar-item {
+::deep .desktop-nav-menu > .fluent-appbar-item > .fluentui-counterbadge-container > a > .stack-vertical {
     /* FluentAppBarItems have a default width. However, we have a wide range of widths
-       in content depending on the language (ie, from traces to seguimientos) so should allow
-       the app bar to fit its content. This means that the width of the app bar will be different for some
-       languages. */
-    width: unset;
+       in content depending on the language (ie, from traces to seguimientos) so remove the default
+       horizontal padding to allow for more text  */
+    padding: calc(var(--design-unit) * 1px) 0;
 }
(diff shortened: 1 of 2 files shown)


JamesNK:
I don't like this. I can't think of other UIs that have this design where the app bar changes size. I looked at what teams does when it is set to spanish and the app bar size is the same but the text is displayed up to the edges.

Instead to fix this we can remove the left/right padding on the items in the app bar. That allows all content to be displayed:

image

JamesNK (changes requested):
See comment

adamint:
> I don't like this. I can't think of other UIs that have this design where the app bar changes size. I looked at what teams does when it is set to spanish and the app bar size is the same but the text is displayed up to the edges.
> 
> Instead to fix this we can remove the left/right padding on the items in the app bar. That allows all content to be displayed:
> 
> image
> 
> That won't work for every language - I see your Russian example won't fit - but I think we have to draw a line somewhere. This approach gives more room, but keeps a fixed size.
> 
> Maybe we should ask translation folks whether there are shorter words that can be used for Russian?

Sounds good with me. I do agree with 
> I think we have to draw a line somewhere.

especially considering it's only one item.

src/Aspire.Dashboard/Components/Layout/MainLayout.razor.css
```diff
@@ -233,3 +233,10 @@
         margin-left: 0;
     }
 }
+
+::deep .desktop-nav-menu > .fluent-appbar-item > .fluentui-counterbadge-container > a > .stack-vertical {
+    /* FluentAppBarItems have a default width. However, we have a wide range of widths
+       in content depending on the language (ie, from traces to seguimientos) so remove the default
+       horizontal padding to allow for more text  */
+    padding: calc(var(--design-unit) * 1px) 0;
```
JamesNK, line 241:
```suggestion
    /* FluentAppBarItems have a default width. However, we have a wide range of widths
       in content depending on the language (ie, from traces to seguimientos).
       Remove the default horizontal padding to allow for more text  */
    padding-left: 0;
    padding-right: 0;
```

JamesNK (approved):

PR #7229, opened 2025-01-23
Add 9.1 templates

### LINKED ISSUE #7160: [WebToolsE2E][Aspire] After installing Aspire 9.1 templates, 'dotnet new aspire' does not show 9.1 as a valid version for --aspire-version parameter
**INSTALL STEPS**
1. Install VS 17.13 Preview 2.1 
2. Install latest Aspire 9.1.0-preview.1.25066.1

**REPRO STEPS**

1. Open Command Prompt > Run 'Dotnet new aspire --help'.

**ACTUAL**
After installing Aspire 9.1.0-preview.1.25066.1, the version of Aspire still is 9.0.
Image

**EXPECTED**
The Aspire version can match the correct version.

DamianEdwards:
## Description

Adds 9.1 templates to the template package as the default. The 9.0 templates got updated to use fixed package versions. The 8.2 and 9.0 templates remain in the package as that's our current side-by-side strategy (all template versions in one package) until we get to 10.0 at which point we can decide which versions to drop from the package.

Fixes #7160

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes (existing)
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

Screenshot of Aspire version option in VS:

image

Screenshot of Aspire version option in C# Dev Kit:

image


eerhardt:
> The 8.2 and 9.0 templates remain in the package as that's our current side-by-side strategy (all template versions in one package) until we get to 10.0 at which point we can decide which versions to drop from the package.

This seems like it goes against our support policy, doesn't it? We don't support 9.0 anymore after we ship 9.1. So why should we continue to have 9.0 templates, if we no longer support it?

DamianEdwards:
> This seems like it goes against our support policy, doesn't it? We don't support 9.0 anymore after we ship 9.1. So why should we continue to have 9.0 templates, if we no longer support it?

Yes, but it's a huge inconvenience given all the trade-offs we made WRT our approach to template packaging for Aspire. If we remove 9.0 (or 8.2) from the package in the 9.1 update, you can no longer perform Aspire project creation or scaffolding operations in VS or VS Code for 9.0 projects. I can't think of anywhere else in our wider .NET experience where we instantly remove the ability to work with unsupported versions when the new version is released, because doing so would be hugely disruptive. The other advantage of this approach is it makes dogfooding much easier as you can install the daily 9.1 templates package (or even a locally built version) and still work with 9.0 projects/solutions with no issue.

joperezr:
> > The 8.2 and 9.0 templates remain in the package as that's our current side-by-side strategy (all template versions in one package) until we get to 10.0 at which point we can decide which versions to drop from the package.
> 
> This seems like it goes against our support policy, doesn't it? We don't support 9.0 anymore after we ship 9.1. So why should we continue to have 9.0 templates, if we no longer support it?

As @DamianEdwards has shared in the past, one thing is not supporting it, and a different one is blocking people from using it. One good precedent is the newer SDKs, they allow you to compile assemblies and applications that target out of support framework, while of course warning you that those runtimes are out of support. I think that the change here is following that same pattern, which seems good to me.

eerhardt:
What happens in .NET Aspire 10? We drop `8.2` templates? Do we drop all `9.x` templates as well? When do we ever drop old versions?

DamianEdwards:
@eerhardt undecided, that's something we figure out for 10. My guess is we drop 8.x and everything but the last 9.x

src/Aspire.ProjectTemplates/templates/aspire-empty/.template.config/template.json
```diff
@@ -21,7 +21,8 @@
     "type": "solution",
     "editorTreatAs": "solution",
     "aspire-8.2-tfms": "net8.0",
-    "aspire-9.0-tfms": "net8.0;net9.0"
+    "aspire-9.0-tfms": "net8.0;net9.0",
```
joperezr, line 24: Just out of curiosity (not super familiar with these files) how come these tags aren't in all template configs? For instance, aspire-apphost doesn't seem to have this change. Is that intentional?

DamianEdwards: they're only used by the code in VS that drives the solution explorer "Add Aspire orchestration" context menu.

joperezr (approved):
Left a copule of small comments, but this looks good otherwise. Thank you so much for getting this work done @DamianEdwards

src/Aspire.ProjectTemplates/templates/aspire-empty/9.0/AspireApplication.1.ServiceDefaults/AspireApplication.1.ServiceDefaults.csproj
```diff
@@ -10,8 +10,8 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
 
-    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="!!REPLACE_WITH_DOTNET_EXTENSIONS_VERSION!!" />
-    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="!!REPLACE_WITH_LATEST_VERSION!!" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="9.1.0" />
```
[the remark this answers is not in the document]
DamianEdwards: No, we don't update the 9.0 templates from this point on

src/Aspire.ProjectTemplates/templates/aspire-apphost/.template.config/template.json
```diff
@@ -53,41 +58,107 @@
...
+        }
+      ],
+      "defaultValue": "9.1"
+    },
+    "AspireVersionNet8": {
       "type": "parameter",
       "description": "The version of .NET Aspire to use.",
       "displayName": ".NET Aspire version",
       "datatype": "choice",
       "isEnabled": "Framework == net8.0",
```
baronfel, line 91: I think you can remove the `AspireVersionCli` symbol entirely if you loosen this enablement condition to allow `Framework` to be either `net8.0` or `net9.0`. I think this is why `--framework net9.0 --aspire-version 9.0` doesn't actually impact the aspire version to be used.

DamianEdwards: Check my latest commit for what I think *should* work for VS and CLI, but only works in VS. The commit before that is what I had to do to make it work in both.

DamianEdwards:
Due to limitations in the templating experience, I had to change how `aspire-empty` works specifically so that the VS flows for adding Aspire orchestration via File -> New and Solution Explorer context menu work. They require that the `AspireVersion` option always be enabled as that's what they set (they don't know about `AspireVersionNet9`) so **just** for the empty template it's possible to select `net9.0` and Aspire version `8.2` but it will be ignored and use `9.1` instead. The display for that option has been updated to make this clearer: 

image

In the CLI scenario it's described like this:

image


FYI @phenning

PR #7232, opened 2025-01-24
Allow disabling proxy for persistent lifetime container endpoints

danegsta:
## Description

The endpoint port assignment for persistent lifetime containers can be confusing due to Aspire creating proxied endpoints for all resources by default. This means that the ports reported for persistent containers in the Aspire dashboard are only valid while the App Host is running and the only way to access persistent containers while the App Host is not running is to inspect them and figure out what random port the container endpoint was exposed on.

This PR adds a new opt-in (and experimental) resource builder API `.WithEndpointProxySupport(false)` that can disable all proxies for the endpoints of a given container resource. In the long run we'd like to make this the new default behavior for persistent containers, but there is additional service discovery work that will need to happen before we can safely make proxy-less persistent containers the default behavior. 

The main thing to be aware of when applying `.WithEndpointProxySupport(false)` to a container resource is that by default Aspire will attempt to use the (internal container) target port as the host port to expose the service. This significantly increases the risk of port conflicts or other issues allocating ports with the default behavior endpoint behavior. It's recommended that any resources being run with endpoint proxies disabled has explicit (unique) host ports set for all its endpoints.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No


src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
```diff
@@ -1070,6 +1083,15 @@ private void PrepareServices()
...
             }
         }
+
+        if (endpointsWithHostUnset.Any())
+        {
+            _logger.LogWarning("You have unproxied container endpoints without an explicit host port set. By default these endpoints will attempt to bind a host port that matches the container target port. This can lead to port conflicts if multiple containers are using the same target port(s). For containers running with a persistent lifetime or container endpoints with IsProxied disabled, we recommend specifying explicit host ports. For more information on container networking in .NET Aspire see: https://aka.ms/dotnet/aspire/container-networking");
+            foreach (var (resource, endpoint) in endpointsWithHostUnset)
+            {
+                _logger.LogInformation("'{EndpointName}' endpoint for '{ResourceName}' doesn't have a host port set. Attempting to bind host port '{TargetPort}'.", endpoint.Name, resource.Name, endpoint.TargetPort);
+            }
```
danegsta, line 1093: Maybe bump these down to `LogInformation` and `LogDebug` respectively? My only concern is that `LogInformation` from `ApplicationExecutor` doesn't normally show up, so we'd only surface the guidance if a user increases their logging verbosity.

davidfowl, line 1093: Make this single log statement?

davidfowl:
@JamesNK moved stuff around 😄

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -83,6 +83,11 @@ public int? Port
         }
     }
 
+    /// <summary>
+    /// Indicates the the port is set with an explicit value.
+    /// </summary>
+    public bool IsPortSet => _port != null;
```
davidfowl, line 89: Does this need to be public?

danegsta: Not really; switched it to internal.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
danegsta: Switched to appending the details to a single log entry.

danegsta: Logs currently look like this with a warning level. DcpExecutor logs below warning don't currently get surfaced with default log levels:
image

danegsta: Made a large refactor to allow default unproxied endpoints for persistent containers and support dynamic port allocation for unproxied container endpoints.

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -30,8 +27,8 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     /// <param name="port">Desired port for the service.</param>
     /// <param name="targetPort">This is the port the resource is listening on. If the endpoint is used for the container, it is the container port.</param>
     /// <param name="isExternal">Indicates that this endpoint should be exposed externally at publish time.</param>
-    /// <param name="isProxied">Specifies if the endpoint will be proxied by DCP. Defaults to true.</param>
-    public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, string? transport = null, [EndpointName] string? name = null, int? port = null, int? targetPort = null, bool? isExternal = null, bool isProxied = true)
+    /// <param name="isProxied">Specifies if the endpoint will be proxied by DCP. Defaults to null.</param>
+    public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, string? transport = null, [EndpointName] string? name = null, int? port = null, int? targetPort = null, bool? isExternal = null, bool? isProxied = null)
```
danegsta, line 31: We could go with a hybrid approach and still force persistent endpoints to be un-proxied if we don't want these API changes.

src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
```diff
@@ -110,19 +110,16 @@ public static IResourceBuilder<KafkaServerResource> WithKafkaUI(this IResourceBu
...
 
-            builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>((e, ct) =>
+            builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(kafkaUi, (e, ct) =>
             {
                 var kafkaResources = builder.ApplicationBuilder.Resources.OfType<KafkaServerResource>();
 
                 int i = 0;
                 foreach (var kafkaResource in kafkaResources)
                 {
-                    if (kafkaResource.InternalEndpoint.IsAllocated)
```
danegsta, line 120: For container-to-container, we don't need to wait for the endpoint to be allocated.

src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
```diff
@@ -110,19 +110,16 @@ public static IResourceBuilder<KafkaServerResource> WithKafkaUI(this IResourceBu
                 .WithHttpEndpoint(targetPort: KafkaUIPort)
                 .ExcludeFromManifest();
 
-            builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>((e, ct) =>
+            builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(kafkaUi, (e, ct) =>
```
danegsta, line 113: AfterEndpointsAllocatedEvent was being used to queue events before the app model resources were created, but with the changes in the PR for un-proxied container endpoints, it's no longer guaranteed to fire synchronously before resources are created. Using the BeforeResourceStartedEvent works as a replacement.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -869,12 +870,6 @@ internal static IResourceBuilder<T> WithHttpHealthCheckInternal<T>(this IResourc
                 throw new DistributedApplicationException($"The endpoint '{endpointName}' on resource '{builder.Resource.Name}' was not using the '{desiredScheme}' scheme.");
             }
 
-            return Task.CompletedTask;
-        });
-
-        Uri? uri = null;
-        builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(builder.Resource, (@event, ct) =>
```
danegsta, line 876: BeforeResourceStartedEvent may fire before AfterEndpointsAllocatedEvent, so it makes sense to simply get the uri in the AfterEndpointsAllocatedEvent callback.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -351,6 +351,7 @@ public static IResourceBuilder<TDestination> WithReference<TDestination>(this IR
         connectionName ??= resource.Name;
 
         builder.WithRelationship(resource, KnownRelationshipTypes.Reference);
+        (builder as IResourceBuilder<IResourceWithWaitSupport>)?.WaitFor(source);
```
danegsta, line 354: I need to make this a bit smarter; basically we may need to wait for some references if the referenced endpoint may not be allocated before the resources are created.

src/Aspire.Hosting/Dcp/ApplicationExecutor.cs
danegsta: Ended up reworking things to no longer require logging

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
danegsta: Reverted the potential API changes.

src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
danegsta: Reverted these particular changes.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
danegsta: Reverted some of these changes; will need to be part of a larger potential rework for a future release.

davidfowl (approved):

davidfowl:
@danegsta good discussion and tradeoffs 😄. Can you make sure we document and update the relevant issue in 9.1?

PR #7233, opened 2025-01-24
Add snapshot version tracking and improve logging

davidfowl:
- Introduced a new `Id` property in `CustomResourceSnapshot` for unique identification of snapshots.
- Updated `WatchAsync` to filter resource events based on the maximum snapshot ID.
- Modified `PublishUpdateAsync` to increment the snapshot ID for new states.
- Enhanced logging in `PublishUpdateAsync` to include the new snapshot ID for better traceability.

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -18,6 +18,8 @@ public sealed record CustomResourceSnapshot
     private readonly ImmutableArray<HealthReportSnapshot> _healthReports = [];
     private readonly ResourceStateSnapshot? _state;
 
+    internal long Id { get; set; }
+
     /// <summary>
     /// The type of the resource.
     /// </summary>
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -21,7 +21,7 @@ public sealed record CustomResourceSnapshot
     /// <summary>
     /// Monotonically increasing version number for the snapshot.
     /// </summary>
-    internal long Version { get; set; }
+    internal long Version { get; init; }
 
     /// <summary>
     /// The type of the resource.
--- a/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
+++ b/src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
@@ -18,7 +18,10 @@ public sealed record CustomResourceSnapshot
     private readonly ImmutableArray<HealthReportSnapshot> _healthReports = [];
     private readonly ResourceStateSnapshot? _state;
 
-    internal long Id { get; set; }
+    /// <summary>
+    /// Monotonically increasing version number for the snapshot.
+    /// </summary>
+    internal long Version { get; set; }
 
     /// <summary>
     /// The type of the resource.
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -375,7 +375,7 @@ public class ResourceNotificationService : IDisposable
             var newState = stateFactory(previousState);
 
             // Increment the snapshot version, this is a per resource version.
-            newState = newState with { Version = notificationState.LastVersion++ };
+            newState = newState with { Version = notificationState.GetNextVersion() };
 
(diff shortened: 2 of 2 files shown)


src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -326,10 +316,30 @@ void WriteToChannel(ResourceEvent resourceEvent) =>
             OnResourceUpdated += WriteToChannel;
         }
 
+        long id = 0;
```
JamesNK, line 319: nit :smile:
```suggestion
        var id = 0L;
```

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -326,10 +316,30 @@ void WriteToChannel(ResourceEvent resourceEvent) =>
...
+
+                yield return new ResourceEvent(resource, resourceId, state.Value.LastSnapshot);
+            }
+        }
+
         try
         {
             await foreach (var item in channel.Reader.ReadAllAsync(cancellationToken).ConfigureAwait(false))
             {
+                if (item.Snapshot.Id <= id)
```
JamesNK, line 338: When could this happen? Multiple publishes from different threads?

If this is unexpected, should a log be written?

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -18,6 +18,8 @@ public sealed record CustomResourceSnapshot
     private readonly ImmutableArray<HealthReportSnapshot> _healthReports = [];
     private readonly ResourceStateSnapshot? _state;
 
+    internal long Id { get; set; }
```
JamesNK, line 21: init instead of set?
```suggestion
    // An internal id for debugging purposes
    internal long Id { get; init; }
```

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: It's something we can't avoid without more synchronization (which I am trying to avoid), so it isn't worth logging (well maybe in trace we can?). We're missing events in some cases, so this is trying to make sure we don't get old events.

JamesNK: How does that happen? PublishUpdateAsync happens in a lock, which writes to the happen in order, and the channel reads in order. When do things go out of order?

davidfowl: The lock is per resource, not globally. We return the last snapshot before we subscribe for updates, in between those 2 operations where we can drop new publishes.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -326,10 +315,35 @@ void WriteToChannel(ResourceEvent resourceEvent) =>
...
+        }
+
         try
         {
             await foreach (var item in channel.Reader.ReadAllAsync(cancellationToken).ConfigureAwait(false))
             {
+                // Skip events that are older than the max version we have seen so far. This avoids duplicates.
+                if (versionsSeen.TryGetValue((item.Resource, item.ResourceId), out var maxVersionSeen) && item.Snapshot.Version <= maxVersionSeen)
+                {
+                    versionsSeen.Remove((item.Resource, item.ResourceId));
```
JamesNK, line 343: A comment saying why it is ok to remove this would be good.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -360,6 +374,9 @@ public Task PublishUpdateAsync(IResource resource, string resourceId, Func<Custo
 
             var newState = stateFactory(previousState);
 
+            // Increment the snapshot version, this is a per resource version.
+            newState = newState with { Version = notificationState.LastVersion++ };
```
JamesNK, line 378: nit: Make LastVersion a private field, and have a method called `GetNextVersion()` on notification state that increments and returns a value?

JamesNK (approved):
I think this is ok.

I think we should consider having some kind of stress test for the notification service that blasts it with load in various situations and verifies it still works. Doesn't need to happen now.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: This whole type is private 😄

Issue #7215, opened 2025-01-23
DCP not cleaning up networks


davidfowl:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=926163
Build error leg or test failing: Aspire.Hosting.MongoDB.Tests.MongoDbFunctionalTests.Aspire.Hosting.MongoDB.Tests.MongoDbFunctionalTests.VerifyWaitForOnMongoBlocksDependentResources
Pull request: https://github.com/dotnet/aspire/pull/7185

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

karolz-ms:
So far I occasionally see leaked containers when running `Aspire.Hosting.Tests/DistributedApplicationTests`

davidfowl:
Adding to the pile https://github.com/dotnet/aspire/issues/7366

davidfowl:
This one has dcp logs

## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=937402
Build error leg or test failing: Aspire.Hosting.Tests.WaitForTests.DependencyWithGreaterThan1ReplicaAnnotationWaitsForAll
Pull request: https://github.com/dotnet/aspire/pull/7367

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md).



```json
{
  "ErrorMessage": "",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

karolz-ms:
This is great specimen, lots of container network cleanup failures, looking...

davidfowl:
I reverted @ReubenBond's [PR](https://github.com/dotnet/aspire/pull/7371) which really had nothing to do with this specifically, but added 15 tests and that seemed to trigger this failure condition consistently.

karolz-ms:
[...]

PR #7237, opened 2025-01-24
Run this after not before

davidfowl:
Contributes to https://github.com/dotnet/aspire/issues/7215


davidfowl:
Not waiting for a build.

PR #7242, opened 2025-01-24
Add role assignment to CosmosDB Entra ID

eerhardt:
We missed adding it to the infra. Adding the role so it gets put in the bicep file.


davidfowl (approved):

captainsafia (approved):
iconic bug fix ✨

PR #7243, opened 2025-01-25
Use BeforeResourceStartedEvent instead of AfterEndpointsAllocatedEvent when scheduling setup before resource creation

danegsta:
## Description

We're using the `AfterEndpointsAllocatedEvent` in several places as a callback to do setup before resources are created. While it's true that historically `AfterEndpointsAllocatedEvent` has always been triggered before resources are created, that's mainly due to a limitation in the endpoints model (we don't currently support dynamically allocated host ports for un-proxied container endpoints).

The appropriate event to schedule additional model configuration before a resource is created is `BeforeResourceStartedEvent` which is sent just before the resource is created.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
+++ b/src/Aspire.Hosting.Kafka/KafkaBuilderExtensions.cs
@@ -110,19 +110,16 @@ public static class KafkaBuilderExtensions
                 .WithHttpEndpoint(targetPort: KafkaUIPort)
                 .ExcludeFromManifest();
 
-            builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>((e, ct) =>
+            builder.ApplicationBuilder.Eventing.Subscribe<BeforeResourceStartedEvent>(kafkaUi, (e, ct) =>
             {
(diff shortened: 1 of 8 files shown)


davidfowl:
I don't think this works because the UI is a singleton that works with all of the target resources of that type.

JamesNK:
`BeforeResourceStartedEvent` is now called when a resource is restarted. Keep in mind for whether this is a good change.

danegsta:
> `BeforeResourceStartedEvent` is now called when a resource is restarted. Keep in mind for whether this is a good change.

That depends on whether its consistently fired before the resource is created in the orchestrator. If so, it shouldn't cause any issues with these changes, we'd just do another pass of setup housekeeping before the new container gets created.

PR #7249, opened 2025-01-25
Support referencing existing Service Bus resources in references

captainsafia:
Opening this PR to sanity check the approach before adding it to other resource types. Some things of note:

- Support for referencing existing resources in other subscription IDs/resource groups isn't implemented due to a gap in the underlying provisioning APIs. See https://github.com/Azure/azure-sdk-for-net/issues/47980. There are workarounds for the but the juice might not be worth the squeeze.
- Separate APIs for publishing/running with an existing resource backed by an annotation and explicit checks in the `ConfigureInfrastructure` calls for each resource.


davidfowl:
> Opening this PR to sanity check the approach before adding it to other resource types. Some things of note:

This should be a draft right?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -30,23 +30,33 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
...
             var skuParameter = new ProvisioningParameter("sku", typeof(string))
             {
                 Value = "Standard"
             };
             infrastructure.Add(skuParameter);
 
-            var serviceBusNamespace = new AzureProvisioning.ServiceBusNamespace(infrastructure.AspireResource.GetBicepIdentifier())
+            AzureProvisioning.ServiceBusNamespace? serviceBusNamespace;
+            if (infrastructure.AspireResource.TryGetExistingResource(out var existingAnnotation) &&
+                builder.ExecutionContext.IsPublishMode == existingAnnotation.IsPublishMode)
```
davidfowl, line 43: Why is there a check for the model?

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// The name of the existing resource.
+    /// </summary>
+    public string Name { get; } = name;
+
+    /// <summary>
+    /// True if the existing resource is reference in publish mode.
+    /// </summary>
+    public bool IsPublishMode { get; } = isPublishMode;
```
davidfowl, line 22: Why is this needed?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -30,23 +30,33 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
...
         {
             var skuParameter = new ProvisioningParameter("sku", typeof(string))
             {
                 Value = "Standard"
             };
             infrastructure.Add(skuParameter);
 
-            var serviceBusNamespace = new AzureProvisioning.ServiceBusNamespace(infrastructure.AspireResource.GetBicepIdentifier())
+            AzureProvisioning.ServiceBusNamespace? serviceBusNamespace;
+            if (infrastructure.AspireResource.TryGetExistingResource(out var existingAnnotation) &&
```
captainsafia, line 42: Is there a way to resolve the execution context from the `AzureResourceInfrastructure` type. Is there a way to do this

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Represents a resource that is not managed by Aspire's provisioning or
+/// container management layer.
+/// </summary>
+public class ExistingResourceAnnotation(string name, bool isPublishMode = false) : IResourceAnnotation
```
davidfowl, line 12: Is this supposed to be Azure specific?

captainsafia: Based on what we decided, yes. Are you indicating that the type name should indicate it's an Existing*Azure*ResourceAnnotation?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
davidfowl: There may not be but this method has access to the builder in a closure right?

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
captainsafia: To distinguish between `RunAsExisting` vs `PublishAsExisting`.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
davidfowl: It's in the wrong assembly.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
davidfowl: I don't this that needs to be part of the annotation. This should just have the data required to reference the resource. It doesn't need the mode. Run/PublishAs are just idioms in the app model driven by the current ExecutionContext mode.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
davidfowl: This should be ExistingAzureResourceAnnotation and it needs the subscription id, resource group and name as properties.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
captainsafia: How would we indicate that behavior needs to be gated on the `ExecutionContext` in this case?

davidfowl: I see why you say that becase these are implemented as a generic annotation instead of methods on the specific builder. Maybe we should reconsider that?

davidfowl: Actually, I'm dumb this is much simpler than that. Is RunAsExisting, check the mode, if it's run add the attribute, same for PublishAsExisting. The annotation doesn't need the mode, it just needs to know if there's data about where this resource is.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
captainsafia: I remove subscription ID and resource group from this model because of:

> Support for referencing existing resources in other subscription IDs/resource groups isn't implemented due to a gap in the underlying provisioning APIs. See https://github.com/Azure/azure-sdk-for-net/issues/47980. There are workarounds for the but the juice might not be worth the squeeze.

We can't materialize these into scopes in the existing reference in the Bicep identifier at the moment so it doesn't make sense to expose an API for something that isn't supported at the moment.

We can add them later easily enough but if we have them now they'd be non-functional.

davidfowl: I think we should at a minimum add the resource group so that we can play with what it would look like in a custom resource even though the Azure.Provisioning libs can't directly do it yet (we can try it out with the bicep resource). The reason it's important is because in a normal aspire deployment, the resource group is created on the fly, so at a minimum, using exisiting resource would look like a name + a different resource group in the same subscription.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
```diff
@@ -291,13 +291,13 @@ private async Task ProcessResourceAsync(IConfiguration configuration, Lazy<Task<
 
             resourceLogger.LogWarning("No provisioner found for {resourceType} skipping.", resource.GetType().Name);
         }
-        else if (!provisioner.ShouldProvision(configuration, resource.AzureResource))
+        else if (!resource.AzureResource.IsExisting() && !provisioner.ShouldProvision(configuration, resource.AzureResource))
```
davidfowl, line 294: Why do you need these checks?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -30,23 +30,33 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
     {
         builder.AddAzureProvisioning();
 
-        var configureInfrastructure = static (AzureResourceInfrastructure infrastructure) =>
+        var configureInfrastructure = (AzureResourceInfrastructure infrastructure) =>
```
davidfowl, line 33: Can you revert this?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
davidfowl: You can remove this too.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <typeparam name="T">The type of the resource.</typeparam>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="name">The name of the existing resource.</param>
+    /// <returns>The resource builder with the existing resource annotation added.</returns>
+    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, string name)
+        where T : IResource
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        builder.Resource.Annotations.Add(new ExistingAzureResourceAnnotation(name));
```
davidfowl, line 39:
```suggestion
        if (builder.ApplicationBuilder.ExecutionContext.IsRunMode)
        {
            builder.Resource.Annotations.Add(new ExistingAzureResourceAnnotation(name));
        }
```

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
captainsafia: Yep, with the idea to set the annotation conditionally in the extension method we can remove this.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
captainsafia: To avoid provisioning a resource or reading the config that has been cached in user-secrets when the same config name is stored. Without this, starting with:

```csharp
var bus = builder.AddServiceBus("messages");
```

And moving to:

```csharp
var bus = builder.AddServiceBus("messages").RunAsExisting("cached-messaging");
```

would still load the config for the Aspire-managed "messages" resource from user-secrets.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
captainsafia: > The reason it's important is because in a normal aspire deployment, the resource group is created on the fly, so at a minimum, using exisiting resource would look like a name + a different resource group in the same subscription.

Hmmm...I see. Are you cool with iterating on this in a separate PR?

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// <summary>
+    /// Marks the resource as an existing resource when the application is running.
+    /// </summary>
+    /// <typeparam name="T">The type of the resource.</typeparam>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="name">The name of the existing resource.</param>
+    /// <returns>The resource builder with the existing resource annotation added.</returns>
+    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, string name)
+        where T : IResource
```
eerhardt, line 35: Should this be IAzureResource?

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        return builder;
+    }
+
+    /// <summary>
+    /// Gets the <see cref="ExistingAzureResourceAnnotation" /> if the resource is an existing resource.
+    /// </summary>
+    /// <param name="resource">The resource to check.</param>
+    /// <param name="existingResource">The existing resource annotation if the resource is an existing resource.</param>
+    /// <returns>True if the resource is an existing resource, otherwise false.</returns>
+    public static bool TryGetExistingResource(this IResource resource, [NotNullWhen(true)] out ExistingAzureResourceAnnotation? existingResource)
```
eerhardt, line 73:
```suggestion
    public static bool TryGetExistingResourceAnnotation(this IResource resource, [NotNullWhen(true)] out ExistingAzureResourceAnnotation? existingResource)
```
?

This isn't really getting the "existing resource", right?

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -0,0 +1,194 @@
...
+            param location string = resourceGroup().location
+
+            param sku string = 'Standard'
+
+            param principalType string
+
+            param principalId string
+
+            resource messaging 'Microsoft.ServiceBus/namespaces@2024-01-01' existing = {
+              name: 'existing-resource-name'
```
eerhardt, line 46: It feels like this needs to be a parameter to the bicep, and not a hard-coded name into the bicep.

src/Aspire.Hosting/ApplicationModel/ExistingResourceAnnotation.cs
davidfowl: No, I think this pr is where we iterate

src/Aspire.Hosting.Azure/Provisioning/Provisioners/AzureProvisioner.cs
davidfowl: Shouldn't that just work? I'm confused what this done.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// Gets the <see cref="ExistingAzureResourceAnnotation" /> if the resource is an existing resource.
+    /// </summary>
+    /// <param name="resource">The resource to check.</param>
+    /// <param name="existingAzureResourceAnnotation">The existing resource annotation if the resource is an existing resource.</param>
+    /// <returns>True if the resource is an existing resource, otherwise false.</returns>
+    public static bool TryGetExistingAzureResourceAnnotation(this IResource resource, [NotNullWhen(true)] out ExistingAzureResourceAnnotation? existingAzureResourceAnnotation)
+    {
+        ArgumentNullException.ThrowIfNull(resource);
+
+        existingAzureResourceAnnotation = resource.Annotations.OfType<ExistingAzureResourceAnnotation>().FirstOrDefault();
```
eerhardt, line 77: For annotaitons, we usually have either:

* respect multiple (not applicable here)
* expect single
* Last one wins

https://github.com/dotnet/aspire/blob/706bbbf4d12957ac0ada81f4247f65c2e8b54c88/src/Aspire.Hosting/ApplicationModel/ResourceExtensions.cs#L277-L279

I don't think using the "first" one is typical.

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -0,0 +1,203 @@
...
+              parent: messaging
+            }
+
+            output serviceBusEndpoint string = messaging.properties.serviceBusEndpoint
+            """;
+        Assert.Equal(expectedBicep, BicepText);
+    }
+
+    [Fact]
+    public async Task RequiresPublishAsExistingInPublishMode()
```
eerhardt, line 74: (nit) Is this test name correct? the test calls `RunAsExisting`

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -0,0 +1,203 @@
...
+              parent: messaging
+            }
+
+            output serviceBusEndpoint string = messaging.properties.serviceBusEndpoint
+            """;
+        Assert.Equal(expectedBicep, BicepText);
+    }
+
+    [Fact]
+    public async Task AddExistingAzureServiceBusInPublishModeMode()
```
eerhardt, line 143: (nit) `InPublishModeMode`?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -38,15 +38,25 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
             };
             infrastructure.Add(skuParameter);
 
-            var serviceBusNamespace = new AzureProvisioning.ServiceBusNamespace(infrastructure.AspireResource.GetBicepIdentifier())
+            AzureProvisioning.ServiceBusNamespace? serviceBusNamespace;
+            if (infrastructure.AspireResource.TryGetExistingAzureResourceAnnotation(out var existingAnnotation))
             {
-                Sku = new AzureProvisioning.ServiceBusSku()
+                var existingResourceName = existingAnnotation.NameParameter.AsProvisioningParameter(infrastructure, $"{infrastructure.AspireResource.GetBicepIdentifier()}ExistingResourceName");
```
eerhardt, line 44: nit - Is `ExistingResourceName` too verbose?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusExtensions.cs
```diff
@@ -86,7 +96,7 @@ public static IResourceBuilder<AzureServiceBusResource> AddAzureServiceBus(this
                         infrastructure.Add(cdkRule);
```
eerhardt, line 96: Does generating these Queues and Topics work with an existing resource?

src/Aspire.Hosting.Azure/ExistingAzureResourceAnnotation.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Represents a resource that is not managed by Aspire's provisioning or
+/// container management layer.
+/// </summary>
+public class ExistingAzureResourceAnnotation(IResourceBuilder<ParameterResource> nameParameter) : IResourceAnnotation
+{
+    /// <summary>
+    /// Gets the name of the existing resource.
+    /// </summary>
+    public IResourceBuilder<ParameterResource> NameParameter { get; } = nameParameter;
+}
```
davidfowl, line 18:
```suggestion
public class ExistingAzureResourceAnnotation(ParameterResource nameParameter) : IResourceAnnotation
{
    /// <summary>
    /// Gets the name of the existing resource.
    /// </summary>
    public ParameterResource NameParameter { get; } = nameParameter;
}
```

captainsafia: We don't have an `AsProvisioningParameter` overload that takes the `ParameterResource` directly although that seems like an oversight given what the implementation currently does.

https://github.com/dotnet/aspire/blob/5008bfb621ee0283fdbe5c144a2d58341ff01e54/src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs#L67-L78

I think we can probably just add another overload that takes `ParameterResource` and have the existing one call into it.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
captainsafia: Good point -- the intent here was to follow the expect single pattern.

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
captainsafia: Yeah, the test name is correct although confusing. I was intending to say that you need to call `PublishAsExisting` in publish-mode otherwise the API effectively no-ops.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <param name="nameParameter">The name of the existing resource.</param>
+    /// <param name="resourceGroupParameter"></param>
+    /// <returns>The resource builder with the existing resource annotation added.</returns>
+    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+        where T : IAzureResource
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        // Throw if ExistingResourceAnnotation already exists on resource
+        if (builder.Resource.Annotations.OfType<ExistingAzureResourceAnnotation>().SingleOrDefault() is not null)
```
eerhardt, line 41:
```suggestion
        if (builder.Resource.IsExisting())
```

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="nameParameter">The name of the existing resource.</param>
+    /// <param name="resourceGroupParameter"></param>
+    /// <returns>The resource builder with the existing resource annotation added.</returns>
+    public static IResourceBuilder<T> PublishAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+        where T : IAzureResource
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        if (builder.Resource.Annotations.OfType<ExistingAzureResourceAnnotation>().SingleOrDefault() is not null)
```
eerhardt, line 67:
```suggestion
        if (builder.Resource.IsExisting())
```

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        return resource.Annotations.OfType<ExistingAzureResourceAnnotation>().SingleOrDefault() is not null;
+    }
+
+    /// <summary>
+    /// Marks the resource as an existing resource when the application is running.
+    /// </summary>
+    /// <typeparam name="T">The type of the resource.</typeparam>
+    /// <param name="builder">The resource builder.</param>
+    /// <param name="nameParameter">The name of the existing resource.</param>
+    /// <param name="resourceGroupParameter"></param>
```
eerhardt, line 33:
```suggestion
    /// <param name="resourceGroupParameter">The name of the existing resource group, or <see langword="null"/> to use the current resource group.</param>
```

src/Aspire.Hosting.Azure/ExistingAzureResourceAnnotation.cs
```diff
@@ -0,0 +1,24 @@
...
+public sealed class ExistingAzureResourceAnnotation(ParameterResource nameParameter,
+    ParameterResource? resourceGroupParameter = null) : IResourceAnnotation
+{
+    /// <summary>
+    /// Gets the name of the existing resource.
+    /// </summary>
+    public ParameterResource NameParameter { get; } = nameParameter;
+
+    /// <summary>
+    /// Gets the name of the existing resource group.
```
eerhardt, line 21:
```suggestion
    /// Gets the name of the existing resource group. If <see langword="null"/>, use the current resource group.
```

src/Aspire.Hosting.Azure/AzureBicepResource.cs
```diff
@@ -134,11 +139,18 @@ public virtual string GetBicepTemplateString()
...
         using var template = GetBicepTemplateFile(Path.GetDirectoryName(context.ManifestPath), deleteTemporaryFileOnDispose: false);
         var path = template.Path;
 
+        if (Scope.Count == 0)
+        {
+            context.Writer.WriteString("type", "azure.bicep.v0");
+        }
+        else
+        {
+            context.Writer.WriteString("type", "azure.bicep.v1");
```
eerhardt, line 151: We should probably update our json schema as well for the new values.

src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
```diff
@@ -76,6 +76,36 @@ public static ProvisioningParameter AsProvisioningParameter(this IResourceBuilde
         return GetOrAddParameter(infrastructure, parameterName, parameterResourceBuilder.Resource.Secret);
```
eerhardt, line 76: Can we update the existing `IResourceBuilder` method to call the new one?

src/Aspire.Hosting.Azure/AzureBicepResource.cs
```diff
@@ -47,6 +47,11 @@ public class AzureBicepResource(string name, string? templateFile = null, string
     /// </summary>
     public TaskCompletionSource? ProvisioningTaskCompletionSource { get; set; }
 
+    /// <summary>
+    /// The scope of the resource.
+    /// </summary>
+    public Dictionary<string, object> Scope { get; } = [];
```
eerhardt, line 53: Can we document this a little better? I'm not sure what the key nor value of this dictionary represents from `The scope of the resource.`.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -105,22 +105,29 @@ await notificationService.PublishUpdateAsync(resource, state =>
         return true;
     }
 
-    public override async Task GetOrCreateResourceAsync(AzureBicepResource resource, ProvisioningContext context, CancellationToken cancellationToken)
+    public override async Task GetOrCreateResourceAsync(AzureBicepResource resource, ProvisioningContext context,CancellationToken cancellationToken)
```
eerhardt, line 108:
```suggestion
    public override async Task GetOrCreateResourceAsync(AzureBicepResource resource, ProvisioningContext context, CancellationToken cancellationToken)
```

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -420,12 +429,16 @@ private static async Task<bool> ExecuteCommand(ProcessSpec processSpec)
         return null;
     }
 
-    internal static string GetChecksum(AzureBicepResource resource, JsonObject parameters)
+    internal static string GetChecksum(AzureBicepResource resource, JsonObject parameters, JsonObject? scope = null)
```
eerhardt, line 432: Do we ever call this without setting `scope`? Can it just be:

```suggestion
    internal static string GetChecksum(AzureBicepResource resource, JsonObject parameters, JsonObject? scope)
```

src/Aspire.Hosting.Azure/AzureBicepResource.cs
eerhardt: Mostly, I'm wondering:

1. why there would ever be more than 1 (this is a Dictionary)
2. What is the key to this dictionary used for? Are there well-known keys?

eerhardt (approved):
Looks really good. Nice work @captainsafia.

src/Aspire.Hosting.Azure/AzureBicepResource.cs
eerhardt: @davidfowl - thoughts? Are we keeping the json schema up to date?

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
captainsafia: Yes, the only place where we didn't was in the test cases but we can update those.

src/Aspire.Hosting.Azure/AzureBicepResource.cs
captainsafia: I've updated the comment here with a remark that contains more details. I'll answer the questions here for posterities sake.

1. It's a dictionary to support adding support for defining resources in separate resource groups _and_ subscriptions in the future.
2. The `resourceGroup` is the only well-known key here. AZD will disregard everything else.

src/Aspire.Hosting.Azure/AzureBicepResource.cs
captainsafia: I've updated the schema in https://github.com/dotnet/aspire/pull/7249/commits/48e1498be3241f03323355904648311ee75b6080. Let me know if the delta looks good.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -139,7 +146,7 @@ await notificationService.PublishUpdateAsync(resource, state => state with
         {
             // This could be done as a bicep template that imports the other bicep template but this is
             // quick and dirty for now
-            var keyVaults = context.ResourceGroup.GetKeyVaults();
+            var keyVaults = resourceGroup.GetKeyVaults();
```
davidfowl, line 149: Testing this code path is interesting...

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <returns>The resource builder with the existing resource annotation added.</returns>
+    public static IResourceBuilder<T> PublishAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+        where T : IAzureResource
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        if (builder.Resource.IsExisting())
+        {
+            throw new InvalidOperationException($"Resource {builder.Resource.Name} is already marked as an existing resource.");
+        }
```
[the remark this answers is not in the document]
captainsafia: I feel like it makes it clearer to reason through the scope a resource is grouped to if we only support one `XAsExisting` invocation on a resource.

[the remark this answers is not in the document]
davidfowl: We need to think about this like a pipeline. It should not throw

src/Aspire.Hosting.Azure/AzureBicepResource.cs
```diff
@@ -47,6 +47,20 @@ public class AzureBicepResource(string name, string? templateFile = null, string
     /// </summary>
     public TaskCompletionSource? ProvisioningTaskCompletionSource { get; set; }
 
+    /// <summary>
+    /// The scope of the resource that will be configured in the main Bicep file.
+    /// </summary>
+    /// <remarks>
+    /// The scope is a weakly-typed collection of key-value pairs that can be used to configure
+    /// the Bicep scope that is emitted by AZD in the module definition for a resource.
```
davidfowl, line 55: Remove azd from the docs

src/Aspire.Hosting.Azure/AzureBicepResource.cs
```diff
@@ -47,6 +47,20 @@ public class AzureBicepResource(string name, string? templateFile = null, string
...
+    /// By default, this scope dictionary will include a single `resourceGroup` key that will
+    /// map to the scope that is generated in the Bice template (for example,
+    /// `scope: resourceGroup(resourceGroupValue)`). This property is only emitted for schema
+    /// versions azure.bicep.v1. Only the `resourceGroup` key is respected as a sub-property although
+    /// more keys may be supported in the future, such as a subscriptionId.
+    /// </remarks>
+    public Dictionary<string, object> Scope { get; } = [];
```
davidfowl, line 62: Why isn't this a type?

PR #7251, opened 2025-01-25
Loose objects with strong views

davidfowl:
- This fundamentally changes the resource model to be all in on annotations. Resources are a set of annotatations and strongly typed "views" can be modeled over those annotations. Each resource has a `ResourceTypeAnnotation` which stores the *actual* resource type. This means that runtime code should never use is or as for resource type checks but should instead use the new ResourceKind API to determine the type of the resource.
- This enables cleanly switching between resource types while keeping the veneer of strong typing at the top level.

TODO:
- Update tests

Issues found:
- We still need to design resources intentionally when want the ability to apply thier configuration to another resource. e.g. RunAsContainer for AzureRedis -> Redis container. This is trivial to do with the change, but doesn't solve the we need to duplicate the logic to configure redis.
- Identity, the identity of the underlying resource needs to be based on the annotation collection, not the object instance.


src/Aspire.Hosting.Azure/PublicAPI.Unshipped.txt
```diff
@@ -40,3 +41,8 @@ static Aspire.Hosting.AzureProvisioningResourceExtensions.AsProvisioningParamete
...
 static Aspire.Hosting.AzureProvisioningResourceExtensions.ConfigureInfrastructure<T>(this Aspire.Hosting.ApplicationModel.IResourceBuilder<T!>! builder, System.Action<Aspire.Hosting.Azure.AzureResourceInfrastructure!>! configure) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<T!>!
 static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
+static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
+static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
+static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
+static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
+static Aspire.Hosting.AzureResourceExtensions.GetBicepIdentifier(this Aspire.Hosting.ApplicationModel.IAzureResource! resource) -> string!
```
davidfowl, line 48: The analyzer is so buggy...

src/Aspire.Hosting.Redis/PublicAPI.Unshipped.txt
```diff
@@ -1,8 +1,10 @@
...
 static Aspire.Hosting.RedisBuilderExtensions.WithHostPort(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Redis.RedisInsightResource!>! builder, int? port) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Redis.RedisInsightResource!>!
 static Aspire.Hosting.RedisBuilderExtensions.WithRedisInsight(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.RedisResource!>! builder, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Redis.RedisInsightResource!>!>? configureContainer = null, string? containerName = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.RedisResource!>!
+static Aspire.Hosting.RedisBuilderExtensions.WithRedisInsight(this Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.RedisResource!>! builder, System.Action<Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.Redis.RedisInsightResource!>!>? configureContainer = null, string? containerName = null) -> Aspire.Hosting.ApplicationModel.IResourceBuilder<Aspire.Hosting.ApplicationModel.RedisResource!>!
```
davidfowl, line 10: Same comment above.

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -42,9 +42,27 @@ public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicat
...
+    /// <summary>
+    /// Configures a container resource for Redis container.
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="port"></param>
+    /// <param name="configureContainer"></param>
+    /// <returns></returns>
+    /// <exception cref="DistributedApplicationException"></exception>
+    /// <exception cref="InvalidOperationException"></exception>
+    public static IResourceBuilder<RedisResource> ApplyRedis(this IResourceBuilder<RedisResource> builder, int? port = null, Action<IResourceBuilder<RedisResource>>? configureContainer = null)
```
davidfowl, line 57: @eerhardt I don't think this will be the final API shape but the idea here is to get the redis defaults applied to an existing builder.

eerhardt: We don't expect "normal" customers to call this method right? Maybe it shouldn't be an extension method so it doesn't clutter up the Intellisense for normal AppHost developers.

danmoseley:
Closing as this is stale. Please reopen if/when you continue with it.

PR #7254, opened 2025-01-26
Allow tags in `WithImage` and `AddContainer` calls

### LINKED ISSUE #4355: Support combined image and tags syntax in WithImage extension method

afscrome:
## Description

As a user new to aspire, but some experience of docker, I'd expect to be able to take any image I pass to `docker run` and pass the same string to `.WithImage(...)` and `.AddContainer(...)`.  Today this doesn't' happen if you include a tag (or digest).  i.e. `WithImage("redis:latest")`.  This is a bit of a pit of failure for new users workign on aspire, so I'm keen to fix this to improve users' first experiences with Aspire.

This PR supersedes #5835 - that Pr tried to do too many things, and got caught up with the sum of those fixes resulting in breaking changes.  This PR focuses really narrowly on solving the new user scenario, and none of the other problems (such as leaving image tag and inserting `latest` later on, or splitting the registry out from the image).  This PR does not preclude fixing those at a later date or in follow on PRs.

Fixes #4355

Contains two potentially api breaking changes:

1. I introduce new exceptions, however these only throw in cases that the container would have failed to start previously.  (e.g. WithImage("foo:bar", "latest") --> ambigious tag exception.   I'd argue that this is really a clarification of error rather than a new one, although if that's a concern, we could pick a rule about which tag wins.

2. I had to change the tag parameter from `string tag = "latest"` to `string? tag = null`.  However my implementation continues to set `ContainerImageAnnotation.Tag` to latest if a tag is not provided on the image property, so there shouldn't be any observable difference in any case except for those that previously didn't work.  I think I could make it work w/out this change, although we would have to pick a tie break rule for 1 as we wouldn't be able to distinguish between 
    ```cs
    WithImage("foo:bar") // tag = bar
    WithImage("foo:bar", "latest") // tag is ambiguous
    ```

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [X] Maybe the API analysers are detecting the change in parameter default value as a problem, but this change is not observable outside as the method continues to set the tag to `latest`.
      - [ ] Yes
      - [ ] No
  - [ ] No


tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
```diff
@@ -11,6 +11,8 @@ namespace Aspire.Hosting.Containers.Tests;
 
 public class ContainerResourceTests
 {
+    //TODO: AddContainer tests here
```
davidfowl, line 14: ??

davidfowl, line 14:
```suggestion
```

tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
```diff
@@ -11,6 +11,7 @@ namespace Aspire.Hosting.Containers.Tests;
 
 public class ContainerResourceTests
 {
+
```
davidfowl, line 14:
```suggestion
```

tests/Aspire.Hosting.Containers.Tests/ContainerResourceTests.cs
afscrome: Oops, was a note to myself.  Thanks for catching.

davidfowl:
I think this change needs more tests showing the outcome of calling WithImage() after one was already set to change the existing images on containers.
e.g.
https://github.com/dotnet/eShop/blob/1ba1cc7931296eea35932f4c04ef32d813f21a04/src/eShop.AppHost/Program.cs#L10-L13

tests/Aspire.Hosting.Containers.Tests/ContainerResourceBuilderTests.cs
```diff
@@ -104,6 +104,92 @@ public void WithImageSHA256ThrowsIfNoImageAnnotation()
...
+    [Fact]
+    public void WithImageThrowsWithConflictingTagAndDigest()
+    {
+        using var builder = TestDistributedApplicationBuilder.Create();
+        var container = builder.AddResource(new TestContainerResource("testcontainer"));
+
+        Assert.Throws<ArgumentOutOfRangeException>(() => container.WithImage("image@sha246:abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcd", "tag"));
+    }
+
+    [Fact]
```
afscrome, line 144: I don't like these tests - particularly the "Overwrite tag with latest if none is specified" part, but they represent the existing behaviour.

I would like to change this (see #5834), but it does introduce some possible breaking changes, so I've left it as is to let this PR narrowly focus on the first time user use case.

tests/Aspire.Hosting.Containers.Tests/ContainerResourceBuilderTests.cs
```diff
@@ -104,6 +104,116 @@ public void WithImageSHA256ThrowsIfNoImageAnnotation()
...
+        AssertImageComponents(container, null, expectedImage, expectedTag, expectedSha256);
+    }
+
+    [Fact]
+    public void WithImageThrowsWithConflictingTag()
+    {
+        using var builder = TestDistributedApplicationBuilder.Create();
+        var container = builder.AddResource(new TestContainerResource("testcontainer"));
+
+        Assert.Throws<ArgumentOutOfRangeException>(() => container.WithImage("image:tag", "anothertag"));
```
davidfowl, line 132: Why out of range? This exception is strange. Why not NotSupported or InvalidOperation (the goto 😄 )

tests/Aspire.Hosting.Containers.Tests/ContainerResourceBuilderTests.cs
davidfowl: At least we will have coverage when we decide to break it 😄

davidfowl (approved):

davidfowl:
Lets see what breaks!

cc @balachir @IEvangelist @joperezr There are some breaking changes here (but reduced to a minimum) FYI.

PR #7261, opened 2025-01-27
Refactor health check service to be more responsive

### LINKED ISSUE #6112: Allow Health Check intervals to be configurable

The Health Checks used as part of wait for currently run every 5 seconds.  If you have a long chain of WaitFor dependencies, these delays can add up to significant amounts of times waiting for no real reason.

I get why a 5 second interval may be wanted in production to avoid generating too much load, but in local dev I don't care about the overhead of running extra health checks and would much rather run them more frequently - I'm tempted to even drop them as low as 100ms.

I'd like to see configuration options to tune the 5 and 30 second health check intervals made into configuration options
https://github.com/dotnet/aspire/blob/cd1f9fd4e5427a1dfcd0bc1eacbd699acdd22874/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs#L64 and https://github.com/dotnet/aspire/blob/cd1f9fd4e5427a1dfcd0bc1eacbd699acdd22874/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs#L116 

Or if we can detect local dev vs published applications, perhaps we could pick some more aggressive defaults for local dev, whilst keeping the safer defaults for published applications.

### Additional context

_No response_

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/6112

The health check service immediately checks health when a resource moves to a running state. However, if the intial report is unhealthy then it waits 5 seconds before checking again. This PR refactors the health check service to be more responsive.

* If a health report isn't healthy then the delay interval starts at 1 second and increases by 1 second per failure up to the existing 5 sec. Healthy delay is still 30 seconds.
* Refactored the health check loop to use an instant interrupt instead of polling the state once per second. The `DelayAsync` method is complicated, but the actual health monitoring loop is much simpler.
* Allow setting the health delays from tests so time based tests run a lot faster. No more 30 second tests.
* No more blocking in tests: `AutoResetEvent` 👎 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
+++ b/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
@@ -13,20 +13,53 @@ namespace Aspire.Hosting.Health;
 
 internal class ResourceHealthCheckService(ILogger<ResourceHealthCheckService> logger, ResourceNotificationService resourceNotificationService, HealthCheckService healthCheckService, IServiceProvider services, IDistributedApplicationEventing eventing, TimeProvider timeProvider) : BackgroundService
 {
-    private readonly Dictionary<string, ResourceEvent> _latestEvents = new();
-    private readonly Dictionary<string, ResourceMonitorState> _resourcesStartedMonitoring = new();
+    private readonly Dictionary<string, ResourceMonitorState> _resourceMonitoringStates = new();
 
-    internal class ResourceMonitorState(CancellationToken serviceStoppingToken)
+    internal class ResourceMonitorState(ResourceEvent initialEvent, CancellationToken serviceStoppingToken)
     {
         private readonly CancellationTokenSource _cts = CancellationTokenSource.CreateLinkedTokenSource(serviceStoppingToken);
+        private TaskCompletionSource? _delayInterruptTcs;
+        private CancellationTokenSource? _delayCts;
 
         public CancellationToken CancellationToken => _cts.Token;
+
+        public ResourceEvent LatestEvent { get; private set; } = initialEvent;
 
         public void StopResourceMonitor()
         {
             _cts.Cancel();
         }
+
+        public void SetLatestEvent(ResourceEvent resourceEvent)
+        {
+            LatestEvent = resourceEvent;
+            _delayInterruptTcs?.TrySetResult();
+        }
+
+        internal async Task DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        {
+            if (_delayCts == null || !_delayCts.TryReset())
+            {
+                _delayCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            }
+            _delayInterruptTcs = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            var completedTask = await Task.WhenAny(Task.Delay(delay, _delayCts.Token), _delayInterruptTcs.Task).ConfigureAwait(false);
+
+            if (completedTask != _delayInterruptTcs.Task)
+            {
+                // Task.Delay won. Check CTS to see if it won because of cancellation.
+                _delayCts.Token.ThrowIfCancellationRequested();
+            }
+            else
+            {
+                _delayCts.Cancel();
+            }
+        }
     }
+
+    // Internal for testing.
+    internal TimeSpan HealthyHealthCheckInterval { get; set; } = TimeSpan.FromSeconds(30);
 
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -13,21 +13,57 @@ namespace Aspire.Hosting.Health;
...
+
+        internal async Task<bool> DelayAsync(TimeSpan delay, CancellationToken cancellationToken)
+        {
+            if (_delayCts == null || !_delayCts.TryReset())
+            {
+                _delayCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            }
+            _delayInterruptTcs = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            var completedTask = await Task.WhenAny(Task.Delay(delay, _delayCts.Token), _delayInterruptTcs.Task).ConfigureAwait(false);
```
davidfowl, line 47: This is old school. We have `WaitAsync` now. It throws an exception but it looks cleaner 😄

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -13,21 +13,57 @@ namespace Aspire.Hosting.Health;
 
 internal class ResourceHealthCheckService(ILogger<ResourceHealthCheckService> logger, ResourceNotificationService resourceNotificationService, HealthCheckService healthCheckService, IServiceProvider services, IDistributedApplicationEventing eventing, TimeProvider timeProvider) : BackgroundService
 {
-    private readonly Dictionary<string, ResourceEvent> _latestEvents = new();
-    private readonly Dictionary<string, ResourceMonitorState> _resourcesStartedMonitoring = new();
+    private readonly Dictionary<string, ResourceMonitorState> _resourceMonitoringStates = new();
 
-    internal class ResourceMonitorState(CancellationToken serviceStoppingToken)
+    internal class ResourceMonitorState(ResourceEvent initialEvent, CancellationToken serviceStoppingToken)
```
davidfowl, line 18: nit: move this to the bottom of the file.

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -13,21 +13,57 @@ namespace Aspire.Hosting.Health;
...
+
         public void StopResourceMonitor()
         {
             _cts.Cancel();
         }
+
+        public void SetLatestEvent(ResourceEvent resourceEvent)
+        {
+            LatestEvent = resourceEvent;
+            _delayInterruptTcs?.TrySetResult();
```
davidfowl, line 36: Do you want the delay to catch missed updates? SetLatestEvent and DelayAsync aren't synchronized. That means a call to SetLatestEvent  then Delay async won't yield.

I would suggest using a channel (instead of a tcs) if you don't want to miss updates. It would also remove allocations.

JamesNK: Missing updates seems fine. There isn't a 1:1 relationship between the state of a resource and its health result, so timing between events will never be precisie. Everything is eventually consistent.

JamesNK: I changed my mind. Previously a missed update when healthy would be caught by the polling in a second. Now there is always a 30 second wait until interrupted.

I added some locking and improved the interrupt decision to consider the resource version and state. I'm not 100% sure why a change in the snapshot should cause health checks to rerun, no tests cover that situation, but I think state change (such as `Running` to another state) is what it makes sense to consider.

davidfowl (approved):
Nit about adding comments but LGTM

src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -38,32 +29,42 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             await foreach (var resourceEvent in resourceEvents.ConfigureAwait(false))
             {
                 var resourceName = resourceEvent.Resource.Name;
+                ResourceMonitorState? state;
 
-                _latestEvents[resourceName] = resourceEvent;
+                lock (_resourceMonitoringStates)
```
eerhardt, line 34: This code isn't multi-threaded, right? There is just 1 thread/loop processing these events. So is there a need to lock here?

eerhardt: I guess there is an internal `GetResourceMonitorState` that is used for tests that could be on a different thread.

JamesNK: Yes. I didn't want tests to fail because the dictionary was in a bad state during a get.

The lock will never be contended in the real world, and this isn't performance critical code, so it seemed better to add some safety in the cause of test stability.

PR #7262, opened 2025-01-27
Fix messy state changes on resource restart

JamesNK:
## Description

A restarted resource's state is set to `Starting` when an event is raised, but deleting the resource in DCP switches it back to a terminal state, e.g. `Exited`. Looks messy in the dashboard.

Fix is to delete the resource and then raise the resource starting event.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/DcpExecutor.cs
@@ -1775,8 +1775,8 @@ internal sealed class DcpExecutor : IDcpExecutor
                 }, cancellationToken).ConfigureAwait(false);
             }
 
-            // Raise event after resource has been deleted. This is required the event sets the status to "Starting" and resources being
-            // deleted will temporarily override the status to a terminal state.
+            // Raise event after resource has been deleted. This is required because the event sets the status to "Starting" and resources being
+            // deleted will temporarily override the status to a terminal state, such as "Exited".
             await _executorEvents.PublishAsync(new OnResourceStartingContext(cancellationToken, resourceType, matchingResource.ModelResource, matchingResource.DcpResource.Metadata.Name)).ConfigureAwait(false);
 
             await _kubernetesService.CreateAsync(resource, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -1777,6 +1775,10 @@ await execution.ExecuteAsync(async (attemptCancellationToken) =>
                 }, cancellationToken).ConfigureAwait(false);
             }
 
+            // Raise event after resource has been deleted. This is required the event sets the status to "Starting" and resources being
+            // deleted will temporarily override the status to a terminal state.
```
eerhardt, line 1779:
```suggestion
            // Raise event after resource has been deleted. This is required because the event sets the status to "Starting" and resources being
            // deleted will temporarily override the status to a terminal state, such as "Exited".
```

src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -1696,8 +1696,6 @@ public async Task StartResourceAsync(string resourceName, CancellationToken canc
 
         try
         {
-            await _executorEvents.PublishAsync(new OnResourceStartingContext(cancellationToken, resourceType, matchingResource.ModelResource, matchingResource.DcpResource.Metadata.Name)).ConfigureAwait(false);
```
eerhardt, line 1699: This used to get raised for ExecutableReplicaSets before. Does it still get raised with this change?

JamesNK: ExecutableReplicaSets is legacy code. It should be removed. But I'll do that in its own PR.

davidfowl (approved):

PR #7275, opened 2025-01-27
Add check to make sure version is valid before comparing to avoid issues with wildcard or fixed versions

### LINKED ISSUE #7102: Using the aspire 9.0.0 SDK with a prerelease nuget version expression fails
```xml
<Project Sdk="Microsoft.NET.Sdk;Microsoft.NET.Sdk.Publish">
	
	<Sdk Name="Aspire.AppHost.Sdk" Version="9.0.0" />

	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFramework>net8.0</TargetFramework>
		<Nullable>enable</Nullable>
		<UserSecretsId>ff7ad17b-2d94-4ac9-a472-b1b236c733c7</UserSecretsId>
		<IsPublishable>true</IsPublishable>
		<NoWarn>NETSDK1023;NETSDK1057</NoWarn>
		<UseAppHost>false</UseAppHost>
	</PropertyGroup>

	<ItemGroup>
		<PackageReference Include="Aspire.Hosting.AppHost" Version="9.*-*" />
	</ItemGroup>

</Project>
```

```
dotnet build C:\dev\git\AspireApps10\AspireApps10.sln /property:GenerateFullPaths=true /consoleloggerparameters:NoSummary /p:Configuration=Debug /p:Platform="Any CPU" 
  Determining projects to restore...
C:\Users\davifowl\.nuget\packages\aspire.apphost.sdk\9.0.0\Sdk\Sdk.targets(80,20): error MSB4184: The expression "[MSBuild]::VersionLessThan(9.*-*, 8.2.0)" cannot be evaluated. Version string was not in a correct format. [C:\dev\git\AspireApps10\samples\AspireApp1.DevHost\AspireApp1.DevHost.csproj]
```

### LINKED ISSUE #7194: Aspire SDK fails build when using exact versions of dependency Aspire.Hosting.AppHost

### Describe the bug

Our company build requires that all external packages are pinned to a specific versions e.g. `[9.0.0]`. It fails the build otherwise.

The Aspire SDK does not honour the square brackets.
The following package reference (note the brackets):
```
  <ItemGroup>
    <PackageReference Include="Aspire.Hosting.AppHost" Version="[9.0.0]"  />
  </ItemGroup>
```

Produces the compile error
```
error MSB4184: The expression "[MSBuild]::VersionLessThan([9.0.0], 8.2.0)" cannot be evaluated. Version string was not in a correct format
```

### Expected Behavior

The pinned version should work - strip `[]` from the SemVer.

### Steps To Reproduce

Square-bracket the package version of `Aspire.Hosting.AppHost` e.g.:

```
<PackageReference Include="Aspire.Hosting.AppHost" Version="[9.0.0]"  />

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by joperezr**
Fixes #7102 
Fixes #7194 

## Description

This should add a check to the aspire sdk to guard against edge cases for folks using versions with wildcards (like 9.*) or exact versions like `[9.0.0]`.

Fixes # (issue)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.AppHost.Sdk/SDK/Sdk.in.targets
+++ b/src/Aspire.AppHost.Sdk/SDK/Sdk.in.targets
@@ -77,7 +77,12 @@
       <_AppHostVersion>%(_AppHostPackageReference.Version)</_AppHostVersion>
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
+    <PropertyGroup Condition="'$(_AppHostVersion)' != ''">
+      <!-- Check if the version is in the format major.minor.patch -->
+      <_IsValidVersion>$([System.Text.RegularExpressions.Regex]::IsMatch('$(_AppHostVersion)', '^\d+\.\d+\.\d+$'))</_IsValidVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and '$(_IsValidVersion)' == 'true' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
       <!-- If we find the version to Aspire.Hosting.AppHost package but it is lower than 8.2.0, then we fall back
       to use the Dashboard and DCP packages that match the version of the installed workload for backwards compatibility.
       This results in the same behavior that we had before moving Dashboard and DCP out of the workload, since the version
--- a/tests/Aspire.Workload.Tests/BuildAndRunTemplateTests.cs
+++ b/tests/Aspire.Workload.Tests/BuildAndRunTemplateTests.cs
@@ -158,6 +158,32 @@ public partial class BuildAndRunTemplateTests : WorkloadTestsBase
         }
     }
 
+    [Theory]
+    [InlineData("9.*-*")]
+    [InlineData("[9.0.0]")]
+    public async Task CreateAndModifyAspireAppHostTemplate(string version)
+    {
+        string id = GetNewProjectId(prefix: $"aspire_apphost_{version.Replace("*", "wildcard").Replace("[", "").Replace("]", "")}");
+        await using var project = await AspireProject.CreateNewTemplateProjectAsync(id, "aspire-apphost", _testOutput, buildEnvironment: BuildEnvironment.ForDefaultFramework);
+
+        ModifyProjectFile(project, version);
+
+        await project.BuildAsync();
+        await project.StartAppHostAsync();
+
+        static void ModifyProjectFile(AspireProject project, string version)
+        {
+            var projectName = Directory.GetFiles(project.AppHostProjectDirectory, "*.csproj").FirstOrDefault();
+            Assert.False(string.IsNullOrEmpty(projectName));
+
+            var projectContents = File.ReadAllText(projectName);
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/Aspire.AppHost.Sdk/SDK/Sdk.in.targets [hunk 1]
```diff
@@ -77,7 +77,12 @@
       <_AppHostVersion>%(_AppHostPackageReference.Version)</_AppHostVersion>
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
+    <PropertyGroup Condition="'$(_AppHostVersion)' != ''">
+      <!-- Check if the version is in the format major.minor.patch -->
+      <_IsValidVersion>$([System.Text.RegularExpressions.Regex]::IsMatch('$(_AppHostVersion)', '^\d+\.\d+\.\d+$'))</_IsValidVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(_AppHostVersion)' != '' and '$(_IsValidVersion)' == 'true' and $([MSBuild]::VersionLessThan('$(_AppHostVersion)', '8.2.0'))">
```
eerhardt, line 85: This won't work for customers who have `[8.1.0]` as their version. Are we OK with that?

Also, at what point do we think we can remove this code since it relates to an out of support version.

eerhardt (approved):
I almost half think we should remove this code altogether, but if we need to leave it in for backcompat (and are OK with the limitations of this fix), this at least unblocks people trying to use the latest.

src/Aspire.AppHost.Sdk/SDK/Sdk.in.targets
joperezr: Yeah, thought about that too, but those folks would just get the error like they used to before this fix. As you mention, the main goal we want for this limited fix is to support folks that want to use latest.

joperezr:
My last push should have fixed the test being added. Enabling auto-merge to get this change in.

src/Aspire.AppHost.Sdk/SDK/Sdk.in.targets
DamianEdwards: I'm surprised there isn't a built-in function or task to cover this scenario (semantic version checking). I assume we spoke to the relevant SMEs?

joperezr: @rainersigwald would you mind sharing your perspective here? For context, We have this check somewhere in a targets that checks if a particular version of a packagereference is less than 8.2.0, but this is failing for folks that have the packagereference like `9.*-*`. Is there a way to do this with a built-in function other than MSBuild VersionLessThan?

rainersigwald: No, we don't have a function for that. AFAIK .NET still doesn't have one we could wrap, though I guess we could call into NuGet like we do for the TF-compat functions. Feel free to file a request at dotnet/msbuild.

PR #7297, opened 2025-01-28
Fix flaky db tests

davidfowl:
- Passwords were not working well, in fact, it was a fluke they work in some cases

Fixes #7296

PR #7305, opened 2025-01-28
Fix AddParameter usage in tests

eerhardt:
When adding the parameter first, and then setting the config, it accidently works because an exception is thrown at first and then the value is retrieved later when the config is set.

Fix this by ensuring the configuration is set first, before calling AddParameter.

For hosting tests for hosting integrations that aren't part of the core/base APIs, I changed the tests to be simple and just pass the constant value into AddParameter.

This is a follow-up to https://github.com/dotnet/aspire/pull/7297.

### DIFF
--- a/tests/Aspire.Hosting.Containers.Tests/WithDockerfileTests.cs
+++ b/tests/Aspire.Hosting.Containers.Tests/WithDockerfileTests.cs
@@ -25,8 +25,8 @@ public class WithDockerfileTests(ITestOutputHelper testOutputHelper)
 
         var (tempContextPath, tempDockerfilePath) = await CreateTemporaryDockerfileAsync(includeSecrets: true);
 
-        var parameter = builder.AddParameter("secret", secret: true);
         builder.Configuration["Parameters:secret"] = "open sesame from env";
+        var parameter = builder.AddParameter("secret", secret: true);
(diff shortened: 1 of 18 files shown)


tests/Aspire.Hosting.Containers.Tests/WithDockerfileTests.cs
```diff
@@ -160,8 +160,8 @@ public async Task WithDockerfileResultsInBuildAttributeBeingAddedToManifest()
         });
         builder.Services.AddLogging(b => b.AddXunit(testOutputHelper));
 
-        var parameter = builder.AddParameter("message");
         builder.Configuration["Parameters:message"] = "hello";
+        var parameter = builder.AddParameter("message");
```
sebastienros, line 164: Why can't it be done with a single call like in several other files?

eerhardt: It could, but I left the usages in the "core/base API" tests to use the configuration for test coverage. Where for the hosting integrations that aren't part of the "core/base API" tests I changed to be simple, since these are the ones that are typically cloned to other hosting tests.

sebastienros (approved):

PR #7315, opened 2025-01-29
Make WaitFor support replicas

### LINKED ISSUE #6083: Allow `WaitForCompletion` with multiple replica instances

`WaitForCompletion` doesn't work when the dependency has multiple replicas: https://github.com/dotnet/aspire/blob/05cbb1cf5e446d5e6fd0c5ebac63fa708af3f964/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs#L188-L191

When all replicas have started, wait for completion should be true.

### Additional context

_No response_

davidfowl:
## Description

Adding replica support to WaitFor*

Fixes #6083

## Checklist
- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -132,55 +132,69 @@ public class ResourceNotificationService : IDisposable
     {
         var resourceLogger = _resourceLoggerService.GetLogger(resource);
         resourceLogger.LogInformation("Waiting for resource '{Name}' to enter the '{State}' state.", dependency.Name, KnownResourceStates.Running);
-
         await PublishUpdateAsync(resource, s => s with { State = KnownResourceStates.Waiting }).ConfigureAwait(false);
-        var resourceEvent = await WaitForResourceCoreAsync(dependency.Name, re => IsContinuableState(re.Snapshot), cancellationToken: cancellationToken).ConfigureAwait(false);
-        var snapshot = resourceEvent.Snapshot;
 
-        if (snapshot.State?.Text == KnownResourceStates.FailedToStart)
+        var names = dependency.GetResolvedResourceNames();
+        var tasks = new Task[names.Length];
+
+        for (var i = 0; i < names.Length; i++)
         {
-            resourceLogger.LogError(
-                "Dependency resource '{ResourceName}' failed to start.",
-                dependency.Name
-                );
-
-            throw new DistributedApplicationException($"Dependency resource '{dependency.Name}' failed to start.");
-        }
-        else if (snapshot.State!.Text == KnownResourceStates.Finished || snapshot.State!.Text == KnownResourceStates.Exited)
-        {
-            resourceLogger.LogError(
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -132,55 +132,69 @@ private async Task WaitUntilHealthyAsync(IResource resource, IResource dependenc
...
-        // otherwise we don't care about their health status.
-        if (dependency.TryGetAnnotationsOfType<HealthCheckAnnotation>(out var _))
+        await Task.WhenAll(tasks).ConfigureAwait(false);
+
+        async Task Core(string displayName, string resourceId)
         {
-            resourceLogger.LogInformation("Waiting for resource '{Name}' to become healthy.", dependency.Name);
-            await WaitForResourceHealthyAsync(dependency.Name, cancellationToken).ConfigureAwait(false);
-        }
+            var resourceEvent = await WaitForResourceCoreAsync(dependency.Name, re => re.ResourceId == resourceId && IsContinuableState(re.Snapshot), cancellationToken: cancellationToken).ConfigureAwait(false);
```
eerhardt, line 150: Why isn't this


```suggestion
            var resourceEvent = await WaitForResourceCoreAsync(displayName, re => re.ResourceId == resourceId && IsContinuableState(re.Snapshot), cancellationToken: cancellationToken).ConfigureAwait(false);
```

?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -204,47 +218,57 @@ public async Task<ResourceEvent> WaitForResourceHealthyAsync(string resourceName
...
-                snapshot.State.Text,
-                snapshot.ExitCode,
-                exitCode
-                );
-
-            throw new DistributedApplicationException(
-                $"Resource '{dependency.Name}' has entered the '{snapshot.State.Text}' state with exit code '{snapshot.ExitCode}', expected '{exitCode}'."
-                );
-        }
+            var resourceEvent = await WaitForResourceCoreAsync(dependency.Name, re => re.ResourceId == resourceId && IsKnownTerminalState(re.Snapshot), cancellationToken: cancellationToken).ConfigureAwait(false);
```
eerhardt, line 239: Same question here:


```suggestion
            var resourceEvent = await WaitForResourceCoreAsync(displayName, re => re.ResourceId == resourceId && IsKnownTerminalState(re.Snapshot), cancellationToken: cancellationToken).ConfigureAwait(false);
```

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
eerhardt: Is it because it explicitly needs to be the Aspire Resource.Name?

davidfowl: Even though they are the same, I want to keep display name for the display only.

eerhardt (approved):

PR #7328, opened 2025-01-30
Added github action to run a subset of tests

davidfowl:
Moving out CI tests to github actions (initial experiment https://github.com/dotnet/aspire/pull/7073)

### DIFF
--- a/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
+++ b/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
@@ -56,7 +56,7 @@ public class ProjectSpecificTests(ITestOutputHelper _testOutput)
         await app.StopAsync();
     }
 
-    [Fact(Skip = "The project reference for this test is skipped in CI.")]
+    [Fact]
     [RequiresDocker]
     [RequiresTools(["func"])]
     public async Task AzureFunctionsTest()
--- a/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
+++ b/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
@@ -8,7 +8,6 @@ using SamplesIntegrationTests;
 using SamplesIntegrationTests.Infrastructure;
 using Xunit;
 using Xunit.Abstractions;
-using System.Reflection;
 
 namespace Aspire.Playground.Tests;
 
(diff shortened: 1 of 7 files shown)


tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
```diff
@@ -56,7 +56,7 @@ await WaitForAllTextAsync(app,
         await app.StopAsync();
     }
 
-    [Fact]
+    [Fact(Skip = "The project reference for this test is skipped in CI.")]
```
joperezr, line 59: @captainsafia looks like this test hasn't been running on CI as the P2P has been conditioned out when `-ci` flag is passed to the build, is that intentional?

https://github.com/dotnet/aspire/blob/1edf95d2f236fccff2d7b9c9cb54d7da88eb7b2e/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj#L51C54-L51C76

This change is just making it so that it is explicitly skipped so GHA doesn't try to run it but please let me know if that's wrong.

joperezr:
I think this should be good to go now, and I'm planning on merging this later today. There are some pending items that I want to address in a follow-up:

- [ ] Remove duplication of skipping tests explicitly and instead just get ActiveIssue attribute to work in GHA
- [ ] Add GitHubActions logger to dotnet test for better integration with actions.
- [ ] Add more test projects to the list of things we run in GHA

FYI: @davidfowl

joperezr (approved):

tests/Aspire.Playground.Tests/ProjectSpecificTests.cs
captainsafia: > @captainsafia looks like this test hasn't been running on CI as the P2P has been conditioned out when -ci flag is passed to the build, is that intentional?

💀 No...I'm a little surprised at this. Is that `-ci` flag actually set. [Here](https://dev.azure.com/dnceng-public/public/_test/analytics?definitionId=274&contextType=build) are the test analytics for this test over the last 30 days off the main branch:

image

joperezr: That is for the whole workitem (meaning the whole test assembly as opposed to the single test) though, right?

joperezr: `ci` flag is set in our pipeline builds so that P2P wouldn't be included, meaning this test would fail like it did in this PR, so I'm guessing it is getting skipped by some different means.

captainsafia: > That is for the whole workitem (meaning the whole test assembly as opposed to the single test) though, right?

I accidentally cropped it from the image but its for the `AzureFunctionsTest` in particular:

image

I think we should remove the constraint in the project file since this test should be running.

joperezr: I did try that and it wasn't working for some reason. Anyway, for the sake of incremental progress I'll go ahead and not run playground tests in GHA for now and we can add that back in a follow up. That will enable the rest of the tests to start running in GHA.

joperezr: I'd be curious for what is happening in helix, I do have a hunch that the test is not running since I'm sure we pass -ci to our tests, so perhaps the Requires func attribute is making the test still shown as run but actually skipped dynamically? I can take a closer look

joperezr: Interesting, looks like playground tests actually don't build on the build machine and instead get built directly on Helix, and in there it seems like they don't pass -ci, so the test might indeed be running after all

PR #7331, opened 2025-01-30
Enable ServiceBus Emulator tests in CI

eerhardt:
Checking to see if these tests are more stable now.


eerhardt:
@sebastienros @davidfowl - these tests passed 2x in a row. Going for 3 now. When do you think we should try turning them back on?

eerhardt:
passed 3x

sebastienros:
Now that they have the correct timeouts and WaitFors it's probably helping. I am ok re-enabling them and monitor.

eerhardt:
Running one more time. If it passes 4x in a row, let's merge and see what happens when it runs a bunch. We can always revert this change as soon as we see them flaking out.

sebastienros (approved):

eerhardt:
One last time. Enabling auto-merge now and it will merge on pass.

Issue #6788, opened 2024-11-25
`AddStandardResilienceHandler` and healthchecks result in noisy logs with TestHost


afscrome:
The default integration test template calls `ConfigureHttpClientDefaults` with `AddStandardResilienceHandler`

https://github.com/dotnet/aspire/blob/b45b0461b789c64d1b1dfa421f63e2e7a09540af/src/Aspire.ProjectTemplates/templates/aspire-nunit/9.0/IntegrationTest1.cs#L19-L22

This however produces a lot of noise in 9.0 with the built in health checks this ends up producing a lot of number of distracting messages in the test host. e.g. 

```
    info: Polly[3]
      Execution attempt. Source: '-standard//Standard-Retry', Operation Key: '', Result: '200', Handled: 'False', Attempt: '0', Execution Time: 1061.2159ms
```

Are these resiliency headers still a good recommendation for tests in 9.0?  In 8.x before healthchecks & readiness, request resiliency was highly needed to make sure the apps had started up properly, but with 9.0's health check support?  (Particularly if the sample tests replace `WaitForResourceAsync` with `WaitForResourceHealthyAsync`)

afscrome:
Here is a slightly more egregious example of noise - this comes from the build server where it's pulling container images for the first time, and so has a lot more waiting (and thus failing of health checks)
```
info: Aspire.Hosting.DistributedApplication[0]
      Aspire version: 9.0.0+01ed51919f8df692ececce51048a140615dc759d
info: Aspire.Hosting.DistributedApplication[0]
      Distributed application starting.
info: Aspire.Hosting.DistributedApplication[0]
      Application host directory is: /mnt/vss/_work/1/s/src/REDACTED
warn: Polly[3]
      Execution attempt. Source: '-standard//Standard-Retry', Operation Key: '', Result: 'An error occurred while sending the request.', Handled: 'True', Attempt: '0', Execution Time: 31.3691ms
      System.Net.Http.HttpRequestException: An error occurred while sending the request.
       ---> System.Net.Http.HttpIOException: The response ended prematurely. (ResponseEnded)
         at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
         --- End of inner exception stack trace ---
         at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

PR #7335, opened 2025-01-30
Add APIs to DistributedApplicationTestingBuilder for testing without a separate AppHost project

ReubenBond:
## Description

This PR implements public APIs for scenario which appears _extensively_ in our own tests (through `TestDistributedApplicationBuilder`) and modifies our own tests to use that new API instead of the bespoke implementation we currently have.

* Adds synchronous `Create` APIs to `DistributedApplicationTestingBuilder` which do not require passing in a type corresponding to an external `AppHost` project.
* Guts `TestDistributedApplicationBuilder` and uses the new `DistributedApplicationTestingBuilder` APIs instead.
* Implements `IDistributedApplicationBuilder` (IDAB) on `IDistributedApplicationTestingBuilder` (IDATB) so that extension methods, etc, on IDAB apply to IDATB as well.
* Makes all test projects Aspire app hosts by including a reference to `Aspire.Hosting.AppHost` in their project files. This ensures they will have DCP accessible. 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes (this improves coverage by dogfooding our testing APIs in more places)
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
        - Docs issue: https://github.com/dotnet/docs-aspire/issues/2534
  - [ ] No


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -228,14 +274,14 @@ private void OnBuilderCreatingCore(DistributedApplicationOptions applicationOpti
 
     private void OnBuilderCreatedCore(DistributedApplicationBuilder applicationBuilder)
     {
+        var services = applicationBuilder.Services;
+        services.AddHttpClient();
+        services.ConfigureHttpClientDefaults(http => http.AddStandardResilienceHandler());
```
davidfowl, line 279: I think @DamianEdwards turned this off by default on purpose.

src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs
```diff
@@ -107,6 +109,32 @@ public static async Task<IDistributedApplicationTestingBuilder> CreateAsync(Type
...
+
+    /// <summary>
+    /// Creates a new instance of <see cref="IDistributedApplicationTestingBuilder"/>.
+    /// </summary>
+    /// <param name="args">The command line arguments to pass to the entry point.</param>
+    /// <param name="configureBuilder">The delegate used to configure the builder.</param>
+    /// <returns>
+    /// A new instance of <see cref="IDistributedApplicationTestingBuilder"/>.
+    /// </returns>
+    public static IDistributedApplicationTestingBuilder Create(string[] args, Action<DistributedApplicationOptions, HostApplicationBuilderSettings> configureBuilder)
```
davidfowl, line 130: YESS!

src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs
```diff
@@ -233,66 +273,159 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
+                configureBuilder(applicationOptions, hostBuilderOptions);
+            });
+
+            builder.Services.AddHttpClient();
+            builder.Services.ConfigureHttpClientDefaults(http => http.AddStandardResilienceHandler());
+
+            return builder;
+        }
+
+        private static Assembly FindApplicationAssembly()
```
davidfowl, line 300: 👀

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -331,7 +331,7 @@ private static IResourceBuilder<ProjectResource> WithProjectDefaults(this IResou
         else
         {
             var appHostDefaultLaunchProfileName = builder.ApplicationBuilder.Configuration["AppHost:DefaultLaunchProfileName"]
-                ?? Environment.GetEnvironmentVariable("DOTNET_LAUNCH_PROFILE");
+                ?? builder.ApplicationBuilder.Configuration["DOTNET_LAUNCH_PROFILE"];
```
davidfowl, line 334: Good catch

src/Aspire.Hosting.Testing/DistributedApplicationTestingBuilder.cs
afscrome: 😍😍😍😍  I was just about to write an issue asking for such an API.

THANK YOU!.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
afscrome: I've found the `AddStandardResilienceHandler` to cause a lot of noise in logs - #6788, I'd prefer not to see this added in a way that can't be avoided.  If your app makes heavy use of WaitFor / HealthChecks, the resiliency handler may not be needed, so the `resiliency` may translate to `unnesescary retries on failures`.

If keeping `AddStandardResilienceHandler` by default, remove it from the standard integration test template as it's no longer needed: https://github.com/dotnet/aspire/blob/e6d499928142356e9a709f6b258d13b5d8179df8/src/Aspire.ProjectTemplates/templates/aspire-xunit/9.1/IntegrationTest1.cs#L21

davidfowl (approved):

PR #7337, opened 2025-01-30
Add the ability to set parent relationships in the dashboard

### LINKED ISSUE #7259: Nest sidecar type resources by default in Dashboard

9.1's dashboard is adding support for showing parent / child resources as nested in the UI.  This works by default for IResourceWithParent implementations (such as databases), but there are other instances where this could also be done - particularly for sidecar type containers such as redis insights, redis commander, pg admin etc.

Image

As well as built in resources, I'd also like to be able to specify this on my own executables and containers - e.g. a migrator executable resource that runs `dotnet ef update`.

I'd like to see an API on IResourceBuilder<T> which I can use to set the parent - e.g. `migrator.WithParent(database)`, and to have this call set on appropriate resources built in to Aspire.

### Additional context

```
<PackageReference Include="Aspire.Hosting.AppHost" Version="9.1.0-preview.1.25075.3" />
```

I have been able to make this somewhat work with the following code, but it's very temperamental - only works about roughly 10% of the time - trying to publish resource notifications for aspire managed resources seems to be fighting with Aspire's built in orchestration.

```cs
builder.Eventing.Subscribe<BeforeResourceStartedEvent>(migrator.Resource, async (evt, ct) =>
{
    var resourceNotificationService = evt.Services.GetRequiredService<ResourceNotificationService>();

    await resourceNotificationService.PublishUpdateAsync(migrator.Resource, x => x with
    {
        Properties = [
            ..x.Properties,
            new("resource.parentName", sql.Resource.Name)
            ]
    });
});
```

eerhardt:
## Description

Add a new API - WithParentRelationship that can be set to nest resources.

Fix an issue in snapshot properties where it was overwriting previous properties by setting to
a new collection on publish. Needs to update/append to the property collection.

Allow Projects to be parents, not just Containers.

Fix #7259

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -64,8 +64,7 @@ public static class AzureFunctionsProjectResourceExtensions
             }
             else
             {
-                storage = builder.AddAzureStorage(storageResourceName)
-                    .RunAsEmulator().Resource;
+                storage = builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
             }
         }
 
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -64,7 +64,9 @@ public static class AzureFunctionsProjectResourceExtensions
             }
             else
             {
-                storage = builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
+                storage = builder.AddAzureStorage(storageResourceName)
+                    .WithParentRelationship(resource)
+                    .RunAsEmulator().Resource;
             }
         }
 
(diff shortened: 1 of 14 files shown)


eerhardt:
I found 3 more places we should set this by default - our azure emulators that need to spin up other containers.

image

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -64,7 +64,9 @@ public static class AzureFunctionsProjectResourceExtensions
             }
             else
             {
-                storage = builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
+                storage = builder.AddAzureStorage(storageResourceName)
+                    .WithParentRelationship(resource)
```
davidfowl, line 68: Not this one, this is global.
```suggestion
```

src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -64,7 +64,8 @@ public static class AzureFunctionsProjectResourceExtensions
             }
             else
             {
-                storage = builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
+                storage = builder.AddAzureStorage(storageResourceName)
+                    .RunAsEmulator().Resource;
```
davidfowl, line 68:
```suggestion
                storage = builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
```

davidfowl (approved):

davidfowl:
I assume executables also work.

eerhardt:
> I assume executables also work.

Yes, they do.

PR #7343, opened 2025-01-31
Revert "Make WaitFor support replicas (#7315)"

davidfowl:
This reverts commit fe035626eef5ff124a7c73b301ee233fd67049d1.
Seeing failures after merging this. Will see if this makes things green and look into why some tests hang.


davidfowl:
Hmm it doesn't look like it's this change.

PR #7363, opened 2025-02-01
[Testing] Populate AppHost environment from launch profile and support launch profile override

### LINKED ISSUE #5093: Aspire.Hosting.Testing should populate DOTNET_LAUNCH_PROFILE
When `Aspire.Hosting.Testing`'s `DistributedApplicationTestingBuilder` creates and builds a DistributedApplicationBuider, it should pass in the `DOTNET_LAUNCH_PROFILE` configuration setting so AppHost code can switch off the launch profile, like documented in https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_launch_profile.

We use this in our samples here:

https://github.com/dotnet/aspire-samples/blob/08c14cdc112c74811c44976064b8d85e39e8bd93/samples/AspireWithNode/AspireWithNode.AppHost/Program.cs#L16-L20

But that check doesn't work when running in an Aspire.Hosting.Testing test because that configuration setting isn't being populated by the testing infrastructure.

cc @ReubenBond @mitchdenny @radical

ReubenBond:
## Description

Fixes #5093

This PR simulates launch profile loading for the testing host. It does not mutate the process' environment since that could cause issues with concurrently running tests. Instead, it propagates the environment variables specified in the AppHost's launch profile to the AppHost's configuration.

This PR sets the `DOTNET_LAUNCH_PROFILE` environment variable to the name of the first launch profile if not overridden. It does not consider the `DOTNET_LAUNCH_PROFILE` from the test runner's environment.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
+++ b/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
@@ -137,7 +137,7 @@ public class DistributedApplicationFactory(Type entryPoint, string[] args) : IDi
         OnBuilt(application);
     }
 
-    internal static void PreConfigureBuilderOptions(
+    private static void PreConfigureBuilderOptions(
         DistributedApplicationOptions applicationOptions,
         HostApplicationBuilderSettings hostBuilderOptions,
         string[] args,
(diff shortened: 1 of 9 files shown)


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -148,57 +148,125 @@ internal static void PreConfigureBuilderOptions(
...
+        if (hostBuilderOptions.Configuration is not null)
+        {
+            existingConfig.AddConfiguration(hostBuilderOptions.Configuration);
+        }
+
         var additionalConfig = new Dictionary<string, string?>();
         SetDefault("DcpPublisher:ContainerRuntimeInitializationTimeout", "00:00:30");
         SetDefault("DcpPublisher:RandomizePorts", "true");
         SetDefault("DcpPublisher:DeleteResourcesOnShutdown", "true");
         SetDefault("DcpPublisher:ResourceNameSuffix", $"{Random.Shared.Next():x}");
```
davidfowl, line 169: We should consider making some of these first class.

davidfowl: Yep, file an issue for this one.

davidfowl (approved):

PR #7381, opened 2025-02-03
Add docker info

davidfowl:
Dump docker info into the azdo pipeline


davidfowl:
@karolz-ms  Server Version: 20.10.24

@joperezr our docker versions are really old...

PR #7393, opened 2025-02-04
Fix EventHubs GetNamespaceFromSettings

### LINKED ISSUE #4886: Result from RandomNumberGenerator.GetHexString is discarded
This line of code here in Aspire.Azure.Messaging.EventHubs

https://github.com/dotnet/aspire/blob/681f2e754dc849f96abafcd2829783e69155abf1/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs#L64

Looks off because the return value is discarded. It generates a random hex string, then just throws it away.

I am not sure what the right thing to do with that is, otherwise I would offer a pull request.

Either something should be done with the result of `GetHexString`, or the whole `else` block can be removed entirely.

eerhardt:
## Description

We were calling RandomNumberGenerator.GetHexString and not using the result. The original intention was to use a random container name if we couldn't see ".servicebus" in the host name.

Instead of using a random name, just sanitize the host name and use it, so it is deterministc.

Fix #4886

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

cc @oising

### DIFF
--- a/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
+++ b/src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
@@ -1,7 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Security.Cryptography;
 using Aspire.Azure.Common;
 using Aspire.Azure.Messaging.EventHubs;
 using Azure.Core;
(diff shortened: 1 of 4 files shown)


davidfowl (approved):

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
```diff
@@ -56,15 +55,15 @@ protected static string GetNamespaceFromSettings(AzureMessagingEventHubsSettings
...
-                ns = ns[..ns.IndexOf(".servicebus")];
+                ns = ns[..serviceBusIndex];
             }
             else
             {
-                // Use a random prefix if no meaningful name is found e.g., "localhost", "127.0.0.1".
-                // This is used to create blob containers names that are unique in the referenced storage account.
-                RandomNumberGenerator.GetHexString(12, true);
+                // sanitize the namespace if it's not a servicebus namespace
+                ns = ns.Replace(".", "-");
```
sebastienros, line 66: Was wondering if it be based on an allow-list, assuming we only expect but this will be a domain here so in theory this should be fine.

sebastienros (approved):

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
oising: Huh. How did that get orphaned? Am I tagged because git blame tattled on me?

src/Components/Aspire.Azure.Messaging.EventHubs/EventProcessorClientComponent.cs
```diff
@@ -97,17 +97,14 @@ private static BlobContainerClient GetBlobContainerClient(
...
         bool shouldTryCreateIfNotExists = false;
 
+        // Do we have a container name provided in the settings?
         if (string.IsNullOrWhiteSpace(settings.BlobContainerName))
         {
+            // If not, we'll create a container name based on the namespace, event hub name and consumer group
             var ns = GetNamespaceFromSettings(settings);
 
-            // Do we have a container name provided in the settings?
-            if (string.IsNullOrWhiteSpace(settings.BlobContainerName))
```
oising, line 105: Oops.

src/Components/Aspire.Azure.Messaging.EventHubs/EventHubsComponent.cs
eerhardt: > How did that get orphaned?

Just a simple mistake (not yours 😉).

> Am I tagged because git blame tattled on me?

I just tagged as an FYI since you've been heavily involved in the EventHubs code.

PR #7396, opened 2025-02-04
Switch to strongly-type resource scope for existing resources

captainsafia:
Follow-up to #7249. Use strongly-typed resource scope for existing resources instead of Dictionary.

### DIFF
--- a/src/Aspire.Hosting.Azure/ResourceScope.cs
+++ b/src/Aspire.Hosting.Azure/ResourceScope.cs
@@ -1,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Azure;
+
+/// <summary>
+/// Represents the scope associated with the resource.
+/// </summary>
+/// <param name="resourceGroup">The name of the existing resource group.</param>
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting.Azure/ResourceScope.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.Azure;
+
+/// <summary>
+/// Represents the scope associated with the resource.
+/// </summary>
+/// <param name="resourceGroup">The name of the existing resource group.</param>
+public sealed class ResourceScope(object resourceGroup)
```
davidfowl, line 10:
```suggestion
public sealed class AzureResourceScope(object resourceGroup)
```

src/Aspire.Hosting.Azure/ResourceScope.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Represents the resource group to encode in the scope.
+    /// </summary>
+    public object ResourceGroup { get; } = resourceGroup;
```
davidfowl, line 15: Why is this an object?

davidfowl (approved):

PR #7409, opened 2025-02-04
Hide project arguments in dashboard, show secrets as passwords in the source column

### LINKED ISSUE #6459: In command-line args, detect secrets used in arguments and hide them

### Describe the bug

https://github.com/dotnet/aspire/pull/6212#issuecomment-2403886745

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

### LINKED ISSUE #6835: [WebToolsE2E][Aspire] The "Source" shows wrong suffix "dotnet run --no-build --project XXX Debug --no-launch-profile" in Aspire Dashboard after running the Aspire App in CLI
**REGRESSION INFO:** Work in Aspire 9.0.0

**INSTALL STEPS**
2. Install VS 17.12.3
3. Install Aspire 9.1.0-preview.1.24577.7
    - Apply dotnet9 feed

**REPRO STEPS**
1. Open CMD, create an Aspire Starter App using following commands:
```
dotnet new aspire-starter -o AspireStarterApp
cd AspireStarterApp/AspireStarterApp.AppHost
```
2. Dotnet run the project in CMD, and open URL in the browser.
`dotnet run`

**Note**
1. The issue doesn't repro when debugging an Aspire Starter project in VS

**ACTUAL**
The "Source" shows wrong suffix: `dotnet run --no-build --project C:\Users\v-ritagao\AspireStarter\AspireStarter.ApiService\AspireStarter.ApiService.csproj -c Debug --no-launch-profile`

Image

**EXPECTED**
Image

adamint:
## Description

Based on Karol and Drew's comments, I inverted the logic and added two properties, `resource.appArgs` and `resource.AppArgsParams`. The first contains launch args not added by DCP, and the second contains secrets that can be inserted into `resource.appArgs`. `resource.appArgs` will contain placeholders for all secrets. example:

The properties passed from AppHost to dashboard will be marked as sensitive if there is any secret in them. Processing in the dashboard turns placeholders into "passwords" in the source column - users can see all arguments by opening the text visualizer.

I extended the `ExpressionResolver` to return a type containing the resolved value as well as whether the expression contained a sensitive value, in the case of parameters appearing in the expressions.

Before:
image

After:
image

You can see that secrets do actually appear in the text visualizer
image

Fixes #6835 and fixes #6459

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

### DIFF
--- a/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
+++ b/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
@@ -33,8 +33,8 @@ public class ResourceSourceViewModel(string value, string? contentAfterValue, st
                 var argumentsString = string.Join(" ", launchArguments);
                 if (resource.TryGetAppArgsFormatParams(out var formatParams))
                 {
-                    var hiddenArgs = string.Format(CultureInfo.InvariantCulture, argumentsString, formatParams.Select(_ => "***"));
-                    var shownArgs = string.Format(CultureInfo.InvariantCulture, argumentsString, formatParams);
+                    var hiddenArgs = string.Format(CultureInfo.InvariantCulture, argumentsString, formatParams.Select(_ => "***").ToArray<object?>());
+                    var shownArgs = string.Format(CultureInfo.InvariantCulture, argumentsString, formatParams.ToArray<object?>());
                     commandLineInfo = (ArgumentsString: hiddenArgs, programPath is null ? argumentsString : $"{programPath} {shownArgs}");
                 }
                 else
--- a/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
+++ b/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
@@ -1,9 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Globalization;
-using System.Text.RegularExpressions;
-
 namespace Aspire.Dashboard.Model;
 
 public class ResourceSourceViewModel(string value, List<LaunchArgument>? contentAfterValue, string valueToVisualize, string tooltip)
@@ -28,13 +25,13 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
             else
             {
                 var argumentsString = string.Join(" ", launchArguments);
-                if (resource.TryGetAppArgsFormatParams(out var formatParams))
+                if (resource.TryGetAppArgsSensitivity(out var areArgumentsSensitive))
                 {
                     var arguments = launchArguments
-                        .Select(arg => new LaunchArgument(arg, !Regex.IsMatch(arg, "{\\d+}")))
+                        .Select((arg, i) => new LaunchArgument(arg, IsShown: !areArgumentsSensitive[i]))
                         .ToList();
-                    var launchArgsString = string.Format(CultureInfo.InvariantCulture, argumentsString, [.. formatParams]);
-                    commandLineInfo = (Arguments: arguments, launchArgsString);
+
+                    commandLineInfo = (Arguments: arguments, argumentsString);
                 }
                 else
                 {
--- a/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
+++ b/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Globalization;
+
 namespace Aspire.Dashboard.Model;
 
 public class ResourceSourceViewModel(string value, string? contentAfterValue, string valueToVisualize, string tooltip)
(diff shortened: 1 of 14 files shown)


src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -18,6 +18,12 @@ public class ResourceSourceViewModel(string value, string? contentAfterValue, st
 
         if (resource.TryGetExecutableArguments(out var arguments))
         {
+            if (resource.TryGetProjectArguments(out var projectArgs))
+            {
+                var foundHostArgCount = arguments.TakeWhile((arg, i) => i < projectArgs.Length && string.Equals(arg, projectArgs[i], StringComparison.Ordinal)).Count();
```
drewnoakes, line 23: If I understand this correctly, the annotation in `DcpExecutor` has new code to add the existing args to `project.args`, only for them to be carefully removed here. Why not simply add only the args you want to show to the annotation?

At the least, please document what's going on here, as it seems like the logic in the dashboard is coupled to some behaviour in the app host, and it's not obvious why that is.

karolz-ms: +1 on @drewnoakes suggestion to "invert" the logic and put into the annotation only the arguments that should show up in the dashboard.

Other than this, I see no issues with this change

adamint: Thank you both! I've updated this to put an annotation on the dcp resource of the launch arguments. I also included information on whether they contained secrets

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -1,47 +1,79 @@
...
+            {
+                commandLineInfo = (null, null);
+            }
+            else
+            {
+                var argumentsString = string.Join(" ", launchArguments);
+                if (resource.TryGetAppArgsFormatParams(out var formatParams))
+                {
+                    var arguments = launchArguments
+                        .Select(arg => new LaunchArgument(arg, !Regex.IsMatch(arg, "{\\d+}")))
```
karolz-ms, line 34: So I know this is not super likely to happen... but what if the program invocation parameter needs to be, literally, something like `{12}`? Isn't this going to be a problem here?

What if we did the `(sensitive value) --> placeholder` substitution before we write the annotation onto the DCP object? Then the view model could read the annotation directly with minimal processing.

karolz-ms: If we need both the "masked" as well as "fully resolved" arguments (depending on the view selected by the user in the dashboard), we could store two versions of the argument list/string in the annotation.

adamint: > what if the program invocation parameter needs to be, literally, something like {12}? Isn't this going to be a problem here

Yeah, that's a good point. I decided instead to avoid the format string entirely, and use an array of booleans representing sensitivity (seemingly protobuf won't serialize an immutablearray<bool>, so it's immutablearray<int>)

`resource.appArgs`: `one two three SECRET five`
`resource.appArgsSensitivity` `[false, false, false, true, false]`

> What if we did the (sensitive value) --> placeholder substitution before we write the annotation onto the DCP object? Then the view model could read the annotation directly with minimal processing.

With the change above, the substitution is no longer required :)

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -77,24 +77,41 @@ await EvalEndpointAsync(endpointReference, EndpointProperty.Port).ConfigureAwait
...
     {
         var value = await vp.GetValueAsync(cancellationToken).ConfigureAwait(false);
 
+        if (vp is ParameterResource pr)
+        {
+            return new ResolvedValue(value, pr.Secret);
+        }
+
+        // No need to do extra work, since the below will only be valid for containers.
+        if (!sourceIsContainer)
```
karolz-ms, line 110: The `sourceIsContainer` parameter complicates the `ResolveInternalAsync` logic below. Could this parameter be eliminated? What do we gain by having it? For example, is value provider (`vp`) going to ever be `HostUrl` if we are not looking at a container?

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -128,30 +145,32 @@ await EvalEndpointAsync(endpointReference, EndpointProperty.Port).ConfigureAwait
...
-        return value switch
+        return (value, sourceIsContainer) switch
         {
-            ConnectionStringReference cs => await ResolveInternalAsync(cs.Resource.ConnectionStringExpression).ConfigureAwait(false),
-            IResourceWithConnectionString cs => await ResolveInternalAsync(cs.ConnectionStringExpression).ConfigureAwait(false),
-            ReferenceExpression ex => await EvalExpressionAsync(ex).ConfigureAwait(false),
-            EndpointReference endpointReference => await EvalEndpointAsync(endpointReference, EndpointProperty.Url).ConfigureAwait(false),
-            EndpointReferenceExpression ep => await EvalEndpointAsync(ep.Endpoint, ep.Property).ConfigureAwait(false),
-            IValueProvider vp => await EvalValueProvider(vp).ConfigureAwait(false),
+            (ConnectionStringReference cs, true) => await ResolveInternalAsync(cs.Resource.ConnectionStringExpression).ConfigureAwait(false),
```
karolz-ms, line 159: I am afraid that after this change the code is not covering all the cases that the previous implementation did. For example, in the previous implementation the `ConnectionStringReference` was matching unconditionally. In the new implementation it only matches for containers.

karolz-ms: Also see my comment about introducing `sourceIsContainer` parameter above

src/Aspire.Hosting/Dcp/ResourceSnapshotBuilder.cs
```diff
@@ -147,6 +158,31 @@ public CustomResourceSnapshot ToSnapshot(Executable executable, CustomResourceSn
...
+        var anySensitive = false;
+        foreach (var annotation in launchArgumentAnnotations)
+        {
+            if (annotation.IsSensitive)
+            {
+                anySensitive = true;
+            }
+
+            launchArgsBuilder.Add(annotation.Argument);
+            argsAreSensitiveBuilder.Add(Convert.ToInt32(annotation.IsSensitive));
```
karolz-ms, line 180: or just `annotation.IsSensitive ? 1 : 0`

src/Aspire.Hosting/Dcp/ResourceSnapshotBuilder.cs
```diff
@@ -147,6 +158,31 @@ public CustomResourceSnapshot ToSnapshot(Executable executable, CustomResourceSn
         };
     }
 
+    private static (ImmutableArray<string> Args, ImmutableArray<int>? ArgsAreSensitive, bool IsSensitive)? GetLaunchArgs(CustomResource resource)
```
karolz-ms, line 161: Is there a particular reason why the sensitivity flags are an `int` array and not `bool` array?

karolz-ms (commented):
I like most of it, but still have some questions, HTH

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
karolz-ms: LGTM on that front

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
karolz-ms: Nice

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
adamint: > ConnectionStringReference was matching unconditionally. In the new implementation it only matches for containers

Previously, that code path would also never be run for non-containers. If you see the old version of ExpressionResolver#166 (the only entrypoint into the resolver), non-container cases only resolve the value from the value provider
image

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
adamint: It seems like yes, `WithOtlpExporter`  -> `AddOtlpEnvironment` will add a HostUrl, which can be resolved even in executables. 

The goal of the parameter was to avoid the value substitution on executables, to preserve the previous behavior

src/Aspire.Hosting/Dcp/ResourceSnapshotBuilder.cs
adamint: Not sure if it's the protobuf library or gRPC, but an ImmutableArray<bool> was not able to be serialized/deserialized, and instead was being treated as a string. using int was a workaround

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
karolz-ms: Got it. Thanks for clarification

karolz-ms (approved):

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -3,45 +3,74 @@
...
     public string Value { get; } = value;
-    public string? ContentAfterValue { get; } = contentAfterValue;
+    public List<LaunchArgument>? ContentAfterValue { get; } = contentAfterValue;
     public string ValueToVisualize { get; } = valueToVisualize;
     public string Tooltip { get; } = tooltip;
 
     internal static ResourceSourceViewModel? GetSourceViewModel(ResourceViewModel resource)
     {
-        var executablePath = resource.TryGetExecutablePath(out var path) ? path : null;
+        (List<LaunchArgument>? Arguments, string? ArgumentsString)? commandLineInfo;
```
drewnoakes, line 15: Should this be non-nullable? I can't see where it's ever assigned to null.

```suggestion
        (List<LaunchArgument>? Arguments, string? ArgumentsString) commandLineInfo;
```

Some of the code below does actually check for null, but I don't think there's a problem.

adamint:
Unrelated tests are failing

PR #7415, opened 2025-02-05
Make CosmosDB resource names consistent.

eerhardt:
Also move the new resources into the same namespace as the emulator. Unfortunately the top level CosmosDBResource is in the top-level Aspire.Hosting namespace.

### DIFF
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -3,7 +3,7 @@
 
 using Aspire.Hosting.ApplicationModel;
 
-namespace Aspire.Hosting.Azure.CosmosDB;
+namespace Aspire.Hosting.Azure;
 
 /// <summary>
 /// Represents an Azure Cosmos DB Database Container.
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
```diff
@@ -3,20 +3,20 @@
 
 using Aspire.Hosting.ApplicationModel;
 
-namespace Aspire.Hosting.Azure.CosmosDB;
+namespace Aspire.Hosting.Azure;
```
eerhardt, line 6: Anyone have opinions on this? `AzureCosmosDBResource` is in `Aspire.Hosting`. The emulator is in `Aspire.Hosting.Azure`.

Everything else in Azure is under `ASpire.Hosting.Azure`. For example, `AzureStorageResource` under `Aspire.Hosting.Azure`. `AzureCosmosDBResource` is the only one that got stuck in the top-level namespace.

davidfowl (approved):

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
davidfowl: This looks fine.

Issue #7407, opened 2025-02-04
EventHubs, ServiceBus, and CosmosDB Hosting integrations should create Resources for children


eerhardt:
With:

* [Add support for managed identity in Azure Cosmos DB hosting component (dotnet/aspire#7092)](https://github.com/dotnet/aspire/pull/7092)
* [Add custom application model for EventHubs emulator (dotnet/aspire#7005)](https://github.com/dotnet/aspire/pull/7005/files#diff-a635266e9f2e54dafd297b24b2e099e0c442140c7080ef4338d9f07b5dd903b1)
* [Add Service Bus emulator support (dotnet/aspire#6737)](https://github.com/dotnet/aspire/pull/6737)

We added Hosting app models for CosmosDB Databases, EventHub Hubs, and ServiceBus Topics and Queues.

We didn't make these app models derive from "Resource", which means they lose out on things like:

* being referenceable
* being able to "wait for"

For example, with EventHubs [Allow specifying an Event Hub to use as the default EntityPath for namespace connection string](https://github.com/dotnet/aspire/pull/7105), we are having to work around these limitations. But if EventHub "Hub" was a resource, we could make an app `WithReference` on the EventHub "Hub". See https://github.com/dotnet/aspire/pull/7105#discussion_r1934206406 for more information.

## Proposal

We should make these child types inherit from "Resource" in the Aspire Hosting integrations. Redoing what we did in the above 9.1 PRs before we ship 9.1 publicly.

The biggest problem is that we already shipped `AddDatabase`, `AddHub`, `AddQueue` APIs in 9.0, but these APIs return the wrong type. They return the top-level parent Resource when they need to be returning the new child Resource. We can't break these APIs in 9.1, so we will need another way to solve this. In the PRs above, we created `WithDatabase`, `WithHub`, etc. but that only worked because the children aren't "Resource"s. Going forward, we want `Add` APIs that return the new child.

To work around this issue, we will create new `Add` methods that have a different name. We will remove the newly added `With` methods and still Obsolete the existing Add methods that were shipped in 9.0.

I can think of 2 patterns for these new `Add` methods:

1. `AddCosmosDatabase` - Add a redundant parent name into the method.
2. `AddDatabaseResource` - append `Resource` to differentiate

[...]

PR #7422, opened 2025-02-05
Allow CosmosDB connection strings to work with Database and Container keys

eerhardt:
## Description

With #7408, Cosmos Database and Containers are Aspire Resources now, which means they can be referenced via WithReference.

In the future, the Cosmos Host integration will append these values to the connection string. This will allow the database and container names to be specified outside of the application. (Today they are hard-coded in both, or specified separately in config.) We can also add new client APIs to register a Database or Container object in DI.

This is a stop-gap to support these values in .NET Aspire 9.1 client integration so it doesn't break applications when we start adding these values in the future.

Contributes to #7407

## Checklist

- Is this feature complete?
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Shared/Cosmos/CosmosUtils.cs
+++ b/src/Shared/Cosmos/CosmosUtils.cs
@@ -47,7 +47,8 @@ internal static class CosmosUtils
             ConnectionString = connectionString
         };
 
-        // Strip out the database and container from the connection string to
+        // Strip out the database and container from the connection string in order
+        // to tell if we are left with just AccountEndpoint.
         if (connectionBuilder.TryGetValue("Database", out var _))
         {
             connectionBuilder["Database"] = null;
--- a/src/Shared/Cosmos/CosmosUtils.cs
+++ b/src/Shared/Cosmos/CosmosUtils.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Data.Common;
+using System.Diagnostics;
 
 namespace Aspire.Hosting.Azure.CosmosDB;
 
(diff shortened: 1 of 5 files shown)


src/Shared/Cosmos/CosmosUtils.cs
```diff
@@ -23,4 +24,63 @@ internal static bool IsEmulatorConnectionString(string? connectionString)
...
+        {
+            return new CosmosConnectionInfo(accountEndpoint, null);
+        }
+
+        var connectionBuilder = new DbConnectionStringBuilder()
+        {
+            ConnectionString = connectionString
+        };
+
+        // Strip out the database and container from the connection string to
```
eerhardt, line 50:
```suggestion
        // Strip out the database and container from the connection string in order
        // to tell if we are left with just AccountEndpoint.
```

src/Shared/Cosmos/CosmosUtils.cs
```diff
@@ -23,4 +24,64 @@ internal static bool IsEmulatorConnectionString(string? connectionString)
...
+    /// </remarks>
+    internal static CosmosConnectionInfo ParseConnectionString(string connectionString)
+    {
+        Uri? accountEndpoint;
+        if (Uri.TryCreate(connectionString, UriKind.Absolute, out accountEndpoint))
+        {
+            return new CosmosConnectionInfo(accountEndpoint, null);
+        }
+
+        var connectionBuilder = new DbConnectionStringBuilder()
```
davidfowl, line 45: future: If we're going all in on connection strings, it might be worth shipping a public API that uses `DbConnectionStringBuilder` internally.

davidfowl (approved):

src/Shared/Cosmos/CosmosUtils.cs
davidfowl: e.g. https://github.com/dotnet/aspire/issues/7427

davidfowl: nit: `CosmosConnectionInfo` and `CosmosConnectionInfo.ParseConnectionString`

eerhardt: Agreed. Sorry for the confusion, I was using the existing internal names above. Not what I would use for public API.

PR #7424, opened 2025-02-06
Add existing resource support to more Azure resources

captainsafia:
This PR adds support for the existing Azure resource feature introduced in #7249 to more resources.

We introduce a `CreateExistingOrNewProvisionableResource` existing method that encapsulates the logic for wiring up an existing or new resource definition based on the existence of the annotation.

Note: SQL Server and Azure Redis will be done in a separate PR since they have some unique interactions with role assignments/access perms.

### DIFF
--- a/src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
@@ -5,6 +5,7 @@ using System.Diagnostics.CodeAnalysis;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Azure;
 using Azure.Provisioning;
+using Azure.Provisioning.Primitives;
 
 namespace Aspire.Hosting;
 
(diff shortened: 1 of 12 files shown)


src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
```diff
@@ -183,6 +184,40 @@ public static ProvisioningParameter AsProvisioningParameter(this ReferenceExpres
...
+    /// <summary>
+    /// Encapsulates the logic for creating an existing or new <see cref="ProvisionableResource"/>
+    /// based on whether or not the <see cref="ExistingAzureResourceAnnotation" /> exists on the resource.
+    /// </summary>
+    /// <typeparam name="T">The type of <see cref="ProvisionableResource"/> to produce.</typeparam>
+    /// <param name="infrastructure">The <see cref="AzureResourceInfrastructure"/> that will contain the <see cref="ProvisionableResource"/>.</param>
+    /// <param name="createExisting">A callback to create the existing resource.</param>
+    /// <param name="createNew">A callback to create the new resource.</param>
+    /// <returns>The provisioned resource.</returns>
+    public static T CreateExistingOrNewProvisionableResource<T>(this AzureResourceInfrastructure infrastructure, Func<string, ProvisioningParameter, T> createExisting, Func<AzureResourceInfrastructure, T> createNew)
```
davidfowl, line 196: I'm struggle a bit with this being a public API but it's not the end of the world I guess...

davidfowl (approved):

src/Aspire.Hosting.Azure/AzureProvisioningResourceExtensions.cs
eerhardt: What if we didn't make it an extension method? Then it wouldn't show up for normal developers, only for devs that were looking for it (which probably will only ever be us).

captainsafia: I considered internal and IVT for this but that felt a little bi too much.

I do like the idea of not making it an extension method to hide it a bit better.

PR #7434, opened 2025-02-06
Support string or ParameterResource arguments for AsExisting APIs

captainsafia:
Follow-up to #7249. Expand support for strings and parameter resoruces as argument types.

### DIFF
--- a/src/Aspire.Hosting.Azure/ExistingAzureResourceAnnotation.cs
+++ b/src/Aspire.Hosting.Azure/ExistingAzureResourceAnnotation.cs
@@ -9,16 +9,21 @@ namespace Aspire.Hosting.Azure;
 /// Represents a resource that is not managed by Aspire's provisioning or
 /// container management layer.
 /// </summary>
-public sealed class ExistingAzureResourceAnnotation(ParameterResource nameParameter,
-    ParameterResource? resourceGroupParameter = null) : IResourceAnnotation
+public sealed class ExistingAzureResourceAnnotation(object name, object? resourceGroup = null) : IResourceAnnotation
 {
     /// <summary>
     /// Gets the name of the existing resource.
     /// </summary>
-    public ParameterResource NameParameter { get; } = nameParameter;
+    /// <remarks>
+    /// Supports a <see cref="string"/> or a <see cref="ParameterResource"/> via runtime validation.
+    /// </remarks>
+    public object Name { get; } = name;
 
     /// <summary>
     /// Gets the name of the existing resource group. If <see langword="null"/>, use the current resource group.
     /// </summary>
-    public ParameterResource? ResourceGroupParameter { get; } = resourceGroupParameter;
+    /// <remarks>
+    /// Supports a <see cref="string"/> or a <see cref="ParameterResource"/> via runtime validation.
+    /// </remarks>
+    public object? ResourceGroup { get; } = resourceGroup;
 }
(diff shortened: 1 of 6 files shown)


src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -28,40 +28,57 @@ public static bool IsExisting(this IResource resource)
     /// </summary>
     /// <typeparam name="T">The type of the resource.</typeparam>
     /// <param name="builder">The resource builder.</param>
-    /// <param name="nameParameter">The name of the existing resource.</param>
-    /// <param name="resourceGroupParameter">The name of the existing resource group, or <see langword="null"/> to use the current resource group.</param>
+    /// <param name="name">The name of the existing resource.</param>
+    /// <param name="resourceGroup">The name of the existing resource group, or <see langword="null"/> to use the current resource group.</param>
     /// <returns>The resource builder with the existing resource annotation added.</returns>
-    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, object name, object? resourceGroup = null)
```
davidfowl, line 34: well keeping the strong type here is good

src/Aspire.Hosting.Azure/ExistingAzureResourceAnnotation.cs
```diff
@@ -9,16 +9,21 @@ namespace Aspire.Hosting.Azure;
...
+public sealed class ExistingAzureResourceAnnotation(object name, object? resourceGroup = null) : IResourceAnnotation
 {
     /// <summary>
     /// Gets the name of the existing resource.
     /// </summary>
-    public ParameterResource NameParameter { get; } = nameParameter;
+    /// <remarks>
+    /// Supports a <see cref="string"/> or a <see cref="ParameterResource"/> via runtime validation.
+    /// </remarks>
+    public object Name { get; } = name;
```
eerhardt, line 20: Can we instead just have the new overload call the `AddParameter` method that takes a hard-coded string value?

Then everything is built on Parameters underneath, but the API can take `string` or `Parameter`.

eerhardt: https://github.com/dotnet/aspire/blob/3c823d994e1f4d16fef1ede4b7e3fc430ea7e00a/src/Aspire.Hosting/ParameterResourceBuilderExtensions.cs#L50-L57

is the overload I'm talking about.

captainsafia: We'd need to figure out if the `Name` pointed to an _actual_ `ParameterResource` that needed to be mapped to a provisioning parameter in the Bicep/manifest or a faux one where we could just extract the value. Is there a way to do that with `AddParameter`?

eerhardt: Would it be terrible if it always created the parameter underneath (in the manifest and bicep) and the parameter's value was hard-coded in the manifest?

captainsafia: `azd` has some special behavior that will provide an enhanced prompt for selecting the resource group when we detect it as a parameter in the resource group. I suspect that mapping it this way will cause that logic to trigger when it shouldn't?

davidfowl: I don't think that's the end of the world, but it's no different from what we do with environment variables, arguments, bicep parameters etc.

davidfowl (approved):

Issue #6457, opened 2024-10-23
Provide a way to specify the relative URL for the resource (to be used by the dashboard)


raffaeler:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

I am using Aspire 9 RC2.
The dashboard always shows the root address like 'https://localhost:4321/'.
Anyway, the `http`/`https` entry-points for the resource may be different than the root.
While I can change it in my own projects, this is not possible for standard container images.

### Describe the solution you'd like

When I use `AddContainer` to add a local instance of the Qdrant container, I want the dashboard showing the link for the dashboard which is `http://localhost:6333/dashboard` instead of just `http://localhost:6333/` as I obtain now with `WithEndpoint`.

Example:
```csharp
var vectordb = builder.AddContainer("vectorDB", "qdrant/qdrant")
    .WithEndpoint(6333, 6333, "http", "dashboard")
    .WithEndpoint(6334, 6334)
    .WithBindMount(...);
```

![Image](https://github.com/user-attachments/assets/f61d4c8e-d7ce-454d-938f-b79456d4e900)

adamint:
What shows up under +1? @raffaeler

raffaeler:
Just the other endpoint that I added  to make grpc available (BTW I marked it as "tcp")

davidfowl:
Related to https://github.com/dotnet/aspire/issues/5209

leslierichardson95:
Using 9.1 to at least start the work needed to address this issue.

davidfowl:
YESS!

davidfowl:
cc @DamianEdwards related

rickykaare:
I really like what you guys are building with Aspire, and we are using it a lot for our local environment.

Having links directly to the CosmosDB emulator and the Scalar UI of our containerized services, is very high on our wish list.

For now we've found the following work-around helpful, and perhaps it can be helpful to others too:
https://gist.github.com/rickykaare/70c5fd8d626802c3b5abedad5f4e982b

Example:
```csharp
var cosmosdb = builder
    .AddAzureCosmosDB("cosmosdb")
    .WithEndpointPath("emulator", "/_explorer/index.html")
    .RunAsEmulator(c => c
        .WithLifetime(ContainerLifetime.Persistent)
        .WithDataVolume());
```

Be aware that this was build as a work-around for local development only.

PR #7442, opened 2025-02-06
Add display name and priority support to endpoints

### LINKED ISSUE #7104: Add support for "Display Name" for endpoints in the Dashboard
With [Add support for CosmosDB vnext-preview emulator (dotnet/aspire#7048)](https://github.com/dotnet/aspire/pull/7048), the Cosmos DB emulator resource has 2 endpoints:

* The data "API" endpoint
* A data explorer that allows developers to see what is stored in the emulator

The problem is you don't know which link is which:

Image

Instead of displaying just a URL, we should have support for giving a display name to an endpoint, and showing the display name in the Dashboard.

adamint:
## Description

Adds two optional properties to `Url` and `EndpointAnnotation` - namely, DisplayName and Priority, that are used in the dashboard to determine 1) display order, and 2) display name. 

Links still show up on the dashboard, just with the text replaced with the display name if set.
image

Additionally, adds a third **Display name** column to the dashboard that shows the display name of the endpoint if it is set.
(no display names)
image

(pgAdmin has a display name)
image

Also sets the endpoint display name on the CosmosDB data explorer to "Data Explorer".

Fixes #7104 

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No **I think?**


davidfowl:
I'm trying to figure out if we should split the visual impact of the endpoint vs the endpoint definition. They should be different annotations.

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -31,7 +31,9 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     /// <param name="targetPort">This is the port the resource is listening on. If the endpoint is used for the container, it is the container port.</param>
     /// <param name="isExternal">Indicates that this endpoint should be exposed externally at publish time.</param>
     /// <param name="isProxied">Specifies if the endpoint will be proxied by DCP. Defaults to true.</param>
-    public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, string? transport = null, [EndpointName] string? name = null, int? port = null, int? targetPort = null, bool? isExternal = null, bool isProxied = true)
+    /// <param name="displayName">Display name of the endpoint, to be displayed in the Aspire Dashboard.</param>
+    /// <param name="priority">Integer to control visual ordering of endpoints in the Aspire Dashboard.</param>
```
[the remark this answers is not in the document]
adamint: It's documented elsewhere, but apparently not here. I'll also add info on ties. Sort order could be used instead!

adamint:
> I'm trying to figure out if we should split the visual impact of the endpoint vs the endpoint definition. They should be different annotations.

Why should they? If display name is a friendly name for the endpoint, that seems like an integral part of the definition itself

davidfowl:
This data doesn't go into the manifest right? I am thinking that there's a core annotation that is used for networking and another part of it that is used for display in the dashboard.

davidfowl:
Thinking out loud, maybe it's fine to create a nested display properties on `EndpointAnnotion`  where we store these properties. I don't think we add to the chaos that is part of these methods an maybe we treat these like properties you set with 

```C#
WithEndpoint("name", e => e.DisplayName = "")
```

JamesNK:
What do other people think about having the grid in details showing a name and display name column? It feels like we lose out on the simplicit of a name/value grid.

image

---

Instead, what about showing the display name as the link text in the value column? The visualizer would open to show the underlying address. That makes the details consistent with what is shown in the grid (which shows the link using the display name), lets endpoints grid still be name/value, and it makes the code simpler.

On the other hand, it makes it more difficult to see the address port in details at a glance, which people might want.

Thoughts?

davidfowl:
Not a fan of the 2 columns

adamint:
> What do other people think about having the grid in details showing a name and display name column? It feels like we lose out on the simplicit of a name/value grid.
> 
> image
> 
> Instead, what about showing the display name as the link text in the value column? The visualizer would open to show the underlying address. That makes the details consistent with what is shown in the grid (which shows the link using the display name), lets endpoints grid still be name/value, and it makes the code simpler.
> 
> On the other hand, it makes it more difficult to see the address port in details at a glance, which people might want.
> 
> Thoughts?

Yeah, there's a tension between showing all the data and adding more visual complexity. 

> On the other hand, it makes it more difficult to see the address port in details at a glance, which people might want.

was the reason I initially added the third column - I'd prefer to be able to see the endpoint URL without needing to hover over the value, but don't have a strong opinion here

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
```diff
@@ -36,7 +39,8 @@ public static List<DisplayedEndpoint> GetEndpoints(ResourceViewModel resource, b
         // - other urls
         // - endpoint name
         var orderedEndpoints = endpoints
-            .OrderByDescending(e => e.Url?.StartsWith("https") == true)
+            .OrderByDescending(e => e.SortOrder ?? 0)
```
JamesNK, line 42: Unit test this works as expected. There should be a test that combines all sort order possibilities: 0, null, positive number, negative number.

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -131,6 +131,15 @@ message Url {
     // Determines if this url shows up in the details view only by default.
     // If true, the url will not be shown in the list of urls in the top level resources view.
     bool is_internal = 3;
+    // Display properties of the Url
+    optional EndpointDisplayProperties display_properties = 4;
+}
+
+message EndpointDisplayProperties {
```
JamesNK, line 138: Rename to `UrlDisplayProperties`. The message this is used on is called `Url` so it should follow its naming standard.

JamesNK:
Add a display name to some endpoints in TestShop. I'll pull the branch and try it out tomorrow.

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
```diff
@@ -31,7 +31,8 @@ public sealed class EndpointAnnotation : IResourceAnnotation
     /// <param name="targetPort">This is the port the resource is listening on. If the endpoint is used for the container, it is the container port.</param>
     /// <param name="isExternal">Indicates that this endpoint should be exposed externally at publish time.</param>
     /// <param name="isProxied">Specifies if the endpoint will be proxied by DCP. Defaults to true.</param>
-    public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, string? transport = null, [EndpointName] string? name = null, int? port = null, int? targetPort = null, bool? isExternal = null, bool isProxied = true)
+    /// <param name="displayProperties">Function to configure Aspire Dashboard display properties for this endpoint.</param>
+    public EndpointAnnotation(ProtocolType protocol, string? uriScheme = null, string? transport = null, [EndpointName] string? name = null, int? port = null, int? targetPort = null, bool? isExternal = null, bool isProxied = true, Action<EndpointDisplayProperties>? displayProperties = null)
```
davidfowl, line 35: undo this Its a breaking change.

src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
```diff
@@ -101,6 +101,11 @@ public EndpointReferenceExpression Property(EndpointProperty property)
     /// </summary>
     public string Url => AllocatedEndpoint.UriString;
 
+    /// <summary>
+    /// Gets the display properties for this endpoint.
+    /// </summary>
+    public EndpointDisplayProperties DisplayProperties => EndpointAnnotation.DisplayProperties;
```
davidfowl, line 107: Who uses this?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -14,6 +14,7 @@
 using Azure.Provisioning.KeyVault;
 using Microsoft.Azure.Cosmos;
 using Microsoft.Extensions.DependencyInjection;
+using EndpointDisplayProperties = Aspire.Hosting.ApplicationModel.EndpointDisplayProperties;
```
eerhardt, line 17: Why is this necessary?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -533,10 +533,11 @@ public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> build
...
+    /// <param name="displayProperties">Function to configure Aspire Dashboard display properties for this endpoint.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
     /// <exception cref="DistributedApplicationException">Throws an exception if an endpoint with the same name already exists on the specified resource.</exception>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>")]
-    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, [EndpointName] string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null) where T : IResourceWithEndpoints
+    public static IResourceBuilder<T> WithEndpoint<T>(this IResourceBuilder<T> builder, int? port = null, int? targetPort = null, string? scheme = null, [EndpointName] string? name = null, string? env = null, bool isProxied = true, bool? isExternal = null, Action<EndpointDisplayProperties>? displayProperties = null) where T : IResourceWithEndpoints
```
davidfowl, line 540: Undo these changes, they're all breaking. There's already an api that takes a callback

src/Aspire.Hosting/ApplicationModel/EndpointAnnotation.cs
eerhardt: There's no real need to set this in the constructor. Callers can just update the property after construction.

src/Aspire.Hosting/ApplicationModel/EndpointReference.cs
adamint: ResourceSnapshotBuilder, #239, #246, #253

src/Aspire.Hosting/ResourceBuilderExtensions.cs
adamint: done

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
adamint: It's not, removed. Thanks

src/Aspire.Hosting/ApplicationModel/CustomResourceSnapshot.cs
```diff
@@ -177,8 +177,16 @@ public sealed record EnvironmentVariableSnapshot(string Name, string? Value, boo
 /// <param name="Name">Name of the url.</param>
 /// <param name="Url">The full uri.</param>
 /// <param name="IsInternal">Determines if this url is internal.</param>
+/// <param name="DisplayProperties">The UI display properties for the url.</param>
 [DebuggerDisplay("{Url}", Name = "{Name}")]
-public sealed record UrlSnapshot(string Name, string Url, bool IsInternal);
+public sealed record UrlSnapshot(string Name, string Url, bool IsInternal, UrlDisplayPropertiesSnapshot? DisplayProperties = null);
```
davidfowl, line 182: Make UrlDisplayPropertiesSnapshot a property. This is a breaking change.

davidfowl (approved):
Address the breaking change and this is good to go!

playground/TestShop/TestShop.AppHost/ResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting;
+
+public static class ResourceBuilderExtensions
+{
+    public static IResourceBuilder<T> WithModifiedEndpoints<T>(this IResourceBuilder<T> builder, Action<EndpointAnnotation> callback) where T : IResourceWithEndpoints
```
davidfowl, line 8: Do you need this?

playground/TestShop/TestShop.AppHost/Program.cs
```diff
@@ -38,6 +38,10 @@
 
 builder.AddProject<Projects.MyFrontend>("frontend")
        .WithExternalHttpEndpoints()
+       .WithModifiedEndpoints(c =>
+       {
+           c.DisplayProperties.DisplayName = "TestShop frontend access";
+       })
```
JamesNK, line 44: This seems like a bit of a sledge hammer to give names to endpoints from the launch profile.

@davidfowl Do you want to be able to give names to individual endpoints from the launch profile?

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -63,14 +63,50 @@
...
+                                ShowHover="true">
+                    <AspireTemplateColumn Sortable="true" SortBy="@(GridSort<DisplayedEndpoint>.ByAscending(i => i.Name))" Title="@ControlStringsLoc[nameof(ControlsStrings.NameColumnHeader)]">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.NameColumnHeader)]"
+                                   Value="@context.Name"
+                                   EnableHighlighting="@(!string.IsNullOrEmpty(_filter))"
+                                   HighlightText="@_filter" />
+                    </AspireTemplateColumn>
+                    @if (hasAnyEndpointDisplayNames)
+                    {
+                        <AspireTemplateColumn Sortable="true" SortBy="@(GridSort<DisplayedEndpoint>.ByAscending(i => i.Name))" Title="@ControlStringsLoc[nameof(ControlsStrings.DisplayNameColumnHeader)]">
```
JamesNK, line 91: Is this a typo?
```suggestion
                        <AspireTemplateColumn Sortable="true" SortBy="@(GridSort<DisplayedEndpoint>.ByAscending(i => i.DisplayName))" Title="@ControlStringsLoc[nameof(ControlsStrings.DisplayNameColumnHeader)]">
```

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -63,14 +63,50 @@
...
+                            @if (context.DisplayName is not null)
```
JamesNK, line 92: Why not have a GridValue if there isn't a display name? The column should display `-` like other empty grid items. I believe GridValue handles doing that for you.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -63,14 +63,50 @@
...
+                            {
+                                <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.DisplayNameColumnHeader)]"
+                                           Value="@context.DisplayName"
+                                           EnableHighlighting="@(!string.IsNullOrEmpty(_filter))"
+                                           HighlightText="@_filter" />
+                            }
+                        </AspireTemplateColumn>
+                    }
+                    <AspireTemplateColumn Sortable="true" SortBy="@_endpointValueSort" Title="@ControlStringsLoc[nameof(ControlsStrings.PropertyGridValueColumnHeader)]">
+                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.NameColumnHeader)]"
```
JamesNK, line 102:
```suggestion
                        <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.PropertyGridValueColumnHeader)]"
```

playground/TestShop/TestShop.AppHost/Program.cs
davidfowl: I think so yes.

src/Aspire.Dashboard/ResourceService/Partials.cs
```diff
@@ -93,7 +93,7 @@ ImmutableArray<UrlViewModel> GetUrls()
             return (from u in Urls
                     let parsedUri = Uri.TryCreate(u.FullUrl, UriKind.Absolute, out var uri) ? uri : null
                     where parsedUri != null
-                    select new UrlViewModel(u.Name, parsedUri, u.IsInternal))
+                    select new UrlViewModel(u.Name, parsedUri, u.IsInternal, u.DisplayProperties is not null ? new UrlDisplayPropertiesViewModel(u.DisplayProperties.HasDisplayName ? u.DisplayProperties.DisplayName : null, u.DisplayProperties.HasSortOrder ? u.DisplayProperties.SortOrder : null) : null))
```
JamesNK, line 96: I think we can simplify things here. A UrlViewModel should always have UrlDisplayPropertiesViewModel. Create one from the values available.

Something like:
```suggestion
                    select new UrlViewModel(u.Name, parsedUri, u.IsInternal, new UrlDisplayPropertiesViewModel((u.DisplayProperties?.HasDisplayName ?? false) ? u.DisplayProperties.DisplayName : null, (u.DisplayProperties?.HasSortOrder ?? false) ? u.DisplayProperties.SortOrder : null)))
```

Then remove logic that worries about null display properties VM from other code.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
```diff
@@ -324,18 +324,22 @@ public sealed class UrlViewModel
...
         ArgumentNullException.ThrowIfNull(url);
 
         Name = name;
         Url = url;
         IsInternal = isInternal;
+        DisplayProperties = displayProperties;
     }
 }
 
+public record UrlDisplayPropertiesViewModel(string? DisplayName, int? SortOrder);
```
JamesNK, line 341: What happens if displayname is an empty string? I think in that case we treat it the same as null (so ignore the display name and just use the URL).

src/Aspire.Dashboard/ResourceService/Partials.cs
JamesNK: I think it would be simpler to not have the concept of null at all in the view model.
* Empty string display name = no display name
* 0 and null sort orders are already equivelent so just have sort order of `int` and default to 0.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
JamesNK: https://github.com/dotnet/aspire/pull/7442#discussion_r1947333828

davidfowl:
I'd also want to use this in the `BicepProvisioner` to set the display for the deployment url.

src/Aspire.Dashboard/Model/ResourceViewModel.cs
adamint: removed nullability

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
adamint: Fair point, displaying GridValue now regardless

davidfowl:
> https://github.com/dotnet/aspire/pull/7442#issuecomment-2649124592

Don't block on this.

davidfowl:
@adamint We decided to move this to the next release to make it work with https://github.com/dotnet/aspire/issues/6457

adamint:
Moved to 9.2 - to be solved alongside https://github.com/dotnet/aspire/issues/6457

PR #7453, opened 2025-02-07
Allow EventHub connection strings to work with EntityPath and ConsumerGroup keys

eerhardt:
## Description

With #7438, EventHub Hub and Consumer Groups are Aspire Resources now, which means they can be referenced via WithReference.

In the future, the EventHubs Host integration will append these values to the connection string. This will allow the hub and consumer group names to be specified outside of the application. (Today they are hard-coded in both, or specified separately in config.) When they flow through the connection string, the EventHubName and ConsumerGroup properties are populated on the settings object.

Contributes to #7407

## Checklist

- Is this feature complete?
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
+++ b/src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
@@ -129,27 +129,10 @@ public abstract class AzureMessagingEventHubsSettings : IConnectionStringSetting
 
             // if we got here, it's a full connection string
             ConnectionString = connectionString;
-            //ConnectionString = connectionBuilder.ConnectionString;
         }
     }
 
     internal virtual void SetConsumerGroup(string? consumerGroup) { }
-    //internal virtual void ParseConnectionProperties(DbConnectionStringBuilder connectionBuilder) { }
-
-    //internal static bool TryParseConsumerGroup(DbConnectionStringBuilder connectionBuilder, out string? consumerGroup)
-    //{
-    //    if (connectionBuilder.TryGetValue("ConsumerGroup", out var group))
-    //    {
-    //        consumerGroup = group?.ToString();
-    //        // remove ConsumerGroup from the connection builder since it isn't a connection property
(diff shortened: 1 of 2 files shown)


src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
```diff
@@ -89,13 +90,66 @@ void IConnectionStringSettings.ParseConnectionString(string? connectionString)
...
+    //        consumerGroup = group?.ToString();
+    //        // remove ConsumerGroup from the connection builder since it isn't a connection property
+    //        // in regular Event Hubs connection strings
+    //        //connectionBuilder.Remove("ConsumerGroup");
+    //        return true;
+    //    }
+
+    //    consumerGroup = null;
+    //    return false;
+    //}
```
davidfowl, line 152: left over?

```suggestion
```

src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
```diff
@@ -89,13 +90,66 @@ void IConnectionStringSettings.ParseConnectionString(string? connectionString)
...
+                    // if all that's left is Endpoint or Endpoint+EntityPath,
+                    // it is a fully qualified namespace
+                    FullyQualifiedNamespace = endpoint.ToString();
+                    return;
+                }
+            }
+
+            // if we got here, it's a full connection string
+            ConnectionString = connectionString;
+            //ConnectionString = connectionBuilder.ConnectionString;
```
davidfowl, line 132: Left over?

```suggestion
```

src/Components/Aspire.Azure.Messaging.EventHubs/AzureMessagingEventHubsSettings.cs
eerhardt: 😊

PR #7477, opened 2025-02-08
Add Azure Event Hubs persistence functional test

sebastienros:
## Description

This is the first functional test verifying that persistence is effective for a container. I might have missed more efficient ways to do this. Currently only checking that the same container is used when a different app is started. Once approved we can add the same thing to other containers.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No


tests/Aspire.Hosting.Azure.Tests/AzureEventHubsExtensionsTests.cs
```diff
@@ -583,4 +583,62 @@ public void AzureEventHubsAppliesAzureFunctionsConfiguration()
...
+
+        var resourceEvent = await rns.WaitForResourceAsync("resource", e => e.Snapshot.State?.Text == KnownResourceStates.Running, cts.Token);
+        var ehContainerId1 = resourceEvent.Snapshot.Properties.FirstOrDefault(x => x.Name == "container.id")?.Value?.ToString();
+
+        resourceEvent = await rns.WaitForResourceAsync("resource-storage", e => e.Snapshot.State?.Text == KnownResourceStates.Running, cts.Token);
+        var storageContainerId1 = resourceEvent.Snapshot.Properties.FirstOrDefault(x => x.Name == "container.id")?.Value?.ToString();
+
+        await app.StopAsync(cts.Token).WaitAsync(TimeSpan.FromMinutes(1));
+
+        using var builder2 = TestDistributedApplicationBuilder.Create(testOutputHelper);
```
eerhardt, line 615: This test could use this new functionality:

https://github.com/dotnet/aspire/issues/4878#issuecomment-2648623757

That way we knew the app was stopped.

cc @davidfowl @karolz-ms

sebastienros: It's on by default. But I decided to disable it on purpose, it's not required for the test (we check we don't have a different container, the state of the previous one doesn't matter) and not waiting makes the test faster (19s -> 12s)

tests/Aspire.Hosting.Azure.Tests/AzureEventHubsExtensionsTests.cs
```diff
@@ -583,4 +584,73 @@ public void AzureEventHubsAppliesAzureFunctionsConfiguration()
...
+        var rns = app.Services.GetRequiredService<ResourceNotificationService>();
+
+        await rns.WaitForResourceHealthyAsync(resource.Resource.Name, cts.Token);
+
+        var resourceEvent = await rns.WaitForResourceAsync("resource", e => GetContainerId(e) is not null, cts.Token);
+        var ehContainerId1 = GetContainerId(resourceEvent);
+
+        // NOTE: Waiting for Running state doesn't guaranty that the container-id property is set (for this resource at least)
+
+        resourceEvent = await rns.WaitForResourceAsync("resource-storage", e => GetContainerId(e) is not null, cts.Token);
```
sebastienros, line 616: @karolz-ms 

I had some problems locally with this. I am trying to read the `"container-id"` property of the snapshot. However if I wait for the Running state and then read the id right after it returns null most of the time. For the previous resource (a parent) it works all the time. So now I use the actual value as the predicate. Also these containers were already running on my machine (persistent from a previous run) so the container always has an id and is running.

karolz-ms: It is a little surprising that a container in Running state does not have a container ID set since on DCP side this is done as one change that should be atomic.  But what you are doing here certainly makes sense.

The other thing to keep in mind is that even for running containers, the sequence of events represents the lifetime of the Container object in DCP. So the first event will always be a "Created" event with no container ID, and then DCP will notice that the existing Docker container can be reused, so it will fill the ContainerID property and change its Container object state to Running. And that change will be delivered as a separate event.

tests/Aspire.Hosting.Azure.Tests/AzureEventHubsExtensionsTests.cs
```diff
@@ -583,4 +584,73 @@ public void AzureEventHubsAppliesAzureFunctionsConfiguration()
...
+        await app.StartAsync(cts.Token);
+
+        var rns = app.Services.GetRequiredService<ResourceNotificationService>();
+
+        await rns.WaitForResourceHealthyAsync(resource.Resource.Name, cts.Token);
+
+        var resourceEvent = await rns.WaitForResourceAsync("resource", e => e.Snapshot.HealthStatus == HealthStatus.Healthy, cts.Token);
+        var ehContainerId1 = GetContainerId(resourceEvent);
+
+        // NOTE: Waiting for Running state doesn't guaranty that the container-id property is set (for this resource at least)
```
eerhardt, line 614:
```suggestion
        // NOTE: Waiting for Running state doesn't guarantee that the container-id property is set (for this resource at least)
```

sebastienros:
Closing, in the meantime I added MySql_WithPersistentLifetime_ReusesContainers and Postgres_WithPersistentLifetime_ReusesContainers which provide a better pattern to follow.

PR #7494, opened 2025-02-09
Follow-up fixes to RunAsExisting flow

captainsafia:
- Inline Tags configuration for key vault
- Set resource scope in user-secrets config
- Remove fallback to default user group if existing resource group was not found
- Fix resolution of resource group name from parameter resource

### DIFF
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
@@ -541,7 +541,7 @@ internal sealed class BicepProvisioner(
 
     internal static async Task SetScopeAsync(JsonObject scope, AzureBicepResource resource, CancellationToken cancellationToken = default)
     {
-        // Resole the scope from the AzureBicepResource if it has already been set
+        // Resolve the scope from the AzureBicepResource if it has already been set
         // via the ConfigureInfrastructure callback. If not, fallback to the ExistingAzureResourceAnnotation.
         var targetScope = resource.Scope;
         if (targetScope is null
--- a/src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
+++ b/src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
@@ -114,9 +114,9 @@ internal sealed class BicepProvisioner(
             existingResource.ResourceGroup is { } existingResourceGroup)
         {
             var existingResourceGroupName = existingResourceGroup is ParameterResource parameterResource
-                ? parameterResource.Name
+                ? parameterResource.Value
                 : (string)existingResourceGroup;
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -535,12 +541,22 @@ internal static async Task SetParametersAsync(JsonObject parameters, AzureBicepR
 
     internal static async Task SetScopeAsync(JsonObject scope, AzureBicepResource resource, CancellationToken cancellationToken = default)
     {
-        scope["resourceGroup"] = resource.Scope?.ResourceGroup switch
+        // Resole the scope from the AzureBicepResource if it has already been set
```
davidfowl, line 544:
```suggestion
        // Resolve the scope from the AzureBicepResource if it has already been set
```

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -114,9 +114,9 @@ public override async Task GetOrCreateResourceAsync(AzureBicepResource resource,
             existingResource.ResourceGroup is { } existingResourceGroup)
         {
             var existingResourceGroupName = existingResourceGroup is ParameterResource parameterResource
-                ? parameterResource.Name
+                ? parameterResource.Value
```
davidfowl, line 117: I knew it was doing this from the error I was getting

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -114,9 +114,9 @@ public override async Task GetOrCreateResourceAsync(AzureBicepResource resource,
...
                 : (string)existingResourceGroup;
-            resourceGroup = await context.GetResourceGroup(existingResourceGroupName, resourceLogger, cancellationToken).ConfigureAwait(false);
+            resourceGroup = await context.Subscription.GetResourceGroupAsync(existingResourceGroupName, cancellationToken).ConfigureAwait(false);
```
davidfowl, line 119: Does this handle if the resource group doesn't exist?

captainsafia: Depends on what you mean by handle. Before it would fallback to the default group. Now, it'll fail and you'll see something like this in the logs:

```
Provisioning sbemulator...
 Error provisioning sbemulator.
Azure.RequestFailedException: Resource group 'rg-aspire-deploy-not-real' could not be found.
Status: 404 (Not Found)
ErrorCode: ResourceGroupNotFound

Content:
{"error":{"code":"ResourceGroupNotFound","message":"Resource group 'rg-aspire-deploy-not-real' could not be found."}}
```

davidfowl (approved):

PR #7497, opened 2025-02-10
Display launchSettings.json args in dashboard

### LINKED ISSUE #7475: Launch profile args not displayed in dashboard

### Describe the bug

Launch profile args won't be displayed in the dashboard UI after this fix: https://github.com/dotnet/aspire/pull/7409

These are the args I'm talking about:

https://github.com/dotnet/aspire/blob/1100e7e88da73ca86ae39150695df90bd8ff7a64/src/Aspire.Hosting/Dcp/DcpExecutor.cs#L860-L869

Should they be displayed?

A fix could be to move that logic to `CreateExecutableAsync` and add them to the args along with the custom annotation.

cc @davidfowl @adamint 

### Expected Behavior

Launch profile args are displayed in the dashboard UI

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/7475

Before (no args in dashboard):

image

After:

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/DcpExecutor.cs
@@ -964,7 +964,7 @@ internal sealed class DcpExecutor : IDcpExecutor, IAsyncDisposable
         {
             throw new InvalidOperationException($"Expected an Executable resource, but got {er.DcpResource.Kind} instead");
         }
-        ExecutableSpec spec = exe.Spec;
+        var spec = exe.Spec;
 
         // An executable can be restarted so args must be reset to an empty state.
         // After resetting, first apply any dotnet project related args, e.g. configuration, and then add args from the model resource.
--- a/src/Aspire.Hosting/Dcp/DcpExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/DcpExecutor.cs
@@ -859,17 +859,6 @@ internal sealed class DcpExecutor : IDcpExecutor, IAsyncDisposable
                     // This means we need to apply the launch profile settings manually--the invocation parameters here,
                     // and the environment variables/application URLs inside CreateExecutableAsync().
                     projectArgs.Add("--no-launch-profile");
-
-                    var launchProfile = project.GetEffectiveLaunchProfile()?.LaunchProfile;
-                    if (launchProfile is not null && !string.IsNullOrWhiteSpace(launchProfile.CommandLineArgs))
-                    {
-                        var cmdArgs = CommandLineArgsParser.Parse(launchProfile.CommandLineArgs);
-                        if (cmdArgs.Count > 0)
-                        {
-                            projectArgs.Add("--");
-                            projectArgs.AddRange(cmdArgs);
-                        }
-                    }
                 }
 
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -994,8 +974,29 @@ private async Task CreateExecutableAsync(AppResource er, ILogger resourceLogger,
         {
             spec.Args.AddRange(projectArgs);
         }
-        spec.Args.AddRange(args.Select(a => a.Value));
-        er.DcpResource.SetAnnotationAsObjectList(CustomResource.ResourceAppArgsAnnotation, args.Select(a => new AppLaunchArgumentAnnotation(a.Value, isSensitive: a.IsSensitive)));
+
+        var launchArgs = new List<(string Value, bool IsSensitive)>();
+
+        // If the executable is a project then include any command line args from the launch profile.
+        if (er.ModelResource is ProjectResource project)
```
davidfowl, line 981: We have to make sure this does not change any behavior when there's IDE in use. In that situation the args is handled by the IDE even if we show them in the dashboard.

JamesNK: I tested when running TestShop from IDE. Should I check I get the same behavior when running form the command line? I don't understand the scenario you're talking about.

davidfowl: The code before has 2 branches, in IDE mode command line lanuch profile (and command line args) are processed 
by the IDE https://github.com/dotnet/aspire/blob/e574625ccf3f8c126e9822ee5015c6434c09ba87/src/Aspire.Hosting/Dcp/DcpExecutor.cs#L863-L872.

In your change that is happening in both the IDE and non IDE case regardless. We can process the launch profile for args, but if we want the behavior to match today, we shouldn't provide the args to the DCP resource (in this one case).

If we change this (which I am not 100% against), the IDE will never set arguments from the launch profile when doing processing. This means all of the weird VS tokens you can put in launch settings that only work in VS will stop working.

davidfowl: As a test, put $(SolutionDir) in the args and run before and after your change and see if it gets resolved

JamesNK: Ok, I'll look at this more tomorrow. If we do change something, I think it should probably be a 9.2 change. Give us time to see if it causes any issues.

davidfowl (approved):

PR #7518, opened 2025-02-10
Revert redis password change

### LINKED ISSUE #7429: [WebToolsE2E][Aspire] Using ‘azd up’ to deploy aspire starter with redis project fails with error: generating bicep from manifest: argument 1 cannot contain connection strings, secured parameters, or secret outputs. Use environment variables instead.
**REGRESSION INFO:** Worked in Aspire 9.0.0

**INSTALL STEPS**

2. Install SDK 9.0.200
   - Apply internal Feed
3. Install latest Aspire 9.1.0-preview.1.25105.6
   - Apply dotnet9 Feed
4. Install latest released azd [1.12.0]
5. Install Docker Desktop 4.38.0

**REPRO STEPS**

1. Open CMD, create an Aspire Starter App with redis using following commands:
```
dotnet new aspire-starter --use-redis-cache -o aspire01
cd aspire01/aspire01.AppHost
```
2. Run the following code to deploy
```
azd init
azd up
```
**ACTUAL**
ERROR: generating bicep from manifest: argument 1 cannot contain connection strings, secured parameters, or secret outputs. Use environment variables instead
Image

**NOTE**
- This issue didn't repro on Aspire Starter without Redis 

**EXPECTED**
Deployment Success

eerhardt:
## Description

Reverts [Add Password To Redis (dotnet/aspire#4642)](https://github.com/dotnet/aspire/pull/4642).

`azd` doesn't allow for secret parameters to be used in command line args of containers. See the [code comment here](https://github.com/Azure/azure-dev/blob/5f05eaf81dddbe74cf6d1d8481335e5ed32cb59f/cli/azd/pkg/apphost/generate.go#L1779-L1784).

```
		// Unlike environment variables, ACA doesn't provide a way to pass secret values without baking them into the args
		// array directly. We don't want folks to accidentally bake the plaintext value of these secrets into the container
		// definition, so for now, we block this.
		//
		// This logic is similar to what we do in buildEnvBlock to detect when we need to take values and treat them as ACA
		// secrets.
```

For .NET Aspire 9.1, we will revert back to the 9.0 behavior for Redis containers. We will revisit this issue in 9.2 to decide how we can password protect Redis containers in ACA.

Fixes #7429

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] Yes
  - [x] No

### DIFF
--- a/playground/AspireEventHub/EventHubs.AppHost/aspire-manifest.json
+++ b/playground/AspireEventHub/EventHubs.AppHost/aspire-manifest.json
@@ -22,6 +22,10 @@
         "principalId": ""
       }
     },
+    "eventhub": {
+      "type": "value.v0",
+      "connectionString": "Endpoint={eventhubns.outputs.eventHubsEndpoint};EntityPath=eventhub"
+    },
     "consumer": {
       "type": "project.v0",
       "path": "../EventHubsConsumer/EventHubsConsumer.csproj",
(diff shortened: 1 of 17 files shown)


playground/AspireEventHub/EventHubs.AppHost/aspire-manifest.json
```diff
@@ -22,14 +22,18 @@
         "principalId": ""
       }
     },
+    "eventhub": {
```
eerhardt, line 25: FYI - These are caused by other changes that didn't update the playground manifests.

davidfowl (approved):

ReubenBond (approved):

PR #7523, opened 2025-02-11
Allow ServiceBus connection strings to work with EntityPath key

eerhardt:
## Description

In the future, the ServiceBus Host integration will append this value to the connection string. This will allow the queue, topic, and subscription names to be specified outside of the application. (Today they are hard-coded in both, or specified separately in config.) We can also add new client APIs to register a Sender or Processor in DI.

Note that the Subscription can be passed in the EntityPath using the format {topic}/Subscriptions/{subscription}.

This is a stop-gap to support these values in .NET Aspire 9.1 client integration so it doesn't break applications when we start adding these values in the future.

Contributes to #7407

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
+++ b/src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Data.Common;
 using Aspire.Azure.Common;
 using Azure.Core;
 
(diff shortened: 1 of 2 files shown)


src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
```diff
@@ -87,11 +88,35 @@ void IConnectionStringSettings.ParseConnectionString(string? connectionString)
...
+            {
+                ConnectionString = connectionString
+            };
+
+            // Note: Strip out the EntityPath from the connection string in order
+            // to tell if we are left with just Endpoint.
+            // The EntityPath can contain a queue or topic name. And if it references a topic,
+            // it can contain {topic}/Subscriptions/{subscription}. See https://github.com/Azure/azure-sdk-for-net/pull/27070
+
+            if (connectionBuilder.TryGetValue("EntityPath", out var _))
```
sebastienros, line 104: Can be removed, equivalent to `connectionBuilder.Remove("EntityPath");` which returns `bool`

sebastienros (approved):

src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
eerhardt: In the future we will use the EntityPath value to initialize the queue/topic name, and optionally the subscription name.

PR #7496, opened 2025-02-10
Improve error handling of health monitoring background tasks

JamesNK:
## Description

WaitAsync with a timeout throws `TimeoutException` when reached. Even with a catch block around it, the exception is still reported in the debug console. Fix is to replace with `Task.WhenAny` and handle the result accordingly.

Fixes https://github.com/dotnet/aspire/issues/7486

Before:

image

After:

image

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Is this introducing a breaking change?
      - [ ] Yes
        - Link to aspire-docs issue (please use this [`breaking-change` template](https://github.com/dotnet/docs-aspire/issues/new?template=04-breaking-change.yml)):
      - [ ] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)):
  - [x] No

PR #7527, opened 2025-02-11
Reduce thrown exceptions

JamesNK:
Use `WaitAsync` again. Also, avoid throwing `DistributedApplicationException` when restarting resource.

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpPipelineBuilder.cs
+++ b/src/Aspire.Hosting/Dcp/DcpPipelineBuilder.cs
@@ -10,16 +10,16 @@ namespace Aspire.Hosting.Dcp;
 
 internal static class DcpPipelineBuilder
 {
-    public static ResiliencePipeline BuildDeleteRetryPipeline(ILogger logger)
+    public static ResiliencePipeline<bool> BuildDeleteRetryPipeline(ILogger logger)
     {
-        var ensureDeleteRetryStrategy = new RetryStrategyOptions()
+        var ensureDeleteRetryStrategy = new RetryStrategyOptions<bool>()
         {
             BackoffType = DelayBackoffType.Exponential,
             Delay = TimeSpan.FromMilliseconds(200),
             UseJitter = true,
             MaxRetryAttempts = 10, // Cumulative time for all attempts amounts to about 15 seconds
             MaxDelay = TimeSpan.FromSeconds(3),
-            ShouldHandle = new PredicateBuilder().Handle<Exception>(),
+            ShouldHandle = args => ValueTask.FromResult(!args.Outcome.Result),
             OnRetry = (retry) =>
             {
                 logger.LogDebug("Retrying check for deleted resource. Attempt: {Attempt}. Error message: {ErrorMessage}", retry.AttemptNumber, retry.Outcome.Exception?.Message);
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
```diff
@@ -342,26 +342,16 @@ internal async Task<bool> DelayAsync(ResourceEvent? currentEvent, TimeSpan delay
...
-                {
-                    _delayCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
-                }
                 _delayInterruptTcs = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+                delayInterruptedTask = _delayInterruptTcs.Task;
             }
 
-            var completedTask = await Task.WhenAny(Task.Delay(delay, _delayCts.Token), _delayInterruptTcs.Task).ConfigureAwait(false);
+            // Don't throw to avoid writing the thrown exception to the debug console.
+            // See https://github.com/dotnet/aspire/issues/7486
```
danmoseley, line 350: can't they get this from the change description? then the codebase doesn't clutter up with old close issue URLs

... but if this is team style, fair enough

JamesNK: We do it occassionally in aspnetcore and here.

This isn't a problem that's usually considered (exceptions notification in debug console) which is why I linked to the issue.

davidfowl:
backport this plz

davidfowl:
Is the error showing up in the console output? That will make people report bugs (especially if it wasn't happening in 9.0).

JamesNK:
The `DistributedApplicationException` error on restart appears on the dev console, but that isn't a regression.

The regression in 9.1 was many `TimeoutException` instances being thrown in the dev console, and that's already fixed in 9.1.

JamesNK:
> Dev console, are you talking about the debug window?

Yes. See screenshots here: https://github.com/dotnet/aspire/pull/7496

> Does it show up in the dashboard anywhere or in the app host console?

They just show up in the IDE debug console. These exceptions are all caught and ignored, but that doesn't stop them from appearing in the debug console.

I created an issue about reducing how many times we throw exceptions to improve this situation: https://github.com/dotnet/aspire/issues/7528

src/Aspire.Hosting/Dcp/DcpPipelineBuilder.cs
```diff
@@ -10,24 +10,24 @@ namespace Aspire.Hosting.Dcp;
...
-        var ensureDeleteRetryStrategy = new RetryStrategyOptions()
+        var ensureDeleteRetryStrategy = new RetryStrategyOptions<bool>()
         {
             BackoffType = DelayBackoffType.Exponential,
             Delay = TimeSpan.FromMilliseconds(200),
             UseJitter = true,
             MaxRetryAttempts = 10, // Cumulative time for all attempts amounts to about 15 seconds
             MaxDelay = TimeSpan.FromSeconds(3),
-            ShouldHandle = new PredicateBuilder().Handle<Exception>(),
+            ShouldHandle = args => ValueTask.FromResult(!args.Outcome.Result),
```
eerhardt, line 22: What about when an exception occurs that isn't Http NotFound? Should we retry on that?

eerhardt (approved):
I just had one question, the rest of the changes look fine to me.

src/Aspire.Hosting/Dcp/DcpPipelineBuilder.cs
JamesNK: No, the retry here is specifically just for not found.

PR #7567, opened 2025-02-12
Run tests on release branches

eerhardt:


wtgodbe (approved):

wtgodbe:
I'm not sure if this is true for github workflows, but this may need to be backported to the target release branches as well

PR #7568, opened 2025-02-12
[IGNORE] Try enabling playground tests

radical:
- **Try enabling playground tests**
- **enable workload tests for PRs**


davidfowl:
why did all of the whitespace change?

tests/Shared/WorkloadTesting/TemplateCustomHive.cs
```diff
@@ -100,12 +100,17 @@ public static string GetPackagePath(string builtNuGetsPath, string templatePacka
     public static async Task<CommandResult> InstallTemplatesAsync(string packagePath, string customHiveDirectory, string dotnet)
     {
         var installCmd = $"new install --debug:custom-hive {customHiveDirectory} {packagePath}";
+        string workingDir = BuildEnvironment.IsRunningOnCI
+                                ? Path.GetTempPath()
+                                : Path.Combine(BuildEnvironment.TempDir, "templates", "working"); // avoid running from the repo
+        if (!Directory.Exists(workingDir))
```
danmoseley, line 106: doesn't matter here, but just FYI for product code CreateDirectory works fine if the directory already exists, so you don't need to do Exists

PR #7588, opened 2025-02-13
Bump the priority of processing paramters to avoid infinite recursion

davidfowl:
## Description

Second part to https://github.com/dotnet/aspire/pull/7500. Since we changed the connection string resource to be a parameter resource (that points to itself), we needed to bump the priority of processing parameters VS connection strings much like we did in manifest parsing.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No

### DIFF
--- a/src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
+++ b/src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
@@ -634,6 +634,13 @@ internal sealed class AzureContainerAppsInfrastructure(
                     return (url, secretType);
                 }
 
+                if (value is ParameterResource param)
+                {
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -634,6 +634,13 @@ BicepValue<string> GetHostValue(string? prefix = null, string? suffix = null)
                     return (url, secretType);
                 }
 
+                if (value is ParameterResource param)
```
JamesNK, line 637: Consider a comment saying that it's important that this is resolved before connection strings.

But if a test catches a regression then not required.

JamesNK (approved):

PR #7599, opened 2025-02-13
Reapply Redis password

### LINKED ISSUE #3838: Revisit Redis password options

Alirexaa:
## Description

[Add Password To Redis (dotnet/aspire#464)](https://github.com/dotnet/aspire/pull/4642) has been reverted because of [[WebToolsE2E][Aspire] Using ‘azd up’ to deploy aspire starter with redis project fails with error: generating bicep from manifest: argument 1 cannot contain connection strings, secured parameters, or secret outputs. Use environment variables instead. (dotnet/aspire#7429)](https://github.com/dotnet/aspire/issues/7429).

This PR addresses the mentioned issue by changing the container entrypoint and providing the password from an environment variable. 

For more information see https://github.com/dotnet/aspire/issues/3838#issuecomment-2655175621

Fixes #3838

@eerhardt @davidfowl 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [ ] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -61,7 +91,57 @@ public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicat
...
+
+            builder.WithEnvironment(context =>
+            {
+                if (builder.Resource.PasswordParameter is { } password)
+                {
+                    context.EnvironmentVariables["REDIS_PASSWORD"] = password;
+                }
+            });
+
+            builder.WithAnnotation(new CommandLineArgsCallbackAnnotation(context =>
```
davidfowl, line 112: You can just do WithArgs here right?

Alirexaa: I did.

eerhardt (approved):
Thanks again for this contribution @Alirexaa! Nice work here.

I merged main and refactored the code a little bit to simplify it a bit.
But when deploying the app to ACA I found an issue that while Aspire doesn't use commas `,` in its generated passwords, `azd` will. So I updated the code to not allow special characters in the password.

This change LGTM - we should be able to finally fix this long standing security issue. Thanks again!

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -608,6 +608,7 @@ BicepValue<string> GetHostValue(string? prefix = null, string? suffix = null)
                     EndpointProperty.Url => GetHostValue($"{scheme}://", suffix: isHttpIngress ? null : $":{port}"),
                     EndpointProperty.Host or EndpointProperty.IPV4Host => GetHostValue(),
                     EndpointProperty.Port => port.ToString(CultureInfo.InvariantCulture),
+                    EndpointProperty.HostAndPort=> GetHostValue(suffix: isHttpIngress ? null : $":{port}"),
```
eerhardt, line 611: FYI - @davidfowl - I found a bug that we missed here and fixed it so all the generate-manifests work again.

eerhardt: I didn't know if I should switch off `isHttpIngress` or not. If you explicitly ask for `HostAndPort` and this is the httpIngress, should we still tack on `:80`?

eerhardt: Rethinking this, I think it should always put the port. The reasoning the URL can drop the port is because it contains the scheme at the begining. `http://` indicates what the port is. If you say `HostAndPort` we should always give you the host and the port.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -608,6 +608,7 @@ BicepValue<string> GetHostValue(string? prefix = null, string? suffix = null)
                     EndpointProperty.Url => GetHostValue($"{scheme}://", suffix: isHttpIngress ? null : $":{port}"),
                     EndpointProperty.Host or EndpointProperty.IPV4Host => GetHostValue(),
                     EndpointProperty.Port => port.ToString(CultureInfo.InvariantCulture),
+                    EndpointProperty.HostAndPort=> GetHostValue(suffix: $":{port}"),
```
eerhardt, line 611:
```suggestion
                    EndpointProperty.HostAndPort => GetHostValue(suffix: $":{port}"),
```

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
davidfowl: I agree, it should include host an port

PR #7604, opened 2025-02-13
Allow for nested children in the dashboard

### LINKED ISSUE #7580: Dashboard only allows for one level of nested children

### Describe the bug

With more complex resources, like CosmosDB, ServiceBus, etc., they have 2 levels of child resources.

```C#

var entries = db.AddContainer("entries", "/id");
```

`cosmos` is the parent of `db` is the parent of `entries`.

However, in the dashboard both `entries` and `db` show up as direct children of `cosmos`:

Image

### Expected Behavior

We should make `entries` a child of `db`.

### Steps To Reproduce

Run an AppHost with the above code and view the dashboard.

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

cc @JamesNK

eerhardt:
## Description

Originally the parent-child lookup contained the "root" parent and all the descendants in a flat list. This doesn't show up well in the dashboard because grandchildren are parented directly under their root.

Fix this by making the parent-child lookup only contain direct children, and recursively parent the descendants on their direct parent.

Fix #7580

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
+++ b/src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
@@ -238,6 +238,11 @@ internal sealed class ApplicationOrchestrator
                 StopTimeStamp = stopTimeStamp,
                 Properties = s.Properties.SetResourceProperty(KnownProperties.Resource.ParentName, parentName)
             }).ConfigureAwait(false);
+
+            // the parent name needs to be an instance name, not the resource name.
+            // parent the children of the child under the first resource instance.
+            await SetChildResourceAsync(child, child.GetResolvedResourceNames()[0], state, startTimeStamp, stopTimeStamp)
+                .ConfigureAwait(false);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
```diff
@@ -238,6 +238,11 @@ await _notificationService.PublishUpdateAsync(child, s => s with
                 StopTimeStamp = stopTimeStamp,
                 Properties = s.Properties.SetResourceProperty(KnownProperties.Resource.ParentName, parentName)
             }).ConfigureAwait(false);
+
+            // the parent name needs to be an instance name, not the resource name.
+            // parent the children of the child under the first resource instance.
+            await SetChildResourceAsync(child, child.GetResolvedResourceNames()[0], state, startTimeStamp, stopTimeStamp)
```
JamesNK, line 244: This isn't going to work well with replicas. But that isn't new.

JamesNK (approved):
Tested in TestShop and cosmos E2E and they both work:

image

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
eerhardt: How would you expect it to work with replicas? Which replica should the child go under?

JamesNK: I don't think it can work today. Lets see if it comes up before worrying about it.

PR #7635, opened 2025-02-16
fix casing of CosmosDB

danmoseley:
This breaks the build for me on Ubuntu. I assume the official Linux builds are using a case-insensitive file system (?)


sebastienros (approved):

sebastienros:
Weird it never broke anybody before. I it possible it's a change in the SDK (if you have 9.0.200 locally)?

danmoseley:
> Weird it never broke anybody before. I it possible it's a change in the SDK (if you have 9.0.200 locally)?

Could be, if the official builds are using older.

PR #7643, opened 2025-02-17
Update Microsoft.Extensions.AI to 9.3.0-preview.1.25114.11

stephentoub:
cc: @eerhardt

PR #7657, opened 2025-02-18
[release/9.1] Update Microsoft.Extensions.AI to 9.3.0-preview.1.25114.11 (#7643)

eerhardt:
Backport of #7643 to release/9.1

/cc @stephentoub

## Customer Impact

When we ship 9.1, users won't be getting the latest versions of our dependencies. Specifically, they won't be getting the latest version of Microsoft.Extensions.AI. And MEAI has had breaking API changes (because they are still in preview) which may cause problems if users try using the new version with Aspire.

This needed to update to all .NET package versions because a new MEAI brings in new versions transitively. CPM doesn't like it when you have mismatched versions. So I brought them all up to date with the release last week.

## Testing

Existing tests

## Risk

Relatively low. Just updating to the newest versions. Tests are all passing.

## Regression?

No.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -9,9 +9,9 @@
     <DefaultTargetFramework>net8.0</DefaultTargetFramework>
     <AllTargetFrameworks>$(DefaultTargetFramework);net9.0</AllTargetFrameworks>
     <!-- dotnet 8.0 versions for running tests -->
-    <DotNetRuntimePreviousVersionForTesting>8.0.12</DotNetRuntimePreviousVersionForTesting>
+    <DotNetRuntimePreviousVersionForTesting>8.0.13</DotNetRuntimePreviousVersionForTesting>
     <!-- dotnet 8.0 versions for running tests - used for workload tests -->
-    <DotNetSdkPreviousVersionForTesting>8.0.404</DotNetSdkPreviousVersionForTesting>
+    <DotNetSdkPreviousVersionForTesting>8.0.406</DotNetSdkPreviousVersionForTesting>
     <UseVSTestRunner>true</UseVSTestRunner>
     <!-- Enable to remove prerelease label. -->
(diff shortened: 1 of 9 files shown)


stephentoub (approved):

eng/Versions.props
```diff
@@ -35,46 +35,43 @@
...
-    <MicrosoftExtensionsConfigurationAbstractionsVersion>9.0.1</MicrosoftExtensionsConfigurationAbstractionsVersion>
-    <MicrosoftExtensionsConfigurationBinderVersion>9.0.1</MicrosoftExtensionsConfigurationBinderVersion>
-    <MicrosoftExtensionsDependencyInjectionAbstractionsVersion>9.0.1</MicrosoftExtensionsDependencyInjectionAbstractionsVersion>
-    <MicrosoftExtensionsLoggingAbstractionsVersion>9.0.1</MicrosoftExtensionsLoggingAbstractionsVersion>
-    <MicrosoftExtensionsOptionsVersion>9.0.1</MicrosoftExtensionsOptionsVersion>
-    <MicrosoftExtensionsPrimitivesVersion>9.0.1</MicrosoftExtensionsPrimitivesVersion>
-    <MicrosoftExtensionsHttpVersion>9.0.1</MicrosoftExtensionsHttpVersion>
-    <SystemFormatsAsn1Version>9.0.1</SystemFormatsAsn1Version>
-    <SystemTextJsonVersion>9.0.1</SystemTextJsonVersion>
+    <MicrosoftExtensionsHostingAbstractionsVersion>9.0.2</MicrosoftExtensionsHostingAbstractionsVersion>
```
sebastienros, line 64: Should we bump abstraction if we require a new feature? (expected answer: yeah it's necessary in this case)

eerhardt: Our current plan is to update to the latest all the time (except for runtime-band serviced packages stay on 8.0 / 9.0 split).

sebastienros (approved):

PR #7662, opened 2025-02-18
Hide secrets in source tooltip, correct ExpressionResolver logic for non-containers

### LINKED ISSUE #7632: Password in executable args is shown on the Resources page
I have an executable resource (`dotnet ef`) with arguments that contain a connection string for another resource (database) that includes a password. The connection string including password is being displayed on the dashboard on the Resources page in the table:

Image

I've tried various different ways of adding the argument but they all result in the password being shown. Note that the argument is manually escaped with quotes.

```csharp
...

var chatAppDb = postgres.AddDatabase("chatappdb");

var efMigrate = builder.AddExecutable("ef-migrate", "dotnet", "../ChatApp1", "ef", "database", "update", "--no-build", "--connection")
    .WithArgs(context => context.Args.Add(ReferenceExpression.Create($"\"{chatAppDb.Resource}\"")))
    .WaitFor(postgres);
```

### LINKED ISSUE #7634: Secret arguments are displayed in tooltip

### Describe the bug

When adding arguments to a resource as a secret it is correctly redacted on the dashboard resources page. However the tooltip shows the secret without any user action. I would have expected it to not show up.

Image

Going into the details popup it's also displayed without user interaction. I don't think secrets should show unless an explicit interaction like "show secrets" or click a special "eye" icon is done. Same as what is done for ENVs in the properties.

Image

Example: doing a demo with a connection string and unintentionally showing secrets.

### Expected Behavior

Execute a special gesture to reveal secrets on screen.

### Steps To Reproduce

```c#
var connectionString = builder.AddParameter("connectionString", "my-secret", secret: true);
var exe = builder.AddExecutable("myexe", "-arg1", "--connectionstring", connectionString);
```

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

adamint:
## Description

Took the opportunity to clean up my code in `GetSourceViewModel` as well to add a type `CommandLineInfo` instead of relying on a tuple. Hides secrets in tooltips using the same masking characters as appear in the GridValue.

Also removes logic in `ExpressionResolver` that resolves some types of expressions only for containers. All expressions except `EndpointReferenceExpression` and `EndpointReference` should be resolved for all resource types. The only other difference between containers and other types of resources is how HostUrl is resolved (special-cased for containers). 

Fixes #7634, fixes #7632

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

### DIFF
--- a/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
+++ b/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
@@ -14,71 +14,22 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
 
     internal static ResourceSourceViewModel? GetSourceViewModel(ResourceViewModel resource)
     {
-        CommandLineInfo? commandLineInfo;
-
-        // If the resource contains launch arguments, these project arguments should be shown in place of all executable arguments,
-        // which include args added by the app host
-        if (resource.TryGetAppArgs(out var launchArguments))
-        {
-            if (launchArguments.IsDefaultOrEmpty)
-            {
-                commandLineInfo = null;
-            }
-            else
-            {
-                var argumentsString = string.Join(" ", launchArguments);
-                if (resource.TryGetAppArgsSensitivity(out var areArgumentsSensitive))
-                {
-                    var arguments = launchArguments
-                        .Select((arg, i) => new LaunchArgument(arg, IsShown: !areArgumentsSensitive[i]))
-                        .ToList();
-
-                    commandLineInfo = new CommandLineInfo(
-                        Arguments: arguments,
-                        ArgumentsString: argumentsString,
-                        TooltipString: string.Join(" ", arguments.Select(arg => arg.IsShown
-                            ? arg.Value
-                            : DashboardUIHelpers.GetMaskingText(6).Text)));
-                }
-                else
-                {
-                    commandLineInfo = new CommandLineInfo(Arguments: launchArguments.Select(arg => new LaunchArgument(arg, true)).ToList(), ArgumentsString: argumentsString, TooltipString: argumentsString);
-                }
-            }
-        }
-        else if (resource.TryGetExecutableArguments(out var executableArguments) && !resource.IsProject())
-        {
-            var arguments = executableArguments.IsDefaultOrEmpty ? [] : executableArguments.Select(arg => new LaunchArgument(arg, true)).ToList();
-            var argumentsString = string.Join(" ", executableArguments);
-
-            commandLineInfo = new CommandLineInfo(Arguments: arguments, ArgumentsString: argumentsString, TooltipString: argumentsString);
-        }
-        else
-        {
-            commandLineInfo = null;
-        }
+        var commandLineInfo = GetCommandLineInfo(resource);
 
         // NOTE projects are also executables, so we have to check for projects first
         if (resource.IsProject() && resource.TryGetProjectPath(out var projectPath))
         {
-            return commandLineInfo is not null
-                ? new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: commandLineInfo.Arguments, valueToVisualize: $"{projectPath} {commandLineInfo.ArgumentsString}", tooltip: $"{projectPath} {commandLineInfo.TooltipString}")
-                // default to project path if there is no executable path or executable arguments
-                : new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: null, valueToVisualize: projectPath, tooltip: projectPath);
+            return CreateResourceSourceViewModel(Path.GetFileName(projectPath), projectPath, commandLineInfo);
         }
 
         if (resource.TryGetExecutablePath(out var executablePath))
         {
-            return commandLineInfo is not null
-                ? new ResourceSourceViewModel(value: Path.GetFileName(executablePath), contentAfterValue: commandLineInfo.Arguments, valueToVisualize: $"{executablePath} {commandLineInfo.ArgumentsString}", tooltip: $"{executablePath} {commandLineInfo.TooltipString}")
-                : new ResourceSourceViewModel(value: Path.GetFileName(executablePath), contentAfterValue: null, valueToVisualize: executablePath, tooltip: executablePath);
+            return CreateResourceSourceViewModel(Path.GetFileName(executablePath), executablePath, commandLineInfo);
         }
 
         if (resource.TryGetContainerImage(out var containerImage))
         {
-            return commandLineInfo is not null
-                ? new ResourceSourceViewModel(value: containerImage, contentAfterValue: commandLineInfo.Arguments, valueToVisualize: $"{containerImage} {commandLineInfo.ArgumentsString}", tooltip: $"{containerImage} {commandLineInfo.TooltipString}")
-                : new ResourceSourceViewModel(value: containerImage, contentAfterValue: null, valueToVisualize: containerImage, tooltip: containerImage);
+            return CreateResourceSourceViewModel(containerImage, containerImage, commandLineInfo);
         }
 
         if (resource.Properties.TryGetValue(KnownProperties.Resource.Source, out var property) && property.Value is { HasStringValue: true, StringValue: var value })
@@ -87,6 +38,53 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
         }
 
         return null;
+
+        static CommandLineInfo? GetCommandLineInfo(ResourceViewModel resourceViewModel)
+        {
+            // If the resource contains launch arguments, these project arguments should be shown in place of all executable arguments,
+            // which include args added by the app host
+            if (resourceViewModel.TryGetAppArgs(out var launchArguments))
+            {
+                if (launchArguments.IsDefaultOrEmpty)
+                {
+                    return null;
+                }
+
+                var argumentsString = string.Join(" ", launchArguments);
+                if (resourceViewModel.TryGetAppArgsSensitivity(out var areArgumentsSensitive))
+                {
+                    var arguments = launchArguments
+                        .Select((arg, i) => new LaunchArgument(arg, IsShown: !areArgumentsSensitive[i]))
+                        .ToList();
+
+                    return new CommandLineInfo(
+                        Arguments: arguments,
+                        ArgumentsString: argumentsString,
+                        TooltipString: string.Join(" ", arguments.Select(arg => arg.IsShown
+                            ? arg.Value
+                            : DashboardUIHelpers.GetMaskingText(6).Text)));
+                }
+
+                return new CommandLineInfo(Arguments: launchArguments.Select(arg => new LaunchArgument(arg, true)).ToList(), ArgumentsString: argumentsString, TooltipString: argumentsString);
+            }
+
+            if (resourceViewModel.TryGetExecutableArguments(out var executableArguments) && !resourceViewModel.IsProject())
+            {
+                var arguments = executableArguments.IsDefaultOrEmpty ? [] : executableArguments.Select(arg => new LaunchArgument(arg, true)).ToList();
+                var argumentsString = string.Join(" ", executableArguments);
+
+                return new CommandLineInfo(Arguments: arguments, ArgumentsString: argumentsString, TooltipString: argumentsString);
+            }
+
+            return null;
+        }
+
+        static ResourceSourceViewModel CreateResourceSourceViewModel(string value, string path, CommandLineInfo? commandLineInfo)
+        {
+            return commandLineInfo is not null
+                ? new ResourceSourceViewModel(value: value, contentAfterValue: commandLineInfo.Arguments, valueToVisualize: $"{path} {commandLineInfo.ArgumentsString}", tooltip: $"{path} {commandLineInfo.TooltipString}")
+                : new ResourceSourceViewModel(value: value, contentAfterValue: null, valueToVisualize: path, tooltip: path);
+        }
     }
 
     private record CommandLineInfo(List<LaunchArgument> Arguments, string ArgumentsString, string TooltipString);
--- a/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
+++ b/src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
@@ -1,6 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Net;
+using Aspire.Dashboard.Utils;
+
 namespace Aspire.Dashboard.Model;
 
 public class ResourceSourceViewModel(string value, List<LaunchArgument>? contentAfterValue, string valueToVisualize, string tooltip)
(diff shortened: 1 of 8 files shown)


src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -31,46 +34,50 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
...
+                    commandLineInfo = new CommandLineInfo(Arguments: launchArguments.Select(arg => new LaunchArgument(arg, true)).ToList(), ArgumentsString: argumentsString, TooltipString: argumentsString);
                 }
             }
         }
         else if (resource.TryGetExecutableArguments(out var executableArguments) && !resource.IsProject())
         {
-            var arguments = executableArguments.IsDefaultOrEmpty ? null : executableArguments.Select(arg => new LaunchArgument(arg, true)).ToList();
-            commandLineInfo = (Arguments: arguments, string.Join(' ', executableArguments));
+            var arguments = executableArguments.IsDefaultOrEmpty ? [] : executableArguments.Select(arg => new LaunchArgument(arg, true)).ToList();
+            commandLineInfo = new CommandLineInfo(Arguments: arguments, ArgumentsString: string.Join(' ', executableArguments), TooltipString: string.Join(' ', executableArguments));
```
eerhardt, line 53:
```suggestion
            var argumentsString = string.Join(' ', executableArguments);
            commandLineInfo = new CommandLineInfo(Arguments: arguments, ArgumentsString: argumentsString, TooltipString: argumentsString);
```

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -154,16 +154,14 @@ async Task<ResolvedValue> EvalValueProvider(IValueProvider vp)
     /// </summary>
     async ValueTask<ResolvedValue> ResolveInternalAsync(object? value)
     {
-        return (value, sourceIsContainer) switch
```
eerhardt, line 157: Does the summary comment need to be updated at all?

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -154,16 +154,14 @@ async Task<ResolvedValue> EvalValueProvider(IValueProvider vp)
...
-            (IResourceWithConnectionString cs, true) => await ResolveInternalAsync(cs.ConnectionStringExpression).ConfigureAwait(false),
-            (ReferenceExpression ex, false) => await EvalExpressionAsync(ex).ConfigureAwait(false),
-            (ReferenceExpression ex, true) => await EvalExpressionAsync(ex).ConfigureAwait(false),
-            (EndpointReference endpointReference, true) => new(await EvalEndpointAsync(endpointReference, EndpointProperty.Url).ConfigureAwait(false), false),
-            (EndpointReferenceExpression ep, true) => new(await EvalEndpointAsync(ep.Endpoint, ep.Property).ConfigureAwait(false), false),
-            (IValueProvider vp, false) => await EvalValueProvider(vp).ConfigureAwait(false),
-            (IValueProvider vp, true) => await EvalValueProvider(vp).ConfigureAwait(false),
+            ConnectionStringReference cs => await ResolveInternalAsync(cs.Resource.ConnectionStringExpression).ConfigureAwait(false),
+            IResourceWithConnectionString cs => await ResolveInternalAsync(cs.ConnectionStringExpression).ConfigureAwait(false),
```
eerhardt, line 160: @davidfowl @mitchdenny @davidebbo - do we even need these 2 cases? Since `ConnectionStringReference` and `IResourceWithConnectionString` implement `IValueProvider`, shouldn't this just fall to that case?

(Note we don't necessarily need to change this in this PR, but I'm just trying to understand if I have the logic correct. If we are going to backport this to 9.1, we may not want to risk a change here.)

eerhardt: Nevermind, I think I understand why this is necessary - because we want to intercept any "host" references inside the ConnectionString and redirect them to the containerHostName.

Would it ever work if someone implemented an IValueProvider that contained a "host" reference, but wasn't one of these?

eerhardt (approved):
LGTM. I just had some questions for my learning.

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
adamint: It does, thanks. Updated

adamint:
Example with a connection string, like in @DamianEdwards's example in #7632 - 
image
image

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -148,22 +151,28 @@ async Task<ResolvedValue> EvalValueProvider(IValueProvider vp)
         return new ResolvedValue(value, false);
     }
 
+    async Task<ResolvedValue> ResolveConnectionStringReferenceAsync(ConnectionStringReference cs)
+    {
+        // GetValueAsync will throw if the connection string is not optional and does not exist.
+        // Invoke GetValueAsync to make sure the value isn't resolved, since we are manually resolving the expression.
+        await ((IValueProvider)cs).GetValueAsync().ConfigureAwait(false);
```
JamesNK, line 158: This should be passed the cancellation token.

adamint:
@eerhardt I added some additional special cases for certain resource kinds in ExpressionEvaluator, enough changes that need a re-review

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
JamesNK:
```suggestion
        await ((IValueProvider)cs).GetValueAsync(cancellationToken).ConfigureAwait(false);
```

danmoseley:
@adamint please get this in 9.1 today if possible.

adamint:
> @adamint please get this in 9.1 today if possible.

Waiting on re-review. @eerhardt @davidfowl

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -148,22 +151,28 @@ async Task<ResolvedValue> EvalValueProvider(IValueProvider vp)
         return new ResolvedValue(value, false);
     }
 
+    async Task<ResolvedValue> ResolveConnectionStringReferenceAsync(ConnectionStringReference cs)
+    {
+        // GetValueAsync will throw if the connection string is not optional and does not exist.
+        // Invoke GetValueAsync to make sure the value isn't resolved, since we are manually resolving the expression.
```
eerhardt, line 157: Can we beef up this comment a little bit, so it is more clear why we are calling GetValueAsync? Is it only to produce the exception and not duplicate code?

tests/Aspire.Dashboard.Tests/Model/ResourceSourceViewModelTests.cs
```diff
@@ -94,7 +96,7 @@ void AddStringProperty(string propertyName, string? propertyValue)
                 value: "project",
                 contentAfterValue: [new LaunchArgument("arg2", true), new LaunchArgument("--key", true), new LaunchArgument("secret", false)],
                 valueToVisualize: "path/to/project arg2 --key secret",
-                tooltip: "path/to/project arg2 --key secret"));
+                tooltip: $"path/to/project arg2 --key {WebUtility.HtmlDecode(DashboardUIHelpers.GetMaskingText(6).Value)}"));
```
eerhardt, line 99: Are there other tests that should be added for this change?

eerhardt (approved):
The changes look good to me. Just a couple minor comments.

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -12,15 +15,15 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
 
     internal static ResourceSourceViewModel? GetSourceViewModel(ResourceViewModel resource)
     {
-        (List<LaunchArgument>? Arguments, string? ArgumentsString) commandLineInfo;
+        CommandLineInfo? commandLineInfo;
```
JamesNK, line 18: Can the code for building the CommandLineInfo be split into its own method?

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -31,46 +34,52 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
...
-            {
-                return new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: arguments, valueToVisualize: $"{projectPath} {fullCommandLine}", tooltip: $"{projectPath} {fullCommandLine}");
-            }
-
-            // default to project path if there is no executable path or executable arguments
-            return new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: commandLineInfo.Arguments, valueToVisualize: projectPath, tooltip: projectPath);
+            return commandLineInfo is not null
+                ? new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: commandLineInfo.Arguments, valueToVisualize: $"{projectPath} {commandLineInfo.ArgumentsString}", tooltip: $"{projectPath} {commandLineInfo.TooltipString}")
+                // default to project path if there is no executable path or executable arguments
+                : new ResourceSourceViewModel(value: Path.GetFileName(projectPath), contentAfterValue: null, valueToVisualize: projectPath, tooltip: projectPath);
```
JamesNK, line 68: There is duplication in creating the ResourceSourceViewModel in the three paths. Does it make sense to have a helper method?

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
```diff
@@ -31,46 +34,52 @@ public class ResourceSourceViewModel(string value, List<LaunchArgument>? content
                         .Select((arg, i) => new LaunchArgument(arg, IsShown: !areArgumentsSensitive[i]))
                         .ToList();
 
-                    commandLineInfo = (Arguments: arguments, argumentsString);
+                    commandLineInfo = new CommandLineInfo(
+                        Arguments: arguments,
+                        ArgumentsString: argumentsString,
+                        TooltipString: string.Join(" ", arguments.Select(arg => arg.IsShown
+                            ? arg.Value
+                            : WebUtility.HtmlDecode(DashboardUIHelpers.GetMaskingText(6).Value))));
```
JamesNK, line 42: Add a unit test to ResourceSourceViewModel tests that tests this.

This PR could be backported so it should be well tested.

JamesNK (commented):
Dashboard changes look good. Need some tests and a little clean up.

Is there a sample where this can be manually tested?

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
JamesNK, line 42: `WebUtility.HtmlDecode` here is a bit hacky. `GetMaskingText` was indented for text that's included in the Razor page. I'm guessing the decode is neccessary to avoid the already escaped characters being escaped again.

Perhaps `GetMaskingText` should return a record that has escaped and unescaped strings on it. You would then used the unescaped string here. Could do this as a follow up change.

JamesNK:
Is there a sample where this can be manually tested?

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
adamint: Sure, changed to a record, TextMask

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
adamint: Added a helper method CreateResourceSourceViewModel

src/Aspire.Dashboard/Model/ResourceSourceViewModel.cs
adamint: there is a case in ResourceSourceViewModel_ReturnsCorrectValue_TestData that tests this specific case!

adamint:
> Is there a sample where this can be manually tested?

There is not, as there is no sample using ef migrate or this feature. We could add one? but I'm unsure exactly how to replicate the use case in #7632 - @DamianEdwards maybe for suggestion, or could you share the app you were referencing there?

I've been manually testing using the following, inspired by #7632 -

Update any service in TestShop.AppHost/Program.cs to include `.WithArgs(catalogDb.Resource)`

JamesNK (approved):

tests/Aspire.Dashboard.Tests/Model/ResourceSourceViewModelTests.cs
adamint: Added those tests in `ExpressionResolverTests`

src/Aspire.Hosting/ApplicationModel/ExpressionResolver.cs
```diff
@@ -148,22 +151,29 @@ async Task<ResolvedValue> EvalValueProvider(IValueProvider vp)
         return new ResolvedValue(value, false);
     }
 
+    async Task<ResolvedValue> ResolveConnectionStringReferenceAsync(ConnectionStringReference cs)
+    {
+        // We are substituting our own logic for ConnectionStringReference's GetValueAsync.
+        // However, ConnectionStringReference#GetValueAsync will throw if the connection string is not optional but is not present.
+        // To avoid duplicating that logic, we can defer to ConnectionStringReference#GetValueAsync, which will throw if needed.
+        await ((IValueProvider)cs).GetValueAsync(cancellationToken).ConfigureAwait(false);
```
davidfowl, line 159: I would be worried about side effects. We should only do this in the null case to avoid double evaluation in the common case.

mitchdenny:
@adamint I was taking a closer look at this PR and I noticed a bug. Lets say I do this:

builder.Configuration["ConnectionStrings:connstring"] = "dummyvalue";
var connstring = builder.AddConnectionString("connstring");
builder.AddExecutable("pinger", "path/to/ping", "localhost", connstring);

When this renders in the console it'll look like this. The connectionstring type name itself is being rendered into the source column. I don't think that is the intention? It should just show the secret placeholder value.

image

mitchdenny:
Also issue #7634 suggests that we should also have a show secrets button in the copy dialog. I'm not 100% sure I agree that we need to do that - but was that a deliberate choice. If it was just skipped for expediency for 9.1 and we still want to do it then we should re-open that issue or open a new one to make sure we don't loose track.

adamint:
> Also issue #7634 suggests that we should also have a show secrets button in the copy dialog. I'm not 100% sure I agree that we need to do that - but was that a deliberate choice. If it was just skipped for expediency for 9.1 and we still want to do it then we should re-open that issue or open a new one to make sure we don't loose track.

 It's a deliberate choice, I'll make a comment in #7634 as well - opening the text visualizer _is_ an explicit action. With secret properties in the resource details view, users wouldn't otherwise be able to hide secrets without them being initially masked, of course. Here, secrets are already hidden in the source column; if users want to see all args, they can expand the column or hover to show the (masked) tooltip.

PR #7669, opened 2025-02-19
Bump timeout for non-helix tests

radical:
- **Bump timeout for non-helix tests**
- **Use blame hang timeout for tests same as used for PRs**


radical:
Running a CI build for this at https://dev.azure.com/dnceng-public/public/_build/results?buildId=955925&view=results .

karolz-ms (approved):

PR #7677, opened 2025-02-19
Support Existing Azure AppInsights, OpenAI, and CosmosDB

### LINKED ISSUE #7665: Missing Azure integrations that don't support "AsExisting"

### Describe the bug

We have 3 Azure integrations that don't support AsExisting today:

* AppInsights
* CongitiveServices
* CosmosDB

### Expected Behavior

When using "AsExisting" on these 3 services should support pointing to the existing resource.

### Steps To Reproduce

Run code:

```C#
var azureAIName = builder.AddParameter("azureai-name");
var azureAIRG = builder.AddParameter("azureai-rg");

var azureAI = builder.AddAzureOpenAI("azureai")
    .AsExisting(azureAIName, azureAIRG)
```

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

_No response_

eerhardt:
## Description

These resources don't support existing resources yet. Adding support to finish all the Azure resources for existing.

Note that for CosmosDB and Redis, we cannot explicitly set DisableLocalAuth on an existing resource because bicep doesn't support it.

Fix #7665

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
@@ -324,16 +324,26 @@ public static class AzureCosmosExtensions
     private static void ConfigureCosmosDBInfrastructure(AzureResourceInfrastructure infrastructure)
     {
         var azureResource = (AzureCosmosDBResource)infrastructure.AspireResource;
+        bool disableLocalAuth = !azureResource.UseAccessKeyAuthentication;
 
-        var cosmosAccount = new CosmosDBAccount(infrastructure.AspireResource.GetBicepIdentifier())
-        {
-            Kind = CosmosDBAccountKind.GlobalDocumentDB,
-            ConsistencyPolicy = new ConsistencyPolicy()
+        var cosmosAccount = AzureProvisioningResource.CreateExistingOrNewProvisionableResource(infrastructure,
+            (identifier, name) =>
             {
-                DefaultConsistencyLevel = DefaultConsistencyLevel.Session
+                var resource = new ExistingCosmosDBAccount(identifier, disableLocalAuth)
+                {
+                    Name = name,
+                };
+                return resource;
             },
-            DatabaseAccountOfferType = CosmosDBAccountOfferType.Standard,
-            Locations =
+            (infrastructure) => new CosmosDBAccount(infrastructure.AspireResource.GetBicepIdentifier())
+            {
+                Kind = CosmosDBAccountKind.GlobalDocumentDB,
+                ConsistencyPolicy = new ConsistencyPolicy()
+                {
+                    DefaultConsistencyLevel = DefaultConsistencyLevel.Session
+                },
+                DatabaseAccountOfferType = CosmosDBAccountOfferType.Standard,
+                Locations =
                 {
                     new CosmosDBAccountLocation
                     {
@@ -341,9 +351,9 @@ public static class AzureCosmosExtensions
                         FailoverPriority = 0
                     }
                 },
-            Tags = { { "aspire-resource-name", infrastructure.AspireResource.Name } }
-        };
-        infrastructure.Add(cosmosAccount);
+                DisableLocalAuth = disableLocalAuth,
+                Tags = { { "aspire-resource-name", infrastructure.AspireResource.Name } }
+            });
 
         foreach (var database in azureResource.Databases)
         {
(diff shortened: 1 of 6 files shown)


src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -428,6 +434,28 @@ private static void ConfigureCosmosDBInfrastructure(AzureResourceInfrastructure
...
+internal sealed class ExistingCosmosDBAccount : CosmosDBAccount
+{
+    public ExistingCosmosDBAccount(string bicepIdentifier, bool disableLocalAuth, string? resourceVersion = null) : base(bicepIdentifier, resourceVersion)
+    {
+        // only explicitly set DisableLocalAuth if WithAccessKeyAuthentication was called.
+        // We don't want to disable local auth on an existing resource if it is already enabled
+        // as that might break other uses of the resource.
+        if (disableLocalAuth is false)
+        {
+            DisableLocalAuth = false;
```
eerhardt, line 452: @captainsafia - I see for Redis that we always set it, is that intentional?

https://github.com/dotnet/aspire/blob/acdf5012407c2ab0259f4b6f075acb6c668949f3/src/Aspire.Hosting.Azure.Redis/AzureRedisExtensions.cs#L287-L294

https://github.com/dotnet/aspire/blob/acdf5012407c2ab0259f4b6f075acb6c668949f3/src/Aspire.Hosting.Azure.Redis/AzureRedisExtensions.cs#L205-L208

captainsafia: Yes, this was intentional although you make a good point in your comment here and we probably want to constrain the Redis resource in the same way.

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -1254,6 +1275,314 @@ param keyVaultName string
             }
             """;
 
+        output.WriteLine(BicepText);
```
captainsafia, line 1278: Is this just here to help with test output?

eerhardt: It's there to help when the bicep changes and you need to know what changed. You can copy the actual bicep from the test log and paste it into the test code. That will show you all the differences. This is better/easier than getting the diffs one at a time from the test failure:

image

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -134,6 +138,8 @@ param principalId string
 
             output serviceBusEndpoint string = messaging.properties.serviceBusEndpoint
             """;
+
+        output.WriteLine(BicepText);
```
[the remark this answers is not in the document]
eerhardt: It isn't redundant. It is helpful to diagnose changes in behavior.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
eerhardt: It turns out this doesn't work end-to-end. If you set this property on an existing resource you get:

> Error BCP173: The property "properties" cannot be used in an existing resource declaration. [https://aka.ms/bicep/core-diagnostics#BCP173]

See https://github.com/Azure/azure-sdk-for-net/issues/48352#issuecomment-2669386450

eerhardt: I've updated CosmosDB and Redis to no longer do this. I will make a follow up for Postgres since it has the same problem.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -324,26 +324,34 @@ public static IResourceBuilder<AzureCosmosDBResource> WithAccessKeyAuthenticatio
...
                     new CosmosDBAccountLocation
                     {
                         LocationName = new IdentifierExpression("location"),
                         FailoverPriority = 0
                     }
                 },
-            Tags = { { "aspire-resource-name", infrastructure.AspireResource.Name } }
-        };
-        infrastructure.Add(cosmosAccount);
+                DisableLocalAuth = disableLocalAuth,
```
captainsafia, line 352: +1 to this. Setting this directly on the resource instead of after the fact reads better as well.

captainsafia (approved):

PR #7697, opened 2025-02-19
Add support for specifying an image pull policy for containers

### LINKED ISSUE #7350: Expose Image Pull Policy on Containers

Currently if you reference the container via a mutable tag (e.g. `redis:latest`), then docker will only pull the latest version of that tag if one doesn't exist locally.  This could mean that your `latest` tag is many versions behind if you haven't explicitly done a `docker pull redis:latest` to refresh your local version.

Docker Compose and Kubernetes expose Image Pull Policies that can be used to configure this behavior, it would be nice if Aspire exposes the same.

This functionality is available through `--pull` on `docker container create`.  Alternatively it can be implemented by explicitly running `docker pull` before creating the container image

From a user perspective, I imagine this being a `WithImagePullPolicy` extension method on IResourceBuilder
```cs
container
    .WithImagePullPolicy(ImagePullPolicy.Always) // Always | Missing (default) | Never
```

The naming of the method and arguments matching what are used by Kubernetes & docker for this same capability:
- Docker Compose: https://docs.docker.com/reference/compose-file/services/#pull_policy
- Kubernetes: https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

### Additional context

We're looking to create a number of internal components re-distributing our production services as `ContainerResources` via shared nuget packages.  These services follow a trunk based CD model and could be released several times a day - because of this we want our aspire component to reference a mutable container tag (e.g. `latest` or `production`) and not have to update the version number in our nuget package every release, and then get all consumers to )

In our early days of rolling this out, we've already encountered a couple of issues at this point which we root causes as being caused by the consumer having an ancient version of the container that they used for some ad-hoc testing 1-2 years ago!  Having the ability to set an image pull policy to always would greatly help.

danegsta:
## Description

Adds a new ContainerImagePullPolicyAnnotation and WithImagePullPolicy builder API to allow configuring the container image pull policy for a given container resource. The supported pull policy settings are `Default`, `Always`, and `Missing`. A pull policy of `Always` will result in the container image always being pulled when the container is run, regardless of whether a version of the image is cached locally. `Missing` will only pull images if there is no cached version locally (and therefore you may not be using the latest version of the image). `Default` represents the default behavior if no policy is specified. Currently that will be the default behavior for your container runtime (which is generally the same as `Missing` for both Docker and Podman), however that could change in future Aspire releases.

When building a new image from a Dockerfile, `Always` will also force pulling any base images as part of the build, `Missing` will use the default behavior of only pulling images that aren't available locally, and `Default` will again match the runtime's default build process behavior.

Example:

```C#
container.WithImagePullPolicy(ImagePullPolicy.Always);
```

Fixes #7350

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No (it would be difficult to validate in unit tests whether Docker had pulled an image)
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
  - [ ] No


src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    Always,
+    /// <summary>
+    /// Pull the image only if it does not already exist.
+    /// </summary>
+    Missing,
+    /// <summary>
+    /// Never pull the image.
+    /// </summary>
+    Never
```
eerhardt, line 24: Why would we want `Never`?

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
```diff
@@ -0,0 +1,37 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Image pull policies for container resources.
+/// </summary>
+public enum PullPolicy
```
eerhardt, line 11: Should we add a `Default = 0` enum value?

I've hoped that we would periodically pull new images (say once a day) in the future. If we don't put a 0 enum value now,  `Always` will be the 0 enum value.

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
danegsta: It's definitely an edge case setting; it's supported in DCP, but I'm happy to leave it out of the Aspire types until such time as we get feedback asking for it.

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
danegsta: Adding a default makes sense to make future API/behavior changes more straightforward.

davidfowl:
Does this mean we’ll get the pulling status separately in the ux?

danegsta:
> Does this mean we’ll get the pulling status separately in the ux?

This PR won't give any pulling status updates, just allow controlling the behavior of when new images are pulled. Getting image pull progress from the Docker or Podman CLI is non-trivial as they only provide detailed breakdowns when running in a TTY environment.

One thing we may want to consider is making container images a first class DCP resource so that we can control when to pull them independent of running the container that requires the image; that way we can differentiate between pulling an image and general container startup when reporting status (even if we can't give detailed progress on the image pull).

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
WeihanLi: Should we consider naming it `ContainerImagePullPolicy` etc, think the `PullPolicy` is a little generic

danegsta: I updated the enum and annotation to `ImagePullPolicy` and `ContainerImagePullPolicyAnnotation` respectively.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
```diff
@@ -317,6 +317,20 @@ public static IResourceBuilder<T> WithLifetime<T>(this IResourceBuilder<T> build
...
     }
 
+    /// <summary>
+    /// Sets the pull policy for the container resource.
+    /// </summary>
+    /// <typeparam name="T">The resource type.</typeparam>
+    /// <param name="builder">Builder for the container resource.</param>
+    /// <param name="pullPolicy">The pull policy behavior for the container resource.</param>
+    /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    public static IResourceBuilder<T> WithImagePullPolicy<T>(this IResourceBuilder<T> builder, ImagePullPolicy pullPolicy) where T : ContainerResource
```
eerhardt, line 327: Are there tests we can write for this?

At a minimum at least calling this API and ensuring the annotation is there. Ideally, if we can tell what data gets sent to DCP, we could ensure the image pull policy is set correctly.

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <summary>
+/// Annotation that controls the image pull policy for a container resource.
+/// </summary>
+[DebuggerDisplay("Type = {GetType().Name,nq}")]
+public sealed class ContainerImagePullPolicyAnnotation : IResourceAnnotation
+{
+    /// <summary>
+    /// Gets or sets the image pull policy for the container resource.
+    /// </summary>
+    public required ImagePullPolicy ImagePullPolicy { get; set; }
```
eerhardt, line 36: @mitchdenny - do we usually have immutable annotation classes? Should this be a ctor parameter and a get-only property?

eerhardt (commented):
Just 2 comments from me. Once those are addressed, I think this change looks good to go in.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
danegsta: Unit tests added

src/Aspire.Hosting/ApplicationModel/ContainerPullPolicyAnnotation.cs
danegsta: I based the implementation off ContainerImageAnnotation, which uses this pattern, but there are others that use constructors. I'd argue record types could be a good fit for the annotations if we ever decided like making that much of a breaking change.

davidfowl (approved):

PR #7698, opened 2025-02-19
Account for standard environment variable prefixes when simulating launch profile

ReubenBond:
## Description

Environment variables prefixed with `DOTNET_` or `ASPNETCORE_` should be propagated into configuration in non-prefixed form. For example, setting `DOTNET_ENVIRONMENT=Testing` via an environment variable should set `ENVIRONMENT=Testing` in configuration.

In the Aspire.Hosting.Testing package, we simulate loading the app host's launch profile, including environment variables. We do not set environment variables in the process since that would leak between concurrently running tests. Instead, we read the environment variables from the launch profile into configuration. This PR changes how we propagate the environment variables into configuration so that we check for those two well-known prefixes and flow them into configuration without those prefixes with the precedence order of _no prefix_ > `DOTNET_` > `ASPNETCORE_`.

In particular, this allows setting the environment name from launch profile to work as expected.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
+++ b/src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
@@ -162,7 +162,6 @@ public class DistributedApplicationFactory(Type entryPoint, string[] args) : IDi
         };
         applicationOptions.Args = hostBuilderOptions.Args;
 
-        hostBuilderOptions.EnvironmentName = Environments.Development;
         hostBuilderOptions.ApplicationName = entryPointAssembly.GetName().Name ?? string.Empty;
         applicationOptions.AssemblyName = entryPointAssembly.GetName().Name ?? string.Empty;
         applicationOptions.DisableDashboard = true;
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -264,6 +263,18 @@ private static void PostConfigureBuilderOptions(
                 foreach (var (key, value) in envVars)
                 {
                     SetDefault(key, value);
+
+                    // See https://github.com/dotnet/runtime/blob/8edaf7460777e791b6279b395a68a77533db2d20/src/libraries/Microsoft.Extensions.Hosting/src/HostApplicationBuilder.cs#L96
```
eerhardt, line 267: Why aren't we just calling `configuration.AddEnvironmentVariables(prefix: "DOTNET_");` and `configuration.AddEnvironmentVariables(prefix: "ASPNETCORE_");` ?

eerhardt: Oh, because these env vars come from the launch profile, and aren't actually set in the environment variables.

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -264,6 +263,18 @@ private static void PostConfigureBuilderOptions(
...
+                    if (key.StartsWith("DOTNET_"))
```
eerhardt, line 268:
```suggestion
                    if (key.StartsWith("DOTNET_", StringComparison.OrdinalIgnoreCase))
```

src/Aspire.Hosting.Testing/DistributedApplicationFactory.cs
```diff
@@ -264,6 +263,18 @@ private static void PostConfigureBuilderOptions(
...
+                    {
+                        SetDefault(key["DOTNET_".Length..], value);
+                    }
+
+                    // See https://github.com/dotnet/aspnetcore/blob/4ce2db7b8d85c07cad2c59242edc19af6a91b0d7/src/DefaultBuilder/src/WebApplicationBuilder.cs#L38
+                    if (key.StartsWith("ASPNETCORE_"))
```
eerhardt, line 274:
```suggestion
                    if (key.StartsWith("ASPNETCORE_", StringComparison.OrdinalIgnoreCase))
```

eerhardt (approved):
LGTM

sebastienros:
Tests were actually failing there too

image

PR #7699, opened 2025-02-20
Fix Azure PostgreSQL AsExisting

### LINKED ISSUE #7694: Azure PostgreSQL AsExisting doesn't deploy successfully

### Describe the bug

When using AddAzurePostgresFlexibleServer and AsExisting, the bicep we generate doesn't work. You get an error on startup:

```
pg.module.bicep(12,3) : Error BCP173: The property "properties" cannot be used in an existing resource declaration. [https://aka.ms/bicep/core-diagnostics#BCP173]
```

### Expected Behavior

I expect to use an existing Azure PostgreSQL server in my app.

### Steps To Reproduce

```C#
var serverName = "myserver";
var rg = "myresourcegroup";

var db = builder.AddAzurePostgresFlexibleServer("pg")
                .RunAsExisting(serverName , rg)
                .AddDatabase("db");
```

When you F5 you get the above error when it tries provisioning.

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

cc @captainsafia

eerhardt:
## Description

In bicep, it is not supported to set properties on "existing" resources. When using AsExisting on an Azure PostgreSQL resource, we are setting authConfig properties to enable/disable password and activeDirectory auth. This is causing errors when trying to deploy because the bicep is invalid.

Fix this by not setting these properties on existing resources. Instead for activeDirectory/Entra ID auth, we will add the principle as an admin on the server. For password auth on an existing resource, we expect the user to pass the correct username/password parameters to connect to the database server.

Fix #7694

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
+++ b/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
@@ -394,14 +394,19 @@ public static class AzurePostgresExtensions
             keyVault.Name = kvNameParam;
             infrastructure.Add(keyVault);
 
-            postgres.AuthConfig = new PostgreSqlFlexibleServerAuthConfig()
+            // bicep doesn't allow for setting properties on existing resources. So we don't set auth properties here.
+            // The administratorLogin and administratorLoginPassword are expected to match what is already configured on the server
+            if (!postgres.IsExistingResource)
             {
-                ActiveDirectoryAuth = PostgreSqlFlexibleServerActiveDirectoryAuthEnum.Disabled,
-                PasswordAuth = PostgreSqlFlexibleServerPasswordAuthEnum.Enabled
-            };
+                postgres.AuthConfig = new PostgreSqlFlexibleServerAuthConfig()
+                {
+                    ActiveDirectoryAuth = PostgreSqlFlexibleServerActiveDirectoryAuthEnum.Disabled,
+                    PasswordAuth = PostgreSqlFlexibleServerPasswordAuthEnum.Enabled
+                };
 
-            postgres.AdministratorLogin = administratorLogin;
-            postgres.AdministratorLoginPassword = administratorLoginPassword;
+                postgres.AdministratorLogin = administratorLogin;
+                postgres.AdministratorLoginPassword = administratorLoginPassword;
+            }
 
             var secret = new KeyVaultSecret("connectionString")
             {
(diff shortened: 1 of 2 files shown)


captainsafia (approved):

src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs
```diff
@@ -430,11 +435,14 @@ private static void ConfigurePostgreSqlInfrastructure(AzureResourceInfrastructur
...
-            postgres.AuthConfig = new PostgreSqlFlexibleServerAuthConfig()
+            if (!postgres.IsExistingResource)
             {
-                ActiveDirectoryAuth = PostgreSqlFlexibleServerActiveDirectoryAuthEnum.Enabled,
-                PasswordAuth = PostgreSqlFlexibleServerPasswordAuthEnum.Disabled
-            };
+                postgres.AuthConfig = new PostgreSqlFlexibleServerAuthConfig()
+                {
+                    ActiveDirectoryAuth = PostgreSqlFlexibleServerActiveDirectoryAuthEnum.Enabled,
+                    PasswordAuth = PostgreSqlFlexibleServerPasswordAuthEnum.Disabled
```
mitchdenny, line 443: I think disabling password auth there is a potential issue. Lets say you have an existing Postgres instance which is setup with production data and there are multiple apps hanging off it. Then our deployment comes along and turns off password auth.

We can't assume that is the right thing to do for resources that we don't own.

captainsafia: From my understanding, with these code changes, we're not touching the authentication configuration at all if it's an existing resource. If it's configured with password auth and you try to use a managed identity via your app host config, you'll get an error at the provisioning step at the moment.

The current implementations puts the onus on the developer to  make sure that the auth setup between the the existing resource and the app host config lines up.

Let me know if I missed something here.

davidfowl: That’s an interesting problem though. We have to let people set the type of auth they want (and maybe try to detect if that’s disabled?)

eerhardt: > with these code changes, we're not touching the authentication configuration at all if it's an existing resource.

This is correct. We aren't changing the auth configuration on an existing resource, after these changes. The expectation is as @captainsafia states - if you have an existing server, you need to try to connect to it using the same auth as what is already enabled. If only username/password, you need to have that username/password. If only Entra ID is enabled, you need to use Entra ID.

The one thing we will do is for Entra ID we will add the current app / user to the database.

> We have to let people set the type of auth they want 

This is the intention.

> (and maybe try to detect if that’s disabled?)

We aren't trying to detect yet. But if there is an enhancement we can make in the future, we can try.

PR #7701, opened 2025-02-20
[CI] re-enable windows non-helix tests on public pipeline rolling builds

radical:
Issue: `##[error]Artifact Windows_CodeCoverageResults was not found for build ....`

The underlying problem is that pipeline (non-helix) tests are not being run on Windows at all, since the recent pipeline changes. This was detected now due to the above error because all other tests are passing now, and the CodeCoverage step is run which failed because no codecoverage results were found for windows.

### DIFF
--- a/eng/pipelines/azure-pipelines-public.yml
+++ b/eng/pipelines/azure-pipelines-public.yml
@@ -135,6 +135,7 @@ stages:
                 repoTestResultsPath: $(Build.Arcade.TestResultsPath)
                 isWindows: true
                 runHelixTests: ${{ contains(testVariant, 'helix') }}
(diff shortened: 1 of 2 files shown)


radical:
CI run - https://dev.azure.com/dnceng-public/public/_build/results?buildId=957257&view=results

radical:
Another run after adding a missing `RequiresDocker` on a test: https://dev.azure.com/dnceng-public/public/_build/results?buildId=957283&view=results

danmoseley (approved):
let's see if it works

radical:
ugh.. something else to check. Don't merge yet :|

PR #7711, opened 2025-02-20
Fix arg annotations when running in IDE mode

JamesNK:
## Description

Apply behavior from https://github.com/dotnet/aspire/blob/main/docs/specs/IDE-execution.md#launch-profile-processing-project-launch-configuration when displaying args in the dashboard.

Two changes:
1. When projects are executed by the IDE, the launch args and app host args aren't merged together. Launch args are only displayed if there are no app host args.
2. If there are no args then `project.Spec.Args` is left as null. That wasn't happening but appears to have special behavior according to the [IDE execution spec](https://github.com/dotnet/aspire/blob/main/docs/specs/IDE-execution.md#launch-profile-processing-project-launch-configuration). Maybe something along the toolchain automatically converts an empty array to null?

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Dcp/DcpExecutor.cs
+++ b/src/Aspire.Hosting/Dcp/DcpExecutor.cs
@@ -966,31 +966,43 @@ internal sealed class DcpExecutor : IDcpExecutor, IAsyncDisposable
         }
         var spec = exe.Spec;
 
+        spec.Args = null;
+
         // An executable can be restarted so args must be reset to an empty state.
         // After resetting, first apply any dotnet project related args, e.g. configuration, and then add args from the model resource.
-        spec.Args = [];
-        if (er.DcpResource.TryGetAnnotationAsObjectList<string>(CustomResource.ResourceProjectArgsAnnotation, out var projectArgs))
+        if (er.DcpResource.TryGetAnnotationAsObjectList<string>(CustomResource.ResourceProjectArgsAnnotation, out var projectArgs) && projectArgs.Count > 0)
         {
+            spec.Args ??= [];
             spec.Args.AddRange(projectArgs);
         }
+
+        (var args, var failedToApplyArgs) = await BuildArgsAsync(resourceLogger, er.ModelResource, cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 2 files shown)


JamesNK:
@captainsafia I think this might remove the need for https://github.com/dotnet/aspire/pull/7683

captainsafia (commented):
Not an expert but LGTM overall -- it does appear that the double `--port` fix can be reverted with this change set.

src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -966,31 +966,49 @@ private async Task CreateExecutableAsync(AppResource er, ILogger resourceLogger,
...
 
-            var launchProfileArgs = GetLaunchProfileArgs(project.GetEffectiveLaunchProfile()?.LaunchProfile);
-            launchArgs.AddRange(launchProfileArgs.Select(a => (a, isSensitive: false, annotationOnly)));
+                var launchProfileArgs = GetLaunchProfileArgs(project.GetEffectiveLaunchProfile()?.LaunchProfile);
+                launchArgs.AddRange(launchProfileArgs.Select(a => (a, isSensitive: false, annotationOnly)));
+            }
         }
 
-        (var args, var failedToApplyArgs) = await BuildArgsAsync(resourceLogger, er.ModelResource, cancellationToken).ConfigureAwait(false);
         launchArgs.AddRange(args.Select(a => (a.Value, a.IsSensitive, annotationOnly: false)));
```
adamint, line 1003: I know this won't actually do anything if line 992 is true, and so the effective launch args will be correct, but did a double take - could a comment or a condition be added to clarify?

adamint (approved):
just that one question

karolz-ms (approved):
nice

PR #7716, opened 2025-02-20
Resolve DistributedApplicationResourceBuilder<ResourceWithConnectionStringSurrogate> correctly in args evaluation

adamint:
## Description

ResourceWithConnectionStringSurrogate is wrapped in its builder when we are processing args, and so we should evaluate the underlying resource.

The right fix might be to do that for all DistributedApplicationResourceBuilder<T>, but I'm not confident that would be correct. I can't find other cases where a builder is passed instead of the resource, does anyone else know? @mitchdenny for thoughts? keeping this as a draft during discussion

Fixes https://github.com/dotnet/aspire/pull/7662#issuecomment-2670994303

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [X] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

### DIFF
--- a/tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
+++ b/tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
@@ -263,17 +263,29 @@ public class ResourceExtensionsTests
     public async Task GetArgumentValuesAsync_ReturnsCorrectValuesForSpecialCases()
     {
         var builder = DistributedApplication.CreateBuilder();
-        var surrogate = new ResourceWithConnectionStringSurrogate("ResourceWithConnectionStringSurrogate", _ => "ConnectionString", null);
-        var secretParameter = new ParameterResource("SecretParameter", _ => "SecretParameter", true);
-        var nonSecretParameter = new ParameterResource("NonSecretParameter", _ => "NonSecretParameter");
+        var surrogate = builder.AddResource(new ResourceWithConnectionStringSurrogate("ResourceWithConnectionStringSurrogate", _ => "ConnectionString", null));
+        var secretParameter = builder.AddResource(new ParameterResource("SecretParameter", _ => "SecretParameter", true));
+        var nonSecretParameter = builder.AddResource(new ParameterResource("NonSecretParameter", _ => "NonSecretParameter"));
 
-        var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
-            .WithArgs(builder.AddResource(surrogate))
-            .WithArgs(builder.AddResource(secretParameter))
-            .WithArgs(builder.AddResource(nonSecretParameter));
-        var args = await container.Resource.GetArgumentValuesAsync().DefaultTimeout();
+        var containerArgs = await builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
+            .WithArgs(surrogate)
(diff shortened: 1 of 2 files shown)


mitchdenny:
So I had a look at .NET Aspire 9.0 and this problem still existed there so this isn't a regression. But it is a gap that we need to fill. I think your solution is fine, but it needs some test cases to make sure that we don't accidentally regress this behavior.

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -259,6 +259,27 @@ public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesUsingMa
...
+        var surrogate = new ResourceWithConnectionStringSurrogate("ResourceWithConnectionStringSurrogate", _ => "ConnectionString", null);
+        var secretParameter = new ParameterResource("SecretParameter", _ => "SecretParameter", true);
+        var nonSecretParameter = new ParameterResource("NonSecretParameter", _ => "NonSecretParameter");
+
+        builder.AddResource(surrogate);
+        builder.AddResource(secretParameter);
+        builder.AddResource(nonSecretParameter);
+
+        var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
+            .WithArgs(surrogate)
```
mitchdenny, line 275: We should be passing in the IResourceBuilder<T> here instead of the actual resource itself.

adamint: yeah you're right, sorry

tests/Aspire.Hosting.Tests/ResourceExtensionsTests.cs
```diff
@@ -259,6 +259,27 @@ public async Task GetEnvironmentVariableValuesAsyncReturnCorrectVariablesUsingMa
...
+        var builder = DistributedApplication.CreateBuilder();
+        var surrogate = new ResourceWithConnectionStringSurrogate("ResourceWithConnectionStringSurrogate", _ => "ConnectionString", null);
+        var secretParameter = new ParameterResource("SecretParameter", _ => "SecretParameter", true);
+        var nonSecretParameter = new ParameterResource("NonSecretParameter", _ => "NonSecretParameter");
+
+        builder.AddResource(surrogate);
+        builder.AddResource(secretParameter);
+        builder.AddResource(nonSecretParameter);
+
+        var container = builder.AddContainer("elasticsearch", "library/elasticsearch", "8.14.0")
```
mitchdenny, line 270: The matrix we need to test is:

- ```AddProject(...).WithArgs(connectionStringBuilder);```
- ```AddExecutable(..., connectionStringBuilder);```
- ```AddExecutable(...).WithArgs(context => { context.Args.Add(connectionStringBuilder) });
- ```AddContainer(...).WithArgs(context => { context.Args.Add(connectionStringbuilder) });

mitchdenny (approved):
LGTM. Thanks for jumping on this @adamint

PR #7723, opened 2025-02-20
Add policy to close & re-open github-action PRs

### LINKED ISSUE #7719: Tests don't run on PRs opened by the backport bot
This is because our tests use github actions, and, by-design, github actions won't be triggered on PRs that were created via github action (to prevent recursive workflows): https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

We should figure out a way around this

wtgodbe:
Attempt to fix https://github.com/dotnet/aspire/issues/7719

Add a policy that closes & re-opens PRs opened via github action, in order to trigger actions on those PRs. 

Don't merge yet, need to make sure this doesn't infinitely loop

### DIFF
--- a/.github/policies/eventResponderTask.GithubActionPROpened.yml
+++ b/.github/policies/eventResponderTask.GithubActionPROpened.yml
@@ -20,14 +20,14 @@ configuration:
           label: 'Closed Github-Action PR'
       - closeIssue
       description: Close and label Github-Action PRs
-      - if:
-        - payloadType: Pull_Request
-        - isAction:
-            action: Closed
-        - hasLabel:
-            label: 'Closed Github-Action PR'
-        then:
-        - removeLabel:
-            label: 'Closed Github-Action PR'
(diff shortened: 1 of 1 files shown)


.github/policies/eventResponderTask.GithubActionPROpened.yml
```diff
@@ -0,0 +1,38 @@
+id: eventResponderTask.GithubActionPROpened
```
eerhardt, line 1: It would be nice to add some comments in here describing why it is doing what it is doing.

wtgodbe: Done

mitchdenny:
Wow, I can't believe we need this hack!

mitchdenny (approved):

davidfowl:
What's this waiting on/

Issue #7653, opened 2025-02-18
Run command args should completeley override launch profile args when set


vmykagapuz:
**REGRESSION INFO**

1. Worked in Aspire 9.0.0

**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2
2. Install VS 17.14 P1
3. Install Aspire 9.1.0 GA (Apply Feed)
4. Install Docker/Podman

**NOTE**
1. This issue reproduces for both 9.0/8.0 framework.

**REPRO STEPS**
1. Create .NET Aspire Starter App > .NET 9.0 (Standard Term Support) > Aspire 9.1 > Create
2. Right click solution > New project > Azure functions > FunctionApp1 > .NET 9.0 Isolated > Http trigger > Check the "Enlist in Aspire orchestrations > Create.
3. Build and run.

ACTUAL
The "--port 7235" is duplicated
![Image](https://github.com/user-attachments/assets/a4af25eb-f62b-4b49-9243-cab9559405ae)

EXPECTED
The text should not be duplicated.

eerhardt:
cc @captainsafia

davidfowl:
Also @adamint

davidfowl:
This should be in 9.1 @danmoseley

davidfowl:
I think this is a bug the app model code we added. If arguments are specified then the launch profile is ignored and the dashboard should reflect that.

@JamesNK this is related to your change

JamesNK:
It's definitely related to displaying command line args from launchSettings.json. Could it be that the port is just being passed in twice? Once in the launchSettings.json and again as an arg configured in Aspire.

JamesNK:
I tested in the AzureFunctionsEndToEnd sample.

As reported, the functions app has double ports:

![Image](https://github.com/user-attachments/assets/a0471c75-1248-4380-b035-a6fc195d21cb)

But if you add `WithAnnotation(new ExcludeLaunchProfileAnnotation())` to disable the launch profile:

```cs
var funcApp = builder.AddAzureFunctionsProject<Projects.AzureFunctionsEndToEnd_Functions>("funcapp")
    .WithExternalHttpEndpoints()
    .WithReference(eventHub).WaitFor(eventHub)
#if !SKIP_UNSTABLE_EMULATORS
    .WithReference(serviceBus).WaitFor(serviceBus)
    .WithReference(cosmosDb).WaitFor(cosmosDb)
#endif
    .WithReference(blob)
    .WithReference(queue)
[...]

PR #7726, opened 2025-02-21
Revert "Avoid setting --port twice for Azure Functions (#7683)"

JamesNK:
This reverts commit 97d5c1c75ec99c3b068b02eafcc34e8a8db17102.

Change is no longer neccessary with https://github.com/dotnet/aspire/pull/7711

Confirmation that double args aren't shown:

image

From Azure Functions EndToEnd sample.

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -114,6 +114,17 @@ public static class AzureFunctionsProjectResourceExtensions
                 // Set the storage connection string.
                 ((IResourceWithAzureFunctionsConfig)resource.HostStorage).ApplyAzureFunctionsConfiguration(context.EnvironmentVariables, "AzureWebJobsStorage");
             })
+            .WithArgs(context =>
+            {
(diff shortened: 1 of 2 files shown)


danmoseley (approved):

JamesNK:
Note: This change isn't in `release/9.1`. No need to backport this PR.

eerhardt:
Are we sure we should be reverting this change? From @captainsafia on https://github.com/dotnet/aspire/issues/7653#issuecomment-2671883969

> Hmmm...I see. I think the change on the Functions side is still valid since we shouldn't be populating args at all if we don't need to.

JamesNK:
There is no urgency is merging this. Will wait until David, Safia have commented.

mitchdenny (approved):

captainsafia:
Let's not merge the revert. Although the DCP issue was the root cause, this was still exposing a place where we were setting the `--port` redundently.

It also happens that I need to move the `WithArgs` to the `WithFunctionsHttpEndpoint` for some launch profile-based processing for HTTPs so this is a code move that would've happened anyways.

davidfowl:
See https://github.com/dotnet/aspire/issues/7751

captainsafia:
> See #7751

Posted a comment on this. I don't think reverting my change solves this. It seems like there's a different change that is causing the `--` to be inserted here.

PR #7733, opened 2025-02-21
Require resource group on existing azure resources

eerhardt:
Since, by default, publishing a .NET Asipre app will create a new resource group, not specifying a resource group won't work - since the current resource group hasn't been created yet. Because of this, require callers to explicitly specify the resource group to use. They can still say `null` to mean "use the current" if that's really what they want.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] Yes
      - [x] No

### DIFF
--- a/src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
@@ -32,7 +32,7 @@ public static class ExistingAzureResourceExtensions
     /// <param name="nameParameter">The name of the existing resource.</param>
     /// <param name="resourceGroupParameter">The name of the existing resource group, or <see langword="null"/> to use the current resource group.</param>
     /// <returns>The resource builder with the existing resource annotation added.</returns>
-    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+    public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter)
         where T : IAzureResource
     {
         ArgumentNullException.ThrowIfNull(builder);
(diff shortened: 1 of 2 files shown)


src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
```diff
@@ -74,7 +74,7 @@ public static IResourceBuilder<T> RunAsExisting<T>(this IResourceBuilder<T> buil
     /// <param name="nameParameter">The name of the existing resource.</param>
     /// <param name="resourceGroupParameter">The name of the existing resource group, or <see langword="null"/> to use the current resource group.</param>
     /// <returns>The resource builder with the existing resource annotation added.</returns>
-    public static IResourceBuilder<T> PublishAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter = null)
+    public static IResourceBuilder<T> PublishAsExisting<T>(this IResourceBuilder<T> builder, IResourceBuilder<ParameterResource> nameParameter, IResourceBuilder<ParameterResource>? resourceGroupParameter)
```
eerhardt, line 77: Do we want to make the same change for the `RunAsExisting` methods? If someone specifies a `Azure:ResourceGroup` in their user secrets (and not a Prefix), the resource group can be existing. So it is more likely to work if you don't specify a resource group.

mitchdenny:
We can probably let this go in as is but have we considered using a fully qualified resource ID as well?

Given that scenario how would we disambiguate the APIs (since both would just be parameters).

mitchdenny:
In case it wasn't obvious the reason for my question was around using existing resources that might be in a different sub which is actually pretty common.

eerhardt:
> We can probably let this go in as is but have we considered using a fully qualified resource ID as well?

I believe we talked about that orginally, but went with "resourceName" and "resourceGroup" separately.

> Given that scenario how would we disambiguate the APIs (since both would just be parameters).

This change actually makes it easier becuase in the future we can add an overload that takes a single string which is the full resource ID. If we are going to do that, we should also make this same change to the `RunAsExisting` methods, so we can have the single string version in the future.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
eerhardt: Given @mitchdenny's comment https://github.com/dotnet/aspire/pull/7733#issuecomment-2675562736, I think we should make the same change to RunAsExisting. It gives us room to take a single string to mean the whole resourceID in the future.

captainsafia (approved):

mitchdenny:
> This change actually makes it easier becuase in the future we can add an overload that takes a single string which is the full resource ID. If we are going to do that, we should also make this same change to the `RunAsExisting` methods, so we can have the single string version in the future.

Very good point.

src/Aspire.Hosting.Azure/ExistingAzureResourceExtensions.cs
mitchdenny: Agreed

PR #7700, opened 2025-02-20
Add protocol argument to WithEndpoint

sebastienros:
## Description

Adds an optional argument to set the protocol on new endpoints. Currently defaults to tcp with no way to change it but use the overload that take a lamda.

Based on [this discussion](https://github.com/dotnet/aspire/discussions/7170)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [ ] No
- Does the change make any security assumptions or guarantees?
  - [ ] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [ ] No
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
    - Is this introducing a breaking change?
      - [ ] Yes
        - Link to aspire-docs issue (please use this [`breaking-change` template](https://github.com/dotnet/docs-aspire/issues/new?template=04-breaking-change.yml)):
      - [ ] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)):
  - [x] No

PR #7735, opened 2025-02-21
Fix NodeJS Tests

eerhardt:
These tests assumed that the test environment was always "development" which is no longer the case after #7698.


radical:
> These tests assumed that the test environment was always "development" which is no longer the case after https://github.com/dotnet/aspire/pull/7700.

That PR hasn't merged yet. Did you mean to reference a different one?

ReubenBond (approved):

PR #7736, opened 2025-02-21
Add a Final Results job for branch protection

eerhardt:
To ensure all tests pass before merging, we add a single "Final Results" job after all the tests that ensures the tests pass. That way the branch protection rule can be set on this one result.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,7 +229,7 @@ jobs:
           name: testresults-${{ matrix.name }}
           path: testresults/**
 
-  results: # This job does is used for the branch protection. It ensures all the above tests passed
+  results: # This job is used for branch protection. It ensures all the above tests passed
     if: ${{ always() }}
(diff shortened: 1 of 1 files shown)


mitchdenny (approved):
LGTM. Once this is in an backported we should enable the check in the branch policy.

eerhardt:
> Once this is in an backported

9.1 is in an even worse state because of [Tests don't run on PRs opened by the backport bot (dotnet/aspire#7719)](https://github.com/dotnet/aspire/issues/7719)

radical:
This seems to be working now, but needs a re-review.

eerhardt (commented):
Since I opened this PR I can't approve it. But I'm good with the changes. Thanks, @radical for making them.

Since @mitchdenny approved above and I approved the subsequent changes, I think we are good to merge.

eerhardt:
I've updated the branch protections for main and release/* to require "Final Results" to pass before merging.

Issue #7437, opened 2025-02-06
[Test] Fix AzureFunctionsTest on GitHub Actions


ReubenBond:
In #7423, we enabled tests in GitHub Actions and disabled non-Helix Azure DevOps tests.
Unfortunately, some tests fail on GitHub Actions.

See:
* https://github.com/dotnet/aspire/actions/runs/13184837065/job/36804421373?pr=7423

Failing test:
🟥 <a href="https://github.com/dotnet/aspire/blob/9fb6bd9200c80254c649e3e0308088d9c5667065/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs#L61">AzureFunctionsTest</a>

```yml
System.Threading.Tasks.TaskCanceledException : Task was canceled before these messages were found: 'Worker process started and initialized.'
---- System.Threading.Tasks.TaskCanceledException : A task was canceled.
   at Aspire.Hosting.Tests.Utils.LoggerNotificationExtensions.WaitForAllTextAsync(DistributedApplication app, IEnumerable`1 logTexts, String resourceName, CancellationToken cancellationToken) in /_/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs:line 108
   at Aspire.Playground.Tests.ProjectSpecificTests.AzureFunctionsTest() in /_/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs:line 70
   at Aspire.Playground.Tests.ProjectSpecificTests.AzureFunctionsTest() in /_/tests/Aspire.Playground.Tests/ProjectSpecificTests.cs:line 129
--- End of stack trace from previous location ---
----- Inner Stack Trace -----
   at Aspire.Hosting.Tests.Utils.LoggerNotificationExtensions.WaitForAllTextAsync(DistributedApplication app, IEnumerable`1 logTexts, String resourceName, CancellationToken cancellationToken) in /_/tests/Aspire.Hosting.Tests/Utils/LoggerNotificationExtensions.cs:line 92
```

davidfowl:
cc @captainsafia This is tracking the issue we were investigating

captainsafia:
Here's a summary of what we've found out so far from looking at the logs related to this failure.

- The test environment does appear to be using a version of the .NET SDK that contains the `ComputeRunArguments` target we take a dependency on.
- The Functions project is built successfully in the pre-build step.
- The Functions SDK that contains the MSBuild that extends the `ComputeRunArguments` target is loaded into the test environment.
[...]

PR #7748, opened 2025-02-23
Rebuild Azure Functions project before test run

captainsafia:
Addresses https://github.com/dotnet/aspire/issues/7437 by rebuilding the Functions project again after the build of test projects to force resolution of trigger metadata correctly before the Functions test runs.

This is a workaround for https://github.com/Azure/azure-functions-dotnet-worker/issues/2969.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -187,6 +187,7 @@ jobs:
         run: |
           ./build.sh -restore -ci -build -projects ${{ github.workspace }}/${{ matrix.project }} /bl
 
+      # Workaround for bug in Azure Functions Worker SDK. See https://github.com/Azure/azure-functions-dotnet-worker/issues/2969.
       - name: Rebuild for Azure Functions project
         if: matrix.name == 'Playground'
         env:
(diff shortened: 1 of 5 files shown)


Directory.Build.props
```diff
@@ -50,4 +50,8 @@
     <AspireDashboardDir>$(MSBuildThisFileDirectory)/artifacts/bin/Aspire.Dashboard/$(Configuration)/net8.0/</AspireDashboardDir>
   </PropertyGroup>
 
+  <!-- Workaround for resolving Git packaging related targets inner Functions build. -->
+  <Target Name="GitVersion" />
+  <Target Name="GitInfo" />
```
davidfowl, line 55: Remove this and set CI=false on your step

.github/workflows/tests.yml
```diff
@@ -187,6 +187,13 @@ jobs:
         run: |
           ./build.sh -restore -ci -build -projects ${{ github.workspace }}/${{ matrix.project }} /bl
 
+      - name: Rebuild for Azure Functions project
```
eerhardt, line 190: Can we add a comment to an issue tracking this? We will want to remove this as soon as we can.

radical:
Can we add the explanation of the problem, and the workaround being used here to the PR description? I'm guessing that it would be essentially what you said in https://github.com/dotnet/aspire/issues/7437#issuecomment-2682916884 .

captainsafia:
@eerhardt @radical Added more docs here.

eerhardt (approved):
Assuming the tests pass. LGTM.

PR #7763, opened 2025-02-25
Return AspireOpenAIClientBuilder from AddOpenAIClientFromConfiguration

### LINKED ISSUE #7759: AddOpenAIClientFromConfiguration should return AspireAzureOpenAIClientBuilder

### Describe the bug

`AddOpenAIClientFromConfiguration` didn't get updated as part of [Add AddAzureOpenAIChatClient (dotnet/aspire#6225)](https://github.com/dotnet/aspire/pull/6225). It still returns `void` instead of returning an `AspireAzureOpenAIClientBuilder`.

We should update this API to conform to the same pattern used in `AddAzureOpenAIClient` and `AddOpenAIClient`.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### .NET Version info

_No response_

### Anything else?

cc @sebastienros @SteveSandersonMS

eerhardt:
## Description

This API was introduced concurrently when we introduced the AspireOpenAIClientBuilder API, so it wasn't using it.

Fix #7759

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No


SteveSandersonMS (approved):
Looks perfect. Thanks for sorting this out @eerhardt!

PR #7781, opened 2025-02-26
Optimize required unix file modes

sebastienros:
## Description

- This sets the minimal required file mode for bind mounts, and ensure the tests run with the required file mode for executing non-root containers.
- This delays the preparation of volume mounts such that these can be added in the `AfterEndpointsAllocatedEvent`.
 - Update PostgresQL extensions to use the Aspire Store to store configuration.
 - Update MySql extensions to use the Aspire Store to store configuration.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
+++ b/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
@@ -103,8 +103,6 @@ public static class MySqlBuilderExtensions
 
         containerName ??= $"{builder.Resource.Name}-phpmyadmin";
 
-        var configurationTempFileName = Path.GetTempFileName();
-
         var phpMyAdminContainer = new PhpMyAdminContainerResource(containerName);
         var phpMyAdminContainerBuilder = builder.ApplicationBuilder.AddResource(phpMyAdminContainer)
                                                 .WithImage(MySqlContainerImageTags.PhpMyAdminImage, MySqlContainerImageTags.PhpMyAdminTag)
--- a/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
+++ b/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
@@ -240,8 +240,7 @@ public static class MySqlBuilderExtensions
         // This temporary file is not used by the container, it will be copied and then deleted
         var filePath = Path.GetTempFileName();
 
-        using var stream = new FileStream(filePath, FileMode.Create);
-        using var writer = new StreamWriter(stream);
+        using var writer = new StreamWriter(filePath);
 
         writer.WriteLine("<?php");
         writer.WriteLine();
--- a/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
+++ b/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
@@ -14,6 +14,7 @@ namespace Aspire.Hosting;
 public static class MySqlBuilderExtensions
 {
     private const string PasswordEnvVarName = "MYSQL_ROOT_PASSWORD";
+    private const UnixFileMode FileMode644 = UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.GroupRead | UnixFileMode.OtherRead;
 
     /// <summary>
     /// Adds a MySQL server resource to the application model. For local development a container is used.
@@ -109,7 +110,6 @@ public static class MySqlBuilderExtensions
                                                 .WithImage(MySqlContainerImageTags.PhpMyAdminImage, MySqlContainerImageTags.PhpMyAdminTag)
                                                 .WithImageRegistry(MySqlContainerImageTags.Registry)
                                                 .WithHttpEndpoint(targetPort: 80, name: "http")
-                                                .WithBindMount(configurationTempFileName, "/etc/phpmyadmin/config.user.inc.php")
                                                 .ExcludeFromManifest();
 
         builder.ApplicationBuilder.Eventing.Subscribe<AfterEndpointsAllocatedEvent>((e, ct) =>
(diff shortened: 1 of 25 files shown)


src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -1206,6 +1188,8 @@ private async Task CreateContainerAsync(AppResource cr, ILogger resourceLogger,
             spec.Ports = BuildContainerPorts(cr);
         }
 
+        spec.VolumeMounts = BuildBindMounts(modelContainerResource);
```
sebastienros, line 1191: The idea is to be able to add container mount annotations during the `AftertEndpointAllocated`. The same way endpoints are built in `CreateContainerAsync`. Otherwise we can't create file mounts with deterministic names based on the file content (for persistent containers).

NB: In the case of MySql/PgSql the target ports are currently used, which are static. Resolution is done thought host name. This means that at least this part of the configuration in invariant across restarts.

eerhardt (commented):
Can we write a persistence container test for MySQL and PostgreSQL?

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -149,25 +149,41 @@ public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
...
+
+                    Directory.CreateDirectory(bindMountPath);
+                    Directory.CreateDirectory($"{bindMountPath}/data");
+                    Directory.CreateDirectory($"{bindMountPath}/log");
+                    Directory.CreateDirectory($"{bindMountPath}/secrets");
+
+                    File.SetUnixFileMode(bindMountPath, BindMountPermissions);
+                    File.SetUnixFileMode($"{bindMountPath}/data", BindMountPermissions);
+                    File.SetUnixFileMode($"{bindMountPath}/log", BindMountPermissions);
+                    File.SetUnixFileMode($"{bindMountPath}/secrets", BindMountPermissions);
```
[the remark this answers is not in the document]
eerhardt: Oh, I see why this doesn't work - its because the umask will ensure the folder isn't writeable by others when it was created. So you need to create the directory, and then increase the permissions in 2 steps.

We should clean this code up - probably using the old structure of always creating the directory, and then on non-windows increasing the permissions.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -235,4 +234,39 @@ public static IResourceBuilder<MySqlServerResource> WithInitBindMount(this IReso
...
+        using var stream = new FileStream(filePath, FileMode.Create);
+        using var writer = new StreamWriter(stream);
+
+        writer.WriteLine("<?php");
+        writer.WriteLine();
+        writer.WriteLine("$i = 0;");
+        writer.WriteLine();
+        foreach (var mySqlInstance in mySqlInstances)
+        {
+            if (mySqlInstance.PrimaryEndpoint.IsAllocated)
```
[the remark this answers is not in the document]
sebastienros: I don't think it is, but I didn't want to remove it during the refactoring in case there was another reason.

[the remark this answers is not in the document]
eerhardt: ok. let's remove it separately then in a follow up change.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -112,11 +112,6 @@ public static IResourceBuilder<SqlServerServerResource> WithDataBindMount(this I
         {
             var path = Path.Combine(source, dir);
 
-            if (!Directory.Exists(path))
```
eerhardt, line 115: I'm not sure if this warrants a breaking change, but we should add docs to tell users what they need to do to make this work on Linux.

src/Aspire.Hosting/Dcp/DcpExecutor.cs
```diff
@@ -1646,4 +1630,27 @@ private static List<ContainerPortSpec> BuildContainerPorts(AppResource cr)
 
         return ports;
     }
+
+    private static List<VolumeMount> BuildBindMounts(IResource container)
```
davidfowl, line 1634:
```suggestion
    private static List<VolumeMount> BuildContainerMounts(IResource container)
```

eerhardt (approved):
This change looks really nice. Nice work here! Thanks for the persistence to figure this all out.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -101,25 +101,15 @@ public static IResourceBuilder<SqlServerServerResource> WithDataVolume(this IRes
     /// <param name="source">The source directory on the host to mount into the container.</param>
     /// <param name="isReadOnly">A flag that indicates if this is a read-only mount.</param>
     /// <returns>The <see cref="IResourceBuilder{T}"/>.</returns>
+    /// <remarks>
+    /// The container starts up as non-root and the <paramref name="source"/> directory must be readable by the user that the container runs as.
```
eerhardt, line 105:
```suggestion
    /// The container starts up as non-root and the <paramref name="source"/> directory must be readable and writeable by the user that the container runs as.
```

Issue #7150, opened 2025-01-19
AddAzureFunctionsProject always running http instead of https


eolamisan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

AddAzureFunctionsProject always starts to azure function in http.
How to make it run with https.

### Expected Behavior

Allow to run the azure function with https

### Steps To Reproduce

```
using Aspire.Hosting;

var builder = DistributedApplication.CreateBuilder(args);

builder.AddAzureFunctionsProject<Projects.api>("api");
```

### .NET Version info

.NET SDK:
 Version:           9.0.102
 Commit:            cb83cd4923
 Workload version:  9.0.100-manifests.4a54b1a6
 MSBuild version:   17.12.18+ed8c6aec5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26100
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.102\

.NET workloads installed:
 [aspire]
   Installation Source: VS 17.12.35707.178
   Manifest Version:    8.2.2/8.0.100
   Manifest Path:       C:\Program Files\dotnet\sdk-manifests\8.0.100\microsoft.net.sdk.aspire\8.2.2\WorkloadManifest.json
   Install Type:              Msi

Configured to use loose manifests when installing new manifests.

Host:
  Version:      9.0.1
  Architecture: x64
  Commit:       c8acea2262

.NET SDKs installed:
  8.0.100 [C:\Program Files\dotnet\sdk]
  9.0.102 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 7.0.20 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 9.0.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 7.0.20 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
[...]

PR #7789, opened 2025-02-26
Support processing --useHttps flag on Functions applications

captainsafia:
## Description

The Functions host supports passing a `--useHttps` flag to support binding to `https://localhost:*` instead of `http://localhost:*`. This change supports picking up this flag and mapping to a target endpoint with the correct scheme.

There's likely more work to be done here (validate what happens with custom certs, etc.) but this wires up the Aspire-specific things at least.

Conributes to https://github.com/dotnet/aspire/issues/7150

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [X] No. Follow-up changes expected.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [X] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -151,6 +151,7 @@ public static class AzureFunctionsProjectResourceExtensions
         }
         var launchProfile = builder.Resource.GetEffectiveLaunchProfile();
         int? port = null;
+        var useHttps = false;
         if (launchProfile is not null)
         {
             var commandLineArgs = CommandLineArgsParser.Parse(launchProfile.LaunchProfile.CommandLineArgs ?? string.Empty);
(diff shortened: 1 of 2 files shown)


davidfowl:
We should be supporting both http and https. Does that work with the host?

captainsafia:
> We should be supporting both http and https. Does that work with the host?

AFAIK, the Functions host doesn't support being able to listen on HTTP and HTTPS ports at the same time.

davidfowl:
I don't know if that is a blocker or not, but the tests are failing. So the idea is that when somebody opts into this they only get https. We ignore this during deployment right?

captainsafia:
> I don't know if that is a blocker or not, but the tests are failing. So the idea is that when somebody opts into this they only get https. We ignore this during deployment right?

Test failure was a test bug. This is ignored during deployment.

mitchdenny:
Would it be better to add an extension method like `UseHttps(...)` to opt into this?

mitchdenny:
Hrm maybe not. If the idiomatic thing to do for functions is to update your launch profile to use HTTPS then maybe we don't don't to require people to use a gesture in the AppModel.

captainsafia:
> Hrm maybe not. If the idiomatic thing to do for functions is to update your launch profile to use HTTPS then maybe we don't don't to require people to use a gesture in the AppModel.

Yeah, I think it'll make more sense to the typical user for now for the command line arguments users are already used to configuring to be the primary drivers of behaviors of the Functions project in Aspire.

mitchdenny (approved):

PR #7798, opened 2025-02-27
Report health checks error from CheckHealthAsync to the user

JamesNK:
## Description

A user reported that an error when creating an `IHealthCheck` was not reported to the user. Health checks for the resource would hang in a wating state forever.

`ResourceHealthCheckService` now reports when there is an error calling the underlying health check service.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
+++ b/src/Aspire.Hosting/Health/ResourceHealthCheckService.cs
@@ -57,13 +57,15 @@ internal class ResourceHealthCheckService(ILogger<ResourceHealthCheckService> lo
                             {
                                 await MonitorResourceHealthAsync(state).ConfigureAwait(false);
                             }
-                            catch (OperationCanceledException)
-                            {
-                                // Ignore cancellation errors.
-                            }
                             catch (Exception ex)
                             {
-                                logger.LogDebug(ex, "Unexpected error ended health monitoring for resource '{Resource}'.", resourceName);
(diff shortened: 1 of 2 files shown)


davidfowl:
I'm not sure this should be the default. How noisy is this? By default? Did you run some of the samples where containers take a long time to be healthy?

JamesNK:
> I'm not sure this should be the default. How noisy is this? By default? Did you run some of the samples where containers take a long time to be healthy?

I'm confused. What is noisy? Why would containers taking a long time to be healthy have any impact here.

Here are the functional changes in this PR:

* If `HealthCheckService.CheckHealthAsync` throws an error then the resource health reports are set as failed. This will only happen if there is an error creating the `IHealthCheck`, so health checks will fail straight away and probably stay failed until the developer fixes their app. This is better than the current behavior when the health checks stay in a waiting state forever.
* If there is an unexpected error that causes health monitoring loop to break then a log is written at an error level. This makes it visible compared to the previous debug level. This shouldn't happen if code is behaving correctly.

mitchdenny (approved):
LGTM

PR #7800, opened 2025-02-27
Resources graph

### LINKED ISSUE #2595: Graph view of resource relationships

JamesNK:
## Description

Fixes https://github.com/dotnet/aspire/issues/2595

Follow up tasks that won't be done in this PR:
* Add a configuration option to disable the resource graph. Useful when the dashboard resource screen is used and there aren't relationships between resources
* Save the selected view to browser storage instead of the query string. There will be external places that link to resources page and they should automatically show the last selected view

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


JamesNK:
Looking for code reviews and feedback from people trying it out.

src/Aspire.Dashboard/Model/ResourceGraph/ResourceGraphMapper.cs
```diff
@@ -0,0 +1,102 @@
...
+                resolvedNames.Add(match.Name);
+            }
+        }
+
+        var endpoint = ResourceEndpointHelpers.GetEndpoints(r, includeInternalUrls: false).FirstOrDefault();
+        var resolvedEndpointText = ResolvedEndpointText(endpoint);
+        var resourceName = ResourceViewModel.GetResourceName(r, resourcesByName);
+        var color = ColorGenerator.Instance.GetColorHexByKey(resourceName);
+
+        var icon = r.ResourceType switch
```
davidfowl, line 44: We're going to want to allow the resource to specify this.

davidfowl:
Tried this out in a codespace, maybe we should remove the urls, it looks like noise in the ui when it's a long url.

image

davidfowl:
The selection UI feels a little too close to the grid:

image

JamesNK:
> Tried this out in a codespace, maybe we should remove the urls, it looks like noise in the ui when it's a long url.

Could truncate the URL with `...`

davidfowl:
The URL doesn't feel useful to show if you can't click it. Maybe we can reconsider when we get display names.

En3Tho:
I think that urls wouldn't seem so noisy if they had a "whiter" color e.g. more recognizable and there wouldn't be a an additional circle around a service. It has this greyish color and url having the same they just conflict with each other and that circle kinda eats away all of the space you have there (e.g. everything that sticks ouside of the circle feels like a buggy ui).

I think instead of a grey circle/outline it's better to make icons larger and draw that "running"/"stopped" sign of those larger icons.
Grey outline can/should be removed
Under the icon a service name should be displayed
Under service name url may be displayed and it's color should be recognizable if you expect info about url to be useful and easily accessible

src/Aspire.Dashboard/Model/ResourceGraph/ResourceGraphMapper.cs
DamianEdwards: @glennc is exploring resource-provided icons

davidfowl (approved):

PR #7801, opened 2025-02-27
CLI bootstrap

mitchdenny:
## Description

This is just bootstrapping some CLI code into the repo. The testing is probably not the approach that we'll go with in the end but I wanted to have something. Did some work to the devcontainer so we can use it naturally in the terminal after it is built.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No


davidfowl:
Excellent! I do think we should make this tool aot compatible from the jump

davidfowl (approved):

mitchdenny:
We can, although I'm not sure what value it'll provide because it'll almost certainly need to execute the apphost in most scenarios which implies runtime being available.

eerhardt:
> We can, although I'm not sure what value it'll provide because it'll almost certainly need to execute the apphost in most scenarios which implies runtime being available.

Faster startup. Less memory. Also, it can use a different version of the runtime/SDK than what is needed to execute the apphost. They won't need to align - like we do with the Dashboard. No need to "allow roll forward".

eerhardt:
But it would mean that we would need to complicate our builds a bit. For example, if we want to build for macOS, we will need to do it on macOS. Same for linux.

PR #7820, opened 2025-02-28
Add some persistent lifetime tests

sebastienros:
## Description

Adds persistent lifetime tests for MySql and PostgreSql

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
+++ b/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
@@ -123,18 +123,15 @@ public static class MySqlBuilderExtensions
             if (mySqlInstances.Count() == 1)
             {
                 var singleInstance = mySqlInstances.Single();
-                if (singleInstance.PrimaryEndpoint.IsAllocated)
+                var endpoint = singleInstance.PrimaryEndpoint;
+                phpMyAdminContainerBuilder.WithEnvironment(context =>
                 {
-                    var endpoint = singleInstance.PrimaryEndpoint;
-                    phpMyAdminContainerBuilder.WithEnvironment(context =>
-                    {
-                        // PhpMyAdmin assumes MySql is being accessed over a default Aspire container network and hardcodes the resource address
-                        // This will need to be refactored once updated service discovery APIs are available
-                        context.EnvironmentVariables.Add("PMA_HOST", $"{endpoint.Resource.Name}:{endpoint.TargetPort}");
-                        context.EnvironmentVariables.Add("PMA_USER", "root");
-                        context.EnvironmentVariables.Add("PMA_PASSWORD", singleInstance.PasswordParameter.Value);
-                    });
-                }
+                    // PhpMyAdmin assumes MySql is being accessed over a default Aspire container network and hardcodes the resource address
+                    // This will need to be refactored once updated service discovery APIs are available
+                    context.EnvironmentVariables.Add("PMA_HOST", $"{endpoint.Resource.Name}:{endpoint.TargetPort}");
+                    context.EnvironmentVariables.Add("PMA_USER", "root");
+                    context.EnvironmentVariables.Add("PMA_PASSWORD", singleInstance.PasswordParameter.Value);
+                });
             }
             else
             {
--- a/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
+++ b/src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
@@ -248,20 +248,17 @@ public static class MySqlBuilderExtensions
         writer.WriteLine();
         foreach (var mySqlInstance in mySqlInstances)
         {
-            if (mySqlInstance.PrimaryEndpoint.IsAllocated)
-            {
-                var endpoint = mySqlInstance.PrimaryEndpoint;
-                writer.WriteLine("$i++;");
-                // PhpMyAdmin assumes MySql is being accessed over a default Aspire container network and hardcodes the resource address
-                // This will need to be refactored once updated service discovery APIs are available
-                writer.WriteLine($"$cfg['Servers'][$i]['host'] = '{endpoint.Resource.Name}:{endpoint.TargetPort}';");
-                writer.WriteLine($"$cfg['Servers'][$i]['verbose'] = '{mySqlInstance.Name}';");
-                writer.WriteLine($"$cfg['Servers'][$i]['auth_type'] = 'cookie';");
-                writer.WriteLine($"$cfg['Servers'][$i]['user'] = 'root';");
-                writer.WriteLine($"$cfg['Servers'][$i]['password'] = '{mySqlInstance.PasswordParameter.Value}';");
-                writer.WriteLine($"$cfg['Servers'][$i]['AllowNoPassword'] = true;");
-                writer.WriteLine();
-            }
+            var endpoint = mySqlInstance.PrimaryEndpoint;
+            writer.WriteLine("$i++;");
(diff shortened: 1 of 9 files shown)


davidfowl (approved):

tests/Aspire.Hosting.Tests/Utils/TestDistributedApplicationBuilder.cs
```diff
@@ -52,6 +54,9 @@ private static IDistributedApplicationTestingBuilder CreateCore(string[] args, A
     {
         var builder = DistributedApplicationTestingBuilder.Create(args, (applicationOptions, hostBuilderOptions) => configureOptions?.Invoke(applicationOptions));
 
+        // Load the user secrets file for Aspire.Hosting.Tests
+        builder.Configuration.AddUserSecrets<AspireStoreTests>();
```
eerhardt, line 58: Why is this necessary? I don't see any reasoning in the comments or in the PR description.

eerhardt: If this really is necessary, it shouldn't reference `AspireStoreTests`. But instead:

```suggestion
        builder.Configuration.AddUserSecrets<TestDistributedApplicationBuilder>();
```

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
```diff
@@ -248,20 +248,17 @@ private static string WritePhpMyAdminConfiguration(IEnumerable<MySqlServerResour
         writer.WriteLine();
         foreach (var mySqlInstance in mySqlInstances)
         {
-            if (mySqlInstance.PrimaryEndpoint.IsAllocated)
```
eerhardt, line 251: Can we make the same change elsewhere? Line 126 above. And in PostgreSQL and Redis?

tests/Aspire.Hosting.MySql.Tests/MySqlFunctionalTests.cs
```diff
@@ -442,4 +442,53 @@ await pipeline.ExecuteAsync(async token =>
...
     }
+
+    [Fact]
+    [RequiresDocker]
+    public async Task MySql_WithPersistentLifetime_ReusesContainers()
+    {
+        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
+
+        using var builder1 = TestDistributedApplicationBuilder.Create(testOutputHelper);
+        var resource1 = builder1.AddMySql("resource").WithLifetime(ContainerLifetime.Persistent);
```
eerhardt, line 453: We should also add WithPhpMyAdmin. And simiarly WithPgWeb and WithPgAdmin for Postgres. Ensure all the containers are reused across runs.

tests/Aspire.Hosting.Tests/Utils/TestDistributedApplicationBuilder.cs
eerhardt: > The reason is that passwords need to be reused across restarts for persistent containers snapshots to be deterministic. 

We do that in other tests by getting the password from the first resource and flowing it into the 2nd resource.

https://github.com/dotnet/aspire/blob/c36342ccdf439830b26389379454fb9f2184bae8/tests/Aspire.Hosting.PostgreSQL.Tests/PostgresFunctionalTests.cs#L205-L209

I don't want the tests mucking with user secrets on the file system. Especially in my User directory. There could be issues with persisting this information in a non-TEMP folder. We want the tests to be "clean" and not mutate the machine state.

sebastienros: Ok, but it's fine only because the store is unit tested in other places and we'll assume it works fine so this test doesn't have to use this component. It does make it less of a functional test though.

Do the playground tests follow the same principle though?

tests/Aspire.Hosting.MySql.Tests/MySqlFunctionalTests.cs
sebastienros: Done, also added the variants that trigger different config types.

tests/Aspire.Hosting.MySql.Tests/MySqlFunctionalTests.cs
```diff
@@ -442,4 +442,53 @@ await pipeline.ExecuteAsync(async token =>
...
         }, cts.Token);
     }
+
+    [Fact]
+    [RequiresDocker]
+    public async Task MySql_WithPersistentLifetime_ReusesContainers()
+    {
+        var cts = new CancellationTokenSource(TimeSpan.FromMinutes(10));
+
+        using var builder1 = TestDistributedApplicationBuilder.Create(testOutputHelper);
```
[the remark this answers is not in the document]
sebastienros: Added it back, making tests faster, not better. Also created a helper to hide the implementation details.

src/Aspire.Hosting.MySql/MySqlBuilderExtensions.cs
sebastienros: I'll keep Redis and the test for another PR.

tests/Aspire.Hosting.Tests/Utils/TestDistributedApplicationBuilder.cs
sebastienros: Done. One thing I fixed though is that the aspire store path needs to be constant between runs, since it's defined randomly with each builder.

tests/Shared/DistributedApplicationTestingBuilderExtensions.cs
```diff
@@ -24,12 +24,21 @@ public static IDistributedApplicationTestingBuilder WithTestAndResourceLogging(t
...
 
-        builder.Configuration["Aspire:Store:Path"] = Directory.CreateTempSubdirectory().FullName;
+        builder.Configuration["Aspire:Store:Path"] = path ?? Directory.CreateTempSubdirectory().FullName;
+        return builder;
+    }
+
+    public static IDistributedApplicationTestingBuilder WithResourceCleanUp(this IDistributedApplicationTestingBuilder builder, bool? resourceCleanup = null)
+    {
+        // We create the Aspire Store in a folder with user-only access. This way non-root containers won't be able
+        // to access the files unless they correctly assign the required permissions for the container to work.
```
eerhardt, line 39: This seems like it should be updated.

tests/Aspire.Hosting.MySql.Tests/MySqlFunctionalTests.cs
```diff
@@ -442,4 +442,79 @@ await pipeline.ExecuteAsync(async token =>
             Assert.Equal("BatMobile", cars[0].Brand);
         }, cts.Token);
     }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    [RequiresDocker]
+    public async Task MySql_WithPersistentLifetime_ReusesContainers(bool useMultipleInstances)
```
eerhardt, line 450: it may help to add a blurb about why we are testing multiple instances. It took me a second to understand why.

eerhardt (approved):
Just 2 minor comments. This looks really clean. Thanks!

PR #7826, opened 2025-02-28
Enhance resource management for connection strings

### LINKED ISSUE #2964: Support appending custom connection string options to some resources
Today connection strings are owned by the resource type but sometimes they need to be appended to set other options (MARS in sql for example). We should allow apps to append to the connection string expression in WithReference or at the resource level (where it makes sense).

Today this is doable with manual logic but it might need to be more first class.

### LINKED ISSUE #7641: Support specifying a ReferenceExpression to AddConnectionString
This enables the creation of a "Just in time" resource with a custom connection string with slots that have to be filled in from the outside.

```C#

var endpoint = builder.AddParameter("endpoint");
var secret = builder.AddParameter("key", secret: true);

builder.AddOpenAI("oai", endpoint, secret);

builder.Build().Run();

static class ConnectionStringExtensions
{
    public static IResourceBuilder<IResourceWithConnectionString> AddOpenAI(
        this IDistributedApplicationBuilder builder,
        string name,
        IResourceBuilder<ParameterResource> endpoint,
        IResourceBuilder<ParameterResource> key)
    {
        var csb = new ReferenceExpressionBuilder();
        csb.Append($"Endpoint={endpoint.Resource};");
        csb.Append($"Key={key.Resource}");
        return builder.AddConnectionString(name, csb.Build());
    }
}
```

### LINKED ISSUE #7642: Add an interface to allow resources to opt out of WaitFor
Follow up to https://github.com/dotnet/aspire/pull/7314, I'm running into this building custom resources that are just data and have no lifecycle. We should introduce an interface that allows a resource to opt out of being waited on.

davidfowl:
## Description

Introduces the `IResourceWithoutLifetime` interface for resources without a lifetime, such as parameters and connection strings. Updates `ParameterResource` to implement this interface and adds the `ConnectionStringParameterResource` class for better management of connection string parameters.

Implements the `AddConnectionString` method in `ConnectionStringBuilderExtensions` and updates `ApplicationOrchestrator` to handle resources without a lifetime correctly.

Tests have been added and modified to verify the new functionality, and `Program.cs` has been updated to demonstrate the new connection string capabilities, including the addition of a second connection string with a parameter value.

These changes improve the overall resource management in the application, particularly for connection strings and parameters.

Fixes #7641
Fixes #7642
Fixes #2964

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [x] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

### DIFF
--- a/playground/ParameterEndToEnd/ParameterEndToEnd.AppHost/Program.cs
+++ b/playground/ParameterEndToEnd/ParameterEndToEnd.AppHost/Program.cs
@@ -21,9 +21,12 @@ var db = builder.AddSqlServer("sql")
 
 var insertionrows = builder.AddParameter("insertionrows");
 
+var cs = builder.AddConnectionString("cs", $"sql={db};rows={insertionrows}");
+
 builder.AddProject<Projects.ParameterEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
        .WithEnvironment("InsertionRows", insertionrows)
+       .WithReference(cs)
        .WithReference(db);
 
 #if !SKIP_DASHBOARD_REFERENCE
--- a/src/Aspire.Hosting/ApplicationModel/ParameterResource.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ParameterResource.cs
@@ -65,7 +65,8 @@ public class ParameterResource : Resource, IResourceWithoutLifetime, IManifestEx
     public string ValueExpression => $"{{{Name}.value}}";
 
     /// <summary>
-    /// The configuration key for this parameter, the default is
+    /// The configuration key for this parameter. The default format is "ConnectionStrings:{Name}" if the parameter is a connection string,
+    /// otherwise it is "Parameters:{Name}".
     /// </summary>
     internal string ConfigurationKey
     {
--- a/src/Aspire.Hosting/ApplicationModel/ParameterResource.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ParameterResource.cs
@@ -6,11 +6,12 @@ namespace Aspire.Hosting.ApplicationModel;
 /// <summary>
 /// Represents a parameter resource.
 /// </summary>
-public class ParameterResource : Resource, IManifestExpressionProvider, IValueProvider
+public class ParameterResource : Resource, IResourceWithoutLifetime, IManifestExpressionProvider, IValueProvider
 {
     private string? _value;
     private bool _hasValue;
     private readonly Func<ParameterDefault?, string> _valueGetter;
+    private string? _configurationKey;
 
     /// <summary>
     /// Initializes a new instance of <see cref="ParameterResource"/>.
@@ -62,6 +63,15 @@ public class ParameterResource : Resource, IManifestExpressionProvider, IValuePr
     /// Gets the expression used in the manifest to reference the value of the parameter.
     /// </summary>
     public string ValueExpression => $"{{{Name}.value}}";
+
+    /// <summary>
+    /// The configuration key for this parameter, the default is
+    /// </summary>
+    internal string ConfigurationKey
+    {
+        get => _configurationKey ?? (IsConnectionString ? $"ConnectionStrings:{Name}" : $"Parameters:{Name}");
+        set => _configurationKey = value;
+    }
 
     ValueTask<string?> IValueProvider.GetValueAsync(CancellationToken cancellationToken) => new(Value);
 }
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -372,11 +372,10 @@ public class ResourceNotificationService : IDisposable
         var pendingDependencies = new List<Task>();
         foreach (var waitAnnotation in waitAnnotations)
         {
-            if (waitAnnotation.Resource is ParameterResource or ResourceWithConnectionStringSurrogate)
+            if (waitAnnotation.Resource is IResourceWithoutLifetime)
             {
-                // Parameters and connection string resources are inert and don't need to be waited on.
-                // If we add support for parameter resources that can be waited on, we can remove this check.
-                // As of right now, we don't support waiting on parameter resources.
+                // IResourceWithoutLifetime are inert and don't need to be waited on.
+                // If we add support for parameter resources that can be waited on.
                 continue;
             }
 
--- a/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
+++ b/src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
@@ -375,7 +375,6 @@ public class ResourceNotificationService : IDisposable
             if (waitAnnotation.Resource is IResourceWithoutLifetime)
             {
                 // IResourceWithoutLifetime are inert and don't need to be waited on.
-                // If we add support for parameter resources that can be waited on.
                 continue;
             }
 
--- a/src/Aspire.Hosting/ConnectionStringBuilderExtensions.cs
+++ b/src/Aspire.Hosting/ConnectionStringBuilderExtensions.cs
@@ -1,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+using Aspire.Hosting.ApplicationModel;
+
+namespace Aspire.Hosting;
+
+/// <summary>
+/// Provides extension methods for adding connection string resources to an application.
+/// </summary>
+public static class ConnectionStringBuilderExtensions
+{
+    /// <summary>
+    /// Adds a connection string to the distributed application a resource with the specified expression.
+    /// </summary>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="connectionStringExpression">The connection string expression.</param>
+    /// <returns></returns>
+    public static IResourceBuilder<ConnectionStringResource> AddConnectionString(this IDistributedApplicationBuilder builder, [ResourceName] string name, ReferenceExpression connectionStringExpression)
+    {
+        var cs = new ConnectionStringResource(name, connectionStringExpression);
+        return builder.AddResource(cs)
+                      .WithInitialState(new CustomResourceSnapshot
+                      {
+                          ResourceType = "ConnectionString",
+                          // TODO: We'll hide this until we come up with a sane representation of these in the dashboard
+                          State = KnownResourceStates.Hidden,
+                          Properties = []
+                      });
+    }
+
+    /// <summary>
+    /// Adds a connection string to the distributed application a resource with the specified expression.
+    /// </summary>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="connectionStringExpression">The connection string expression.</param>
+    /// <returns></returns>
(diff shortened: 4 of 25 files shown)


davidfowl:
@jamesnk and @adamint, I'd like to follow up on this work with a way to show these resources without a lifetime in the dashboard. Right now, we hide parameters and connection strings in the dashboard but that isn't ideal. There's also no way to unhide resources. 

I'd like to solve a couple of things here:
- Make sure that it is possible to see these in the dashboard
- Make sure we have a status that isn't running that means this resource is "in the right state" that isn't running.

playground/ParameterEndToEnd/ParameterEndToEnd.AppHost/Program.cs
```diff
@@ -21,9 +21,12 @@
 
 var insertionrows = builder.AddParameter("insertionrows");
 
+var cs = builder.AddConnectionString("cs", $"sql={db};rows={insertionrows}");
```
eerhardt, line 24: I like this and I don't like this at the same time. Obviously it looks nice and clean. But I wonder if this is confusing between "is this an interpolated string? or is this a ReferenceExpression?".

tests/Aspire.Hosting.Tests/ExpressionResolverTests.cs
```diff
@@ -39,7 +39,7 @@ public async Task ResolveInternalAsync_ResolvesCorrectly(ExpressionResolverTestD
         data.Add(new ExpressionResolverTestData(true, new ConnectionStringReference(new TestExpressionResolverResource("SecretParameter"), false)), null, ("SecretParameter", true));
 
         // IResourceWithConnectionString resolves differently for ResourceWithConnectionStringSurrogate (as a secret parameter)
```
eerhardt, line 41:
```suggestion
        // IResourceWithConnectionString resolves differently for ConnectionStringParameterResource (as a secret parameter)
```

eerhardt: There are a few references to `ResourceWithConnectionStringSurrogate` left in the code. Can you find and replace them?

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
```diff
@@ -140,11 +145,61 @@ static Task PublishUpdateAsync(ResourceNotificationService notificationService,
         }
     }
 
-    private async Task OnResourcesPrepared(OnResourcesPreparedContext _)
+    private async Task OnResourcesPrepared(OnResourcesPreparedContext context)
```
eerhardt, line 148: ?

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
```diff
@@ -140,11 +145,61 @@ static Task PublishUpdateAsync(ResourceNotificationService notificationService,
...
+                var value = await vp.GetValueAsync(default).ConfigureAwait(false);
+
+                await _notificationService.PublishUpdateAsync(resource, s =>
+                {
+                    return s with
+                    {
+                        Properties = [
+                            ..s.Properties,
+                            new("Value", value) { IsSensitive = resource is ParameterResource p && p.Secret }
+                        ]
```
eerhardt, line 168: Are we guaranteed this is the first time we are setting the "Value" property? If it is already there, won't we be adding 2 "Value" properties this way?

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
```diff
@@ -372,11 +372,10 @@ public async Task WaitForDependenciesAsync(IResource resource, CancellationToken
...
         foreach (var waitAnnotation in waitAnnotations)
         {
-            if (waitAnnotation.Resource is ParameterResource or ResourceWithConnectionStringSurrogate)
+            if (waitAnnotation.Resource is IResourceWithoutLifetime)
             {
-                // Parameters and connection string resources are inert and don't need to be waited on.
-                // If we add support for parameter resources that can be waited on, we can remove this check.
-                // As of right now, we don't support waiting on parameter resources.
+                // IResourceWithoutLifetime are inert and don't need to be waited on.
+                // If we add support for parameter resources that can be waited on.
```
eerhardt, line 378: If we add support.... then what?

playground/ParameterEndToEnd/ParameterEndToEnd.AppHost/Program.cs
davidfowl: We don't *need* to add this overload its a convenience.

src/Aspire.Hosting/ApplicationModel/ResourceNotificationService.cs
davidfowl: Fixed the comment

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
davidfowl: Fixed

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
davidfowl: Left over

tests/Aspire.Hosting.Tests/ExpressionResolverTests.cs
davidfowl: Resolved

src/Aspire.Hosting/ConnectionStringBuilderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+public static class ConnectionStringBuilderExtensions
+{
+    /// <summary>
+    /// Adds a connection string resource to the distributed application with the specified expression.
+    /// </summary>
+    /// <param name="builder">Distributed application builder</param>
+    /// <param name="name">The name of the resource.</param>
+    /// <param name="connectionStringExpression">The connection string expression.</param>
+    /// <returns></returns>
```
mitchdenny, line 18: Might need some more XML docs on this (can be a follow up PR).

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
```diff
@@ -145,6 +150,50 @@ private async Task OnResourcesPrepared(OnResourcesPreparedContext _)
         await PublishResourcesWithInitialStateAsync().ConfigureAwait(false);
     }
 
+    private Task ProcessResourcesWithoutLifetime(AfterEndpointsAllocatedEvent @event, CancellationToken cancellationToken)
```
mitchdenny, line 153: Should these resources raise the `ResourceReadyEvent`. There is a meta question here - its not that these resources don't have a lifetime - its more that their lifetime only has internal dependencies.

davidfowl: These resources don't have a lifetime. Are you saying we should publish the event from here or something else? ResourceReadyEvent will never fire because these are never in a running state.

mitchdenny: In the playground sample in the parameter, the connection string references the DB resource which inherits its lifecycle from its parent. I'm not seeing this as much different other than its not strictly a "child" resource. I don't think that this is a decision we need to block on though, just something we need to consider that we may want to revisit.

src/Aspire.Hosting/ApplicationModel/ParameterResource.cs
```diff
@@ -63,5 +64,15 @@ public string Value
...
 
+    /// <summary>
+    /// The configuration key for this parameter. The default format is "ConnectionStrings:{Name}" if the parameter is a connection string,
+    /// otherwise it is "Parameters:{Name}".
+    /// </summary>
+    internal string ConfigurationKey
+    {
+        get => _configurationKey ?? (IsConnectionString ? $"ConnectionStrings:{Name}" : $"Parameters:{Name}");
+        set => _configurationKey = value;
+    }
```
mitchdenny, line 75: Glad this is internal. I want to avoid strictly tying the concept of a parameter/connection string to the .NET configuration store. They are connected for convienence however.

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
mitchdenny: I stuffed up when debugging this to validate what I was saying. It DOES fire ResourceReadyEvent for SqlDatabases (as an example of a child resource).

davidfowl: That's more of a dependency that inheriting the lifecycle (like WaitFor). It does make me wonder if enabling that does make it more interesting to support WaitFor.

tests/Aspire.Hosting.Tests/AddParameterTests.cs
```diff
@@ -43,54 +42,7 @@ public void ParametersAreHiddenByDefault()
...
-            },
-            prop =>
-            {
-                Assert.Equal("Value", prop.Name);
-                Assert.Equal("pass1", prop.Value);
-            });
-    }
-
-    [Fact]
-    public void MissingParametersAreConfigurationMissing()
```
mitchdenny, line 55: Isn't this test still valid except for the hook bit at the bottom, or is it mostly redundant?

tests/Aspire.Hosting.Tests/WithReferenceTests.cs
```diff
@@ -246,6 +246,52 @@ public async Task ConnectionStringResourceWithConnectionString()
...
+
+        // Call environment variable callbacks.
+        var config = await EnvironmentVariableEvaluator.GetEnvironmentVariablesAsync(projectB.Resource, DistributedApplicationOperation.Run, TestServiceProvider.Instance).DefaultTimeout();
+
+        var servicesKeysCount = config.Keys.Count(k => k.StartsWith("ConnectionStrings__"));
+        Assert.Equal(1, servicesKeysCount);
+        Assert.Equal("Endpoint=http://localhost:3452;Key=secretKey", config["ConnectionStrings__cs"]);
+    }
+
+    [Fact]
```
mitchdenny, line 273: Add a test like we have in the ParameterEndToEnd playground where we use a resource (such as a SQL DB) as an input to the expression.

mitchdenny (approved):
A few suggestions for future enhancements and test improvements, but overall this looks good and seems to work with the run throughs that I made. I can see how this will be useful for rapidly creating wrapper APIs where the input is primarily something like an API key with a few extra bits and pieces bolted on.

tests/Aspire.Hosting.Tests/AddParameterTests.cs
davidfowl: The first test in this file tests the same things, this adds the hook.

davidfowl:
@JamesNK new dashboard failures? Are they related to your changes to tests?

src/Aspire.Hosting/Orchestrator/ApplicationOrchestrator.cs
davidfowl: Updated WaitFor to handle this situation.

mitchdenny (approved):

PR #7842, opened 2025-03-01
[CI] Enumerate test projects at build time for PR validation

radical:
Generate list of tests to run for PR validation at run time

- Move core steps to run the tests to a new reusable workflow
  `run-tests.yml`
- Generate the list of test projects using the new
  `tests/Shared/GetTestProjects.proj`

- Adjust `tests.yml` to essentially use `run-tests.yml` to run the test
  matrix generated above

- Also, introduce a new property to control running tests on github
  actions `$(RunTestsOnGithubActions)`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-name: Run Integration Tests
+name: Integration Tests
 
 on:
   pull_request:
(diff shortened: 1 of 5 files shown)


tests/Shared/GetTestProjects.proj
```diff
@@ -0,0 +1,43 @@
+<Project DefaultTargets="GenerateTestProjectsJsonForGithubActions">
```
davidfowl, line 1: I think I would much rather have this expressed in yaml and than in msbuild.

radical: Having it in msbuild allows us to control at the .csproj level whether to skip the tests, though that might not be a common scenario.

I was trying to do all this in yaml, but generating the json for the matrix was a pain. But I do have the code simplified now, so that json is simpler too, and we can try yaml.

radical: Hm keeping it in msbuild though means that we are sharing the same rules for helix, and for gh actions, which would keep that consistent.

davidfowl: I hear you, but, copilot is kinda crap at writing msbuild and it's amazing at writing yaml. When I look at our csproj files wrapping shell commands its quite unmaintainable. This yaml file is long but very clean and really easy to understand. Maybe this is a fine middle ground but it feels really messy to me.

radical: By "shell commands" do you mean the json being emitted to the file? I should be able to get that out of this project, then it would just be evaluating all the msbuild projects to find the ones that should be built.

The alternative would be to pick up the test projects with globbing, and keeping an exclude list in the yaml.

danmoseley: My bias is David's, an exclude list in the yaml sounds good. I like simple. (If we don't care about Helix)

mitchdenny: I don't have a strong opinion here, but this would have helped me avoid breaking the build last week when I added _Aspire.Cli_.

I know that the Azure SDK uses this to great effect to enable more elaborate integration testing scenarios.

One thing where this matrix generation capability would be useful is in running the same set of tests across both Docker AND Podman container runtimes pretty easily.

But I don't have a strong opinion.

davidfowl: I think we should dynamically generate the list, I am just having a hard time having that logic be in msbuild 😄

mitchdenny (approved):
LGTM. Don't have a strong opinion on MSBuild - but trust your judgement on the engineering system side.

radical:
Taking a hybrid route:
1. Get the list of projects from msbuild
2. Build the matrix json for github actions (not needed for helix at all) to the yaml

tests/Shared/GetTestProjects.proj
davidfowl: Agree msbuild > python. I wouldn't mind if we could move off helix altogether 😄

radical:
The github checks now shows `Integration Tests / _ (Azure.Security.KeyVault) / Azure.Security.KeyVault (pull_request)` instead of `Run Integration Tests / Azure.Security.KeyVault`. This is because we are using a job matrix, and reusing a workflow (`run-tests.yml`).

.github/workflows/tests.yml
```diff
@@ -11,261 +11,45 @@ concurrency:
...
-
+    outputs:
+      tests_matrix: ${{ steps.generate_test_matrix.outputs.tests_matrix }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Set up .NET Core
-        uses: actions/setup-dotnet@v4
+      - uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
radical: Oops, I was using it earlier but dropped it in the latest changes. I'll remove this. thanks

tests/Shared/GetTestProjects.proj
```diff
@@ -0,0 +1,44 @@
...
+      <_TestProjectsToExclude Include="$(RepoRoot)tests\Aspire.Workload.Tests\**\*Tests.csproj" />
+      <_TestProjectsToExclude Include="$(RepoRoot)tests\Aspire.EndToEnd.Tests\**\*Tests.csproj" />
+
+      <_TestProjects Include="$(RepoRoot)tests\**\*Tests.csproj"
+                     Exclude="@(_TestProjectsToExclude)" />
+    </ItemGroup>
+
+    <MSBuild Projects="@(_TestProjects)" Targets="GetRunTestsOnGithubActions">
+      <Output TaskParameter="TargetOutputs" ItemName="ProjectsToTest" />
+    </MSBuild>
```
eerhardt, line 31: Doesn't that target return `true` or empty? What does it mean to have a bunch of `true` items?

.github/workflows/tests.yml
```diff
@@ -11,261 +11,45 @@ concurrency:
...
+      testShortName: ${{ matrix.shortname }}
+      testSessionTimeoutMs: ${{ matrix.testSessionTimeoutMs }}
+      extraTestArgs: ${{ matrix.extraTestArgs }}
 
   results: # This job is used for branch protection. It ensures all the above tests passed
     if: ${{ always() }}
     runs-on: ubuntu-latest
     name: Final Results
-    needs: [test]
+    needs: [_]
```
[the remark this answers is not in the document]
radical: The `_` shows in the github checks list as `Integration Tests / _ (Azure.Security.KeyVault) / Azure.Security.KeyVault (pull_request)`.

I used `_` just to reduce the length of that display string, and have it close to the current state `Run Integration Tests / Azure.Security.KeyVault`.

tests/Shared/GetTestProjects.proj
radical: ugh.. yes, you are correct. Missed this in the rework. thank you!

eerhardt (approved):
LGTM

PR #7849, opened 2025-03-02
Remove lauch.json and lean on dynamic launch config

davidfowl:
We have too many playground projects to hardcode a single run target.


mitchdenny:
We'll see how this goes. I suspect that using the dynamic ones with launch profiles is the easiest way to go for .NET anyway.

mitchdenny (approved):

PR #7861, opened 2025-03-03
Make AppHost shutdown if orphaned by aspire CLI.

mitchdenny:
This PR adds logic to both the app host and the aspire CLI to keep track of the PID of the CLI process. This information is passed along to the apphost so that when the process for the CLI disappears, the apphost will automatically shutdown gracefully. This is to deal with the situation where the CLI is killed and can't signal the apphost to shutdown (remember its DCP that is responsible for killing apphost sub-resources and it kicks in only after the apphost disappears.

This PR isn't quite ready as I'm working through some of the testing side of things to make sure we can reliably test the CLI as part of CI.


src/Aspire.Hosting/Cli/CliOrphanDetector.cs
```diff
@@ -0,0 +1,52 @@
...
+            return !Process.GetProcessById(pid).HasExited;
+        }
+        catch (ArgumentException)
+        {
+            // If Process.GetProcessById throws it means the process in not running.
+            return false;
+        }
+    };
+
+    internal Func<string, string?> GetEnvironmentVariable { get; set; } = Environment.GetEnvironmentVariable;
```
mitchdenny, line 28: Will replace with `IConfiguration` usage.

src/Aspire.Hosting/Cli/CliOrphanDetector.cs
```diff
@@ -0,0 +1,51 @@
...
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        if (configuration[CliProcessIdEnvironmentVariable] is not { } pidString || !int.TryParse(pidString, out var pid))
+        {
+            // If there is no PID environment variable, we assume that the process is not a child process
+            // of the .NET Aspire CLI and we won't continue monitoring.
+            return;
+        }
+
+        Console.WriteLine("CLI PROCESS ID IS: " + pid);
```
davidfowl, line 38: delete this

src/Aspire.Hosting/Cli/CliOrphanDetector.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+namespace Aspire.Hosting.Cli;
+
+internal sealed class CliOrphanDetector(IConfiguration configuration, IHostApplicationLifetime lifetime) : BackgroundService
+{
+    private const string CliProcessIdEnvironmentVariable = "ASPIRE_CLI_PID";
+
+    internal TimeProvider TimeProvider { get; set; } = TimeProvider.System;
```
davidfowl, line 14: Why not inject it?

mitchdenny: Will do ... I thought that we did it this way in other areas but I was wrong. It was also a similar approach to IsRunning just setting an internal property.

davidfowl (approved):

PR #7864, opened 2025-03-03
Include Aspire.Hosting.Analyzers in Aspire.Hosting.AppHost nuget package

eerhardt:
This was broken when the Aspire.Hosting.AppHost package multi-targeted to net8 and net9. The AnalyzerAssemblyPath was no longer being populated in the multi-targeted build.

The fix is to explicitly get the TargetPath from the Analyzer project and use it to add to the analyzers folder.

### DIFF
--- a/src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
+++ b/src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="$(RepoRoot)eng/ReplaceText.targets" />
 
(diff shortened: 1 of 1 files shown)


radical:
Is testing that the analyzer works included in the manual test cases?

radical:
Could we have some simple test to exercise the analyzer, just to sanity check that it is well formed/loads up?

radical (approved):

src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
```diff
@@ -39,13 +39,20 @@
...
   <ItemGroup>
     <!-- Reference the analyzer to ensure it's built but don't reference its output -->
-    <ProjectReference Include="..\Aspire.Hosting.Analyzers\Aspire.Hosting.Analyzers.csproj" OutputItemType="AnalyzerAssemblyPath" ReferenceOutputAssembly="false" Private="true" />
+    <ProjectReference Include="..\Aspire.Hosting.Analyzers\Aspire.Hosting.Analyzers.csproj" ReferenceOutputAssembly="false" Private="true" />
   </ItemGroup>
 
   <!-- Include the analyzer in the package -->
-  <Target Name="IncludeAnalyzerInPackage" AfterTargets="ResolveProjectReferences">
+  <PropertyGroup>
+    <BeforePack>$(BeforePack);IncludeAnalyzerInPackage</BeforePack>
```
DamianEdwards, line 47: Where is this property defined/read? Is this an arcade thing?

radical: In arcade: https://github.com/dotnet/arcade/blob/e52018af7f5bab97b0721382d52badfc7edb54f6/src/Microsoft.DotNet.Arcade.Sdk/tools/AfterSigning.proj#L28-L29

src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
```diff
@@ -39,13 +39,20 @@
...
+  </PropertyGroup>
+  <Target Name="IncludeAnalyzerInPackage">
+    <MSBuild Projects="..\Aspire.Hosting.Analyzers\Aspire.Hosting.Analyzers.csproj" Targets="GetTargetPath">
+      <Output TaskParameter="TargetOutputs" ItemName="_AnalyzerFile" />
+    </MSBuild>
```
radical, line 52: It might be useful to check that `@(_AnalyzerFile)` is not empty. 
nit: Also, this can use a `PropertyName` since it would be getting a single path.

src/Aspire.Hosting.AppHost/Aspire.Hosting.AppHost.csproj
eerhardt: This is the same approach (roughly) that dotnet/runtime uses to package their analyzers:

https://github.com/dotnet/runtime/blob/3d275542626ac6420269168ecbebe2bcb8e6ae54/eng/packaging.targets#L15

https://github.com/dotnet/runtime/blob/3d275542626ac6420269168ecbebe2bcb8e6ae54/eng/packaging.targets#L136-L152

The only difference is that dotnet/runtime has a Target in the analyzers project `GetAnalyzerPackFiles` to allow for multiple files coming from the analyzer. Here we only care about TargetPath, so I just grab it.

eerhardt:
> Could we have some simple test to exercise the analyzer, just to sanity check that it is well formed/loads up?

We have unit tests for the analyzer. https://github.com/dotnet/aspire/tree/main/tests/Aspire.Hosting.Analyzers.Tests

I'm unsure how to automate testing the nuget package here. Manually checking it is what I've been doing.

PR #7872, opened 2025-03-04
Remove ambiguity by removing this overload

davidfowl:
Follow up on https://github.com/dotnet/aspire/pull/7826#discussion_r1975655668. After a discussion with @eerhardt, we're removing this api


mitchdenny (approved):

eerhardt (commented):
Thanks! LGTM

PR #7873, opened 2025-03-04
Fix Aspire CLI test failures on Windows.

mitchdenny:
This PR fixes a build failure happening on main where the Windows command I was using to cause the stub process to wait indefinately was failing, causing the tests not to behave correctly. Once I fixed that issue another issue appeared in a test case which was now throwing a task cancelled exception. I fixed up the test so that we don't need to worry about that (might revisit in the future if it turns out to be an issue).

### DIFF
--- a/tests/Aspire.Cli.Tests/Hosting/CliOrphanDetectorTests.cs
+++ b/tests/Aspire.Cli.Tests/Hosting/CliOrphanDetectorTests.cs
@@ -109,7 +109,7 @@ public class CliOrphanDetectorTests(ITestOutputHelper testOutputHelper)
 
         // Wait until the apphost is spun up and then kill off the stub
         // process so everything is torn down.
-        _ = await resourcesCreatedEventsChannel.Reader.WaitToReadAsync().AsTask().WaitAsync(TimeSpan.FromSeconds(10000));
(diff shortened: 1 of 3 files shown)


davidfowl (approved):

tests/Aspire.Cli.Tests/Hosting/CliOrphanDetectorTests.cs
```diff
@@ -97,9 +98,21 @@ public async Task AppHostExitsWhenCliProcessPidDies()
...
+            resourcesCreatedTcs.SetResult();
+            return Task.CompletedTask;
+        });
+
         using var app = builder.Build();
         var pendingRun = app.RunAsync();
+
+        // Wait until the apphost is spun up and then kill off the stub
+        // process so everything is torn down.
+        await resourcesCreatedTcs.Task.WaitAsync(TimeSpan.FromSeconds(10));
```
danmoseley, line 113: do we care if the wait is exhausted for either of these?

mitchdenny: We care if it takes more than 10 seconds because it probably means it'll wait forever and just waste a build agent for an hour. So its really there just to ensure we fail fast.

PR #7874, opened 2025-03-04
`aspire pack [project] --target [publisher] --output-path [manifest filename]`

mitchdenny:
This PR introduces the `aspire pack` command. It's pretty much pass thru to `dotnet run -- --publisher manifest --output-path [path]` at this point. This is just a first pass at laying down the command a bunch more work will need to be done on the app host side to give publishers more superpowers.

### DIFF
--- a/src/Aspire.Cli/AppHostRunner.cs
+++ b/src/Aspire.Cli/AppHostRunner.cs
@@ -3,11 +3,10 @@
 
 using System.Diagnostics;
 using System.Globalization;
-using Microsoft.Extensions.Hosting;
 
 namespace Aspire.Cli;
 
-internal sealed class AppHostRunner(IHostApplicationLifetime lifetime)
+internal sealed class AppHostRunner
 {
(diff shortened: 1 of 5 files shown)


davidfowl:
The CLI commands should just exit automatically without ceremony. Maybe the host builder is overkill and we only need the IServiceProvider

src/Aspire.Cli/AppHostRunner.cs
```diff
@@ -40,7 +45,7 @@ public async Task<int> RunAppHostAsync(FileInfo appHostProjectFile, Cancellation
             process.Kill(false); // DCP should clean everything else up.
         }
 
-        await hostLifetime.StopAsync(cancellationToken).ConfigureAwait(false);
+        lifetime.StopApplication();
```
davidfowl, line 48: Why is this required?

mitchdenny: Its not. Just tweaked things so process lifetime doesn't wait for RunAsync(...) to return (since we are going to tear down the process anyway).

davidfowl (approved):

Issue #7780, opened 2025-02-26
Telemetry support from the Dashboard


leslierichardson95:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

We're currently missing telemetry for tracking user behavior and dashboard performance, making it difficult to track areas that need improvement and validate success for areas that are successful.

### Describe the solution you'd like

I'd like for us to investigate and implement the steps needed to get telemetry data for the Aspire dashboard so we can better track success for the dashboard and its features, and determine areas for improvement in the future.

joperezr:
Pending Privacy and Security Review

joperezr:
Next step: @adamint @JakeRadMSFT will define the exact list of things we want (key data points) to add for 9.3.

danmoseley:
Waiting on compliance work, which @adamint is driving.

adamint:
superceded by #8925

PR #7884, opened 2025-03-04
Implement dashboard telemetry

adamint:
## Description

Adds the `IAspireTelemetryService`, an opt-out feature in `SettingsDialog`, `TelemetryEnabledComponentBase` to enable automatic telemetry sending when a component renders and has new parameters, and implements `TelemetryEnabledComponentBase` in `ConsoleLogs`.

A cursory amount of telemetry is included in this PR, follow-ups will flesh out the rest of supported telemetry.

The `AspireTelemetryService` communicates with the webtools AspireServer, which handles the posting of telemetry via VS APIs. It works in VS and VS Code instances that contain the updated AspireServer, and never in CLI. APIs in `IAspireTelemetryService` directly parallel the MS.VS.Telemetry data model.
 
Resolves part of #7780 

**To test**: you can test two scenarios, one easy and one more time-consuming.
1) run a playground app in any non-webtools-patched-VS environment. You should **not** see a telemetry entry on the dashboard settings pane.
2) patch [this webtools branch](https://devdiv.visualstudio.com/DevDiv/_git/WebTools/pullrequest/613577) and then run a playground app. You should see a telemetry entry on the settings pane.
image

 
**Note**: this pull request only enables 2 telemetry events (aspire/dashboard/component/initialize and aspire/dashboard/component/render) and for only one page (console logs). Additional pull requests will add those events to other pages and will also add the remaining telemetry events from [here](https://microsoft.sharepoint.com/:fl:/s/a049211a-86f4-4983-b0d2-f3f0214e6fca/Ea1nUm4ucjJNvEt2Mk-v3SABN_oDQq6QDA6fU1W7XFAYCw?e=aQa5CQ&nav=cz0lMkZzaXRlcyUyRmEwNDkyMTFhLTg2ZjQtNDk4My1iMGQyLWYzZjAyMTRlNmZjYSZkPWIlMjFOT05GN1pHR2hFT2tqVTgtRndmY0puLUlBdnVjdzExTnZRU3ltR3JYSjRZUWwtbnNDUjNVVHA2b0RLYktudzhqJmY9MDFaUkdGSkpGTk01Skc0TFRTR0pHM1lTM1dHSkgyN1hKQSZjPSUyRiZhPUxvb3BBcHA%3D) everywhere applicable.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [X] No. Follow-up changes expected.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)): https://github.com/dotnet/docs-aspire/issues/2768
  - [ ] No


src/Aspire.Dashboard/Resources/Dialogs.resx
```diff
@@ -263,9 +263,15 @@
...
   <data name="SettingsRemoveAllButtonText" xml:space="preserve">
     <value>Remove all</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SettingsDialogEnableTelemetry" xml:space="preserve">
+    <value>Enable dashboard telemetry</value>
+  </data>
+  <data name="SettingsDialogTelemetryEnabledInfo" xml:space="preserve">
+    <value>When enabled, Aspire will collect usage data and send it to Microsoft to help improve the Aspire Dashboard.</value>
```
mitchdenny, line 275: We may want to be specific about the things that we DON'T send. For example we aren't sending copies of console legs and/or metrics.

JamesNK: @adamint Talk with Maddy/Damian about what this text should be.

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+
+/// <summary>
+/// A base class for components that would like to opt-in to automatic telemetry events, capturing
+/// 1) initial render time
+/// 2) component properties on each render
+///
+/// <remarks>If overriding <see cref="OnInitializedAsync"/>, <see cref="OnParametersSetAsync"/>/<see cref="OnParametersSet"/>, or <see cref="OnAfterRenderAsync"/>/<see cref="OnAfterRender"/>,
+/// inheritors <b>must</b> the accompanying base methods for telemetry to work properly.</remarks>
+/// </summary>
+public abstract class TelemetryEnabledComponentBase : ComponentBase
```
JamesNK, line 18: I rather this was modeled with composition rather than inheritance.

You can only inherit from one thing. In this situation, telemetry support feels more like adding a capabilitiy to the component rather than it being the fundamental base type. Inheritance is just being used to avoid duplication.

I think this should be an interface, similar to `IPageWithSessionAndUrlState`.

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+            properties[TelemetryPropertyKeys.DashboardComponentId] = new AspireTelemetryProperty(ComponentId);
+
+            var request = new PostOperationRequest(
+                TelemetryEventKeys.ParametersSet,
+                TelemetryResult.Success,
+                null,
+                properties,
+                _loadOperation?.Content?.Correlation is { } loadPageCorrelation ? [loadPageCorrelation] : null);
+
+            TelemetryService.PostUserTaskAsync(request);
```
JamesNK, line 77: No await?

src/Aspire.Dashboard/Telemetry/AspireTelemetryService.cs
```diff
@@ -0,0 +1,328 @@
...
+using System.Net;
+using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
+using Aspire.Dashboard.Configuration;
+using Aspire.Dashboard.Utils;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Telemetry;
+
+public sealed class AspireTelemetryService(IOptions<DashboardOptions> options, ILocalStorage localStorage) : IAspireTelemetryService
```
JamesNK: This is a pretty big type. For big types I prefer a normal constructor.

Directory.Packages.props
```diff
@@ -171,6 +171,8 @@
     <PackageVersion Include="Azure.Identity" Version="1.13.2" />
     <!-- https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3313 -->
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
+    <!-- VS telemetry package for dashboard telemetry -->
+    <PackageVersion Include="Microsoft.VisualStudio.Telemetry" Version="17.14.2" />
```
JamesNK, line 175: What do we use from here? It has a fair number of dependencies.

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+        OnAfterRender(firstRender);
+        return Task.CompletedTask;
+    }
+
+    protected override void OnAfterRender(bool firstRender)
+    {
+        if (_loadOperation?.Content is { OperationId: { } operationId })
+        {
+            _loadOperation = null;
+            TelemetryService.EndUserTaskAsync(new EndOperationRequest(operationId, TelemetryResult.Success));
```
JamesNK, line 92: No await?

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    protected override Task OnParametersSetAsync()
+    {
+        OnParametersSet();
+        return Task.CompletedTask;
+    }
+
+    protected override void OnParametersSet()
+    {
+        PostParametersSetTelemetryAsync();
```
JamesNK, line 61: This is going to trigger with every parameter change? This could happen a lot. Why record this?

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        if (TelemetryService.IsTelemetryEnabled)
+        {
+            var request = new StartOperationRequest(
+                TelemetryEventKeys.InitializeComponent,
+                new AspireTelemetryScopeSettings(new Dictionary<string, AspireTelemetryProperty> {
+                    // Component properties
+                    { TelemetryPropertyKeys.DashboardComponentId, new AspireTelemetryProperty(ComponentId) }
+                }));
+            _loadOperation = await TelemetryService.StartUserTaskAsync(request);
```
JamesNK, line 49: This is adding a HTTP request to every page load. What is the performance impact?

I think physically sending telemetry needs to be taken out of inline dashboard operations. UI and performance of the app shouldn't be impacted by making HTTP requests.

Instead, telemetry events should be written by the UI to a channel, and then there is a background thread reads telemetry events from the channel and sends it to the telemetry backend.

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
adamint: The idea was that so we always know the current state of the page, so you wouldn't have to add up the results of each user action to know what the current state is.

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
adamint: A channel makes sense

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -647,4 +651,13 @@ public ConsoleLogsPageState ConvertViewModelToSerializable()
     {
         return new ConsoleLogsPageState(PageViewModel.SelectedResource?.Name);
     }
+
+    protected override Dictionary<string, AspireTelemetryProperty> GetTelemetryProperties()
+    {
+        return new Dictionary<string, AspireTelemetryProperty>
```
JamesNK, line 657: Is a dictionary necessary? A lighter weight option would be `List<KeyValuePair<string, AspireTelemetryProperty>>`. Or you could add the name to `AspireTelemetryProperty`.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -647,4 +651,13 @@ public ConsoleLogsPageState ConvertViewModelToSerializable()
...
+        {
+            { TelemetryPropertyKeys.ConsoleLogsApplicationName, new AspireTelemetryProperty(PageViewModel.SelectedResource?.Name ?? string.Empty) },
```
JamesNK, line 659: Is it ok to record the resource name? That's PII.

src/Aspire.Dashboard/Telemetry/AspireTelemetryService.cs
```diff
@@ -0,0 +1,364 @@
...
+        {
+            return null;
+        }
+
+        return _httpClient.Value;
+    }
+
+    private static HttpClient? CreateHttpClient(DebugSession debugSession)
+    {
+        if (!HasDebugSession(debugSession, out var debugSessionUri, out var token, out var certData))
```
JamesNK, line 294: Returning null here is weird. I think check whether there is a debug session should happen before here.

But you can still call it here and throw if code mistakenly made its way to this point.

src/Aspire.Dashboard/Telemetry/IAspireTelemetryService.cs
```diff
@@ -0,0 +1,181 @@
...
+    Task<ITelemetryResponse?> EndUserTaskAsync(EndOperationRequest request);
+
+    /// <summary>
+    /// Performs a short-lived operation. Preference this method over <see cref="StartOperationAsync"/> and <see cref="EndOperationAsync"/> if the operation is contained within a single method.
+    /// Use <see cref="PerformUserTaskAsync"/> if the operation is a user task.
+    /// </summary>
+    Task PerformOperationAsync(StartOperationRequest request, Func<Task<OperationResult>> func);
+
+    /// <summary>
+    /// Performs a short-lived user task. Preference this method over <see cref="StartUserTaskAsync"/> and <see cref="EndUserTaskAsync"/> if the operation is contained within a single method.
```
JamesNK, line 75:
```suggestion
    /// Performs a short-lived user task. Prefer this method over <see cref="StartUserTaskAsync"/> and <see cref="EndUserTaskAsync"/> if the operation is contained within a single method.
```

src/Aspire.Dashboard/Telemetry/IAspireTelemetryService.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// </summary>
+    Task<ITelemetryResponse<StartOperationResponse>?> StartUserTaskAsync(StartOperationRequest request);
+
+    /// <summary>
+    /// Ends a long-running user task. This will post the end event and calculate the duration.
+    /// </summary>
+    Task<ITelemetryResponse?> EndUserTaskAsync(EndOperationRequest request);
+
+    /// <summary>
+    /// Performs a short-lived operation. Preference this method over <see cref="StartOperationAsync"/> and <see cref="EndOperationAsync"/> if the operation is contained within a single method.
```
JamesNK, line 69:
```suggestion
    /// Performs a short-lived operation. Prefer this method over <see cref="StartOperationAsync"/> and <see cref="EndOperationAsync"/> if the operation is contained within a single method.
```

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
JamesNK: I see that some methods return a response, e.g. `StartOperationResponse`, which have an ID on them. To support this with a channel, we should create our own ID (can just be an incrementing `long`) and return it from the method. Then there is an internal map between our ID and the ID eventually returned by the response.

Then when we end the operation, we pass in our ID, it's mapped to the response ID, and the external endpoint is called with the response ID.

src/Aspire.Dashboard/DashboardWebApplication.cs
```diff
@@ -230,6 +231,10 @@ public DashboardWebApplication(
         builder.Services.TryAddSingleton<IDashboardClientStatus, DashboardClientStatus>();
         builder.Services.TryAddScoped<DashboardCommandExecutor>();
 
+        // Telemetry
+        builder.Services.AddSingleton<AspireTelemetryService.ITelemetrySender, AspireTelemetryService.TelemetrySender>();
```
JamesNK, line 235: Move telemetry sender types so they're not nested.

src/Aspire.Dashboard/Telemetry/AspireTelemetryService.cs
```diff
@@ -0,0 +1,364 @@
...
+    public async Task InitializeAsync()
+    {
+        logger.LogDebug("Initializing telemetry service.");
+
+        if (_telemetrySupported is not null)
+        {
+            return;
+        }
+
+        _telemetrySupported = await GetTelemetrySupportedAsync(_httpClient.Value, telemetrySender, logger).ConfigureAwait(false);
```
JamesNK, line 38: HttpClient should be moved inside telemetrySender. The telemetry service shouldn't worry about that, the HttpClient is the sender's concern.

tests/Aspire.Dashboard.Components.Tests/Pages/ConsoleLogsTests.cs
```diff
@@ -380,6 +381,7 @@ private void SetupConsoleLogsServices(TestDashboardClient? dashboardClient = nul
         Services.AddSingleton<IDashboardClient>(dashboardClient ?? new TestDashboardClient());
         Services.AddSingleton<DashboardCommandExecutor>();
         Services.AddSingleton<ConsoleLogsManager>();
+        Services.AddScoped<IAspireTelemetryService, TestTelemetryService>();
```
JamesNK, line 384: Add this as a singleton like all other types. Repeat in other test files.

src/Aspire.Dashboard/Components/Dialogs/SettingsDialog.razor
```diff
@@ -46,6 +46,19 @@
             </div>
         </div>
 
+        @if (TelemetryService is { IsTelemetryInitialized: true, IsTelemetrySupported: true })
```
JamesNK, line 49: Shouldn't `IsTelemetrySupported` be enough here? I think by this point if telemetry is supported then it would always be initialized.

src/Aspire.Dashboard/Telemetry/IAspireTelemetryService.cs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// <summary>
+    /// Sets the telemetry enabled state for the current circuit. This will not affect debug session telemetry opt-out, and will be persisted to localStorage.
+    /// </summary>
+    /// <param name="enabled"></param>
+    /// <returns>Whether the state was successfully updated. False if <see cref="IsTelemetrySupported"/> is false.</returns>
+    Task<bool> SetTelemetryEnabledAsync(bool enabled);
+
+    /// <summary>
+    /// Begin a long-running user operation. Preference this over <see cref="PostOperationAsync"/>. If an explicit user task caused this operation to start,
```
JamesNK, line 47:
```suggestion
    /// Begin a long-running user operation. Prefer this over <see cref="PostOperationAsync"/>. If an explicit user task caused this operation to start,
```

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
```diff
@@ -31,6 +32,10 @@ public void Configure(DashboardOptions options)
         options.OtlpApiKey = configuration["AppHost:OtlpApiKey"];
 
         options.AspNetCoreEnvironment = configuration["ASPNETCORE_ENVIRONMENT"] ?? "Production";
+
+        options.TelemetryEnabled = bool.TryParse(configuration["DOTNET_DASHBOARD_ENABLE_TELEMETRY"], out var telemetryEnabled)
```
JamesNK, line 36: Rename this to `DOTNET_DASHBOARD_TELEMETRY_OPTOUT` and invert the logic.

src/Aspire.Dashboard/Resources/Dialogs.resx
```diff
@@ -263,9 +263,18 @@
...
   <data name="SettingsRemoveAllButtonText" xml:space="preserve">
     <value>Remove all</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SettingsDialogEnableTelemetry" xml:space="preserve">
+    <value>Enable dashboard telemetry</value>
+  </data>
+  <data name="SettingsDialogTelemetryEnabledInfo" xml:space="preserve">
+    <value>When enabled, Aspire will collect usage data and send it to Microsoft.</value>
```
danmoseley, line 275: typically these seem to say "anonymous usage data" -- assuming that's true. (there will be a right answer, maybe whaty ou have is the right answer)

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
```diff
@@ -0,0 +1,95 @@
...
+    [Inject]
+    public required NavigationManager NavigationManager { get; init; }
+
+    [Inject]
+    public required IAspireTelemetryService TelemetryService { get; init; }
+
+    protected abstract string ComponentId { get; }
+
+    protected virtual Dictionary<string, AspireTelemetryProperty> GetTelemetryProperties() => [];
+
```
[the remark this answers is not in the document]
danmoseley: ignore, this is a well known issue with cp right now

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -196,6 +196,8 @@ protected override async Task OnInitializedAsync()
 
         _isLoading = false;
 
+        return;
+
```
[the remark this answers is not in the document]
adamint: This is entirely incorrect.

Directory.Packages.props
adamint: Just a few enums. I copied the enums and removed the dependency

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: We have to send a dictionary to the aspire server, so we'll need to be creating a dictionary at some point

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: It's ok to, the property just has to be marked as PII and it'll get hashed and treated like other PII telemetry. Added Pii type, thanks

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
adamint: turned into an interface!

src/Aspire.Dashboard/Components/Pages/TelemetryEnabledComponentBase.cs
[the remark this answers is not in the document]
adamint: changed to post event

src/Aspire.Dashboard/Telemetry/AspireTelemetryService.cs
```diff
@@ -0,0 +1,364 @@
...
+using Aspire.Dashboard.Utils;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Telemetry;
+
+public sealed class AspireTelemetryService(
+    IOptions<DashboardOptions> options,
+    ILocalStorage localStorage,
+    AspireTelemetryService.ITelemetrySender telemetrySender,
+    ILogger<AspireTelemetryService> logger) : IAspireTelemetryService
```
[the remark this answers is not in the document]
JamesNK: Hmm, the problem with having an interface and a separate test implementation is if you need it to behave correctly, you have to hack together a solution in the test implementation.

I'll take a closer look when I test the branch out.

JamesNK:
I want to test it out for myself. Do I need a specific version of VS or VS Code to test this?

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,81 @@
...
+                    {
+                        var result = await requestFunc(client, GetResponseProperty).ConfigureAwait(false);
+                        foreach (var (guid, value) in guids.Zip(result))
+                        {
+                            _responsePropertyMap[guid] = value;
+                        }
+                    }
+                    catch (Exception ex) when (ex is HttpRequestException or JsonException or ArgumentException)
+                    {
+                        logger.LogWarning("Failed to make telemetry request: {ExceptionMessage}", ex.Message);
```
JamesNK, line 38:
```suggestion
                        logger.LogWarning(ex, "Failed to send telemetry request.");
```

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,81 @@
...
+                    }
+
+                    continue;
+
+                    object GetResponseProperty(Guid guid)
+                    {
+                        if (!_responsePropertyMap.TryGetValue(guid, out var value))
+                        {
+                            throw new ArgumentException("Response property not found, maybe a dependent telemetry request failed?", nameof(guid));
```
JamesNK, line 47:
```suggestion
                            throw new InvalidOperationException($"Response property not found. Id: {propertyId}");
```

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,81 @@
...
+                        }
+                    }
+                    catch (Exception ex) when (ex is HttpRequestException or JsonException or ArgumentException)
+                    {
+                        logger.LogWarning("Failed to make telemetry request: {ExceptionMessage}", ex.Message);
+                    }
+
+                    continue;
+
+                    object GetResponseProperty(Guid guid)
```
JamesNK, line 43: guid is a type, not the name of a variable. Is this a property id? Call it `propertyId`.

src/Aspire.Dashboard/Telemetry/IDashboardTelemetryService.cs
```diff
@@ -0,0 +1,240 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Net;
+using System.Text.Json.Serialization;
+
+namespace Aspire.Dashboard.Telemetry;
+
+public interface IDashboardTelemetryService
```
JamesNK, line 9: Split some of the types here into different files. Don't need a file per type. Group them logically.

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,81 @@
...
+                while (_channel.Reader.TryRead(out var operation))
+                {
+                    var (guids, requestFunc) = operation;
+                    try
+                    {
+                        var result = await requestFunc(client, GetResponseProperty).ConfigureAwait(false);
+                        foreach (var (guid, value) in guids.Zip(result))
+                        {
+                            _responsePropertyMap[guid] = value;
+                        }
```
JamesNK, line 34: Add a comment about what is happening here. Why the property map is required, what zip is for, etc.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
danmoseley: I know we already have "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL" and "DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL" but feels like it ought to be "ASPIRE_DASHBOARD_TELEMETRY_OPTOUT". No need to be unnecessarily dotnet centric even more than already. @DamianEdwards thoughts?

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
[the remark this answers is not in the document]
danmoseley: I hit thumbs down -- let's keep doing that when needed

src/Aspire.Dashboard/Telemetry/AspireTelemetryService.cs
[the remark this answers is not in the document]
adamint: Gotcha, I removed the interface now

adamint:
> I want to test it out for myself. Do I need a specific version of VS or VS Code to test this?

As of 3 hours ago, the version of WebTools containing the telemetry server was merged into VS main, so you should be able to use a VS main version more recent.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
danmoseley: For future env vars, we should probably use ASPIRE_ not DOTNET_ and for existing DOTNET_ ones should possibly also accept and document ASPIRE_ ... but that's a problem for later

DamianEdwards: Agree with @JamesNK

DamianEdwards: On naming, I agree also with @danmoseley that we should move to using `ASPIRE_` as our prefix (with back compat for the existing ones of course). We should do that all up in 9.2 I think.

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,53 @@
...
+                TelemetryEventKeys.InitializeComponent,
+                TelemetryResult.Success,
+                properties: new Dictionary<string, AspireTelemetryProperty> {
+                    // Component properties
+                    { TelemetryPropertyKeys.DashboardComponentId, new AspireTelemetryProperty(component.ComponentId) }
+                });
+        }
+    }
+
+    public static void PostParametersSetTelemetry(this IComponentWithTelemetry component)
```
JamesNK, line 42: I thought about this a little more. Recording an event on every parameter change could easily cause a lot of events that don't have much value. But at the same time, we want to capture when the parameters change so updated telemetry properties are sent.

How about this as a solution:

Store the last result from `component.GetTelemetryProperties()`. Then, when parameters change, check if the new properties are different from the old properties. If properties are different call `PostOperation` and update the stored properties. If properties are the same then do nothing.

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        if (component.TelemetryService.IsTelemetryEnabled)
+        {
+            component.InitializeCorrelation = component.TelemetryService.PostUserTask(
+                TelemetryEventKeys.InitializeComponent,
+                TelemetryResult.Success,
+                properties: new Dictionary<string, AspireTelemetryProperty> {
+                    // Component properties
+                    { TelemetryPropertyKeys.DashboardComponentId, new AspireTelemetryProperty(component.ComponentId) }
+                });
```
JamesNK, line 38: The captures the "page load" event right? Should there also be something recording in dispose to indicate that someone is leaving the page?

Maybe this situation is where it makes sense to have starting the page event and ending the page event be correlated. That way you can see how long someone was on a page.

JamesNK:
@adamint I looked at this more closely and played around with the code. I have some changes on a branch:

https://github.com/dotnet/aspire/compare/jamesnk/telemetry-refactor

Specifically, this commit: https://github.com/dotnet/aspire/commit/d1b002b706858105f39aa0a590f5106f29c4b9db

* I think these changes properly decouple the sender and the service. It moves the HTTP client into the sender, and HTTP interactions all happen in the sender. The service becomes a strongly typed wrapper over the sender.
* Removes the need for a concurrent directory to stop inprogress values. Values were being removed from it. It was a leak that would have grown forever. Now a reference to a property is passed around. No more GUIDs.
* Improve the certificate validation. Compare using SHA256 thumbprints.

I don't have a good VS version to actually test these changes work. Will need some manual testing to double check.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -299,3 +300,11 @@ internal bool TryParseOptions([NotNullWhen(false)] out IEnumerable<string>? erro
         return messages is null;
     }
 }
+
+public sealed class DebugSession
+{
+    public string? Address { get; set; }
+    public string? Token { get; set; }
+    public string? ServerCertificate { get; set; }
```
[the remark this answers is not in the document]
JamesNK:
```cs
public sealed class DebugSession
{
    private X509Certificate2? _serverCertificate;

    public string? Address { get; set; }
    public string? Token { get; set; }
    public string? ServerCertificate { get; set; }

    public X509Certificate2? GetServerCertificate() => _serverCertificate;

    internal bool TryParseOptions([NotNullWhen(false)] out string? errorMessage)
    {
        if (!string.IsNullOrEmpty(ServerCertificate))
        {
            byte[] data;
            try
            {
                data = Convert.FromBase64String(ServerCertificate);
            }
            catch (Exception ex)
            {
                errorMessage = $"Error converting server certificate payload from base64 to bytes: {ex.Message}";
                return false;
            }

            try
            {
                _serverCertificate = new X509Certificate2(data);
            }
            catch (Exception ex)
            {
                errorMessage = $"Error reading server certificate as X509Certificate2: {ex.Message}";
                return false;
            }
        }

        errorMessage = null;
        return true;
    }
}
```

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
adamint: That's a good idea. Added

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
danmoseley: Opened https://github.com/dotnet/aspire/issues/8009

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -150,6 +154,10 @@ protected override async Task OnInitializedAsync()
             PageViewModel.Status = Loc[nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsLogsNotYetAvailable)];
         }
 
+        await this.InitializeComponentTelemetryAsync();
+
+        return;
```
JamesNK, line 159: What's this return for?

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,77 @@
...
+    public DashboardTelemetryService TelemetryService { get; }
+
+    public string ComponentType { get; }
+    public string ComponentId { get; }
+    public Dictionary<string, AspireTelemetryProperty> GetTelemetryProperties() => [];
+}
+
+public static class IComponentWithTelemetryExtensions
+{
+    private static readonly Dictionary<string, Dictionary<string, AspireTelemetryProperty>> s_telemetryProperties = [];
```
JamesNK, line 29: This looks like a memory leak. Values are added to it and never removed. I think you reused this idea from what I suggested for storing properties returned from telemetry requests. To be honest it was a bad idea there, and it's a bad idea here.

Instead of doing this, I think there should be a `ComponentTelemetryContext` type. It tracks telemetry state for a component.

* Has component name property
* Has initialize correlation property
* Stores dictionary properties
* Triggers property update when dictionary properties change.

This has the benefit of reducing how much logic sits in the component.

The `GetTelemetryProperties` method on the component changes to `UpdateTelemetryProperties()`

```
    public Dictionary<string, AspireTelemetryProperty> UpdateTelemetryProperties()
    {
        _telemetryContext.UpdateTelemetryProperties([
            new ComponentTelemetryProperty(TelemetryPropertyKeys.MetricsApplicationInstanceId, PageViewModel.SelectedApplication.Id?.InstanceId ?? string.Empty) },
            new AspireTelemetryProperty(TelemetryPropertyKeys.MetricsApplicationInstanceId, PageViewModel.SelectedApplication.Id?.InstanceId ?? string.Empty) }
        ]);
    }
```

* `ComponentTelemetryProperty` can be a struct record that has the name, value, and type.
* Update method on context can be `UpdateTelemetryProperties(ReadOnlySpan<ComponentTelemetryProperty> properties)`. I think that could allow the span to be stackallocated.
* Internally the properties can be stored in a dictionary. If any of the properties change then trigger the telemetry request.

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,202 @@
...
+            try
+            {
+                while (await _channel.Reader.WaitToReadAsync().ConfigureAwait(false))
+                {
+                    while (_channel.Reader.TryRead(out var operation))
+                    {
+                        var (context, requestFunc) = operation;
+                        try
+                        {
+                            await requestFunc(_httpClient, GetResponseProperty).ConfigureAwait(false);
```
JamesNK, line 77:
```suggestion
                            await requestFunc(_httpClient, GetResponseProperty).ConfigureAwait(false);
                            _logger.LogTrace("Telemetry request successfully sent.");
```

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -277,6 +285,8 @@ protected override async Task OnParametersSetAsync()
                 LoadLogs(newConsoleLogsSubscription);
             }
         }
+
+        this.PostComponentTelemetryParametersAsync();
```
JamesNK, line 289: Is this hiding at the bottom of the method, beneath the local method? How come?

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    private static bool HasDebugSession(
+        DebugSession debugSession,
+        [NotNullWhen(true)] out Uri? debugSessionUri,
+        [NotNullWhen(true)] out string? token,
+        [NotNullWhen(true)] out X509Certificate2? serverCert)
+    {
+        if (debugSession.Address is not null && debugSession.Token is not null && debugSession.GetServerCertificate() is { } cert && debugSession.TelemetryOptOut is not true)
+        {
+            debugSessionUri = new Uri($"https://{debugSession.Address}");
```
JamesNK, line 48: From reading the doc here - https://github.com/dotnet/aspire/blob/main/docs/specs/IDE-execution.md#enabling-ide-execution - it seems like https is for when `DEBUG_SESSION_SERVER_CERTIFICATE` is present, otherwise it is http.

Can you update this to match the spec? Also write a unit test that checks the address specified with requests is correct depending on whether `DEBUG_SESSION_SERVER_CERTIFICATE` is set or not.

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            StartSendLoop();
+        }
+
+        return _isEnabled.Value;
+    }
+
+    private async Task<bool> TryStartTelemetrySessionCoreAsync()
+    {
+        if (HasDebugSession(_options.Value.DebugSession, out var debugSessionUri, out var token, out var certData))
```
JamesNK, line 155: cert data is optional - https://github.com/dotnet/aspire/blob/main/docs/specs/IDE-execution.md#enabling-ide-execution

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,202 @@
...
+                _logger.LogInformation("Ending sender loop.");
+            }
+        });
+    }
+
+    private HttpClient CreateHttpClient(Uri debugSessionUri, string token, X509Certificate2 cert)
+    {
+        var handler = new HttpClientHandler
+        {
+            ServerCertificateCustomValidationCallback = (_, c, _, e) =>
```
JamesNK, line 106: Only set this callback if a cert is provided. Otherwise requests will always be http

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+            finally
+            {
+                _logger.LogInformation("Ending sender loop.");
+            }
+        });
+    }
+
+    private HttpClient CreateHttpClient(Uri debugSessionUri, string token, X509Certificate2 cert)
```
JamesNK, line 102: Move code for creating the debug session HttpClient along with  `HasDebugSession` method. It can be placed in a static class, `DebugSessionHelpers`?, in the Models directory.

We may use it for things other than just telemetry.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: There's no local method defined in that method. Because page state might be changing in OnParametersSet, it feels safer to always have this call at the end

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
adamint: I added this ComponentTelemetryContext, with slightly different signatures. What do you think?

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -150,6 +154,10 @@ protected override async Task OnInitializedAsync()
             PageViewModel.Status = Loc[nameof(Dashboard.Resources.ConsoleLogs.ConsoleLogsLogsNotYetAvailable)];
         }
 
+        await this.InitializeComponentTelemetryAsync();
+
+        return;
+
```
JamesNK, line 160: This doesn't do anything. Below it are local methods.
```suggestion
```

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,93 @@
...
+            TelemetryEventKeys.ComponentInitialize,
+            TelemetryResult.Success,
+            properties: new Dictionary<string, AspireTelemetryProperty>
+            {
+                // Component properties
+                { TelemetryPropertyKeys.DashboardComponentId, new AspireTelemetryProperty(component.TelemetryContext.ComponentType) }
+            });
+    }
+
+    public static void DisposeComponentTelemetry(this IComponentWithTelemetry component)
```
JamesNK, line 40: What happens here if there is an exception in the component's initialize method, and `InitializeComponentTelemetryAsync` is never called?

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,93 @@
...
+    public ComponentTelemetryContext TelemetryContext { get; }
+
+    public void UpdateTelemetryProperties()
+    {
+    }
+}
+
+public static class IComponentWithTelemetryExtensions
+{
+    public static async Task InitializeComponentTelemetryAsync(this IComponentWithTelemetry component)
```
JamesNK, line 26: What do you think of moving these static methods can be moved onto TelemetryContext? TelemetryContext would need a reference to the service, but that doesn't seem like a big problem.

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            PostProperties(telemetryService);
+        }
+
+        return anyChange;
+    }
+
+    private void PostProperties(DashboardTelemetryService telemetryService)
+    {
+        _properties[TelemetryPropertyKeys.DashboardComponentId] = new AspireTelemetryProperty(ComponentType);
```
JamesNK, line 83: Seems like this can be done once in a constructor rather than each PostProperties call.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
JamesNK, line 200: This doesn't do anything. Below it are just local methods.

src/Aspire.Dashboard/Model/DebugSessionHelpers.cs
```diff
@@ -0,0 +1,69 @@
...
+        return client;
+    }
+
+    public static bool HasDebugSession(
+        DebugSession debugSession,
+        out X509Certificate2? serverCert,
+        [NotNullWhen(true)] out Uri? debugSessionUri,
+        [NotNullWhen(true)] out string? token)
+    {
+        if (debugSession.Address is not null && debugSession.Token is not null && debugSession.TelemetryOptOut is not true)
```
JamesNK, line 55: TelemetryOptOut shouldn't be tested here. You can have a debug session when telemetry is opted out.

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Diagnostics;
+using System.Net;
+using System.Threading.Channels;
+using Aspire.Dashboard.Configuration;
+using Aspire.Dashboard.Model;
+using Microsoft.Extensions.Options;
+
+namespace Aspire.Dashboard.Telemetry;
+
+public class DashboardTelemetrySender : IDashboardTelemetrySender
```
JamesNK, line 13:
```suggestion
public sealed class DashboardTelemetrySender : IDashboardTelemetrySender
```

src/Aspire.Dashboard/Telemetry/OperationContext.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        var context = new OperationContext
+        {
+            Properties = properties
+        };
+
+        return context;
+    }
+
+    public OperationContextProperty[] Properties { get; init; } = [];
```
JamesNK, line 30: Add a string `Name` property to the context. The value can be the telemetry endpoint. We can use this name in logs and tests

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,157 @@
...
+                while (await _channel.Reader.WaitToReadAsync().ConfigureAwait(false))
+                {
+                    while (_channel.Reader.TryRead(out var operation))
+                    {
+                        var (context, requestFunc) = operation;
+                        try
+                        {
+                            await requestFunc(Client, GetResponseProperty).ConfigureAwait(false);
+
+                            _logger.LogTrace("Telemetry request successfully sent.");
```
JamesNK, line 56: Include name from context (see previous comment about adding it)

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                            // Double check properties are set.
+                            foreach (var property in context.Properties)
+                            {
+                                if (!property.HasValue)
+                                {
+                                    _logger.LogWarning("Unset context property.");
```
JamesNK, line 63: Include name from context (see previous comment about adding it)

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,157 @@
...
+                                }
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogWarning(ex, "Failed to send telemetry request.");
```
JamesNK, line 69: Include name from context (see previous comment about adding it)

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
JamesNK: For example:
```suggestion
                            _logger.LogTrace("Telemetry request {Name} successfully sent.", context.Name);
```

JamesNK:
Just some minor feedback. I think this is almost done.

I think it would be useful to have a component test that verifies the end-to-end of the initialize event, parameter update events, dispose event. Shouldn't be difficult to verify expected operations from using the test telemetry sender. I recommended adding a name to the context so you can assert on that.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
adamint: I prefer to keep the return before local methods for readability. But deleted

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
adamint: I prefer to keep the return before local methods for readability. But deleted

src/Aspire.Dashboard/Components/Pages/IComponentWithTelemetry.cs
adamint: That seems like a good idea, moved

adamint:
> Just some minor feedback. I think this is almost done.
> 
> I think it would be useful to have a component test that verifies the end-to-end of the initialize event, parameter update events, dispose event. Shouldn't be difficult to verify expected operations from using the test telemetry sender. I recommended adding a name to the context so you can assert on that. Test should set a pretend correlation ID on initiailize and verify it is passed to other sent contexts.
> 
> Note that I had problems earlier with unit testing component dispose in bUnit. The test framework doesn't automatically call DisposeAsync. You may need to call dispose manually at the end of the test to trigger the dispose telemetry event, and then assert on it.

Now that almost all the logic is inside the `ComponentTelemetryContext`, I just wrote a unit test on that class. Let me know if you want to test one of the pages too

src/Aspire.Dashboard/Telemetry/DashboardTelemetrySender.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            // If the request fails, we've been given an invalid server address and should assume telemetry is unsupported.
+            _logger.LogWarning(ex, "Failed to request whether telemetry is supported.");
+            return false;
+        }
+    }
+
+    public void MakeRequest(OperationContext context, Func<HttpClient, Func<OperationContextProperty, object>, Task> requestFunc)
+    {
+        _channel.Writer.TryWrite((context, requestFunc));
```
JamesNK, line 147: I think a check here is needed for whether the sender is enabled or not. We don't want someone to mistakenly queue requests here and fill up the channel when not enabled.

Probably should throw if not enabled or not initialized

JamesNK (approved):
This is really good.

Merge once remaining feedback is applied.

JamesNK:
I added an integration test for reading the config values.

adamint:
Holding for merge pending internal reviews.

src/Aspire.Hosting/Dashboard/DashboardOptions.cs
```diff
@@ -31,6 +32,10 @@ public void Configure(DashboardOptions options)
         options.OtlpApiKey = configuration["AppHost:OtlpApiKey"];
 
         options.AspNetCoreEnvironment = configuration["ASPNETCORE_ENVIRONMENT"] ?? "Production";
+
+        options.TelemetryOptOut = bool.TryParse(configuration["DOTNET_DASHBOARD_TELEMETRY_OPTOUT"], out var telemetryOptOut)
```
DamianEdwards, line 36: This configuration key doesn't match the one in the readme.

src/Aspire.Dashboard/Configuration/DashboardOptions.cs
```diff
@@ -305,3 +306,45 @@ internal bool TryParseOptions([NotNullWhen(false)] out IEnumerable<string>? erro
         return messages is null;
     }
 }
+
+public sealed class DebugSession
```
JamesNK, line 310: All the other option types end with `Options`.
```suggestion
public sealed class DebugSessionOptions
```

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor.cs
```diff
@@ -267,6 +275,21 @@ public Task OnViewRelationshipAsync(ResourceDetailRelationship relationship)
         NavigationManager.NavigateTo(DashboardUrls.ResourcesUrl(resource: relationship.Resource.Name));
         return Task.CompletedTask;
     }
+
+    // IComponentWithTelemetry impl
+    public ComponentTelemetryContext TelemetryContext { get; } = new("ResourceDetails");
```
JamesNK, line 280: Add a central constants class to get these strings from.

src/Aspire.Dashboard/Components/Pages/ComponentTelemetryContext.cs
```diff
@@ -15,9 +16,10 @@ public class ComponentTelemetryContext(string componentType) : IDisposable
 
     private DashboardTelemetryService TelemetryService => _telemetryService ?? throw new ArgumentNullException(nameof(_telemetryService), "InitializeAsync has not been called");
 
-    public async Task InitializeAsync(DashboardTelemetryService telemetryService)
+    public async Task InitializeAsync(DashboardTelemetryService telemetryService, IJSRuntime js)
     {
         Properties[TelemetryPropertyKeys.DashboardComponentId] = new AspireTelemetryProperty(componentType);
+        Properties[TelemetryPropertyKeys.UserAgent] = new AspireTelemetryProperty(await telemetryService.GetUserAgentAsync(js));
```
JamesNK, line 22: I haven't got up to `GetUserAgentAsync` yet, but just want to make sure that the value is cached. Add a comment here that GetUserAgentAsync is cached.

src/Aspire.Dashboard/Components/Pages/ComponentTelemetryContext.cs
```diff
@@ -32,7 +34,7 @@ public async Task InitializeAsync(DashboardTelemetryService telemetryService)
             });
     }
 
-    public bool UpdateTelemetryProperties(ReadOnlySpan<ComponentTelemetryProperty> modifiedProperties)
+    public bool UpdateTelemetryProperties(IEnumerable<ComponentTelemetryProperty> modifiedProperties)
```
JamesNK, line 37: FYI this change will cause an allocation when doing `UpdateTelemetryProperties([...])`

src/Aspire.Dashboard/Components/Pages/ComponentTelemetryContext.cs
```diff
@@ -15,9 +16,10 @@ public class ComponentTelemetryContext(string componentType) : IDisposable
 
     private DashboardTelemetryService TelemetryService => _telemetryService ?? throw new ArgumentNullException(nameof(_telemetryService), "InitializeAsync has not been called");
 
-    public async Task InitializeAsync(DashboardTelemetryService telemetryService)
+    public async Task InitializeAsync(DashboardTelemetryService telemetryService, IJSRuntime js)
```
JamesNK, line 19: Inject IJSRuntime into the telemetry service so it doesn't need to be passed in each time? Or does DashboardTelemetryService have the wrong lifetime (singleton and not session?)

src/Aspire.Dashboard/Components/Pages/Error.razor.cs
```diff
@@ -16,4 +17,32 @@ public partial class Error
...
+
+    [Inject]
+    public required DashboardTelemetryService TelemetryService { get; init;  }
+
+    // IComponentWithTelemetry impl
+    public ComponentTelemetryContext TelemetryContext { get; } = new("Error");
+
+    protected override async Task OnInitializedAsync()
+    {
+        await TelemetryService.InitializeAsync();
```
JamesNK, line 29: Doesn't ComponentTelemetryContext need to be initialized?

src/Aspire.Dashboard/Components/Pages/ComponentTelemetryContext.cs
JamesNK: When the dashboard is first visited some info is sent from the client, such as the browser timezone. I think the user agent could be sent at the same time. That batches together client interop calls and improves initial load performance.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -296,6 +302,8 @@ protected override async Task OnParametersSetAsync()
                 LoadLogs(newConsoleLogsSubscription);
             }
         }
+
+        this.UpdateTelemetryProperties();
```
JamesNK, line 306:
```suggestion
        UpdateTelemetryProperties();
```

src/Aspire.Dashboard/Components/Pages/NotFound.razor.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [Inject]
+    public required DashboardTelemetryService TelemetryService { get; init;  }
+
+    // IComponentWithTelemetry impl
+    public ComponentTelemetryContext TelemetryContext { get; } = new("NotFound");
+
+    protected override async Task OnInitializedAsync()
+    {
+        await TelemetryService.InitializeAsync();
```
JamesNK, line 18: Initialize context

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -818,4 +825,22 @@ private async Task ContextMenuClosed(Microsoft.AspNetCore.Components.Web.MouseEv
...
+    public void UpdateTelemetryProperties()
+    {
+        var properties = new List<ComponentTelemetryProperty>
+        {
+            new(TelemetryPropertyKeys.ResourceView, new AspireTelemetryProperty(PageViewModel.SelectedViewKind.ToString(), AspireTelemetryPropertyType.UserSetting))
+        };
+
+        foreach (var resourceTypeGroup in _resourceByName.Values.GroupBy(r => r.ResourceType))
+        {
+            properties.Add(new ComponentTelemetryProperty($"{TelemetryPropertyKeys.ResourceType}.{resourceTypeGroup.Key}", new AspireTelemetryProperty(resourceTypeGroup.Count(), AspireTelemetryPropertyType.Metric)));
```
JamesNK, line 841: Resource type is being used in the key. What happens if the resource type contains dots? Or anything else odd or unusual. Does it need to be escaped?.

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor.cs
```diff
@@ -481,4 +488,18 @@ public class StructuredLogsPageState
...
+
+    public void UpdateTelemetryProperties()
+    {
+        TelemetryContext.UpdateTelemetryProperties([
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSelectedApplication, new AspireTelemetryProperty(PageViewModel.SelectedApplication.Id?.ToString() ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSelectedLogLevel, new AspireTelemetryProperty(PageViewModel.SelectedLogLevel.Id?.ToString() ?? string.Empty, AspireTelemetryPropertyType.UserSetting)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsFilterCount, new AspireTelemetryProperty(ViewModel.Filters.Count.ToString(CultureInfo.InvariantCulture), AspireTelemetryPropertyType.Metric)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsTraceId, new AspireTelemetryProperty(TraceId ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSpanId, new AspireTelemetryProperty(SpanId ?? string.Empty))
+        ]);
```
JamesNK, line 503: Have you checked with people that sending an empty string is the right approch when there is no value? The alternative is not adding the key at all.

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor.cs
```diff
@@ -316,5 +324,16 @@ public void Dispose()
...
+        TelemetryContext.Dispose();
+    }
+
+    // IComponentWithTelemetry impl
+    public ComponentTelemetryContext TelemetryContext { get; } = new(DashboardUrls.TracesBasePath);
+
+    public void UpdateTelemetryProperties()
+    {
+        TelemetryContext.UpdateTelemetryProperties([
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.TraceDetailTraceId, new AspireTelemetryProperty(TraceId)),
```
JamesNK, line 336: What would we use the trace id for? Doesn't seem like useful data.

src/Aspire.Dashboard/Components/Pages/Metrics.razor.cs
```diff
@@ -304,5 +315,22 @@ public void Dispose()
...
+    public ComponentTelemetryContext TelemetryContext { get; } = new(DashboardUrls.MetricsBasePath);
+
+    public void UpdateTelemetryProperties()
+    {
+        TelemetryContext.UpdateTelemetryProperties([
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.ApplicationInstanceId, new AspireTelemetryProperty(PageViewModel.SelectedApplication.Id?.InstanceId ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.MetricsApplicationIsReplica, new AspireTelemetryProperty(PageViewModel.SelectedApplication.Id?.ReplicaSetName is not null)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.MetricsInstrumentsCount, new AspireTelemetryProperty(PageViewModel.Instruments?.Count ?? -1)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.MetricsSelectedMeter, new AspireTelemetryProperty(PageViewModel.SelectedMeter?.MeterName ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.MetricsSelectedInstrument, new AspireTelemetryProperty(PageViewModel.SelectedInstrument?.Name ?? string.Empty)),
```
JamesNK, line 331: Isn't some of this PII?

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor.cs
```diff
@@ -481,4 +488,18 @@ public class StructuredLogsPageState
...
+    public ComponentTelemetryContext TelemetryContext { get; } = new(DashboardUrls.TracesBasePath);
+
+    public void UpdateTelemetryProperties()
+    {
+        TelemetryContext.UpdateTelemetryProperties([
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSelectedApplication, new AspireTelemetryProperty(PageViewModel.SelectedApplication.Id?.ToString() ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSelectedLogLevel, new AspireTelemetryProperty(PageViewModel.SelectedLogLevel.Id?.ToString() ?? string.Empty, AspireTelemetryPropertyType.UserSetting)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsFilterCount, new AspireTelemetryProperty(ViewModel.Filters.Count.ToString(CultureInfo.InvariantCulture), AspireTelemetryPropertyType.Metric)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsTraceId, new AspireTelemetryProperty(TraceId ?? string.Empty)),
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.StructuredLogsSpanId, new AspireTelemetryProperty(SpanId ?? string.Empty))
```
JamesNK, line 502: What here is PII?

src/Aspire.Dashboard/Components/Pages/TraceDetail.razor.cs
JamesNK, line 336: Is TraceId PII?

src/Aspire.Dashboard/Components/Pages/Traces.razor.cs
```diff
@@ -382,4 +382,14 @@ public class TracesPageState
...
         public required IReadOnlyCollection<TelemetryFilter> Filters { get; set; }
     }
+
+    // IComponentWithTelemetry impl
+    public ComponentTelemetryContext TelemetryContext { get; } = new(DashboardUrls.TracesBasePath);
+
+    public void UpdateTelemetryProperties()
+    {
+        TelemetryContext.UpdateTelemetryProperties([
+            new ComponentTelemetryProperty(TelemetryPropertyKeys.ApplicationInstanceId, new AspireTelemetryProperty(PageViewModel.SelectedApplication.Id?.InstanceId ?? string.Empty)),
```
JamesNK, line 392: PII?

src/Aspire.Dashboard/Model/DashboardCommandExecutor.cs
```diff
@@ -36,9 +38,25 @@ public async Task ExecuteAsync(ResourceViewModel resource, CommandViewModel comm
             _executingCommands.Add(executingCommandKey);
         }
 
+        await telemetryService.InitializeAsync().ConfigureAwait(false);
```
JamesNK, line 41: Hmmm, I don't like how telemetry service needs to be initialized everywhere just in case it hasn't already been initialized.

This is a scoped service so it can't just be initialized once when the app starts. We need to find an event after a circuit starts and initialize there. Don't try and fix this now, but I want to make this better.

src/Aspire.Dashboard/Model/DashboardCommandExecutor.cs
```diff
@@ -36,9 +38,25 @@ public async Task ExecuteAsync(ResourceViewModel resource, CommandViewModel comm
...
+
+        var startEvent = telemetryService.StartOperation(TelemetryEventKeys.ExecuteCommand,
+            new Dictionary<string, AspireTelemetryProperty>
+            {
+                { TelemetryPropertyKeys.ResourceType, new AspireTelemetryProperty(resource.ResourceType) },
+                { TelemetryPropertyKeys.CommandName, new AspireTelemetryProperty(command.Name) },
```
JamesNK, line 47: Is this PII?

src/Aspire.Dashboard/Telemetry/DashboardTelemetryService.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        return context;
+    }
+
+    /// <summary>
+    /// Ends a long-running operation. This will post the end event and calculate the duration.
+    /// </summary>
+    public void EndOperation(OperationContextProperty? operationId, TelemetryResult result, string? errorMessage = null)
+    {
+        if (!IsEnabled() || operationId is null)
```
JamesNK, line 114: Seems odd to have the method noop if there isn't an operationId. Why not have the caller check if there is an operationId and make that decision?

Some with EndUserTask.

src/Aspire.Dashboard/Telemetry/DashboardTelemetryService.cs
```diff
@@ -0,0 +1,390 @@
...
+        {
+            { TelemetryPropertyKeys.DashboardVersion, new AspireTelemetryProperty(typeof(DashboardWebApplication).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty) },
+            { TelemetryPropertyKeys.DashboardBuildId, new AspireTelemetryProperty(typeof(DashboardWebApplication).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version ?? string.Empty) },
+        };
+    }
+
+    public ValueTask<string> GetUserAgentAsync(IJSRuntime js)
+    {
+        return js.InvokeAsync<string>("getUserAgent");
+    }
```
JamesNK, line 355: The user agent is used with telemetry but this doesn't seem like the right type to place it on. I guess having it here gives a place to cache the value.

src/Aspire.Dashboard/Telemetry/TelemetryExceptionReporter.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        if (exception is not null)
+        {
+            Task.Run(async () =>
+            {
+                var serviceScope = serviceProvider.CreateAsyncScope();
+                var telemetryService = serviceScope.ServiceProvider.GetRequiredService<DashboardTelemetryService>();
+                await telemetryService.InitializeAsync().ConfigureAwait(false);
+
+                telemetryService.PostFault(
```
JamesNK, line 29: Have you tested this?

You're creating a new scope, which means a brand new telemetry service (along with creating a lot of other scoped DI types) which will send a new get call to the IDE to check if telemetry is available, queing one error message, then disposing everything. I think the telemetry will get sent, because disposing the sender waits for the queue to be empty, but this is way too much to do for each error message.

Loggers have the wrong lifetime. They're singletons and not in a scope, which is why you're having to do this.

Reporting errors should be when there is an unhandled error in Blazor. In other words, when there that yellow bar is displayed. I don't know what the event hook for that is, but I'm sure there is one.

src/Aspire.Dashboard/wwwroot/js/app.js
```diff
@@ -370,3 +370,7 @@ window.downloadStreamAsFile = async function (fileName, contentStreamReference)
     anchorElement.remove();
     URL.revokeObjectURL(url);
 };
+
+window.getUserAgent = function() {
```
JamesNK, line 374: There is getBrowserTimeZone method. This info should be combined with it. It could be called something like `getBrowserInfo`.

Browser details can be called and cached at the same time to avoid many roundtrips to the client.

tests/Aspire.Dashboard.Components.Tests/Pages/StructuredLogsTests.cs
```diff
@@ -124,6 +125,7 @@ public void Render_FiltersWithSpecialCharacters_SuccessfullyParsed()
             builder.Add(p => p.ViewportInformation, viewport);
         });
 
+        await Task.Delay(2000);
```
JamesNK, line 128: Hmmm. You shouldn't do this unless it's nessessary. Is it nessessary?
If you do do it, there should be a comment saying why.

tests/Aspire.Dashboard.Tests/Telemetry/ComponentTelemetryContextTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        public ValueTask<TValue> InvokeAsync<TValue>(string identifier, object?[]? args)
+        {
+            return typeof(TValue) == typeof(string) ? new ValueTask<TValue>((TValue)(object)string.Empty) : new ValueTask<TValue>(default(TValue)!);
+        }
+
+        public ValueTask<TValue> InvokeAsync<TValue>(string identifier, CancellationToken cancellationToken, object?[]? args)
+        {
+            return typeof(TValue) == typeof(string) ? new ValueTask<TValue>((TValue)(object)string.Empty) : new ValueTask<TValue>(default(TValue)!);
+        }
+    }
```
JamesNK, line 85: This is browser/components logic leaking into models. They should be kept separated.

JamesNK:
Next time please merge a very long PR that was already approved and make a new PR. It's a little more work for you, but it's difficult for reviewers to review new commits to an already very long PR without re-reviewing everything.

JamesNK (changes requested):
Next time please merge a very long PR that was already approved, and add further commits in new PR.

It's a little more work for you, but it's difficult for reviewers to review new commits to an already very long PR without re-reviewing everything.

JamesNK (approved):

src/Aspire.Dashboard/Components/Pages/StructuredLogs.razor.cs
adamint: I've checked, it is better to not send at all. Updated

PR #7890, opened 2025-03-04
Default WriteToManifest for AzureBicepResource

eerhardt:
We have this duplicated code every time we want to create an Azure resource. It makes more sense to consolidate it in one spot - when the resource is created.


sebastienros (approved):

davidfowl (approved):
Been meaning to do this!

PR #7894, opened 2025-03-05
Spike WithEntityFrameworkMigration

mitchdenny:
DO NOT MERGE

This explores the idea of adding a `WithEntityFrameworkMigrations(...)` extension method which creates an executable that calls out to `dotnet ef database update`. At the moment this code doesn't work because we are being impacted by the dreaded:

```
/workspaces/aspire/playground/waitfor/WaitForSandbox.ApiService/WaitForSandbox.ApiService.csproj : error MSB4057: The target "GetEFProjectMetadata" does not exist in the project.
Unable to retrieve project metadata. Ensure it's an SDK-style project. If you're using a custom BaseIntermediateOutputPath or MSBuildProjectExtensionsPath values, Use the --msbuildprojectextensionspath option.
```

But conceptually dynamically adding an executable and passing along the connection string works.


mitchdenny:
Ergonomics wise it might make sense to have a WithMigrations(...) extension method which is hand-crafted for each resource type for which we support migrations instead of what I've done here which is trying to make it generic at the Aspire.Hosting package level.

This would be better from an ergonomics point of view because we'd be able to do this:

```
builder.AddPostgres(...).AddDatabase(...).WithEntityFrameworkMigration<Projects.ProjectWithDbContext>();
```

It looks like the MSBuild issues with `dotnet ef database update` and `dotnet ef migrations add` have been solved in .NET 10 - but there is no plan to backport them. So maybe this isn't something we can do before we start on the .NET Aspire 10 round of enhancements?

DamianEdwards:
> It looks like the MSBuild issues with dotnet ef database update and dotnet ef migrations add have been solved in .NET 10 - but there is no plan to backport them. So maybe this isn't something we can do before we start on the .NET Aspire 10 round of enhancements?

We should figure out the details and push to backport if we think it's needed for this scenario.

afscrome:
I have been doing something very similar to this successfully for a while now:

```cs
builder.AddExecutable("dbMigrations", "dotnet", "../My.Migration.Project")
   .WithArgs(
      "ef",
      "database",
      "update",
      "--connection",
      database.Resource.ConnectionStringExpression,
      "--startup-project",
      "../My.Api",
      "-v",
      "--no-build") // As this project is part of the solution, relying on the fact that it has just been built before running the app host
   .WaitFor(postgres)
   .WithParentRelationship(database)
   ;
```

src/Aspire.Hosting/EntityFrameworkMigrationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+                                    context.Args.Add("ef");
+                                    context.Args.Add("database");
+                                    context.Args.Add("update");
+                                    context.Args.Add("--project");
+                                    context.Args.Add(projectMetadata.ProjectPath);
+                                    context.Args.Add("--connection");
+                                    context.Args.Add(builder);
+                                  })
+                                  .WaitFor((IResourceBuilder<IResource>)builder);
+        return builder;
```
afscrome, line 33: What if your app wants to wait for migrations to be completed?  With thisapproach, your app would need to wait on the migrator resource, but the migrator resource is not returned to the consumer?

An to running `dotnet ef migrate` as a seperate process, you could run the migrations in-process via the `ResourceReadyEvent` - as the event is blocking, consumers could just wait on the database resource directly, although it would require a bit more code to write, wire up logs, handle errors etc.

src/Aspire.Hosting/EntityFrameworkMigrationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+                                  .WithArgs(context => {
+                                    context.Args.Add("ef");
+                                    context.Args.Add("database");
+                                    context.Args.Add("update");
+                                    context.Args.Add("--project");
+                                    context.Args.Add(projectMetadata.ProjectPath);
+                                    context.Args.Add("--connection");
+                                    context.Args.Add(builder);
+                                  })
+                                  .WaitFor((IResourceBuilder<IResource>)builder);
```
afscrome, line 32: Does `dotnet ef migrate` emit open telemetry?
```suggestion
                                  .WaitFor((IResourceBuilder<IResource>)builder)
                                  .WithOtlpExporter()
                                  ;
```

mitchdenny: Its a good question.

src/Aspire.Hosting/EntityFrameworkMigrationExtensions.cs
mitchdenny: We are considering this option as well.

davidfowl: Running it in process mean referencing the DbContext and all of the app code from the apphost, which is a non starter.

davidfowl:
@mitchdenny can you file an issue for this?

PR #7896, opened 2025-03-05
Remove Aspire.Cli from unexpected packages list

mitchdenny:
Exclude Aspire.Cli from the set of unexpected packages. This is currently breaking the build.


mitchdenny:
@radical @joperezr the build is currently broken because I set Aspire.Cli to be packable (wasn't picked up on PR validation). Is this the correct approach to solve this problem?

davidfowl (approved):

davidfowl:
Why are these packages still arch specific even though you removed publishaot?

mitchdenny:
They aren't arch specific (at least Aspire.Cli isn't). It just so happens some of these other ones are.

PR #7909, opened 2025-03-05
Run PR validation tests on windows

radical:
- Tests for PR validation are run for Windows, and Linux now.
- Docker dependent tests are disabled on windows here too, same as for helix
- Sort the list of tests


tests/Aspire.Hosting.Tests/SlimTestProgramTests.cs
```diff
@@ -18,6 +19,7 @@ public SlimTestProgramTests(SlimTestProgramFixture slimTestProgramFixture)
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/7923", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnGithubActions), nameof(PlatformDetection.IsWindows))]
```
eerhardt, line 22: I mentioned it in https://github.com/dotnet/aspire/issues/7923, but maybe these should just have `[RequiresDocker]` attributes?

tests/Aspire.Hosting.Testing.Tests/TestingBuilderTests.cs
```diff
@@ -211,6 +211,7 @@ public async Task GetHttpClientBeforeStart(bool genericEntryPoint)
     [InlineData(false, true)]
     [InlineData(true, false)]
     [InlineData(true, true)]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/7930", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnGithubActions), nameof(PlatformDetection.IsWindows))]
```
eerhardt, line 214: Same here - This test requires docker from what I can see.

tests/Aspire.Cli.Tests/Hosting/CliOrphanDetectorTests.cs
```diff
@@ -93,6 +94,7 @@ public async Task CliOrphanDetectorAfterTheProcessWasRunningForAWhileThenStops()
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/7920", typeof(PlatformDetection), nameof(PlatformDetection.IsRunningOnGithubActions), nameof(PlatformDetection.IsWindows))]
```
eerhardt, line 97: I think this test requires docker.

.github/workflows/tests.yml
```diff
@@ -12,25 +12,45 @@ concurrency:
...
     outputs:
       tests_matrix: ${{ steps.generate_test_matrix.outputs.tests_matrix }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Get list of tests
+        env:
+          CI: false
+        run: >
```
eerhardt, line 33: for my learning, why do we sometimes use `|` and sometimes use `>`?

.github/workflows/tests.yml
```diff
@@ -12,25 +12,45 @@ concurrency:
...
+          ${{ matrix.dotnet_script }} build ${{ github.workspace }}/tests/Shared/GetTestProjects.proj
+          /p:TestsListOutputPath=${{ github.workspace }}/artifacts/TestsForGithubActions.list
+          /p:ContinuousIntegrationBuild=true
+
       - name: Generate tests matrix
         id: generate_test_matrix
+        shell: pwsh
```
eerhardt, line 40: powershell on linux... nice!

.github/workflows/run-tests.yml
```diff
@@ -22,76 +22,99 @@ on:
...
-    name: ${{ inputs.testShortName }}
+    name: ${{ matrix.name }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Trust HTTPS development certificate
-        run: ./dotnet.sh dev-certs https --trust
+        if: matrix.os == 'ubuntu-latest'
+        run: ${{ matrix.dotnet_script }} dev-certs https --trust
```
eerhardt, line 46: Why don't the dev-certs need to be trusted on Windows?

.github/workflows/tests.yml
radical: I was using `|` earlier as just continuing what was being used. `|` means that it keeps the newlines, so:
```
- run: |
     foo
     bar
```

.. would be `foo`, and `bar` as separate commands on the shell. And for unix we were using `\` at the end of the lines to concatenate them. But that doesn't work on windows, so I started using `>` - which by default converts the newlines to spaces, so the lines join to become one.

https://yaml-multiline.info

.github/workflows/run-tests.yml
radical: I wasn't able to get that running without prompt popping up. I seem to recall that you can pass `echo yes` or something to it, but couldn't get something working. And .. the tests seem to be passing anyway? So, I have this disabled right now.

.github/workflows/run-tests.yml
```diff
@@ -22,76 +22,99 @@ on:
...
         run: |
-          export OPTION_A=${{ github.workspace }}/tests/Aspire.${{ inputs.testShortName }}.Tests/Aspire.${{ inputs.testShortName }}.Tests.csproj; \
-          export OPTION_B=${{ github.workspace }}/tests/${{ inputs.testShortName }}.Tests/${{ inputs.testShortName }}.Tests.csproj; \
-          echo TEST_PROJECT_PATH=$(test -f $OPTION_A && echo $OPTION_A || echo $OPTION_B); \
-          echo TEST_PROJECT_PATH=$(test -f $OPTION_A && echo $OPTION_A || echo $OPTION_B) >> $GITHUB_ENV
+          $testShortName = "${{ inputs.testShortName }}"
+          $projectPath1 = "${{ github.workspace }}/tests/$testShortName.Tests/$testShortName.Tests.csproj"
+          $projectPath2 = "${{ github.workspace }}/tests/Aspire.$testShortName.Tests/Aspire.$testShortName.Tests.csproj"
+
+          if (Test-Path -Path $projectPath1) {
```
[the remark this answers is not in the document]
radical: I had it that way earlier, but any matrix inputs get serialized to the display line in PR checks. `Integration / test / (Hosting, tests/Aspire.Hosting.Tests/Aspire.Hosting.Tests.csproj)`, which made it less readable. So, I went with this.

tests/Aspire.Hosting.Tests/SlimTestProgramTests.cs
radical: Why are these not failing on the rolling builds then?

eerhardt: That's a good question. Are they running there?

radical: Yeah - 
Image
https://dev.azure.com/dnceng-public/public/_build/results?buildId=965263&view=ms.vss-test-web.build-test-results-tab

.github/workflows/run-tests.yml
```diff
@@ -22,76 +22,99 @@ on:
...
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Trust HTTPS development certificate
-        run: ./dotnet.sh dev-certs https --trust
+        if: matrix.os == 'ubuntu-latest'
+        run: ${{ matrix.dotnet_script }} dev-certs https --trust
 
       - name: Verify Docker is running
+        # nested docker containers not supported on windows
```
danmoseley, line 49: Really?

tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj
```diff
@@ -8,6 +8,8 @@
 
     <!-- no docker support on helix/windows yet -->
     <RunTestsOnHelix Condition="'$(OS)' != 'Windows_NT'">true</RunTestsOnHelix>
+    <!-- no docker support on gitbub-actions/windows yet -->
```
danmoseley, line 11:
```suggestion
    <!-- no docker support on github-actions/windows yet -->
```

danmoseley (approved):

.github/workflows/run-tests.yml
danmoseley: Oh just a GHA thing?

eerhardt (approved):
LGTM

PR #7919, opened 2025-03-06
Bootstrap more publishers

mitchdenny:
This PR introduces a few new packages:

- `Aspire.Hosting.Docker` (to contain a publisher for Docker Compose)
- `Aspire.Hosting.Kubernetes` (to contain a publisher for Kubernetes)

Some tweaks to the app model to support carrying the publisher name through to the execution context to allow for dynamically switching the publisher. A new playground project (copied initially from waitfor) to be the test bed for using custom publishers.

At this point the current publishers don't do anything they are just placeholders. The goal of this PR is to make the minimum changes necessary to make sure that existing publishing behavior continues to work whilst giving us a place to work on the new publisher interaction with the Aspire CLI.

There will be more changes, but this basic change will allow for some parallelism of effort.


src/Aspire.Hosting.Docker/DockerComposePublisher.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using Aspire.Hosting.ApplicationModel;
+using Aspire.Hosting.Publishing;
+
+namespace Aspire.Hosting.Docker;
+
+/// <summary>
+/// TODO
+/// </summary>
+public sealed class DockerComposePublisher : IDistributedApplicationPublisher
```
davidfowl, line 12: 👍🏾

src/Aspire.Hosting.Docker/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Install the package
+
+In your AppHost project, install the .NET Aspire Kubernetes Hosting library with [NuGet](https://www.nuget.org):
+
+```dotnetcli
+dotnet add package Aspire.Hosting.Kubernetes
+```
+
+## Configure Azure Provisioning for local development
```
davidfowl, line 19: ??

mitchdenny: Need to clean up some copy pasta ;)

davidfowl (approved):

src/Aspire.Hosting.Docker/README.md
```diff
@@ -0,0 +1,30 @@
...
+dotnet add package Aspire.Hosting.Docker
+```
+
+## Usage example
+
+Then, in the _Program.cs_ file of `AppHost`, register the publisher:
+
+```csharp
+builder.AddPublisher<DockerComposePublisher>("docker-compose");
+});
```
davidfowl, line 21:
```suggestion
```

davidfowl:
We need to follow up and add args passing to publishers

mitchdenny:
I'm thinking now that System.CommandLine is shipping with look at reintroducing it into the AppHost.

PR #7939, opened 2025-03-06
[CI] Enable Workload tests for PR validation

### LINKED ISSUE #7583: [Test] Re-enable Workload tests on PR runs

radical:
- Extract the existing job to enumerate the test class names to a custom action, so that it can be re-used
- Split the jobs so linux, and windows ones can run independently. And can have separate list of tests.
- Add a new workflow for running Workload tests, with a corresponding new action `enumerate-template-tests`
- Disable tests on windows/CI that need to trust the SSL certificate
- Also, add an artifact with all the test results for `Integration`, and `Template` workflows.

With this PR the `setup` jobs run independently for each OS, thus allowing their test jobs to start running immediately. This helps because Windows job can be a little slower, and doesn't block the Linux test jobs.

Fixes https://github.com/dotnet/aspire/issues/7583 .

### DIFF
--- a/.github/workflows/template-tests.yml
+++ b/.github/workflows/template-tests.yml
@@ -81,3 +81,16 @@ jobs:
       - name: Compute result
         run: |
           [ 0 -eq $(find test-job-result -name 'result-failed-*' | wc -l) ]
+
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: logs-*
+          merge-multiple: true
+          path: testresults
+
+      - name: Upload test result
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: Template-TestResults
+          path: testresults/**/*.trx
--- a/.github/workflows/template-tests.yml
+++ b/.github/workflows/template-tests.yml
@@ -31,8 +31,8 @@ jobs:
       tests_matrix: ${{ steps.generate_test_matrix.outputs.tests_matrix }}
     steps:
       - uses: actions/checkout@v4
-      - id: generate_test_matrix
-        uses: ./.github/actions/enumerate-template-tests
+      - uses: ./.github/actions/enumerate-template-tests
+        id: generate_test_matrix
 
   test_linux:
     name: Linux
--- a/.github/workflows/template-tests.yml
+++ b/.github/workflows/template-tests.yml
@@ -1,0 +1,83 @@
+name: Templates
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'release/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Duplicated jobs so their dependencies are not blocked on both the
+  # setup jobs
+
+  setup_for_template_tests_lin:
+    name: Setup for template tests (Linux)
+    runs-on: ubuntu-latest
+    outputs:
+      tests_matrix: ${{ steps.generate_test_matrix.outputs.tests_matrix }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/enumerate-template-tests
+        id: generate_test_matrix
+
+  setup_for_template_tests_win:
+    name: Setup for template tests (Windows)
+    runs-on: windows-latest
+    outputs:
+      tests_matrix: ${{ steps.generate_test_matrix.outputs.tests_matrix }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: generate_test_matrix
+        uses: ./.github/actions/enumerate-template-tests
+
+  test_linux:
+    name: Linux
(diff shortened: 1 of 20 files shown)


davidfowl:
Can you rename these tests while you’re at it

radical:
> Can you rename these tests while you’re at it

I'll have a separate PR for that. There are a whole bunch of places with `Workload` in the name.

radical:
@eerhardt  The `Final results` job in the `Templates` workflow will need to be marked as `Required` too.

radical:
Currently, we build the packages on each of the template test jobs to save time, instead of building once and uploading+downloading in the test jobs.

radical:
We can explore using https://github.com/actions/cache in the future.

tests/Aspire.Workload.Tests/StarterTemplateProjectNamesTests.cs
```diff
@@ -1,50 +1,78 @@
...
-    public StarterTemplateProjectNamesTests(ITestOutputHelper testOutput)
+    private readonly string _testType;
+    public StarterTemplateProjectNamesTests(string testType, ITestOutputHelper testOutput)
         : base(testOutput)
     {
+        _testType = testType;
     }
 
-    public static TheoryData<string, string> ProjectNamesWithTestType_TestData()
+    public static IEnumerable<object[]> ProjectNamesWithTestType_TestData()
```
eerhardt, line 19: Why the move away from TheoryData?

tests/Aspire.Workload.Tests/StarterTemplateProjectNamesTests.cs
```diff
@@ -1,50 +1,78 @@
...
         _testOutput.WriteLine($"Checking the starter template project");
         await AssertStarterTemplateRunAsync(context, project, config, _testOutput);
 
         _testOutput.WriteLine($"Checking the starter template project tests");
-        await AssertTestProjectRunAsync(project.TestsProjectDirectory, testType, _testOutput, config);
+        await AssertTestProjectRunAsync(project.TestsProjectDirectory, _testType, _testOutput, config);
+    }
+}
+
+// Individual class for each test framework so the tests can run in separate helix jobs
```
eerhardt, line 51: Why are these split, but BuildAndRunTemplateTests aren't?

.github/workflows/template-tests.yml
```diff
@@ -0,0 +1,102 @@
...
+    needs: setup_for_tests_lin
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.setup_for_tests_lin.outputs.tests_matrix) }}
+    with:
+      testShortName: ${{ matrix.shortname }}
+      os: "ubuntu-latest"
+      testProjectPath: tests/Aspire.Workload.Tests/Aspire.Workload.Tests.csproj
+      testSessionTimeoutMs: 1200000
+      testHangTimeout: 12m
```
eerhardt, line 49: With a 20 minute session timeout, is having 12 minute timeout for each test too much?

tests/Aspire.Workload.Tests/StarterTemplateProjectNamesTests.cs
radical: These had a natural split based on the template argument for the test type. I can split the others in a follow up based on how long they take to run. But that timing will be affected once we start using the `actions/cache`.

.github/workflows/tests.yml
```diff
@@ -11,61 +11,64 @@ concurrency:
...
-    runs-on: ${{ matrix.os }}
-    strategy:
-        matrix:
-          include:
-            - os: ubuntu-latest
-              dotnet_script: ./dotnet.sh
-            - os: windows-latest
-              dotnet_script: .\dotnet.cmd
+  # Duplicated jobs so their dependencies are not blocked on both the
+  # setup jobs
```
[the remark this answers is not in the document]
radical: On some runs (yesterday for example) it was 40s vs 4m.

tests/Aspire.Workload.Tests/StarterTemplateProjectNamesTests.cs
radical: Err leftover from a different change. Thank you, fixing it now!

.github/workflows/tests.yml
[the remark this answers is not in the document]
radical: I think it was nuget restore taking time essentially.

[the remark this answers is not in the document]
eerhardt: Hmm, I wonder if splitting this is actually going to help or hurt, since now it is 2 machines doing the restore instead of 1.

eerhardt (approved):
Let's get this in so these tests are protected. We can make fix ups if necessary afterwards.

PR #7940, opened 2025-03-06
Adding spectre.console to package deps and using it in Aspire.CLI

maddymontaquila:
## Description

- Added Spectre.Console package (latest version on nuget)
- Threw a try catch into Program.cs for UseOrFindAppHostProjectFile

cc @mitchdenny

### DIFF
--- a/src/Aspire.Cli/Program.cs
+++ b/src/Aspire.Cli/Program.cs
@@ -120,7 +120,8 @@ public class Program
             {
                 throw new InvalidOperationException("No project file found.");
             }
-            else {
+            else
+            {
                 return new FileInfo(projectFilePath);
             }
             
--- a/src/Aspire.Cli/Program.cs
+++ b/src/Aspire.Cli/Program.cs
@@ -128,7 +128,14 @@ public class Program
         catch (Exception ex)
         {
             Debug.WriteLine(ex.Message);
-            AnsiConsole.MarkupLine("[red bold]The --project option was not specified and no *.csproj files were detected.[/]");
+            if (projectFilePaths.Length > 1)
+            {
+                AnsiConsole.MarkupLine("[red bold]The --project option was not specified and multiple *.csproj files were detected.[/]");
+            }
+            else
+            {
+                AnsiConsole.MarkupLine("[red bold]The --project option was not specified and no *.csproj files were detected.[/]");
+            }
             return new FileInfo(Environment.CurrentDirectory);
         };
     }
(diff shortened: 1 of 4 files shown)


src/Aspire.Cli/Program.cs
```diff
@@ -111,8 +113,31 @@ private static FileInfo UseOrFindAppHostProjectFile(FileInfo? projectFile)
...
-        var projectFilePath = projectFilePaths.Single();
-        return new FileInfo(projectFilePath);
+        try 
+        {
+            var projectFilePath = projectFilePaths?.Single();
+            if (projectFilePath is null)
+            {
+                throw new InvalidOperationException("No project file found.");
+            }
+            else {
```
danmoseley, line 123:
```suggestion
            else
            {
```

danmoseley (approved):

src/Aspire.Cli/Program.cs
```diff
@@ -111,8 +113,24 @@ private static FileInfo UseOrFindAppHostProjectFile(FileInfo? projectFile)
...
+                return new FileInfo(projectFilePath);
+            }
+            
+        }
+        catch (Exception ex)
+        {
+            Debug.WriteLine(ex.Message);
+            AnsiConsole.MarkupLine("[red bold]The --project option was not specified and no *.csproj files were detected.[/]");
+            return new FileInfo(Environment.CurrentDirectory);
+        };
```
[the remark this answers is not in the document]
danmoseley: I agree with this, just check length rather than having Single throw?

src/Aspire.Cli/Program.cs
```diff
@@ -111,8 +113,32 @@ private static FileInfo UseOrFindAppHostProjectFile(FileInfo? projectFile)
...
 
         var projectFilePaths = Directory.GetFiles(Environment.CurrentDirectory, "*.csproj");
-        var projectFilePath = projectFilePaths.Single();
-        return new FileInfo(projectFilePath);
+        try 
+        {
+            var projectFilePath = projectFilePaths?.Single();
+            if (projectFilePath is null)
+            {
+                throw new InvalidOperationException("No project file found.");
```
eerhardt, line 121: I would suggest not using exceptions for control flow. Just write to the console here.

maddymontaquila: this was a hack because it was throwing before it could get to the point where it writes to the console... @mitchdenny will fix it lol

src/Aspire.Cli/Program.cs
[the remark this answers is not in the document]
maddymontaquila: I love when copilot copilots copilot code

mitchdenny (approved):

PR #7944, opened 2025-03-06
Adding workflow to update dependencies weekly

joperezr:
## Description

This adds a workflow which runs weekly and updates the NuGet package dependencies in the project.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] No
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No


.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,48 @@
...
+          git checkout -- .
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: update-dependencies
+          commit-message: "[Automated] Updating package dependencies"
+          title: "[Automated] Update dependencies"
+          body: "Auto-generated update to the package dependencies. In order for this PR to be green, it will require all of the new dependencies to be mirrored to our AzDO NuGet feeds."
```
danmoseley, line 48: should this comment say "but before merging revert changes to XYZ and ABC"

danmoseley: or "consult Jose" :)

danmoseley (approved):

.github/workflows/update-dependencies.yml
joperezr: Sure, I can add that as well.

.github/workflows/update-dependencies.yml
```diff
@@ -45,4 +45,4 @@ jobs:
           branch: update-dependencies
           commit-message: "[Automated] Updating package dependencies"
           title: "[Automated] Update dependencies"
-          body: "Auto-generated update to the package dependencies. In order for this PR to be green, it will require all of the new dependencies to be mirrored to our AzDO NuGet feeds."
\ No newline at end of file
+          body: "Auto-generated update to the package dependencies. In order for this PR to be green, it will require all of the new dependencies to be mirrored to our AzDO NuGet feeds. Any updates to dependencies that begin with `System.*` or `Microsoft.*` should be reverted as those will be updated using regular dependency flow."
```
danmoseley, line 48: what about `Microsoft.Azure.` `Microsoft.Data.SqlClient` `Microsoft.FluentUI` `Microsoft.Playwright` etc .. those won't be updated by codeflow?

Probably it should be a bit more explicit. Maybe it could be clearly commented in the directory.packages.props -- a separate commented section?

joperezr: They are already in their own separate section in Directory.Packages.props but I'll make this more explicit.

.github/workflows/update-dependencies.yml
```diff
@@ -0,0 +1,48 @@
...
+          git checkout -- .
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: update-dependencies
+          commit-message: "[Automated] Updating package dependencies"
+          title: "[Automated] Update dependencies"
+          body: "Auto-generated update to the package dependencies. In order for this PR to be green, it will require all of the new dependencies to be mirrored to our AzDO NuGet feeds. Any updates made outside of the first itemgroup in Directory.Packages.props should be reverted as those are packages that get updated through arcade's dependency flow."
```
eerhardt, line 48: Is there a way for us to ensure that dotnet outdated doesn't update outside of the first itemgroup?

eerhardt (commented):
I assume we are going to turn off dependabot so we don't get duplicated PRs? Should we just do that at the same time?

joperezr:
> I assume we are going to turn off dependabot so we don't get duplicated PRs? Should we just do that at the same time?

Yup, I can take care of that when this is merged.

.github/workflows/update-dependencies.yml
joperezr: I don't think there is other than manually adding either an include or exclude list with the packages that are updateable. We could potentially calculate that include list dynamically as part of the action in a previous step, but not sure if that is overkill (given we will need to review these updates anyway manually)

eerhardt: I guess we will see how annoying this gets. Maybe we will add an exclude list in the future so it doesn't mess with the dependencies to aspnetcore and runtime.

eerhardt:
> Yup, I can take care of that when this is merged.

Why not at the same time?

joperezr:
> > Yup, I can take care of that when this is merged.
> 
> 
> 
> Why not at the same time?

I was under the impression that that was configured via Settings of the GH repo, but I now understand that that is only the dependabot security updates, so I will include in this change the removal of the dependabot yml configuration file.

danmoseley:
Dependabot also updated global tools and maybe other stuff we want to keep not just package references (?)

eerhardt:
> Dependabot also updates global tools and maybe other stuff we want to keep not just package references (?)

Agreed. We can just turn off dependabot for nuget references.

sebastienros:
Here is what I had done for Orchard with dotnet outdated. Maybe one thing to take is the major version lock arguments and the fact it won't try to create a PR if there are no updates detected (but probably the PR actions takes care of it)

https://github.com/OrchardCMS/OrchardCore/compare/main...sebros/outdated

Since this Orchard has been using Renovate which handles everything with all the rules we want.

joperezr:
> Here is what I had done for Orchard with dotnet outdated. Maybe one thing to take is the major version lock arguments and the fact it won't try to create a PR if there are no updates detected (but probably the PR actions takes care of it)
> 
> [OrchardCMS/OrchardCore@main...sebros/outdated](https://github.com/OrchardCMS/OrchardCore/compare/main...sebros/outdated)
> 
> Since this Orchard has been using Renovate which handles everything with all the rules we want.

Thanks for sharing! I specifically decided not to use version lock on major, because we've said that, with our third party dependencies, we are okay with bumping them even across major versions. This is the current policy we use for our dependencies.

joperezr:
This should be good to go now.

davidfowl:
image

Broke the build.

joperezr:
I have #8093 fixing this. FWIW, it only broke the dependabot leg, not the actual build.

PR #7961, opened 2025-03-08
Add Functions host support to AzureTableStorageResource

captainsafia:
Azure Tables are a dependency for Durable Functions in deployments. We need to implement the `IResourceWithAzureFunctionsConfig` on the AzureStorageTableResource in order to support this.


davidfowl (approved):

davidfowl:
No test updates?

captainsafia:
> No test updates?

Our test coverage for these has usually been E2E tests with a trigger. We don't have unit tests for this behavior.

I can add some in a follow-up PR.

PR #7962, opened 2025-03-08
Add WithHttpCommand

### LINKED ISSUE #6649: WithHttpsCommand extension

_No response_

During .NET Conf 2024 @DamianEdwards showed a `WithHttpsCommand` extension method. This allows you to add commands that simply invoke HTTP based API's in your project. This feels like a generally useful extension method that deserves a place in the product.

### Additional context

_No response_

DamianEdwards:
## Description

Adds `WithHttpCommand()` to allow easily adding a resource command that sends an HTTP request to an endpoint.

Fixes #6649 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Ye
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] No
        - Link to aspire-docs issue: dotnet/aspire-docs#2779


src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+            return Task.CompletedTask;
+        });
+
+        builder.WithCommand(commandName, displayName,
+            async context =>
+            {
+                if (!endpoint.IsAllocated)
+                {
+                    return new ExecuteCommandResult { Success = false, ErrorMessage = "Endpoints are not yet allocated." };
+                }
```
davidfowl, line 1343: Does it make sense to instead not enable the command if the endpoint isn't allocated?

DamianEdwards: It does that already (see the `updateState` arg below), but this is the actual execute method so it's too late to disable it. If we somehow hit this case, this code means we'll just return a failed command result.

JamesNK: There could be a delay between state being updated, the user clicking the button, and the request hitting the server. Server validation of state is still good.

playground/TestShop/CatalogDb/ResetDbExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        var resetDbKey = app.Configuration["DatabaseResetKey"];
+        if (!string.IsNullOrEmpty(resetDbKey))
+        {
+            app.MapPost("/reset-db", async ([FromHeader(Name = "Authorization")] string? key, CatalogDbContext dbContext, CatalogDbInitializer dbInitializer, CancellationToken cancellationToken) =>
+            {
+                if (!string.Equals(key, $"Key {resetDbKey}", StringComparison.Ordinal))
+                {
+                    return Results.Unauthorized();
+                }
```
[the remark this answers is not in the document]
DamianEdwards: I can log an issue to track that idea.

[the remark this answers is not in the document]
DamianEdwards: #7984

tests/testproject/TestProject.ServiceA/Program.cs
```diff
@@ -8,6 +8,8 @@
 var app = builder.Build();
 
 app.MapGet("/", () => "Hello World!");
+app.MapGet("/get-only", () => Results.Ok());
```
eerhardt, line 11: Do we need to modify this? What uses this?

DamianEdwards: Nothing, it was added as potential target for a test but I didn't end up using it. Will remove.

src/Aspire.Hosting/ApplicationModel/ResourceCommandAnnotation.cs
```diff
@@ -122,7 +122,19 @@ public static class CommandResults
...
+    /// Produces an unsuccessful result with an error message.
+    /// </summary>
+    /// <param name="errorMessage">An optional error message.</param>
+    public static ExecuteCommandResult Failure(string? errorMessage = null) => new() { Success = false, ErrorMessage = errorMessage };
+
+    /// <summary>
+    /// Produces an unsuccessful result from an <see cref="Exception"/>. <see cref="Exception.Message"/> is used as the error message.
+    /// </summary>
+    /// <param name="exception">The exception to get the error message from.</param>
+    public static ExecuteCommandResult Failure(Exception exception) => Failure(exception.Message);
```
eerhardt, line 137: Can we refactor the existing code that is creating `ExecuteCommandResult` instances to use these new methods? For example in the StressApp?

https://github.com/dotnet/aspire/blob/dc284ba1075b0f77ea445877a2167651fde2dafa/playground/Stress/Stress.AppHost/ResourceBuilderExtensions.cs#L53-L71

Actually, can that whole file go away now with this?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+        HttpMethod? method = null,
+        string? httpClientName = null,
+        Func<HttpRequestMessage, Task>? configureRequest = null,
+        Func<HttpResponseMessage, Task<ExecuteCommandResult>>? getCommandResult = null,
+        string? commandName = null,
+        string? displayDescription = null,
+        string? confirmationMessage = null,
+        string? iconName = null,
+        IconVariant? iconVariant = null,
+        bool isHighlighted = false)
```
eerhardt, line 1215: This is a big signature. What are the odds we are going to add another parameter to this in the future?

DamianEdwards: It's basically the existing signature for `WithCommand` plus the options specific to HTTP commands. What you suggest as an alternative?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+                    {
+                        return await getCommandResult(response).ConfigureAwait(false);
+                    }
+
+                    response.EnsureSuccessStatusCode();
+                    return CommandResults.Success();
+                }
+                catch (Exception ex)
+                {
+                    return CommandResults.Failure(ex);
```
eerhardt, line 1365: Do we need a test for a failure case?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+    /// <para>
+    /// The command will be added to the resource represented by <paramref name="builder"/>.
+    /// </para>
+    /// <para>
+    /// If no <paramref name="endpointSelector"/> is specified, the first HTTP endpoint found on the resource will be used.
+    /// HTTP endpoints with an <c>https</c> scheme are preferred over those with an <c>http</c> scheme. If no HTTP endpoint
+    /// is found on the resource, an exception will be thrown.
+    /// </para>
+    /// <para>
+    /// The supplied <paramref name="endpointSelector"/> may return an endpoint from a different resource to that which the command is being added to.
```
eerhardt, line 1270: Do we have a test for this? Returning an endpoint for a different resource.

DamianEdwards: Yes, see https://github.com/dotnet/aspire/pull/7962/files#diff-39660fc1865df29aaf6dd429fbf374c894ffe5a8330d6660706718438a0758e5R185

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+        where TResource : IResourceWithEndpoints
+    {
+        method ??= HttpMethod.Post;
+
+        endpointSelector ??= PreferredEndpointSelector(builder);
+
+        var endpoint = endpointSelector()
+            ?? throw new DistributedApplicationException($"Could not create HTTP command for resource '{builder.Resource.Name}' as the endpoint selector returned null.");
+
+        commandName ??= $"{endpoint.Resource.Name}-{endpoint.EndpointName}-http-{method.Method.ToLowerInvariant()}";
```
DamianEdwards, line 1316: Just found that this isn't unique enough as multiple calls that resolve to the same command name just override the previous one. I'll have to use the path as well.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+
+        return builder;
+    }
+
+    // These match the default endpoint names resulting from calling WithHttpsEndpoint or WithHttpEndpoint as well as the defaults
+    // created for ASP.NET Core projects with the default launch settings added via AddProject. HTTPS is first so that we prefer it
+    // if found.
+    private static readonly string[] s_httpSchemes = ["https", "http"];
+
+    private static Func<EndpointReference> PreferredEndpointSelector<TResource>(IResourceBuilder<TResource> builder, string[]? endpointNames = null)
```
eerhardt, line 1383: (super nit) - I wonder if this method makes sense all together. It has 2 modes: either you supplied a list of endpointNames, or you didn't. It doesn't really have common code between those 2 cases. Could split it into 2 methods, one that takes a `string[]` and one that doesn't.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+        var endpoint = endpointSelector()
+            ?? throw new DistributedApplicationException($"Could not create HTTP command for resource '{builder.Resource.Name}' as the endpoint selector returned null.");
+
+        commandName ??= $"{endpoint.Resource.Name}-{endpoint.EndpointName}-http-{method.Method.ToLowerInvariant()}";
+
+        var targetRunning = false;
+        builder.ApplicationBuilder.Eventing.Subscribe<BeforeStartEvent>((e, ct) =>
+        {
+            var rns = e.Services.GetRequiredService<ResourceNotificationService>();
+            var watchTask = Task.Run(async () =>
```
eerhardt, line 1322:
```suggestion
            _ = Task.Run(async () =>
```

This makes it obvious that this is "fire and forget".

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
         return builder.WithAnnotation(new ResourceCommandAnnotation(name, displayName, updateState ?? (c => ResourceCommandState.Enabled), executeCommand, displayDescription, parameter, confirmationMessage, iconName, iconVariant, isHighlighted));
     }
 
+    /// <summary>
```
mitchdenny, line 1143: Great XML doc comments!

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+    /// <para>
+    /// The <paramref name="configureRequest"/> callback will be invoked to configure the request before it is sent. This can be used to add headers or a request payload
+    /// before the request is sent.
+    /// </para>
+    /// <para>
+    /// The <paramref name="getCommandResult"/> callback will be invoked after the response is received to determine the result of the command invocation. If this callback
+    /// is not specified, the command will be considered succesful if the response status code is in the 2xx range.
+    /// </para>
+    /// </remarks>
+    public static IResourceBuilder<TResource> WithHttpCommand<TResource>(
```
mitchdenny, line 1201: Add an `<example />` or two.

eerhardt (approved):
Looks reasonable to me. I just had some questions/comments.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+        {
+            var rns = e.Services.GetRequiredService<ResourceNotificationService>();
+            var watchTask = Task.Run(async () =>
+            {
+                await foreach (var resourceEvent in rns.WatchAsync(ct).WithCancellation(ct))
+                {
+                    if (resourceEvent.Resource == endpoint.Resource)
+                    {
+                        var resourceState = resourceEvent.Snapshot.State?.Text;
+                        targetRunning = resourceState == KnownResourceStates.Running || resourceState == KnownResourceStates.RuntimeUnhealthy;
```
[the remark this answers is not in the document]
DamianEdwards: That's a good question. Do we think we should just enable the command once the resource is in the `Running` state then, meaning if it has health checks you can't execute HTTP commands against it unless it's healthy?

mitchdenny (approved):
This looks good to me. One piece of feedback about exposing the command execution context on the request configuration callback. It is a very long over-load but I think everything there is stuff that we'll be asked for if we don't add it.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: I think we do? See https://github.com/dotnet/aspire/pull/7962/files#diff-39660fc1865df29aaf6dd429fbf374c894ffe5a8330d6660706718438a0758e5R71-R75

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: Yeah it used to fall through if you supplied names to looking for the default named ones, but that's not the case anymore.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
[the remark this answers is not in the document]
mitchdenny: I think you should be able to run commands even if the resource is unhealthy. You can use the callback to determine whether a specific command should be enabled in a specific state.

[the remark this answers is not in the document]
DamianEdwards: > You can use the callback to determine whether a specific command should be enabled in a specific state.

The `updateState` callback is not exposed right now. It's an implementation of `WithHttpCommand`. To change it, you'd have to dig it out of the annotation. Do you think we should allow passing that callback through too?

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1139,6 +1140,289 @@ public static IResourceBuilder<T> WithCommand<T>(
...
+    /// </para>
+    /// </remarks>
+    public static IResourceBuilder<TResource> WithHttpCommand<TResource>(
+        this IResourceBuilder<TResource> builder,
+        string path,
+        string displayName,
+        [EndpointName] string? endpointName = null,
+        HttpMethod? method = null,
+        string? httpClientName = null,
+        Func<HttpRequestMessage, Task>? configureRequest = null,
```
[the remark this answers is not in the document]
DamianEdwards: @mitchdenny you vote for `Func<ExecuteCommandContext, HttpRequestMessage, Task>`/`Func<ExecuteCommandContext, HttpResponseMessage, Task<ExecuteCommandResult>>` or should we introduce `HttpCommandConfigureRequestContext` and `HttpCommandGetResultContext`? 😁

src/Aspire.Hosting/ResourceBuilderExtensions.cs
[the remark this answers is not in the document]
DamianEdwards: I added `updateState` so that the command state can be customized and a test to verify it.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: OK I split it up into `NamedEndpointSelector` and `DefaultEndpointSelector`

PR #7963, opened 2025-03-08
Docker Compose publishing

prom3theu5:
## Description

Starting to roll in the expansions to the publishers, to allow output in the various target asset types.
Ignore a lot of the processing within the Output generators for now, lots of changes to come - this is more for transparency / visibility atm.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


src/Aspire.Hosting.Docker/Aspire.Hosting.Docker.csproj
```diff
@@ -7,6 +7,7 @@
     <PackageTags>aspire hosting kubernetes</PackageTags>
     <Description>Kubernetes publishing for .NET Aspire.</Description>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <NoWarn>$(NoWarn);CS8002</NoWarn>
```
prom3theu5, line 10: required due to k8s client not being signed

src/Aspire.Hosting/Aspire.Hosting.csproj
```diff
@@ -73,6 +73,8 @@
     <InternalsVisibleTo Include="Aspire.Hosting.Testing.Tests" />
     <InternalsVisibleTo Include="Aspire.Hosting.Containers.Tests" />
     <InternalsVisibleTo Include="Aspire.Cli.Tests" />
+    <InternalsVisibleTo Include="Aspire.Hosting.Docker" />
+    <InternalsVisibleTo Include="Aspire.Hosting.Kubernetes" />
```
davidfowl, line 77: We don't want IVT to product code, only tests. Why do you need this?

src/Aspire.Hosting.Docker/DockerComposePublisherOptions.cs
```diff
@@ -12,4 +12,9 @@ public sealed class DockerComposePublisherOptions
     /// The container registry to use.
     /// </summary>
     public string? DefaultContainerRegistry { get; set; }
-}
\ No newline at end of file
+
+    /// <summary>
+    /// Gets or sets the file system path where the generated Kubernetes artifacts will be saved.
```
davidfowl, line 17:
```suggestion
    /// Gets or sets the file system path where the generated docker compose artifacts will be saved.
```

src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
```diff
@@ -0,0 +1,224 @@
...
+namespace Aspire.Hosting.Docker;
+
+/// <summary>
+/// Contextual information used for manifest publishing during this execution of the AppHost as docker compose output format.
+/// </summary>
+/// <param name="executionContext">Global contextual information for this invocation of the AppHost.</param>
+/// <param name="outputPath">Output path for assets generated via this invocation of the AppHost.</param>
+/// <param name="logger">The current publisher logger instance.</param>
+/// <param name="cancellationToken">Cancellation token for this operation.</param>
+public sealed class DockerComposePublishingContext(DistributedApplicationExecutionContext executionContext, string outputPath, ILogger logger, CancellationToken cancellationToken = default)
```
davidfowl, line 20: Keep everything internal. We don't want public API yet.

src/Aspire.Hosting/Aspire.Hosting.csproj
prom3theu5: Its not - sorry, just made it easier for me to play around with the annotations before I'd added the execution context ^^.- will be removing

src/Aspire.Hosting.Kubernetes/Generators/HelmOutputGenerator.cs
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        var imageRepo = container.Image ?? throw new DistributedApplicationException("Container image is null, and cannot be published.");
+        var imageTag = "latest";
+        if (imageRepo.Contains(':'))
+        {
+            var idx = imageRepo.LastIndexOf(':');
+            imageTag = imageRepo[(idx + 1)..];
+            imageRepo = imageRepo[..idx];
+        }
```
davidfowl, line 379: Is this usable:

https://github.com/dotnet/aspire/blob/7834ae88d2b3ed8895e9bbb058d21ffd3491ff56/src/Aspire.Hosting/Utils/ContainerReferenceParser.cs

src/Aspire.Hosting.Docker/Aspire.Hosting.Docker.csproj
davidfowl, line 10: What's this?

prom3theu5: Can't use k8s.models without it - as they still don't sign the KubernetesClient library
```bash
0>CSC: Error CS8002 : Referenced assembly 'KubernetesClient, Version=16.0.0.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.
0>------- Finished building project: Aspire.Hosting.Docker. Succeeded: False. Errors: 1. Warnings: 0
```

src/Aspire.Hosting.Kubernetes/Generators/HelmOutputGenerator.cs
prom3theu5: yep - cheers

prom3theu5: its internal unfortunately

davidfowl:
One of the things we will want to do is also look at modeling docker compose files, helm and k8s as resources much like how we do with the azure resources.

src/Aspire.Hosting.Docker/Aspire.Hosting.Docker.csproj
davidfowl: Why is the k8s client a reference here?

src/Aspire.Hosting.Kubernetes/Generators/HelmOutputGenerator.cs
davidfowl: Shared source 😄

src/Aspire.Hosting.Docker/Aspire.Hosting.Docker.csproj
prom3theu5: It was 50/50 on the best way to take the writer.

YamlDotNet doesn't expose a stream based writer like STJ does. It has the concept of an Emitter, but its terrible to use, and results in a massive amount of code to do very little.
Rather than working with just a Dictionary<string, object> at the root, I was just using it as it was used in aspirate, where I'm reusing k8s types, and outputting those transformer as compose entries. However plan is to introduce strong types for sections of the compose and directly serialize those once a Compose object is created

src/Aspire.Hosting.Kubernetes/Generators/HelmOutputGenerator.cs
prom3theu5: Haha - I'll drop it in to each when i get to the image processing stage 👅

davidfowl: https://github.com/dotnet/aspire/tree/main/src/Shared

src/Aspire.Hosting.Docker/DockerComposePublisher.cs
```diff
@@ -4,24 +4,43 @@
...
+            throw new DistributedApplicationException(
+                "The '--output-path [path]' option was not specified even though '--publisher docker-compose' argument was used."
+            );
+        }
+
+        var context = new DockerComposePublishingContext(executionContext, publisherOptions.OutputPath, logger, cancellationToken);
+
+        await context.WriteModel(model).ConfigureAwait(false);
+
+        lifetime.StopApplication();
```
mitchdenny, line 44: I'll have a PR in over the weekend which means you don't need to do this. We'll hoist this up to the layer above.

src/Aspire.Hosting.Docker/DockerComposePublisherLoggerExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Aspire.Hosting.Docker;
+
+/// <summary>
+/// Provides a set of extension methods for interacting with ANSI console
+/// functionality within the Kubernetes hosting context.
+/// This static class facilitates additional behaviors or utility methods
+/// that extend the base ANSI console capabilities.
+/// </summary>
+internal static partial class DockerComposePublisherLoggerExtensions
```
mitchdenny, line 14: What do you think @davidfowl time for us to start adopting this across Aspire?

src/Aspire.Hosting.Docker/DockerComposePublisherOptions.cs
```diff
@@ -12,4 +12,9 @@ public sealed class DockerComposePublisherOptions
     /// The container registry to use.
     /// </summary>
     public string? DefaultContainerRegistry { get; set; }
```
mitchdenny, line 14: This was just a proof of concept configuration parameter, I'm not suggesting you need to have this in the DockerComposePublisher, but I'm guessing we need something like this.

src/Aspire.Hosting.Docker/DockerComposePublisher.cs
prom3theu5: Ok cheers mitch

src/Aspire.Hosting.Kubernetes/KubernetesPublisherOutputType.cs
```diff
@@ -0,0 +1,22 @@
...
+/// </summary>
+internal static class KubernetesPublisherOutputType
+{
+    public const string Helm = "helm";
+    public const string Kustomize = "kustomize";
+
+    public static string InvalidOutputTypeMessage(string outputType) =>
+        $"""
+        The '--output-type [type]' option was invalid. Supplied value was: '{outputType}'.
+        "Valid values are '{Helm}' and '{Kustomize}'.
```
mitchdenny, line 19: Are these sufficiently different that they should be different publishers?

prom3theu5: I don't think so. Flow will be identical, up to writing the output.
All options will be the same for instance

prom3theu5: Although that would probably change if we introduce modelled resources akin to what Azure.Provisioning produces, As we'd probably have the concept of things like HelmValue, HelmDictionaryOfT etc (and the same but Kustomize specific)

prom3theu5:
> One of the things we will want to do is also look at modeling docker compose files, helm and k8s as resources much like how we do with the azure resources. (Cdk like objects for each)

Thats probably a good call. It would - in turn, eventually allow compose and helm charts etc to be added to an AppHost as templates etc much like bicep supports.
We'd probably have to build a similar abstraction with properties wrapped in things like YamlValue, YamlDictionaryOfT etc

davidfowl:
> If we eventually went the AppHost route for addition / inclusion of those though, the provisioner would be pretty "full on" in terms of effort ^^

This is the plan 😄, but we should be incremental here so we can build up before we replicate the wrong pattern everywhere.

davidfowl:
Thoughts on https://cdk8s.io/ @prom3theu5 ?

prom3theu5:
> Thoughts on https://cdk8s.io/ @prom3theu5 ?

I am a fan. Also, c# support was recently added.
Personally, i've always tended to use Pulumi's Automation API though for C# based deployments when i'm asked to bundle things a dev team can use directly for things like local-dev etc [Example](https://github.com/pulumi/automation-api-examples/blob/main/dotnet/LocalProgram/automation/Program.cs) / [Docs](https://www.pulumi.com/docs/iac/using-pulumi/automation-api/)

I've also been following: [Timoni](https://timoni.sh/concepts/) quite closely

davidfowl:
> Are you thinking about including cdk8s projects in the AppHost?

Only if they had C# support and it was extensible (and not going to die 😄).

davidfowl:
We could contribute to it as well.

prom3theu5:
> > Are you thinking about including cdk8s projects in the AppHost?
> 
> Only if they had C# support and it was extensible (and not going to die 😄).

They do now - its not documented, but it was merged in 2 weeks ago ^^

src/Aspire.Hosting.Docker/DockerComposePublisherLoggerExtensions.cs
davidfowl: Ambivalent. Too much effort for little benefit at the moment. Can be done as cleanup on a class by class basis.

prom3theu5:
> These packages are in preview and we’ll keep it there until we get to a good approach. That said I’d like us to iterate on small commits, especially when we have something that we can try out.
> 
> what do you think is the next step here @prom3theu5? We should pick one end to end to flesh out the “problems” and interactions between the cli and publisher (including stare management and prompting etc)

Yep!

Compose is the easiest to work through, once thats in place - its relatively quick to work through the rest.
I've made a start firming that up - still a few issues i need to work through tomorrow night, but we are going to have to capture some input somehow if not all parameters and secrets have values set etc

src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
```diff
@@ -0,0 +1,521 @@
...
+        public async Task ProcessResourceAsync(DistributedApplicationExecutionContext executionContext, CancellationToken cancellationToken)
+        {
+            ProcessEndpoints();
+            ProcessVolumes();
+
+            await ProcessEnvironmentAsync(executionContext, cancellationToken).ConfigureAwait(false);
+            await ProcessArgumentsAsync(cancellationToken).ConfigureAwait(false);
+        }
+
+        private void ProcessEndpoints()
```
davidfowl, line 178: This logic should be dramatically simplified. Azure container apps has all sorts of limitations that cause us to determin what is supported up front. This should be super thin in the docker compose case.

src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
```diff
@@ -0,0 +1,521 @@
...
+                {
+                    var value = await ProcessValueAsync(kv.Value).ConfigureAwait(false);
+
+                    EnvironmentVariables.Add(new ComposeEnvironmentVariable(kv.Key, value.ToString()));
+                }
+
+                if (resource is ProjectResource)
+                {
+                    SetUpAspNetCoreUrlsForProject();
+                }
```
davidfowl, line 341: Why do you need this? This should be part of the project resource.

prom3theu5: Agreed. Will remove. Port allocator just needs to track target ports on the host not the container ports

src/Aspire.Hosting.Yaml/Aspire.Hosting.Yaml.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
davidfowl, line 1: Can this just be shared source? Do we need to expose these types publicly?

src/Aspire.Hosting.Docker/Aspire.Hosting.Docker.csproj
```diff
@@ -11,7 +11,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Aspire.Hosting\Aspire.Hosting.csproj" />
+    <ProjectReference Include="..\Aspire.Hosting.Yaml\Aspire.Hosting.Yaml.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="$(SharedDir)ContainerReferenceParser.cs" Link="Shared\ContainerReferenceParser.cs" />
```
davidfowl, line 18: This still needed?

src/Aspire.Hosting.Yaml/Aspire.Hosting.Yaml.csproj
prom3theu5: Not right now it doesn't have to be exposed. Could move to a yaml folder in shared, and ditch the need for a new package as well as publicly exposing it. Will do that for now until we need to

davidfowl (approved):

mitchdenny:
Thanks for all your efforts @prom3theu5

prom3theu5:
Cheers Mitch 😃

PR #7970, opened 2025-03-10
Move remaining ACA infrastructure generation to Aspire.Hosting.Azure.AppContainers

### LINKED ISSUE #7076: Move remaining ACA infrastructure generation to Aspire.Hosting.Azure.AppContainers
Today we generate bicep for each container app when PublishAsAzureContainerApp is called, but we still defer to azd to do the compute and networking infrastructure generation. We should move this to the same logic that allocates the container app resources. This allows the app host to mutate core infrastructure in C#.

davidfowl:
## Description

This change introduces a new `AzureContainerAppsEnviromentResource` that represents an Azure Container App environment. The goal is to move all of the generated infrastructure to the app host and out of azd. This allows users to model and tweak most things to do with deployment in C#, instead of dropping to bicep via infra synth.

Fixes #7076


src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppEnvironmentResource.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public BicepOutputReference ContainerRegistryManagedIdentityId => new("AZURE_CONTAINER_REGISTRY_MANAGED_IDENTITY_ID", this);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public BicepOutputReference ClientId => new("MANAGED_IDENTITY_CLIENT_ID", this);
```
eerhardt, line 42: Is it weird that "ClientId" and "ManagedIdentityId" point to different identities? Should ManagedIdentityId be renamed to include "Registry"?

src/Aspire.Hosting.Azure.AppContainers/ContainerAppEnvironmentDotnetComponentResource.cs
```diff
@@ -0,0 +1,39 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Azure.Provisioning.Primitives;
+
+namespace Azure.Provisioning.AppContainers;
+
+// This is missing from Azure.Provisioning.AppContainers
+internal sealed class ContainerAppEnvironmentDotnetComponentResource(string bicepIdentifier, string? resourceVersion = null) : ProvisionableResource(bicepIdentifier, new("Microsoft.App/managedEnvironments/dotNetComponents"), resourceVersion)
```
mitchdenny, line 9: When do we get the real provisioning types from Azure SDK?

playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
```diff
@@ -3,8 +3,17 @@
...
+using Azure.Provisioning.AppContainers;
+
 var builder = DistributedApplication.CreateBuilder(args);
 
+builder.AddContainerAppEnvironment("infra")
+       .ConfigureInfrastructure(infra =>
+       {
+           var env = infra.GetProvisionableResources().OfType<ContainerAppManagedEnvironment>().Single();
+           env.WorkloadProfiles.Clear();
+       });
```
mitchdenny, line 15: Thinking about this and the intersection of publishers. We will have a publisher for ACA-based deployments. I wonder if the `AddAzureContainerAppsPublisher(...)` we should have a callback in there where you can get at this rather than adding it as a resource?

src/Aspire.Hosting.Azure.AppContainers/ContainerAppEnvironmentDotnetComponentResource.cs
davidfowl: @tg-msft

playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
```diff
@@ -3,8 +3,17 @@
 
 #pragma warning disable ASPIREACADOMAINS001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
 
+using Azure.Provisioning.AppContainers;
+
 var builder = DistributedApplication.CreateBuilder(args);
 
+builder.AddContainerAppEnvironment("infra")
```
captainsafia, line 10: Are you limited to one container app environment per app host or is it possible to have multiple?

playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
```diff
@@ -39,7 +48,7 @@
        .WithEnvironment("VALUE", param)
        .PublishAsAzureContainerApp((module, app) =>
        {
-           app.ConfigureCustomDomain(customDomain, certificateName);
+           // app.ConfigureCustomDomain(customDomain, certificateName);
```
captainsafia, line 51: What doesn't work about this?

playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
davidfowl: The plan is to support multiple

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppEnvironmentResource.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+/// <param name="name"></param>
+/// <param name="configureInfrastructure"></param>
+public class AzureContainerAppEnvironmentResource(string name, Action<AzureResourceInfrastructure> configureInfrastructure) :
+    AzureProvisioningResource(name, configureInfrastructure)
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public BicepOutputReference ContainerAppEnvironmentId => new("AZURE_CONTAINER_APPS_ENVIRONMENT_ID", this);
```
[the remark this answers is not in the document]
davidfowl: Yep

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -23,4 +32,166 @@ public static IDistributedApplicationBuilder AddAzureContainerAppsInfrastructure
...
+
+        var containerAppEnvResource = new AzureContainerAppEnvironmentResource(name, infra =>
+        {
+            var principalId = new ProvisioningParameter("principalId", typeof(string));
+
+            infra.Add(principalId);
+
+            var principalType = new ProvisioningParameter("principalType", typeof(string));
+
+            infra.Add(principalType);
```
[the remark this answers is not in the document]
davidfowl: It is because azd handles secret outputs by extracting the secret from key vault and storing it in aca manually as a secret. That would happen if you had a secret output value within a larger reference expression.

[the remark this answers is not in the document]
davidfowl: This is practice a problem that went away when we put the full connection strings in the key vault but its still possible if you write some crazy code. I think we can just block this for now...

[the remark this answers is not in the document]
davidfowl: It still required when you are using azd up. The user is the one doing the deployment and they need to wire up role assignments so you can see the dashboard.

playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs
davidfowl: In this PR to simplify things and make progress, I've decided to throw if there's more than one.

playground/AzureContainerApps/AzureContainerApps.AppHost/infra/main.bicep
```diff
@@ -0,0 +1,76 @@
...
+    location: location
+    storage_outputs_name: storage.outputs.name
+  }
+}
+module infra 'infra/infra.module.bicep' = {
+  name: 'infra'
+  scope: rg
+  params: {
+    location: location
+    principalId: ''
```
davidfowl, line 58: @vhvb1989 This is the last issue with this PR. There's no way to access known parameters from main in the manifest so there's have no way to represent that I need the parameter principlId cleanly.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -34,13 +34,15 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
             return;
         }
 
+        // TODO: We need support direct association between a compute resource and the container app environment.
+        // Right now we support a single container app environment is the one we want to use and we'll fall back to 
+        // azd based environment if we don't have one.
```
eerhardt, line 39:
```suggestion
        // TODO: We need to support direct association between a compute resource and the container app environment.
        // Right now we support a single container app environment as the one we want to use and we'll fall back to 
        // azd based environment if we don't have one.
```

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -31,4 +43,266 @@ public static IDistributedApplicationBuilder AddAzureContainerAppsInfrastructure
...
+
+            var pullRa = containerRegistry.CreateRoleAssignment(ContainerRegistryBuiltInRole.AcrPull, identity);
+
+            // There's a bug in the CDK, see https://github.com/Azure/azure-sdk-for-net/issues/47265
+            pullRa.Name = BicepFunction.CreateGuid(containerRegistry.Id, identity.Id, pullRa.RoleDefinitionId);
+            infra.Add(pullRa);
+
+            var pushRa = containerRegistry.CreateRoleAssignment(ContainerRegistryBuiltInRole.AcrPush, identity);
+            pushRa.Name = BicepFunction.CreateGuid(containerRegistry.Id, identity.Id, pushRa.RoleDefinitionId);
+            infra.Add(pushRa);
```
eerhardt, line 100: Why is AcrPush needed?

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -31,4 +43,266 @@ public static IDistributedApplicationBuilder AddAzureContainerAppsInfrastructure
...
+                        Name = "consumption"
+                    }
+                ],
+                AppLogsConfiguration = new()
+                {
+                    Destination = "log-analytics",
+                    LogAnalyticsConfiguration = new()
+                    {
+                        CustomerId = laWorkspace.CustomerId,
+                        SharedKey = laWorkspace.GetKeys().PrimarySharedKey
```
eerhardt, line 125: I know this is how azd does it, but is there a way to not use secrets here? For the future - obviously we won't change this in this PR. But should we consider logging a bug to use MSI/Entra here?

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -31,4 +43,266 @@ public static IDistributedApplicationBuilder AddAzureContainerAppsInfrastructure
...
+            // Add the secret outputs to the key vaults
+            foreach (var (key, value) in kvs)
+            {
+                infra.Add(new ProvisioningOutput(key, typeof(string))
+                {
+                    Value = value.Name
+                });
+            }
+
+            infra.Add(new ProvisioningOutput("MANAGED_IDENTITY_NAME", typeof(string))
```
eerhardt, line 255: Looks like you missed `MANAGED_IDENTITY_CLIENT_ID`

playground/AzureContainerApps/AzureContainerApps.AppHost/infra/main.bicep
vhvb1989: How do you want to express it in the manifest?
 
We can do:
"princiaplId", as raw string, which becomes a reserved word for azd parsing the manifest
 
"{ .princiaplId }" , as a property of a global object.  Manifest in this case knows too much, maybe
 
"{ principalId }" Similar to the raw string, but using the brackets. 
 
"{ _publisher_.principalId }" .  Introduce the concept of a publisher in the manifest. An special way to request functionality for anyone parsing the manifest

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppEnvironmentResource.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// <summary>
+    /// Gets the default domain associated with the Container App Environment.
+    /// </summary>
+    public BicepOutputReference ContainerAppDomain => new("AZURE_CONTAINER_APPS_ENVIRONMENT_DEFAULT_DOMAIN", this);
+
+    /// <summary>
+    /// Gets the managed identity ID associated with the Container App Environment.
+    /// </summary>
+    public BicepOutputReference ManagedIdentityId => new("AZURE_CONTAINER_REGISTRY_MANAGED_IDENTITY_ID", this);
```
eerhardt, line 29: Is this supposed to be `clientId` or `principalId`?

The output name is wrong - it is dupicated with `ContainerRegistryManagedIdentityId` below.

eerhardt: It would be easier to analyze if these output references were grouped together in some coherent manner.

playground/AzureContainerApps/AzureContainerApps.AppHost/infra.module.bicep
```diff
@@ -0,0 +1,173 @@
...
+      name: 'PerGB2018'
+    }
+  }
+  tags: tags
+}
+
+resource cae 'Microsoft.App/managedEnvironments@2024-03-01' = {
+  name: take('cae${uniqueString(resourceGroup().id)}', 24)
+  location: location
+  properties: {
```
eerhardt, line 57: Is this up to date? Doing a diff with azd, I see workloadProfiles is missing (left is azd):

image

But I see code trying to set this.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -70,23 +72,50 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+
+                r.Parameters[key] = value;
+            }
+        }
+
+        // Resolve the known parameters for the container app environment
+        foreach (var r in appModel.Resources.OfType<AzureBicepResource>())
+        {
+            // HACK: This forces parameters to be resolved for any AzureProvisioningResource
+            r.GetBicepTemplateFile();
```
eerhardt, line 95: 🤮. What are the downsides of doing this here and again when really generating it?

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -70,23 +72,50 @@ public async Task BeforeStartAsync(DistributedApplicationModel appModel, Cancell
...
+        static void SetKnownParameterValue(AzureBicepResource r, string key, Func<AzureBicepResource, object> factory)
+        {
+            if (r.Parameters.TryGetValue(key, out var existingValue) && existingValue is null)
+            {
+                var value = factory(r);
+
+                if (value is BicepOutputReference b && b.Resource == r)
+                {
+                    return;
+                }
```
eerhardt, line 85: What is this check for? Can we add a comment here?

It would be odd that a parameter is pointing to an output of the same resource, isn't it?

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
davidfowl: I think that would be an ACA issue

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -31,4 +43,266 @@ public static IDistributedApplicationBuilder AddAzureContainerAppsInfrastructure
...
+            var principalId = new ProvisioningParameter("principalId", typeof(string));
+
+            infra.Add(principalId);
+
+            var tags = new ProvisioningParameter("tags", typeof(object))
+            {
+                Value = new BicepDictionary<string>()
+            };
+
+            infra.Add(tags);
```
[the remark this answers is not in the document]
davidfowl: I think we can do a clean up the bicep pass. None of our generated bicep has descriptions.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
davidfowl: In theory, you can add more parameters in another before start hook.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
davidfowl: I'll add a comment. It really is an odd case (it was because we're using principalId)

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
davidfowl: Do we still use this after your change?

playground/AzureContainerApps/AzureContainerApps.AppHost/infra.module.bicep
davidfowl: Undid my change.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
davidfowl: Removed

playground/AzureContainerApps/AzureContainerApps.AppHost/infra/account/account.module.bicep
```diff
@@ -0,0 +1,49 @@
+@description('The location for the resource(s) to be deployed.')
```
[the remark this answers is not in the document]
davidfowl: Yep!

playground/AzureContainerApps/AzureContainerApps.AppHost/infra/main.bicep
davidfowl: Agreed on a new known parameter "userPrincipalId"

davidfowl:
This is ready for review again @eerhardt @mitchdenny

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppEnvironmentResource.cs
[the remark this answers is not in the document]
davidfowl: I will do a pass in a follow up PR to clean up some of the implementation.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppEnvironmentResource.cs
davidfowl: So this identity is really the keyvault identity (I am going to do a pass to split the identity and improve the naming here). https://github.com/dotnet/aspire/issues/8133

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
davidfowl: We don't this comes from the per container app identity now.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
davidfowl: Removed!

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppExtensions.cs
```diff
@@ -301,6 +301,13 @@ public static IResourceBuilder<AzureContainerAppEnvironmentResource> AddAzureCon
             });
         });
 
+        if (builder.ExecutionContext.IsRunMode)
```
davidfowl, line 304: This is kinda ugly 😄.

eerhardt (approved):
Let's get this in.

PR #7980, opened 2025-03-10
Don't trigger codeql pipeline for preview/rc branches

wtgodbe:


danmoseley (approved):
we need to check what is required of us.

wtgodbe:
Confirmed it's alright to exclude preview branches, but we should still run codeql on RCs

PR #7983, opened 2025-03-10
Add policy to close & re-open github-action PRs

wtgodbe:
Reverts dotnet/aspire#7981, and adds `triggerOnOwnActions: true` so that PolicyBot can respond to its own events. Previous PR didn't work because the re-open step was skipped, as by default PolicyBot will not respond to its own events.

### DIFF
--- a/.github/policies/eventResponderTask.GithubActionPROpened.yml
+++ b/.github/policies/eventResponderTask.GithubActionPROpened.yml
@@ -28,7 +28,6 @@ configuration:
           label: 'Closed Github-Action PR'
       - closeIssue
       description: Close and label Github-Action PRs
-      triggerOnOwnActions: true
     - if:
       - payloadType: Pull_Request
       - isAction:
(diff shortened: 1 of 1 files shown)


.github/policies/eventResponderTask.GithubActionPROpened.yml
```diff
@@ -0,0 +1,44 @@
...
+          action: Opened
+      - isActivitySender:
+          user: github-actions[bot]
+          issueAuthor: False
+      - not:
+          hasLabel:
+            label: 'Re-opened Github-Action PR'
+      then:
+      - addLabel:
+          label: 'Closed Github-Action PR'
```
danmoseley, line 28: I'm wondering what i would take for this to get in a loop. Say, someone deletes this label, will it still close it? If so, then it will begin to loop. I guess the repo won't be destroyed ...

danmoseley (approved):

.github/policies/eventResponderTask.GithubActionPROpened.yml
wtgodbe: I only added `triggerOnOwnActions` to the step that re-opens PRs. The step to close PRs can't be triggered by PolicyBot's own actions, so we should be safe from loops

wtgodbe:
That worked! https://github.com/dotnet/aspire/pull/7985 was closed & re-opened by PolicyBot, and the github action tests started running.

danmoseley:
do any other of the .NET repos need this? I guess not as they use Helix exclusively?

wtgodbe:
> do any other of the .NET repos need this? I guess not as they use Helix exclusively?

Not that I can think of - the mainline repos (sdk, aspnetcore, runtime, windowsdesktop) don't use github actions for CI, as far as I know

PR #7991, opened 2025-03-11
Backchannel between CLI and AppHost

mitchdenny:
This PR introduces a bidirectional RPC (using `StreamJsonRpc`) between the CLI and the AppHost. If the CLI isn't the thing invoking the AppHost then the backchannel doesn't spin up in the AppHost. In this PR they just sit there pinging each other but this will be used as the foundation for the AppHost sending information that the console will display (URLs etc).

I still haven't settled on how I want to expose the ability to call the CLI RPC from the AppHost yet. It might be that I move away from using a background service so that code can call the RPC directly via an interface.

Alternatively, the background service might be very constrained in what it sends and so it basicaly uses information in the app model to send information.

### DIFF
--- a/src/Aspire.Cli/DotNetCliRunner.cs
+++ b/src/Aspire.Cli/DotNetCliRunner.cs
@@ -3,10 +3,12 @@
 
 using System.Diagnostics;
 using System.Globalization;
+using System.Net.Sockets;
+using StreamJsonRpc;
 
 namespace Aspire.Cli;
 
-internal sealed class DotNetCliRunner
+internal sealed class DotNetCliRunner(CliRpcTarget cliRpcTarget)
 {
     internal Func<int> GetCurrentProcessId { get; set; } = () => Environment.ProcessId;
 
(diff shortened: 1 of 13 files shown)


src/Aspire.Hosting/Cli/CliBackchannel.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        var unixSocketPath = configuration.GetValue<string>(UnixSocketPathEnvironmentVariable);
+
+        if (string.IsNullOrEmpty(unixSocketPath))
+        {
+            logger.LogDebug("Aspire CLI backchannel socket path was not specified.");
+            return;
+        }
+
+        logger.LogInformation("Aspire CLI backchannel socket path: {SocketPath}", unixSocketPath);
```
davidfowl, line 27: Add a Task.Yield here to force this to the background

src/Aspire.Cli/DotNetCliRunner.cs
```diff
@@ -3,33 +3,85 @@
...
+            } while (await timer.WaitForNextTickAsync(cancellationToken).ConfigureAwait(false));
+        }
+        catch (OperationCanceledException ex)
+        {
+            logger.LogDebug(ex, "Shutting down AppHost backchannel because of cancellation.");
+            return;
+        }
+        catch (Exception ex)
+        {
+            _ = ex;
```
davidfowl, line 80: Log?

davidfowl (approved):

src/Aspire.Hosting/Cli/CliBackchannel.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            logger.LogDebug("Ignoring ConnectionLostException because of cancellation.");
+            return;
+        }
+        catch (OperationCanceledException ex)
+        {
+            logger.LogDebug(ex, "Shutting down CLI backchannel because of cancellation.");
+            return;
+        }
+    }
```
davidfowl, line 60: Log?

src/Aspire.Hosting/Cli/CliBackchannel.cs
```diff
@@ -0,0 +1,61 @@
...
+            do
+            {
+                var sendTimestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+                var responseTimestamp = await rpc.InvokeAsync<long>("PingAsync", sendTimestamp).ConfigureAwait(false);
+                Debug.Assert(sendTimestamp == responseTimestamp);
+                var roundtripMilliseconds = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - sendTimestamp;
+                logger.LogDebug("PingAsync round trip time is {RoundTripDuration} ms", roundtripMilliseconds);
+            } while(await timer.WaitForNextTickAsync(stoppingToken).ConfigureAwait(false));
+        }
+        catch (StreamJsonRpc.ConnectionLostException) when (stoppingToken.IsCancellationRequested)
```
davidfowl, line 50: Shouldn't it reconnect?

davidfowl: OK well log everything 😄

src/Aspire.Hosting/Cli/CliBackchannel.cs
mitchdenny: I think `LogDebug` is appropriate in this case. We generally don't want to see this output unless we are trying to debug a problem with the backchannel itself. We wouldn't want these errors in general informational output I don't think.

src/Aspire.Hosting/Cli/CliBackchannel.cs
mitchdenny: Adding more logging to both sides.

PR #7994, opened 2025-03-11
Add default icons for resource type

glennc:
## Description

image

- Adds icons to the resource grid that are the same as the ones from the graph.
- Hides the resource type column, resource type can be toggled in options and is the tooltip for the icon.

I started out thinking about how to add custom icons for a resource, so that Redis for example could use the Redis logo. This felt like a good first part of that. After this, if we want, we can add an annotation for an icon and add it to the protocol. Then this would be the fallback for when there is no custom icon or if the resource server doesn't implement custom icons. A dashboard running in prod would not have the icons for example.

I thought this PR would at least let us talk about if we want to do this or not and start progressing on the idea.

Some notes:

- Size: technically we should probably request a size from the icon helper and have it return the right Icon for that size since the actual icon can and does change for each size. However not all the icons exist for all sizes and the differences didn't seem to make a huge deal. So I left the size as the same one as we use in the graph and reduce it with styles.
- Color: I would have liked to not use WithColor and therefore avoid instantiating a new icon for each invocation, but it wasn't obvious to me how to do that with FluentIcon component. All the things I tried got messed with in some way. Maybe someone else can tell me.

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] No
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No (we probably should update screenshots eventually but it's fine)


src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using Microsoft.FluentUI.AspNetCore.Components;
+using Icons = Microsoft.FluentUI.AspNetCore.Components.Icons;
+
+internal static class ResourceIconHelpers
+{
+    /// <summary>
+    /// Maps a resource to a default icon.
+    /// </summary>
+    public static Icon GetIconForResource(ResourceViewModel resource)
```
davidfowl, line 14: These icons were all static before now it's creating them up per call.

mitchdenny:
I think there needs to be a tad more space between the icon and the text.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -385,6 +386,28 @@ private void UpdateMenuButtons()
                 Icon = new Icons.Regular.Size16.EyeOff()
             });
         }
+
+        if (_showResourceTypeColumn)
+        {
+
+             _resourcesMenuItems.Add(new MenuButtonItem
```
JamesNK, line 393:
```suggestion
             _resourcesMenuItems.Add(new MenuButtonItem
```

src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
```diff
@@ -6,18 +6,22 @@
...
-{
-    <FluentIcon
-        Icon="Icons.Regular.Size16.Pin"
-        Title="@Loc[nameof(Columns.PersistentContainerIconTooltip)]"
-        Color="Color.Neutral"
-        Class="persistent-container-icon" />
-}
+<span class="resource-name-container">
+    <FluentIcon Width="16px" Class="resource-icon "
+        Value="@ResourceIconHelpers.GetIconForResource(Resource).WithColor(ColorGenerator.Instance.GetColorHexByKey(FormatName(Resource)))" />
```
JamesNK, line 11: These are 24px icons displayed at 16px. It would be better if the 16px versions were used directly, with a fallback to shrinking a larger size icon if required. The 16px versions could have optimizations for smaller size.

src/Aspire.Dashboard/Components/Pages/Resources.razor
```diff
@@ -136,7 +136,7 @@
                                                     OnRowClick="@(r => r.ExecuteOnDefault(d => ShowResourceDetailsAsync(d.Resource, buttonId: null)))"
                                                     Class="main-grid enable-row-click">
                                         <ChildContent>
-                                            <AspireTemplateColumn ColumnId="@NameColumn" ColumnManager="@_manager" Title="@ControlsStringsLoc[nameof(ControlsStrings.NameColumnHeader)]" Sortable="true" SortBy="@_nameSort" Tooltip="true" TooltipText="@(c => GetResourceName(c.Resource))" Class="expand-col">
+                                            <AspireTemplateColumn ColumnId="@NameColumn" ColumnManager="@_manager" Title="@ControlsStringsLoc[nameof(ControlsStrings.NameColumnHeader)]" Sortable="true" SortBy="@_nameSort" Tooltip="true" TooltipText="@(c => c.Resource.ResourceType)" Class="expand-col">
```
JamesNK, line 139: The tooltip should still include the resource name.
```suggestion
                                            <AspireTemplateColumn ColumnId="@NameColumn" ColumnManager="@_manager" Title="@ControlsStringsLoc[nameof(ControlsStrings.NameColumnHeader)]" Sortable="true" SortBy="@_nameSort" Tooltip="true" TooltipText="@(c => $"{c.Resource.ResourceType} {GetResourceName(c.Resource)})" Class="expand-col">
```

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
```diff
@@ -154,7 +155,7 @@ protected override async Task OnInitializedAsync()
             new GridColumn(Name: NameColumn, DesktopWidth: "1.5fr", MobileWidth: "1.5fr"),
             new GridColumn(Name: StateColumn, DesktopWidth: "1.25fr", MobileWidth: "1.25fr"),
             new GridColumn(Name: StartTimeColumn, DesktopWidth: "1fr"),
-            new GridColumn(Name: TypeColumn, DesktopWidth: "1fr"),
+            new GridColumn(Name: TypeColumn, DesktopWidth: "1fr", IsVisible: () => _showResourceTypeColumn),
```
JamesNK, line 158: What do folks feel about having an option to still show the type column? Would anyone use it?

JamesNK:
> I think there needs to be a tad more space between the icon and the text.

Yes. Add margin-right to `.resource-icon`.

src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
glennc: Yeah because I couldn't find a better way to do colors. If someone has a better way I am all ears. Color is an instance property. Maybe some CSS selector would work and I can spend some time looking for one if we are worried about it.

src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
glennc: I think what this would look like is GetIconForResource takes size and returns the closest size we can to the one you asked for. Most of the icons we use here don't have 16px versions IIRC. So they would be a mix of 20/24px and you would need to know to request a size and also set width. I didn't do it this way because of that complexity in the API. The API would look like you could request a size and you would get what you asked for. if you assumed that you would be surprised. Assuming I'm not wrong about what you had in mind I will do this and try and make it as obvious as I can that you might not get the size you are requesting.

src/Aspire.Dashboard/Components/Pages/Resources.razor.cs
glennc: The main reason I didn't just delete it was that I didn't have a better default sort column unless we made the Icon its own column. It's less usable as it is now in terms of getting back to default sort order if you sort by something else, you would need to make the column visible and then sort by it to get back to the default.

I don't feel strongly about this though. Just seemed like not super useful when the information was there as tooltips and icons.

JamesNK: If we hide and show this column then it should be persisted to local storage. That means visibility is preseved when starting/stopping Aspire.

src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        return resource.ResourceType switch
+        {
+            KnownResourceTypes.Executable => new Icons.Filled.Size24.Database(),
+            KnownResourceTypes.Project => new Icons.Filled.Size24.CodeCircle(),
+            KnownResourceTypes.Container => new Icons.Filled.Size24.Box(),
+            string t => t.Contains("database", StringComparison.OrdinalIgnoreCase) ? new Icons.Filled.Size24.Database() : new Icons.Filled.Size24.SettingsCogMultiple(),
```
adamint, line 21: All databases contain "database" in the type? no "db"?

src/Aspire.Dashboard/wwwroot/css/app.css
```diff
@@ -752,3 +752,7 @@ fluent-switch.table-switch::part(label) {
     max-height: 400px;
     overflow-y: auto;
 }
+
+.resource-icon {
+    vertical-align: text-bottom;
```
adamint, line 757: Use .align-text-bottom unless you are planning on adding additional styling

src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
CZEMacLeod: Would be nice to be able to have an attribute or annotation on the resource type to set the icon from the apphost....

adamint: Yeah, I agree with this. @glennc are you comfortable making that change, or would you like me to?

src/Aspire.Dashboard/wwwroot/css/app.css
glennc: I think we need to add space to the right as well as per the other comments.

src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
glennc: I viewed that as a different PR after this one, because it would require protocol changes. Happy to hand this PR over and do it all in one go if you want though? The logic as it is now copied from the graph. but that would obviously also update if we did the custom icons now.

adamint: If you handed it off, release would be delayed until 9.3. Just want to set expectatiosn

adamint: For now I think this is fine

src/Aspire.Dashboard/Components/ResourcesGridColumns/ResourceNameDisplay.razor
adamint: Reverting my previous thought. I just checked in the fluentui-system-icons repo, and I do not think it's going to be reasonable to request an additional size for the several icons missing a 16px version.

Please use size 16 for the icons that support that size, and downscale the others

davidfowl:
Can we get this in for 9.2?

adamint:
> Can we get this in for 9.2?

I don’t think there are any additional changes needed. Just need to fix merge conflicts and I can approve

JamesNK:
I would try this out and review it again before it's merged.

src/Aspire.Dashboard/Model/ResourceIconHelpers.cs
adamint, line 21: nit: you could use a `when` expression to avoid the ternary operator

JamesNK (approved):

JamesNK:
Resolved merge conflicts.
Did various clean up.
Reused icon name that used for command icons. Added support for specifying a desired size. We'll want to use this when resources can specify an icon name.

JamesNK:
@adamint Can you fix these tests. I'm guessing changing the HTML in the resource grid requires a reaction in the tests.

Issue #6279, opened 2024-10-14
Display more friendly error message for common health checks errors


JamesNK:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

This is what is displayed for a health check that is validates whether an endpoint returns 200:

![Image](https://github.com/user-attachments/assets/c10884d5-8566-45ae-a246-bc2de2c20d17)

No message is returned so health checks defaults to displaying the exception message. This doesn't look user friendly.

The health check should have an error message like "Request to xxx timed out" if there is a timeout.  Or "Request to xxx returned yyy status code".
We should do the same for other build in checks: polling an endpoint for 200, a database for a good response, or message queue, etc.

davidfowl:
Maybe we should do this [here](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks). That way we don't need to do anything special for these health checks.

PR #7999, opened 2025-03-11
Revert "Backchannel between CLI and AppHost (#7991)"

eerhardt:
This reverts commit 811341ddf033bbda9c3d07fa1390bcfa25cf70cf.

## Description

This change is bringing in [StreamJsonRpc](https://www.nuget.org/packages/StreamJsonRpc) as a new dependency to Aspire.Hosting. With this dependency brings a set of analyzers in [Microsoft.VisualStudio.Threading.Analyzers](https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers/) which are being enabled by default in any AppHost project. This was breaking our template tests because the analyzers are causing warnings in our templates. We don't want these analyzers enabled by default in .NET Aspire. Reverting the change to figure out how we can use StreamJsonRpc without enabling these analyzers in AppHost projects.

cc @mitchdenny @AArnott


davidfowl (approved):

AArnott:
What if your Aspire template includes an .editorconfig that turns the analyzers off?

AArnott:
We have an issue that tracks it: [2.0 package shouldn't bring in vs-threading analyzers · Issue #195 · microsoft/vs-streamjsonrpc](https://github.com/microsoft/vs-streamjsonrpc/issues/195)
But ultimately we're limited by nuget. They have 61 up-votes on this issue that is now 8 years old:
[project.assets.json should indicate "analyzer" assets like "compile" and "runtime" · Issue #6279 · NuGet/Home](https://github.com/NuGet/Home/issues/6279)

eerhardt:
> What if your Aspire template includes an .editorconfig that turns the analyzers off?
I see that your .editorconfig does in fact do this. Was that file part of the template that customers get? Was there some other reason it wasn't a sufficient solution for you?

I don't want to bother our customers with this at all. It is an internal implementation detail that we are using StreamJsonRpc.

Maybe one option (short term) is to remove the "compile" assets from our StreamJsonRpc dependency.

AArnott:
> remove the "compile" assets from our StreamJsonRpc dependency.

That wouldn't remove the `analyzers` asset from its own dependency.

Issue #5826, opened 2024-09-22
Support named configuration providers for Aspire resources


captainsafia:
Aspire resource support `Configuration`-based models for configuring resources under well-defined keys. For example, Storage Blob resources support configuration under the `Aspire:Azure:Storage:Blobs` key.

```
{
  "Aspire": {
    "Azure": {
      "Storage": {
        "Blobs": {
          "DisableHealthChecks": true,
          "DisableTracing": false,
          "ClientOptions": {
            "Diagnostics": {
              "ApplicationId": "myapp"
            }
          }
        }
      }
    }
  }
}
```

This configuration model doesn't currently support different configurations for different instances of the same resource type in an application. To address this, we'll update the resource configuration strategy to support instances keyed by name as follows:

```
{
  "Aspire": {
    "Azure": {
      "Storage": {
        "Blobs": {
           "blob1": { .. },
           "blob2": { ..},
        }
      }
    }
  }
}
```

At the moment, the primary motivation for doing this is to provide a resolution for the "dual config dilemma" documented in https://github.com/dotnet/aspire/issues/5563 since the existing configuration provider key name scheme doesn't conflict with Azure WebJob's lookup semantics.

**Tracked Integrations**

- [x] Aspire.Azure.AI.OpenAI
- [x] Aspire.Azure.Data.Tables
- [x] Aspire.Azure.Messaging.EventHubs
- [x] Aspire.Azure.Messaging.ServiceBus
- [x] Aspire.Azure.Message.WebPubSub
- [x] Aspire.Azure.Search.Documents
- [x] Aspire.Azure.Security.KeyVault
- [x] Aspire.Azure.Storage.Blobs
- [x] Aspire.Azure.Storage.Queues
- [x] Aspire.Confluent.Kafka
- [x] Aspire.Elastic.Clients.Elasticsearch
- [x] Aspire.Microsoft.Azure.Cosmos
- [x] Aspire.Microsoft.Data.SqlClient
- [x] Aspire.Microsoft.EntityFrameworkCore.Cosmos
- [x] Aspire.Microsoft.EntityFrameworkCore.SqlServer
- [x] Aspire.Milvus.Client
- [x] Aspire.MongoDB.Driver
- [x] Aspire.MySqlConnector
- [x] Aspire.NATS.Net
- [x] Aspire.Npgsql
- [x] Aspire.Npgsql.EntityFrameworkCore.PostgreSQL
- [ ] Aspire.OpenAI
[...]

PR #8014, opened 2025-03-11
Fix named configuration handling in DbContext-based clients

### LINKED ISSUE #7785: AddAzureCosmosDB is not working with AddCosmosDbContext

### Describe the bug

I have a very simple demo app here: https://github.com/mwiedemeyer/AspireDemo
Updated to Aspire 9.1

In the AppHost I add a CosmosDB resource with 

```
                    .RunAsPreviewEmulator(p => p.WithDataExplorer().WithDataVolume().WithLifetime(ContainerLifetime.Persistent))
                    .AddCosmosDatabase("db")
                    .AddContainer("users", "/id");
```

In my client a try to add a cosmos EF context with:

`builder.AddCosmosDbContext<AppDbContext>("cosmos", "db");`

However it fails, because it cannot find the connection string.

In the env variables of my client it shows only these connection strings mapping to the cosmos db:

- `Aspire__Microsoft__Azure__Cosmos__users__ConnectionString`
- `users`

The client fails with this error message:

`'A DbContext could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:cosmos' or either ConnectionString or AccountEndpoint must be provided in the 'Aspire:Microsoft:EntityFrameworkCore:Cosmos' or 'Aspire:Microsoft:EntityFrameworkCore:Cosmos:AppDbContext' configuration section.'`

So the names do not match. Is there anything I need to do to fix it?
I also tried this one, but it fails either:

`builder.AddCosmosDbContext<AppDbContext>("users", "db");`

### Expected Behavior

I expect the documented version to work like in the description above

### Steps To Reproduce

Run the demo project from here https://github.com/mwiedemeyer/AspireDemo
Then open the console logs for the func and click on the link (something like `http://localhost:7184/api/Function1`)

### Exceptions (if any)

InvalidOperationException

`'A DbContext could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:cosmos' or either ConnectionString or AccountEndpoint must be provided in the 'Aspire:Microsoft:EntityFrameworkCore:Cosmos' or 'Aspire:Microsoft:EntityFrameworkCore:Cosmos:AppDbContext' configuration section.'`

### .NET Version info

.NET SDK:
 Version:           9.0.200
 Commit:            90e8b202f2
 Workload version:  9.0.200-manifests.c4f6226a
 MSBuild version:   17.13.8+cbc39bea8

### Anything else?

Aspire 9.1 just updated after the release yesterday. Working with Visual Studio.

captainsafia:
Contributes towards https://github.com/dotnet/aspire/issues/5826.

Closes https://github.com/dotnet/aspire/issues/7785.

### DIFF
--- a/src/Components/Common/EntityFrameworkUtils.cs
+++ b/src/Components/Common/EntityFrameworkUtils.cs
@@ -14,12 +14,20 @@ internal static class EntityFrameworkUtils
     /// <summary>
     /// Binds the DbContext specific configuration section to settings when available.
     /// </summary>
-    public static TSettings GetDbContextSettings<TContext, TSettings>(this IHostApplicationBuilder builder, string defaultConfigSectionName, Action<TSettings, IConfiguration> bindSettings)
+    public static TSettings GetDbContextSettings<TContext, TSettings>(this IHostApplicationBuilder builder, string defaultConfigSectionName, string? connectionName, Action<TSettings, IConfiguration> bindSettings)
         where TSettings : new()
     {
         TSettings settings = new();
         var configurationSection = builder.Configuration.GetSection(defaultConfigSectionName);
         bindSettings(settings, configurationSection);
+        // If the connectionName is not provided, we've been called in the context
+        // of an Enrich invocation and don't need to bind the connectionName specific settings.
+        // Instead, we'll just bind to the TContext-specific settings.
+        if (connectionName is not null)
+        {
+            var connectionSpecificConfigurationSection = configurationSection.GetSection(connectionName);
+            bindSettings(settings, connectionSpecificConfigurationSection);
+        }
         var typeSpecificConfigurationSection = configurationSection.GetSection(typeof(TContext).Name);
         if (typeSpecificConfigurationSection.Exists()) // https://github.com/dotnet/runtime/issues/91380
         {
--- a/tests/Aspire.Microsoft.EntityFrameworkCore.Cosmos.Tests/AspireAzureEfCoreCosmosDBExtensionsTests.cs
+++ b/tests/Aspire.Microsoft.EntityFrameworkCore.Cosmos.Tests/AspireAzureEfCoreCosmosDBExtensionsTests.cs
@@ -246,6 +246,60 @@ public class AspireAzureEfCoreCosmosDBExtensionsTests
         Assert.Contains("AccountEndpoint", e.Message);
     }
 
+    [Fact]
+    public void AddCosmosDbContext_WithConnectionNameAndSettings_AppliesConnectionSpecificSettings()
+    {
+        var builder = Host.CreateEmptyApplicationBuilder(null);
+
+        var connectionName = "testdb";
+        var databaseName = "testdbname";
+
+        builder.Configuration.AddInMemoryCollection(new Dictionary<string, string?>
+        {
+            [$"ConnectionStrings:{connectionName}"] = ConnectionString,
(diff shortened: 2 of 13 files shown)


tests/Aspire.Microsoft.EntityFrameworkCore.Cosmos.Tests/AspireAzureEfCoreCosmosDBExtensionsTests.cs
```diff
@@ -246,6 +246,60 @@ public void AddAzureCosmosClient_FailsWithError()
...
+        });
+
+        EntityFrameworkCoreCosmosSettings? capturedSettings = null;
+        builder.AddCosmosDbContext<TestDbContext>(connectionName, databaseName, settings =>
+        {
+            capturedSettings = settings;
+        });
+
+        Assert.NotNull(capturedSettings);
+        Assert.Equal(TimeSpan.Parse("60"), capturedSettings.RequestTimeout);
```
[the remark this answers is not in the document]
captainsafia: Copilot, you make a good point but I am lazy and want to copy the exact string that was in the settings.

src/Components/Common/EntityFrameworkUtils.cs
```diff
@@ -14,12 +14,20 @@ internal static class EntityFrameworkUtils
...
+    public static TSettings GetDbContextSettings<TContext, TSettings>(this IHostApplicationBuilder builder, string defaultConfigSectionName, string? connectionName, Action<TSettings, IConfiguration> bindSettings)
         where TSettings : new()
     {
         TSettings settings = new();
         var configurationSection = builder.Configuration.GetSection(defaultConfigSectionName);
         bindSettings(settings, configurationSection);
+        // If the connectionName is not provided, we've been called in the context
+        // of an Enrich invocation and don't need to bind the connectionName specific settings.
+        // Instead, we'll just bind to the TContext-specific settings.
+        if (connectionName is not null)
```
eerhardt, line 26: I was expecting this to go after the TContext specific section (and thus override any settings).

My thinking is to go from "more general to more specific". But I guess we don't really know which is "more specific" - the DbContext type, or the connection name. Do people use the same DbContext on different connections more often than using the same connection between different DbContexts?

@AndriySvyryd @roji - any opinions here on the ordering? Which order should the settings apply? Where lower settings override the settings before them:

* Aspire:Microsoft:EntityFrameworkCore:Cosmos
* Aspire:Microsoft:EntityFrameworkCore:Cosmos:{connectionName}
* Aspire:Microsoft:EntityFrameworkCore:Cosmos:{DbContextType}

**or**

* Aspire:Microsoft:EntityFrameworkCore:Cosmos
* Aspire:Microsoft:EntityFrameworkCore:Cosmos:{DbContextType}
* Aspire:Microsoft:EntityFrameworkCore:Cosmos:{connectionName}

AndriySvyryd: With the way connection string is set configured in Aspire it would be pretty difficult to use the same context with different connection strings, and thus less likely

captainsafia: > It's possible to have multiple context types rely on the same connection, so I think Aspire:Microsoft:EntityFrameworkCore:Cosmos:{DbContextType} should be applied last

That was my intuition as well. It seems like can we can go ahead with the current implementation?

eerhardt: Thanks for the discussion here. This helped convince me.

eerhardt (approved):
LGTM. Thanks!

PR #8015, opened 2025-03-11
[CI] Simplify gathering test projects to run for non-templates on PRs

radical:
Two separate jobs (for Linux, and Windows) are currently used to get the list of non-template projects to run. But only the `Playground` is skipped on Windows. So, this can be simplified to use a single job.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -33,7 +33,8 @@ jobs:
   test:
     runs-on: ${{ inputs.os }}
     timeout-minutes: 60
-    name: ${{ inputs.os }}
+    name: ${{ inputs.testShortName }}
+    if: inputs.os != 'windows-latest' || inputs.testShortName != 'Playground'
     steps:
       - name: Validate arguments
         shell: pwsh
--- a/.github/actions/enumerate-tests/action.yml
+++ b/.github/actions/enumerate-tests/action.yml
@@ -35,6 +35,15 @@ runs:
         $lines = Get-Content $filePath
         $jsonObject = @{
             "shortname" = $lines | Sort-Object
+            "os" = @(
+              @{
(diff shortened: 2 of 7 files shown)


radical:
First attempt.

radical:
As a result of this the way the jobs are displayed on github changes from:
Screenshot 2025-03-12 at 12 12 15
... to ...
Screenshot 2025-03-12 at 12 10 43

danmoseley (approved):
if it works

.github/workflows/run-tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
   test:
     runs-on: ${{ inputs.os }}
     timeout-minutes: 60
-    name: ${{ inputs.os }}
+    name: ${{ inputs.testShortName }}
+    if: inputs.os != 'windows-latest' || inputs.testShortName != 'Playground'
```
eerhardt, line 37: Where were the "playground" tests filtered out previously? Can that be removed now that we are filtering them out here?

Another option could be to update the playground tests to have `[RequiresDocker]` on them, and just always run them, even if they no-op for now.

eerhardt (approved):
LGTM. Thanks!

.github/workflows/run-tests.yml
radical: They are controlled from the project file:
https://github.com/dotnet/aspire/blob/0574cfcd61a28a538acca941bec71880e3b7f95f/tests/Aspire.Playground.Tests/Aspire.Playground.Tests.csproj#L9-L13

I'll move to using the attribute instead of a condition here.

radical:
`main` pipeline validation run - https://dev.azure.com/dnceng-public/public/_build/results?buildId=977721&view=results

PR #8018, opened 2025-03-11
Reapply "Backchannel between CLI and AppHost (#7991)" (#7999)

mitchdenny:
This PR reverts the revert of the backchannel PR. Once we get the updated StreamJsonRpc package we'll update this branch and it should be able to go in safely.


.editorconfig
```diff
@@ -478,4 +478,75 @@ dotnet_diagnostic.IDE0005.severity = silent
...
+dotnet_diagnostic.VSTHRD112.severity = none
+
+# VSTHRD113: Check for System.IAsyncDisposable.
+dotnet_diagnostic.VSTHRD113.severity = none
+
+# VSTHRD114: Avoid returning a null Task.
+dotnet_diagnostic.VSTHRD114.severity = none
+
+# VSTHRD200: Use "Async" suffix in names of methods that return an awaitable type.
+dotnet_diagnostic.VSTHRD200.severity = none
```
eerhardt, line 552: We shouldn't need these changes once we get the updated package. It shouldn't have the analyzer enabled.

mitchdenny: Yep ... leaving until we get the new package though :)

mitchdenny:
@eerhardt @davidfowl using the new prerelease StreamJsonRpc. We should have stable next week so I think this is safe to merge in now and continue building on.

davidfowl (approved):

Issue #7101, opened 2025-01-14
Create an empty database for AddDatabase calls


davidfowl:
This is a big point of confusion for developers using aspire

Kralizek:
I assume this is not going to be solved in 9.1. :(

danmoseley:
Postgres and SQL Server first as they're highest usage rn.

davidfowl:
@sebastienros can you make sure we have issues for the other databases and guidance for authors of integrations adding new databases.

Do we have a way to disable the creation of the database as well?

sebastienros:
> Do we have a way to disable the creation of the database as well?

I believe that one can already write a custom no-op script if they don't want it to be created. But we could add an option/extension for this too.

SoggyBottomBoy:
Maybe not the correct place to ask this question. I just moved from using the SqlServer container integration to AzureSqlServer (which worked great for deploying to Azure, and the database is created). For local debugging I also switched to using RunAsContainer() using AzureSqlServer, however, no database is created and therefore I need to run an entrypoint script to create it.

It would be great if AzureSqlServer().RunAsContainer() also had this behaviour (the database is created if it does not exist).

RussKie:
Closing as all references issues are now completed.

PR #8022, opened 2025-03-12
Create database for Sql Server resource

sebastienros:
## Description

Support database creation for `AddDatabase` in Sql Server container resource. If the database already exists nothing is done. Custom scripts can be provided, using an annotation internally.

Part of https://github.com/dotnet/aspire/issues/7101

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

### DIFF
--- a/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
+++ b/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
@@ -68,8 +68,11 @@ public static class SqlServerBuilderExtensions
                         throw new InvalidOperationException($"Could not open connection to '{sqlServer.Name}'");
                     }
 
+                    var scriptAnnotation = sqlDatabase.Annotations.OfType<ScriptAnnotation>().LastOrDefault();
+
                     using var command = sqlConnection.CreateCommand();
-                    command.CommandText = sqlDatabase.DatabaseCreationScript ?? $"IF ( NOT EXISTS ( SELECT 1 FROM sys.databases WHERE name = @DatabaseName ) ) CREATE DATABASE {quotedDatabaseIdentifier};";
+                    command.CommandText = scriptAnnotation?.Script ??
+                        $"IF ( NOT EXISTS ( SELECT 1 FROM sys.databases WHERE name = @DatabaseName ) ) CREATE DATABASE {quotedDatabaseIdentifier};";
                     command.Parameters.Add(new SqlParameter("@DatabaseName", sqlDatabase.DatabaseName));
                     await command.ExecuteNonQueryAsync(ct).ConfigureAwait(false);
                 }
@@ -101,9 +104,8 @@ public static class SqlServerBuilderExtensions
     /// <param name="builder">The SQL Server resource builders.</param>
     /// <param name="name">The name of the resource. This name will be used as the connection string name when referenced in a dependency.</param>
     /// <param name="databaseName">The name of the database. If not provided, this defaults to the same value as <paramref name="name"/>.</param>
-    /// <param name="databaseCreationScript">A custom database creation script.</param>
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
-    public static IResourceBuilder<SqlServerDatabaseResource> AddDatabase(this IResourceBuilder<SqlServerServerResource> builder, [ResourceName] string name, string? databaseName = null, string? databaseCreationScript = null)
+    public static IResourceBuilder<SqlServerDatabaseResource> AddDatabase(this IResourceBuilder<SqlServerServerResource> builder, [ResourceName] string name, string? databaseName = null)
     {
         ArgumentNullException.ThrowIfNull(builder);
         ArgumentException.ThrowIfNullOrEmpty(name);
@@ -113,7 +115,7 @@ public static class SqlServerBuilderExtensions
         // Use the resource name as the database name if it's not provided
         databaseName ??= name;
 
-        var sqlServerDatabase = new SqlServerDatabaseResource(name, databaseName, builder.Resource) { DatabaseCreationScript = databaseCreationScript };
+        var sqlServerDatabase = new SqlServerDatabaseResource(name, databaseName, builder.Resource);
 
         builder.Resource.AddDatabase(sqlServerDatabase);
 
@@ -167,24 +169,25 @@ public static class SqlServerBuilderExtensions
         ArgumentNullException.ThrowIfNull(builder);
         ArgumentException.ThrowIfNullOrEmpty(source);
 
-        if (!OperatingSystem.IsWindows())
-        {
-            return builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
-        }
-        else
-        {
-            // c.f. https://learn.microsoft.com/sql/linux/sql-server-linux-docker-container-configure?view=sql-server-ver15&pivots=cs1-bash#mount-a-host-directory-as-data-volume
+        return builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
+    }
 
-            foreach (var dir in new string[] { "data", "log", "secrets" })
-            {
-                var path = Path.Combine(source, dir);
+    /// <summary>
+    /// Alters the JSON configuration document used by the emulator.
+    /// </summary>
+    /// <param name="builder">The builder for the <see cref="SqlServerDatabaseResource"/>.</param>
+    /// <param name="script">The SQL script used to create the database.</param>
+    /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
+    /// <remarks>
+    /// <value>Default script is <code>IF ( NOT EXISTS ( SELECT 1 FROM sys.databases WHERE name = @DatabaseName ) ) CREATE DATABASE [&lt;QUOTED_DATABASE_NAME%gt;];</code></value>
+    /// </remarks>
+    public static IResourceBuilder<SqlServerDatabaseResource> WithCreationScript(this IResourceBuilder<SqlServerDatabaseResource> builder, string script)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(script);
 
-                Directory.CreateDirectory(path);
+        builder.WithAnnotation(new ScriptAnnotation(script));
 
-                builder.WithBindMount(path, $"/var/opt/mssql/{dir}", isReadOnly);
-            }
-
-            return builder;
-        }
+        return builder;
     }
 }
--- a/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
+++ b/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
@@ -19,7 +19,7 @@ public static partial class SqlServerBuilderExtensions
 {
     // GO delimiter format: {spaces?}GO{spaces?}{repeat?}{comment?}
     // https://learn.microsoft.com/sql/t-sql/language-elements/sql-server-utilities-statements-go
-    [GeneratedRegex(@"^\s*GO(?<repeat>\s+\d{1,6})?(\s*\-{2,}.*)?\s*$", RegexOptions.CultureInvariant | RegexOptions.IgnoreCase)]
+    [GeneratedRegex(@"^\s*GO\s+(?<repeat>\d{1,6})?(\s*\-{2,}.*)?\s*$", RegexOptions.CultureInvariant | RegexOptions.IgnoreCase)]
     private static partial Regex GoStatements();
 
     /// <summary>
--- a/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
+++ b/src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
@@ -167,6 +167,24 @@ public static class SqlServerBuilderExtensions
         ArgumentNullException.ThrowIfNull(builder);
         ArgumentException.ThrowIfNullOrEmpty(source);
 
-        return builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
+        if (!OperatingSystem.IsWindows())
+        {
+            return builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
+        }
+        else
+        {
+            // c.f. https://learn.microsoft.com/sql/linux/sql-server-linux-docker-container-configure?view=sql-server-ver15&pivots=cs1-bash#mount-a-host-directory-as-data-volume
+
+            foreach (var dir in new string[] { "data", "log", "secrets" })
+            {
+                var path = Path.Combine(source, dir);
+
+                Directory.CreateDirectory(path);
+
+                builder.WithBindMount(path, $"/var/opt/mssql/{dir}", isReadOnly);
+            }
+
+            return builder;
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -132,7 +136,7 @@ public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
             using var builder1 = TestDistributedApplicationBuilder.Create(o => { }, testOutputHelper);
 
             var sqlserver1 = builder1.AddSqlServer("sqlserver");
-            var masterdb1 = sqlserver1.AddDatabase("master");
+            var db1 = sqlserver1.AddDatabase(databaseName);
```
sebastienros, line 139: Not using `master` updates this test to ensure the database is created. `master` was used since it's a default one in sql server.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -68,18 +70,18 @@ public async Task VerifySqlServerResource()
         using var builder = TestDistributedApplicationBuilder.Create(o => { }, testOutputHelper);
 
         var sqlserver = builder.AddSqlServer("sqlserver");
-        var tempDb = sqlserver.AddDatabase("tempdb");
+        var newDb = sqlserver.AddDatabase(databaseName);
```
sebastienros, line 73: Not using `tempdb` updates this test to ensure the database is created. `tempdb` was used since it's a default one in sql server.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -319,4 +323,109 @@ await pipeline.ExecuteAsync(async token =>
...
+            var results = await selectCommand.ExecuteReaderAsync(token);
+            Assert.True(results.HasRows);
+        }, cts.Token);
+    }
+
+    [Fact]
+    [RequiresDocker]
+    public async Task AddDatabaseCreatesDatabaseWithSpecialNames()
+    {
+        const string databaseName = "!'][\"";
```
sebastienros, line 388: @eerhardt this one might itch a bit. This was to ensure the name is correctly encoded in the default script. But I needed to change the connection string creation to encode it too.

tests/Aspire.Hosting.SqlServer.Tests/SqlServerFunctionalTests.cs
```diff
@@ -149,7 +153,7 @@ public async Task WithDataShouldPersistStateBetweenUsages(bool useVolume)
             {
                 bindMountPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
 
-                Directory.CreateDirectory(bindMountPath);
+                var directoryInfo = Directory.CreateDirectory(bindMountPath);
```
danmoseley, line 156: unused?

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -72,12 +110,33 @@ public static IResourceBuilder<SqlServerDatabaseResource> AddDatabase(this IReso
         ArgumentNullException.ThrowIfNull(builder);
         ArgumentException.ThrowIfNullOrEmpty(name);
 
+        Dictionary<string, string> databaseConnectionStrings = [];
```
davidfowl, line 113: Why do you need this? This call has all of the details for a single database right? You can just capture the  databaseConnectionString

src/Aspire.Hosting.SqlServer/SqlServerDatabaseResource.cs
```diff
@@ -23,8 +24,16 @@ public class SqlServerDatabaseResource(string name, string databaseName, SqlServ
...
+    public ReferenceExpression ConnectionStringExpression {
+        get
+        {
+            var connectionStringBuilder = new SqlConnectionStringBuilder();
+            connectionStringBuilder["Database"] = DatabaseName;
+            var connectionString = connectionStringBuilder.ToString();
+
+            return ReferenceExpression.Create($"{Parent};{connectionString}");
+        }
+    }
```
davidfowl, line 36: Formatting. Why the change?

davidfowl:
This generally looks good; I wonder if we should also think about publish mode (maybe not in this PR).

eerhardt:
> I wonder if we should also think about publish mode (maybe not in this PR).

What does this mean exactly? For Azure SQL, we will create the DB with bicep. Are you thinking of some sort of "startup" command that runs on publish of a SQL container?

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -45,6 +47,42 @@ public static IResourceBuilder<SqlServerServerResource> AddSqlServer(this IDistr
...
+                    if (sqlConnection.State != System.Data.ConnectionState.Open)
+                    {
+                        throw new InvalidOperationException($"Could not open connection to '{sqlServer.Name}'");
+                    }
+
+                    var scriptAnnotation = sqlDatabase.Annotations.OfType<ScriptAnnotation>().LastOrDefault();
+
+                    using var command = sqlConnection.CreateCommand();
+                    command.CommandText = scriptAnnotation?.Script ??
+                        $"IF ( NOT EXISTS ( SELECT 1 FROM sys.databases WHERE name = @DatabaseName ) ) CREATE DATABASE {quotedDatabaseIdentifier};";
```
eerhardt, line 75: Why the difference between the name in the WHERE and the name used on the Create Database? Shouldn't they both use the same SqlParameter?

src/Aspire.Hosting/ApplicationModel/ScriptAnnotation.cs
```diff
@@ -0,0 +1,25 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Represents an annotation for defining a script to create a resource.
+/// </summary>
+public sealed class ScriptAnnotation : IResourceAnnotation
```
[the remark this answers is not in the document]
sebastienros: I was torn between a database specific name or not. But in the end it's only a string. Could have been all the way to `StringAnnotation`, so I understand why a more specific one would make sense. If we tell users they can add one or change it. Another advantage of a specific name is that it won't conflict with another annotation that would use the same type for something else, and we would be screwed. So I agree with that: `DatabaseCreationScriptAnnotation` for instance.

NB: To go all the way meta we could have general annotation like `KeyValuePairAnnotation` which would also have a key/value (`"databasecreationscript"`/`"create database [foo]"`). But that's why we have custom type for ;)

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
sebastienros: Parameters can't be used for object names (such as database names, table names, or column names). 
It's used in the query part because in that case it's a string value.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -112,4 +169,23 @@ public static IResourceBuilder<SqlServerServerResource> WithDataBindMount(this I
 
         return builder.WithBindMount(source, "/var/opt/mssql", isReadOnly);
     }
+
+    /// <summary>
+    /// Alters the JSON configuration document used by the emulator.
```
eerhardt, line 174: This doesn't seem right.

eerhardt (commented):
Can we also update the playground app to no longer ensure the DB is created?

https://github.com/dotnet/aspire/blob/8d2aac405912676ae19c41807014e0eb2e7295c9/playground/SqlServerEndToEnd/SqlServerEndToEnd.ApiService/Program.cs#L19-L23

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -45,6 +47,42 @@ public static IResourceBuilder<SqlServerServerResource> AddSqlServer(this IDistr
...
+
+                    using var command = sqlConnection.CreateCommand();
+                    command.CommandText = scriptAnnotation?.Script ??
+                        $"IF ( NOT EXISTS ( SELECT 1 FROM sys.databases WHERE name = @DatabaseName ) ) CREATE DATABASE {quotedDatabaseIdentifier};";
+                    command.Parameters.Add(new SqlParameter("@DatabaseName", sqlDatabase.DatabaseName));
+                    await command.ExecuteNonQueryAsync(ct).ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    @event.Services.GetRequiredService<ILogger<SqlServerServerResource>>().LogError(e, "Failed to create database '{DatabaseName}'", sqlDatabase.DatabaseName);
```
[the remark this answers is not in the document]
sebastienros: It currently shows up in the console. Would it be nice if it were in the database resource log in the dashboard? If so what is the way to do it?

src/Aspire.Hosting/ApplicationModel/ScriptAnnotation.cs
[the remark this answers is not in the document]
eerhardt: Given the summary of the class "for defining a script to create a resource." I thought the class would be named more specific.

eerhardt, line 9: Thinking more about this, we also have a sample for how to customize the initial database:

https://github.com/dotnet/aspire-samples/blob/fc7aa13409821eb1d7b7ccd374b465c97dca5591/samples/DatabaseContainers/DatabaseContainers.AppHost/Program.cs#L38-L45

1. Does this feature work well with this example?
2. I wonder if we really need to support a custom create database script, and instead point people to this route. (possibly with an option for turning this feature off).

cc @DamianEdwards

sebastienros: As a user I would definitely prefer this extension method (and annotation) to customize the create database script than having to deal with bind mounts and understand how the container is configured. The same way we have `WithDataVolume` extensions. Or should we provide (in addition maybe) a method to configure this `"/usr/config/entrypoint.sh"` script in sql server container to make it easier?

Though I would understand if we say we skip this feature (customizing CREATE DATABASE) since there is already a way to do it in the container itself.

eerhardt (approved):

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -1,17 +1,27 @@
...
 
 /// <summary>
 /// Provides extension methods for adding SQL Server resources to the application model.
 /// </summary>
-public static class SqlServerBuilderExtensions
+public static partial class SqlServerBuilderExtensions
 {
+    // GO delimiter format: {spaces?}GO{spaces?}{repeat?}{comment?}
+    // https://learn.microsoft.com/sql/t-sql/language-elements/sql-server-utilities-statements-go
+    [GeneratedRegex(@"^(?:\s*)(GO|go)(?<repeat>\s+\d{1,6})?(?:\s*\-{2,}.*)?(?:\s+)?$", RegexOptions.CultureInvariant)]
```
danmoseley, line 22:
```suggestion
    [GeneratedRegex(@"^\s*(GO|go)(?<repeat>\s+\d{1,6})?\s*(?:-{2,}.*)?\s*$", RegexOptions.CultureInvariant)]
```
i don't think you need the non capturing groups?

danmoseley: uh, I didn't test it though

danmoseley:
```suggestion
    [GeneratedRegex(@"^\s*GO(?<repeat>\s+\d{1,6})?\s*(?:-{2,}.*)?\s*$", RegexOptions.CultureInvariant | RegexOptions.IgnoreCase)]
```
?

sebastienros: Thanks, suggestions applied, and added unit tests.

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -1,17 +1,27 @@
...
 
 /// <summary>
 /// Provides extension methods for adding SQL Server resources to the application model.
 /// </summary>
-public static class SqlServerBuilderExtensions
+public static partial class SqlServerBuilderExtensions
 {
+    // GO delimiter format: {spaces?}GO{spaces?}{repeat?}{comment?}
+    // https://learn.microsoft.com/sql/t-sql/language-elements/sql-server-utilities-statements-go
+    [GeneratedRegex(@"^\s*GO(?<repeat>\s+\d{1,6})?(\s*\-{2,}.*)?\s*$", RegexOptions.CultureInvariant | RegexOptions.IgnoreCase)]
```
danmoseley, line 22:
```suggestion
    [GeneratedRegex(@"^\s*GO\s+(?<repeat>\d{1,6})?(\s*\-{2,}.*)?\s*$", RegexOptions.CultureInvariant | RegexOptions.IgnoreCase)]
```
zero functional effect, but slightly tidier, feel free to ignore

sebastienros: Had to revert it since the number require \s+ but comments are \s*. That would fail the test `"GO--"`

src/Aspire.Hosting.SqlServer/SqlServerBuilderExtensions.cs
```diff
@@ -112,4 +162,87 @@ public static IResourceBuilder<SqlServerServerResource> WithDataBindMount(this I
...
+                {
+                    using var command = sqlConnection.CreateCommand();
+                    command.CommandText = batchBuilder.ToString();
+                    await command.ExecuteNonQueryAsync(ct).ConfigureAwait(false);
+                }
+            }
+        }
+        catch (Exception e)
+        {
+            var logger = serviceProvider.GetRequiredService<ILogger<DistributedApplicationBuilder>>();
```
DamianEdwards, line 244: Are we going to update this to write the log to the resource instead?

sebastienros: I just got taught how to, yes I will change that right away.

ErikEJ:
@sebastienros This is amazing, love the script support. Together with a SQL project, This allows full lifecycle support for a persisted local database.

ErikEJ:
@sebastienros @davidfowl When will this be in the latest "daily" ??  Not in `9.2.0-preview.1.25178.1`

davidfowl:
It should be there.

ErikEJ:
@davidfowl Found it - I will show myself out!

PR #8025, opened 2025-03-12
`aspire add` command

mitchdenny:
Exploration of a `aspire add` command. Usage example:

```bash
aspire add redis
```

This PR also adds support for two switches (`--debug`, without which all log output is supressed so output is purely driven by Spectre.Console, and `--wait-for-debugger` which causes the code to pause until a debugger attaches).

https://github.com/user-attachments/assets/0d31c564-d7b7-4138-8053-06e5fd837899


src/Aspire.Cli/Program.cs
```diff
@@ -270,6 +272,66 @@ private static void ConfigureNewCommand(Command parentCommand)
         parentCommand.Subcommands.Add(command);
     }
 
+    private static Integration GetIntegrationByInteractiveFlow(IEnumerable<Integration> knownIntegrations)
+    {
+        var prompt = new SelectionPrompt<Integration>()
+            .Title("Please select the integration you want to add:")
```
danmoseley, line 278: how does this work when you need to add the hosting package as well?

danmoseley (approved):
LGTM assuming we don't care about error handling at this point.

src/Aspire.Cli/Program.cs
mitchdenny: We will actually only be adding the hosting packages.

src/Aspire.Cli/DotNetCliRunner.cs
```diff
@@ -15,21 +16,37 @@ internal sealed class DotNetCliRunner(ILogger<DotNetCliRunner> logger, CliRpcTar
 
     public async Task<int> RunAsync(FileInfo projectFile, string[] args, CancellationToken cancellationToken)
     {
+
```
davidfowl, line 19:
```suggestion
```

davidfowl (approved):

maddymontaquila:
I love you

PR #8030, opened 2025-03-12
Make secrets in docker compose placeholders

davidfowl:
- Longer term we may have to replace this approach to hoist secrets and any env variable that uses secrets out of the file altogether.

### DIFF
--- a/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
+++ b/tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
@@ -209,6 +209,7 @@ public class DistributedApplicationTests
     }
 
     [Fact]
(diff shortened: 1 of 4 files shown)


mitchdenny (approved):

tests/Aspire.Hosting.Tests/DistributedApplicationTests.cs
```diff
@@ -209,6 +209,7 @@ public async Task ExplicitStart_StartExecutable()
     }
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/aspire/issues/7836")]
```
davidfowl, line 212:
```
```

davidfowl, line 212:
```suggestion
    
```

PR #8040, opened 2025-03-12
Refactor SignalR sample

cqnguy23:
- Replace Github star fetch logic with a simple incremental counter to focus on the broadcasting logic of sample
- Move `BackgroundWorker` to Program.cs and change name to `PeriodicBroadcaster` for clarity
- Move `hubContext` to global scope for reusability


playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -1,21 +1,17 @@
...
-// Add services to the container.
-builder.Services.AddRazorPages();
-builder.Services.AddHttpClient();
-builder.Services.AddSingleton(sp =>
-{
-    return new ServiceManagerBuilder()
+var serviceManager = new ServiceManagerBuilder()
         .WithOptions(option =>
         {
             option.ConnectionString = builder.Configuration["ConnectionStrings:signalrServerless"];
```
vicancy, line 8: builder.Configuration.GetConnectionString("signalrServerless")

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -1,21 +1,17 @@
...
         })
         .BuildServiceManager();
-});
-builder.Services.AddHostedService<BackgroundWorker>();
+var hubContext = await serviceManager.CreateHubContextAsync("myHubName", default);
+
+builder.Services.AddRazorPages();
+builder.Services.AddSingleton(hubContext).AddHostedService<PeriodicBroadcaster>();
```
vicancy, line 14: how it fetches ServiceHubContext in the ctor?

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -47,3 +42,16 @@
 
 app.MapRazorPages();
 app.Run();
+
+internal sealed class PeriodicBroadcaster(ServiceHubContext hubContext) : BackgroundService
+{
+    private int _count;
```
vicancy, line 48: could be a local var inside the method

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -47,3 +42,16 @@
...
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            await hubContext.Clients.All.SendCoreAsync("newMessage", [$"Current count is: {_count++}"], stoppingToken);
```
vicancy, line 53: use SendAsync instead

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -47,3 +42,16 @@
...
+            await Task.Delay(TimeSpan.FromSeconds(2), stoppingToken);
```
vicancy, line 54: `Task.Delay(2000, stoppingToken)`

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -1,21 +1,17 @@
...
-    return new ServiceManagerBuilder()
+var serviceManager = new ServiceManagerBuilder()
         .WithOptions(option =>
         {
             option.ConnectionString = builder.Configuration["ConnectionStrings:signalrServerless"];
         })
         .BuildServiceManager();
-});
-builder.Services.AddHostedService<BackgroundWorker>();
+var hubContext = await serviceManager.CreateHubContextAsync("myHubName", default);
```
vicancy, line 11: move to right above negotiate logic

playground/signalr/SignalRServerlessWeb/Program.cs
cqnguy23: So `.AddHostedService<PeriodicBroadcaster>()` registers `PeriodicBroadcaster` in the DI Container. And since the ctor of `PeriodicBroadcaster` has a dependency which is `ServiceHubContext`,  the DI Container fetches the `ServiceHubContext` singleton and pass it to the ctor when init the `PeriodicBroadcaster` instance

playground/signalr/SignalRServerlessWeb/Program.cs
cqnguy23: This needs to be declared before app build to register singleton for `PeriodicBroadcaster`, so I can't really move it above negotiate logic. I can move it right above `builder.Services.AddSingleton(hubContext).AddHostedService<PeriodicBroadcaster>()`

playground/signalr/SignalRServerlessWeb/Program.cs
vicancy: why not directly add `new PeriodicBroadcaster(hubContext)` as the HostedService?

I am not sure if adding the ServiceHubContext is a best approach. this approach fails quickly when we have multiple hubs.

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -47,3 +43,16 @@
 
 app.MapRazorPages();
 app.Run();
+
+internal sealed class PeriodicBroadcaster(ServiceHubContext hubContext) : BackgroundService
+{
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        var _count = 0;
```
vicancy, line 51: local var does not start with `_`

playground/signalr/SignalRServerlessWeb/Program.cs
cqnguy23: that makes sense, just updated the code

vicancy (approved):

Issue #8038, opened 2025-03-12
WithHttpCommand_ResultsInExpectedResultForStatusCode failing in CI


eerhardt:
See https://github.com/dotnet/aspire/actions/runs/13813895287

[Aspire.Hosting.Tests.WithHttpCommandTests.WithHttpCommand_ResultsInExpectedResultForStatusCode(statusCode: 500, expectSuccess: False)](https://github.com/dotnet/aspire/blob/29a5ad19e10b43b58f4d4092e5b8e2ef6ead23c3/tests/Aspire.Hosting.Tests/WithHttpCommandTests.cs#L111)

```
System.TimeoutException : The operation at /_/tests/Aspire.Hosting.Tests/WithHttpCommandTests.cs:120 timed out after reaching the limit of 5000ms.
   at Microsoft.AspNetCore.InternalTesting.AsyncTestHelpers.TimeoutAfter(Task task, TimeSpan timeout, String filePath, Int32 lineNumber) in /_/tests/Shared/AsyncTestHelpers.cs:line 149
   at Aspire.Hosting.Tests.WithHttpCommandTests.WithHttpCommand_ResultsInExpectedResultForStatusCode(Int32 statusCode, Boolean expectSuccess) in /_/tests/Aspire.Hosting.Tests/WithHttpCommandTests.cs:line 120
--- End of stack trace from previous location ---
```

eerhardt:
Fixed by #8041

PR #8041, opened 2025-03-12
Update timeouts for WithHttpCommand tests

DamianEdwards:
See #8038


eerhardt:
Rerunning one more time. If this passes, then we should merge.

eerhardt (approved):

PR #8042, opened 2025-03-12
Parametrize parameters as env replacements

davidfowl:
Express parameters in the app model as env variable expressions in the compose file

### DIFF
--- a/src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
+++ b/src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
@@ -360,7 +360,7 @@ internal sealed class DockerComposePublishingContext(
 
                 if (value is ParameterResource param)
                 {
-                    return await AllocateParameter(param).ConfigureAwait(false) ?? throw new InvalidOperationException("Parameter name is null");
+                    return AllocateParameter(param);
                 }
 
                 if (value is ConnectionStringReference cs)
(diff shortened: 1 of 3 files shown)


src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
```diff
@@ -420,32 +420,24 @@ private async Task<object> ProcessValueAsync(object value)
...
+            // Treat secrets as environment variable placeholders as for now
+            // this doesn't handle generation of parameter values with defaults
+            var env = parameter.Name.ToUpperInvariant().Replace("-", "_");
 
-            if (res.Secret) 
+            if (parameter.Secret || parameter.Default is null)
             {
-                // Treat secrets as environment variable placeholders as for now
-                // this doesn't handle generation of parameter values with defaults
-                var env = res.Name.ToUpperInvariant().Replace("-", "_");
```
eerhardt, line 441: Do we still need this replacement?

eerhardt (approved):

src/Aspire.Hosting.Docker/DockerComposePublishingContext.cs
davidfowl: looks more env variabley

danhpaiva:
.Net and Docker are amazing!

PR #8047, opened 2025-03-12
Port arcade publishing fix

wtgodbe:
Adds retries back, and sets defaults where necessary


sebastienros (approved):

joperezr:
Correct me if I'm wrong, but these would just get overwritten by the next arcade update, right? Should we make this configurable from dotnet/arcade itself?

wtgodbe:
> Correct me if I'm wrong, but these would just get overwritten by the next arcade update, right? Should we make this configurable from dotnet/arcade itself?

We already pushed the update to Arcade, but wanted to get it into this repo as quickly as possible since Publishing had been failing. We also could have just waited ~1 day for the arcade build to get validated & promoted, then manually triggered the update

PR #8049, opened 2025-03-13
Added a test for docker compose file generation

davidfowl:
Added baseline tests for docker compose generation


mitchdenny (approved):
LGTM. As I'm looking at this I'm wondering whether we should introduce a virtual file system for the field we produce? Initially to make testing easier.

davidfowl:
Yes probably. I almost did that but we lots of tests today that write temp files to disk

PR #8050, opened 2025-03-13
Introduce the azure publisher

davidfowl:
Added the azure publisher 

- Added the azure publisher whose job it is to gather bicep resources into both a `main.bicep` and to gather all deployment target resources into sub folders that will be deployed separately.
- Added tests


src/Aspire.Hosting.Azure/AzurePublisher.cs
```diff
@@ -0,0 +1,129 @@
...
+            var moduleDirectory = outputDirectory.CreateSubdirectory(resource.Name);
+
+            var modulePath = Path.Combine(moduleDirectory.FullName, $"{resource.Name}.bicep");
+
+            File.Copy(file.Path, modulePath, true);
+
+            var identifier = Infrastructure.NormalizeBicepIdentifier(resource.Name);
+            BicepValue<string> scope = new IdentifierExpression(rg.BicepIdentifier);
+
+            if (resource.Scope?.ResourceGroup is string rgName)
```
eerhardt, line 81: This could be a Parameter, not just a `string`

src/Aspire.Hosting.Azure/AzurePublisher.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        SaveToDisk(outputDirectory.FullName, infra);
+
+        return Task.CompletedTask;
+    }
+
+    private static void SaveToDisk(string outputDirectoryPath, Infrastructure infrastructure)
+    {
+        var plan = infrastructure.Build();
```
eerhardt, line 121: Should we be passing the ProvisioningOptions through to here?

src/Aspire.Hosting/PublisherDistributedApplicationBuilderExtensions.cs
```diff
@@ -20,13 +20,16 @@ public static class PublisherDistributedApplicationBuilderExtensions
...
+        where TPublisher : class, IDistributedApplicationPublisher
+        where TPublisherOptions : class
     {
         // TODO: We need to add validation here since this needs to be something we refer to on the CLI.
         builder.Services.AddKeyedSingleton<IDistributedApplicationPublisher, TPublisher>(name);
 
-        builder.Services.Configure<TPublisherOptions>("name", configureOptions!);
+        if (configureOptions is not null)
+        {
+            builder.Services.Configure<TPublisherOptions>("name", configureOptions!);
```
eerhardt, line 31:
```suggestion
            builder.Services.Configure<TPublisherOptions>("name", configureOptions);
```

src/Aspire.Hosting.Azure/AzurePublisher.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        var tags = new ProvisioningVariable("tags", typeof(object))
+        {
+            Value = new BicepDictionary<string>
+            {
+                ["aspire-env-name"] = environmentParam
+            }
+        };
+
+        // REVIEW: Do we want people to be able to change this
```
mitchdenny, line 50: We could potentially provide an override. I think some people dislike the whole rg- thing (I do, it reminds me of warts on variable names).

But I think we should keep it as default for back compat.

davidfowl: Via options for sure.

mitchdenny (approved):

PR #8067, opened 2025-03-13
Introduce --operation switch and inspect mode

mitchdenny:
Replaces this PR: #8033

Introducing a `--operation` switch which takes the options `run`, `publish`, and `inspect`. Introduced tests to insure backwards compatability. In inspect mode the apphost sits idle so that the launcher can make RPC requests to interrogate the application model after it has been (partially) evaluated.

### DIFF
--- a/src/Aspire.Hosting/DistributedApplicationBuilder.cs
+++ b/src/Aspire.Hosting/DistributedApplicationBuilder.cs
@@ -122,19 +122,12 @@ public class DistributedApplicationBuilder : IDistributedApplicationBuilder
             _ => throw new DistributedApplicationException("Invalid operation specified. Valid operations are 'publish', 'inspect', or 'run'.")
         };
 
-        if (operation == DistributedApplicationOperation.Run)
+        return operation switch
         {
-            return new DistributedApplicationExecutionContextOptions(operation!);
-        }
-        else if (operation == DistributedApplicationOperation.Inspect)
-        {
-            return new DistributedApplicationExecutionContextOptions(operation);
-        }
-        else
-        {
-            var publisher = _innerBuilder.Configuration["Publishing:Publisher"];
-            return new DistributedApplicationExecutionContextOptions(operation, publisher);
-        }
+            DistributedApplicationOperation.Run => new DistributedApplicationExecutionContextOptions(operation),
+            DistributedApplicationOperation.Inspect => new DistributedApplicationExecutionContextOptions(operation),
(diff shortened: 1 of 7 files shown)


src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -102,6 +102,41 @@ public DistributedApplicationBuilder(string[] args) : this(new DistributedApplic
...
+
+        var operation = _innerBuilder.Configuration["AppHost:Operation"]?.ToLowerInvariant() switch
+        {
+            "publish" => DistributedApplicationOperation.Publish,
+            "inspect" => DistributedApplicationOperation.Inspect,
+            "run" => DistributedApplicationOperation.Run,
+            _ => throw new DistributedApplicationException("Invalid operation specified. Valid operations are 'publish', 'inspect', or 'run'.")
+        };
+
+        if (operation == DistributedApplicationOperation.Run)
```
karolz-ms, line 125: Missed opportunity to employ pattern matching again (for computing return value)? 😆

tests/Aspire.Hosting.Tests/WithHttpCommandTests.cs
```diff
@@ -325,6 +325,8 @@ public async Task WithHttpCommand_EnablesCommandOnceResourceIsRunning()
         // Arrange
         using var builder = TestDistributedApplicationBuilder.Create(testOutputHelper);
 
+        builder.Configuration["CODESPACES"] = "false";
```
karolz-ms, line 328: Why is this necessary?

mitchdenny: When running this test in a Codespace the URLs will be remapped in a way that this test doesn't expect. May need to verify that the command enable/disable behavior doesn't try to use the codespaces URL and instead uses it from the endpoint annotation.

mitchdenny: /cc @DamianEdwards

src/Aspire.Hosting/DistributedApplicationBuilder.cs
mitchdenny: Updated to a switch statement just for you :)

karolz-ms (approved):

tests/Aspire.Hosting.Tests/WithHttpCommandTests.cs
mitchdenny: https://github.com/dotnet/aspire/issues/8070

davidfowl (approved):

Issue #7907, opened 2025-03-05
Support deep-linking to child resources in CosmosDB, EventHubs, ServiceBus and Storage


eerhardt:
The next step after [EventHubs, ServiceBus, and CosmosDB Hosting integrations should create Resources for children (dotnet/aspire#7407)](https://github.com/dotnet/aspire/issues/7407) is that we should support "deep linking" to the child resources and have corresponding client integration APIs for adding the child directly.

For example, in CosmosDB:

AppHost
```C#
var cosmos = builder.AddAzureCosmosDB("cosmos")
                .RunAsPreviewEmulator(e => e.WithDataExplorer());

var db = cosmos.AddCosmosDatabase("db");
var entries = db.AddContainer("entries", "/id");

builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
       .WithExternalHttpEndpoints()
       .WithReference(cosmos) // adds 'cosmos' connection string pointing to root service
       .WithReference(db) // adds 'db' connection string pointing to the root service + ";Database=db"
       .WithReference(entries); // adds 'entries' connection string pointing to the root service + ";Database=db" + ";Container=entries"
```

Then on the app side, we should have corresponding APIs for directly referencing these children:
```C#
builder.AddAzureCosmosClient("cosmos"); // exists today
builder.AddAzureCosmosDatabaseClient("db"); // new, registers a Microsoft.Azure.Cosmos.Database in DI
builder.AddAzureCosmosContainerClient("entries"); // new, registers a Microsoft.Azure.Cosmos.Container in DI
```

On CosmosDB EF, this will simply the existing code:

```C#
builder.AddCosmosDbContext<TestCosmosContext>("cosmos", "db"); // today
builder.AddCosmosDbContext<TestCosmosContext>("db"); // new - gets all connection information from the 'db' connection string
```

**QUESTION**: What happens if you `AddCosmosDbContext` to a "container"? Today AFAIK EF defaults to using the DbContext TypeName as the container name. Maybe we can change that to the specified container if you deep-link to that container.

We should do similar APIs for EventHubs and ServiceBus as well, where it makes sense.

We should also remember to update the Azure Functions config support as well.

### Resource Types Requiring Support

- [x] CosmosDB
- [x] Service Bus
- [x] EventHubs
- [ ] Azure Storage
- [x] OpenAI https://github.com/dotnet/aspire/pull/6577
- [x] Azure WebPubSub Hub

eerhardt:
[...]

PR #8068, opened 2025-03-14
Support CosmosDB child resources in client integration

captainsafia:
Contributes towards https://github.com/dotnet/aspire/issues/7907.

Note: this doesn't include support for setting the container on the DbContext for Cosmos. Apparently, in addition to setting it on a per-entity basis in the `OnModelCreating` method, you can configure an [IModelInitializedConvention](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.metadata.conventions.imodelinitializedconvention?view=efcore-9.0) that will run implicitl when models are created. However, EF Core only respects these conventions when they are added to its internal service provider and that requires some additional code. I opted not to include this in the initial PR.

### DIFF
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
@@ -13,9 +13,8 @@ var container = db.AddContainer("entries", "/id");
 
 builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
-       .WithReference(cosmos).WaitFor(cosmos)
-       .WithReference(db)
-       .WithReference(container);
+       .WithReference(db).WaitFor(db)
+       .WithReference(container).WaitFor(container);
 
 #if !SKIP_DASHBOARD_REFERENCE
 // This project is only added in playground projects to support development/debugging
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
@@ -9,11 +9,13 @@ var cosmos = builder.AddAzureCosmosDB("cosmos")
                 .RunAsPreviewEmulator(e => e.WithDataExplorer());
 
 var db = cosmos.AddCosmosDatabase("db");
-db.AddContainer("entries", "/id");
+var container = db.AddContainer("entries", "/id");
 
 builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
-       .WithReference(cosmos).WaitFor(cosmos);
+       .WithReference(cosmos).WaitFor(cosmos)
+       .WithReference(db)
+       .WithReference(container);
 
 #if !SKIP_DASHBOARD_REFERENCE
 // This project is only added in playground projects to support development/debugging
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -38,8 +38,17 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
     public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database container
-    void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName) =>
-        ((IResourceWithAzureFunctionsConfig)Parent).ApplyAzureFunctionsConfiguration(target, connectionName);
+    void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
+    {
+        if (Parent.Parent.IsEmulator || Parent.Parent.UseAccessKeyAuthentication)
+        {
+            Parent.Parent.SetConnectionString(target, connectionName, ConnectionStringExpression);
+        }
+        else
+        {
+            Parent.Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+        }
+    }
 
     private static string ThrowIfNullOrEmpty([NotNull] string? argument, [CallerArgumentExpression(nameof(argument))] string? paramName = null)
     {
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -50,6 +50,10 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
         else
         {
             Parent.Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+            target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__DatabaseName"] = Parent.DatabaseName;
+            target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__DatabaseName"] = Parent.DatabaseName;
+            target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__ContainerName"] = ContainerName;
+            target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__ContainerName"] = ContainerName;
         }
     }
 
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -36,9 +36,7 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
     /// Gets the connection string expression for the Azure Cosmos DB Database Container.
     /// </summary>
     public ReferenceExpression ConnectionStringExpression =>
-        Parent.Parent.IsEmulator || Parent.Parent.UseAccessKeyAuthentication
-            ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}")
-            : Parent.Parent.ConnectionStringExpression;
+        ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database container
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -35,7 +35,10 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB Database Container.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}");
+    public ReferenceExpression ConnectionStringExpression =>
+        Parent.Parent.IsEmulator || Parent.Parent.UseAccessKeyAuthentication
+            ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}")
+            : Parent.Parent.ConnectionStringExpression;
 
     // ensure Azure Functions projects can WithReference a CosmosDB database container
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -35,7 +35,7 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB Database Container.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database container
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName) =>
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
@@ -47,7 +47,7 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
         }
         else
         {
-            Parent.Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+            Parent.Parent.SetAccountEndpoint(target, connectionName);
             target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__DatabaseName"] = Parent.DatabaseName;
             target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__DatabaseName"] = Parent.DatabaseName;
             target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__ContainerName"] = ContainerName;
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -35,7 +35,7 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB database.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName) =>
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -35,7 +35,10 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB database.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}");
+    public ReferenceExpression ConnectionStringExpression =>
+        Parent.IsEmulator || Parent.UseAccessKeyAuthentication
+            ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}")
+            : Parent.ConnectionStringExpression;
 
     // ensure Azure Functions projects can WithReference a CosmosDB database
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -38,7 +38,7 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
     public ReferenceExpression ConnectionStringExpression =>
         Parent.IsEmulator || Parent.UseAccessKeyAuthentication
             ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}")
-            : Parent.ConnectionStringExpression;
+            : ReferenceExpression.Create($"AccountEndpoint={Parent.ConnectionStringExpression};Database={DatabaseName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -50,6 +50,8 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
         else
         {
             Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+            target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__DatabaseName"] = DatabaseName;
+            target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__DatabaseName"] = DatabaseName;
         }
     }
 
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -49,7 +49,7 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
         }
         else
         {
-            Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+            Parent.SetAccountEndpoint(target, connectionName);
             target[$"Aspire__Microsoft__EntityFrameworkCore__Cosmos__{connectionName}__DatabaseName"] = DatabaseName;
             target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__DatabaseName"] = DatabaseName;
         }
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
@@ -38,8 +38,17 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
     public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database
-    void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName) =>
-        ((IResourceWithAzureFunctionsConfig)Parent).ApplyAzureFunctionsConfiguration(target, connectionName);
+    void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
+    {
+        if (Parent.IsEmulator || Parent.UseAccessKeyAuthentication)
+        {
+            Parent.SetConnectionString(target, connectionName, ConnectionStringExpression);
+        }
+        else
+        {
+            Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+        }
+    }
 
     private static string ThrowIfNullOrEmpty([NotNull] string? argument, [CallerArgumentExpression(nameof(argument))] string? paramName = null)
     {
--- a/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
+++ b/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
@@ -78,7 +78,7 @@ public class AzureCosmosDBResource(string name, Action<AzureResourceInfrastructu
         }
         else
         {
-            SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
+            SetAccountEndpoint(target, connectionName);
         }
     }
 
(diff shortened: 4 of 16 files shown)


src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
```diff
@@ -23,6 +23,47 @@ public static class AspireAzureEFCoreCosmosExtensions
...
+                    connectionName,
+                    cosmosConnectionInfo.DatabaseName,
+                    configureSettings,
+                    configureDbContextOptions);
+            }
+            else
+            {
+                throw new InvalidOperationException(
+                  "A DbContext could not be configured with this AddCosmosDbContext overload. "
+                  + $"Ensure the connection string '{connectionName}' contains a database name or use the overload that takes a database name as a parameter. ");
```
danmoseley, line 62: will this potentially contain passwords? afaik we don't want them in exception messages..

davidfowl: This is the name, not the value.

danmoseley: I shouldn't review on my phone :)

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -36,6 +36,42 @@ public static void AddAzureCosmosClient(
...
+    /// <param name="configureClientOptions">An optional method that can be used for customizing the <see cref="CosmosClientOptions"/>.</param>
+    /// <remarks>Reads the configuration from "Aspire:Microsoft:Azure:Cosmos" section.</remarks>
+    /// <exception cref="InvalidOperationException">If required ConnectionString is not provided in configuration section</exception>
+    public static void AddAzureCosmosDatabase(
+        this IHostApplicationBuilder builder,
+        string connectionName,
+        Action<MicrosoftAzureCosmosSettings>? configureSettings = null,
+        Action<CosmosClientOptions>? configureClientOptions = null)
+    {
+        AddAzureCosmosClient(builder, configureSettings, configureClientOptions, connectionName, serviceKey: null);
```
eerhardt, line 54: How does this know to add the `Database` as a DI service? What's the difference between this and `AddAzureCosmosClient`?

From looking at the code, it appears to be based on the connection string and not which API was called.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -119,10 +197,46 @@ private static void AddAzureCosmosClient(
...
+            // Configure database and container services if specified
+            if (!string.IsNullOrEmpty(databaseName) && string.IsNullOrEmpty(containerName))
+            {
+                builder.Services.AddSingleton(sp =>
+                {
+                    var client = sp.GetRequiredService<CosmosClient>();
+                    return client.GetDatabase(databaseName);
+                });
+
+            }
```
eerhardt, line 209: Should the `Database` be added when both it and the container are specified?

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
```diff
@@ -23,6 +23,47 @@ public static class AspireAzureEFCoreCosmosExtensions
...
+    public static void AddCosmosDbContext<[DynamicallyAccessedMembers(RequiredByEF)] TContext>(
+        this IHostApplicationBuilder builder,
+        string connectionName,
+        Action<EntityFrameworkCoreCosmosSettings>? configureSettings = null,
+        Action<DbContextOptionsBuilder>? configureDbContextOptions = null) where TContext : DbContext
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrEmpty(connectionName);
+
+        if (builder.Configuration.GetConnectionString(connectionName) is string connectionString)
```
eerhardt, line 46: What happens when the connection string isn't provided? Typically we still try to add the services and we expect the app code in `configureSettings` to provide the connection information - instead of the connection string.

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
```diff
@@ -23,6 +23,47 @@ public static class AspireAzureEFCoreCosmosExtensions
     private const string DefaultConfigSectionName = "Aspire:Microsoft:EntityFrameworkCore:Cosmos";
     private const DynamicallyAccessedMemberTypes RequiredByEF = DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors | DynamicallyAccessedMemberTypes.PublicProperties;
 
+    /// <summary>
+    /// Registers the given <see cref="DbContext" /> as a service in the services provided by the <paramref name="builder"/>.
+    /// Derives the name of the database from the connection string.
```
eerhardt, line 28: Should it also derive the name of the container (if specified)? Is that even possible from the outside of the DbContext?

cc @AndriySvyryd - is it possible to set the CosmosDB container name outside of the DbContext?

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: > From looking at the code, it appears to be based on the connection string and not which API was called.

Correct -- we can probably shift the parsing of the connection string earlier and throw if the string associated with `connectionName` doesn't contain all the parts associated with the resource type you're trying to add.

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
AndriySvyryd: This method should throw if the connection string isn't provided.
Advanced usage where the connection string is set later would have to use `AddDbContext` with [UseCosmos](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.cosmosdbcontextoptionsextensions.usecosmos?view=efcore-9.0#microsoft-entityframeworkcore-cosmosdbcontextoptionsextensions-usecosmos(microsoft-entityframeworkcore-dbcontextoptionsbuilder-system-action((microsoft-entityframeworkcore-infrastructure-cosmosdbcontextoptionsbuilder)))) and `EnrichCosmosDbContext`

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
AndriySvyryd: By default, the container name is derived from the context type name. As @captainsafia mentioned in the description, the default container name can be set using a convention, but it requires some additional infrastructure.

src/Shared/Cosmos/CosmosUtils.cs
```diff
@@ -49,12 +49,12 @@ internal static CosmosConnectionInfo ParseConnectionString(string connectionStri
 
         // Strip out the database and container from the connection string in order
         // to tell if we are left with just AccountEndpoint.
-        if (connectionBuilder.TryGetValue("Database", out var _))
+        if (connectionBuilder.TryGetValue("Database", out var databaseValue))
```
AndriySvyryd, line 52: Where are these connection string options documented?

src/Shared/Cosmos/CosmosUtils.cs
```diff
@@ -49,12 +49,12 @@ internal static CosmosConnectionInfo ParseConnectionString(string connectionStri
...
         {
             connectionBuilder["Database"] = null;
         }
 
-        if (connectionBuilder.TryGetValue("Container", out var _))
+        if (connectionBuilder.TryGetValue("Container", out var containerValue))
```
AndriySvyryd, line 57: Why does this need to affect the `DbContext` configuration?

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: I think this ties a bit into the conversation we were having about what services get configured in the ServiceBus scenario. 

IMO, it makes sense to limit it to just Client + Container for this overload. You can always resolve the database yourself with the client that we inject.

src/Shared/Cosmos/CosmosUtils.cs
eerhardt: I haven't found a great place to document them yet. We can put them somewhere on https://learn.microsoft.com/dotnet/aspire/database/azure-cosmos-db-integration.

This is part of a larger effort of allowing for "deep linking" to child resources in Aspire.
* [EventHubs, ServiceBus, and CosmosDB Hosting integrations should create Resources for children (dotnet/aspire#7407)](https://github.com/dotnet/aspire/issues/7407)
* [Support deep-linking to child resources in CosmosDB, EventHubs, and ServiceBus (dotnet/aspire#7907)](https://github.com/dotnet/aspire/issues/7907)

We only have the connection string as the ability to pass connection information from the AppHost to the apps. So adding more connection string segments is the path we are going down.

AndriySvyryd: We should add checks that the connection string supplied by the user doesn't contain these.

The names should be extracted to static fields

captainsafia: > We should add checks that the connection string supplied by the user doesn't contain these.

https://github.com/dotnet/aspire/pull/8068/commits/442b4e53326f6a2ada43c0535b53bc81911285eb updated the implementation to throw eagerly if the connection string doesn't contain the required segments.

AndriySvyryd: You would also need to add checks the other way around: `AddAzureCosmosClient` should throw if the connection string contains database or container name and `AddAzureCosmosDatabase` should throw if the connection string contains container name. Otherwise, the connection strings produced for child resources will be invalid.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: Per offline discussion, we no longer register the client if you call the specific `AddAzureCosmosDatabase` or `AddAzureCosmosContainer` method.

src/Shared/Cosmos/CosmosUtils.cs
captainsafia: @AndriySvyryd Hmmm....I dunno if we need to through exceptions in this case. Even if a connection string contains a database or container name, it's still a valid connection string to initialize a CosmosClient from since it will have an  endpoint name/connection string. It feels too restrictive to prevent you from an initializing a more general client from a more specific connection string.

AndriySvyryd: Perhaps. So now if you wanted to add all the services you could use the same connection?
```C#
builder.AddAzureCosmosClient("MyCosmosContainer");
builder.AddAzureCosmosDatabase("MyCosmosContainer");
builder.AddAzureCosmosContainer("MyCosmosContainer");
```

AndriySvyryd: I guess I am more concerned about the Hosting side when using the [AddConnectionString](https://learn.microsoft.com/en-us/dotnet/api/aspire.hosting.parameterresourcebuilderextensions.addconnectionstring) method to specify a connection string. If it contains `Database` or `Container` we should throw

src/Shared/Cosmos/CosmosUtils.cs
captainsafia: It doesn't affect the `DbContext` at all at the moment. This is just shared code that ends up getting used in the non-EF scenario.

The EF-specific overloads currently disregard the container value if it's set. This will work correctly assuming we can implement the [IModelInitializedConvention](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.metadata.conventions.imodelinitializedconvention?view=efcore-9.0) API you referenced.

src/Shared/Cosmos/CosmosUtils.cs
captainsafia: Hmmm....I don't see why we should? If an end-user independently does the work of configuring a connection string with `Database` and `Container` clauses should prohibit them from initializing a client. What would be concerning there?

AndriySvyryd: To avoid generating invalid connection strings. To see what  I mean add a test that adds a cosmos resource with a database and a container, then uses the connection string from container resource

src/Shared/Cosmos/CosmosUtils.cs
AndriySvyryd: What do you mean by "correctly"? I don't understand why this needs to have any impact on the `DbContext`. I don't think the additional code we discussed should be added

src/Shared/Cosmos/CosmosUtils.cs
captainsafia: Do you mean having something like the following in the AppHost:

```

var container = db.AddContainer("entries", "/id");

builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
       .WithExternalHttpEndpoints()
       .WithReference(container);
```

Then the following in the ApiService:

```csharp
builder.AddAzureCosmosClient("entries");
```

This would still work because the `ConnectionString_entries` would be something like `AccountKey=keyhere;AccountEndpoint=http://localhost:53590;Database=db;Container=entries` and the connection string builder will pull out just the AccountEndpoint from it.

src/Shared/Cosmos/CosmosUtils.cs
captainsafia: > What do you mean by "correctly"? I

Correctly here means that the container name associated with the connection string we are parsing is set as the container for entities registered in the context.

> I don't think the additional code we discussed should be added

Is your recommendation for the EF integration that we disregard the container name and always assume the default container name is resolved from the name of the DbContext.

AndriySvyryd: > Is your recommendation for the EF integration that we disregard the container name and always assume the default container name is resolved from the name of the DbContext.

Yes, because the default container name is used only if the container name for a given entity type is not set. So changing the default might not have any effect if all container names are configured explicitly. Also, it wouldn't have any effect if the user is relying on a pre-compiled model as by the time this code runs the model is already built.

src/Shared/Cosmos/CosmosUtils.cs
AndriySvyryd: I meant using `AddConnectionString` and adding the container as a child resource, but I see now that it doesn't work that way.

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
```diff
@@ -9,11 +9,13 @@
                 .RunAsPreviewEmulator(e => e.WithDataExplorer());
 
 var db = cosmos.AddCosmosDatabase("db");
-db.AddContainer("entries", "/id");
+var container = db.AddContainer("entries", "/id");
 
 builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
-       .WithReference(cosmos).WaitFor(cosmos);
+       .WithReference(cosmos).WaitFor(cosmos)
```
eerhardt, line 16: (nit) can remove WithReference(cosmos) since the app doesn't use it.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
```diff
@@ -35,7 +35,7 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB Database Container.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Container={ContainerName}");
 
     // ensure Azure Functions projects can WithReference a CosmosDB database container
```
eerhardt, line 40: How is `ApplyAzureFunctionsConfiguration` going to work here? Do we need to add the child connection string there as well?

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
eerhardt, line 62:
```suggestion
                  + $"Ensure the connection string '{connectionName}' contains a database name or use the overload that takes a database name as a parameter.");
```

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
eerhardt: > This method should throw if the connection string isn't provided.

We also support setting the connection properties via the `Action<EntityFrameworkCoreCosmosSettings>? configureSettings` parameter for the cases where apps need to retrieve this information from somewhere other than the connection string section.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -116,33 +260,27 @@ private static void AddAzureCosmosClient(
         }
 
         clientOptions.ApplicationName = cosmosApplicationName;
-        if (serviceKey is null)
+
+        return clientOptions;
+    }
+
+    private static CosmosClient ConfigureDb(string connectionName, MicrosoftAzureCosmosSettings settings, CosmosClientOptions clientOptions)
```
eerhardt, line 267:
```suggestion
    private static CosmosClient GetCosmosClient(string connectionName, MicrosoftAzureCosmosSettings settings, CosmosClientOptions clientOptions)
```

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -101,6 +243,8 @@ private static void AddAzureCosmosClient(
             });
         }
 
+        configureClientOptions?.Invoke(clientOptions);
```
eerhardt, line 246: Why are we calling this here and on line 254?

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -33,11 +33,73 @@ public static void AddAzureCosmosClient(
...
+        this IHostApplicationBuilder builder,
+        string connectionName,
+        Action<MicrosoftAzureCosmosSettings>? configureSettings = null,
+        Action<CosmosClientOptions>? configureClientOptions = null)
+    {
+        var cosmosConnectionInfo = GetCosmosConnectionInfo(builder, connectionName);
+        if (cosmosConnectionInfo is null || string.IsNullOrEmpty(cosmosConnectionInfo.Value.DatabaseName))
+        {
+            throw new InvalidOperationException($"The connection string '{connectionName}' does not exist or is missing the database name.");
+        }
```
eerhardt, line 62: Typically we don't throw up front if the connection string isn't passed, but instead wait to see if the information is set in `Action<MicrosoftAzureCosmosSettings>? configureSettings`. Only then do we fail if we couldn't get the info we needed. The connection string initially populates the settings object, and then the app code gets a chance to modify it as well.

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
AndriySvyryd: > We also support setting the connection properties via the Action<EntityFrameworkCoreCosmosSettings>? configureSettings parameter for the cases where apps need to retrieve this information from somewhere other than the connection string section.

If both are specified which one should take precedence?

captainsafia: > If both are specified which one should take precedence?

I've implemented it such that the name defined in `configureSettings` takes precedence.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: Missed this while refactoring the method. Fixed.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: Delayed the exception to occur when the instance is resolved and updated the settings objects to store the database and/or container name.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
captainsafia: Functions is going to be interesting. I've update the implementation so that:

- We always inject the connection string associated with the `CosmosDbResource` under the connection key that is used by the Functions host.
- For the named configuration that the Aspire integrations use, we inject the connection string associated with the Azure resource.

One thing to note is you can have slight weirdness where your AppHost has:

```csharp
var cosmosDb = builder.AddAzureCosmosDB("cosmosdb")
    .RunAsEmulator();
var database = cosmosDb.AddCosmosDatabase("mydatabase");
var container = database.AddContainer("mycontainer", "/id");

builder.AddAzureFunctionsProject<Projects.AzureFunctionsEndToEnd_Functions>("funcapp")
  .WithReference(container);
```

You can resolve the container via Aspire's DI injection like so:

```csharp
var builder = FunctionsApplication.CreateBuilder(args);

builder.AddAzureCosmosContainer("mycontainer");
```

But your trigger still has to be defined fully like this:

```csharp
[Function(nameof(MyCosmosDbTrigger))]
    public void Run([CosmosDBTrigger(
        databaseName: "mydatabase",
        containerName: "mycontainer",
        CreateLeaseContainerIfNotExists = true,
        Connection = "cosmosdb")] IReadOnlyList<Document> input)
```

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
eerhardt: > I've implemented it such that the name defined in configureSettings takes precedence.

👍 That follows the same pattern in all our client integrations.

* Settings object is bound to the configuration
* Values coming from the connection string are applied to the settings object
* configureSettings is called, so code and inspect and modify it.

src/Components/Aspire.Microsoft.EntityFrameworkCore.Cosmos/AspireAzureEFCoreCosmosExtensions.cs
```diff
@@ -23,6 +23,48 @@ public static class AspireAzureEFCoreCosmosExtensions
...
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrEmpty(connectionName);
+
+        if (builder.Configuration.GetConnectionString(connectionName) is string connectionString)
+        {
+            var cosmosConnectionInfo = CosmosUtils.ParseConnectionString(connectionString);
+            if (cosmosConnectionInfo.DatabaseName is not null)
+            {
+                var targetDatabaseName = cosmosConnectionInfo.DatabaseName;
+                AddCosmosDbContext<TContext>(
```
eerhardt, line 52: I think we should move this call outside of the `if` check. If for some reason you don't have a connection string with `connectionName` this method is going to no-op and then your app is going to fail with "no CosmosDbContext service registered" errors - which will be confusing. Instead, we should always call `AddCosmosDbContext` which will eventually fail because it can't find a connection string (the real problem).

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBContainerResource.cs
```diff
@@ -35,11 +35,20 @@ public class AzureCosmosDBContainerResource(string name, string containerName, s
...
-        ((IResourceWithAzureFunctionsConfig)Parent).ApplyAzureFunctionsConfiguration(target, connectionName);
+    void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
+    {
+        if (Parent.Parent.IsEmulator || Parent.Parent.UseAccessKeyAuthentication)
+        {
+            Parent.Parent.SetConnectionString(target, connectionName, ConnectionStringExpression);
+        }
+        else
+        {
+            Parent.Parent.SetAccountEndpoint(target, connectionName, ConnectionStringExpression);
```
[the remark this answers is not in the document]
captainsafia: Tweaked this by changing the `ConnectionStringExpression` property to omit the `;Database` and `;Container` values when we're using an account endpoint.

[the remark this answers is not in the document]
AndriySvyryd: Then how the database and container name will be supplied when using an account endpoint? Am I missing something?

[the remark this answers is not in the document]
AndriySvyryd: But that's only called for Azure Functions, right? What about just Cosmos DB with an account endpoint?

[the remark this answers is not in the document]
AndriySvyryd: Couldn't we always do it through settings like that instead of trying to cram them in the connection string?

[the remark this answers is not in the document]
eerhardt: We've gone around and around on connection strings in Aspire. We even have a proposal: [Add support for Connection Properties (dotnet/aspire#1765)](https://github.com/dotnet/aspire/issues/1765) to add a more structred "string-string" key value pair object to separate these things out. But the current prevailing approach is to cram the "reference to the resource" into the connection string.

In the end, a string-string key vaule pair vs. one string with the keys and values concatenated is almost the same thing.

The higher level reasoning here is because Aspire operates on `.WithReference`:

https://github.com/dotnet/aspire/blob/89e4cb75f7b3105810e66b0c918d713c4f80bf54/src/Aspire.Hosting/ResourceBuilderExtensions.cs#L345-L361

And here we only know the `source` is `IResourceWithConnectionString`. We don't know all the places a resource might want to stick configuration settings. We solved this half way with Functions for a subset of resources - ones that implement `IResourceWithAzureFunctionsConfig`.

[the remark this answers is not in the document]
AndriySvyryd: Ok, but while the connection string is used, something still appears to be missing for the scenario of Cosmos DB without Access Key authentication with database and container resources when used from a non-functions project.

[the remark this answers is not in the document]
captainsafia: @AndriySvyryd Yep, you're correct. We need to massage everything into a connection string when using child resources on a `WithReference`. Commit incoming.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
```diff
@@ -35,11 +35,25 @@ public class AzureCosmosDBDatabaseResource(string name, string databaseName, Azu
     /// <summary>
     /// Gets the connection string expression for the Azure Cosmos DB database.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
+    public ReferenceExpression ConnectionStringExpression =>
+        Parent.IsEmulator || Parent.UseAccessKeyAuthentication
+            ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression};Database={DatabaseName}")
+            : ReferenceExpression.Create($"AccountEndpoint={Parent.ConnectionStringExpression};Database={DatabaseName}");
```
AndriySvyryd, line 41: Can't `AccountEndpoint` be added in `AzureCosmosDBResource`? This would move the logic dealing with this to one place.

captainsafia: I was leaning towards not massaging things into a connection string until we absolutely needed it. So I do it here in the Database resource and don't use the conditional in the Container resource.

AndriySvyryd (approved):

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBDatabaseResource.cs
eerhardt: I think it would be better if we followed the same pattern as EventHubs:

https://github.com/dotnet/aspire/blob/08b424e1fa0b28aa225a67e2c6a7b5b959037f05/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs#L50-L82

https://github.com/dotnet/aspire/blob/08b424e1fa0b28aa225a67e2c6a7b5b959037f05/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubConsumerGroupResource.cs#L35

The child resources call this method on the parent, and the parent builds up the connection string.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -74,22 +74,36 @@ void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDiction
...
+        target[$"Aspire__Microsoft__Azure__Cosmos__{connectionName}__ConnectionString"] = connectionStringExpression;
+    }
+
+    internal void SetAccountEndpoint(IDictionary<string, object> target, string connectionName)
+    {
+        // Always inject the connection string associated with the top-level resource
+        // for the Azure Functions host.
+        target[$"{connectionName}__accountEndpoint"] = ConnectionStringExpression;
+        // Injected to support Aspire client integration for CosmosDB in Azure Functions projects.
+        // Use the child resource connection string here to support child resource integrations.
```
eerhardt, line 102: What does this comment mean? `Use the child resource connection string here`, but the code uses `this.ConnectionStringExpression`.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -101,48 +236,42 @@ private static void AddAzureCosmosClient(
...
 
+        configureClientOptions?.Invoke(clientOptions);
+
         if (CosmosUtils.IsEmulatorConnectionString(settings.ConnectionString))
         {
             clientOptions.ConnectionMode = ConnectionMode.Gateway;
             clientOptions.LimitToEndpoint = true;
         }
 
-        configureClientOptions?.Invoke(clientOptions);
```
eerhardt, line 110: Why did we change the order of the calls here? The intention is to give the developer the last chance to change options. So by moving this call above the IsEmulatorConnectionString check, we are always going to overwrite the above 2 options.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
captainsafia: Yes, this is copy-pasta. 🍝

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: This wasn't intentionally. I'll try to figure out a way for us to test the ordering behavior here. 🤔

PR #8081, opened 2025-03-14
[CI] Playwright on Linux for PR validation

radical:
- Install playwright on Linux/GithubActions
-  Add RequiresPlaywright attributes, and discoverer and update tests to use that instead of an ActiveIssue
- Collect logs from template test runs on Github Actions

This enables 16 additional template tests to run on Linux/GHA.


RussKie (approved):
Looks sensible

PR #8085, opened 2025-03-15
Create database for Postgres resource

sebastienros:
## Description

Support database creation for `AddDatabase()` in Postgres container resource. If the database already exists nothing is done. Custom scripts can be provided.

Updated PgAdmin container version.

Part of https://github.com/dotnet/aspire/issues/7101

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No

### DIFF
--- a/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
+++ b/src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
@@ -575,7 +575,7 @@ public static class PostgresBuilderExtensions
             }
             catch (Exception e)
             {
-                var logger = serviceProvider.GetRequiredService<ILogger<DistributedApplicationBuilder>>();
+                var logger = serviceProvider.GetRequiredService<ResourceLoggerService>().GetLogger(npgsqlDatabase.Parent);
                 logger.LogError(e, "Failed to create database '{DatabaseName}'", npgsqlDatabase.DatabaseName);
             }
         }
@@ -591,7 +591,7 @@ public static class PostgresBuilderExtensions
             }
             catch (Exception e)
             {
-                var logger = serviceProvider.GetRequiredService<ILogger<DistributedApplicationBuilder>>();
+                var logger = serviceProvider.GetRequiredService<ResourceLoggerService>().GetLogger(npgsqlDatabase.Parent);
                 logger.LogError(e, "Failed to create database '{DatabaseName}'", npgsqlDatabase.DatabaseName);
             }
         }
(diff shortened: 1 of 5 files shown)


src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -488,4 +554,46 @@ private static string WritePgAdminServerJson(IEnumerable<PostgresServerResource>
...
+        {
+            // User-provided script
+
+            try
+            {
+                using var command = npgsqlConnection.CreateCommand();
+                command.CommandText = scriptAnnotation?.Script;
+                await command.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception e)
```
davidfowl, line 592: Don't we want to handle `            catch (PostgresException p) when (p.SqlState == "42P04")` here too? Or does the

src/Aspire.Hosting.PostgreSQL/PostgresDatabaseResource.cs
```diff
@@ -23,9 +24,18 @@ public class PostgresDatabaseResource(string name, string databaseName, Postgres
...
-    public ReferenceExpression ConnectionStringExpression =>
-       ReferenceExpression.Create($"{Parent};Database={DatabaseName}");
+    public ReferenceExpression ConnectionStringExpression
+    {
+        get
+        {
+            var connectionStringBuilder = new DbConnectionStringBuilder
+            {
+                ["Database"] = DatabaseName
+            };
```
davidfowl, line 34: What's special about this that we need to use the builder? Does it escape properly?

sebastienros: It does, and there is a functional test to validate we can pass database name with special chars to ensure the connection string and the default CREATE DATABASE script are correctly escaping.

sebastienros: I did it for Sql Server too btw.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
sebastienros: Since the script could contain other things than create database I didn't want to hide errors. But now it's only supporting create db so it's doing the same thing.

eerhardt (approved):

PR #7575, opened 2025-02-12
Adding test coverage - validate arguments of public methods

Zombach:
Initial issues [#2649](https://github.com/dotnet/aspire/issues/2649)
Issues [#5047](https://github.com/dotnet/aspire/issues/5047)
[message](https://github.com/dotnet/aspire/issues/5047#issuecomment-2278838761)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
  - [x] No
- Does the change require an update in our Aspire docs?
  - [ ] Yes
  - [x] No

Added or Update public API test coverage for:
# Components
- [x] Aspire.Azure.Ai.OpenAi
- [x] Aspire.Azure.Data.Tables
- [x] Aspire.Azure.Messaging.EventHubs
- [x] Aspire.Azure.Messaging.ServiceBus
- [x] Aspire.Azure.Messaging.WebPubSub
- [x] Aspire.Azure.Search.Documents
- [x] Aspire.Azure.Security.KeyVault
- [x] Aspire.Azure.Storage.Blobs
- [x] Aspire.Azure.Storage.Queues
- [x] Aspire.Confluent.Kafka
- [x] Aspire.Elastic.Clients.Elasticsearch
- [x] Aspire.Keycloak.Authentication
- [x] Aspire.Microsoft.Azure.Cosmos
- [x] Aspire.Microsoft.Data.Sql.Client
- [x] Aspire.Microsoft.EntityFrameworkCore.Cosmos
- [x] Aspire.Microsoft.EntityFrameworkCore.SqlServer
- [x] Aspire.Milvus.Client
- [x] Aspire.MongoDB.Driver
- [x] Aspire.MySqlConnector
- [x] Aspire.NATS.Net
- [x] Aspire.Confluent.Kafka
- [x] Aspire.Npgsql.EntityFrameworkCore.PostgreSQL
- [x] Aspire.Npgsql
- [x] Aspire.OpenAI
- [x] Aspire.Oracle.EntityFrameworkCore
- [x] Aspire.Pomelo.EntityFrameworkCore.MySql
- [x] Aspire.Qdrant.Client
- [x] Aspire.RabbitMQ.Client
- [x] Aspire.Seq
- [x] Aspire.StackExchange.Redis.DistributedCaching
[...]

PR #8098, opened 2025-03-16
share-argument-exception-extension

Zombach:
## Description

[message](https://github.com/dotnet/aspire/pull/7575#discussion_r1976482906)

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No


Zombach:
@danmoseley Hello. Please look. Made a check share code

src/Aspire.Hosting.Garnet/GarnetBuilderExtensions.cs
```diff
@@ -54,7 +54,7 @@ public static class GarnetBuilderExtensions
     public static IResourceBuilder<GarnetResource> AddGarnet(this IDistributedApplicationBuilder builder, [ResourceName] string name,
         int? port)
     {
-        return builder.AddGarnet(name, port, password: null);
+        return AddGarnet(builder, name, port, password: null);
```
danmoseley, line 57: why this change?

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -39,7 +39,7 @@ public static class RedisBuilderExtensions
     /// </remarks>
     public static IResourceBuilder<RedisResource> AddRedis(this IDistributedApplicationBuilder builder, [ResourceName] string name, int? port)
     {
-        return builder.AddRedis(name, port, null);
+        return AddRedis(builder, name, port, null);
```
danmoseley, line 42: same question. style?

src/Components/Common/ArgumentExceptionExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+internal static class ArgumentExceptionExtensions
+{
+    /// <summary>
+    /// Adds a check Throw ArgumentException if argument is null or empty.
+    /// And return argument
+    /// </summary>
+    /// <param name="argument"></param>
+    /// <param name="paramName"></param>
+    /// <returns>argument not null</returns>
+    public static string ThrowIfNullOrEmpty(
```
danmoseley, line 18: style wise, not sure whether folks want all strings and string arrays in the codebase to have new methods or not. (i don't think I care, but in dotnet/runtime they would prefer explicit static methods)

@mitchdenny any opinion?

danmoseley: this team does love extension methods...

src/Aspire.Hosting.Redis/RedisResource.cs
```diff
@@ -16,6 +16,8 @@ public class RedisResource(string name) : ContainerResource(name), IResourceWith
     /// <param name="password">A parameter that contains the Redis server password.</param>
     public RedisResource(string name, ParameterResource password) : this(name)
     {
+        ArgumentNullException.ThrowIfNull(password);
```
danmoseley, line 19: where you've added new checks, do they need tests?

src/Aspire.Hosting.Garnet/GarnetBuilderExtensions.cs
Zombach: The problem is. What if we perform the line "Return Builder.adddganet (Name, Port, Password: Null);"
And Builder will come to us == null. Then we will get NULLREFERENEEXCEPTION

And if we use this option "Return Addganet (Builder, Name, Port, Password: Null);"
The same as in the first case, we will cause the overload of the method.
But in this case, we will not turn to Builder. Inside the overload, there is an audit on Argumentnullexception, which will check Builder and give a valid error.

src/Aspire.Hosting.Redis/RedisResource.cs
Zombach: I will add today. I will also add tests to "Builder.addgarnet ..." => "Addganet (Builder ..."
I just wanted to understand, this option is suitable for the implementation of the Share check

src/Components/Common/ArgumentExceptionExtensions.cs
danmoseley: I like that too.

Zombach:
@danmoseley If anything is ready, you can fill.

mitchdenny:
This looks good to me. Just need the covering tests for all the other resource types.

Zombach:
> This looks good to me. Just need the covering tests for all the other resource types.

And what exactly do I need to cover the tests in this MR?
I did not quite understand.
In my previous MR [#7575](https://github.com/dotnet/aspire/pull/7575) i have already covered these elements.

danmoseley (approved):

Zombach:
Here I came across several non -critical mistakes in tests.
Which depend on the operating system and on the localization of the system

image
In this screenshot, there is an error in due to '.' <=> ','

image
Here the problem arises due to the localization of the operating system on which the test is launched. 
Although the expected values ​​and the result of the test are equal in meaning, but still they are in different languages
Launch settings file does not contain 'not-exist' "··· Файл параметров запуска не содержит профиля "not-e"···

image
Also the problem of localization

I don't know how critical it is. But these are still errors

danmoseley:
Tests should pass in all locales

RussKie:
I _think_ there are (at least) two ways of solving this.

1. Tactical - set the culture for each test and unset at it at the end. E.g. something like this pseudocode:
    ```cs
        try
        {
             _originalCulture = Thread.CurrentThread.CurrentCulture;
             _originalUICulture = Thread.CurrentThread.CurrentUICulture;
 
             CultureInfo.DefaultThreadCurrentCulture = Culture;
             Thread.CurrentThread.CurrentCulture = "en-US";
             Thread.CurrentThread.CurrentUICulture = "en-US"
   
            // ... body of the test
        }
        finally
        {
             Thread.CurrentThread.CurrentCulture = _originalCulture;
             Thread.CurrentThread.CurrentUICulture = _originalUICulture;
        }
    ```
2. Pass the culture as a theory argument.
3. Strategic - add an attribute which can be applies to a test/fixture. Something like https://github.com/dotnet/winforms/commit/9a26e660c6f7f3663dac601a4bba64961e046fa1 (though it'd be simpler since there are no unmanaged resources here).

danmoseley:
@Zombach not about this PR directly, but chatting offline, I think we want to hold off on any future large "argument validation" PR's without agreement first. Yes, code changes should validate arguments. A PR fixing a particular instance is probably fine. But any more large scale ones don't make sense right now as we're biasing for speed, and it brings the risk of regressions.

Zombach:
> @Zombach not about this PR directly, but chatting offline, I think we want to hold off on any future large "argument validation" PR's without agreement first. Yes, code changes should validate arguments. A PR fixing a particular instance is probably fine. But any more large scale ones don't make sense right now as we're biasing for speed, and it brings the risk of regressions.

Yes, I will make Culture fix in a separate PR.
And i will roll back this one to the previous state (before Culture fix)

src/Aspire.Hosting.Azure.CognitiveServices/Aspire.Hosting.Azure.CognitiveServices.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)src\Shared\AzureRoleAssignmentUtils.cs" />
+    <Compile Include="$(ComponentsDir)\Common\ArgumentExceptionExtensions.cs" Link="ArgumentExceptionExtensions.cs" />
```
RussKie, line 13: Instead of doing this in each project file, I suggest considering a different approach.

* If this file needs to be linked in each project, then do this injection at Directory.Build.props/targets (or alike).
* If this file needs to be conditionally linked in project, then do a conditional injection at Directory.Build.props/targets (or alike) controlled by an opt-in switch in each project.
For example: https://github.com/dotnet/extensions/blob/1e4f6c51798be661b8d8743dd464ca9f54b0c9a3/src/Libraries/Microsoft.AspNetCore.HeaderParsing/Microsoft.AspNetCore.HeaderParsing.csproj#L15C6-L15C41

RussKie, line 13: Why not put the shared file in the Shared folder (like the line above)?

src/Components/Common/ArgumentExceptionExtensions.cs
```diff
@@ -0,0 +1,45 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Aspire;
+
+internal static class ArgumentExceptionExtensions
```
RussKie, line 9: These extensions aren't very discoverable. I very much suggest taking an inspiration from https://github.com/dotnet/extensions/tree/1e4f6c51798be661b8d8743dd464ca9f54b0c9a3/src/Shared/Throw, and do the same way (both in terms of API naming and the location of the file).

Zombach: Ooo. This option is really good. I will do the other day

danmoseley:
Let's merge this after the 9.3 snap

danmoseley:
@Zombach we are now open for 9.4 changes. can you resolve conflicts - is this ready to go?

PR #8102, opened 2025-03-17
pin GHA

danmoseley:
1. Pin GHA except for arcade one, to avoid exploits like https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised
2. Remove dependabot as its only remaining use is auto updating GHA, and it's easy to merge that without sufficient scrutiny. Better to stay on whatever is working.


danmoseley:
https://github.com/dotnet/aspire/issues/8103

davidfowl:
Who updates these shas?

danmoseley:
Who? Nobody, unless dependabot is set up for it. But we don't want to update them, at least without a reason.

jeffhandley (approved):

PR #8120, opened 2025-03-17
Add secret warning alert to text visualizer the first time you open it with a secret

### LINKED ISSUE #7713: In the dashoard make secrets revealed explicitly

I would like to see the dashboard follow this updated pattern to visualize text values with secrets:

1- In the resource properties, the environment variables and arguments are already redacted. We can "reveal", "show the secrets" by clicking on the `Show value` icon. This is very explicit, and the action of showing secrets is intentional.

Before we show the secrets:

Image

Once the secrets are shown:

Image

However the icon right to its left has no tooltip, it's looks like a magnifier. It doesn't seem to have a label, aria, or whatever information explaining what it does. When it's clicked it reveals a new windows showing the unredacted value, and other ways to visualize it (Formats) as shown in the following screenshot

Image

In my opinion I didn't choose to reveal the redacted value. I like the extra window but I would like the secrets to be masked until I decide it's not. Or a tooltip/visual information stating that it will reveal secrets when clicked, even if it's a confirmation when it's known to contain secrets.

2- The main page has the same behavior with command line arguments:

The arguments are correctly redacted, but the icon to open the text explorer, the same as in the previous point, has a tooltip that is the values with the text redacted, still no information about what it's about to reveal secrets if I click on it.

Image

1- Define a tooltip on this "data explorer" icon
2- Keep redacted information in the default view mode of this window
3- Create a button with the same "Show value" icon as in the properties that makes it explicit that secrets will be revealed.
4- Nit: "Show value" and the icon convey enough meaning right now, but to be picky we could change the tooltip of the action like "Show secret value" (like in KeyVault?), "Reveal redacted values", ...

Optionally if the data explorer can't redact secrets for technical reasons (like meta-data is lost), then make it clear that this popup shows secrets, with a better icon and a clear tooltip. But to me it would not be the ideal solution.

### Additional context

_No response_

adamint:
## Description

Adds an initial UI where the user has to explicitly acknowledge that secrets will be revealed automatically in the text visualizer in subsequent opens. An entry is stored in localStorage indicating that this UI has been showed to users and acknowledged, and will not appear again. @blowdart

This is the current UI, but definitely can be changed.
image

Fixes #7713

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No


src/Aspire.Dashboard/Components/Controls/ExceptionDetails.razor
```diff
@@ -29,6 +29,6 @@
 
     private void OnExceptionDetailsClicked(MouseEventArgs e)
```
JamesNK, line 30:
```suggestion
    private async Task OnExceptionDetailsClicked()
```
+ await opening the dialog.

I don't see a reason for this method not to be async.

JamesNK:
I tested in resource details and the masked environment variables didn't count as secrets.

image

JamesNK:
I opened from resource details on a property marked as a secret and it also displayed without showing the consent.

What counts as a secret?

image

JamesNK:
I think the flag passed to the dialog shouldn't be "ContainsSecret". It should be something like "IsSourceMasked".

If someone has unmasked all environment variables for example, then it doesn't make sense to show the secret warning when viewing it in details. They've already shown the value.

tests/Aspire.Dashboard.Components.Tests/Controls/TextVisualizerDialogTests.cs
```diff
@@ -136,16 +142,37 @@ public async Task Render_TextVisualizerDialog_ResolveTheme_LineClassesChange()
...
+        cut.WaitForAssertion(() => Assert.True(cut.HasComponent<TextVisualizerDialog>()));
+
+        Assert.True(cut.HasComponent<FluentMessageBar>());
+        Assert.False(cut.HasComponent<Virtualize<TextVisualizerDialog.StringLogLine>>());
+
+        cut.Find(".text-visualizer-unmask-content").Click();
+
+        cut.WaitForAssertion(() => Assert.False(cut.FindComponent<TextVisualizerDialog>().Instance.ShowContainsSecretsWarning));
+        Assert.False(cut.HasComponent<FluentMessageBar>());
+        Assert.True(cut.HasComponent<Virtualize<TextVisualizerDialog.StringLogLine>>());
```
JamesNK, line 167: Also assert that the expected value was written to local storage.

tests/Aspire.Dashboard.Components.Tests/Controls/TextVisualizerDialogTests.cs
```diff
@@ -136,16 +142,37 @@ public async Task Render_TextVisualizerDialog_ResolveTheme_LineClassesChange()
...
-        await dialogService.ShowDialogAsync<TextVisualizerDialog>(new TextVisualizerDialogViewModel(xml, string.Empty), []);
+        await dialogService.ShowDialogAsync<TextVisualizerDialog>(new TextVisualizerDialogViewModel(xml, string.Empty, false), []);
+        cut.WaitForAssertion(() => Assert.True(cut.HasComponent<TextVisualizerDialog>()));
 
         Assert.NotEmpty(cut.FindAll(".theme-a11y-dark-min"));
     }
 
-    private IRenderedFragment SetUpDialog(out IDialogService dialogService, ThemeManager? themeManager = null)
+    [Fact]
+    public async Task Render_TextVisualizerDialog_WithSecret_ShowsWarningFirstOpenAsync()
```
JamesNK, line 152: Add a second test that verifies the warning isn't displayed if the required value is set in local storage

adamint:
> I think the flag passed to the dialog shouldn't be "ContainsSecret". It should be something like "IsSourceMasked".
> 
> If someone has unmasked all environment variables for example, then it doesn't make sense to show the secret warning when viewing it in details. They've already shown the value.

I changed the approach a little bit. The flag should be true if the GridValue's IsMasked is true, or if it is a value like one with secrets that is partially masked (with IsMasked being false). This previously just worked with the source column for secrets, now it also works for the environment variables.

JamesNK (approved):

JamesNK:
I made a minor change to add a min height to the content and display button at bottom.

PR #8121, opened 2025-03-17
Add support for creating files and folders in a container via `docker/podman cp`

danegsta:
## Description

Adds a new `WithContainerFiles` API method that can be used to create files and folders in a container when it is run. This is intended as an alternative to existing usage of `WithBindMount` to copy temporary files into a container for configuration purposes. This method offers several advantages over `WithBindMount` in that the ownership and permissions of individual files and folders can be controlled without having to resort to applying permissions to local temporary files.

The design is intended to deal with locally ephemeral files, so does not read from the local file system. Due to the size constraints of resources registered for orchestration, it only makes sense for simple configuration files as the entire contents of the file have to be passed as part of resource creation. As this API is intended as somewhat of an advanced scenario, the design errs on the side of allowing granular control of individual file and folder permissions and uses a tree structure to represent the created (or updated) file system.

It's important to ensure the created file system (including file contents) is idempotent across runs as the specific configuration will contribute to the lifecycle hash used to identify changes in `ContainerLife.Persistent` containers.

A (simple) example of usage from the `WithPgAdmin` resource builder extension (this replaces the previous behavior that depended on creating and bind mounting temporary files):

```C#
pgAdminContainerBuilder.WithContainerFiles(
    destinationPath: "/pgadmin4",
    callback: (context, _) =>
    {
        var appModel = context.ServiceProvider.GetRequiredService<DistributedApplicationModel>();
        var postgresInstances = builder.ApplicationBuilder.Resources.OfType<PostgresServerResource>();

        return Task.FromResult<IEnumerable<ContainerFileSystemItem>>([
            new ContainerFile
            {
                Name = "servers.json",
                Contents = WritePgAdminServerJson(postgresInstances),
            },
        ]);
    });
```

An example of the non-callback API:
```C#
pgAdminContainerBuilder.WithContainerFiles(
    destinationPath: "/tmp",
    entries: [
        new ContainerFile
        {
            Name = "helloworld.sh",
            Contents = "echo 'hello world'",
            Mode = UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.UserExecute,
        },
    ]);
```

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No - While we could consider adding documentation, this API is intended as an advanced scenario for resource builders rather than developers consuming said resources in their App Host.


src/Aspire.Hosting/Dcp/Model/Container.cs
```diff
@@ -283,6 +287,146 @@ internal sealed class ContainerPortSpec
...
+    {
+        return HashCode.Combine(obj.Destination, obj.DefaultOwner, obj.DefaultGroup, obj.Mode);
+    }
+}
+
+internal sealed class ContainerFileSystemEntryComparer : IEqualityComparer<ContainerFileSystemEntry>
+{
+    public bool Equals(ContainerFileSystemEntry? x, ContainerFileSystemEntry? y)
+    {
+        if (x is null || y is null)
```
karolz-ms, line 403: Is .NET convention to return false if both operands are null?

karolz-ms (approved):

src/Aspire.Hosting/Dcp/Model/Container.cs
sebastienros: yes

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -162,39 +161,21 @@ public static IResourceBuilder<T> WithPgAdmin<T>(this IResourceBuilder<T> builde
...
+                // Generate the contents of the servers.json file
+                var serversConfigFile = WritePgAdminServerJson(postgresInstances);
 
-                    pgAdminContainerBuilder.WithBindMount(configJsonPath, "/pgadmin4/servers.json");
-                }
-                finally
-                {
-                    try
+                // Create a servers.json file in the container with necessary permissions
+                pgAdminContainerBuilder.WithCreateFile("/pgadmin4", new()
```
mitchdenny, line 170: `WithCreateFile` seems like an odd name for this API given what it is really doing is `docker cp`.

mitchdenny:
Its good that we are getting this capability. I have a few thoughts/questions:

1. Should we have the file system types or just have annotations that capture all the details directly?
2. What happens with file content in terms of clean up. Is it temporarily put on disk and then removed?
3. The `WithCreateFile` API seems like a weird name to me. I'd probably do something like `WithContainerFile` or something.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
sebastienros: I would say that `ContainerCreateFileAnnotation` is an implementation detail and here we care about providing a custom file.

But we could also say that `docker cp` is an implementation detail and from a user's perspective we want to create a file in the container.

Personally, I may go with `WithContainerFile(name, callback)` this way it hides both implementation details ;)

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
```diff
@@ -699,6 +699,76 @@ public static IResourceBuilder<T> WithBuildSecret<T>(this IResourceBuilder<T> bu
...
+    ///                     Mode = UnixFileMode.UserExecute | UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.GroupRead | UnixFileMode.GroupWrite,
+    ///                 },
+    ///             },
+    ///         },
+    ///     },
+    ///     defaultOwner: 1000,
+    ///     defaultMode: UnixFileMode.UserRead | UnixFileMode.UserWrite | UnixFileMode.GroupRead | UnixFileMode.GroupWrite);
+    /// </code>
+    /// </example>
+    public static IResourceBuilder<T> WithCreateFile<T>(this IResourceBuilder<T> builder, string destinationPath, List<ContainerFileSystemItem> entries, int defaultOwner = 0, int defaultGroup = 0, UnixFileMode defaultMode = UnixFileMode.None) where T : ContainerResource
```
mitchdenny, line 747: Yeah, so I've been thinking about this some more. I think what we need to do with this API is make it take an async callback (maybe we can have some simple wrappers) which is evaluated just before the resource is created.

This would allow for very late configuration scenarios where you might want to grab some state from the app model to generate file content.

So I am thinking something like:

```csharp
public static IResourceBuilder<T> WithContainerFiles(this IResourceBuilder<T> builder, Func<ContainerFilesContext, Task> callback);
```

The usage might look something like this:

```csharp
var c = builder.AddContainer(...)
               .WithContainerFiles(async context => {
                  _ = context.Model; // Access to app model.
                  _ = context.ServiceProvider = // Access to DI container.
               });
```

This would allow us to do things like interrogate other resources that might have already started to download configurations from them and then inject them into another container.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -296,54 +277,19 @@ public static IResourceBuilder<PostgresServerResource> WithPgWeb(this IResourceB
...
-                {
-                    try
-                    {
-                        Directory.Delete(tempDir, true);
-                    }
-                    catch
-                    {
-                    }
-                }
+                        Name = "bookmarks",
```
sebastienros: In `DirectoryInfo` it's named `path`.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
mitchdenny: You would end up with multiple container files annotations on a resource, and the callbacks would be invoked in order on the same context. The context itself would have methods that allow you to attach files/directories.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -162,39 +161,21 @@ public static IResourceBuilder<T> WithPgAdmin<T>(this IResourceBuilder<T> builde
...
                     {
-                        File.Delete(tempConfigFile);
-                    }
-                    catch
-                    {
-                    }
-                }
+                        new ContainerFile
+                        {
+                            Name = "servers.json",
```
sebastienros, line 174: In `FileInfo` it's named `FileName`

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: I like `WithContainerFile(s)`.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -296,54 +277,19 @@ public static IResourceBuilder<PostgresServerResource> WithPgWeb(this IResourceB
...
-                    {
-                        Directory.Delete(tempDir, true);
-                    }
-                    catch
-                    {
-                    }
-                }
+                        Name = "bookmarks",
+                        Mode = FileMode644,
+                        Entries = bookmarkFiles,
```
sebastienros: It feels weird to pass files and a folder. I would have expected to pass file with relative paths instead. And a directory entry to just be the entry and its permissions.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -440,22 +383,20 @@ private static string WritePgWebBookmarks(IEnumerable<PostgresDatabaseResource>
...
                     """;
 
-            hash.Append(Encoding.UTF8.GetBytes(fileContent));
-
-            File.WriteAllText(Path.Combine(dir, $"{postgresDatabase.Name}.toml"), fileContent);
+            bookmarkFiles.Add(new ContainerFile
+            {
+                Name = $"{postgresDatabase.Name}.toml",
+                Contents = fileContent,
+                Mode = FileMode755,
```
sebastienros: Only the folder would need 755, right? As it needs to be enumerated, but maybe I lack knowledge and the files too.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
mitchdenny: You could also have a simplified version of the method with the following signature:

```csharp
public static IResourceBuilder<T> WithContainerFiles(this IResourceBuilder<T> builder, params (string, int) files);
```

Usage would be as follows usage might be as follows:

```csharp
var c = builder.AddContainer(...)
               .WithContainerFiles([
                 ("blah.txt", 777),
                 ("foo.bin", 644)
               ]);
```

I'm less certain about this API and we could live without it as I think this API in general is a more advanced scenario.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
```diff
@@ -162,39 +161,21 @@ public static IResourceBuilder<T> WithPgAdmin<T>(this IResourceBuilder<T> builde
...
-                    // Deterministic file path for the configuration file based on its content
-                    var configJsonPath = aspireStore.GetFileNameWithContent($"{builder.Resource.Name}-servers.json", tempConfigFile);
-
-                    // Need to grant read access to the config file on unix like systems.
-                    if (!OperatingSystem.IsWindows())
-                    {
-                        File.SetUnixFileMode(configJsonPath, FileMode644);
-                    }
+                // Generate the contents of the servers.json file
+                var serversConfigFile = WritePgAdminServerJson(postgresInstances);
```
sebastienros, line 167: Variable name is confusing, now it's a file content, not a "File".
We can assume these files are fine in memory, but I kind of used a physical file on purpose before.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
sebastienros, line 747: `List<ContainerFileSystemItem>` is too constraining, I believe we would prefer IEnumerable in this case.

src/Aspire.Hosting/ApplicationModel/ContainerCreateFileAnnotation.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// The default permissions for files/directories to be created or updated in the container. 0 will be treated as 0600.
+    /// </summary>
+    public UnixFileMode DefaultMode { get; init; }
+
+    /// <summary>
+    /// The list of file system entries to create in the container.
+    /// </summary>
+    public List<ContainerFileSystemItem> Entries { get; init; } = new();
```
sebastienros, line 102: If Copilot told you that, it's because it's not up to date

```suggestion
    public List<ContainerFileSystemItem> Entries { get; init; } = [];
```

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
danegsta: > 1. Should we have the file system types or just have annotations that capture all the details directly?

I thought about having nested annotations, but didn't see that we'd used that pattern anywhere else, so was a bit hesitant to introduce it here. 
> 2. What happens with file content in terms of clean up. Is it temporarily put on disk and then removed?

We build a tar file in memory and then stream that to stdin for the `docker cp` command, so there's no temporary files to cleanup (technically `podman cp` does create and cleanup a temporary file themselves, but I want to create a PR against them to remove that behavior since the temp file they create is never actually used).
> 3. The WithCreateFile API seems like a weird name to me. I'd probably do something like WithContainerFile or something.

`WithContainerFiles` seems like a solid choice. The destination path may seem a bit odd, but it is technically necessary as you have to give a base path in the container you want to add or modify your files/folders and any folders you add/update will have their ownership and permissions updated if they already exist (whereas the destination path won't be modified). With a callback model, we'd likely need to support nested callbacks for directories and files:

```C#
.WithContainerFiles("/", async context =>
{
    context.AddDirectory(".pgweb" d =>
    {
        d.AddDirectory("bookmarks", d =>
        {
            d.AddFile("somefile.toml", "contents", mode: FileMode644);
        });
    });
});
```

sebastienros: I think we need to be able to pass streams if not a file path (even/too). Because files might be big enough that we don't want to load them in memory at once.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: Correct, I had these reversed.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: From a `docker cp` standpoint, it isn't really feasible to do relative paths as we need to apply permissions in the tar file on a folder by folder and file by file basis, otherwise base directory permissions get a bit weird. Basically if we want to have more granular control, we have to treat it as a full file system tree.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: This API is specifically intended to replace using physical files in situations where we need to control the ownership and permissions. `docker cp` specifically doesn't create any kind of link back to the host filesystem so bind mounts are still appropriate when wanting to work with physical files from the host filesystem.

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
danegsta: Main reason I made it a List was to enable being lazy and instantiating the argument with new(), but that can easily be changed.

danegsta: Trying to stream in files or pass in filesystem is theoretically possible, but adds substantial complexity vs. ephemeral files. The current API is really intended as a targeted replacement for temporary files intended to be bind mounted into a container as configuration. For situations with larger files or assets, bind mounts offer better performance, reliability, and persistence than we'd likely get out of a `docker cp` based solution for existing files.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: DirectoryInfo and FileInfo both seem to have a Name property that corresponds in purpose to this Name property.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
davidfowl: Me too!

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
danegsta: @mitchdenny I’ve implemented a callback version of the API and annotation, but need to update the functional tests based on the change.

danegsta: Tests and API are updated.

src/Aspire.Hosting.PostgreSQL/PostgresBuilderExtensions.cs
danegsta: One of the key things I think this API brings is the direct manipulation of the container file system prior to startup, which opens up a lot of advanced scenarios, but necessitates the more complex tree focused design vs. a simple file/path model.

mitchdenny (approved):

src/Aspire.Hosting/ContainerResourceBuilderExtensions.cs
danegsta: Chatted with @mitchdenny offline; I'm updating the default permission behavior to match that of *nix systems (umask based default permissions).

davidfowl: Is the API in the description the only APi? It’s super verbose. Where’s the easy mode APi?

danegsta: There's a non-callback version that's a little simplified, but most of the complexity comes from the fact that the API is letting us do direct file system manipulation in the container (and that manipulation has to happen at a specific existing path in the container). We're manipulating a file tree, so we ended up with a tree structure in the API.

If we did add a simpler API method, it'd be a bit weird in any slightly complicated scenario (here's an example based on the pgweb bookmark file creation):

```C#
// Creates new bookmark files under the newly created /.pgweb/bookmarks/
WithContainerFiles(
    // We're creating folders and files relative to the root of the filesystem (/)
    destinationPath: "/",
    files: [
        // We need to specify the full path for each file since the .pgweb/bookmarks/ portion of the path is newly created
        (".pgweb/bookmarks/someserver.toml", "contents"),
        (".pgweb/bookmarks/otherserver.toml", "contents"),
    ]);
```

PR #8136, opened 2025-03-18
Making `aspire run` pretty

mitchdenny:
This PR does a few things:

1. When you execute `aspire run` without any special arguments it will output the dashboard login URL followed by a table of resources and their types and endpoints.

2. It extends the `--wait-for-debugger` switch so that when it launches the apphost it adds an environment variable which triggers the apphost to also pause for the debugger to be attached.

3. Suppresses all other output from the console unless you are using `--debug`. When `--debug` is used pretty console output will not be generated and instead you'll get the combined log output of the CLI and the apphost that is being invoked.

https://github.com/user-attachments/assets/655a9b11-aec9-4d53-b5c3-d260fcf2cf8f


davidfowl:
Show the health status! This looks amazing!

maddymontaquila:
I made it prettier 🤣


https://github.com/user-attachments/assets/03fecfee-8331-4814-9e19-e919174fa5be

davidfowl (approved):

PR #8138, opened 2025-03-18
Update to xunit.v3

Youssef1313:


Youssef1313:
Hi @radical @eerhardt 
Can I have a review here please? Thanks!

tests/Shared/RepoTesting/Aspire.RepoTesting.props
```diff
@@ -2,6 +2,10 @@
   <PropertyGroup Condition="'$(TestsRunningOutsideOfRepo)' == 'true'">
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+
+    <!-- Arcade doesn't support such test runner yet. -->
```
danmoseley, line 6: is there an arcade issue for this? (and note in there to remove this then)

danmoseley (approved):
thanks.

danmoseley:
@eerhardt any concerns about merging this? 

looking at https://xunit.net/docs/getting-started/v3/whats-new there are many new features but nothing pops that we need immediately

tests/Shared/RepoTesting/Aspire.RepoTesting.props
Youssef1313: I just opened an issue.

```suggestion
    <!-- Arcade doesn't support such test runner yet. -->
    <!-- https://github.com/dotnet/arcade/issues/15654 -->
```

Youssef1313:
> @eerhardt any concerns about merging this?
> 
> looking at https://xunit.net/docs/getting-started/v3/whats-new there are many new features but nothing pops that we need immediately

The support for Microsoft.Testing.Platform could be something (I'm planning to do it after this PR is merged)

eerhardt:
I'm confused at the goal here. Is the goal for all our tests in this repo to use xunit v3? If so, this PR doesn't appear to do that. I opened our tests with this branch and I'm seeing:

image

Youssef1313:
Hmm.. That was unexpected. Let me check what's going on 😕

radical:
Those xunit versions and references are likely coming from Arcade.

radical:
Does `ActiveIssue` not work anymore? Is that because Discoverers are not supported?

We should check out how this behaves with helix, and azdo too. It might also be a good idea to see what Arcade has to say about moving to v3, since we are still using that.

As for useful features, the `dynamic test skipping` could be good, but not a must-have.

Youssef1313:
I'll get to ActiveIssue. It should be possible to work.

danmoseley:
@Youssef1313 thanks for the work here, but given the isze of the changes (post my signoff) I think we should discuss in an issue first whether we want this change. I will close this, feel free to open an issue proposing xunit v3, explaining the advantage, anything we lose or gain. If we go ahead, we'll want to be confident we didn't lose any coverage, including test counts  before and after, and also time taken to run tests before and after to check perf.

PR #8140, opened 2025-03-18
Support CosmosDB and Redis role assignments on app specific managed identity

### LINKED ISSUE #8142: AsExisting to a Redis resource doesn't create a unique access policy name
When trying to use an existing Redis resource that was created by another Aspire app, I'm getting an error during deployment:

```
ERROR: error executing step command 'provision': deployment failed: error deploying infrastructure: deploying to subscription:

Deployment Error Details:
InvalidAccessPolicyAssignment: An invalid access policy assignment was detected due to duplicate entries. Please review and ensure that each assignment has a unique combination of assignmentName, policyName, and objectId to avoid such duplicates. For more information, please refer to the following link: https://aka.ms/redis/MicrosoftEntraAuthenticationPrerequisites.
RequestID=6caf2145-4221-4133-83de-108cb45a447f

TraceID: 34acc9d60e6a16bca27d9cf689f07607
```

The error message is a little misleading (the objectId is different between the 2 deployments), but the issue is that we aren't creating a unique name for the role assignment:

https://github.com/dotnet/aspire/blob/60acd6718739bacaf9e6d12ae2f2441884204274/src/Aspire.Hosting.Azure.Redis/AzureRedisExtensions.cs#L264-L270

Since we aren't setting the Name property, it is getting defaulted:

https://github.com/dotnet/aspire/blob/624e7a4e73231eb7389704d958e5244bd4b4c7fe/playground/bicep/BicepSample.AppHost/redis.module.bicep#L29-L30

When this runs inside the same resource group (since it needs to run in the existing Redis's resource group), the same `name` is being generated, causing the duplicate error.

To fix this, we should generate a unique name based on principalId, like we do for other role assignments:

https://github.com/dotnet/aspire/blob/624e7a4e73231eb7389704d958e5244bd4b4c7fe/playground/bicep/BicepSample.AppHost/sb.module.bicep#L25

## Repro

Deploy Redis resource via an Aspire app.

In another Aspire app, try to attach to that Redis resource as existing:
```C#
var redis = builder.AddAzureRedis("redis")
    .PublishAsExisting("redis-name", "rg-name-redis");
```

cc @captainsafia @sebastienros

eerhardt:
## Description

We currently don't support an easy API for modifying the role assignments on database resources. There are no role types that can be used from the Azure.Provisioning libraries today. So for now we will just move our existing role assignments to the ACA roles bicep module.

To support database resources, we add a new virtual method on AzureProvisioningResource - `AddRoleAssignments` passing in a context object containing the information on the role assignments to add.

Also needed to split all role assignments out of the {app}-roles bicep module because of an issue with CosmosDB failing if the role assignment is added in the same bicep file as the managed identity it is pointing to. This simplifies the code because the role assignment generation is the same whether the resource is existing or not.

Contributes to #6161
Fix #8142

## Checklist

- Is this feature complete?
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] Yes
      - Ongoing
  - [x] Yes
        - Link to aspire-docs issue [[New article]: Aspire 9.2 Azure WithRoleAssignments docs (dotnet/docs-aspire#2788)](https://github.com/dotnet/docs-aspire/issues/2788)


playground/Redis/Redis.ApiService/Program.cs
```diff
@@ -1,14 +1,31 @@
...
+
+var configurationOptions = ConfigurationOptions.Parse(
+    builder.Configuration.GetConnectionString("redis") ??
+    throw new InvalidOperationException("Could not find a 'redis' connection string."));
+
+if (configurationOptions.EndPoints.Any(azureOptionsProvider.IsMatch))
+{
+    await configurationOptions.ConfigureForAzureWithTokenCredentialAsync(
+        new DefaultAzureCredential());
+}
```
davidfowl, line 22: 🤮

eerhardt: LOL, I'm reverting this change anyway.

captainsafia (approved):
This looks good given the gnarliness you had to work around with the way role assignments are managed for these resource types.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -314,96 +316,110 @@ private void ProcessRoleAssignments()
...
+                }
+
+                if (RoleAssignmentCustomizations.TryGetValue(azureResource, out var customizations))
+                {
+                    foreach (var customization in customizations)
+                    {
+                        customization(infra, azureResource);
+                    }
+
+                    // hook up any known parameters
```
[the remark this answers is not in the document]
eerhardt: I was able to refactor this so we don't need this hookup anymore. Instead, when the property is referenced, the parameter will get allocated.

tests/Aspire.Hosting.Azure.Tests/AzureRedisExtensionsTests.cs
```diff
@@ -1,23 +1,46 @@
...
-    public async Task AddAzureRedis()
+    /// <summary>
+    /// Test both with and without ACA infrastructure because the role assignments
+    /// are handled differently between the two. This ensures that the bicep is generated
+    /// consistently regardless of the infrastructure used in RunMode.
+    /// </summary>
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task AddAzureRedis(bool useAcaInfrastructure)
```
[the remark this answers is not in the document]
eerhardt: Added.

tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs
```diff
@@ -271,17 +271,18 @@ public async Task AddAzureFunctionsProject_WorksWithAddAzureContainerAppsInfrast
 
         await ExecuteBeforeStartHooksAsync(app, default);
 
-        var projRoles = Assert.Single(model.Resources.OfType<AzureProvisioningResource>().Where(r => r.Name == $"funcapp-roles"));
```
captainsafia, line 274: Should we validate that the funcapp-identity resource emits the right manifest/bicep here? Particularly around the principal ID being an output that's consumed as a parameter elsewhere?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -36,7 +36,8 @@ public static IResourceBuilder<AzureCosmosDBResource> AddAzureCosmosDB(this IDis
         builder.AddAzureProvisioning();
 
         var resource = new AzureCosmosDBResource(name, ConfigureCosmosDBInfrastructure);
-        return builder.AddResource(resource);
+        return builder.AddResource(resource)
+            .WithAnnotation(new DefaultRoleAssignmentsAnnotation(new HashSet<RoleDefinition>()));
```
captainsafia, line 40: What does an empty `DefaultRoleAssignmentsAnnotation` actually help us achieve here? This is just here to pass through to `AppliedRoleAssignmentsAnnotation` in the non-ACA scenario, right?

captainsafia (approved):
LGTM! Reads much easier than the annotations-based implementation. 😅

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -322,101 +321,69 @@ private void ProcessRoleAssignments()
...
+                if (RoleAssignments.TryGetValue(azureResource, out var roles))
                 {
-                    var roleAssignment = CreateRoleAssignment(prefix, provisionable, role.Id, role.Name, appIdentityId, RoleManagementPrincipalType.ServicePrincipal, appIdentityPrincipalId);
-                    infra.Add(roleAssignment);
+                    var context = new AddRoleAssignmentsContext(
+                        infra,
+                        roles,
+                        new(() => RoleManagementPrincipalType.ServicePrincipal),
+                        new(() => new BicepOutputReference("principalId", containerAppIdentityResource).AsProvisioningParameter(infra, parameterName: AzureBicepResource.KnownParameters.PrincipalId)),
+                        new(() => new BicepOutputReference("principalName", containerAppIdentityResource).AsProvisioningParameter(infra, parameterName: AzureBicepResource.KnownParameters.PrincipalName)));
```
davidfowl, line 381: Using Lazy in the public API is hella ugly. Would be cleaner if there was an `IRoleAssignmentsContext` and AddRoleAssignmentsContext as internal.

eerhardt:
Merging now because this change and https://github.com/dotnet/aspire/pull/8171 will conflict. So merging will help avoid conflicts now.

I will address the above feedback in a follow up PR today.

tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs
eerhardt: I don't think that is necessary everywhere. We validate that in other places - specifically in the AzureContainerAppsTests.

https://github.com/dotnet/aspire/blob/122579052d497681ba156bb23058a70434eb1dd7/tests/Aspire.Hosting.Azure.Tests/AzureContainerAppsTests.cs#L2427-L2464

Feels like overkill to spread that all over.

Note that this test does verify that the principalId comes from the identity resource.

https://github.com/dotnet/aspire/blob/122579052d497681ba156bb23058a70434eb1dd7/tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs#L286

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -322,101 +321,69 @@ private void ProcessRoleAssignments()
...
             }
 
-            // existing resources need to be referenced in separate modules so the Scope can be set correctly
-            private List<AzureBicepResource> CreateExistingResourceRoleAssignments(AzureProvisioningOptions provisioningOptions, AzureProvisioningResource containerAppIdentityResource)
+            private List<AzureBicepResource> CreateRoleAssignmentsResources(AzureProvisioningOptions provisioningOptions, AzureProvisioningResource containerAppIdentityResource)
             {
-                var existingResourceRoleAssignments = new List<AzureBicepResource>();
-                foreach (var (a, roles) in GetExistingResourceRoles())
+                var roleAssignments = new List<AzureBicepResource>();
+                foreach (var a in RoleAssignments.Keys)
```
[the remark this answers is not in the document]
eerhardt: Fixed in [Follow up on database role assignments (dotnet/aspire#8212)](https://github.com/dotnet/aspire/pull/8212).

src/Aspire.Hosting.Azure/AddRoleAssignmentsContext.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    public BicepValue<RoleManagementPrincipalType> PrincipalType => _getPrincipalType.Value;
+
+    /// <summary>
+    /// Gets the principal ID to use in the role assignment.
+    /// </summary>
+    public BicepValue<Guid> PrincipalId => _getPrincipalId.Value;
+
+    /// <summary>
+    /// Gets the principal name to use in the role assignment.
```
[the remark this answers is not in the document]
eerhardt: Added in [Follow up on database role assignments (dotnet/aspire#8212)](https://github.com/dotnet/aspire/pull/8212).

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
eerhardt: Fixed in [Follow up on database role assignments (dotnet/aspire#8212)](https://github.com/dotnet/aspire/pull/8212)

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
captainsafia: > In ACA scenarios, the AzureResourcePreparer will copy the DefaultRoleAssignmentsAnnotation to a regular (empty) RoleAssignmentAnnotation, which will cause the role to be written to the role bicep file.

Ah, I [see where this happens](https://source.dot.net/#Aspire.Hosting.Azure/AzureResourcePreparer.cs,153).

> Thoughts?

IMO, an empty one feels better than a fake role assignment. We clarified in an offline discussion why we need `DefaultRoleAssignments` vs `RoleAssignment` here.

PR #8158, opened 2025-03-19
Rename Workload tests to Template tests

radical:
Aspire does not use a workload anymore, and the `Aspire.Workload.Tests` are essentially just template tests now. This renames the project, and any references to the workload in the code, and in the build infrastructure.

Also, we drop the unused package reference to `Microsoft.NET.Runtime.WorkloadTesting.Internal` which is no longer used as we don't install a workload for testing.


radical:
Validation azdo build -https://dev.azure.com/dnceng-public/public/_build/results?buildId=986547&view=results

tests/Shared/Aspire.Templates.Testing.props
```diff
@@ -0,0 +1,3 @@
+<Project>
+  <Import Project="$(RepoRoot)src\Shared\Workload.props" />
```
eerhardt, line 2: Why is this still named "Workload"?

tests/Aspire.EndToEnd.Tests/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- This triggers the import of targets needed for workload testing -->
```
eerhardt, line 3:
```suggestion
    <!-- This triggers the import of targets needed for template testing -->
```

eerhardt (approved):

tests/Shared/Aspire.Templates.Testing.props
radical: This seemed related to the actual workload, but I'm not sure. @joperezr Any idea about this?

PR #8182, opened 2025-03-19
Remove role assignments on implicit storage account

### LINKED ISSUE #8143: Functions project provisioning storage even when `WithHostStorage` is set in Aspire 9.2

### Describe the bug

Using Aspire 9.2 bits, I'm seeing an additional storage account being provisioned as the default host storage, even though I have called `WithHostStorage`, passing in a customized storage account. 

From an internal discussion, this may be related to my adding `builder.AddAzureContainerAppsInfrastructure()`, which was required due to an initial error upon upgrade to 9.2:

> The application model does not support role assignments. Ensure you are using a publisher that supports role assignments, for example AddAzureContainerAppsInfrastructure.

### Expected Behavior

Only the storage account I've specified is provisioned. The deployment is not blocked by Policy in my case.

### Steps To Reproduce

1. Use VS to scaffold a new .NET 9 Functions project enlisted in Aspire, just using an HTTP trigger
2. Update packages to Aspire 9.2 versions
3. modify the app host to the following:

```csharp

builder.AddAzureContainerAppsInfrastructure();

var storage = builder.AddAzureStorage("mystorage").RunAsEmulator();

 storage.ConfigureInfrastructure<AzureStorageResource>(infra =>
{
    var sa = infra.GetProvisionableResources().OfType<StorageAccount>().FirstOrDefault(r => r.BicepIdentifier == infra.AspireResource.GetBicepIdentifier());
    sa.AllowBlobPublicAccess = false;
});

builder.AddAzureFunctionsProject<Projects.FunctionAppAspire92>("functionappaspire92")
    .WithHostStorage(storage)
    .WithRoleAssignments(storage, StorageBuiltInRole.StorageBlobDataOwner);

builder.Build().Run();
```

4. Attempt to Deploy to ACA.

My situation leads to an error due to Azure Policy, but if that consideration were removed, it seems that you would have two storage accounts when only one was desired.

### Exceptions (if any)

> Error deployment failed: error deploying infrastructure: deploying to subscription:
> 
> Deployment Error Details:
> InvalidTemplateDeployment: The template deployment failed because of policy violation. Please see details for more information.
> RequestDisallowedByPolicy: Resource 'funcstorage80823lre5eq5f' was disallowed by policy. Policy identifiers: [REDACTED]

The policy in question is set on my subscription to disallow blob public access, That's why I have the `ConfigureInfrastructure` call in the example. I'm attempting to ensure the storage is compliant and can be deployed. The system will successfully provision a separate storage account for "mystorage", but this error is referring to an additional storage account I would not expect to be provisioned.

### .NET Version info

SDK 9.0.200

Runtime 9.0.3

### Anything else?

Aspire package version: **9.2.0-preview.1.25168.1** (2025-03-18)

captainsafia:
Closes https://github.com/dotnet/aspire/issues/8143.

Prior to this change, we still maintained a `RoleAssignmentAnnotation` for the implicit host storage resource even when `WithHostStorage` was called to set it explicitly. As a result of this, the `AzureResourcePreparer` would emit role assignments against this stale resource pointer. This declaration would be emitted into the manifest and result in a Bicep module for the stale implicit host storage still being created.

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -59,6 +59,17 @@ public static class AzureFunctionsProjectResourceExtensions
                 {
                     removeStorage = false;
                 }
+                else
+                {
+                    // Remove the role assignments associated with the implicit host storage resource
+                    // if an explicit one has been provided. Assume that users must configure their
+                    // own role assignments for the explicit storage resource.
+                    if (item.TryGetAnnotationsOfType<RoleAssignmentAnnotation>(out var roleAssignments) &&
+                        roleAssignments.SingleOrDefault(assignment => assignment.Target == storage) is { } roleAssignmentAnnotation)
+                    {
+                        item.Annotations.Remove(roleAssignmentAnnotation);
+                    }
+                }
             }
 
             if (removeStorage)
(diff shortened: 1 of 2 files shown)


eerhardt (approved):
LGTM. Thanks!

PR #8184, opened 2025-03-19
Add a pause button to console logs, structured logs, traces, and metrics pages

### LINKED ISSUE #7610: Pause/resume button on console logs, structured logs, traces and metrics

9.1 added remove buttons to these pages. The feature turns out to be pretty low cost. Compared to the effort involved, I think people will love it.

On the same theme, what people will love just as much are pause buttons to stop updating the UI while they look at logs/telemetry.

We add pause buttons to these pages. It toggles between pause and play. It sits to the left of the remove button.

There are two ways we can do pause:

1. Pause receiving data. Data received while the page is paused is discarded. Clicking resume starts receiving data again and updating the UI.
2. Or pause the UI. When pause is clicked we make a note of the time, and any data after that time is filtered out in the UI. Clicking resume removes the time filter and the UI updates to display data that was received during the pause.

Also with both approaches we'd need to update the metrics page graphs to stop updating the X asis with the current time)

Each approach has its pros and cons. I don't think either is terribly difficult to implement.

### Additional context

_No response_

adamint:
## Description

Adds a pause/resume button to those 4 pages to the left of the clear signals button.
image

For console logs, an entry appears when a pause is active noting how many logs have been filtered out for that resource. 
image

If, after a pause has ended, one or more logs were filtered, text appears in the console log noting the dates of the pause and how many logs were filtered. These lines are filtered out when downloading logs. Additionally, line numbers increment by the # of filtered logs, as can be seen in the screenshot below.
image

For metrics, the X axis of a visible graph stays consistent during a pause. 

Fixes #7610

## Checklist

- Is this feature complete?
  - [X] Yes. Ready to ship.
  - [X] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [X] No


src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor
```diff
@@ -53,14 +53,3 @@ else
         </FluentTabs>
     </div>
 }
-
-@code {
```
adamint, line 57: moved to code behind to match the rest of the parameters

JamesNK:
There is a design decision that needs to be made here that I mentioned on the issue: whether the pause happens in the UI, or collecting telemetry. That needs to be figured out first.

src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
```diff
@@ -79,7 +91,7 @@ private async Task UpdateDataAsync()
         while (await timer!.WaitForNextTickAsync())
         {
             _instrument = GetInstrument();
-            if (_instrument == null)
+            if (_instrument == null || PauseManager.MetricsPaused)
```
JamesNK, line 94: This stops the graph from updating while you are on the page, but if you navigate away and back again then the time updates.

I think the pause manager needs to track when metrics was paused. Then when someone navigates back to the page then that time is used as the "current time" when rendering the graph.

The time scale on this page is "The last 15 minutes" (for example). When paused that time should refer to the last 15 minutes since the pause time.

src/Aspire.Dashboard/Components/Controls/LogViewer.razor.cs
```diff
@@ -36,6 +36,9 @@ public sealed partial class LogViewer
     [Parameter]
     public bool IsTimestampUtc { get; set; }
 
+    [Parameter]
+    public DateTime? PausedAt { get; set; }
```
JamesNK, line 40: Not used anymore?

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -480,11 +483,15 @@ private void LoadLogs(ConsoleLogsSubscription newConsoleLogsSubscription)
                         }
 
                         var logEntry = logParser.CreateLogEntry(content, isErrorOutput);
-                        if (timestampFilterDate is null || logEntry.Timestamp is null || logEntry.Timestamp > timestampFilterDate)
+                        if (logEntry.Timestamp is not null &&
+                            ((timestampFilterDate is not null && !(logEntry.Timestamp > timestampFilterDate))
+                            || PauseManager.ConsoleLogsPausedRanges.Any(range => range.IsOverlapping(logEntry.Timestamp.Value))))
```
JamesNK, line 488: Can you rewrite this check to use a method instead of LINQ. Having N+1 LINQ queries might have some perf impact.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -600,6 +607,21 @@ private async Task ClearConsoleLogs(ApplicationKey? key)
...
+    {
+        PauseManager.ConsoleLogsPaused = isPaused;
+        if (isPaused)
+        {
+            PauseManager.ConsoleLogsPausedRanges.Add(new DateTimeRange(start: DateTime.UtcNow, end: null));
+        }
+        else
+        {
+            PauseManager.ConsoleLogsPausedRanges.Last().End = DateTime.UtcNow;
+        }
```
JamesNK, line 620: Hmmm, I worry about thread safety of `ConsoleLogsPausedRanges`

How about making ConsoleLogsPaused have a private setter, and then add a `SetConsoleLogsPaused(bool isPaused)` method. Inside that method you would update the ranges as needed. Just encapsulates the code logic a little more.

Also, the method I mentioned [here](https://github.com/dotnet/aspire/pull/8184/files#r2006604953) could be on the PauseManager and the pause ranges collection could be made internal.

Instead of updating the list with a new value, I think a new collection should be created when a pause starts. That prevents any problem of the collection being iterated over while it is changed.

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
```diff
@@ -271,6 +274,11 @@ private void RaiseSubscriptionChanged(List<Subscription> subscriptions)
 
     public void AddLogs(AddContext context, RepeatedField<ResourceLogs> resourceLogs)
     {
+        if (_pauseManager.StructuredLogsPaused)
+        {
+            return;
```
JamesNK, line 279: Write a trace level message to the logs that structured logs were ignored because of the pause.

Repeat for traces and metrics

JamesNK, line 279: Could you add a unit test to the telemetry repository tests that tests pausing? I think just one test where everything is paused, and then you attempt to write the 3 kinds of telemetry, and then check that there is no new data.

src/Aspire.Dashboard/Components/Controls/PauseResumeButton.razor
```diff
@@ -0,0 +1,16 @@
+﻿@using Aspire.Dashboard.Resources
+@inject IStringLocalizer<ControlsStrings> Loc
+
+<FluentButton Appearance="Appearance.Lightweight"
+              Title="@(IsPaused ? Loc[nameof(ControlsStrings.ResumeButtonTitle)] : Loc[nameof(ControlsStrings.PauseButtonTitle)])"
```
JamesNK, line 5: I like the tooltip. I noticed the remove button doesn't have a tooltip. Could you update it to have one ("Remove data") while you're here?

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
```diff
@@ -29,6 +29,10 @@
 
             <ClearSignalsButton SelectedResource="PageViewModel.SelectedOption" HandleClearSignal="ClearConsoleLogs" />
 
+            <PauseResumeButton IsPaused="@PauseManager.ConsoleLogsPaused"
+                               IsPausedChanged="@IsPausedChangedAsync"
+                               Disabled="@(PageViewModel.SelectedResource is null)" />
```
JamesNK, line 34: Should the button be disabled? You're not viewing data yet, but I don't think that means you shouldn't be able to pause/resume. You can still click "Remove all" if not viewing anything:

image

src/Aspire.Dashboard/Components/Pages/Metrics.razor
```diff
@@ -51,6 +51,10 @@
             <ClearSignalsButton SelectedResource="PageViewModel.SelectedApplication"
                                 HandleClearSignal="ClearMetrics" />
 
+            <PauseResumeButton
+                @bind-IsPaused="@PauseManager.MetricsPaused"
+                Disabled="@(PageViewModel.SelectedInstrument is null)" />
```
JamesNK, line 56: Same as https://github.com/dotnet/aspire/pull/8184/files#r2006616340. Should it be disabled?

adamint: Re-enabled!

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
adamint: Re-enabled!

src/Aspire.Dashboard/Components/Controls/Chart/ChartContainer.razor.cs
adamint: I have updated the `PauseManager` and `ChartBase` to use the paused time as `inProgressDataTime` and `_currentDataStartTime` on initialization

src/Aspire.Dashboard/Components/Controls/LogViewer.razor
```diff
@@ -1,39 +1,71 @@
...
-                                    <span class="timestamp" title="@FormatHelpers.FormatDateTime(TimeProvider, timestamp, MillisecondsDisplay.Full, CultureInfo.CurrentCulture)">@GetDisplayTimestamp(timestamp)</span>
-                                }
-                                @if (context.Type == LogEntryType.Error)
-                                {
-                                    hasPrefix = true;
-                                    <fluent-badge appearance="accent">stderr</fluent-badge>
-                                }
-                                @((MarkupString)((hasPrefix ? "&#32;" : string.Empty) + (context.Content ?? string.Empty)))
+            <Virtualize Items="@logEntries.GetEntries()" ItemSize="20" OverscanCount="100" TItem="LogEntry">
+                @if (context is { Type: LogEntryType.Pause, Timestamp: { } startTimestamp} && PauseManager.TryGetConsoleLogPause(startTimestamp, out var pause))
```
adamint, line 17: Split out handling the pause so that a div is not added for a previous pause when there were no filtered logs for that app

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor
```diff
@@ -27,24 +27,30 @@
                             @bind-SelectedResource:after="HandleSelectedOptionChangedAsync"
                             LabelClass="toolbar-left" />
 
+            <PauseIncomingDataSwitch IsPaused="@PauseManager.ConsoleLogsPaused" IsPausedChanged="@IsPausedChangedAsync" />
             <ClearSignalsButton SelectedResource="PageViewModel.SelectedOption" HandleClearSignal="ClearConsoleLogs" />
 
-            @foreach (var command in _highlightedCommands)
+            @if (_highlightedCommands.Count > 0)
```
adamint, line 33: Nothing in this block actually changed other than adding a Commands label when there are highlighted commands to separate from the built-in signals/pause commands

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -474,15 +483,34 @@ private void LoadLogs(ConsoleLogsSubscription newConsoleLogsSubscription)
                     foreach (var (lineNumber, content, isErrorOutput) in batch)
                     {
                         // Set the base line number using the reported line number of the first log line.
-                        if (_logEntries.EntriesCount == 0)
+                        if (_logEntries.BaseLineNumber is null && !_logEntries.GetEntries().Any(e => e.Type is not LogEntryType.Pause))
```
adamint, line 486: uses linq but will only be running a couple times

JamesNK:
I left console logs paused for a while. Then I navigateed away from the page and came back:

image

There should have been some logs inbetween these pauses. But I believe there is a 10,000 log limit so they were filtered out.

However, the row number starts at 1. It should take into account previous logs, and logs filtered out.

JamesNK:
> <Log capture paused between 1:55:19.156 pm and 1:55:38.079 pm 190 log entries filtered out>

Could you add a comma between the date and count. The "currently" paused text has this.

> <Currently paused log capture, 9269 entries filtered out>

JamesNK:
> <Currently paused log capture, 9269 entries filtered out>

Add the start date to this. e.g.

> <Currently paused log capture starting at XXXXX, 9269 entries filtered out>

JamesNK:
I think we need to make it a bit more obvious that logs are paused on structured logs and traces. The Pause vs Unpause icon doesn't jump out of the screen. I imagine there will be at least some folks who pause, get confused that results aren't updating, then it takes then a while to figure out they early paused.

We have the footer which displays a count of matching results:

image

Next to it, in bold, red text, add "Currently paused structured logs capture starting at XXXXX".

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
```diff
@@ -804,6 +815,12 @@ public Dictionary<string, int> GetLogsFieldValues(string attributeName)
 
     public void AddMetrics(AddContext context, RepeatedField<ResourceMetrics> resourceMetrics)
     {
+        if (_pauseManager.AreMetricsPaused(out _))
+        {
+            _logger.LogTrace("{LogCount} incoming metric(s) was ignored because of an active pause.", resourceMetrics.Count);
```
JamesNK, line 820:
```suggestion
            _logger.LogTrace("{Count} incoming metric(s) was ignored because of an active pause.", resourceMetrics.Count);
```

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
```diff
@@ -271,6 +276,12 @@ private void RaiseSubscriptionChanged(List<Subscription> subscriptions)
 
     public void AddLogs(AddContext context, RepeatedField<ResourceLogs> resourceLogs)
     {
+        if (_pauseManager.StructuredLogsPaused)
+        {
+            _logger.LogTrace("{LogCount} incoming structured log(s) was ignored because of an active pause.", resourceLogs.Count);
```
JamesNK, line 281:
```suggestion
            _logger.LogTrace("{Count} incoming structured log(s) was ignored because of an active pause.", resourceLogs.Count);
```

src/Aspire.Dashboard/Otlp/Storage/TelemetryRepository.cs
```diff
@@ -826,6 +843,12 @@ public void AddMetrics(AddContext context, RepeatedField<ResourceMetrics> resour
 
     public void AddTraces(AddContext context, RepeatedField<ResourceSpans> resourceSpans)
     {
+        if (_pauseManager.TracesPaused)
+        {
+            _logger.LogTrace("{LogCount} incoming trace(s) was ignored because of an active pause.", resourceSpans.Count);
```
JamesNK, line 848:
```suggestion
            _logger.LogTrace("{Count} incoming trace(s) was ignored because of an active pause.", resourceSpans.Count);
```

JamesNK:
> There should have been some logs inbetween these pauses. But I believe there is a 10,000 log limit so they were filtered out. I think that limit is on the server and it would require extra API calls to do filtering there.
> 
> However, the row number starts at 1. It should take into account previous logs, and logs filtered out.

Sometimes the line number is updated to skip values, other times it isn't:

image

adamint:
> > There should have been some logs inbetween these pauses. But I believe there is a 10,000 log limit so they were filtered out. I think that limit is on the server and it would require extra API calls to do filtering there.
> > However, the row number starts at 1. It should take into account previous logs, and logs filtered out.
> 
> Sometimes the line number is updated to skip values, other times it isn't:
> 
> image

Hmm, I'm not able to reproduce this?

JamesNK:
Please try again and keep testing. There are definitely bugs here. Use the stress playground and the actions that add many logs.

Another example:
image

src/Aspire.Dashboard/Model/PauseManager.cs
```diff
@@ -0,0 +1,121 @@
...
+            if (range.IsOverlapping(entry.Timestamp.Value))
+            {
+                foundRange = range;
+                break;
+            }
+        }
+
+        if (foundRange is not null && foundRange.FilteredLogsByApplication.GetValueOrDefault(application)?.Contains(entry) is not true)
+        {
+            ImmutableInterlocked.Update(
```
JamesNK, line 74: Every filtered log is going to recreate a constantly growing immutable has set?

src/Aspire.Dashboard/Components/Controls/LogViewer.razor
```diff
@@ -1,39 +1,76 @@
...
```
JamesNK, line 17: Make the log pause a property on the log entry. Using the start date as a key to look it up from the manager is hacky.

src/Aspire.Dashboard/Model/PauseManager.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                return ranges.SetItem(lastRange.Start, lastRange with { End = timestamp });
+            }
+        });
+    }
+}
+
+public record ConsoleLogPause(DateTime Start, DateTime? End, ImmutableDictionary<string, ImmutableHashSet<LogEntry>> FilteredLogsByApplication)
+{
+    public int GetFilteredLogCount(string? application)
```
JamesNK, line 110: I don't think the count should be stored here. We're possibly storing an unbounded number of log entries as long as the user doesn't unpause. This could use up all memory easly.

This should simply store the start/end date. The count can be calcuated where it is used: on the console logs page

JamesNK:
I didn't take a close look at the console pausing logic until now. I think the way counts are calculated and stored needs to be rewritten. We can't hold onto all the log entries in memory until the user unpauses. All memory could be consumed.

JamesNK:
> LogEntry.CreatePause(pauseStart)

When a pause log entry is created it have some kind of pause view model, and it's associated with the range and has a count. When something is filtered out, the count on the pause view model is incremented.

Keep the logic needed for this page (the counts) on this page. Don't put it into global state.

JamesNK:
I think line number of pause log entries also need to be rethought.

* I think pause log entry line number should always be zero.
* Code that looks to get the line number should skip pause log entries.
* A pause entry could be added first. So logic that checks whether a log entry is added first (and use the initial base number) needs to also consider whether there are earlier pause entries.

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -474,15 +483,33 @@ private void LoadLogs(ConsoleLogsSubscription newConsoleLogsSubscription)
...
                         {
                             _logEntries.BaseLineNumber = lineNumber;
                         }
```
JamesNK, line 489:
```suggestion
                        _logEntries.BaseLineNumber ??= lineNumber;
```

src/Aspire.Dashboard/Components/Pages/ConsoleLogs.razor.cs
```diff
@@ -474,17 +482,21 @@ private void LoadLogs(ConsoleLogsSubscription newConsoleLogsSubscription)
...
+                        var logEntry = logParser.CreateLogEntry(content, isErrorOutput);
+                        if (logEntry.Timestamp is not null && timestampFilterDate is not null && !(logEntry.Timestamp > timestampFilterDate))
                         {
-                            _logEntries.BaseLineNumber = lineNumber;
+                            continue;
                         }
 
-                        var logEntry = logParser.CreateLogEntry(content, isErrorOutput);
-                        if (timestampFilterDate is null || logEntry.Timestamp is null || logEntry.Timestamp > timestampFilterDate)
+                        if (logEntry.Timestamp is not null && _logEntries.GetPauseEntries().Select(e => e.Pause).Cast<LogPauseViewModel>().FirstOrDefault(pause => pause.Contains(logEntry.Timestamp.Value)) is { } pause)
```
adamint, line 493: If the user refreshes the page, may add several prior pauses to log entries. Each needs to be checked. I added a list of pause entries to LogEntries to speed up this call

JamesNK (approved):

JamesNK:
I think it's much better.

I fixed a bug when you remove log entries while a pause is active. I noticed a race condition bug on log entries so added locks around modifying it. Also did some minor clean up.

PR #8199, opened 2025-03-20
Run only quarantined tests in outer loop

RussKie:
Allow quarantine flaky tests, which excludes the test from being executed by the normal CI pipelines. The quarantined tests get executed on preset schedule by a cron job for some period of time, after which they could manually unquarantined, if their stability improves.

The quarantine attribute and the discoverer as well as some of the MSBuild infra are copied from dotnet/aspnetcore.

The test execution is implemented in an Arcade SDK way, which should make it easy to maintain and enhance, if required.

The quarantined test can be invoked with the following command:

    ./build.cmd -projects ./tests/Shared/SolutionTests.proj
                -restore -build -test
                /p:RunQuarantinedTests=true

Setting `RunQuarantinedTests=false` (which is default, so can be omitted) will execute all the non-quarantined tests.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -15,7 +15,7 @@
   </ItemGroup>
 
   <Import Condition="'$(SampleProject)' == 'true' or '$(CI)' != 'true' " Project="eng\Versions.dev.targets" />
-  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' " Project="eng\Versions.targets" />
+  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' and '$(RunTestsOnGithubActions)' != 'true' " Project="eng\Versions.targets" />
 
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -15,7 +15,7 @@
   </ItemGroup>
 
   <Import Condition="'$(SampleProject)' == 'true' or '$(CI)' != 'true' " Project="eng\Versions.dev.targets" />
-  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' and '$(RunTestsOnGithubActions)' != 'true' " Project="eng\Versions.targets" />
+  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' " Project="eng\Versions.targets" />
 
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
--- a/eng/Testing.props
+++ b/eng/Testing.props
@@ -1,0 +1,19 @@
+<Project>
+  <PropertyGroup Condition="'$(UseVSTestRunner)' != 'true'">
+    <_QuarantinedTestRunAdditionalArgs>-trait "quarantined=true"</_QuarantinedTestRunAdditionalArgs>
+    <_NonQuarantinedTestRunAdditionalArgs>-notrait "quarantined=true"</_NonQuarantinedTestRunAdditionalArgs>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(UseVSTestRunner)' == 'true'">
+    <_QuarantinedTestRunAdditionalArgs>--filter "quarantined=true"</_QuarantinedTestRunAdditionalArgs>
+    <_NonQuarantinedTestRunAdditionalArgs>--filter "quarantined!=true"</_NonQuarantinedTestRunAdditionalArgs>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <BlameHangTimeout Condition="'$(BlameHangTimeout)' == ''">10m</BlameHangTimeout>
+    <_BlameArgs>--blame --blame-hang-timeout $(BlameHangTimeout) --blame-crash</_BlameArgs>
+
+    <TestRunnerAdditionalArguments Condition="'$(RunQuarantinedTests)' == ''">$(TestRunnerAdditionalArguments) $(_NonQuarantinedTestRunAdditionalArgs) $(TestRunnerAdditionalArguments) $(_BlameArgs)</TestRunnerAdditionalArguments>
+    <TestRunnerAdditionalArguments Condition="'$(RunQuarantinedTests)' == 'true'">$(TestRunnerAdditionalArguments) $(_QuarantinedTestRunAdditionalArgs) $(TestRunnerAdditionalArguments) $(_BlameArgs)</TestRunnerAdditionalArguments>
+  </PropertyGroup>
+</Project>
--- a/eng/Testing.targets
+++ b/eng/Testing.targets
@@ -1,0 +1,3 @@
+<Project>
+
+</Project>
--- a/tests/Shared/SolutionTests.proj
+++ b/tests/Shared/SolutionTests.proj
@@ -1,0 +1,30 @@
+<Project Sdk="Microsoft.Build.Traversal">
+  <!--
+    This file is used to build all the eligible test projects in the repo.
+    It is used to execute the tests in the outer loop and in the CI pipeline.
+
+    Generally, this project should invoked in the following way:
+
+      ./build.cmd -projects ./tests/Shared/SolutionTests.proj -restore -build -test /p:RunQuarantinedTests=[true|false]
+
+   -->
+  <PropertyGroup>
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\Shared\**\*Tests.csproj" />
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\testproject\**\*Tests.csproj" />
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\TestingAppHost1\**\*Tests.csproj" />
+
+    <!-- This runs in a separate job -->
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\Aspire.Workload.Tests\**\*Tests.csproj" />
+
+    <_TestProjectsToInclude Include="$(RepoRoot)tests\Aspire.Hosting.Tests\*Tests.csproj" />
+
+    <!-- Add all the test projects we want to build as project references, so the traversal SDK can build them -->
+    <ProjectReference Include="@(_TestProjectsToInclude)"
+                      Exclude="@(_TestProjectsToExclude)" />
+  </ItemGroup>
+
+</Project>
--- a/.github/workflows/tests-outerloop.yml
+++ b/.github/workflows/tests-outerloop.yml
@@ -44,6 +44,8 @@ jobs:
         run: ${{ env.DOTNET_SCRIPT }} dev-certs https --trust
 
       - name: Run quarantined tests
+        env:
+          CI: false
         run: |
           ${{ env.BUILD_SCRIPT }} -projects ${{ github.workspace }}/tests/Shared/SolutionTests.proj -restore -build -test -c Release /p:RunQuarantinedTests=true /bl:${{ github.workspace }}/artifacts/log/Release/test-quarantined.binlog
 
--- a/.github/workflows/tests-outerloop.yml
+++ b/.github/workflows/tests-outerloop.yml
@@ -1,0 +1,173 @@
+# Executes quarantined tests in the outerloop
+name: Outerloop Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 14 * * *' # 6am PST (14:00 UTC)
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+
+  test:
+    name: ${{ matrix.os.title }}
+    runs-on: ${{ matrix.os.name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - name: ubuntu-latest
+            title: Linux
+          - name: windows-latest
+            title: Windows
+    steps:
+      - name: Setup vars (Linux)
+        if: ${{ matrix.os.name == 'ubuntu-latest' }}
+        run: |
+          echo "DOTNET_SCRIPT=./dotnet.sh" >> $GITHUB_ENV
+          echo "BUILD_SCRIPT=./build.sh" >> $GITHUB_ENV
+
+      - name: Setup vars (Windows)
+        if: ${{ matrix.os.name == 'windows-latest' }}
+        run: |
+          echo "DOTNET_SCRIPT=.\dotnet.cmd" >> $env:GITHUB_ENV
+          echo "BUILD_SCRIPT=.\build.cmd" >> $env:GITHUB_ENV
+
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Trust HTTPS development certificate (Linux)
+        if: matrix.os.name == 'ubuntu-latest'
+        run: ${{ env.DOTNET_SCRIPT }} dev-certs https --trust
+
+      - name: Run quarantined tests
+        run: |
+          ${{ env.BUILD_SCRIPT }} -projects ${{ github.workspace }}/tests/Shared/SolutionTests.proj -restore -build -test -c Release /p:RunQuarantinedTests=true /bl:${{ github.workspace }}/artifacts/log/Release/test-quarantined.binlog
+
+      - name: Keep only relevant test logs
+        if: always()
+        shell: pwsh
+        run: |
+          # Define the directory to search for log files
+          $logDirectory = "${{ github.workspace }}/artifacts/log/**/TestLogs"
+
+          # Define the text to search for in the log files
+          $searchText = "No test matches the given testcase filter"
+          $resultsFilePattern = "Results File: (.+)"
+
+          # Get all .log files in the specified directory and its subdirectories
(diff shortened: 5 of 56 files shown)


Directory.Build.targets
```diff
@@ -15,7 +15,7 @@
   </ItemGroup>
 
   <Import Condition="'$(SampleProject)' == 'true' or '$(CI)' != 'true' " Project="eng\Versions.dev.targets" />
-  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' " Project="eng\Versions.targets" />
+  <Import Condition="'$(SampleProject)' != 'true' and '$(CI)' == 'true' and '$(IsGitHubActions)' != 'true' " Project="eng\Versions.targets" />
```
RussKie, line 18: This import caused me a fair a bit of grief (e.g. [here](https://github.com/RussKie/aspire/actions/runs/13960498451/job/39080809909#step:6:500)):
```
/home/runner/work/aspire/aspire/eng/Versions.targets(49,67): error MSB4057: The target "GitInfo" does not exist in the project. [/home/runner/work/aspire/aspire/src/Aspire.AppHost.Sdk/Aspire.RuntimeIdentifier.Tool/Aspire.RuntimeIdentifier.Tool.csproj]
/home/runner/work/aspire/aspire/eng/Versions.targets(49,67): error MSB4057: The target "GitInfo" does not exist in the project. [/home/runner/work/aspire/aspire/src/Aspire.AppHost.Sdk/Aspire.RuntimeIdentifier.Tool/Aspire.RuntimeIdentifier.Tool.csproj]
/home/runner/work/aspire/aspire/eng/Versions.targets(49,67): error MSB4057: The target "GitInfo" does not exist in the project. [/home/runner/work/aspire/aspire/src/Aspire.Cli/Aspire.Cli.csproj]
/home/runner/work/aspire/aspire/eng/Versions.targets(49,67): error MSB4057: The target "GitInfo" does not exist in the project. [/home/runner/work/aspire/aspire/src/Aspire.Hosting.Analyzers/Aspire.Hosting.Analyzers.csproj]
```

In general, these two imports look scary and nothing other dotnet/* repos use (except for maui). I'm questioning the need for these, but that's a separate discussion.
These are definitely not required for the outer loop runs.

RussKie:
Here's a test run to show it'll look: https://github.com/RussKie/aspire/actions/runs/13963811532

radical:
- Does `RunQuarantinedTests=true` mean that the quarantined tests will run in addition to all the other tests?
- I think it would be better to run essentially the tests in the same way as in the regular workflow, so they can run in parallel, and have the same test environment. I'll have a PR up soon to combine the `Templates`, and `Integrations` workflows, which should make reusing them easier.

Directory.Build.targets
joperezr: We should be able to clean this up.

Aspire.sln
```diff
@@ -3915,7 +3915,7 @@ Global
 		{39AD3DBA-6B79-44FC-9637-232544D0CA46} = {27381127-6C45-4B4C-8F18-41FF48DFE4B2}
 		{0C97F423-DAAE-4B63-B1B1-CB85E2840B98} = {27381127-6C45-4B4C-8F18-41FF48DFE4B2}
 		{23075FC1-D893-434D-9A20-02870B96ABA7} = {C424395C-1235-41A4-BF55-07880A04368C}
-		{F371C3EE-8AB3-4261-A5F2-9A1FF77ADC19} = {C424395C-1235-41A4-BF55-07880A04368C}
+		{F371C3EE-8AB3-4261-A5F2-9A1FF77ADC19} = {4981B3A5-4AFD-4191-BF7D-8692D9783D60}
```
joperezr, line 3918: Is this intentional?

radical:
I would suggest re-using the existing workflows that we have for tests, with the extra property passed to the test runs. This would mean that the flaky tests will be running with everything else in the same environment as they do on PRs, having the same load on resources.

cc @davidfowl @eerhardt

eng/Testing.props
```diff
@@ -0,0 +1,22 @@
+<Project>
+  <PropertyGroup Condition="'$(RunQuarantinedTests)' == 'true'">
+    <ArtifactsLogDir>$(ArtifactsDir)log\$(Configuration)\Quarantined\</ArtifactsLogDir>
```
joperezr, line 3: Out of curiosity, why do we need them to go into a separate directory

eng/Testing.props
```diff
@@ -0,0 +1,22 @@
...
+    <ArtifactsTestResultsDir>$(ArtifactsDir)TestResults\$(Configuration)\Quarantined\</ArtifactsTestResultsDir>
+    <TestRunnerAdditionalArguments></TestRunnerAdditionalArguments>
```
joperezr, line 5: Can we add a comment explaining what are we clearing out here?

eng/Testing.targets
```diff
@@ -0,0 +1,15 @@
+<Project>
+  <PropertyGroup Condition=" '$(SkipTests)' == '' and '$(IsTestProject)' == 'true' ">
```
joperezr, line 2:
```suggestion
  <PropertyGroup Condition=" '$(SkipTests)' != 'true' and '$(IsTestProject)' == 'true' ">
```
Super minor NIT but this way it will still work even if someone happens to do `/p:SkipTests=false`

joperezr: Oh nvm, I guess I was misunderstanding this.

eng/Testing.targets
```diff
@@ -0,0 +1,15 @@
...
+    <SkipTests>true</SkipTests>
+
+    <!-- Only run tests if the build is running on GitHub Actions  -->
+    <SkipTests Condition=" '$(IsGitHubActions)' == 'true' and '$(RunTestsOnGithubActions)' == 'true' ">false</SkipTests>
+
+    <!-- Only run tests if the build is running on Helix infra  -->
+    <SkipTests Condition=" '$(IsHelix)' == 'true' and '$(RunTestsOnHelix)' == 'true' ">false</SkipTests>
+
+    <!-- Otherwise, run tests on AzDO CI agents -->
+    <SkipTests Condition=" '$(RunTestsOnGithubActions)' != 'true' and '$(RunTestsOnHelix)' != 'true' ">true</SkipTests>
```
joperezr, line 13: Not sure I'm following when will this be needed. Looks from above that we are always automatically setting to true and only switching to false in the above two cases, so when this runs it seems like if the condition is true then SkipTests would already be true. Am I missing something?

tests/Aspire.Cli.Tests/Hosting/CliBackchannelTests.cs
```diff
@@ -50,6 +51,7 @@ public async Task AppHostRespondsToPingWithMatchingTimestamp()
 
     [Fact]
     [ActiveIssue("https://github.com/dotnet/aspire/issues/8113")]
+    [QuarantinedTest("https://github.com/dotnet/aspire/issues/8113")]
```
joperezr, line 54: What happens when a test is both ActiveIssue and QuarantinedTest? Is there an intention of separating a test that is flaky vs a test that is just failing consistently? The original discussion mentioned that there was value in separating plain failing tests vs flaky ones, and having the flaky use QuarantinedTest and failing tests using ActiveIssue.

joperezr: Not saying this is how it has to work, I'm interested in what you think we should do here.

tests/Shared/SolutionTests.proj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.Build.Traversal">
+  <!--
+    This file is used to build all the eligible test projects in the repo. 
+    It is used to execute the tests in the outer loop and in the CI pipeline.
+
+    Generally, this project should invoked in the following way:
+
+      ./build.cmd -projects ./tests/Shared/SolutionTests.proj -restore -build -test /p:IsGitHubActions=true /p:RunQuarantinedTests=[true|false]
```
joperezr, line 8: Can you help me understand a bit better why this is needed? I think that running `build.cmd -restore -build -test ...` already works and will test all the projects in the repo as we have this: https://github.com/dotnet/aspire/blob/fca6ddb80eaca68d0c694c187223c7006b8d6119/eng/Build.props#L7. If what you want is to make it so we only build tests and not the product, feels like we could reuse that Build.props and just add conditions that can be set via build.cmd command for this without having to introduce a new file. As an example, we already have `build.cmd -build -testnobuild` which builds all of the product but not the tests.

RussKie:
> * Does `RunQuarantinedTests=true` mean that the quarantined tests will run in addition to all the other tests?

No, the flow will only run the quarantined tests.

> * I think it would be better to run essentially the tests in the same way as in the regular workflow, so they can run in parallel, and have the same test environment. I'll have a PR up soon to combine the `Templates`, and `Integrations` workflows, which should make reusing them easier.

I disagree. 

The quarantined flow has different requirements to a check run for each build such as 
1. it runs on a dedicated schedule;
2. it doesn't have time constraints - i.e., it can run for 20 minutes, if necessary.
3. we can't know up front (without a significant very custom effort) which test project contain quarantined tests.

Following from the above:
- all affected tests can run serially, if required (points 1 and 2);
- no need to use excessive compute (point 2 and 3).

Aspire.sln
RussKie: Yes, moved Aspire.Components.Common.Tests.csproj to the "tests" v-folder because it is used across all tests.
image

I'm of an option the project should be renamed to not imply it is a project that contains tests (e.g., to "Aspire.TestsUtilities.csproj"), as its current names implies just that.

eng/Testing.targets
RussKie: Good catch, the value must be inverted. Thank you

eng/Testing.props
RussKie: We don't, I'm cleaning this out.

tests/Shared/SolutionTests.proj
RussKie: The reason I created a dedicated proj file was to execute only the required test projects, following the established filtering pattern in 
https://github.com/dotnet/aspire/blob/b3cc78f869a2a709454c78bac7d2ebe228e6f949/tests/Shared/GetTestProjects.proj#L17-L27

I treat this is an intermediary step to address the immediate task. The current implementations are currently a) fluid and b) complicated, and it doesn't feel appropriate to splice in these.

I'm not 100% clear on the target state, but I think if each test project declared what test infra it's expected to be executed (e.g., by a virtue of having an MSBuild property), it could allow us to simplify the current infra.

tests/Aspire.Cli.Tests/Hosting/CliBackchannelTests.cs
RussKie: That's a good question.

I thought these two can conicide, but after some tests (and some lost hair) I came to a conclusion that, essentially, these two attributes clash. `ActiveIssue` appears to act as a conditional attribute by skipping tests for a specified platform. And this prevents tests to be executed in the outerloop.

RussKie:
Here's how it'd look:

image

image

radical:
> I disagree.

IMHO, the problem here is that the tests are not without side-effects. The tests can be using resources like docker containers, creating/deleting networks, or can have dcp maybe taking too long to quit etc. All of which means that if the individual tests are run by themselves then they are running in a very different environment. This would mean the signal about the tests passing or failing might not really apply to the real test environment in the main CI.

Maybe we can identity the test projects that have at least one `Quarantined` test, and run those with `run-tests.yml`, same as the other tests. And in the reporting section only look at the ones that were Quarantined.

RussKie:
> The tests can be using resources like docker containers, creating/deleting networks, or can have dcp maybe taking too long to quit etc. All of which means that if the individual tests are run by themselves then they are running in a very different environment. This would mean the signal about the tests passing or failing might not really apply to the real test environment in the main CI.

This would indicate that the tests are poorly implemented, and they should be either redesigned to make stable or removed entirely. In such shape the tests are pretty much unusable to maintainers or contributors.
We can't/shouldn't make any assumptions or requirements that tests should only be executed in a vacuum.

> Maybe we can identity the test projects that have at least one Quarantined test, and run those with run-tests.yml, same as the other tests. And in the reporting section only look at the ones that were Quarantined.

It will likely be complex and very likely unreliable. Please feel free to take over this task.

radical:
My opinion is really an *assumption* based on looking at flaky tests in the past. But I haven't tried running them separately.

We can go with running them like this to start with. And you can track if re-enabling these in the main tests shows any issues related to the difference in test environment. And we can try to improve those tests.

radical:
re:moving-Aspire.Components.Common.Tests, the condition for archiving the tests needs to be updated so it is not triggered for test support projects.

Aspire.sln
RussKie: Fixed in #8279

radical:
@RussKie how do I trigger a run of the `tests-outerloop` workflow to check that?

radical:
Also, this is running all the tests on the `pipeline tests` job instead of just the non-helix ones.
https://dev.azure.com/dnceng-public/public/_build/results?buildId=1001556&view=logs&j=80af4256-dfad-5412-889b-4c79f0868294&t=41fa98d3-a032-5354-aa47-99008c7f4a7b

RussKie:
I don't think it's possible in this repo until it's merged.
I've been running actions from the main in my fork (https://github.com/RussKie/aspire/actions/workflows/tests-outerloop.yml).

radical:
> I don't think it's possible in this repo until it's merged. I've been running actions from the main in my fork ([RussKie/aspire/actions/workflows/tests-outerloop.yml](https://github.com/RussKie/aspire/actions/workflows/tests-outerloop.yml)).

It seems to be failing there. For the sake of validating this, we can drop the `schedule` from the outerloop workflow and have it executed on pull_request like others. And once we are sure that it works, we can add back the `schedule` and `trigger`.

RussKie:
I undone some changes, and now AzDO and Helix tests run as before -> https://dev.azure.com/dnceng-public/public/_build/results?buildId=1001618&view=results. Total tests: 8,388, which is lower than before due to filtering out the quarantined tests.

Looking at getting the outerloop run on GHA again.

RussKie:
> Those tests were being filtered out earlier through the `ActiveIssue` attribute, so I would expect the count to remain the same.

I don't think so. `ActiveIssueAttribute` conditionally skips tests, `QuarantinedTestAttribute` doesn't do that.

RussKie:
> > > Those tests were being filtered out earlier through the `ActiveIssue` attribute, so I would expect the count to remain the same.
> > 
> > 
> > I don't think so. `ActiveIssueAttribute` conditionally skips tests, `QuarantinedTestAttribute` doesn't do that.
> 
> Maybe I'm missing something. This suggests that the `failing` trait is added which we explicitly filter out. [dotnet/arcade@`5768bb1`/src/Microsoft.DotNet.XUnitExtensions.Shared/Attributes/ActiveIssueAttribute.cs#L68](https://github.com/dotnet/arcade/blob/5768bb1b57dfd7d9ac9a84749da0537162277e56/src/Microsoft.DotNet.XUnitExtensions.Shared/Attributes/ActiveIssueAttribute.cs#L68)
> 
> Isn't that what is being done with the `quarantined` attribute+trait?

I don't think so. I copied the dotnet/aspnetcore implementation directly:
- https://github.com/dotnet/aspnetcore/blob/main/src/Testing/src/xunit/QuarantinedTestTraitDiscoverer.cs
- https://github.com/dotnet/aspnetcore/blob/main/src/Testing/src/xunit/QuarantinedTestAttribute.cs

RussKie:
> Aspire.Templates.Tests	0	0	62	93

Should be fixed now.

RussKie:
> ⚠️ Aspire.Elastic.Clients.Elasticsearch.Tests	33	36	❌	❌

I don't see these tests to be run on AzDO. E.g., [this rolling build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1001771&view=results) didn't run these tests.
The tests are completely disabled on Helix:
https://github.com/dotnet/aspire/blob/7b7356077e91f99612e893fe340f61b90d2b6089/tests/Aspire.Elastic.Clients.Elasticsearch.Tests/Aspire.Elastic.Clients.Elasticsearch.Tests.csproj#L6-L8

RussKie:
Here's a full build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1001862&view=results; Total tests: 8,986

Here's a GHA run on my fork: https://github.com/RussKie/aspire/actions/runs/14209204488

radical (approved):
This looks amazing! Thank you for all the hard work on this, and persevering through the helix, and CI issues and headaches!

LGTM! 👍

tests/Shared/SolutionTests.proj
```diff
@@ -0,0 +1,30 @@
...
+    <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\Shared\**\*Tests.csproj" />
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\testproject\**\*Tests.csproj" />
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\TestingAppHost1\**\*Tests.csproj" />
+
+    <!-- This runs in a separate job -->
+    <_TestProjectsToExclude Include="$(RepoRoot)tests\Aspire.Workload.Tests\**\*Tests.csproj" />
```
radical, line 21: `Aspire.Workload.Tests` -> `Aspire.Templates.Tests`

tests/Shared/SolutionTests.proj
```diff
@@ -0,0 +1,30 @@
...
+
+    <_TestProjectsToInclude Include="$(RepoRoot)tests\Aspire.Hosting.Tests\*Tests.csproj" />
```
radical, line 23: I think we missed this. It should be including all the test projects, right?

RussKie: Bah... /facepalm

PR #8201, opened 2025-03-20
Implement pull based model resource states in `aspire` CLI.

mitchdenny:
This PR implements a pulll based model for resource states in the Aspire CLI. It changes the CLI so that it listens for connections on a unix socket and once the AppHost is launched the CLI attempts to connect to it.

This change leaves in roughly the same state as we were before it started but with a pull vs. push model.


src/Aspire.Hosting/Backchannel/AppHostRpcTarget.cs
```diff
@@ -0,0 +1,69 @@
...
+
+internal class AppHostRpcTarget(ILogger<AppHostRpcTarget> logger, ResourceNotificationService resourceNotificationService, IServiceProvider serviceProvider)
+{   
+    public async IAsyncEnumerable<(string Resource, string Type, string State, string[] Endpoints)> GetResourceStatesAsync([EnumeratorCancellation]CancellationToken cancellationToken)
+    {
+        var resourceEvents = resourceNotificationService.WatchAsync(cancellationToken);
+
+        await foreach (var resourceEvent in resourceEvents.WithCancellation(cancellationToken).ConfigureAwait(false))
+        {
+            if (resourceEvent.Resource.Name == "aspire-dashboard")
```
davidfowl, line 23: Maybe also skip hidden resources?

davidfowl (approved):
beautiful

src/Aspire.Hosting/Backchannel/AppHostRpcTarget.cs
mitchdenny: Will address this in follow-up PR. I also need to deal with health status better and replicas.

PR #8202, opened 2025-03-20
[CI] Combine Templates, and Integrations test workflows

radical:
[CI] Combine Templates, and Integrations test workflows

Since both the workflows reuse `run-tests.yml`, we can combine these two
into one workflow. This allows having a single place to check the
results, and final counts for the tests.

- Disable EndToEnd, and Playground tests for windows in the project files since we are collecting the list of tests separately for each OS

- unrelated: Correctly rename Aspire.Templates.Tests/WorkloadTestsBase.cs to TemplateTestsBase.cs


.github/workflows/tests-integration.yml
```diff
@@ -1,5 +1,4 @@
-# Executes the integration tests on all platforms
-name: Integration
+name: Tests
```
joperezr, line 1: Minor nit, but we should also rename this file as it is no longer just integration tests. Also, any reason why we can't combine this with run-tests.yml?

radical: `run-tests.yml` is the workflow being re-used for the various test/os combinations that we have here.

radical:
Before this can be merged the existing `Required` checks would need to be removed. And maybe add the new `Tests` one.

RussKie (approved):

radical:
@eerhardt @joperezr The `Required` checks need to be changed for this PR to be merged.

PR #8203, opened 2025-03-20
[SignalR-Playground] Ignore null fields when serializing negotiate response

cqnguy23:
- Add a serializing option to ignore null fields for negotiate response as a workaround for this [bug](https://github.com/dotnet/aspnetcore/issues/60935) in SignalR Client .NET Library

### DIFF
--- a/playground/signalr/SignalRServerlessWeb/Program.cs
+++ b/playground/signalr/SignalRServerlessWeb/Program.cs
@@ -1,3 +1,4 @@
+using System.Text.Json;
 using System.Text.Json.Serialization;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR.Management;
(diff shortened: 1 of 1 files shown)


IEvangelist:
Hi @cqnguy23,

This bug doesn't exist in the JavaScript SignalR client; therefore, this change isn't needed—unless the playground sample code is updated to use the .NET SignalR client.

cqnguy23:
> Hi @cqnguy23,
> 
> This bug doesn't exist in the JavaScript SignalR client; therefore, this change isn't needed—unless the playground sample code is updated to use the .NET SignalR client.

It's not necessary to run this playground sample, but I think it would prevent users from getting error in case they were looking at this sample as a reference, and they are using the .NET SignalR client instead.

IEvangelist (approved):

playground/signalr/SignalRServerlessWeb/Program.cs
```diff
@@ -1,9 +1,14 @@
+using System.Text.Json.Serialization;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR.Management;
 
 var builder = WebApplication.CreateBuilder(args);
 
 builder.Services.AddRazorPages();
+builder.Services.ConfigureHttpJsonOptions(options =>
+{
+    options.SerializerOptions.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
```
vicancy, line 10: this changes the default behavior for all the apis. `var ingoreNullOption = new JsonSerializerOptions...` for negotiate is enough?

vicancy (approved):

PR #8219, opened 2025-03-20
Remove ConfigureAwait from the Cli

eerhardt:
## Description

Since the CLI is a console application, requiring ConfigureAwait everywhere isn't necessary.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] No
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -473,7 +473,7 @@ dotnet_diagnostic.IDE0161.severity = silent
 # IDE0005: Remove unused usings. Ignore for shared src files since imports for those depend on the projects in which they are included.
 dotnet_diagnostic.IDE0005.severity = silent
 
-[{*.razor.cs,src/Aspire.Dashboard/Components/**.cs}]
+[{*.razor.cs,src/Aspire.Dashboard/Components/**.cs,src/Aspire.Cli/**.cs}]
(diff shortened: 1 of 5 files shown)


joperezr (commented):
Probably a dumb question, but isn't also there performance reasons to use `ConfigureAwait(false)` as you would then not need to capture the synchronization context?

joperezr (commented):
I understand that this is mostly useful on libraries, but at least the rule of thumb that I used to follow is to always use ConfigureAwait unless you explicitly know that you want to use the same context.

eerhardt:
> but isn't also there performance reasons to use ConfigureAwait(false) as you would then not need to capture the synchronization context?

Console applications use the default SynchronizationContext (unless you set one explicitly, which we don't).

https://devblogs.microsoft.com/dotnet/configureawait-faq/#when-should-i-use-configureawait(false)

> Even with these special cases, the general guidance stands and is a very good starting point: use ConfigureAwait(false) if you’re writing general-purpose library / app-model-agnostic code, and otherwise don’t.

joperezr (approved):
Got it, thanks for sharing the link. This LGTM.

PR #8224, opened 2025-03-20
Implement publisher advertisement and CLI integration

mitchdenny:
Introduce publisher advertisement events and enhance CLI functionality to retrieve publishers. Update the application to handle different execution modes and improve backchannel connection handling.

We now fire an event when the CLI asks the AppHost for publishers to advertise themselves. This is currently hooked up inside the `AddPublisher` method, and this will probably be where publishers declare the arguments that they accept and their mapping to options.

At the moment the `GetPublishersAsync` RPC call just returns an array of strings but once I start working on option mapping it will be a richer set of objects that are returned.

https://github.com/user-attachments/assets/a9930e9f-7147-4931-80b0-0f71c8714465


src/Aspire.Hosting/Publishing/PublisherAdvertisementEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+
+internal sealed class PublisherAdvertisementEvent : IDistributedApplicationEvent
+{
+    public void AddAdvertisement(string name)
+    {
+        var advertisement = new PublisherAdvertisement(name);
+        _advertisements.Add(advertisement);
+    }
+
+    private readonly List<PublisherAdvertisement> _advertisements = new ();
```
davidfowl, line 16:
```suggestion
    private readonly List<PublisherAdvertisement> _advertisements = [];
```

src/Aspire.Cli/Backchannel/AppHostBackchannel.cs
```diff
@@ -77,4 +81,30 @@ public async Task ConnectAsync(string socketPath, CancellationToken cancellation
...
+
+        var publishers = await rpc.InvokeWithCancellationAsync<string[]>(
+            "GetPublishersAsync",
+            Array.Empty<object>(),
+            cancellationToken).ConfigureAwait(false);
+
+        return publishers;
+    }
+
+    public void Dispose()
```
davidfowl, line 99: What's this for?

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class AzureContainerAppsInfrastructure(
 {
     public async Task BeforeStartAsync(DistributedApplicationModel appModel, CancellationToken cancellationToken = default)
     {
-        if (executionContext.IsRunMode)
+        if (!executionContext.IsPublishMode)
```
davidfowl, line 32: I don't like this at all. Can we just not run any of these in inspect mode?

src/Aspire.Hosting/Publishing/PublisherAdvertisementEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+
+internal sealed class PublisherAdvertisementEvent : IDistributedApplicationEvent
+{
+    public void AddAdvertisement(string name)
+    {
+        var advertisement = new PublisherAdvertisement(name);
+        _advertisements.Add(advertisement);
+    }
+
+    private readonly List<PublisherAdvertisement> _advertisements = [];
```
davidfowl, line 16: Does thread safety matter?

mitchdenny: This event is blocking sequential so no.

src/Aspire.Hosting.Azure.AppContainers/AzureContainerAppsInfrastructure.cs
mitchdenny: I'll look at selectively disabling.

src/Aspire.Cli/Backchannel/AppHostBackchannel.cs
mitchdenny: It was either this or making an RPC call to get the apphost to tear itself down. This seemed the simpler option.

davidfowl:
Conflcits!

maddymontaquila:
Fit n finish things either for this PR or a follow up issue -

- When you don't specify a publisher, we shouldn't show the warning that " ' ' " wasn't found
- When you do specify, it should prob be "Finding publisher..." or "Getting publisher..." instead of available ones because it makes me feel like you're searching for the whole list again
- are you planning on caching so we don't have to look up every time?

tests/Aspire.Hosting.Tests/Backchannel/AppHostBackchannelTests.cs
```diff
@@ -0,0 +1,241 @@
...
+using Microsoft.VisualStudio.Threading;
+using StreamJsonRpc;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Aspire.Hosting.Backchannel;
+
+public class AppHostBackchannelTests(ITestOutputHelper outputHelper)
+{
+
```
davidfowl, line 18:
```suggestion
```

davidfowl (approved):

PR #8227, opened 2025-03-21
[tests] PR validation speed ups

radical:
Improves `Tests` workflow time from ~14 mins to ~10 mins.
### Build packages once on Linux, and use with tests

This reduces the run time by avoiding package builds on Windows, and reduces total compute consumption.

### Merge github action enumerate-template-tests into enumerate-tests

These two tasks were called sequentially, and each would
checkout-code+install-dotnet before doing any work. By combining the
tasks, we get an improved run time.

### Avoid building test projects when TestNoBuild!=true

Saves on build unnecessary build times.


.github/workflows/run-tests.yml
```diff
@@ -28,6 +28,10 @@ on:
         required: false
         type: boolean
         default: false
+      requiresTestSdk:
```
joperezr, line 31: Are these inputs documented anywhere? If not, I wonder if it would be good to just add comments on top of each briefly explaining what they are for and what do they control. Some are really obvious, but some others (like requiresTestSdk for instance) might not be super obvious.

.github/workflows/run-tests.yml
```diff
@@ -83,14 +87,21 @@ jobs:
         if: inputs.os == 'ubuntu-latest'
         run: docker info
 
-      - name: Build with packages
+      - name: Download built nugets
         if: ${{ inputs.requiresNugets }}
-        env:
-          CI: false
-        run: ${{ env.BUILD_SCRIPT }} -restore -build -ci -pack -testnobuild /bl
+        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4.1.9
```
joperezr, line 92: Just curious, any reason why some of the actions we reference use tag names (like setup-dotnet above) but some others use direct hashes?

radical: @danmoseley could answer that

eng/Build.props
```diff
@@ -4,7 +4,7 @@
     <ProjectToBuild Include="$(RepoRoot)eng\dcppack\**\*.csproj" />
     <ProjectToBuild Include="$(RepoRoot)eng\dashboardpack\**\*.csproj" />
     <ProjectToBuild Include="$(RepoRoot)playground\**\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)tests\**\*.csproj"  />
+    <ProjectToBuild Include="$(RepoRoot)tests\**\*.csproj" Condition="'$(TestNoBuild)' != 'true'" />
```
joperezr, line 7: Interesting, does this mean that the -testnobuild switch wasn't really working before?

tests/Aspire.EndToEnd.Tests/IntegrationServicesFixture.cs
```diff
@@ -38,7 +38,7 @@ public IntegrationServicesFixture(IMessageSink diagnosticMessageSink)
     {
         _diagnosticMessageSink = diagnosticMessageSink;
         _testOutput = new TestOutputWrapper(messageSink: _diagnosticMessageSink);
-        BuildEnvironment = new(useSystemDotNet: !TestsRunningOutsideOfRepo);
+        BuildEnvironment = new(useSystemDotNet: true);
```
joperezr, line 41: Can you help me understand this change? Does this mean that integration tests will no longer require the test sdk ever? Also, does it mean it won't test against real nugets?

eng/Build.props
radical: Yeah, we weren't doing anything with that.

tests/Shared/GetTestProjects.proj
```diff
@@ -19,6 +19,8 @@
       <_TestProjectsToExclude Include="$(RepoRoot)tests\testproject\**\*Tests.csproj" />
       <_TestProjectsToExclude Include="$(RepoRoot)tests\TestingAppHost1\**\*Tests.csproj" />
 
+      <!-- This runs in a separate job -->
```
joperezr, line 22: We don't have to block this PR on it, but it would be good to have a README somewhere that shows how our matrix in GHA is composed, and which workflow controls which types of test projects.

tests/Aspire.EndToEnd.Tests/IntegrationServicesFixture.cs
radical: Yes, it won't require the test sdk ever. Since we don't have a workload anymore, we don't need to install anything into a test sdk, so a system installed sdk can be used. But it will still be running against real nugets when running outside the repo, same as it does now.

joperezr (approved):
Added some comments but this looks good overall. Thanks @radical!

Issue #5209, opened 2024-08-07
API Proposal: Add support for dynamic launch  IResourceBuilder.WithLaunchUrl()


kkirkfield:
## Proposed API

```cs
// aspire/src/Aspire.Hosting/ResourceBuilderExtensions.cs

using Aspire.Hosting.Lifecycle;

namespace Aspire.Hosting;

public static class ResourceBuilderExtensions
{
    public static IResourceBuilder<TResource> WithHttpLaunchUrl<TResource>(
        this IResourceBuilder<TResource> builder,
        string relativePath)
        where TResource : IResourceWithEndpoints =>
        builder.WithLaunchUrl("http", relativePath);

    public static IResourceBuilder<TResource> WithHttpsLaunchUrl<TResource>(
        this IResourceBuilder<TResource> builder,
        string relativePath)
        where TResource : IResourceWithEndpoints =>
        builder.WithLaunchUrl("https", relativePath);

    public static IResourceBuilder<TResource> WithLaunchUrl<TResource>(
        this IResourceBuilder<TResource> builder,
        string endpointName,
        string relativePath)
        where TResource : IResourceWithEndpoints
    {
        if (!builder.ApplicationBuilder.ExecutionContext.IsRunMode)
        {
            return builder;
        }

        var endpoint = builder.GetEndpoint(endpointName);

        if (endpoint is null)
        {
            return builder;
        }

        builder.ApplicationBuilder.Services.AddLifecycleHook(_ =>
            new LaunchUrlLifecycleHook(endpoint, relativePath));

        return builder;
    }
}
```

```cs
// aspire/src/Aspire.Hosting/Lifecycle/LaunchUrlLifecycleHook.cs

using System.Diagnostics;

namespace Aspire.Hosting.Lifecycle;

internal sealed class LaunchUrlLifecycleHook(
    EndpointReference endpoint,
    string relativePath)
    : IDistributedApplicationLifecycleHook
{
    public Task AfterEndpointsAllocatedAsync(
        DistributedApplicationModel appModel,
        CancellationToken cancellationToken = default)
    {
        var startInfo = new ProcessStartInfo()
        {
            FileName = new Uri(new Uri(endpoint.Url), relativePath).ToString(),
            UseShellExecute = true
        };

        Process.Start(startInfo);
        return Task.CompletedTask;
    }
}
```

## Usage Examples

Here is an example that launches a Scalar UI for interacting with Open API.

```cs
// AppHost/Program.cs

[...]

PR #8229, opened 2025-03-21
Support specifying custom URLs for resources

### LINKED ISSUE #4319: Allow specifying an address hostname for project/service binding
@davidfowl @mitchdenny This issue I'm trying to solve (or feature enhancement) is related to not being able to provide a specific hostname/address alias to resource.

> _While I'm not trying to externally access anything orchestrated locally I think there maybe some fundamental overlap with discussions here: #2887 and #3146 and pull request: #3492_ 

I have a couple use cases:

1. I have several UI projects that rely on OIDC and specific domains are whitelisted in Okta for redirects. _Of course I can lobby to statically add a whitelist of localhost:port but that requires another team that manages that infrastructure to get involved._
2. We have multi-tenancy logic that utilizes the request's host vanity domain to serve as a parameter to alter context and behavior at runtime. So with the same hosted instance a request with *.vanity-A.com would do something slightly different than *.vanity-B.com

So it would be awesome to be able supply additional address bindings specifically for local hosting that ALSO display on the dashboard. (_This was something I recall doing with Tye_)
image

An existing `tye.yaml`
```yaml
- name: ui
  project: ../src/App.UI/App.UI.csproj
  replicas: 1
  bindings:
  - name: http
    host: localhost
    protocol: http
  - name: https
    host: localhost
    protocol: https
  - name: local
    host: my.vanity-a.com
    protocol: https
    port: 1337
  env:
  - ASPNETCORE_ENVIRONMENT=Development
```

### A "kinda" workaround
I've been adapting an existing distributed application into the Aspire model. My goal has been to do this incrementally without being invasive or making modifications to the existing projects to prevent any refactor disruptions or coordination. I have been successful enough so far, **so kudos on baking in that flexibility from the start!** 

The workaround I have found like below so far appears to allow a workable path but not ideal of course from the dashboard or verbosity.

```c#
var ui = builder.AddProject<Projects.MyAccount_UI>("ui", null)
    .WithReference(api)
    .WithHttpsEndpoint(port: 7098, isProxied: false)
    .WithEnvironment("ASPNETCORE_ENVIRONMENT", "Development")
    // Need to specify address. Environment customization allows us to provide a hostname other
    // than localhost for Okta whitelist and general compatibility with current codebase state.
    .WithEnvironment(async context =>
    {
        if (context.ExecutionContext.IsRunMode)
        {
            var aspnetUrls = (await (context.EnvironmentVariables["ASPNETCORE_URLS"] as ReferenceExpression)!.GetValueAsync(context.CancellationToken));
            context.EnvironmentVariables["ASPNETCORE_URLS"] = aspnetUrls!.Replace("localhost", "my.vanity-a.com");
        }
    })
    // This does cause another addition endpoint to render under the resource in the dashboard but only localhost, no hostname address parameter
    .WithAnnotation(new EndpointAnnotation(protocol: ProtocolType.Tcp, "https", name: "my-vanity-b", port: 18001, isExternal: true, isProxied: false));
```

### LINKED ISSUE #6457: Provide a way to specify the relative URL for the resource (to be used by the dashboard)

### LINKED ISSUE #7104: Add support for "Display Name" for endpoints in the Dashboard

DamianEdwards:
## Description

This change introduces support for resource URLs. Resource endpoints get automatically projected into URLs, which can be customized, and extra URLs can be added that aren't associated with endpoints at all.

Fixes #6457
Fixes #7104
Fixes #4319
Contributes to #5209

https://github.com/user-attachments/assets/30dda79a-b97a-4190-a91e-e1d9e99240a3

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [ ] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [x] No
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] No
    - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)): [pending]


DamianEdwards:
@JamesNK right now the dashboard is not showing URLs that aren't associated with an endpoint in the resource details grid as there's currently a requirement for property grid items to have a name as that's used as the key. @davidfowl is suggesting we change the current "Endpoints" grid to instead just show URLs, and re-think the columns to match, but we'll need to figure out how to handle items with no name. Would appreciate your help there.

JamesNK:
I think we change "Endpoints" to "Urls" everywhere.

image

In the resource details we could change the columns to:

"Link" and "Endpoint name"

* "Link" would be a hyperlink with the display name as the text, and fallback to the URL if there is no display name.
* "Endpoint name" would be the optional endpoint name. If there is no endpoint name then the table cell would display "-" like other columns.

Anything can be the row key, as long as it is unique. I'm guessing it is possible to have the same URL multiple times so that couldn't be the key. There wasn't a good key for volumns so we used the index of the volumne in the volumne collection: https://github.com/dotnet/aspire/blob/e9aa3c292877c232f427ef2e42813f42adc5faae/src/Aspire.Dashboard/Model/ResourceViewModel.cs#L348
I think you could do the same thing here.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -63,14 +63,47 @@
...
+                                   HighlightText="@_filter" />
+                    </AspireTemplateColumn>
+                    @if (hasAnyEndpointDisplayNames)
+                    {
+                        <AspireTemplateColumn Sortable="true" SortBy="@(GridSort<DisplayedEndpoint>.ByAscending(i => i.DisplayName))" Title="@ControlStringsLoc[nameof(ControlsStrings.DisplayNameColumnHeader)]">
+                            <GridValue ValueDescription="@ControlStringsLoc[nameof(ControlsStrings.DisplayNameColumnHeader)]"
+                                       Value="@context.DisplayName"
+                                       EnableHighlighting="@(!string.IsNullOrEmpty(_filter))"
+                                       HighlightText="@_filter" />
+                        </AspireTemplateColumn>
```
JamesNK, line 96: I'm not a fan of having the URL and display name as separate columns. I'd rather there was just a "Link" column which displayed a hyperlink. If there is a display name then it is the text for the link.

Clicking on more details or copy would use the URL value.

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
```diff
@@ -11,7 +11,7 @@ internal static class ResourceEndpointHelpers
     /// <summary>
     /// A resource has services and endpoints. These can overlap. This method attempts to return a single list without duplicates.
     /// </summary>
-    public static List<DisplayedEndpoint> GetEndpoints(ResourceViewModel resource, bool includeInternalUrls = false)
+    public static List<DisplayedEndpoint> GetEndpoints(ResourceViewModel resource, bool includeInternalUrls = false, bool includeNonEndpointUrls = false)
```
JamesNK, line 14: Rename type to ResourceUrlHelpers. And method to GetUrls. And DisplayedEndpoint to DisplayedUrl. etc

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
```diff
@@ -24,13 +24,21 @@ public static List<DisplayedEndpoint> GetEndpoints(ResourceViewModel resource, b
...
+                if (!includeNonEndpointUrls && string.IsNullOrEmpty(url.EndpointName))
+                {
+                    continue;
+                }
+
                 endpoints.Add(new DisplayedEndpoint
                 {
-                    Name = url.Name,
-                    Text = url.Url.OriginalString,
+                    Name = url.EndpointName ?? "",
```
adamint, line 34: Why does this default to empty string instead of null?

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
```diff
@@ -41,7 +49,8 @@ public static List<DisplayedEndpoint> GetEndpoints(ResourceViewModel resource, b
         // - other urls
         // - endpoint name
         var orderedEndpoints = endpoints
-            .OrderByDescending(e => e.Url?.StartsWith("https") == true)
+            .OrderByDescending(e => e.SortOrder)
+            .ThenByDescending(e => e.Url?.StartsWith("https") == true)
```
adamint, line 53: Maybe add scheme as a field to DisplayedEndpoint and use that here instead?

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -298,15 +298,25 @@ public static IResourceBuilder<AzureCosmosDBEmulatorResource> WithDataExplorer(t
...
+                endpoint.UriScheme = "http";
+                endpoint.TargetPort = 1234;
+                endpoint.Port = port;
+            })
+            .WithUrls(context =>
+            {
+                var url = context.Urls.FirstOrDefault(u => u.Endpoint?.EndpointName == endpointName);
+                if (url is not null)
+                {
+                    url.DisplayText = "Data Explorer";
```
adamint, line 314: We would need to justify not having this value localized or provide a relatively simple way for users to set localized values

adamint: For known endpoints like this, we should override the display text on the dashboard with a localized value like is done in `src/Aspire.Dashboard/ResourceService/Partials.cs#124` with known commands.

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
DamianEdwards: @adamint just go with "Link"?

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
DamianEdwards: Being updated to default to `"-"` when there's no `EndpointName`.

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
DamianEdwards: Instead of the "https" check? This was already here :smile:

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
DamianEdwards: I went with "Link", take a look in the video in the PR description.

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
adamint: Oops I didn't realize it was already there. Might as well change it regardless!

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
```diff
@@ -299,15 +299,25 @@ public static IResourceBuilder<AzureCosmosDBEmulatorResource> WithDataExplorer(t
...
+                endpoint.UriScheme = "http";
+                endpoint.TargetPort = 1234;
+                endpoint.Port = port;
+            })
+            .WithUrls(context =>
+            {
+                var url = context.Urls.FirstOrDefault(u => u.Endpoint?.EndpointName == endpointName);
+                if (url is not null)
+                {
+                    url.DisplayText = Resources.DataExplorerUrlText;
```
adamint, line 315: I don't think this will work, because the requesting browser may be using a different locale than hosting. I think it should go in the place I linked before

DamianEdwards: Ah right. We need this to be localized in the dashboard itself then, as a known string?

src/Aspire.Dashboard/Model/ResourceEndpointHelpers.cs
DamianEdwards: I think this logic is just specifically looking to ensure https URLs are shown first so I don't think it requires the scheme to passed through separately. At this point we're just dealing with URL strings so there's no need to keep the scheme separate IMO.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBExtensions.cs
adamint: Right. I think we'll have to use a combo of resource name + endpoint name?

DamianEdwards: OK rigged it up.

JamesNK:
I think the link column should come first. It makes more sense to me than optional columns come after ones that are always present.

image

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -259,18 +280,35 @@
...
+    private static RenderFragment RenderUrlValue(DisplayedUrl vm, string filter)
     {
+        var highlighting = !string.IsNullOrEmpty(filter) && vm.Text != "-";
+        
+        // If there's no URL, e.g. this is a tcp:// URI, just show the text
         if (vm.Url is null)
         {
+            if (highlighting)
+            {
+                return@<FluentHighlighter HighlightedText="@filter" Text="@vm.Text" />;
```
JamesNK, line 292:
```suggestion
                return @<FluentHighlighter HighlightedText="@filter" Text="@vm.Text" />;
```

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -259,18 +280,35 @@
...
+            }
             return @<span>@vm.Text</span>;
         }
+        // If the URL and text are the same, render a link for the URL
         else if (string.Equals(vm.Url, vm.Text, StringComparison.Ordinal))
         {
+            if (highlighting)
+            {
+                return@<a href="@vm.Url" target="_blank"><FluentHighlighter HighlightedText="@filter" Text="@vm.Url" /></a>;
```
JamesNK, line 301:
```suggestion
                return @<a href="@vm.Url" target="_blank"><FluentHighlighter HighlightedText="@filter" Text="@vm.Url" /></a>;
```

src/Aspire.Dashboard/Components/Controls/ResourceDetails.razor
```diff
@@ -259,18 +280,35 @@
...
+            }
             return @<a href="@vm.Url" target="_blank">@vm.Url</a>;
         }
+        // Otherwise, render a link with the text as the anchor text & title as the URL
         else
         {
+            if (highlighting)
+            {
+                return@<a href="@vm.Url" title="@vm.Url" target="_blank"><FluentHighlighter HighlightedText="@filter" Text="@vm.Text" /></a>;
```
JamesNK, line 310:
```suggestion
                return @<a href="@vm.Url" title="@vm.Url" target="_blank"><FluentHighlighter HighlightedText="@filter" Text="@vm.Text" /></a>;
```

src/Aspire.Hosting/ApplicationModel/ResourceUrlsCallbackContext.cs
```diff
@@ -0,0 +1,38 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+///
```
JamesNK, line 10: todo

src/Aspire.Hosting/ApplicationModel/ResourceUrlsCallbackContext.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets the CancellationToken associated with the callback context.
+    /// </summary>
+    public CancellationToken CancellationToken { get; } = cancellationToken;
+
+    /// <summary>
+    /// An optional logger to use for logging.
+    /// </summary>
+    public ILogger Logger { get; set; } = NullLogger.Instance;
```
JamesNK, line 32: This feels a bit odd. Does it follow a pattern from other hosting code?

src/Aspire.Hosting/Dashboard/proto/resource_service.proto
```diff
@@ -124,8 +124,8 @@ message EnvironmentVariable {
 }
 
 message Url {
-    // The name of the url
-    string name = 1;
+    // The name of endpoint associated with the url
+    optional string endpoint_name = 1;
```
JamesNK, line 128: FYI this change to the proto is binary compatible. It's ok to make

src/Aspire.Hosting/ApplicationModel/ResourceUrlsCallbackContext.cs
DamianEdwards: Yeah it's the same patten used in the environment callback context but I didn't set it to an instance in the `ApplicationOrchestrator`. Will fix.

JamesNK:
Two column widths in URLs grid isn't consistent with others:

image

It should be consistent.

JamesNK:
@mitchdenny @davidfowl There is a lot in this PR. I'm focusing on dashboard changes and taking a brief look at hosting changes.

It would be good for hosting experts (ya'll) to double check hosting changes.

DamianEdwards:
@JamesNK 

> Two column widths in URLs grid isn't consistent with others:

Given the column order will be reversed so the link is first and the endpoint name second, and I'd expect the link to be longer than the endpoint name in most cases, do you still think it makes sense to have the column widths match that of a standard name/value property grid?

JamesNK:
I don't think either of them will be that long. A good URL display name will be short so it fits in the main grid so it should fit here. And most values here will still be the underlying URL anyway.

I think the consisency of grid widths, i.e. looking good, trumps showing a bit more of very long display names IMO.

DamianEdwards:
> I think the consisency of grid widths, i.e. looking good, trumps showing a bit more of very long display names IMO.

I'm not sure I agree but I'm fine with making them match until such time that we see evidence of it being a problem.

src/Aspire.Dashboard/Model/ResourceGraph/ResourceGraphMapper.cs
```diff
@@ -77,9 +77,9 @@ public static ResourceDto MapResource(ResourceViewModel r, IDictionary<string, R
...
     }
 
-    private static string ResolvedEndpointText(DisplayedEndpoint? endpoint)
+    private static string ResolvedEndpointText(DisplayedUrl? endpoint)
     {
-        var text = endpoint?.Text ?? endpoint?.Url;
+        var text = endpoint?.Url;
         if (string.IsNullOrEmpty(text))
         {
             return "No endpoints";
```
JamesNK, line 85: This should say "No URLs". And while you're here, could you move the text to a resource file. I should have done that 😄

src/Aspire.Dashboard/Model/ResourceUrlHelpers.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Diagnostics;
+using Aspire.Dashboard.Components.Controls;
+
+namespace Aspire.Dashboard.Model;
+
+internal static class ResourceUrlHelpers
+{
+    /// <summary>
+    /// A resource has services and endpoints. These can overlap. This method attempts to return a single list of URLs without duplicates.
+    /// </summary>
```
JamesNK, line 13: Unrelated, but this comment hasn't been valid for a while.
```suggestion
```

src/Aspire.Dashboard/Model/ResourceGraph/ResourceGraphMapper.cs
DamianEdwards: Actually the resource graph only looks at endpoint-associated URLs right now. When it was showing all URLs it would pick the first one and then show the display text un-hyperlinked which wasn't useful at all. Having it show the actual endpoint address is more useful IMO.

src/Aspire.Hosting/ProjectResourceBuilderExtensions.cs
```diff
@@ -458,6 +458,38 @@ private static IResourceBuilder<ProjectResource> WithProjectDefaults(this IResou
...
+                        if (context.Resource.TryGetEndpoints(out var endpoints))
+                        {
+                            foreach (var endpoint in endpoints)
+                            {
+                                if (endpoint.FromLaunchProfile)
+                                {
+                                    var url = context.Urls.FirstOrDefault(u => string.Equals(u.Endpoint?.EndpointName, endpoint.Name, StringComparisons.EndpointAnnotationName));
+                                    if (url is not null)
+                                    {
+                                        if (launchUri.IsAbsoluteUri)
```
davidfowl, line 476: Follow up: Discuss if an absolute launch urls wins over all over urls and becomes the only ones that shows up.

src/Shared/KnownUrls.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting;
+
+internal static class KnownUrls
+{
+    public static class DataExplorer
```
davidfowl, line 8: This is the most .NET thing ever. Why is this shared it's cosmos specific.

davidfowl (approved):
@DamianEdwards do a follow up and fix the AzureProvisioner to use this for deployments

src/Shared/KnownUrls.cs
DamianEdwards: Cos localization has to happen in the context of the dashboard. We do this for known commands too. We can add stuff to this for any other known link text we want to localize.

DamianEdwards:
> do a follow up and fix the AzureProvisioner to use this for deployments

@davidfowl can you elaborate? Which URLs should the provisioner use exactly and for what scenarios?

Issue #6669, opened 2024-11-14
Consider building an Azure Postgres client integration


davidfowl:
We should simplify the life of developers using aspire if we built an azure postgres integration that supported postgres containers for local dev and managed identity based azure postgres when deployed.

We need to fix this for EF and the EF Npgsql client integrations.

See the instructions here:

https://learn.microsoft.com/en-us/dotnet/aspire/database/postgresql-integration?tabs=dotnet-cli#add-azure-authenticated-npgsql-client

- [x] [Aspire Azure Npgsql](https://github.com/dotnet/aspire/pull/8230)
- [ ] Aspire Azure EF Npgsql

vincent-obi:
Hi @davidfowl , the current documentation does not explain the requirement to explicitly add a token as the password to the connection string generated  by AspireHost:
https://learn.microsoft.com/en-us/dotnet/aspire/database/azure-postgresql-entity-framework-integration?tabs=dotnet-cli#client-integration
I was able to get this working using the following:
```
if (builder.Environment.IsDevelopment())
{
    // For development use the connection string and password provided by the AppHost
    builder.Services.AddDbContext<InnovationOpsDbContext>(options =>
    options.UseNpgsql(builder.Configuration.GetConnectionString("db")
        ?? throw new InvalidOperationException("Connection string 'db' not found.")));
}
else
{
    // For production we need to use the Azure Identity library to get a token for the connection string
    builder.Services.AddDbContext<InnovationOpsDbContext>((sp, options) =>
    {
        var connection = builder.Configuration.GetConnectionString("db")
            ?? throw new InvalidOperationException("Connection string 'db' not found.");
        var credentials = new DefaultAzureCredential();
        var token = credentials.GetToken(new TokenRequestContext(["https://ossrdbms-aad.database.windows.net/.default"]));
        connection += $";password={token.Token}";

        options.UseNpgsql(connection);
    });
}

```

davidfowl:
cc @IEvangelist

PR #8230, opened 2025-03-21
Azure PostgreSQL client

sebastienros:
## Description

Adding Aspire.Azure.Npgsql client ~component~ _integration_ that can use a `TokenCredential` for authenticating to an Azure PostgreSQL database.

Part of https://github.com/dotnet/aspire/issues/6669

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [x] Yes
      - [ ] Yes
      - [x] No
      - [x] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [x] Yes
      - [ ] Yes
      - [x] No
        - https://github.com/dotnet/docs-aspire/issues/2883
  - [ ] No


davidfowl:
cc @roji

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+                }
+                else
+                {
+                    throw new InvalidOperationException("Could not determine username from token claims");
+                }
+            }
+
+            if (string.IsNullOrEmpty(dataSourceBuilder.ConnectionStringBuilder.Password))
+            {
+                // The token is not cached since it it refreshed for each new physical connection, or when it has expired.
```
roji, line 95: Isn't the ideal thing here to **not** re-retrieve for every physical open, i.e. cache the token and only retrieve it again once we're getting close to expiry?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+        builder.Services.AddNpgsqlDataSource(settings.ConnectionString ?? string.Empty, dataSourceBuilder =>
+        {
+            // The connection string required the username to be provided. Since it will depend on the Managed Identity that is used
+            // we attempt to get the username from the access token.
+
+            var credential = settings.Credential ?? new DefaultAzureCredential();
+
+            if (string.IsNullOrEmpty(dataSourceBuilder.ConnectionStringBuilder.Username))
+            {
+                var token = credential.GetToken(s_databaseForPostgresSqlTokenRequestContext, default);
```
roji, line 81: Is this a sync I/O call? Shouldn't necessarily be a problem as this is a one-time startup thing, but still...

roji (commented):
Nice to see a nice out-of-the-box experience here :) Though I'm sad that this is an Aspire thing - other non-Aspire users have the same problem and we should fix the problem for everyone. But anyway.

Implementation-wise, see note about caching the token. Also, how does this integrate with e.g. the Aspire EF component? Does the EF component simply resolve the NpgsqlDataSource that this new project registers there (with the Azure auth logic)?

src/Components/Aspire.Azure.Npgsql/AssemblyInfo.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Azure.Npgsql;
+using Aspire;
+
+[assembly: ConfigurationSchema("Aspire:Azure:Npgsql", typeof(AzureNpgsqlSettings))]
+
+[assembly: LoggingCategories(
```
eerhardt, line 9: Is there a way we can share this?

src/Components/Aspire.Azure.Npgsql/AssemblyInfo.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire.Azure.Npgsql;
+using Aspire;
+
+[assembly: ConfigurationSchema("Aspire:Azure:Npgsql", typeof(AzureNpgsqlSettings))]
```
eerhardt, line 7: Hmm, I was thinking this was more like how output caching and distributed caching work with Redis. It isn't a "full fledged" client integration on its own. Instead, it is an "extension" to client integration. So it uses the same schema location as the root library.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+        else if (payloadJson.TryGetProperty("unique_name", out var uniqueName))
+        {
+            username = uniqueName.GetString();
+        }
+
+        return username != null;
+    }
+
+    private static string AddBase64Padding(string base64) => (base64.Length % 4) switch
```
eerhardt, line 151: I thought you mentioned this is unnecessary and can be removed?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
eerhardt: Yes. But we have this problem elsewhere too. The root issue is that DI doesn't mix with async.

See also Long-lived Connections in DI (dotnet/aspire#565)

eerhardt: The only other option is to make our API async and someone needs to "await" in their Main. That is going to be necessary for Azure Redis, because the API we will need to use is async.

[https://github.com/Azure/Microsoft.Azure.StackExchangeRedis/?tab=readme-ov-file#usage](https://github.com/Azure/Microsoft.Azure.StackExchangeRedis/?tab=readme-ov-file#usage)

playground/AzurePostgres/AzurePostgres.Api/AzurePostgres.Api.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
eerhardt, line 1: Let's not make a whole new playground app for this.

I've updated https://github.com/dotnet/aspire/commit/c8315db2fa540c05a84060b23b21fc54389b96dd to allow both Azure and a Postgres container. You should be able to use that in your testing.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
roji: Yeah, makes sense. Probably not very important as it's one-time startup.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,157 @@
...
+        payload = AddBase64Padding(payload);
+
+        // Decode the payload from Base64Url
+        var decodedBytes = Convert.FromBase64String(payload);
+
+        // Parse the decoded payload as JSON
+        var reader = new Utf8JsonReader(decodedBytes);
+        var payloadJson = JsonElement.ParseValue(ref reader);
+
+        // Try to get the username from 'upn', 'preferred_username', or 'unique_name' claims
```
eerhardt, line 134: @roji @lossyrob - I'm trying to use this code in an app that uses a UserAssignedIdentity and it isn't working. Looking at the JWT token, I see these properties set:

```json
{
    "aud": "https://ossrdbms-aad.database.windows.net",
    "iss": "https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/",
    "iat": 1742596049,
    "nbf": 1742596049,
    "exp": 1742682749,
    "aio": "xxx",
    "appid": "xxx",
    "appidacr": "xxx",
    "idp": "https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/",
    "idtyp": "app",
    "oid": "xxxx",
    "rh": "xxx.",
    "sub": "xxx",
    "tid": "xxxx",
    "uti": "xxxx",
    "ver": "1.0",
    "xms_idrel": "7 20"
}
```

at first I thought I needed `appid`, but I think that is wrong as it is the "ClientId" of the UserAssignedIdentity. Looking at the Azure Postgres server, and comparing it's "Authentication" page in the portal with the values above, I think I need to use `oid`, which matches the `Object (principal) ID` of the UserAssignedIdentity. But that still fails with:

```
2025-03-21T17:52:13 stdout F       Npgsql.PostgresException (0x80004005): 28P01: password authentication failed for user "XXX-XXX-XXX-XXX-XXX"
2025-03-21T17:52:13 stdout F          at Npgsql.Internal.NpgsqlConnector.ReadMessageLong(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
2025-03-21T17:52:13 stdout F          at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
2025-03-21T17:52:13 stdout F          at Npgsql.Internal.NpgsqlConnector.AuthenticateSASL(List`1 mechanisms, String username, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.Internal.NpgsqlConnector.Authenticate(String username, NpgsqlTimeout timeout, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.Internal.NpgsqlConnector.<Open>g__OpenCore|214_1(NpgsqlConnector conn, SslMode sslMode, NpgsqlTimeout timeout, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.Internal.NpgsqlConnector.Open(NpgsqlTimeout timeout, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.PoolingDataSource.OpenNewConnector(NpgsqlConnection conn, NpgsqlTimeout timeout, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.PoolingDataSource.<Get>g__RentAsync|33_0(NpgsqlConnection conn, NpgsqlTimeout timeout, Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.NpgsqlConnection.<Open>g__OpenAsync|42_0(Boolean async, CancellationToken cancellationToken)
2025-03-21T17:52:13 stdout F          at Npgsql.NpgsqlConnection.Open()
2025-03-21T17:52:13 stdout F          at Npgsql.NpgsqlDataSource.OpenConnection()
2025-03-21T17:52:13 stdout F          at Program.<>c.<<Main>$>b__0_0(NpgsqlDataSource dataSource) in D:\git\aspire2\playground\AzurePostgres\AzurePostgres.Api\Program.cs:line 13
2025-03-21T17:52:13 stdout F          at lambda_method1(Closure, Object, HttpContext)
2025-03-21T17:52:13 stdout F          at Microsoft.AspNetCore.HostFiltering.HostFilteringMiddleware.Invoke(HttpContext context)
2025-03-21T17:52:13 stdout F          at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
2025-03-21T17:52:13 stdout F         Exception data:
2025-03-21T17:52:13 stdout F           Severity: FATAL
2025-03-21T17:52:13 stdout F           SqlState: 28P01
2025-03-21T17:52:13 stdout F           MessageText: password authentication failed for user "XXX-XXX-XXX-XXX-XXX"
2025-03-21T17:52:13 stdout F           File: auth.c
2025-03-21T17:52:13 stdout F           Line: 407
2025-03-21T17:52:13 stdout F           Routine: auth_failed
```

where `XXX-XXX-XXX-XXX-XXX` is my UserAssignedIdentity's `Object (principal) ID`.

Ideas here?

eerhardt: Looks like I've hit this issue before. We actually need the "PrincipalName" in the connection string. But when the app is a UserAssignedIdentity, the JWT token doesn't appear to have it. Is there a way to get the "PrinicpalName"?

eerhardt: Looks like if I make a request for scope `"https://management.azure.com/.default"`, and then read the `xms_mirid` property, that will give me the full name of the identity:

```
"xms_mirid":"/subscriptions/XXX-XXX-XXXX/resourcegroups/rg-eerhardt-XXXX/providers/Microsoft.ManagedIdentity/userAssignedIdentities/api_identity-XXXX"
```

Parsing the last segment of that string out will give us the PrincipalName. Is this the easiest way?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
sebastienros: I believe we should rely on the connection pool to cache the token indirectly. Or do you mean that this lambda is invoked on each logical connection (ado.net one)?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
sebastienros: Not in the case of JWT tokens.

src/Components/Aspire.Azure.Npgsql/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire;
+using Aspire.Npgsql;
+using Azure.Core;
+
+[assembly: ConfigurationSchema("Aspire:Npgsql", typeof(NpgsqlSettings))]
+[assembly: ConfigurationSchema("Aspire:Npgsql:Credential", typeof(TokenCredential))]
```
sebastienros, line 9: @eerhardt is that correct?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
roji: No, the password provider lambda indeed only gets invokved when a new physical (unpooled) connection needs to be created (that's when authentication occurs). So perf-wise, caching is indeed not super critical here as it doesn't affect pooled connections in any way.

However, it's still a good idea to reduce the time it takes to establish new physical connections - this can affect startup responsiveness and make things slower than they have to be. As an example, PostgreSQL recently added a new SSL protocol flow to knock out an additional network roundtrip from physical connection setup, so if we can avoid adding a needless one, we should...

Any specific reason to not cache here? It seems pretty trivial...

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
NickCraver: I'm confused at the one-time comment, it seems off vs. what we see in behavior. I'm torn here because while you want startup to go as fast as possible, continuing without base dependencies working also creates a world of unknown hurt afterwards that's harder to track down.

Given that in any rotation situation we need a token on startup and what should be a background refresh (delivering a token async, but almost always effectively sync with the many caching layers), it seems like a base situation of "here are all my rotating promises" and somewhere in startup there could effectively be a `.EnsureDependenciesAsync()` before going forward with the rest of app startup, so that what's wrong is super clear.

We have all libraries depending on the tree for `TokenCredential` already in the structure here, and I don't picture that changing, so perhaps some collection registered `TokenCredential`s live in could be in `Azure.Core` and we await that collection having all members populated as a first-class thing during startup. Is there an existing event `Azure.Core` could register to, to make this happen automatically perhaps?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
NickCraver: For what its worth, generally token sources are already pretty heavily cache underneath already. A reason _not_ to cache would be rapid invalidation upstream which needs to be accounted for - if we hit auth issues we'd need to be sure to purge any cache.

roji: Hey @NickCraver 👋 

That's a pure Azure question - can you confirm that azure Credentials indeed cache? If they do - and it's documented behavior we can rely on - then I agree we should not be caching here.

> A reason not to cache would be rapid invalidation upstream which needs to be accounted for - if we hit auth issues we'd need to be sure to purge any cache.

This could maybe be dealt with via automatic retrying of the open; in any case, I haven't heard of authentication tokens being invalidated before their published expiry time... I get the scenario in principle, but is this something that actually happens?

I'm asking since reducing physical connection time is also an important goal for good application latency - some applications scale down their connection pool in times of less load, and when they ramp up again it could be important that not every physical connection needlessly fetch new tokens while they already have a perfectly good one in hand...

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
roji: Now I'm confused too: this specific piece of code (unlike the PasswordProvider we're discussing in the other comment) really does seem to happen exactly once at startup, and only for getting the Username from Azure in case it isn't specified by the user. The username doesn't rotate (or at least, this PR doesn't allow it to) - the only thing that happens repeatedly is the PasswordProvider, which is for the PG password field (auth token), not for the username.

Or am I misreading things?

NickCraver: The underlying token call is shared for both cases, if it's the same credential instance  - unless we're changing here then this and later calls are really "who gets there first?" (this during startup as I understand it) then refreshes underneath.

roji: Sure, it's the same underlying token - my only point was that this is a special, first-time and once-only fetching of the auth token, only for the purposes of determining the username (and only if the user didn't specify it on the connection string). That in itself doesn't seem problematic to me - the only issue is the fact that this is done via synchronous I/O instead of asynchronously, because of DI limitations.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
NickCraver: If we're talking MSI, then yes - documented here: https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/managed-identities-faq#are-managed-identities-tokens-cached ...and I can confirm this directly from several interactions too. There is quite heavy caching at several layers. However, `TokenCredential` _could_ be anything so depending on scope that could not be true in some other use case.

As for invalidation, there are 2 aspects: authentication and authorization. The latter RBAC could be revoked at any time effectively invalidating the credential. I'm less familiar if tokens themselves are invalidated but given the latter, I'm not sure the former effectively matters.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
NickCraver: Gotcha - same page. I think that DI piece is worth solving - we're making serial remote calls in the current structure purely because of how DI is laid out when we really could all be parallel (e.g. Redis + Postgres, etc.) at startup. I know we're talking tokens here but even discovering service endpoints and such is in the same vein and something we often see. We could be getting tokens, discovering clusters, etc. all in parallel async here which is mostly network latency and aren't really embracing that first class when the number of cases for doing so seems to keep increasing.

src/Components/Aspire.Azure.Npgsql/AssemblyInfo.cs
eerhardt:
```suggestion
```

eerhardt: We can probably get rid of this whole thing since it references `Aspire.Npgsql` underneath, which will bring in this configuration schema.

src/Components/Aspire.Npgsql/AspirePostgreSqlNpgsqlExtensions.cs
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#pragma warning disable IDE0130 // Namespace "Microsoft.Extensions.Hosting" does not match folder structure
```
eerhardt, line 4: What is this needed for?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,179 @@
...
+            configureDataSourceBuilder?.Invoke(dataSourceBuilder);
+        });
+    }
+
+    private static AzureNpgsqlSettings ConfigureSettings(Action<AzureNpgsqlSettings>? userConfigureSettings, NpgsqlSettings settings)
+    {
+        var azureSettings = new AzureNpgsqlSettings();
+        CopySettings(settings, azureSettings);
+        userConfigureSettings?.Invoke(azureSettings);
+        CopySettings(azureSettings, settings);
```
eerhardt, line 88: Why is this extra copy needed?

src/Components/Telemetry.md
```diff
@@ -48,6 +48,29 @@ Aspire.Azure.Messaging.WebPubSub:
 - Metric names:
   - none (currently not supported by the Azure SDK)
 
+Aspire.Azure.Npgsql:
```
eerhardt, line 51: I wonder if we could just add this name to the `Aspire.Npgsql` section, so we don't need to dupe here.

src/Components/Aspire.Npgsql/AssemblyInfo.LoggingCategories.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Aspire;
+
+[assembly: LoggingCategories(
```
eerhardt, line 6: We shouldn't need to split this out. Azure.Npgsql -> Npgsql, so the user will get the configuration schema for the underlying library as well.

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
eerhardt: I've updated the latest code to handle this: https://github.com/dotnet/aspire/pull/8230/commits/2f49ff809243fa386eca93c9cdab602c28f53137

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
```diff
@@ -0,0 +1,215 @@
...
+    {
+        // The connection string required the username to be provided. Since it will depend on the Managed Identity that is used
+        // we attempt to get the username from the access token.
+
+        var credential = settings.Credential ?? new DefaultAzureCredential();
+
+        if (string.IsNullOrEmpty(dataSourceBuilder.ConnectionStringBuilder.Username))
+        {
+            // ensure to use the management scope, so the token contains user names for all managed identity types - e.g. user and service principal
+            var token = credential.GetToken(new([AzureManagementScope]), default);
```
lossyrob, line 118: Commented on this line in another discussion thread:

> One case to consider: is there instances where a user would have permissions for the AzureDatabaseForPostgresSqlScope, and not AzureManagementScope ?

https://github.com/dotnet/aspire/pull/8230#discussion_r2012103423

src/Components/Aspire.Npgsql/AspirePostgreSqlNpgsqlExtensions.cs
eerhardt: Should we disable this warning in general?

src/Components/Aspire.Azure.Npgsql/AspireAzureNpgsqlExtensions.cs
NickCraver: @roji It's cached even at the local wire server level (described as backend there) - can grab time and walk through how the infra works if helpful!

roji: Ah I see :) Thanks @NickCraver, no need to dive into the details. If you're saying that in most typical scenarios this gets cached and there's no need for us to add caching on top at this layer, that's more than good enough for me!

eerhardt (approved):
LGTM. Nice work here!

src/Components/Aspire.Npgsql/AspirePostgreSqlNpgsqlExtensions.cs
eerhardt: I removed these lines. I see them as "Messages" not warnings. We should disable this in the root .editorconfig if we care about these.

PR #8231, opened 2025-03-21
Support deep-linking Azure Service Bus child resources in clients

captainsafia:
Contributes towards https://github.com/dotnet/aspire/issues/7907.

- Set `EntityPath` property on ServiceBus connection strings using `EntityPath=queueName`, `EntityPath=topicName`, `EntityPath=topicName/Subscriptions/subscriptions`
- Add `AddServiceBusSender`, `AddServiceBusProcessor`, and `AddServiceBusReceiver` APIs and keyed versions to client integrations
- Update values injected for Azure Functions integration to pass through child resources


tests/Aspire.Azure.Messaging.ServiceBus.Tests/AspireServiceBusExtensionsTests.cs
```diff
@@ -196,6 +197,584 @@ public void AddAzureServiceBusClient_EnsuresConnectionStringIsCorrect(string con
...
+    {
+        var builder = Host.CreateEmptyApplicationBuilder(null);
+        PopulateConfiguration(builder.Configuration, connectionString);
+
+        // Check that settings are respected over value in in connection string
+        builder.AddAzureServiceBusSender("sb", settings =>
+        {
+            if (expectedTopicName == null)
+            {
+                // ServiceBus doesn't disambiguate between queue and topic names
```
captainsafia, line 261: One gross aspect of using the EntityPath to capture the connection info is that it doesn't do a good job of disambiguating between topic name and queue name. The client integration currently favors `QueueName` over `TopicName` when initializing single-value ServiceBus clients like the `ServiceBusSender`. This means that if you want to override the topic name in settings, you have to unset `QueueName` then set `TopicName` to the desired value.

We could work around this by adding some additional resolution logic that checks to see if the QueueName and TopicName are different from each other and if they are if the value is different from the one defined in the connection string. If they are, favor the value that's been overriden in settings.

That solution is kinda hacky. We could alternatively use individual properties instead of `EntityPath` but that would require re-writing most of this PR. 😅

davidfowl: Can we use another name them? QueueName=?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusTopicResource.cs
```diff
@@ -39,7 +39,10 @@ public string TopicName
     /// <summary>
     /// Gets the connection string expression for the Azure Service Bus Topic.
     /// </summary>
-    public ReferenceExpression ConnectionStringExpression => Parent.ConnectionStringExpression;
+    public ReferenceExpression ConnectionStringExpression =>
+        Parent.IsEmulator
+        ? ReferenceExpression.Create($"{Parent.ConnectionStringExpression}EntityPath={TopicName}")
+        : ReferenceExpression.Create($"Endpoint={Parent.ConnectionStringExpression};EntityPath={TopicName}");
```
eerhardt, line 45: Can we follow the same pattern as

https://github.com/dotnet/aspire/blob/08b424e1fa0b28aa225a67e2c6a7b5b959037f05/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsResource.cs#L50-L82

src/Components/Aspire.Azure.Messaging.ServiceBus/AspireServiceBusExtensions.cs
```diff
@@ -61,10 +124,124 @@ public static void AddKeyedAzureServiceBusClient(
...
+        protected override bool GetMetricsEnabled(AzureMessagingServiceBusSettings settings)
+            => false;
+
+        protected override bool GetTracingEnabled(AzureMessagingServiceBusSettings settings)
+            => !settings.DisableTracing;
+
+        protected override void BindClientOptionsToConfiguration(IAzureClientBuilder<TClient, TOptions> clientBuilder, IConfiguration configuration)
+        {
+            throw new NotImplementedException();
+        }
```
eerhardt, line 241:
```suggestion
```

This shouldn't be necessary.

tests/Aspire.Azure.Messaging.ServiceBus.Tests/AspireServiceBusExtensionsTests.cs
eerhardt: @JoshLove-msft @jsquire - do you have any opinions here? I think we have 3 viable options:

* What we have now - 2 properties - `QueueName` and `TopicName`
* `EntityName`
* `QueueOrTopicName` - same as the Azure SDK uses in the API

captainsafia: I'm leaning towards `QueueOrTopicName` myself given the only time a topic is used to initialize a sub-client without a subscription is in the `ServiceBusSender` which can take both.

src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
```diff
@@ -114,9 +130,36 @@ void IConnectionStringSettings.ParseConnectionString(string? connectionString)
                 return;
             }
 
-            // if we got here, it's a full connection string
-            // use the original connection string since connectionBuilder was modified above.
-            ConnectionString = connectionString;
+            // Use the connection string with the entitypath removed to
+            // support configuring processors/receivers/etc from the service
+            // bus client with names derived from settings.
+            ConnectionString = connectionBuilder.ConnectionString;
```
eerhardt, line 136: This surprises me a little bit. You used a connection that had one Queue on it, but want to override it with a new Queue. I guess this is consistent with what we do elsewhere - like the new CosmosDB change. I am a little concerned that we aren't using the original connection string.

Is there any change to the existing code if you just call `AddAzureServiceBusClient` and the connection string had an EntityPath before, but now it doesn't?

tests/Aspire.Azure.Messaging.ServiceBus.Tests/AspireServiceBusExtensionsTests.cs
jsquire: I like `QueueOrTopicName`.  we've seen a large number of developers be confused by the references to `entity` or `entity name` and not equate that with a queue/topic.

src/Components/Aspire.Azure.Messaging.ServiceBus/AspireServiceBusExtensions.cs
```diff
@@ -84,43 +261,117 @@ protected override IAzureClientBuilder<ServiceBusClient, ServiceBusClientOptions
...
+                }
+
+                if (string.IsNullOrEmpty(settings.QueueName) && string.IsNullOrEmpty(settings.TopicName))
+                {
+                    throw new InvalidOperationException($"A ServiceBusSender could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:{connectionName}' or specify a 'QueueName' or 'TopicName' in the '{configurationSectionName}' configuration section.");
+                }
+
+                var client = !string.IsNullOrEmpty(connectionString) ?
+                    new ServiceBusClient(connectionString) :
+                    new ServiceBusClient(settings.FullyQualifiedNamespace);
```
eerhardt, line 293: I wonder if there should be a way to be able to configure the options on the `ServiceBusClient` being used here.

Is there a way the user can configure the root client's Options?

eerhardt: Maybe we try reading the ServiceBusClient from DI? And if it is there, we use it.

src/Components/Aspire.Azure.Messaging.ServiceBus/AspireServiceBusExtensions.cs
```diff
@@ -317,17 +305,17 @@ protected override IAzureClientBuilder<ServiceBusReceiver, ServiceBusReceiverOpt
...
                 }
 
-                if (string.IsNullOrEmpty(settings.QueueName) && (string.IsNullOrEmpty(settings.TopicName) || string.IsNullOrEmpty(settings.SubscriptionName)))
+                if (string.IsNullOrEmpty(settings.QueueOrTopicName) && string.IsNullOrEmpty(settings.SubscriptionName))
                 {
-                    throw new InvalidOperationException($"A ServiceBusReceiver could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:{connectionName}' or specify a 'QueueName' or 'TopicName' and 'SubscriptionName' in the '{configurationSectionName}' configuration section.");
+                    throw new InvalidOperationException($"A ServiceBusReceiver could not be configured. Ensure valid connection information was provided in 'ConnectionStrings:{connectionName}' or specify a 'QueueOrTopicName' or 'SubscriptionName' in the '{configurationSectionName}' configuration section.");
```
JoshLove-msft, line 310: For subscriptions, don't you need both the QueueOrTopicName and the SubscriptionName?

captainsafia: Yeah, this error message is kinda hard to word now that we've combined the queue and topic name.

If you provide just the `QueueOrTopicName`, then it creates a receiver assuming that the value represents a queue.

If you provide a subscription ID, it assumes that it's a topic + subscription pairing.

Maybe the right phrasing here is `QueueOrTopicName` and/or `SubscriptionName`.

src/Components/Aspire.Azure.Messaging.ServiceBus/AzureMessagingServiceBusSettings.cs
captainsafia: Yes, this bit of code is designed to workaround the [ValidateEntityName](https://github.com/Azure/azure-sdk-for-net/blob/ecc9cd50619413f39f99949efd03e5d0fa89b7a4/sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs#L722) implementation in the SDK that assumes the QueueOrTopicName and SubscriptionName must match those in the connection string. By removing the entity path from the connection string, we circumvent the name checks it does.

> Is there any change to the existing code if you just call AddAzureServiceBusClient and the connection string had an EntityPath before, but now it doesn't?

I guess one breaking change here is that you are no longer getting the validation on teh entity names for any subclients that you were generating.

src/Components/Aspire.Azure.Messaging.ServiceBus/AspireServiceBusExtensions.cs
JoshLove-msft: `specify a 'QueueOrTopicName' and if using a subscription, the 'SubscriptionName' ...`?

src/Aspire.Hosting.Azure.ServiceBus/AzureServiceBusQueueResource.cs
```diff
@@ -97,7 +97,24 @@ public string QueueName
 
     // ensure Azure Functions projects can WithReference a ServiceBus queue
     void IResourceWithAzureFunctionsConfig.ApplyAzureFunctionsConfiguration(IDictionary<string, object> target, string connectionName)
-        => ((IResourceWithAzureFunctionsConfig)Parent).ApplyAzureFunctionsConfiguration(target, connectionName);
+    {
+        if (Parent.IsEmulator)
```
eerhardt, line 101: Thoughts on consolidating this code into the parent SB Resource as well? Just like we did for conneciton string?

src/Components/Aspire.Azure.Messaging.ServiceBus/AspireServiceBusExtensions.cs
```diff
@@ -61,10 +124,112 @@ public static void AddKeyedAzureServiceBusClient(
...
+                });
+        }
+
+        protected override void BindSettingsToConfiguration(AzureMessagingServiceBusSettings settings, IConfiguration config)
+        {
+            config.Bind(settings);
+        }
+
+        protected override bool GetHealthCheckEnabled(AzureMessagingServiceBusSettings settings)
+            => !string.IsNullOrEmpty(settings.HealthCheckQueueName) || !string.IsNullOrEmpty(settings.HealthCheckTopicName);
```
eerhardt, line 220: Now that we have a Queue or Topic name, I wonder if we should add the Health Check based on that Queue/Topic.

playground/AzureServiceBus/ServiceBus.AppHost/Program.cs
```diff
@@ -5,10 +5,10 @@
 
 var serviceBus = builder.AddAzureServiceBus("sbemulator");
 
-serviceBus.AddServiceBusQueue("queueOne", "queue1")
+var queue = serviceBus.AddServiceBusQueue("queueOne", "queue1")
     .WithProperties(queue => queue.DeadLetteringOnMessageExpiration = false);
 
-serviceBus.AddServiceBusTopic("topicOne", "topic1")
+var topic = serviceBus.AddServiceBusTopic("topicOne", "topic1")
```
eerhardt, line 11: This is actually the Subscription.

captainsafia:
@davidfowl @eerhardt Ping for review.

eerhardt (approved):
LGTM. Thanks!

playground/AzureServiceBus/ServiceBusWorker/Program.cs
```diff
@@ -1,10 +1,26 @@
...
+});
+builder.Services.AddSingleton(sp =>
+{
+    var client = sp.GetRequiredService<ServiceBusClient>();
+    return client.CreateProcessor("queue1", new ServiceBusProcessorOptions
+    {
+        MaxConcurrentCalls = 1, // Process one message at a time
+        AutoCompleteMessages = true
+    });
+});
```
davidfowl, line 23: What's this jank?

captainsafia: Since we're not doing a client integration to inject the Processors/Senders/Receivers, this showcases how someone could do that with the child resource connection string and existing DI APIs.

davidfowl: 😢

davidfowl: future?

captainsafia: Maybe?

One thing that makes Service Bus unique is that the child resources on the hosting side represent data entities (queues, topics, subscriptions) but the "child" concept on the client side represents actors like senders receivers this means that for any given child resource client integration there are two distinct options, one for the service bus and one for the processor receiver, etc.

PR #8235, opened 2025-03-21
Support deep-link for Azure WebPubSub Hubs in clients

captainsafia:
Contributes to #7907.

### DIFF
--- a/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
+++ b/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
@@ -151,20 +151,9 @@ public static class AzureWebPubSubExtensions
     /// <param name="builder">The builder for the distributed application.</param>
     /// <param name="hubName">The hub name. Hub name is case-insensitive.</param>
     /// <returns></returns>
-    [Obsolete("This method is obsolete because it has been superceded by the AddHub(IResourceBuilder<AzureWebPubSubResource>, string, string?) overload.", true)]
     public static IResourceBuilder<AzureWebPubSubHubResource> AddHub(this IResourceBuilder<AzureWebPubSubResource> builder, [ResourceName] string hubName)
     {
-        ArgumentNullException.ThrowIfNull(builder);
-        ArgumentException.ThrowIfNullOrEmpty(hubName);
-
-        AzureWebPubSubHubResource? hubResource;
-        if (!builder.Resource.Hubs.TryGetValue(hubName, out hubResource))
-        {
-            hubResource = new AzureWebPubSubHubResource(hubName, builder.Resource);
-            builder.Resource.Hubs[hubName] = hubResource;
-        }
-        var hubBuilder = builder.ApplicationBuilder.CreateResourceBuilder(hubResource);
-        return hubBuilder;
+        return AddHub(builder, hubName, hubName);
     }
 
     /// <summary>
@@ -185,7 +174,7 @@ public static class AzureWebPubSubExtensions
         AzureWebPubSubHubResource? hubResource;
         if (!builder.Resource.Hubs.TryGetValue(hubName, out hubResource))
         {
-            hubResource = new AzureWebPubSubHubResource(hubName, builder.Resource);
+            hubResource = new AzureWebPubSubHubResource(name, hubName, builder.Resource);
             builder.Resource.Hubs[hubName] = hubResource;
         }
         var hubBuilder = builder.ApplicationBuilder.CreateResourceBuilder(hubResource);
--- a/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
+++ b/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
@@ -146,6 +146,28 @@ public static class AzureWebPubSubExtensions
     }
 
     /// <summary>
+    /// Add hub settings
+    /// </summary>
+    /// <param name="builder">The builder for the distributed application.</param>
+    /// <param name="hubName">The hub name. Hub name is case-insensitive.</param>
+    /// <returns></returns>
+    [Obsolete("This method is obsolete because it has been superceded by the AddHub(IResourceBuilder<AzureWebPubSubResource>, string, string?) overload.", true)]
+    public static IResourceBuilder<AzureWebPubSubHubResource> AddHub(this IResourceBuilder<AzureWebPubSubResource> builder, [ResourceName] string hubName)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrEmpty(hubName);
+
+        AzureWebPubSubHubResource? hubResource;
+        if (!builder.Resource.Hubs.TryGetValue(hubName, out hubResource))
+        {
+            hubResource = new AzureWebPubSubHubResource(hubName, builder.Resource);
+            builder.Resource.Hubs[hubName] = hubResource;
+        }
+        var hubBuilder = builder.ApplicationBuilder.CreateResourceBuilder(hubResource);
+        return hubBuilder;
+    }
+
+    /// <summary>
     /// Adds an Azure Web Pub Sub hub resource to the application model.
     /// </summary>
     /// <param name="builder">The Azure WebPubSub resource builder.</param>
--- a/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
+++ b/src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
@@ -24,7 +24,7 @@ public static class AzureWebPubSubExtensions
     /// <returns>A reference to the <see cref="IResourceBuilder{T}"/>.</returns>
     /// <remarks>
     /// By default references to the Azure Web PubSub resource will be assigned the following roles:
-    /// 
+    ///
     /// - <see cref="WebPubSubBuiltInRole.WebPubSubServiceOwner"/>
     ///
     /// These can be replaced by calling <see cref="WithRoleAssignments{T}(IResourceBuilder{T}, IResourceBuilder{AzureWebPubSubResource}, WebPubSubBuiltInRole[])"/>.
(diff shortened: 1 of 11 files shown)


src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
```diff
@@ -74,6 +74,29 @@ public static void AddKeyedAzureWebPubSubServiceClient(
...
+    /// to support scenarios where multiple Hubs are referenced in the same application.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder" /> to read config from and add services to.</param>
+    /// <param name="connectionName">The name of the component to retrieve the connection string from the ConnectionStrings configuration section.</param>
+    /// <param name="configureSettings">An optional method that can be used for customizing the <see cref="AzureMessagingWebPubSubSettings"/>. It's invoked after the settings are read from the configuration.</param>
+    /// <param name="configureClientBuilder">An optional method that can be used for customizing the <see cref="IAzureClientBuilder{WebPubSubServiceClient, WebPubSubServiceClientOptions}"/>.</param>
+    /// <remarks>Reads the configuration from "Aspire.Azure.Messaging.WebPubSub:{name}" section.</remarks>
+    /// <exception cref="InvalidOperationException">Thrown when neither <see cref="AzureMessagingWebPubSubSettings.ConnectionString"/> nor <see cref="AzureMessagingWebPubSubSettings.Endpoint"/> is provided.</exception>
+    public static void AddKeyedAzureWebPubSubServiceClient(
```
captainsafia, line 88: cc: @vicancy I added an `AddKeyedAzureWebPubSubServiceClient` overload that takes a connection name and uses it as the service key to support scenarios where multiple child hubs are linked and you need to disambiguate between the two.

Let me know what you think of this.

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubHubResource.cs
```diff
@@ -11,13 +11,18 @@ namespace Aspire.Hosting.ApplicationModel;
...
     private readonly AzureWebPubSubResource _webpubsub = webpubsub ?? throw new ArgumentNullException(nameof(webpubsub));
     /// <summary>
     /// Gets the parent AzureWebPubSubResource of this AzureWebPubSubHubSettingResource.
     /// </summary>
     public AzureWebPubSubResource Parent => _webpubsub;
 
+    /// <summary>
+    /// Gets the connection string template for the manifest for Azure Web PubSub.
+    /// </summary>
+    public ReferenceExpression ConnectionStringExpression => ReferenceExpression.Create($"Endpoint={Parent.Endpoint};Hub={Name}");
```
eerhardt, line 25: Can we get a Hosting test for this?

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
```diff
@@ -74,6 +74,29 @@ public static void AddKeyedAzureWebPubSubServiceClient(
         new WebPubSubComponent().AddClient(builder, DefaultConfigSectionName, configureWithServiceKeyAsDefaultHubName, configureClientBuilder, connectionName: connectionName, serviceKey: serviceKey);
     }
 
+    /// <summary>
+    /// Registers <see cref="WebPubSubServiceClient"/> as a singleton for given <paramref name="connectionName"/> in the services provided by the <paramref name="builder"/>. This
+    /// overload does not require a service key and uses the connection name as the service key
+    /// to support scenarios where multiple Hubs are referenced in the same application.
```
eerhardt, line 80: Should we document that we expect the `connectionName` to be the name of a `Hub` resource?

eerhardt (approved):
Looks good to me. Would be nice to get @vicancy's thoughts/approval as well.

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
```diff
@@ -74,6 +74,29 @@ public static void AddKeyedAzureWebPubSubServiceClient(
...
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder" /> to read config from and add services to.</param>
+    /// <param name="connectionName">The name of the component to retrieve the connection string from the ConnectionStrings configuration section.</param>
+    /// <param name="configureSettings">An optional method that can be used for customizing the <see cref="AzureMessagingWebPubSubSettings"/>. It's invoked after the settings are read from the configuration.</param>
+    /// <param name="configureClientBuilder">An optional method that can be used for customizing the <see cref="IAzureClientBuilder{WebPubSubServiceClient, WebPubSubServiceClientOptions}"/>.</param>
+    /// <remarks>Reads the configuration from "Aspire.Azure.Messaging.WebPubSub:{name}" section.</remarks>
+    /// <exception cref="InvalidOperationException">Thrown when neither <see cref="AzureMessagingWebPubSubSettings.ConnectionString"/> nor <see cref="AzureMessagingWebPubSubSettings.Endpoint"/> is provided.</exception>
+    public static void AddKeyedAzureWebPubSubServiceClient(
+        this IHostApplicationBuilder builder,
+        string connectionName,
```
vicancy, line 90: what if we have 2 wps services with the same hub name which sounds a valid case, e.g. in the Host app:
```
var wps1 = builder.AddAzureWebPubSub("wps1");
var hub1 = wps1.AddHub("chat");
var wps2 = builder.AddAzureWebPubSub("wps2");
var hub2 = wps2.AddHub("chat");
var web = ...AddReference(hub1).AddReference(hub2);

```

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
```diff
@@ -74,6 +74,29 @@ public static void AddKeyedAzureWebPubSubServiceClient(
...
+        Action<AzureMessagingWebPubSubSettings>? configureSettings = null,
```
vicancy, line 91: what if this action sets `settings.HubName`? shall we throw if there is a conflict?

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
captainsafia: The `connectionName` doesn't end up being the name of the hub resource, just the service key. We derive the name of the hub resource from the connection string denoted by `connectionName` though.

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
captainsafia: Based on the current behavior (see `ExplicitHubNameOverridesConnectionStringHub` test), we favor the `HubName` that is set in settings if a user provides `configureSettings`. This matches the precendence order that we use for other client integrations.

src/Components/Aspire.Azure.Messaging.WebPubSub/AspireWebPubSubExtensions.cs
captainsafia: We don't have an overload for the WebPubSub client integration that supports providing a hub name that differs from the resource name. We do this for Azure Event Hubs (see https://github.com/dotnet/aspire/blob/9480d26740a699497aab0d61260a1f739ce786a6/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs#L139) and it allows you to have two hubs with the same hub name but different resource names to distinguish them on the Aspire side.

vicancy: shall we do the same for Web PubSub Hub now when it supports deep-link? Previously since this it is not reference-able, there is no need to allow resource name to be customizable. Besides the same hub name scenario, another concern is that currently there is an implicit restriction that if WebPubSub resource name should not be the same as the hub name otherwise it is overwritten:
```
var wpsChat = builder.AddAzureWebPubSub("chat");
var chatHub =  builder.AddAzureWebPubSub("wps1").AddHub("chat");
var web = ....AddReference(wpsChat).AddReference(chatHub); // the latter one overwrites the former one?
```

captainsafia: This is done in https://github.com/dotnet/aspire/pull/8235/commits/e97054f0c25936f3b753be89c93115492c8837ca. I placed on obsoletion marker on the old API in https://github.com/dotnet/aspire/pull/8235/commits/33e797663d12cc91cfd32cb6a4ddbee6c89f9f40.

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -151,6 +151,7 @@ public static IResourceBuilder<AzureWebPubSubResource> AddAzureWebPubSub(this ID
     /// <param name="builder">The builder for the distributed application.</param>
     /// <param name="hubName">The hub name. Hub name is case-insensitive.</param>
     /// <returns></returns>
+    [Obsolete("This method is obsolete because it has been superceded by the AddHub(IResourceBuilder<AzureWebPubSubResource>, string, string?) overload.", true)]
```
eerhardt, line 154: Why do we need to Obsolete this? Why can't we just have both overloads?

* `AddHub(this IResourceBuilder<AzureWebPubSubResource> builder, [ResourceName] string hubName)`
* `AddHub(this IResourceBuilder<AzureWebPubSubResource> builder, [ResourceName] string name, string hubName)`

And the top one calls the bottom one passing `hubName` for both `name` and `hubName`.

src/Aspire.Hosting.Azure.WebPubSub/AzureWebPubSubExtensions.cs
```diff
@@ -166,6 +167,31 @@ public static IResourceBuilder<AzureWebPubSubHubResource> AddHub(this IResourceB
...
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentException.ThrowIfNullOrEmpty(name);
+
+        // Use the resource name as the hub name if it's not provided
+        hubName ??= name;
+
+        AzureWebPubSubHubResource? hubResource;
+        if (!builder.Resource.Hubs.TryGetValue(hubName, out hubResource))
+        {
+            hubResource = new AzureWebPubSubHubResource(hubName, builder.Resource);
```
eerhardt, line 188: You actually want to split the properties out on the child resource, like we do for EventHubs:

https://github.com/dotnet/aspire/blob/5fd2b10c5fc644ae3aac08a0ca895d7fd72af86c/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubResource.cs#L19-L25

The `name` is the resource name (and can only have letters, numbers, and hyphens), and can't be duplicated in the model. You can only have 1 resource named "foo", even if they have different parents.

PR #8237, opened 2025-03-21
Refactor database explorer resource builders to not prefix their names the name of the first resource they are added to.

### LINKED ISSUE #2060: Make PgAdmin, RedisCommander, and incoming MyPhpAdmin opt in per resource.
Currently if we use `WithPgAdmin` (for example) on one resource, it opts all postgres resources in. It would be better to opt a particular resource into working with PgAdmin on a resource by resource basis.

paulomorgado:
## Description

Refactor database explorer resource builders to not prefix their names the name of the first resource they are added to.

Kept the behavior where, when a `containerName` is passed, that's the one that is used.

Fixes #2060

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [x] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - Probably


davidfowl:
@paulomorgado You might need to follow the pattern here https://github.com/dotnet/aspire/blob/8ad1d4b3d0a297f1b09ca62dfd62baf1e454ba5e/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs#L230-L234

This is to avoid conflicts with differnet apphosts.

paulomorgado:
> @paulomorgado You might need to follow the pattern here
> 
> https://github.com/dotnet/aspire/blob/8ad1d4b3d0a297f1b09ca62dfd62baf1e454ba5e/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs#L230-L234
> 
> This is to avoid conflicts with differnet apphosts.

I'm sorry @davidfowl, but I don't get what you mean. I only changed the resource name, which is internal only to each instance of app host.

davidfowl:
You're right. I guess there's no volume associated here.

sebastienros:
> when a containerName is passed, that's the one that is used.

Isn't it the current behavior? And if none is passed then a unique name is created by using the initiating resource.

Now with you PR two resources will have the same name if `containerName` is not set, meaning you have to set it if you have at least two. That might be ok but that's an insidious breaking change.

paulomorgado:
@sebastienros,

> Isn't it the current behavior? And if none is passed then a unique name is created by using the initiating resource.

No. If none is passed, the name is dependent on the name of the first resource adding this dependency.

> Now with you PR two resources will have the same name if `containerName` is not set, meaning you have to set it if you have at least two. That might be ok but that's an insidious breaking change.

There will only be 1 and only 1 resource of these. there's no way of duplicating the names that was not possible before.

paulomorgado:
@davidfowl, @sebastienros,

Is there anything missing here?

sebastienros:
I will take another look, assigning to 9.3 so I don't forget

sebastienros:
I reverted the changes for Milvus and MongoDB because they won't handle multiple calls (two resources are then declared with the same name). You can test it too, I did, and if you look at the other implementations (PgWeb, PhpMyAdmin, ...) where you did the change, you can see that the second call of the method return the existing container if there is one, so these work fine. These will also handle resources coming from other builders, meaning that if you create multiple dbs, a single explorer will render all these dbs.

sebastienros (approved):

src/Aspire.Hosting.Redis/RedisBuilderExtensions.cs
```diff
@@ -220,7 +220,7 @@ public static IResourceBuilder<RedisResource> WithRedisInsight(this IResourceBui
         }
         else
         {
-            containerName ??= $"{builder.Resource.Name}-insight";
+            containerName ??= "insight";
```
DamianEdwards, line 223: To match scheme of Redis Command above?
```suggestion
            containerName ??= "redisinsight";
```

sebastienros: Thanks, also improved a test since there was none for this one.

DamianEdwards (approved):

PR #8240, opened 2025-03-21
Mark AzurePublisherOptions as Experimental

eerhardt:
See https://github.com/dotnet/aspire/pull/7811#discussion_r2004870333

## Checklist

- Is this feature complete?
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] No
        - Link to aspire-docs issue [[New article]: Add ASPIREAZURE001 to aspire/diagnostics docs (dotnet/docs-aspire#2882)](https://github.com/dotnet/docs-aspire/issues/2882)

### DIFF
--- a/src/Aspire.Hosting.Azure/AzurePublisher.cs
+++ b/src/Aspire.Hosting.Azure/AzurePublisher.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics.CodeAnalysis;
 using Aspire.Hosting.ApplicationModel;
 using Aspire.Hosting.Publishing;
 using Azure.Provisioning;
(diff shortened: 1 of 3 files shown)


src/Aspire.Hosting.Azure/AzurePublisher.cs
```diff
@@ -13,6 +14,7 @@
 
 namespace Aspire.Hosting.Azure;
 
+[Experimental("ASPIREAZURE001", UrlFormat = "https://aka.ms/dotnet/aspire/diagnostics#{0}")]
```
joperezr, line 17: I know this is just following precedent of other experimental attributes, but in hindsight, I wonder if it would've been better to just have these attributes be all ASPIRE0xxx and just increment the number. That seems to be the pattern that some of the other repos using experimental do anyway.

joperezr (commented):
Separate from this PR, we should make sure that we are documenting these attributes and that the forward links work. (PS I see you added and linked a docs issue, my comment is just to do a pass for all of these)

joperezr (approved):

src/Aspire.Hosting.Azure/AzurePublisher.cs
eerhardt: @mitchdenny - thoughts on this? I think you set the precedent as you were the first to make experimental APIs.

https://github.com/dotnet/aspire/pull/3383/commits/9f499dbe81351c3c0f2598f7e0d294254555fe57

PR #8247, opened 2025-03-22
Rename publisher methods for clarity and add overloads for defaults

### LINKED ISSUE #8110: Support default publisher names for `AddDockerCompose` (and friends)
At the moment when we call `AddDockerCompose` we have to give it a publisher name. We should support a default name so people don't have to think about it (especially since we'll have the ability for `aspire deploy` to get a list of supported publishers.

mitchdenny:
Rename methods for Docker Compose and Kubernetes publishers to improve clarity and add overloads for default configurations. This enhances usability and maintains consistency across publisher methods.

Fixes #8110

### DIFF
--- a/playground/publishers/Publishers.AppHost/Program.cs
+++ b/playground/publishers/Publishers.AppHost/Program.cs
@@ -9,9 +9,9 @@ var builder = DistributedApplication.CreateBuilder(args);
 
 builder.AddAzureContainerAppEnvironment("env");
 
-builder.AddDockerComposePublisher("docker-compose");
+builder.AddDockerComposePublisher();
 
-builder.AddKubernetesPublisher("k8s");
+builder.AddKubernetesPublisher();
 
 #pragma warning disable ASPIREAZURE001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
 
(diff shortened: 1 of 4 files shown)


playground/publishers/Publishers.AppHost/Program.cs
```diff
@@ -9,9 +9,9 @@
 
 builder.AddAzureContainerAppEnvironment("env");
 
-builder.AddDockerCompose("docker-compose");
```
davidfowl, line 12: Use the default names here

playground/publishers/Publishers.AppHost/Program.cs
```diff
@@ -9,9 +9,9 @@
...
+builder.AddDockerComposePublisher("docker-compose");
```
davidfowl, line 12:
```suggestion
builder.AddDockerComposePublisher();
```

playground/publishers/Publishers.AppHost/Program.cs
```diff
@@ -9,9 +9,9 @@
...
 
-builder.AddKubernetes("k8s");
+builder.AddKubernetesPublisher("k8s");
```
davidfowl, line 14:
```suggestion
builder.AddKubernetesPublisher();
```

davidfowl (approved):

PR #8248, opened 2025-03-22
Add minimal dotnet watch support to CLI

### LINKED ISSUE #7897: Add support to aspire CLI for `aspire dev --watch` to enable watch support.
We need to check the version of the SDK to ensure that it is 9.0.200 otherwise it should error.

mitchdenny:
Introduce support for a watch mode in the CLI, allowing for continuous monitoring of project changes.

DRAFT because this currently isn't working because dotnet watch doesn't seem to be able to successfully rebuild the API service project:

image

Also noticed that when I launch the WaitForSandbox in dotnet watch mode the API service project doesn't start (works if you start it auotmatically). So there are probably a few underlying issues here with watch support.

Fixes: #7897

### DIFF
--- a/playground/waitfor/WaitForSandbox.ApiService/Program.cs
+++ b/playground/waitfor/WaitForSandbox.ApiService/Program.cs
@@ -19,6 +19,7 @@ app.MapGet("/", async (MyDbContext dbContext) =>
     // since we've proven connectivity to the others for now.
     var entry = new Entry();
     await dbContext.Entries.AddAsync(entry);
+    await dbContext.Entries.AddAsync(entry);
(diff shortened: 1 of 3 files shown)


mitchdenny:
OK so API service isn't starting because dbsetup never finishes because its been watched by dotnet watch (duh).

mitchdenny:
... and the issue with API service not rebuilding occurs even without the CLI, so it shouldn't block this PR. Things like blazor updates work fine. So I think this PR is OK to go in.

davidfowl:
That’s by design. Dotnet watch in this mode doesn’t work at all with top level statements 🥲

davidfowl (approved):

PR #8249, opened 2025-03-22
Automatically determine default template version from CLI version

### LINKED ISSUE #8198: [WebToolsE2E][Aspire] `aspire new` tries to use a non-existent project templates version by default.
**INSTALL STEPS**

1. Clean machine: Win11 x64 ENU
2. Install SDK 9.0.202
3. Install the latest Aspire 9.2.0-preview.1.25163.6

    - Apply dotnet9 feed

**REPRO STEPS**
1. Open CMD, create a folder, and create an Aspire App in it:
```
mkdir HelloAspire && cd HelloAspire
aspire new
```

**ACTUAL**
The command doesn't work and displays "'aspire' is not recognized as an internal or external command,
operable program or batch file."
Image

mitchdenny:
Implement a method to retrieve the default template version based on the CLI's informational version, replacing the hardcoded value.

Fixes #8198

### DIFF
--- a/src/Aspire.Cli/Utils/VersionHelper.cs
+++ b/src/Aspire.Cli/Utils/VersionHelper.cs
@@ -1,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Cli.Utils;
+
+internal static class VersionHelper
+{
+    public static string GetDefaultTemplateVersion()
+    {
+        // Write some code that gets the informational assembly version of the current assembly and returns it as a string.
+        var assembly = typeof(VersionHelper).Assembly;
+        var informationalVersion = assembly
+            .GetCustomAttributes(typeof(System.Reflection.AssemblyInformationalVersionAttribute), false)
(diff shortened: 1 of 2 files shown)


src/Aspire.Cli/Utils/VersionHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+    public static string GetDefaultTemplateVersion()
+    {
+        // Write some code that gets the informational assembly version of the current assembly and returns it as a string.
+        var assembly = typeof(VersionHelper).Assembly;
+        var informationalVersion = assembly
+            .GetCustomAttributes(typeof(System.Reflection.AssemblyInformationalVersionAttribute), false)
+            .OfType<System.Reflection.AssemblyInformationalVersionAttribute>()
+            .FirstOrDefault()?.InformationalVersion;
+
+        return informationalVersion ?? throw new InvalidOperationException("Unable to retrieve the assembly version.");
```
davidfowl, line 17: No default? Do you really want this to throw?

davidfowl (approved):

src/Aspire.Cli/Utils/VersionHelper.cs
mitchdenny: There should always be an informational version.

PR #8260, opened 2025-03-23
k8s first pass

prom3theu5:
no ingress support directly yet

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] Yes
  - [ ] No
- Did you add public API?
  - [ ] Yes
      - [ ] Yes
      - [ ] No
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No
  - [ ] Yes
      - [ ] Yes
      - [ ] No
  - [ ] No


src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
```diff
@@ -0,0 +1,368 @@
...
+    private void ProcessEndpoints()
+    {
+        if (!resource.TryGetEndpoints(out var endpoints))
+        {
+            return;
+        }
+
+        foreach (var endpoint in endpoints)
+        {
+            var port = endpoint.TargetPort ?? 80;
```
davidfowl, line 105: We should only default to 80 for http no? Otherwise it should fail?

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
```diff
@@ -0,0 +1,368 @@
...
+                    continue;
+                }
+
+                // Move connection strings to secrets
+                if (key.IsConnectionString())
+                {
+                    var expression = key.ToHelmSecretExpression(resource.Name);
+                    Secrets[key] = new HelmExpressionValue(expression, stringValue);
+                    continue;
+                }
```
davidfowl, line 196: I don't think we should do this. Only move values that themselves container secrets to secrets (secret parameters).

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
```diff
@@ -0,0 +1,368 @@
...
+
+                if (stringValue.IsHelmSecretExpression())
```
davidfowl, line 198: How does this happen?

prom3theu5: this occurs after helm expressions have been constructed in variables,
So for instance, if we take the Parameter param1 from the 'myapp' test app, it has a kubernetes secret generated, which includes a helm expression of '{{ .Values.secrets.myapp.param1 }}' as its value

Then the final values file for the helm chart with deploy time ability to set values, can supply that value

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
prom3theu5: I had that originally - the problem there though is ultimately this has to go in a configMap or a secret, and having it included in a configMap like all other env vars are didn't feel best, as its value contains references to other secrets (password for example).

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
prom3theu5: aye we should probably take the scheme into account here

src/Aspire.Hosting.Kubernetes/KubernetesPublisherOptions.cs
```diff
@@ -10,4 +10,66 @@ namespace Aspire.Hosting.Kubernetes;
 /// </summary>
 public sealed class KubernetesPublisherOptions : PublishingOptions
 {
+    /// <summary>
+    /// Gets or sets the name of the Helm chart to be generated.
+    /// </summary>
+    public string HelmChartName { get; set; } = "aspire";
```
davidfowl, line 16: Should this be the name of the application instead of aspire?

prom3theu5: yeah! I just defaulted it to aspire, I thought the cli would probably pass it through

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
davidfowl: This is what we do in the azure container apps case. The value is ony secret if it has a secret (transitively). This removes all conventions around key names so that things like azure functions and other future conventions will always work. No need to look for a prefix.

prom3theu5:
Gonna have to rework the unit tests - they work in xunit in my ide (Rider), but fail through dotnet test ran from cli. Strange...
I'll remove the embedded resources tomorrow, and extract them into string constants like the compose test does.
Will also remove the reusable method and just test each file directly, as its too hard to see which file actually fails there

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
davidfowl: I don't fully understand how. This is processing the environment variables of a resource, right? The value can't be a help expression at this stage unless the user or some process directly set a helm expression in env. What am I missing?

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
prom3theu5: Makes sense ^^ - Will rework tomorrow to check the existing Secrets collection for an expression match on the contents of the value first.
6:30am now haha

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
davidfowl: Ah I seem this runs on the result of the `ProcessValue` call? Why not use a strongly typed object to return the helm expression instead of converting to a string first?

prom3theu5: Its because a little above, ProcessValueAsync is called, which processes, and transforms the end `stringValue`, so when you get to that line, its already gone through AllocateParameter etc

davidfowl:
This is looking really promising @prom3theu5 !

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
prom3theu5: Thats a good shout, as we already construct one (HelmExpressionValue) which is just a struct with the Expression, and Value

prom3theu5:
@davidfowl you ok with the dual context? (One for publisher, one for the state of a processed resource) There were so many cross cutting concerns in the single context felt better to split them into two and isolate concerns

davidfowl:
Yep! it looks much cleaner.

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
afscrome: Should we default to a port number > 1024?  Ports below 1024 require root permissions or reconfiguring the `net.ipv4.ip_unprivileged_port_start` security context.

prom3theu5: Good call. Will switch to 8080 for http scheme, 8443 for https and 9000 for everything else

tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs
```diff
@@ -9,48 +9,85 @@
...
+                .WithReference(cs)
+                .WithVolume("logs", "/logs")
+                .WithArgs("--cs", cs.Resource);
+
+            builder.AddProject<TestProject>("project1", launchProfileName: null)
+                .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EXCEPTION_LOG_ATTRIBUTES", "true")
+                .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_EMIT_EVENT_LOG_ATTRIBUTES", "true")
+                .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_OTLP_RETRY", "in_memory")
+                .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_ASPNETCORE_DISABLE_URL_QUERY_REDACTION", "true")
+                .WithEnvironment("OTEL_DOTNET_EXPERIMENTAL_HTTPCLIENT_DISABLE_URL_QUERY_REDACTION", "true")
```
davidfowl, line 67: Why do you need this?

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
```diff
@@ -0,0 +1,374 @@
...
+                }
+            }
+        }
+    }
+
+    private void ProcessEnvironmentHelmExpression(HelmExpressionWithValue helmExpression, string key)
+    {
+        switch (helmExpression)
+        {
+            case {IsHelmSecretExpression: true, ValueContainsSecretExpression: false}:
```
davidfowl, line 206: nit: Formatting

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
```diff
@@ -0,0 +1,374 @@
...
+            return;
+        }
+
+        foreach (var endpoint in endpoints)
+        {
+            var port = endpoint.TargetPort ?? endpoint.UriScheme switch
+            {
+                "http" => 8080,
+                "https" => 8443,
+                _ => 9000,
```
davidfowl, line 108: What port is this supposed to be? The internal port? I think you can only choose a default for project resources. Otherwise, this has to throw right? The container needs to be configured to listen on this port.

prom3theu5: The target port is returned, if something like sqlServer for example was configured, it should have its target port set as 1433 shouldn't it?
This was just the fall back for something that doesn't have a target port set, which probably means endpoint annotations on the resource are wrong
We can choose to only apply this for project resources, or else "explode" with an exception and end publishing?

prom3theu5: something like

```csharp
 foreach (var endpoint in endpoints)
        {
            var internalPort = endpoint.TargetPort;

            if (resource is ProjectResource _)
            {
                internalPort ??= endpoint.UriScheme switch
                {
                    "http" => 8080,
                    "https" => 8443,
                    _ => throw new NotSupportedException($"Unsupported scheme: {endpoint.UriScheme}"),
                };
            }

            if (internalPort is null)
            {
                throw new InvalidOperationException("TargetPort for endpoint must be set");
            }

            EndpointMappings[endpoint.Name] = new(endpoint.UriScheme, resource.Name, internalPort.Value, internalPort.Value, endpoint.Name);
        }
        ```

File: tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs (hunk 5 above)
Comment (prom3theu5) [reply tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs:67]: The test context doesn't appear to set them when you run from cli, but they exist in the output config maps and at run time as they're set by the AppHost aren't they?

File: src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs (hunk 7 above)
Comment (davidfowl) [reply src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs:108]: Yea, either this or it becomes a parameter.

Comment (prom3theu5) [reply src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs:108]: there's also "TargetPortEnvironmentVariable" to take into account, It's internal atm - but we'd probably want to make it public here so we can see if a value has been set too?

Comment (davidfowl) [reply src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs:108]: That shouldn't be required. When it is set the correct data is expressed in the model as an environment variable:

https://github.com/dotnet/aspire/blob/e5014c7bedb48ca663b188d58260eb9d6210f349/src/Aspire.Hosting/ResourceBuilderExtensions.cs#L566-L569

All we need to do is model endpoints appropriately and it should all hang together. 

The answer here might be to avoid doing anything clever and using a parameter if the value is not set.

File: tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs (hunk 5 above)
Comment (davidfowl) [reply tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs:67]: These are set by the AddProject call. Something funky is going on. See

https://github.com/dotnet/aspire/blob/b3cc78f869a2a709454c78bac7d2ebe228e6f949/tests/Aspire.Hosting.Docker.Tests/DockerComposePublisherTests.cs#L37-L38

https://github.com/dotnet/aspire/blob/b3cc78f869a2a709454c78bac7d2ebe228e6f949/tests/Aspire.Hosting.Docker.Tests/DockerComposePublisherTests.cs#L82-L84

No magic.

File: src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs (hunk 7 above)
Comment (prom3theu5) [reply src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs]: I've just done something along those lines in this: https://github.com/dotnet/aspire/pull/8260/commits/ce22f8a07e5db7d84764392056c6203ce57be955

However - if you see the message in the commit, I'm having a little issue with separation of env vars.
As you linked above, if we set WithEndpoint(env: "Port"), this forces an EV callback to populate an EnvVar called Port.
But i don't see a way without having access to

CleanShot 2025-03-24 at 01 17 03

to know that this env var value that has been specified in the AppHost is meant to be referenced by the endpoint. When its populated back into the ev - this "link" is lost if you are just interrogating the annotations

File: tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs (hunk 5 above)
Comment (prom3theu5) [reply tests/Aspire.Hosting.Kubernetes.Tests/KubernetesPublisherTests.cs]: Test wasn't using the Test App builder, it was using the standard - sorted now :)

File: src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs (hunk 7 above)
Comment (prom3theu5) [reply src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs]: Might just be tired and missing something obvious lol

File: src/Aspire.Hosting.Kubernetes/Resources/Int32OrStringV1.cs [hunk 8]
```diff
@@ -0,0 +1,168 @@
...
+
+/// <summary>
+/// Represents a value that can be either a 32-bit integer or a string.
+/// </summary>
+/// <remarks>
+/// This class provides functionality to handle values that could be either
+/// an integer or a string. It supports implicit and explicit conversions,
+/// equality comparisons, and YAML serialization/deserialization handling.
+/// </remarks>
+public sealed record Int32OrStringV1(int? Number = null, string? Text = null) : IEquatable<int>, IEquatable<string>
```
davidfowl, line 16: DU! 😄 As a follow up I would just use object if this was internal storage.

davidfowl (approved):
Lots of follow ups but great start!

src/Aspire.Hosting.Kubernetes/KubernetesResourceContext.cs
davidfowl: Yea, you shouldn't need to do anything special to make it work.

prom3theu5: Didn't realise the EnvVar had a reference expression, missed that yesterday ^^
Definitely will be a follow up on this some time this week 

I opted to not use port allocator here purely because ClusterIP this wouldn't be an issue - its only if we had a specific NodePort binding to the same port we'd run into an issue, but two pods in two separate deployments can use the same port no problem

PR #8265, opened 2025-03-24
Clean up build scripts

RussKie:
* Remove remnants of test coverage infra
* Do not force to restore/build for test flows
  image

### DIFF
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -9,10 +9,6 @@ Param(
   [switch]$testnobuild,
   [ValidateSet("x86","x64","arm","arm64")][string[]][Alias('a')]$arch = @([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant()),
 
-  # Run tests with code coverage
-  [Parameter(ParameterSetName='CommandLine')]
-  [switch] $testCoverage,
-
   [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
 )
 
@@ -43,7 +39,6 @@ function Get-Help() {
   Write-Host "  -sign                   Sign build outputs."
   Write-Host "  -test (-t)              Incrementally builds and runs tests."
   Write-Host "                          Use in conjunction with -testnobuild to only run tests."
-  Write-Host "  -testCoverage           Run unit tests and capture code coverage information."
   Write-Host ""
 
   Write-Host "Libraries settings:"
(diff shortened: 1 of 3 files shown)


eng/build.ps1
```diff
@@ -9,10 +9,6 @@ Param(
   [switch]$testnobuild,
   [ValidateSet("x86","x64","arm","arm64")][string[]][Alias('a')]$arch = @([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant()),
 
-  # Run tests with code coverage
-  [Parameter(ParameterSetName='CommandLine')]
-  [switch] $testCoverage,
```
radical, line 14: Won't these two build scripts get overwritten by the next Arcade update?

radical:
> * Do not force to restore/build for test flows
>   image

I don't understand the change for this. Why is this command line expected to work? It is trying to build an arbitrary `.proj` file, and there shouldn't be any corresponding tests found.
Also, to skip build for tests you can pass `-testnobuild` which `eng/build.sh` has but doesn't seem to be doing anything in `eng/build.ps1. Maybe we should fix that.

RussKie:
> I don't understand the change for this. Why is this command line expected to work? It is trying to build an arbitrary `.proj` file, and there shouldn't be any corresponding tests found.

The current implementation will forcefully run both `restore` and `build` when `test` operation is being executed. This is wrong, we must be able to only execute tests without re-restoring/rebuilding the solution.

```diff
diff --git a/eng/build.ps1 b/eng/build.ps1
index 7a1670e01..7a36b2762 100644
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -77,13 +77,15 @@ if ($vs) {
 
 if (!$actionPassedIn) {
+  Write-Error "No action specified. Please specify one of the following: $actions"
+  exit 1
   $arguments = "-restore -build"
 }
 ```

```powershell
PS D:\Development\dotnet-aspire> .\build.cmd -test
D:\Development\dotnet-aspire\eng\build.ps1 : No action specified. Please specify one of the following: b build r restore rebuild sign testnobuild publish clean
At line:1 char:1
+ & 'D:\Development\dotnet-aspire\eng\build.ps1' -test
```

> Also, to skip build for tests you can pass `-testnobuild` which `eng/build.sh` has but doesn't seem to be doing anything in `eng/build.ps1. Maybe we should fix that.

This is an orthogonal issue.

radical (approved):

RussKie:
> It looks like this change broke the build as we missed updating the AzDO pipelines (which we still use for rolling and internal builds) to not use code coverage during its invocation. I'm temporarily reverting this to unblock the build.
> 
> @RussKie thanks for working on this! When you have time, can you please work on cleaning up the AzDO pipelines yamls as well so we can take the full change please?

I was overzealous in cleaning dotnet-tools.json file and deleted "dotnet-coverage" declaration. It needs to be reinstated for everything to work:

image

RussKie:
This also highlights the gap in our gates - why are we not running the same tests in AzDO for pull-requests?

/cc: @radical

radical:
> This also highlights the gap in our gates - why are we not running the same tests in AzDO for pull-requests?
> 
> 
> 
> /cc: @radical 

We are doing this to have faster CI times for PRs. Once we have fast test run times on our azdo pipeline line too, we could change this.
This does leave the gap, but the reasoning is that faster CI is more useful than the azdo pipeline sometimes breaking.

We could improve this by using matrix jobs on the azdo pipeline to run tests in parallel, instead of using helix.

For the short term, if you are making infrastructure changes then it is useful to run the pipeline manually for your branch. It would be even better to enable this with "/azp run".

PR #8266, opened 2025-03-24
Add Cosmos database builder pattern for child containers

captainsafia:
This PR introduces new APIs to the Azure Cosmos integration to support interacting with multiple containers against the same database. Our goal is to reuse client configuration for multiple containers. Our current set of APIs makes it possible to reuse client information if there is always one database -> one container.

```csharp
builder.AddAzureCosmosDatabase("db");
builder.AddAzureCosmosContainer("entries"); // Uses the same client that initialized "db"
```

This falls apart when there are multiple containers that are uniquely identified by their service key. In this case, we can assume that all containers share the same database but that might not always be the case.

```csharp
builder.AddAzureCosmosDatabase("cosmos1");
builder.AddAzureCosmosDatabase("cosmos2");
// Which database do these container have a relationship with?
builder.AddKeyedAzureCosmosContainer("users");
builder.AddKeyedAzureCosmosContainer("products");
builder.AddKeyedAzureCosmosContainer("orders");
```

The `WithCosmosDatabaseBuilder` API introduced here makes it possible to chain the database->container relationships:

```csharp
builder.WithAzureCosmosDatabase("cosmos1")
    .AddContainer("users")
    .AddContainer("orders");
builder.WithAzureCosmosDatabase("cosmos2")
    .AddContainer("products");
```

In application code, these services can be resolved as follows:

```
app.MapGet("/users", ([FromKeyedServices("users")] Container container) => { });
app.MapGet("/orders", ([FromKeyedServices("orders")] Container container) => { });
app.MapGet("/products", ([FromKeyedServices("products")] Container container) => { });
```

### DIFF
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
@@ -8,9 +8,9 @@ using Newtonsoft.Json;
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
-builder.WithAzureCosmosDatabase("db")
-    .AddContainer("entries")
-    .AddContainer("users");
+builder.AddAzureCosmosDatabase("db")
+    .AddKeyedContainer("entries")
+    .AddKeyedContainer("users");
 builder.AddCosmosDbContext<TestCosmosContext>("db", configureDbContextOptions =>
 {
     configureDbContextOptions.RequestTimeout = TimeSpan.FromSeconds(120);
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
@@ -8,8 +8,9 @@ using Newtonsoft.Json;
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
-builder.AddAzureCosmosDatabase("db");
-builder.AddAzureCosmosContainer("entries");
+builder.WithAzureCosmosDatabase("db")
+    .AddContainer("entries")
+    .AddContainer("users");
 builder.AddCosmosDbContext<TestCosmosContext>("db", configureDbContextOptions =>
 {
     configureDbContextOptions.RequestTimeout = TimeSpan.FromSeconds(120);
@@ -18,14 +19,13 @@ builder.AddCosmosDbContext<TestCosmosContext>("db", configureDbContextOptions =>
 var app = builder.Build();
 
 app.MapDefaultEndpoints();
-app.MapGet("/", async (Database db, Container container) =>
+
+static async Task<object> AddAndGetStatus<T>(Container container, T newEntry)
 {
-    // Add an entry to the database on each request.
-    var newEntry = new Entry() { Id = Guid.NewGuid().ToString() };
     await container.CreateItemAsync(newEntry);
 
-    var entries = new List<Entry>();
-    var iterator = container.GetItemQueryIterator<Entry>(requestOptions: new QueryRequestOptions() { MaxItemCount = 5 });
+    var entries = new List<T>();
+    var iterator = container.GetItemQueryIterator<T>(requestOptions: new QueryRequestOptions() { MaxItemCount = 5 });
 
     var batchCount = 0;
     while (iterator.HasMoreResults)
@@ -40,10 +40,22 @@ app.MapGet("/", async (Database db, Container container) =>
 
     return new
     {
-        batchCount = batchCount,
+        batchCount,
         totalEntries = entries.Count,
-        entries = entries
+        entries
     };
+}
+
+app.MapGet("/", async ([FromKeyedServices("entries")] Container container) =>
+{
+    var newEntry = new Entry() { Id = Guid.NewGuid().ToString() };
+    return await AddAndGetStatus(container, newEntry);
+});
+
+app.MapGet("/users", async ([FromKeyedServices("users")] Container container) =>
+{
+    var newEntry = new User() { Id = $"user-{Guid.NewGuid()}" };
+    return await AddAndGetStatus(container, newEntry);
 });
 
 app.MapGet("/ef", async (TestCosmosContext context) =>
@@ -57,6 +69,12 @@ app.MapGet("/ef", async (TestCosmosContext context) =>
 });
 
 app.Run();
+
+public class User
+{
+    [JsonProperty("id")]
+    public string? Id { get; set; }
+}
 
 public class Entry
 {
--- a/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
+++ b/playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
@@ -9,14 +9,13 @@ var cosmos = builder.AddAzureCosmosDB("cosmos")
                 .RunAsPreviewEmulator(e => e.WithDataExplorer());
 
 var db = cosmos.AddCosmosDatabase("db");
-var entries = db.AddContainer("entries", "/id");
-var users = db.AddContainer("users", "/id");
+var entries = db.AddContainer("entries", "/id", "staging-entries");
+db.AddContainer("users", "/id");
 
 builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
        .WithReference(db).WaitFor(db)
-       .WithReference(entries).WaitFor(entries)
-       .WithReference(users).WaitFor(users);
+       .WithReference(entries).WaitFor(entries);
 
 #if !SKIP_DASHBOARD_REFERENCE
 // This project is only added in playground projects to support development/debugging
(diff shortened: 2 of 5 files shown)


playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
```diff
@@ -9,12 +9,14 @@
                 .RunAsPreviewEmulator(e => e.WithDataExplorer());
 
 var db = cosmos.AddCosmosDatabase("db");
-var container = db.AddContainer("entries", "/id");
+var entries = db.AddContainer("entries", "/id");
+var users = db.AddContainer("users", "/id");
```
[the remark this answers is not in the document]
captainsafia: Interesting obsevation, Copilot! We're introducing an `AddContainer` API in the client integration that might cause people to be confused by the `AddContainer` API that we use in the hosting side.

I like the name `AddContainer` though and would like to avoid dismabiguating between the two with a different name. Maybe `WithContainer`? But I like the way `With CosmosDatabase Add Container` reads from a natural language perspective.

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
```diff
@@ -9,12 +9,14 @@
                 .RunAsPreviewEmulator(e => e.WithDataExplorer());
 
 var db = cosmos.AddCosmosDatabase("db");
-var container = db.AddContainer("entries", "/id");
+var entries = db.AddContainer("entries", "/id");
```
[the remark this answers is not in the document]
captainsafia: This is a dupe of your other comment.

playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
```diff
@@ -8,8 +8,9 @@
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
-builder.AddAzureCosmosDatabase("db");
-builder.AddAzureCosmosContainer("entries");
+builder.WithAzureCosmosDatabase("db")
+    .AddContainer("entries")
```
eerhardt, line 12: I wonder if this should be called `WithKeyedContainer`.

1. We use "With" when you return the same builder, "Add" when you return the new thing being added.
2. Adding "Keyed" makes it obvious that the service will be keyed.

eerhardt: And the root method should be `AddAzureCosmosDatabase`.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -197,7 +197,46 @@ public static void AddKeyedAzureCosmosContainer(
...
+    /// This method can be called multiple times to register multiple containers against the same database.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder" /> to read config from and add services to.</param>
+    /// <param name="name">The name of the component, which is used as the <see cref="ServiceDescriptor.ServiceKey"/> of the service and also to retrieve the connection string from the ConnectionStrings configuration section.</param>
+    /// <returns>A <see cref="CosmosDatabaseBuilder"/> that can be used for further chaining.</returns>
+    /// <remarks>
+    /// Containers registered with this method will always use the account and database associated
+    /// with the parent <see cref="CosmosDatabaseBuilder"/>. The connection string for the container.
+    /// </remarks>
+    public static CosmosDatabaseBuilder AddContainer(
```
eerhardt, line 232: Is there a reason this isn't just a normal method on the CosmosDatabaseBuilder class?

playground/CosmosEndToEnd/CosmosEndToEnd.ApiService/Program.cs
captainsafia: > And the root method should be AddAzureCosmosDatabase.

I like the prefix `With` for this API because it gives a better clue that you're registering containers within the context of a specific database. 

Also, with:

> "Add" when you return the new thing being added.

The `WithAzureCosmosDatabase` doesn't register any services currently so using the `Add` prefix would violate this convention.

eerhardt: My thinking was that we'd have:

```C#
builder.AddAzureCosmosDatabase("db")
    .WithKeyedContainer("entries")
    .WithKeyedContainer("users");
```

And it would register 3 services:

1. The unkeyed Cosmos `Database`.
2. Two keyed Cosmos `Container` services.

And if I want to register keyed Databases, I would use:

```C#
builder.AddKeyedAzureCosmosDatabase("db")
    .WithKeyedContainer("entries")
    .WithKeyedContainer("users");
```

captainsafia: > And it would register 3 services:

> The unkeyed Cosmos Database.
> Two keyed Cosmos Container services.

I find this a little confusing given that our convention in the other types has been that they only register the service that they are associated with. For example, currently when you call `AddKeyedAzureCosmosDatabase`, you don't also get a `CosmosClient` registered as well.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
captainsafia: I was prematurely optimizing in the event that we wanted to change the signature of the inner method but I'm not gung-ho about this.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -197,7 +197,46 @@ public static void AddKeyedAzureCosmosContainer(
...
+    /// Returns a <see cref="CosmosDatabaseBuilder"/> for the given connection name that can be used to
+    /// register multiple containers against the same database.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder" /> to read config from and add services to.</param>
+    /// <param name="connectionName">The connection name to use to find a connection string.</param>
+    /// <param name="configureSettings">An optional method that can be used for customizing the <see cref="MicrosoftAzureCosmosSettings"/>. It's invoked after the settings are read from the configuration.</param>
+    /// <param name="configureClientOptions">An optional method that can be used for customizing the <see cref="CosmosClientOptions"/>.</param>
+    /// <remarks>Reads the configuration from "Aspire:Microsoft:Azure:Cosmos:{name}" section.</remarks>
+    /// <exception cref="InvalidOperationException">If required ConnectionString is not provided in configuration section</exception>
+    public static CosmosDatabaseBuilder WithAzureCosmosDatabase(
```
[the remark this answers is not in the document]
captainsafia: I had two thoughts here:

1. I'd like to maintain a distinction between methods that actually register a database service into the DI container vs. ones that reuse the same client for a given context.
2. Keeping the two methods felt like a good way to track usage and obsolete/deprecate a given pattern based on what we see here.

captainsafia:
@davidfowl @eerhardt Ping for review on latest API.

src/Components/Aspire.Microsoft.Azure.Cosmos/AspireMicrosoftAzureCosmosExtensions.cs
```diff
@@ -132,72 +94,80 @@ public static void AddKeyedAzureCosmosClient(
...
     /// <param name="configureSettings">An optional method that can be used for customizing the <see cref="MicrosoftAzureCosmosSettings"/>. It's invoked after the settings are read from the configuration.</param>
     /// <param name="configureClientOptions">An optional method that can be used for customizing the <see cref="CosmosClientOptions"/>.</param>
     /// <remarks>Reads the configuration from "Aspire:Microsoft:Azure:Cosmos:{name}" section.</remarks>
     /// <exception cref="InvalidOperationException">If required ConnectionString is not provided in configuration section</exception>
-    public static void AddKeyedAzureCosmosDatabase(
-       this IHostApplicationBuilder builder,
-       string name,
-       Action<MicrosoftAzureCosmosSettings>? configureSettings = null,
-       Action<CosmosClientOptions>? configureClientOptions = null)
+    public static void AddKeyedAzureCosmosContainer(
```
eerhardt, line 105: Do we still need/want these direct "AddContainer" APIs? Or is it good enough to just have the new AddCosmosDatabase + builder APIs?

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
```diff
@@ -9,12 +9,14 @@
...
+var users = db.AddContainer("users", "/id");
 
 builder.AddProject<Projects.CosmosEndToEnd_ApiService>("api")
        .WithExternalHttpEndpoints()
        .WithReference(db).WaitFor(db)
-       .WithReference(container).WaitFor(container);
+       .WithReference(entries).WaitFor(entries)
+       .WithReference(users).WaitFor(users);
```
eerhardt, line 19: You really only need to `.WithReference(db).WaitFor(db);`. The ApiService doesn't use the `entries` and `users` references, so we can remove them here.

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/Program.cs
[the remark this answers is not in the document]
eerhardt: The way we've been disambiguating between the hosting APIs and the client APIs is that the client APIs always end in `Client`. e.g. `AddAzureCosmosClient`.

Maybe we should follow that same naming pattern in the new APIs we are introducing here. I should have caught this in the previous PR.

eerhardt (approved):
LGTM. Thanks!

PR #8272, opened 2025-03-24
Only add .msi in Signing.props when missing

akoeplinger:
arcade 10.0 which is used in the VMR build added .msi to FileExtensionSignInfo so the aspire build now complains about a duplicate entry. Only add the entry if it's not already there.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -3,7 +3,9 @@
   <ItemGroup>
     <FileExtensionSignInfo Update=".nupkg" CertificateName="NuGet" />
     <FileExtensionSignInfo Update=".zip" CertificateName="None" />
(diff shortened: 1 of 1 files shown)


radical (approved):

akoeplinger:
I know, but we could in theory also apply that patch in the VMR directly when we switch to flat flow if aspire isn't going to be updated until it is removed.

ViktorHofer:
Good point. Could we already do that now?

akoeplinger:
hmm not sure if the sync would break, we should ask Premek

joperezr:
Sorry for the delay, but as said above the plan is to not update it in the SDK anymore until it is removed, so while technically you can add this change to the branch that sdk was pulling from, the intent is to never ship nor build from there again.

PR #8276, opened 2025-03-24
Introduce CommandOptions & HttpCommandOptions

DamianEdwards:
## Description

React to feedback on API review: https://github.com/dotnet/aspire/pull/7811#discussion_r2010581063

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] Yes
      - [x] Yes
      - [x] Yes
  - [x] No
  - [x] Yes
      - [x] Yes
        - https://github.com/dotnet/docs-aspire/issues/2888

### DIFF
--- a/src/Aspire.Hosting/ApplicationModel/HttpCommandOptions.cs
+++ b/src/Aspire.Hosting/ApplicationModel/HttpCommandOptions.cs
@@ -1,0 +1,42 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>
+/// Optional configuration for resource HTTP commands added with <see cref="ResourceBuilderExtensions.WithHttpCommand{TResource}(IResourceBuilder{TResource}, string, string, string?, HttpCommandOptions?)"/>."/>
+/// </summary>
+public class HttpCommandOptions : CommandOptions
+{
+    internal static new HttpCommandOptions Default { get; } = new HttpCommandOptions();
+
+    /// <summary>
+    /// Gets or sets a callback that selects the HTTP endpoint to send the request to when the command is invoked.
+    /// </summary>
+    public Func<EndpointReference>? EndpointSelector { get; init; }
+
+    /// <summary>
+    /// Gets or sets the HTTP method to use when sending the request. Defaults to <c>POST</c>.
+    /// </summary>
+    public HttpMethod Method { get; init; } = HttpMethod.Post;
+
+    /// <summary>
+    /// Gets or sets the name of the HTTP client to use when creating it via <see cref="IHttpClientFactory.CreateClient(string)"/>.
+    /// </summary>
+    public string? HttpClientName { get; init; }
+
+    /// <summary>
+    /// Gets or sets a callback to be invoked to configure the request before it is sent.
+    /// </summary>
+    public Func<HttpCommandRequestContext, Task>? PrepareRequest { get; init; }
+
+    /// <summary>
+    /// Gets or sets a callback to be invoked after the response is received to determine the result of the command invocation.
+    /// </summary>
+    public Func<HttpCommandResultContext, Task<ExecuteCommandResult>>? GetCommandResult { get; init; }
+
+    /// <summary>
+    /// Gets or sets the name of command. The name uniquely identifies the command.
+    /// </summary>
+    public string? CommandName { get; init; }
+}
--- a/src/Aspire.Hosting/ApplicationModel/HttpCommandOptions.cs
+++ b/src/Aspire.Hosting/ApplicationModel/HttpCommandOptions.cs
@@ -8,35 +8,35 @@ namespace Aspire.Hosting.ApplicationModel;
 /// </summary>
 public class HttpCommandOptions : CommandOptions
 {
-    internal static new HttpCommandOptions Default { get; } = new HttpCommandOptions();
+    internal static new HttpCommandOptions Default { get; } = new();
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


src/Aspire.Hosting/ApplicationModel/HttpCommandOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// <summary>
+    /// Gets or sets a callback to be invoked after the response is received to determine the result of the command invocation.
+    /// </summary>
+    public Func<HttpCommandResultContext, Task<ExecuteCommandResult>>? GetCommandResult { get; set; }
+
+    /// <summary>
+    /// Gets or sets the name of command. The name uniquely identifies the command.
+    /// </summary>
+    public string? CommandName { get; set; }
```
JamesNK, line 41: WithCommand already has a name. How does this interact with it? Also, why is command name specific to HttpCommandOptions? Shouldn't it be on the base CommandOptions?

DamianEdwards: `CommandOptions` is for things that are optional as it's replacing all the optional parameters. You still need to pass a name to `WithCommand`. In the case of `WithHttpCommand`, a name is not required as a default will be generated for you, so it's optional on `HttpCommandOptions`.

JamesNK: It feels weird that there are different ways for doing it for other commands and HTTP commands.

Maybe the name parameter is optional on WithHttpCommand. If you want to specify it then provide a value for the optional parameter.

DamianEdwards: Sounds good.

DamianEdwards: @JamesNK updated.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1338,6 +1338,7 @@ public static IResourceBuilder<T> WithCommand<T>(
     /// <param name="path">The path to send the request to when the command is invoked.</param>
     /// <param name="displayName">The display name visible in UI.</param>
     /// <param name="endpointName">The name of the HTTP endpoint on this resource to send the request to when the command is invoked.</param>
+    /// <param name="commandName">The name of command. The name uniquely identifies the command.</param>
```
JamesNK, line 1341:
```suggestion
    /// <param name="commandName">Optional name of the command. The name uniquely identifies the command. If a name isn't specified then the name is inferred using the command's endpoint and HTTP method.</param>
```

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1227,7 +1228,49 @@ public static IResourceBuilder<T> WithHttpsHealthCheck<T>(this IResourceBuilder<
...
+        commandOptions ??= CommandOptions.Default;
+
+        // Replace existing annotation with the same name.
+        var existingAnnotation = builder.Resource.Annotations.OfType<ResourceCommandAnnotation>().SingleOrDefault(a => a.Name == name);
+        if (existingAnnotation is not null)
+        {
+            builder.Resource.Annotations.Remove(existingAnnotation);
+        }
+
+        return builder.WithAnnotation(new ResourceCommandAnnotation(name, displayName, commandOptions.UpdateState ?? (c => ResourceCommandState.Enabled), executeCommand, commandOptions.Description, commandOptions.Parameter, commandOptions.ConfirmationMessage, commandOptions.IconName, commandOptions.IconVariant, commandOptions.IsHighlighted));
```
mitchdenny, line 1265: `commandOPtions.UpdateState ?? CommandOptions.Default.UpdateState` ?

mitchdenny (approved):
LGTM

src/Aspire.Hosting/ResourceBuilderExtensions.cs
```diff
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Data;
```
IEvangelist, line 4: Why is this namespace needed?

IEvangelist (approved):

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: `CommandOptions.Default.UpdateState` is `null` by default so that it can be detected whether one was set or not.

src/Aspire.Hosting/ResourceBuilderExtensions.cs
DamianEdwards: Not sure but the build would fail if it wasn't required.

DamianEdwards: Not sure why the analyzer isn't picking this one up but you're right, it isn't needed.

DamianEdwards:
@IEvangelist RE the pattern-matching suggestion, it can't be used with `KnownResourceStates` as pattern matching requires constant values when comparing to strings.

IEvangelist:
> @IEvangelist RE the pattern-matching suggestion, it can't be used with `KnownResourceStates` as pattern matching requires constant values when comparing to strings.

@DamianEdwards Ah, ok! I that these were an `enum`...that's the `KnownResourceState` type I was thinking, which caters nicely to the compile time const constraint on the pattern matching suggestion. I'm seeing that `KnownResourceStates.*` fields as `static readonly` instead of `const`, which would allow for pattern matching.

PR #8280, opened 2025-03-25
Remove the need for AppliedRoleAssignmentsAnnotation

### LINKED ISSUE #6161: Improve Azure Managed Identity Defaults

eerhardt:
## Description

Instead of generating the role assignments inline in the resource's bicep, always generate the role assignments in a separate file - even during local provisioning and when not using targeted role assignments.

Fix #6161

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No


src/Aspire.Hosting.Azure/AzureResourcePreparer.cs
```diff
@@ -376,4 +381,62 @@ private static void AppendAppliedRoleAssignmentsAnnotation(IResource resource, I
...
         }
     }
+
+    private static void CreateGlobalRoleAssignments(DistributedApplicationModel appModel, AzureProvisioningOptions provisioningOptions)
+    {
+        var azureResourceSnapshot = appModel.Resources.OfType<AzureProvisioningResource>().ToArray(); // avoid modifying the collection while iterating
+
+        foreach (var azureResource in azureResourceSnapshot)
+        {
+            if (azureResource.TryGetLastAnnotation<AppliedRoleAssignmentsAnnotation>(out var appliedRoleAssignments))
```
eerhardt, line 391: With this change, we can make the `AppliedRoleAssignmentsAnnotation` internal, and rename it to anything we want.

Right now the annotaiton is still needed to support the case where you have 2 apps targeting an Azure resource, and one app requires role X and the other app requires role Y. Here, we will generate a single file `{azureResource}-roles` with all the specified roles. The annotation is used to gather all the targeted roles in the first step, and then later during `CreateGlobalRoleAssignments`, use the annotations to generate the bicep.

The other option to support this scenario would be to generate 2 role files `{app1}-roles-{azureResource}` and `{app2}-roles-{azureResource}`.

eerhardt: Rethinking this, I'm not sure we should make this change. I think it may be too breaking since someone could be using ConfigureInfrastructure on their Azure resource, and moving these role assignments out to a different bicep module would mean their code would no longer work.

davidfowl: Does this mean that every new azure resource also needs this logic forever? When we do remove it?

eerhardt: Moving forward with this change means that every new azure resource won't need this logic. The `AppliedRoleAssignmentsAnnotation` will become internal, and can eventually be removed on more refactoring.

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
```diff
@@ -1005,9 +751,7 @@ public async Task SupportsExistingAzureSignalRWithResourceGroup()
               "connectionString": "Endpoint=https://{signalR.outputs.hostName};AuthType=azure",
```
captainsafia, line 751: Can we add a snapshot test for the roles file in this case? I don't think we need to set a scope on the roles bicep (since we're declaring the role assignments in the same resource group as the app) but we might want to verify that in a test.

src/Aspire.Hosting.Azure/Provisioning/Provisioners/BicepProvisioner.cs
```diff
@@ -94,7 +94,7 @@ await notificationService.PublishUpdateAsync(resource, state =>
 
             return state with
             {
-                State = new("Running", KnownResourceStateStyles.Success),
+                State = new("Provisioned", KnownResourceStateStyles.Success),
```
captainsafia, line 97: I see the following states while a ServiceBus is being provisioned:

image

Seems weird that the top-level resource is provisioned but the underlying resources are pending and starting?

davidfowl (approved):
Brings tears to my eyes

tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs
eerhardt: Opened [Add test to ensure the scope is set correctly on an existing resource's role assignments bicep. (dotnet/aspire#8353)](https://github.com/dotnet/aspire/pull/8353) for this.

> I don't think we need to set a scope on the roles bicep (since we're declaring the role assignments in the same resource group as the app) but we might want to verify that in a test.

We do need to set a scope on the roles bicep, since we need the `existing` resource in order to add role assignments to it.

So just like before, where the role assignments were in the same bicep module as the "existing" resource, with the scope set, we need to set the scope on the new resource as well.

PR #8171, opened 2025-03-19
Introduce AzureKeyVaultSecretReference

davidfowl:
## Description

- This is a replacement for the magic secret outputs implementation. Resources that support keys push those keys into an explicitly defined per resource key vault and now into any key vault that is provided. The IKeyVaultSecretReference is a new primitive reference type that can is understood natively by compute environments like ACA and the azure preparer.

Fixes #8134

## Checklist

- Is this feature complete?
  - [ ] Yes. Ready to ship.
  - [x] No. Follow-up changes expected.
- Are you including unit tests for the changes and scenario tests if relevant?
  - [ ] Yes
  - [x] No
- Did you add public API?
  - [x] Yes
    - If yes, did you have an API Review for it?
      - [ ] Yes
      - [ ] No
    - Did you add `<remarks />` and `<code />` elements on your triple slash comments?
      - [ ] Yes
      - [ ] No
  - [ ] No
- Does the change make any security assumptions or guarantees?
  - [x] Yes
    - If yes, have you done a threat model and had a security review?
      - [ ] Yes
      - [x] No
  - [ ] No
- Does the change require an update in our Aspire docs?
  - [x] Yes
    - Is this introducing a breaking change?
      - [x] Yes
        - Link to aspire-docs issue (please use this [`breaking-change` template](https://github.com/dotnet/docs-aspire/issues/new?template=04-breaking-change.yml)): https://github.com/dotnet/docs-aspire/issues/2889
      - [ ] No
        - Link to aspire-docs issue (please use this [`doc-idea` template](https://github.com/dotnet/docs-aspire/issues/new?template=02-docs-request.yml)):
  - [ ] No

PR #8282, opened 2025-03-25
Fix Cosmos child resource connection strings with access key auth

### LINKED ISSUE #8274: CosmosDB database and container connection strings need to be in secrets when the CosmosDB is using AccessKeyAuthentication
When we started supporting connection strings to the child Database and Container resources in CosmosDB (#8068 and #8236), we generate the connection string expression incorrectly when `UseAccessKeyAuthentication` is used.

https://github.com/dotnet/aspire/blob/db8ebf28c53e4c423287aeda3218e901d252ed20/src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs#L135-L154

When using AccessKeyAuth, the whole connection string including the children portions needs to go into the KeyVault secret. Look at PostgreSQL for an example:

https://github.com/dotnet/aspire/blob/db8ebf28c53e4c423287aeda3218e901d252ed20/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresExtensions.cs#L420-L443

https://github.com/dotnet/aspire/blob/db8ebf28c53e4c423287aeda3218e901d252ed20/src/Aspire.Hosting.Azure.PostgreSQL/AzurePostgresFlexibleServerResource.cs#L96-L108

captainsafia:
Closes https://github.com/dotnet/aspire/issues/8274.


davidfowl:
Wait until https://github.com/dotnet/aspire/pull/8171/files goes in

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -136,7 +136,14 @@ internal ReferenceExpression GetConnectionString(string? databaseName = null, st
 
         if (IsEmulator || UseAccessKeyAuthentication)
         {
-            builder.AppendFormatted(ConnectionStringExpression);
+            if (databaseName is null)
+            {
+                builder.Append($"{new BicepSecretOutputReference(GetKeyValueSecretName(databaseName, containerName), this)}");
```
davidfowl, line 141: This should be using the new keyvault reference. The new format is connectionstrings--{resourcename}. Don't add any hierearchy, just use resource names.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -136,7 +136,14 @@ internal ReferenceExpression GetConnectionString(string? databaseName = null, st
...
         {
-            builder.AppendFormatted(ConnectionStringExpression);
+            if (ConnectionStringSecretOutput is not null)
+            {
+                builder.Append($"{ConnectionStringSecretOutput.Resource.GetSecretReference(GetKeyValueSecretName(resourceName))}");
+            }
+            else
+            {
+                builder.AppendFormatted(ConnectionStringExpression);
+            }
```
eerhardt, line 146: It probably doesn't make sense to share the `if (IsEmulator || UseAccessKeyAuthentication)` check anymore, since we then check for `if (ConnectionStringSecretOutput is not null)` inside.

Also, if `UseAccessKeyAuthentication == true`, we just want to use the ConnectionStringSecretOutput and not append `;Database=xx` and `;Container=xx` to the connection string.

So I'd write this code as something like:

```C#
        if (UseAccessKeyAuthentication)
        {
            builder.Append($"{ConnectionStringSecretOutput.Resource.GetSecretReference(GetKeyValueSecretName(resourceName))}");
        }
        else
        {
            if (IsEmulator)
            {
                builder.AppendFormatted(ConnectionStringExpression);
            }
            else
            {
                builder.Append($"AccountEndpoint={ConnectionStringExpression}");
            }

            if (!string.IsNullOrEmpty(databaseName))
            {
                builder.Append($";Database={databaseName}");

                if (!string.IsNullOrEmpty(containerName))
                {
                    builder.Append($";Container={containerName}");
                }
            }
        }
```

tests/Aspire.Hosting.Azure.Tests/AzureCosmosDBExtensionsTests.cs
```diff
@@ -240,6 +247,120 @@ param principalId string
...
+        var cosmos = builder.AddAzureCosmosDB("cosmos")
+            .WithAccessKeyAuthentication();
+        var database = cosmos.AddCosmosDatabase("db1");
+        var container = database.AddContainer("container1", "id");
+
+        builder.AddContainer("api", "myimage")
+            .WithReference(cosmos);
+
+        Assert.Equal("{cosmos.secretOutputs.connectionString}", cosmos.Resource.ConnectionStringExpression.ValueExpression);
+        Assert.Equal("{cosmos.secretOutputs.connectionString};Database=db1", database.Resource.ConnectionStringExpression.ValueExpression);
```
eerhardt, line 264: This is not correct. We should be using the whole connection string from the secrets.

src/Aspire.Hosting.Azure.CosmosDB/AzureCosmosDBResource.cs
```diff
@@ -125,7 +125,7 @@ public override void AddRoleAssignments(IAddRoleAssignmentsContext roleAssignmen
         AzureCosmosExtensions.AddContributorRoleAssignment(infra, cosmosAccount, principalId);
     }
 
-    internal ReferenceExpression GetConnectionString(string? databaseName = null, string? containerName = null)
+    internal ReferenceExpression GetConnectionString(string resourceName, string? databaseName = null, string? containerName = null)
```
eerhardt, line 128: I wonder if we should rename this to `GetChildConnectionString`. And then `resourceName` could be `childResourceName`.

playground/CosmosEndToEnd/CosmosEndToEnd.AppHost/aspire-manifest.json
```diff
@@ -11,11 +11,11 @@
     },
     "db": {
       "type": "value.v0",
-      "connectionString": "{cosmos.outputs.connectionString}"
+      "connectionString": "AccountEndpoint={cosmos.outputs.connectionString};Database=db"
```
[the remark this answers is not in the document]
captainsafia: Run an update across all the files.

playground/publishers/Publishers.AppHost/api-identity.module.bicep
```diff
@@ -0,0 +1,15 @@
+@description('The location for the resource(s) to be deployed.')
```
eerhardt, line 1: Maybe let's revert this app. It currently its set up to generate a manifest through `eng\refreshManifests.ps1` since it doesn't have a generate-manifests launch profile.

And all this bicep is duped in the `infra` folder.

eerhardt (approved):
This looks good. I just had 2 comments/thoughts.

playground/publishers/Publishers.AppHost/api-identity.module.bicep
captainsafia: That explains the huge delta here. I originally thought we weren't proactive about updating the manifests here. Yet another case for https://github.com/dotnet/aspire/issues/8292.

PR #8290, opened 2025-03-25
Fix Azure Functions Role Assignments to Host Storage

eerhardt:
## Description

With #8127, a mistake was made in that WithRoleAssignments only works if the app is using ACA infrastructure. If the app isn't using our ACA infrastructure, WithRoleAssignments throws an exception at startup.

To make Azure Functions work correctly both with and without ACA infrastructure we use the internal WithDefaultRoleAssignments method to set the default role assignments on the Host Storage, which will be respected for both.

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] No

### DIFF
--- a/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
+++ b/src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
@@ -46,7 +46,19 @@ public static class AzureFunctionsProjectResourceExtensions
             .OfType<AzureStorageResource>()
             .FirstOrDefault(r => r.Name == storageResourceName);
 
-        storage ??= builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
+        if (storage is null)
+        {
+            storage = builder.AddAzureStorage(storageResourceName)
+                // Azure Functions blob triggers require StorageAccountContributor access to the host storage
+                // account when deployed. We assign this role to the implicit host storage resource.
+                .WithDefaultRoleAssignments(StorageBuiltInRole.GetBuiltInRoleName,
+                    StorageBuiltInRole.StorageBlobDataContributor,
+                    StorageBuiltInRole.StorageTableDataContributor,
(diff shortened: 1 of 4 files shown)


src/Aspire.Hosting.Azure.Functions/AzureFunctionsProjectResourceExtensions.cs
```diff
@@ -46,7 +46,19 @@ public static IResourceBuilder<AzureFunctionsProjectResource> AddAzureFunctionsP
             .OfType<AzureStorageResource>()
             .FirstOrDefault(r => r.Name == storageResourceName);
 
-        storage ??= builder.AddAzureStorage(storageResourceName).RunAsEmulator().Resource;
+        if (storage is null)
```
captainsafia, line 49: Nice!

tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs
```diff
@@ -312,21 +312,21 @@ param principalId string
               scope: funcstorage634f8
```
captainsafia, line 312: Would setting `skipPreparer: false` in https://github.com/dotnet/aspire/blob/f5e596bb038ac396487da4350f78223ebdb9dd26/tests/Aspire.Hosting.Azure.Tests/AzureFunctionsTests.cs#L631 have helped catch this?

eerhardt: It might have. But it wouldn't have been able to test what we needed it to, because these tests are targeting `WithAddAzureContainerAppsInfrastructure`, which needed to Execute all the startup hooks, not just the AzurePreparer.

eerhardt: I've added a test that would have caught this bug.

captainsafia (approved):
Thanks for adding the test!

PR #8302, opened 2025-03-25
Clean up build scripts

RussKie:
* Remove remnants of test coverage infra
* Do not force to restore/build for test flows


RussKie:
Take #2, re-reverting #8265, which accidentally deleted dotnet-coverage .NET tool declaration.

PR #8307, opened 2025-03-26
Show endpoints in `aspire run` table as the localhost addresses

### LINKED ISSUE #8205: Endpoints shown by `aspire run` should be localhost addresses, not codespaces URIs (when in codespaces)
The only reason we include the dashboard URL with the codespace URL is because the redirection logic strips off the login token for the dashboard URL. This isn't an issue for the other endpoints which don't need to retain this.

mitchdenny:
Switch to displaying the localhost URLs in the endpoints column in the `aspire run` CLI command.

Fixes #8205 

image

### DIFF
--- a/src/Aspire.Cli/Backchannel/AppHostBackchannel.cs
+++ b/src/Aspire.Cli/Backchannel/AppHostBackchannel.cs
@@ -42,13 +42,13 @@ internal sealed class AppHostBackchannel(ILogger<AppHostBackchannel> logger, Cli
         return url;
     }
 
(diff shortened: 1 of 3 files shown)


maddymontaquila:
I'm assuming we'll leave the friendly name URL stuff that Damian worked on for the proper dashboard right?

maddymontaquila (approved):

mitchdenny:
Yes. There are two things driving this. The first is just addressing some visual noise, but the second is an issue with the way that Codespaces/Devcontainers trigger port forwarding. This makes it more reliable.

PR #8310, opened 2025-03-26
Add Xunit.Abstractions as global using for tests

Youssef1313:
This will make the diff less noisy for the migration to xunit.v3, and this PR on itself is still easy and straightforward to review.

### DIFF
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -19,6 +19,7 @@
 
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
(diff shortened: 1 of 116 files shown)


tests/Directory.Build.props
```diff
@@ -19,6 +19,7 @@
 
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
+    <Using Include="Xunit.Abstractions" />
```
RussKie, line 22: Please place it in own ItemGroup to increase maintainability

Youssef1313: Done

danmoseley (approved):

PR #8323, opened 2025-03-26
Obsolete EventHubs Emulator WithDataBindMount and WithDataVolume

### LINKED ISSUE #8008: Azure EventHubs Emulator WithDataBindMount doesn't do anything
We have WithDataBindMount and WithDataVolume on the EventHubs emulator:

https://github.com/dotnet/aspire/blob/3b08e15ccc3b1fa37c211772b299929d45a17176/src/Aspire.Hosting.Azure.EventHubs/AzureEventHubsExtensions.cs#L366-L390

However, it appears that this folder isn't where the EventHubs emulator writes persistent data to. Looking at the docs:

https://learn.microsoft.com/en-us/azure/event-hubs/overview-emulator#known-limitations

>  Note
> After a container restart, data and entities don't persist in the emulator.

We should obsolete these methods if they don't enable a scenario.

See https://github.com/dotnet/aspire/pull/4209/files#r1619294488 for some discussion on this topic.

cc @sebastienros @oising

eerhardt:
These methods don't work correctly, so obsoleting and eventually removing them.

Fix #8008


sebastienros (approved):
Didn't find any docs about them

PR #8333, opened 2025-03-26
[CI] Fix pipeline tests on rolling builds

### LINKED ISSUE #8330: [CI] Pipeline tests not running at all on the rolling builds
None of the pipeline tests are running on the rolling builds.

This broke somewhere after https://dev.azure.com/dnceng-public/public/_build/results?buildId=991969&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76 for PR #8265 .

I will try to fix it in https://github.com/dotnet/aspire/pull/8329 .

cc @RussKie

radical:
Use `$(VSTestNoBuild)` property which is used by vstest instead of
arbitrary `$(TestNoBuild)`. VSTest prevents a build of the test projects
when the property is `true`.

This fixes the running tests on Azdo pipeline builds by not skipping the test projects completely when `-testnobuild` is used.

Fixes https://github.com/dotnet/aspire/issues/8330 .

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Build with packages
         env:
           CI: false
-        run: ./build.sh -restore -build -ci -pack -testnobuild /bl /p:InstallBrowsersForPlaywright=false
+        run: ./build.sh -restore -build -ci -pack /bl /p:InstallBrowsersForPlaywright=false /p:SkipTestProjects=true
 
       - name: Upload built NuGets
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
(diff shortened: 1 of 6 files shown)


radical:
Validated in https://dev.azure.com/dnceng-public/public/_build/results?buildId=995222

RussKie:
> Use `$(VSTestNoBuild)` property which is used by vstest instead of
> arbitrary `$(TestNoBuild)`. VSTest prevents a build of the test projects
> when the property is `true`.

👍 
Sadly, this property doesn't appear to be documented anywhere, but it's part of [`Microsoft.TestPlatform.targets`](https://github.com/microsoft/vstest/blob/782a46231902762286f6958631437d635bfdd249/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets#L17).

RussKie:
I don't believe `-testnobuild` is needed after #8302. Previously running `build.cmd -test` would be equivalent to running `build.cmd -restore -build -test`. Now it will only run tests.

RussKie (approved):

radical:
> I don't believe `-testnobuild` is needed after #8302. Previously running `build.cmd -test` would be equivalent to running `build.cmd -restore -build -test`. Now it will only run tests.

Are you suggesting to drop the parameter completely?

PR #8336, opened 2025-03-27
Change DOTNET_ to ASPIRE_ in config variables

### LINKED ISSUE #8009: change DOTNET_ to ASPIRE_ in config variables
Variables that are unique to Aspire (ie., not things like DOTNET_ROOT) should use ASPIRE_ not DOTNET_ as Aspire is a platform and toolset  for building polyglot apps, this seems more appropriate.

At the moment it seems like that's these - all should have 

DOTNET_ASPIRE_CONTAINER_RUNTIME  // change to ASPIRE_CONTAINER_RUNTIME, etc
DOTNET_ASPIRE_DEPENDENCY_CHECK_TIMEOUT
DOTNET_ASPIRE_SERVICE_STARTUP_WATCH_TIMEOUT
DOTNET_ASPIRE_SHOW_DASHBOARD_RESOURCES
DOTNET_DASHBOARD_CONFIG_FILE_PATH // change to ASPIRE_DASHBOARD_CONFIG_FILE_PATH
DOTNET_DASHBOARD_CORS_ALLOWED_ORIGINS
DOTNET_DASHBOARD_FILE_CONFIG_DIRECTORY
DOTNET_DASHBOARD_FRONTEND_BROWSERTOKEN
DOTNET_DASHBOARD_OTLP_ENDPOINT_URL
DOTNET_DASHBOARD_OTLP_HTTP_ENDPOINT_URL
DOTNET_DASHBOARD_RESOURCE_SERVER_ENDPOINT_URL
DOTNET_DASHBOARD_RESOURCESERVICE_APIKEY
DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS

this needs to be done while still recognizing the old names. I see env vars all over the code - it might be a good time to centralize into some utility class.

incidentally noticed this is missing an underscore
https://github.com/dotnet/aspire/blob/07b7c7465d2f5e3ada7055bf757ab724112d37fe/playground/TestShop/MyFrontend/Program.cs#L13

DamianEdwards:
## Description

Changes from `DOTNET_` to `ASPIRE_` in all environment variable and config key names. In some cases it replaces it with `ASPIRE_`. Old keys still work of course, but new ones are preferred.

Fixes #8009 

## Checklist

- Is this feature complete?
  - [x] Yes. Ready to ship.
  - [x] Yes
- Did you add public API?
  - [x] No
  - [x] No
  - [x] Yes
      - [x] No
        - Link to aspire-docs issue: https://github.com/dotnet/docs-aspire/issues/2901


src/Aspire.ProjectTemplates/templates/aspire-apphost/8.2/Properties/launchSettings.json
```diff
@@ -10,8 +10,8 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
         "DOTNET_ENVIRONMENT": "Development",
-        "DOTNET_DASHBOARD_OTLP_ENDPOINT_URL": "https://localhost:21000",
-        "DOTNET_RESOURCE_SERVICE_ENDPOINT_URL": "https://localhost:22000"
+        "ASPIRE_DASHBOARD_OTLP_ENDPOINT_URL": "https://localhost:21000",
```
eerhardt, line 13: does it make sense to update the old templates? Where the new env var names won't be respected?

src/Aspire.Dashboard/Configuration/PostConfigureDashboardOptions.cs
```diff
@@ -31,20 +31,37 @@ public void PostConfigure(string? name, DashboardOptions options)
         {
             options.Otlp.GrpcEndpointUrl = otlpGrpcUrl;
         }
+        else if (_configuration[DashboardConfigNames.Legacy.DashboardOtlpGrpcUrlName.ConfigKey] is { Length: > 0 } legacyOtlpGrpcUrl)
```
eerhardt, line 34: Can these use the new IConfigurationExtensions methods that checks both?

src/Aspire.ProjectTemplates/templates/aspire-apphost/8.2/Properties/launchSettings.json
DamianEdwards: didn't mean to, will roll that back.

src/Aspire.Dashboard/Configuration/PostConfigureDashboardOptions.cs
JamesNK: I don't think we'd want to make falling back on an empty string the default behavior. I worry it would cause some issue somewhere where a string could be empty.

But in this situation, what about adding an optional parameterto the new methods? Should be a minor change and would make the code here cleaner.

DamianEdwards: Yep sounds good.

src/Shared/IConfigurationExtensions.cs
```diff
@@ -8,6 +8,35 @@ namespace Aspire;
...
+        return defaultValue;
+    }
+
+    public static bool? GetBool(this IConfiguration configuration, string primaryKey, string secondaryKey)
+    {
+        var value = configuration.GetBool(primaryKey) ?? configuration.GetBool(secondaryKey);
+        return value;
+    }
+    
+    public static string? GetString(this IConfiguration configuration, string primaryKey, string secondaryKey)
```
JamesNK, line 34: > public static string? GetString(this IConfiguration configuration, string primaryKey, string secondaryKey, bool fallbackOnEmpty = false)

?

src/Aspire.Dashboard/Configuration/PostConfigureDashboardOptions.cs
DamianEdwards: OK done.

src/Shared/IConfigurationExtensions.cs
DamianEdwards: Went with that approach, thanks.

src/Shared/IConfigurationExtensions.cs
```diff
@@ -8,6 +8,46 @@ namespace Aspire;
...
+    public static bool? GetBool(this IConfiguration configuration, string primaryKey, string secondaryKey)
+    {
+        var value = configuration.GetBool(primaryKey) ?? configuration.GetBool(secondaryKey);
+        return value;
+    }
+
+    public static string? GetString(this IConfiguration configuration, string primaryKey, string secondaryKey, bool fallbackOnEmpty = false)
+    {
+        var primaryValue = configuration.GetValue(typeof(string), primaryKey, null);
+        if (primaryValue is not null && !fallbackOnEmpty || primaryValue is string { Length: > 1 })
```
JamesNK, line 37: This was skipping strings with 1 character.
```suggestion
        if (primaryValue is not null && !fallbackOnEmpty || primaryValue is string { Length: > 0 })
```

src/Shared/IConfigurationExtensions.cs
```diff
@@ -8,6 +8,46 @@ namespace Aspire;
...
+        {
+            return (string)primaryValue;
+        }
+
+        var secondaryValue = configuration.GetValue(typeof(string), secondaryKey, null);
+        if (secondaryValue is not null && !fallbackOnEmpty || secondaryValue is string { Length: > 1 })
```
JamesNK, line 43:
```suggestion
        if (secondaryValue is not null && !fallbackOnEmpty || secondaryValue is string { Length: > 0 })
```

src/Shared/IConfigurationExtensions.cs
DamianEdwards: doh, thanks

src/Aspire.Hosting/DistributedApplication.cs
```diff
@@ -194,12 +194,12 @@ public static IDistributedApplicationBuilder CreateBuilder(DistributedApplicatio
 
     private static void WaitForDebugger()
     {
-        if (Environment.GetEnvironmentVariable("ASPIRE_WAIT_FOR_DEBUGGER") == "true")
+        if (Environment.GetEnvironmentVariable(KnownConfigNames.WaitForDebugger) == "true")
```
JamesNK, line 197: Is there a reason Environment is used directly in a couple of places instead of IConfiguration? cc @mitchdenny

src/Aspire.Hosting/DistributedApplicationBuilder.cs
```diff
@@ -293,10 +293,15 @@ public DistributedApplicationBuilder(DistributedApplicationOptions options)
...
 
                     // Determine the resource service API key.
-                    if (_innerBuilder.Configuration[KnownConfigNames.DashboardResourceServiceClientApiKey] is not { Length: > 0 } apiKey)
+                    var apiKey = _innerBuilder.Configuration[KnownConfigNames.DashboardResourceServiceClientApiKey];
+                    if (apiKey is not { Length: > 0 })
                     {
-                        // No API key was specified in configuration, so generate one.
-                        apiKey = TokenGenerator.GenerateToken();
+                        apiKey = _innerBuilder.Configuration[KnownConfigNames.Legacy.DashboardResourceServiceClientApiKey];
+                        if (apiKey is not { Length: > 0 })
```
JamesNK, line 300: Could this use GetString with the empty fallback now?

src/Shared/DashboardConfigNames.cs
```diff
@@ -6,12 +6,13 @@ namespace Aspire.Hosting;
 internal static class DashboardConfigNames
 {
     public static readonly ConfigName DashboardFrontendUrlName = new("ASPNETCORE_URLS");
```
JamesNK, line 8:
```suggestion
    public static readonly ConfigName DashboardFrontendUrlName = new(KnownConfigNames.AspNetCoreUrls);
```

src/Aspire.Cli/DotNetCliRunner.cs
```diff
@@ -107,14 +108,14 @@ public async Task<int> ExecuteAsync(string[] args, IDictionary<string, string>?
         var socketPath = GetBackchannelSocketPath();
         if (backchannelCompletionSource is not null)
         {
-            startInfo.EnvironmentVariables["ASPIRE_BACKCHANNEL_PATH"] = socketPath;
+            startInfo.EnvironmentVariables[KnownConfigNames.UnixSocketPath] = socketPath;
```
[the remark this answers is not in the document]
JamesNK: Edit: I see these aren't strictly related to just CLI. Still, structure of names is something to keep in mind. I think it's good that all dashboard parameters have a consistent prefix.

JamesNK (approved):
:shipit:

PR #8352, opened 2025-03-27
Remove IKeyVaultResource.IdOutputReference from public API

eerhardt:
This property isn't used and can be removed.


davidfowl (approved):

davidfowl:
Well.. ok

